PR #42, opened 2019-02-14
Update README.md

rootsongjc:
- Replace the image with relative path
- Add code syntax


leecalcote:
Excellent. Thank you, @rootsongjc!

PR #48, opened 2019-03-01
Add ErrorDetails message

venilnoronha:
This adds a new message type named `ErrorDetails` which is to be used for capturing errors in a structured manner. The first use-case for this type is in the `ApplyRuleResponse` type.

### DIFF
--- a/meshes/meshops.pb.go
+++ b/meshes/meshops.pb.go
@@ -3,13 +3,12 @@
 
 package meshes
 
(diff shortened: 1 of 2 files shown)


girishranganathan:
This is addressed as part of the event updates that were added in a few weeks back. We can close this one.

venilnoronha:
Cool, thanks.

PR #89, opened 2019-04-12
updating install script + readme

girishranganathan:
- storing docker compose file in the home folder
- changed ctl file name
- removed run instructions to point to l5io/meshery

### DIFF
--- a/meshery
+++ b/meshery
@@ -30,7 +30,9 @@ print_help() {
    [start]     runs 'docker-compose' to spin up Meshery and each of each service mesh adapters.
    [stop]      stops all Meshery containers, removes their instances and prunes their connected volumes.
    [update]    polls Docker Hub for new Meshery container images and pulls if new version are available.
-   [logs]      prints history of Meshery's container logs and begins tailing them."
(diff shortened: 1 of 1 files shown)


meshery
```diff
@@ -72,35 +83,42 @@ case $1 in
...
+        else
+            docker-compose -f $dockerComposeFile logs -f
+        fi
     ;;
     update)
-        echo "Updating Meshery..."
-        docker-compose pull
-        echo "Meshery is now up-to-date."
+        if [ ! -f $dockerComposeFile ]; then
+            echo "Unable to download Meshery's docker-compose.yaml file. Please try again later."
```
leecalcote: Potentially, confusing message given that, yes, the user is downloading docker-compose.yaml, but doesn't know that we're renaming it to meshery.yaml necessarily. Maybe best to rename docker-compose.yaml in our repo to meshery.yaml.

girishranganathan: We do need folks who can cloned the repo to be able to run `docker-compose up -d`. If we change the file name then they will have to add a `-f <file name>`. For now I am leaving it as it is but have updated the error message to just say "config".

PR #90, opened 2019-04-13
Labels

leecalcote:
@girishranganathan is there a list of material-ui icons available to import? How do you identify the specific name of an icon to import, like this one? - https://fontawesome.com/icons/external-link-alt?style=solid

### DIFF
--- a/ui/components/Navigator.js
+++ b/ui/components/Navigator.js
@@ -12,7 +12,7 @@ import SettingsInputComponentIcon from '@material-ui/icons/SettingsInputComponen
 import CollectionsIcon from '@material-ui/icons/Collections';
 import ShowChartIcon from '@material-ui/icons/ShowChart';
 import TimerIcon from '@material-ui/icons/Timer';
-import LinkIcon from '@material-ui/icons/Link';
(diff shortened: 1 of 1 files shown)


girishranganathan:
@leecalcote, https://material.io/tools/icons/?style=baseline, https://material-ui.com/style/icons/ - are some sources I have been using.

leecalcote:
@girishranganathan I'm missing the explicit listing of the reference to use for importing a specific icon. An example being https://material.io/tools/icons/?icon=exit_to_app&style=baseline, which is close enough to https://fontawesome.com/icons/external-link-alt?style=solid that it will work to denote a link that exits the user from the site. In either case, I'm not sure where the reference to their _actual_ name can be found (e.g. does https://fontawesome.com/icons/external-link-alt?style=solid convert to __ExternalLinkAlt__? and https://material.io/tools/icons/?icon=exit_to_app&style=baseline convert to __ExitToApp__?)

girishranganathan:
@leecalcote, you are right to the most part. Though font awesome and material ui don't have to follow the same name/convention. The name you find on the material ui site usually maps to a camel cased version of their react component. I usually check in the local npm package if such a thing exists which validates it.

PR #234, opened 2019-08-30
Feature/sub menus

girishranganathan:


.gitignore
```diff
@@ -40,3 +40,5 @@ log.txt
 **/_site
 **/.sass-cache
 **/.jekyll-metadata
+
+mesheryctl/.meshery/
```
leecalcote: @girishranganathan can we move to line 3 or so under #Meshery?

Makefile
```diff
@@ -38,6 +38,7 @@ run-local:
 	SAAS_BASE_URL="https://meshery.layer5.io" \
 	PORT=9081 \
 	DEBUG=true \
+	ADAPTER_URLS="mesherylocal.layer5.io:10000 mesherylocal.layer5.io:10001" \
```
leecalcote: Are we missing 10002? What is ADAPTER_URLS used for?

.gitignore
girishranganathan: done

Makefile
girishranganathan: Purely for local dev. This is different from the one used in docker-compose.

girishranganathan:
@leecalcote can you please fix the DCO issue in your commit?

Issue #91, opened 2019-04-14
Ensure meshery containers are running before `open localhost`


leecalcote:
**Describe the bug**
The current `meshery start` cmd does not wait for the meshery application to be up before launching the user's browser in an attempt access meshery. Users experience a 404 message. 

**To Reproduce**
1. Execute `meshery start`
2. Find that the meshery app is unavailable in your freshly opened browser.

**Expected behavior**
Users expect to be able to interact with the meshery application when the page is first loaded in their browser.

leecalcote:
@sakomws here's one for `mesheryctl`

PR #240, opened 2019-09-02
#91 check for meshery_1 container running before opening browser

anandrkskd:


mesheryctl/cmd/start.go
```diff
@@ -43,14 +45,68 @@ var startCmd = &cobra.Command{
...
+	count := 0
+	for _, x := range s {
+		if strings.TrimRight(x, " ") == "meshery_meshery_1" {
+			count++
+		}
+	}
+	if count < 1 && atempt < 1 {
+		//checks for meshery_meshery_1 container running or not if not then restarting meshery
+		fmt.Println("service cannot be started, attempting to restart the service!")
+		exec.Command("mesheryctl", "stop")
```
leecalcote: When invoking another `mesheryctl` command, can we do so by calling the Cobra command / function directly? (e.g. `stopCmd.Run()`)?

anandrkskd: okay Understood

mesheryctl/cmd/start.go
```diff
@@ -43,14 +45,68 @@ var startCmd = &cobra.Command{
 
 		fmt.Println("Starting Meshery...")
 		start := exec.Command("docker-compose", "-f", dockerComposeFile, "up", "-d")
+		time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
girishranganathan: This can be done in 2 parts:
- checking if the container is running. For this, I would strongly recommend using the docker Go client and NOT use `docker ps` or something along those lines.
- check if the port 9081 is open and is accepting connections by using the default http client to access `http://localhost:9081`

mesheryctl/cmd/start.go
```diff
@@ -43,14 +45,68 @@ var startCmd = &cobra.Command{
...
+		//testing purpose
+		//exec.Command("docker", "kill", "meshery_meshery_1")
 		start.Stdout = &out
```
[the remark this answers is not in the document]
girishranganathan: It looks like it is/has been.
I would recommend staying away from `fmt.Print` or any of its functions to print/log anything and leveraging a logging library like `logrus` (OR) at least the default `log` package.

mesheryctl/cmd/start.go
```diff
@@ -43,14 +45,68 @@ var startCmd = &cobra.Command{
...
+		check(s, atempt)
+	} else if atempt >= 1 {
+		//showing logs if restart doesn't work
+		fmt.Println("service cannot be started. Showing meshery logs")
+		//\
+		fmt.Println("Starting Meshery logging . . .")
+		cmd_log := exec.Command("docker-compose", "-f", dockerComposeFile, "logs", "-f")
+		cmdReader, err := cmd_log.StdoutPipe()
+		if err != nil {
+			log.Fatal(err)
```
girishranganathan: let us try to be consistent with using `log` or `logrus` throughout

mesheryctl/cmd/start.go
```diff
@@ -43,14 +45,68 @@ var startCmd = &cobra.Command{
 
 		fmt.Println("Starting Meshery...")
 		start := exec.Command("docker-compose", "-f", dockerComposeFile, "up", "-d")
+		time.Sleep(2 * time.Second)
+		//testing purpose
```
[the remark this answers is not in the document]
girishranganathan: We would like to follow the Go testing convention described [here](https://golang.org/pkg/testing/).

mesheryctl/cmd/start.go
leecalcote: @girishranganathan let's be explicit. We're standardizing on `logrus`, correct?

girishranganathan: We have been using `logrus` throughout Meshery. Not in MesheryCTL. It would definitely better to use one across the board.
Lets go with `logrus`.

anandrkskd: okay understood

leecalcote:
@anandrkskd we'd do well to leverage Docker's restart policies, if we can - https://docs.docker.com/config/containers/start-containers-automatically/

mesheryctl/cmd/start.go
```diff
@@ -42,33 +45,75 @@ var startCmd = &cobra.Command{
...
 
-		fmt.Println("Starting Meshery...")
+		log.Info("Starting Meshery...")
 		start := exec.Command("docker-compose", "-f", dockerComposeFile, "up", "-d")
 		start.Stdout = &out
 		start.Stderr = &stderr
 
 		if err := start.Run(); err != nil {
-			fmt.Println(stderr.String())
+			log.Println(stderr.String())
```
girishranganathan: can you please change Println() to Error()?

mesheryctl/cmd/start.go
```diff
@@ -42,33 +45,75 @@ var startCmd = &cobra.Command{
...
 
-		fmt.Println("Opening Meshery in your broswer. If Meshery does not open, please point your browser to http://localhost:9081 to access Meshery.")
-		ostype, err := exec.Command("uname", "-s").Output()
+		checkFlag := 0 //flag to check
+
+		//connection to docker-client
+		cli, err := client.NewEnvClient()
 		if err != nil {
-			log.Fatal("[WARNING] Unable to detect OS type. Warning message: \n", err)
+			panic(err)
```
girishranganathan: May be we should follow a pattern, either use panic or log.Fatal? Lets stick with log.Fatal

mesheryctl/cmd/start.go
```diff
@@ -42,33 +45,75 @@ var startCmd = &cobra.Command{
...
-			// Meshery running on Linux host
-			exec.Command("xdg-open", url).Start()
-		} else {
+		//check for container meshery_meshery_1 running status
+		for _, container := range containers {
+			if "/meshery_meshery_1" == container.Names[0] {
+				log.Info("Opening Meshery in your broswer. If Meshery does not open, please point your browser to http://localhost:9081 to access Meshery.")
+				ostype, err := exec.Command("uname", "-s").Output()
+				if err != nil {
+					log.Fatal("[WARNING] Unable to detect OS type. Warning message: \n", err)
```
girishranganathan: We don't want the "[WARNING]", bcoz that is going to confuse.

mesheryctl/cmd/start.go
```diff
@@ -42,33 +45,75 @@ var startCmd = &cobra.Command{
...
+			}
+		}
 
-			// Asssume Meshery running on MacOS host
-			exec.Command("open", url).Start()
+		//if meshery_meshery_1 failed to start showing logs
+		//code for logs
+		if checkFlag == 1 {
+			log.Info("Starting Meshery logging . . .")
+			cmdlog := exec.Command("docker-compose", "-f", dockerComposeFile, "logs", "-f")
```
girishranganathan: we don't need a "-f" right after docker-compose, I think.

mesheryctl/cmd/status.go
```diff
@@ -28,12 +27,12 @@ var statusCmd = &cobra.Command{
 	Short: "Check Meshery status",
 	Long:  `Check status of Meshery and Meshery adapters.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("Meshery containers status . . .\n")
+		log.Info("Meshery containers status . . .\n")
 		out, err := exec.Command("docker-compose", "-f", dockerComposeFile, "ps").Output()
 		if err != nil {
 			log.Fatal("[ERROR] Please, install docker-compose. The error message: \n", err)
```
girishranganathan: I think it will good to follow a pattern, whether or not to include [ERROR] in fatal msgs. . . lets take it out.

mesheryctl/cmd/status.go
```diff
@@ -28,12 +27,12 @@ var statusCmd = &cobra.Command{
...
 		}
-		fmt.Printf("%s\n", out)
+		log.Printf("%s\n", out)
```
girishranganathan: change Printf to Infof? and we probably don't want \n

Logrus does a good job with printing each log in separate lines. Can you please also get rid of all the "\n" in print statements in mesheryctl?

anandrkskd: Yeah thanks for the feedback, and consider it done.
Thanks @girishranganathan.

mesheryctl/cmd/start.go
anandrkskd: I built mesheryctl after removing -f for the command, and it's not wokring. 

INFO[0001] Starting Meshery...                          
ERRO[0002] No such command: .meshery/meshery.yaml

Commands:
  build              Build or rebuild services
  bundle             Generate a Docker bundle from the Compose file
  config             Validate and view the 
...
and other docker-compose command.
So I think  going forward with -f would be better.

leecalcote:
Fixes #91

PR #309, opened 2019-09-30
new lint tool

girishranganathan:
- testing a new lint tool
- updated prometheus page to load a default config and also will only show the panels and button when preconditions are met

Signed-off-by: Girish Ranganathan <girish.rranganathan@gmail.com>

### DIFF
--- a/revive-lint.toml
+++ b/revive-lint.toml
@@ -2,7 +2,7 @@ ignoreGeneratedHeader = false
 severity = "warning"
 confidence = 0.8
 errorCode = 1
-warningCode = 0
+warningCode = 1
 
 [rule.blank-imports]
 [rule.context-as-argument]


revive-lint.toml
```diff
@@ -0,0 +1,25 @@
...
+confidence = 0.8
+errorCode = 0
+warningCode = 0
+
+[rule.blank-imports]
+[rule.context-as-argument]
+[rule.context-keys-type]
+[rule.dot-imports]
+[rule.error-return]
+#[rule.error-strings]
```
leecalcote: Too bad it doesn't get any more granular, but this is probably nicer than the other way around.

girishranganathan: yup

Issue #310, opened 2019-10-01
Update CONTRIBUTING.md in each repo 


subhamkrai:
update contributing.md in each repo which redirects to main meshery repo’s CONTRIBUTING.md with an updated youtube link.

leecalcote:
Fixes #310

leecalcote:
Thanks, @subhamkrai .

PR #311, opened 2019-10-01
#310 updated CONTRIBUTING.md

subhamkrai:
Signed-off-by: subhamkrai <subhamkumarrai03@gmail.com>


CONTRIBUTING.md
```diff
@@ -1,3 +1,7 @@
+# <a name="contributing">Contributing to meshery</a>
```
leecalcote: @subhamkrai given that these *are* the master contribution instructions, I don't know that we need this section here, but rather need this section in the adapter repos (in their contributing.mds).

PR #334, opened 2019-10-06
resolved link broken

### LINKED ISSUE #329: link broken
https://meshery.layer5.io/docs/installation.md    does not works

ashiscs:


leecalcote:
@ashiscs thanks. I think there's some confusion as to what a 404 error and a custom 404 page is. Read this article - https://support.google.com/webmasters/answer/93641?hl=en (or search for others) to better understand why we have this custom 404 page and we wouldn't want to change it's title. I think that including links to prominent resources can be helpful and that we may want to include some in our current 404, so there may be something to do here, but we wouldn't want to holistically change our 404 page into one that assumes that people were looking for the installation doc as they may have been looking for any number of other pages.

leecalcote:
@ashiscs you may want to tackle this issue instead - #345

leecalcote:
@ashiscs, will you confer with @gunjanmimo in terms of whether to close this PR? I think he was able to knock out this issue.

ashiscs:
Yes sir,this issue is fixed.

Issue #371, opened 2019-10-15
update meshery deployments to support k8s 1.16+


chrisevett:
**Enhancement Description**
meshery deployments need to update to have the apps/v1 apiversion in order to run on k8s 1.16

I will have a PR open on this shortly

leecalcote:
Closing this issue as @chrisevett knocked it out of the park. Thank you so much, @chrisevett!

PR #372, opened 2019-10-15
update apiversion for deployments to apps/v1 to support k8s 1.16

chrisevett:
Fix for #371 .


leecalcote:
@chrisevett thank you!

chrisevett:
> 
> 
> @chrisevett thx a lot. PR looks great.
> One request, can you please move these into the parent folder? lets get the other yaml files replaced.

Is that going to be backwards compatible?

leecalcote:
@chrisevett you're right, we'll lose compatibility here. This may be a secondary concern for the time being given the community's lack of bandwidth. What do you recommend as a version compatibility strategy? - a version per folder?

PR #381, opened 2019-10-22
Feature/async lt

girishranganathan:


leecalcote:
Load testing is now run asynchronously. This is very cool, @girishranganathan! 

There are a few extra notifications (in blue) that probably aren't necessary, but this is a great feature.

Issue #389, opened 2019-10-23
Increase verbosity of `mesheryctl update` to show new docker image pull


leecalcote:
**Issue Description**
When a user runs `mesheryctl update` and a new Docker image is pulled to the user's local machine, `meshery` (bash script) would show the user that a new image was being pulled. `mesheryctl` does not. It leaves the user to wonder what is going on, like so:

```
$ mesheryctl update

INFO[0000] Updating Meshery now...
INFO[0008] Meshery is now up-to-date
```

**Enhancement Description**
Increase the verbosity of the command line output from the [mesheryctl](https://github.com/layer5io/meshery/tree/master/mesheryctl) `update` command to show the user the live downloading of new images and the other `update` activities (if present).

leecalcote:
Nice work, @anandrkskd. :)

PR #398, opened 2019-10-24
Feature/anandrkskd/389

anandrkskd:
#389  updated with Increased verbosity of `mesheryctl update` to show new docker image pull


downloadMesheryCandidate.sh
```diff
@@ -0,0 +1,52 @@
...
+#echo ${A:1}
+
+
+#NAME="mesheryctl-$MESHERY_VERSION"
+URL="https://github.com/layer5io/meshery/releases/download/${MESHERY_VERSION}/mesheryctl_${MESHERY_VERSION:1}_${OSEXT}_${OSARCHITECTURE}.zip"
+#URL="https://github.com/layer5io/meshery/releases/download/v0.1.6/mesheryctl_0.1.6_Darwin_i386.zip"
+#printf "Downloading %s from %s ..." "$NAME" "$URL"
+#curl -L "$URL" | tar xz
+
+curl -L ${URL} -o /tmp/meshery.zip
```
leecalcote: @anandrkskd can we use the user's current working directory as the download location?  `/tmp` isn't available on MacOS by default nor on Windows.

downloadMesheryCandidate.sh
```diff
@@ -0,0 +1,52 @@
...
+unzip /tmp/meshery.zip
```
leecalcote: @anandrkskd what do we do in the even the user doesn't have `unzip`?

downloadMesheryCandidate.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/sh
```
leecalcote: @anandrkskd let's use`#!/usr/bin/env bash` to improve portability

leecalcote:
@anandrkskd will you breakout `mesheryctl/cmd/update.go` into a separate PR?

leecalcote:
@anandrkskd I like the branch naming scheme you’ve used here. Thumbs up.

anandrkskd:
SIr @leecalcote , I have removed the bash script for now, and this PR will be used for increased verbosity of mesheryctl update.

downloadMesheryCandidate.sh
anandrkskd: Sir @leecalcote  we are using zip to compress so I'm not sure if tar would work!

PR #480, opened 2019-11-09
Incorporate make docs command from Docs Makefile

raghavaggarwal99:
make docs command


Makefile
```diff
@@ -73,3 +73,15 @@ run-ui-dev:
 # Builds the user interface on your local machine.
 build-ui:
 	cd ui; npm run build && npm run export; cd ..
+	
+#Incorporating Make docs commands from the Docs Makefile	
+jekyll=bundle exec jekyll
+
+site:
```
leecalcote: @raghavaggarwal99 can we change from `site` to `docs` or to `docs-site`?

Makefile
```diff
@@ -73,3 +73,15 @@ run-ui-dev:
...
+	$(jekyll) serve --drafts --livereload
+
+build:
+	$(jekyll) build --drafts --livereload
+
+docker:
```
leecalcote: @raghavaggarwal99 we'll have a conflict of `docker` in line 9. We may need to change line 86 to `docker-docs`

Issue #506, opened 2019-11-25
Add Meshery adapter for CPX to list of adapters


leecalcote:
**Current State:**
The list of adapters is missing an entry for CPX.

The [CPX adapter](https://github.com/layer5io/meshery-cpx) should be listed as an alpha adapter with link to its the [adapter section of the Meshery documentation](https://meshery.layer5.io/docs/installation/adapters).

---
**Contributor Resources**
- [Meshery documentation site](https://meshery.layer5.io/docs/)
- [Meshery documentation source](https://github.com/layer5io/meshery/tree/master/docs)
- [Instructions for contributing to documentation](https://github.com/layer5io/meshery/blob/master/CONTRIBUTING.md#documentation-contribution-flow)

leecalcote:
@dheerajng you might consider grabbing this one.

dheerajng:
Created a [PR](https://github.com/layer5io/meshery/pull/517) for the same.

leecalcote:
Nice work, @dheerajng!

PR #517, opened 2019-12-13
#506 Updating list of adapter with Meshery Adapter for CPX.

dheerajng:
Signed-off-by: dheerajng <dhiraj.gedam@citrix.com>


leecalcote:
Fixes #506

Issue #530, opened 2019-12-21
Hide the adapter link in the nav menu if there is only one adapter for a type


girishranganathan:
**Current Behavior**
The navigation menu on the left of the page in Meshery contains a section for each adapter type. Under the specific types, the connected adapters are displayed as a list.

**Desired Behavior**
If there is only one adapter for a type (like istio, linkerd), then hide the list item and make the adapter type itself clickable and land on the adapter page.

PR #534, opened 2019-12-22
Hide adapter in nav

Nikhil-Ladha:
Referenced issue: #530 

### DIFF
--- a/ui/components/Navigator.js
+++ b/ui/components/Navigator.js
@@ -292,6 +292,22 @@ class Navigator extends React.Component {
       });
     }
 
+    updateAdaptersLink() {
+      const self = this;
+      categories.forEach((cat, ind) => {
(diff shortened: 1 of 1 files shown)


leecalcote:
@Nikhil-Ladha I believe that this is an improvement, however, you will want to confirm that the individual adapters are only listed as a submenu item when there is more than one adapter of the same type present. e.g. when two Istio adapters are running and configured in Meshery's settings, then show two sub-menu items, however, if only one Istio adapter is running and configured in Meshery, then do not show any sub-menu items; only show "Istio"

Nikhil-Ladha:
@leecalcote sir, think it will work the way you are saying, but still you can check it once if it's working or not. If not then it would be helpful if you could share me the screenshot showing the same.

PR #548, opened 2020-01-01
Docs/subhamkrai/#545

subhamkrai:


subhamkrai:
> @subhamkrai it's a bit deceiving, but the fix that is needed here is to either update `published: false` to `published: true` or to remove the ["performance" entry in the toc.yml](https://github.com/layer5io/meshery/blob/master/docs/_data/toc.yml#L31) until the performance page has been written.

Okay I'm on it.

Issue #554, opened 2020-01-04
UI Infinite loop caused when adding a local adapter 127.0.0.1:10000


girishranganathan:
**Description**
When we tried to add a local istio adapter at 127.0.0.1:10000, the UI would crash and take us on an infinite loop.

**Expected Behavior**
Adapter should added fine without any issues.

**Causes**
Recent changes to the nav menu updates have a missed null/undefined check.

**Fix**
Adding extra undefined checks helped mitigate this issue.

girishranganathan:
@raghavaggarwal99 helped us find this issue and also resolve it. He will create a PR with the fix.

PR #556, opened 2020-01-04
Fix: UI Infinite loop caused when adding a local adapter 127.0.0.1:10000

raghavaggarwal99:
Signed-off-by: raghavaggarwal99 <ragha170103048@iitg.ac.in>


raghavaggarwal99:
to the issue #554

raghavaggarwal99:
@leecalcote  done!

Issue #461, opened 2019-11-04
Merge MeshConfigComponent.js and MeshOutClusterConfigComponent.js to remove duplicate code


girishranganathan:
**Description**
MeshConfigComponent.js was replicated to create a new component MeshOutClusterConfigComponent.js to make the settings page work with the new UX for the page. Both the components now have a lot of duplicate code.

**Expected Behavior**
We should just probably update MeshConfigComponent.js in a way that it can be made to behave in different ways based on the different props given to it. 
GrafanaCustomChart component is an example. I am pretty sure we have a few more examples of such components.

leecalcote:
Completed. ✔️

PR #560, opened 2020-01-05
Removed duplicate code and merged them

Nikhil-Ladha:
#461 Fixed!


leecalcote:
@Nikhil-Ladha will you look into the Meshery UI build check error?

Nikhil-Ladha:
@leecalcote in the build process the issue is that the build check is looking for `MeshConfigComponent.js` file, but now it has been removed in this pr and merged with `MeshOutClusterConfig.js` file. So, i guess the build check process has to be modified.

Nikhil-Ladha:
Or another thing i may try is that instead of changing both the file to `MeshOutClusterConfig.js` i will merge them in `MeshConfigComponent.js` and then make the PR?

Nikhil-Ladha:
@leecalcote as expected that was only the issue the missing file `MeshConfigComponent,js`. Now it's good!

Issue #449, opened 2019-11-01
mesheryctl: Add support for `mesheryctl update` to upgrade mesheryctl itself


leecalcote:
**Current Behavior**
Currently `mesheryctl update` downloads new versions of the Meshery containers as described in the [Meshery Installation & Packaging](https://docs.google.com/document/d/1inncJ56SKEzIMIUikzj3XdCWF4d0xdtzOcq6J4GQYz0/edit#heading=h.hxpazjese5dc) document.

**Desired Behavior**
Ideally, `mesheryctl update` will also upgrade the binary itself (displacing itself with the latest version of the binary).

anandrkskd:
I will be working on this issue.

anandrkskd:
In the new updateCmd code I am facing a permission issue [My Working Branch](https://github.com/anandrkskd/meshery/tree/bug/anandrkskd/449)
**behaviour**

- update command detect the client os and arch the downloads the suitable latest mesheryctl zip for the system.
- downloaded file is stored in users **.meshery** directory in HOME of user,
- files are then unzipped to "/usr/local/bin/" dir.

leecalcote:
@NupurThakur27 have a look at `gh` and how it provides provides this functionality. 

```
A new release of gh is available: 0.7.0 → v0.8.0
https://github.com/cli/cli/releases/tag/v0.8.0
```

https://github.com/cli/cli/blob/master/update/update.go

nitishm:
@leecalcote want me to start working this?

leecalcote:
@nitishm that'd be great, yes.

KarenEfereyan:
@nitishm can I work with you? I really want to learn how this is done. It seems more than just a documentation issue. More like code involved.

anirudhjain75:
Is this up for grabs ? @leecalcote

anirudhjain75:
Regarding having mesheryctl show latest version as message / notification to user on invoking commands. gh makes uses of github release for cli repo and manages version using github api. You can try executing GET req on `https://api.github.com/repos/meshery/meshery/releases/latest`. This API results the latest release results of meshery repo. The issue with using the same approach as gh is that `mesheryctl` doesn't have its own repo which makes version management a bit tricky.

anirudhjain75:
`mesheryctl` for now seems to be using the meshery version as it version tag. so things should be fine with making an implementation to check for updates similar to gh cli.

anirudhjain75:
[...]

PR #563, opened 2020-01-05
mesheryctl updateCmd upgradation

anandrkskd:
#449 This new feature now solve the issue of auto upgradation of mesheryctl binary.
Now mesheryctl can download the new mesheryctl binary on its own with **mesheryctl update**  command.


mesheryctl/cmd/update.go
```diff
@@ -30,6 +37,74 @@ var updateCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
```
leecalcote: @anandrkskd let's update the short and long descriptions to include the fact the `mesheryctl` will be updated, too. 

See the text on this page for inspiration - https://meshery.layer5.io/docs/guides/mesheryctl#meshery-lifecycle-management

mesheryctl/cmd/update.go
```diff
@@ -30,6 +37,74 @@ var updateCmd = &cobra.Command{
...
+			}
+
+			version := fmt.Sprint("%v", num)
+			downloadMesheryURL = fmt.Sprintf(downloadMesheryURL+"/%v/mesheryctl_%v_%v_%v.zip", num, version[3:], os, arch)
+		}
+
+		//If sudo is needed to update mesheryctl
+		user, err := user.Current()
+		if err != nil {
+			panic(err)
```
leecalcote: @anandrkskd do we want to panic or use Log.Fatal and exit?

mesheryctl/cmd/update.go
```diff
@@ -30,6 +37,74 @@ var updateCmd = &cobra.Command{
...
+		}
+		if user.Username == "root" {
```
leecalcote: At this point, why do we need to handle the root user use case any differently than an unprivileged user?

mesheryctl/cmd/update.go
```diff
@@ -30,6 +37,74 @@ var updateCmd = &cobra.Command{
...
+			if _, err := os.Stat(mesheryFolder); os.IsNotExist(err) {
+				_ = os.Mkdir(mesheryFolder, 0777)
+			}
+		}
```
leecalcote: We're not handling the case of an non-administrative user, are we? Seems like we would do so here, if so.

mesheryctl/cmd/update.go
```diff
@@ -30,6 +37,74 @@ var updateCmd = &cobra.Command{
...
+
+		//download mehseryctl binary
+		if err := downloadFile(mesherybinary, downloadMesheryURL); err != nil {
+			log.Fatal("update cmd: ", err)
```
leecalcote: Instead of "update cmd:", can we use something more descriptive, like "Unable to download mesheryctl release bundle:"?

mesheryctl/cmd/update.go
```diff
@@ -30,6 +37,74 @@ var updateCmd = &cobra.Command{
...
+		}
+
+		//Unzip mesheryctl binary
+		uz := unzip.New(mesherybinary, "/usr/local/bin/")
+		err = uz.Extract()
+		if err != nil {
+			log.Warn(err)
+			log.Warn("skiping mesheryctl binary update!")
```
leecalcote: "skiping" is misspelled.
If we're skipping due to a permission error, let's include this warning. Otherwise, if we're skipping because the version # is the same as what the user already has on their system, then no need for a Warning, just an informational log will suffice.

Let's check to see if the current `mesheryctl` is the same as the remote (latest) `mesheryctl` version and provide an informational log stating that we're skipping, because the user already has the latest `mesheryctl`.

In order to handle the issue of lack of permissions, let's prompt the user at this point for their sudo passwd and re-execute this unzip under that privileged context.

mesheryctl/cmd/update.go
```diff
@@ -30,6 +37,74 @@ var updateCmd = &cobra.Command{
...
+			//log.Warn("Unable to update mesheryctl: permission denied")
+			//log.Println("Hint: try using sudo!")
+		}
+
+		//update docker image
 		if _, err := os.Stat(dockerComposeFile); os.IsNotExist(err) {
 			if err := downloadFile(dockerComposeFile, fileURL); err != nil {
 				log.Fatal("update cmd: ", err)
```
leecalcote: Can we improve this Fatal log message? "Could not update Meshery containers:"

mesheryctl/cmd/update.go
anandrkskd: Sir that portion of code solves the issue for use of sudo to use update command(mainly for linux user's)

mesheryctl/cmd/update.go
anandrkskd: Understood Sir

mesheryctl/cmd/update.go
anandrkskd: Understood!

leecalcote:
@girishranganathan this is ready for a second round of review given @anandrkskd's updates.

mesheryctl/cmd/update.go
```diff
@@ -15,8 +15,15 @@
 package cmd
 
 import (
+	"encoding/json"
+	"fmt"
+	"github.com/artdarek/go-unzip"
```
girishranganathan: @anandrkskd PR looks great 👏 
One request: can we please move away from this go-unzip library and use the [archive/zip](https://golang.org/pkg/archive/zip/) package.
Here is an article with an example: https://golangcode.com/unzip-files-in-go/

anandrkskd: Understood!

leecalcote:
Thanks for these updates, @anandrkskd. @girishranganathan have a final look?

leecalcote:
@girishranganathan what's the verdict here?

girishranganathan:
@anandrkskd, we totally appreciate your work here. 
But, I would say lets park this code here for now. For Mac and Linux, `brew` does provide a good way to roll back, stick to a specific version, and do upgrades in a nice way. For `Windows`, scoop might be a better route for now.
Lets revisit this branch when we decide to proceed to do upgrade this way.

PR #527, opened 2019-12-20
Signedoff: raghavaggarwal99<ragha170103048@iitg.ac.in>

raghavaggarwal99:
Signed-off-by: raghavaggarwal99 <ragha170103048@iitg.ac.in>

Issue #394, opened 2019-10-23
Update docs site footer with current social buttons


leecalcote:
**Enhancement Description**
Update [the current docs footer](https://github.com/layer5io/meshery/blob/master/docs/_includes/footer.html) to use [the same social buttons](https://github.com/layer5io/layer5/blob/master/_includes/partials/hori-social.html) as those used on layer5.io as shown [in this screencast](https://youtu.be/FZ4Mi68CsGM).

anusha5695:
Hi @leecalcote : Is this still up for grabs?

leecalcote:
Hi @anusha5695, I missed your question earlier. Yes, help here would be very nice. @gunjanmimo has given this a try in PR #432, which is good. However, hasn't quite hit the mark yet in terms of the original ask to use [the same social buttons](https://github.com/layer5io/layer5/blob/master/_includes/partials/hori-social.html) as those used on layer5.io as shown [in this screencast](https://youtu.be/FZ4Mi68CsGM). 

You're help here would be much appreciated!

schillingjerrid:
@leecalcote I can work on this if you want. Do we want the tooltips? They won't really work going below the buttons like they do on the layer5.io site but I can put them above the buttons.

leecalcote:
@schillingjerrid yes, sir. Please.

schillingjerrid:
@leecalcote Yes to the tooltips?

leecalcote:
@schillingjerrid Yes, if the tooltips fit, it would be great to have them - whether they are below or above the buttons.

schillingjerrid:
@leecalcote Sounds good to me.

raghavaggarwal99:
Doesn't #527 solve this issue?

schillingjerrid:
Hey guys, sorry I haven't been working on this recently. I have a branch with the changes made but I've been on holiday for the last few weeks and haven't finished testing it. @raghavaggarwal99 's branch looks good! I'll un-assign myself from the issue.

leecalcote:
@schillingjerrid no problem. Thanks for signaling here. We're very pleased to get what time of yours that we can. :)

PR #568, opened 2020-01-06
Raghav/color icons

raghavaggarwal99:
Signedoff: raghavaggarwal99<ragha170103048@iitg.ac.in>


raghavaggarwal99:
to the issue #394 and #527

leecalcote:
When I build this branch, here's what I see - 

Screen Shot 2020-01-06 at 8 02 11 AM

raghavaggarwal99:
Screenshot from 2020-01-06 19-34-04

leecalcote:
Ha! :) I'm scratching my head. I wonder if all of you local changes aren't pushed, @raghavaggarwal99. 

Note that there are two merge conflicts, which may be causing the issue as well - https://github.com/layer5io/meshery/pull/568/conflicts

raghavaggarwal99:
here you go again @leecalcote

docs/_includes/footer.html
```diff
@@ -36,7 +36,7 @@
   <div class="footer-icons-list">
 
     <div class="footer-icons">
-      <a class="social__link social__link--mail" href="mailto:community@layer5.io" width="50px" height="50px">
+      <a class="social__link social__link--mail" href="mailto:community@layer5.io">
```
girishranganathan: @raghavaggarwal99 this dint seem to fix the issue because the icons have completely disappeared from the screen when I tested it. @leecalcote reported the same behaviour.

girishranganathan: May be there is a cache issue on your end. Can you please clear your browser cache and try once?

raghavaggarwal99: Yes so the issue is so it's showing correctly in firefox private window but not in google chrome

leecalcote: @raghavaggarwal99 thanks for working through this. Be sure to also note the positioning of the footer, which should be centered on the page.

raghavaggarwal99:
here @girishranganathan @leecalcote  done!!

docs/assets/css/main.css
```diff
@@ -10457,7 +10458,9 @@ button.bg-dark:focus {
 }
 
 .text-center {
-    text-align: center!important
+    text-align: center!important;
+    justify-content: center;
+    margin-left: 550px;
```
leecalcote: I haven’t had a chance to build this latest update, yet. I wonder if this static margin centers the footer content irrespective of browser width. @raghavaggarwal99 in your tests, has the firm maintained center when the browser is resized (from mobile to desktop width)?

girishranganathan:
@raghavaggarwal99 I checked out the changes. They definitely look much better. Wonderful progress. 
It will be better if we can get it centred better and more over not have it in a fixed position so that the footer icons will be responsive with the page as well.

raghavaggarwal99:
@girishranganathan can you check again now?

raghavaggarwal99:
I just changed their names cause it was used somewhere else , properties are same

Issue #561, opened 2020-01-05
UI: Service Mesh icon color too light to be visible


leecalcote:
**Description**
The service mesh icons were recently lightened to beautify the navigation menu (which is great), but has had the unintended consequence of washing them out in other UI views.

**Expected Behavior**
Meshery should display icons in context of the background color. We should either have two sets of icons (white w/transparency and full color) or we should manipulate the icon SVG values to suite the background on which they are displayed.

**Screenshots**
<img width="991" alt="Screen Shot 2020-01-05 at 11 42 17 AM" src="https://user-images.githubusercontent.com/7570704/71783749-ed49ae80-2fb0-11ea-9a18-e137b17dd5e0.png">

<img width="871" alt="Screen Shot 2020-01-05 at 11 42 02 AM" src="https://user-images.githubusercontent.com/7570704/71783750-f175cc00-2fb0-11ea-86bc-6b9bc78a838e.png">

<img width="209" alt="Screen Shot 2020-01-05 at 11 41 38 AM" src="https://user-images.githubusercontent.com/7570704/71783758-f76bad00-2fb0-11ea-9d87-fe4e872d3495.png">

<img width="893" alt="Screen Shot 2020-01-05 at 11 41 50 AM" src="https://user-images.githubusercontent.com/7570704/71783766-166a3f00-2fb1-11ea-9822-f8d9a96dc3b7.png">

gunjanmimo:
sir what is the exact file for this issue?

leecalcote:
I'm tagging @raghavaggarwal99 as he can chime in here as well.

raghavaggarwal99:
Hi @gunjanmimo i think it's done, sorry dude.

PR #569, opened 2020-01-06
Raghav/adapter visible icons

raghavaggarwal99:
Here i separated out all the adapter icons in svg formats. In the left side of dashboard, only white versions will be visible and when you connect to the adapter, you can see the adapter icon colored.


raghavaggarwal99:
to the issue #561

leecalcote:
@raghavaggarwal99 the change to split out white and colored logos looks great.

Screen Shot 2020-01-07 at 8 56 29 PM

You'll want to remove the other unaffiliated (footer, css, docs) changes included in this PR. We're good to merge afterwards.

raghavaggarwal99:
Those changes are for the footer in docs. so that is also done it's no worry

Issue #529, opened 2019-12-21
Hide the readonly field with the adapter url if there is only one adapter for a type


girishranganathan:
**Current Behavior**
Each connected adapter has its own page in Meshery. In these pages there is a read-only field which shows the adapter URL.

**Desired Behavior**
Even though the read-only field with the adapter URL is helpful, we do not want to show it when there is only one adapter of the given type

raghavaggarwal99:
hi @Nikhil-Ladha is this solved?

Nikhil-Ladha:
> hi @Nikhil-Ladha is this solved?

Yes @raghavaggarwal99 it is

ashiscs:
@leecalcote sir, I think this issue is already fixed and needs to be closed!

PR #570, opened 2020-01-06
Fixed hiding adapter-url field for single adapter

Nikhil-Ladha:
#529 Fixed!


ui/components/MeshConfigComponent.js
```diff
@@ -270,6 +281,33 @@ class MeshConfigComponent extends React.Component {
     }, self.handleError);
   }
 
+  handleKubernetesClick = () => {
```
girishranganathan: surprised this is showing in as new. . .

Nikhil-Ladha: Even i am surprised, i guess i will have to make new one!

PR #574, opened 2020-01-08
Fix for hiding adapter url for single adapter

Nikhil-Ladha:
#529 Issue fixed


Nikhil-Ladha:
Okay, I'll check on that.

Issue #565, opened 2020-01-06
404 returned for any page other than main index


leecalcote:
**Description**
Authenticated users with active sessions are given a 404 page not found when refreshing their web browser on any page other than the main index page (localhost:9081).

raghavaggarwal99:
hi @girishranganathan can you tell me a little how to start it?

girishranganathan:
@raghavaggarwal99 thx for showing interest in this issue. 
The issue is that due to some recent library changes the built UI folder structure has changed.
After you run `npm build-ui`, you will find a folder `ui/out`. In this folder we used to have individual folders for the UI pages with their own `index.html`. 
Serving this from Go was easily done using: https://github.com/meshery/meshery/blob/6ddc38d550de9acbe0aa475b6cd0a708e51fc770/router/server.go#L58

Now instead of folders they have changed it to generate html files for each page, __without__ the folders. We will have to change our Go code here to instead serve the html files with/without the `.html` extension. 
https://github.com/meshery/meshery/blob/6ddc38d550de9acbe0aa475b6cd0a708e51fc770/router/server.go#L58 

Hope this helps you get started.

PR #581, opened 2020-01-10
Raghav/go/serve

raghavaggarwal99:


leecalcote:
fixes #565

Double check your commit sign-offs, if you would, @raghavaggarwal99.

PR #620, opened 2020-01-22
Init commit of Meshery chart

Aisuko:
Signed-off-by: Aisuko <urakiny@gmail.com>

The chart for Meshery, also we have many issues need to confirm. But, maybe we can push the init commit just for Meshery.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -1,0 +1,3 @@
+# Meshery installation
+
+This directory contians the Meshery installation configuration in several different flavors.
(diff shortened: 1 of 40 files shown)


Aisuko:
@leecalcote It's pleasure that helps Meshery community more powerful and complete.  In my idea, I must keep the Meshery official install in Kubernetes can be used for some users who new in Meshery. So,  I do keep the `deployment_yaml`. 
The second point is if we need all in one yaml file,  maybe we can use `helm template ` to render that one.  If do this we can just maintain the chart rather than maintain `deployment_yaml` and `Meshery chart`. What's your advice, sir?

Aisuko:
> I think it is done properly

Thank you. You are welcome. Please kindly check #TODO label in the chart. There may need your advice.

PR #622, opened 2020-01-22
Improve documentation

anirudhjain75:
Help user onboarding running kubernetes on Minikube instead of docker desktop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,9 @@ brew install mesheryctl
 mesheryctl start
 ```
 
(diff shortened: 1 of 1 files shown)


leecalcote:
Thanks for jumping right in, @anirudhjain75! 👍 

Give consideration as to whether or not calling out the longer list of [Supported Platforms](https://meshery.layer5.io/docs/installation/platforms) might be appropriate.

Issue #612, opened 2020-01-19
Strip off /graph from grafana and prometheus endpoint


subhamkrai:
Strip off any extraneous portion of the URL beyond the port number (e.g. http://kubernetes.docker.internal:32291/graph ) strip off /graph of both Prometheus and grafana connection endpoint.

girishranganathan:
@subhamkrai responding to your slack message here so that you will have a relevant record:
there are 2 ways you can tackle this issue.
in the UI or in the backend
i will say just do it in the backend handler
there is one for prometheus and one for grafana
you just need to add a check to see if the URL ends with /graph and if it does, remove it

feel free to reach out if you have any questions

subhamkrai:
Sure I'll

leecalcote:
@subhamkrai this issue still seems to remain with Prometheus endpoints. When I enter `http://x.x.x.x:port-number/graph`, this error occurs:

```
meshery_1           | time="2020-04-19T05:38:11Z" level=error msg="unable to get data from URL: http://192.168.1.15:30168/graph/api/v1/status/config due to status code: 404"
meshery_1           | time="2020-04-19T05:38:11Z" level=error msg="unable to connect to prometheus: unable to fetch data from url: http://192.168.1.15:30168/graph/api/v1/status/config"
```

However, when I enter `http://x.x.x.x:port-number`, the connection is made without issue. Will you investigate?

leecalcote:
@subhamkrai here's an example from v0.3.18:

![Untitled](https://user-images.githubusercontent.com/7570704/86525594-be70a000-be4e-11ea-9b5e-960f06f9ebde.gif)

@Nikhil-Ladha recently provided a fix for reloading component data.

leecalcote:
@subhamkrai did we ever confirm that the issue shown this this screen capture is fixed?

PR #623, opened 2020-01-23
Enhancement/subhamkrai/#612

subhamkrai:
tried to strip off  /graph form endpoints of graphana and Prometheus.

### DIFF
--- a/handlers/grafana_handlers.go
+++ b/handlers/grafana_handlers.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/gorilla/sessions"
@@ -28,8 +29,13 @@ func (h *Handler) GrafanaConfigHandler(w http.ResponseWriter, req *http.Request,
 	if req.Method == http.MethodPost {
 		grafanaURL := req.FormValue("grafanaURL")
 		grafanaAPIKey := req.FormValue("grafanaAPIKey")
-		if strings.Contains(grafanaURL, "/graph") {
-			grafanaURL = strings.TrimSuffix(grafanaURL, "/graph")
+
+		u, err := url.Parse(grafanaURL)
+		if err != nil {
+			return
+		}
+		if strings.Contains(grafanaURL, u.RequestURI()) {
+			grafanaURL = strings.TrimSuffix(grafanaURL, u.RequestURI())
 		}
 
 		prefObj.Grafana = &models.Grafana{
--- a/handlers/prometheus_handlers.go
+++ b/handlers/prometheus_handlers.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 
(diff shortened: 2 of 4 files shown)


ashiscs:
fixes #612

handlers/prometheus_handlers.go
```diff
@@ -33,6 +34,11 @@ func (h *Handler) PrometheusConfigHandler(w http.ResponseWriter, req *http.Reque
 			http.Error(w, "unable to connect to prometheus", http.StatusInternalServerError)
 			return
 		}
+
+		if strings.Contains(promURL, "/graph") {
+			promURL = strings.TrimSuffix(promURL, "/graph")
```
leecalcote: @subhamkrai you've got the right idea. 👍 Ultimately, what we'll want here is to strip off any text after the ip address and port number. So, the truncator should leave the first portion inside of the square brackets (`[]`) below alone, but remove everything after that - `[http://x.x.x.x:xxxx]/foo/bar/anything`

You'll need to avoid matching a specific string like "/graph".

handlers/grafana_handlers.go
```diff
@@ -28,8 +29,13 @@ func (h *Handler) GrafanaConfigHandler(w http.ResponseWriter, req *http.Request,
 	if req.Method == http.MethodPost {
 		grafanaURL := req.FormValue("grafanaURL")
 		grafanaAPIKey := req.FormValue("grafanaAPIKey")
-		if strings.Contains(grafanaURL, "/graph") {
-			grafanaURL = strings.TrimSuffix(grafanaURL, "/graph")
+
+		u, err := url.Parse(grafanaURL)
+		if err != nil {
+			return
```
leecalcote: @subhamkrai thanks. What will happen if the parser errors out? We may want to send a signal to the UI to let them know that it is an invalid URL.

handlers/grafana_handlers.go
```diff
@@ -28,8 +29,13 @@ func (h *Handler) GrafanaConfigHandler(w http.ResponseWriter, req *http.Request,
...
+		}
+		if strings.Contains(grafanaURL, u.RequestURI()) {
+			grafanaURL = strings.TrimSuffix(grafanaURL, u.RequestURI())
```
leecalcote: This looks great. 👍

handlers/grafana_handlers.go
subhamkrai: Yes can do that too

subhamkrai:
> It would be nice to confirm the behavior when the Parse() fails.

Yes, I'll have to see that part

leecalcote:
Thank you for identifying which issue this fixes, @ashiscs

leecalcote:
@subhamkrai this issue still seems to remain with Prometheus endpoints. When I enter `http://x.x.x.x:port-number/graph`, this error occurs:

```
meshery_1           | time="2020-04-19T05:38:11Z" level=error msg="unable to get data from URL: http://192.168.1.15:30168/graph/api/v1/status/config due to status code: 404"
meshery_1           | time="2020-04-19T05:38:11Z" level=error msg="unable to connect to prometheus: unable to fetch data from url: http://192.168.1.15:30168/graph/api/v1/status/config"
```

However, when I enter `http://x.x.x.x:port-number`, the connection is made without issue. Will you investigate?

Issue #603, opened 2020-01-15
[mesheryctl] Add meshery server in mesheryctl version 


anandrkskd:
**Current Behavior**
Currently `mesheryctl system version` only _accurately_ displays GitSHA and mesheryctl client version information. The server information shown is inaccurate.

**Prologue**
Meshery will provide a hierarchy of context-specific `version` commands, going from root (`mesheryctl`) to subcommand (`mesheryctl mesh`) to a specific type of that subcommand  (`mesh --type istio --version`).

**Desired Behavior**
Enhance `mesheryctl system version` to use the Meshery REST API provided by #963.

**Acceptance Tests**
1. `mesheryctl system version` should also include Meshery server's version (release tag) and Git SHA (commit ID).
1.  `mesheryctl --version` should reports this same information(_future_: and eventually information for versions of all types of objects under lifecycle management).

**Related**
- meshery server build - #962
- meshery server api - #963

leecalcote:
This issue is related to https://github.com/meshery/meshery/issues/765 - "[mesheryctl] GitSHA in `mesheryctl version` should specify client vs server"

PR #626, opened 2020-01-24
Server version added to versionCmd 

anandrkskd:
Signed-off-by: anandrkskd <anandrkskd@gmail.com>
#603 
Previously only client version was displayed when version command was used.
The new update will show the client version as well as the server version.


mesheryctl/cmd/version.go
```diff
@@ -20,8 +20,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// Build - holds the build info
-var Build string
+// BuildClient - holds the build info of Client
+var BuildClient string
+
+// BuildServer - holds the build info of server
```
leecalcote: Good commenting.

mesheryctl/main.go
```diff
@@ -22,7 +22,8 @@ var version = "dev"
 var commitsha = "SHA"
 
 func main() {
-	cmd.Build = version
+	cmd.BuildClient = version
+	cmd.BuildServer = version
```
leecalcote: It looks like you're setting the server version to always be the same of the client. The point of this feature is to acknowledge that in a given Meshery deployment, the two of these may differ in version number. `mesheryctl` needs to have a way of retrieving the server version from Meshery itself.

leecalcote: @anandrkskd does this make sense?

anandrkskd: Yes sir, Got it, But I had conversation with @girishranganathan about it on the development meeting, and he asked me to just add the server version same as client for now and when we will develop the api for meshery server then we will replace it with actual server version and for now we are also publishing meshery and mesheryctl binary together so both of them carry the same version.
But I guess, for now we can add a new serververion tag in goreleaser for server version in versionCmd.

leecalcote:
Makes sense, @anandrkskd. Will you open up an issue to track the need to implement this server-side?

leecalcote:
And maybe another to track the need to update goreleaser?

PR #631, opened 2020-01-27
Update .gitignore

mayankshah1607:
- Update .gitignore to ignore `.idea` folder generated by IntelliJ based code editors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,5 @@ dist/
 ui/npm-debug.log
 
 Gemfile.lock
+
+**/.idea


leecalcote:
@chandresh-pancholi what do you think? Look good?

.gitignore
```diff
@@ -46,4 +46,6 @@ dist/
 
 ui/npm-debug.log
 
-Gemfile.lock
\ No newline at end of file
+Gemfile.lock
+
+**/.idea
```
chandresh-pancholi: @leecalcote is it necessary to add IDE specific files in .gitignore?

leecalcote: Great question. If we don't, it's inevitable that someone's .vscode or .idea IDE / project settings make their way into a PR or two. If there are standard conventions that we consider those contributing to the project would benefit from as template settings for their IDE, we may want to provide those - - doing so in the CONTRIBUTING.md or similar location, but not as a versioned file (as part of the repos' code, because personalized changes that individuals make would then be lost).

mayankshah1607: I agree with @leecalcote . It was kinda inconvenient that I had to remove the .idea folder anytime I pushed some changes. Would be nice to have it in the .gitignore; I'm a fan of intelliJ based editors like Goland!

chandresh-pancholi: LGTM

PR #637, opened 2020-01-28
Update `logs` command to show appropriate output

mayankshah1607:
This PR addresses https://github.com/layer5io/meshery/issues/625

- Add method `isMesheryRunning()` to determine if Meshery containers are
found before proceeding with the `logs` command.

`meshery logs` command now shows the following output if meshery is not running:

```console
$ meshery logs

No logs to show. Meshery is not running.


```

### DIFF
--- a/mesheryctl/cmd/logs.go
+++ b/mesheryctl/cmd/logs.go
@@ -30,6 +30,11 @@ var logsCmd = &cobra.Command{
 	Short: "Print logs",
 	Long:  `Print history of Meshery's container logs and begin tailing them.`,
 	Run: func(cmd *cobra.Command, args []string) {
+		if ok := isMesheryRunning(); !ok {
+			log.Error("No logs to show. Meshery is not running.")
+			return
+		}
+
 		log.Info("Starting Meshery logging...")
 
 		if _, err := os.Stat(dockerComposeFile); os.IsNotExist(err) {
--- a/mesheryctl/cmd/vars.go
+++ b/mesheryctl/cmd/vars.go
@@ -131,3 +131,11 @@ func installprereq() {
 	}
 	log.Info("Prerequisite Docker Compose is installed.")
 }
+
+func isMesheryRunning() bool {
+	op, err := exec.Command("docker-compose", "-f", dockerComposeFile, "ps").Output()
+	if err != nil {
+		return false
+	}
+	return strings.Contains(string(op), "meshery")
+}


mesheryctl/cmd/logs.go
```diff
@@ -30,6 +30,11 @@ var logsCmd = &cobra.Command{
 	Short: "Print logs",
 	Long:  `Print history of Meshery's container logs and begin tailing them.`,
 	Run: func(cmd *cobra.Command, args []string) {
+		if ok := isMesheryRunning(); !ok {
+			log.Error("No logs to show. Meshery is not running.")
```
leecalcote: Nice.

mayankshah1607:
@anandrkskd Done! mind taking a look?

mesheryctl/cmd/vars.go
```diff
@@ -131,3 +131,33 @@ func installprereq() {
 	}
 	log.Info("Prerequisite Docker Compose is installed.")
 }
+
+func isMesheryRunning() bool {
+	ps := exec.Command("docker-compose", "-f", dockerComposeFile, "ps")
+	grep := exec.Command("grep", "meshery")
```
leecalcote: @mayankshah1607, this is an improvement and works on macOS and Linux, but is an area where we will either need to leverage a universally available tool (e.g. parse the output in go) or will need to check for OS arch and ensure that an OS-specific utility is used for Windows, macOS and Linux compatibility of `mesheryctl`.

mayankshah1607: @leecalcote yep, that's a good one. Should have thought of that. Will get back shortly with a fix :)

mesheryctl/cmd/vars.go
```diff
@@ -131,3 +132,36 @@ func installprereq() {
 	}
 	log.Info("Prerequisite Docker Compose is installed.")
 }
+
+func isMesheryRunning() bool {
+	if runtime.GOOS == "linux" {
+		ps := exec.Command("docker-compose", "-f", dockerComposeFile, "ps")
+		grep := exec.Command("grep", "meshery")
```
leecalcote: @mayankshah1607 good update. I think we're ok to include Darwin here as well. `grep` comes with the OS.

Ultimately, the safest and singularly compatible route is to parse the stream / string in Go.

mayankshah1607: @leecalcote yep. Could you take a look now?
Tried a few things that might work for windows, but as you said, the safest approach is to parse the output during runtime itself. I think it would work universally now :)

mayankshah1607: Thins also look neat now as we no longer deal with piping

PR #641, opened 2020-01-29
Add new completion cmd

mayankshah1607:
fixes https://github.com/layer5io/meshery/issues/636

- Add new command `completion` that generates auto-completion scripts
for zsh / bash
- Add helper `getCompletion()`

### DIFF
--- a/mesheryctl/cmd/completion.go
+++ b/mesheryctl/cmd/completion.go
@@ -61,7 +61,7 @@ func getCompletion(sh string, parent *cobra.Command) (string, error) {
 	if len(sh) == 0 {
 		err = errors.New("Shell not specified.")
 	}
-	if len(args) > 1 {
+	if len(sh) > 1 {
 		err = errors.New("Too many arguments. Expected only the shell type.")
 	}
 	switch sh {


leecalcote:
@mayankshah1607 that was quick. 👍 

FYI - https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

mayankshah1607:
Thanks @leecalcote ! 
I followed up with the links and resources mentioned in the issue. I must point out that `kubectl completion` for zsh did not seem to work for me; and neither do the changes in this PR. It seems like this is an issue with the scripts generated for zsh. Refer https://github.com/spf13/cobra/issues/881

@anandrkskd @leecalcote would you mind testing out this branch on a different shell just to confirm my understanding?

leecalcote:
@mayankshah1607 I tested this on macOS with zsh and bash. It didn't seem to work. That said, I build the `mesheryctl` binary in this branch and sourced the completion script from that binary, while I have `mesheryctl` installed via brew on the system. I assume that the completion script doesn't have any dependency on this PR's version of `mesheryctl`, but will / should work with the version of `mesheryctl` installed on my system.

mayankshah1607:
@leecalcote I think your assumption is right. The completion scripts shouldn't depend on the PR's version of `mesheryctl`. What do you think could be going wrong here? My guess is that maybe we shouldn't source it. Can't find much documentation on this.

leecalcote:
@mayankshah1607 some reference material - https://www.kubernetes.io/docs/reference/kubectl/cheatsheet

leecalcote:
Even better reference - https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/completion/completion.go

mesheryctl/cmd/completion.go
```diff
@@ -0,0 +1,85 @@
...
+
+// getCompletion checks given arguments and executes command
+func getCompletion(sh string, parent *cobra.Command) (string, error) {
+	var err error
+	var buf bytes.Buffer
+
+	if len(sh) == 0 {
+		err = errors.New("Shell not specified.")
+	}
+	if len(sh) > 1 {
```
mayankshah1607: @leecalcote I'm not sure if this is required. Line 43-44 handles this scenario and would throw an error before this situation is encountered.

leecalcote: Oh, ok. Sounds good. Feel free to revert. 

While here, given thought to the consistently of approach across the rest of the cammands, if you would. Consider the UX in terms of messages to the user.

kushthedude:
Any updates over here ?

mayankshah1607:
Hi, @kushthedude ! 
Nothing as of now, feel free to take up this one :)

leecalcote:
Fyi - https://github.com/spf13/cobra/issues/881#issuecomment-609491600

PR #661, opened 2020-02-06
meshery: providers are now a choice on start

girishranganathan:
#### Description
This PR contains implementation for bringing the concept of providers right up front in the system. Users can now choose the provider they want to work with and the respective provider will be chosen on the fly.
As a result of this feature most of the handlers had to be tweaked to embrace the dynamic choice of the provider.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,6 +127,8 @@ jobs:
     steps:
     - name: Check out code
       uses: actions/checkout@master
+    - name: Unshallow
+      run: git fetch --prune --unshallow
     - name: Set up Go
       uses: actions/setup-go@v1
       with:


Makefile
```diff
@@ -48,26 +48,6 @@ run-local-cloud:
...
-run-local-cloud-noauth:
-	cd cmd; go clean; rm meshery; go mod tidy; go build -tags draft -a -o meshery; \
-	SAAS_BASE_URL="http://mesherylocal.layer5.io:9876" \
-	PORT=9081 \
-	DEBUG=true \
-	NO_AUTH=true \
-	ADAPTER_URLS=$(ADAPTER_URLS) \
-	./meshery; \
-	cd ..
-
```
leecalcote: @girishranganathan do these go away because, there is no local cloud anymore? Only local provider?

Makefile
```diff
@@ -90,14 +70,20 @@ proto:
...
 
-# Runs the UI interface on your local machine.
+# Runs the Meshery UI interface on your local machine.
 run-ui-dev:
 	cd ui; npm run dev; cd ..
 
-# Builds the user interface on your local machine.
+# Runs the Provider UI interface on your local machine.
+run-provider-ui-dev:
+	cd provider-ui; npm run dev; cd ..
```
leecalcote: @girishranganathan the locally running provider ui and the regular ui collide on port 3000.

provider-ui/ui_dev_server.js
```diff
@@ -0,0 +1,47 @@
...
+  });
+  res.end('Unexpected issue.');
+});
+
+
+app.prepare().then(() => {
+  createServer((req, res) => {
+    const { pathname } = parse(req.url, true)
+    if (pathname.startsWith("/api") || pathname.startsWith("/logout") || pathname.startsWith("/login")){
+        proxy.web(req, res, { target: 'http://localhost:9081' });
```
leecalcote: @girishranganathan How will this behave when deployed in-cluster?

provider-ui/ui_dev_server.js
```diff
@@ -0,0 +1,47 @@
...
+    //   return
+    // }
+    // // assigning `query` into the params means that we still
+    // // get the query string passed to our application
+    // // i.e. /blog/foo?show-comments=true
+    // app.render(req, res, '/blog', Object.assign(params, query))
+
+  }).listen(port, err => {
+    if (err) throw err
+    console.log(`> Ready on http://localhost:${port}`)
```
leecalcote: Should we variabilize the hostname?

leecalcote:
Screen Shot 2020-02-10 at 4 14 22 PM
How can these two independent selections be collapsed into one selection?

Makefile
girishranganathan: Local cloud still exists. It is if you want to run Meshery-cloud on your local and want Meshery to point to it.

Makefile
girishranganathan: I have changed the port a different one now.

girishranganathan:
This is where you can start: https://github.com/layer5io/meshery/blob/d34f4d62d328a6cd0ba2a23c5d5977049d24591b/provider-ui/components/ProviderComponent.js#L57.
Right now I started off with separate variables for remote and local. You can merge them and make this component much simple.

leecalcote:
@Nikhil-Ladha could you look at consolidating these two separate dropdowns into one dropdown list?

PR #674, opened 2020-02-13
`mesheryctl`: Add sudo with the mv command instead of running bash with sudo

### LINKED ISSUE #643: Installation: Request for sudo password is lost among other script output
**Current Behavior**
While running the latest `meshery` installation script on macOS in zsh, users don't necessary see the prompt for sudo permissions. See screenshot.

**Screenshot**
Screen Shot 2020-01-31 at 12 06 41 PM

Related to #642

hinupurthakur:
Signed-off-by: nupur-thakur <nupur.s.thakur@gmail.com>

#### Description
This change was required as the command for installing `mesheryctl` was unable to take password as the password message was getting lost in the curl command.

After this change :
The the command for installing meshery using bash needs to be updated as well to `curl -L https://git.io/meshery | bash -  `  instead of  `curl -L https://git.io/meshery | sudo bash - `

This PR fixes #643 

#### Notes to reviewers:
This script is supposed to work for macOS users with sudo-less `mv` command so please check (on macOS) if any error occurs.

### DIFF
--- a/meshery
+++ b/meshery
@@ -62,7 +62,12 @@ unzip ${PWD}/meshery.zip
 
 printf "\nInstalling mesheryctl in /usr/local/bin.\n"
 WHOAMI=$(whoami)
-if mv ${PWD}/mesheryctl /usr/local/bin/mesheryctl ; then
+if [ $OSEXT = "Linux" ] ; then
+  MV_CMD=$(sudo mv ${PWD}/mesheryctl /usr/local/bin/mesheryctl)
+else
+  MV_CMD=$(mv ${PWD}/mesheryctl /usr/local/bin/mesheryctl)
+fi
+if $MV_CMD ; then
   echo "mesheryctl installed"
 else
 	echo "Cannot install mesheryctl. Check permissions of $WHOAMI for /usr/local/bin."


nupur-thakur:
Check the script by using the command `curl -L https://raw.githubusercontent.com/nupur-thakur/meshery/bug/nupur-thakur/643/meshery | bash -`

meshery
```diff
@@ -62,7 +62,12 @@ unzip ${PWD}/meshery.zip
 
 printf "\nInstalling mesheryctl in /usr/local/bin.\n"
 WHOAMI=$(whoami)
-if mv ${PWD}/mesheryctl /usr/local/bin/mesheryctl ; then
+if [ $OSEXT = "Linux" ] ; then
+  MV_CMD=$(sudo mv ${PWD}/mesheryctl /usr/local/bin/mesheryctl)
```
leecalcote: Thank you, @nupur-thakur! It looks like this will do the trick. A slightly more generic approach (that potentially covers other hiccups) is to attempt a `mv`, catching failure on error due to lack of permissions, executing a `sudo mv` thereafter.

nupur-thakur: Okay Sir, I understand what you are saying but As per I know It's compulsory to have sudo with mv command for /usr/local/bin folder in Linux environment. 
But May be in bigger picture your approach is better so should I change the script according to your suggestions?

leecalcote: @nupur-thakur I'm a fan of continual, incremental improvement and am inclined to move forward here. 

The rub in doing so is that we won't be addressing the root cause of the issue - permissions. If this issue only ever occurs in `uname`="Linux", then we're ok. To the extent this isn't the case (that permission is an issue in other environments / situations), this issue will rear its head again. Conceptually, if we address the permission issue directly, we won't see the issue again.

Alternatively, we could accept this PR and log an issue to circle back to improve here at a future point (if ever we arrive at that point). I'll let you decide as its not a critical issue and one that we might not ever see again considering the solution you've provided in this PR.

nupur-thakur: @leecalcote Sir, Now I completely understand that we want to address the issue of permission not for a specific environment but overall whereas previously I was looking at the problem as environment specific. 
I will make appropriate changes and create another pull request with updated documents as well.

PR #701, opened 2020-02-25
Performance: Sorting of "service meshes" list based on adapters present

### LINKED ISSUE #587: Performance: Sorting of "service meshes" list based on adapters present
**Current Behavior**
When users visit the Performance test page, they are presented an alphabetically sorted list of service meshes with which they may associate to their test.

**Desired Behavior**
The list of entries in the “Service Mesh” drop-down needs to be a bit more intelligent in that it should rearrange the ordering of the list based on which adapters the user is running. For example, if the user only has the Istio adapter loaded, Istio should be listed first and there should be a small, horizontal line separating service meshes that have adapters loaded and those that don’t.

**Mockup**
Untitled

kanishkarj:
Signed-off-by: kanishkarj <kanishkarj@hotmail.com>

#### Description

This PR fixes #587    

#### Notes to reviewers:

Please do run this and verify if its working fine. I have attached the screenshots of it in my system below.

With Istio configured
Without any adapters configured

#### Checklist

- [ ] [DCO](https://github.com/layer5io/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin) signed.
- [ ] Title of the PR starts with area name, e.g. `meshery:`, `mesheryctl:`, `ui:`.

### DIFF
--- a/ui/components/MesheryPerformanceComponent.js
+++ b/ui/components/MesheryPerformanceComponent.js
@@ -4,7 +4,7 @@ import Button from '@material-ui/core/Button';
 import Typography from '@material-ui/core/Typography';
 import { withStyles } from '@material-ui/core/styles';
 import Grid from '@material-ui/core/Grid';
-import { NoSsr, Tooltip, MenuItem, IconButton, CircularProgress, FormControl, FormLabel, RadioGroup, FormControlLabel, Radio } from '@material-ui/core';
+import { NoSsr, Tooltip, MenuItem, IconButton, CircularProgress, FormControl, FormLabel, RadioGroup, FormControlLabel, Radio, Divider } from '@material-ui/core';
 import TextField from '@material-ui/core/TextField';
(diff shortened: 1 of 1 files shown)


leecalcote:
I wonder if this next item is a behavior specific to running in dev (running locally using `make run`),  but another item to look at is that adapters other than Istio seem to not get represented.

Untitled

kanishkarj:
> I wonder if this next item is a behavior specific to running in dev (running locally using `make run`), but another item to look at is that adapters other than Istio seem to not get represented.
> 
> Untitled

@leecalcote  I have attached a screenshot of it. Also, I tested it with `make build-ui && make run-local` and it works. Could you try that once ?

Screenshot_2020-02-26 Meshery

leecalcote:
Odd in that, yes, I did a `make build-ui` first, which is why in the screen capture that I did, the divider shows under the Istio entry in the dropdown list.

PR #716, opened 2020-03-09
mesheryctl: mesheryctl performance subcommand

kanishkarj:
#### Description

This PR implements the performance testing subcommand in `mesheryctl`. All the requirements of `perf` subcommand in the _Meshery CLI Commands and Documentation_ has been implemented except the file subcommand. 

#### Notes to reviewers:

Please compile mesheryctl and try running this on the latest Meshery. A sample of how to test is given below : 

```bash
./mesheryctl perf --url https://github.com/ --duration 3s --qps 10 --parallel-requests 10
```

#### Checklist

### DIFF
--- a/mesheryctl/cmd/perf.go
+++ b/mesheryctl/cmd/perf.go
@@ -60,9 +60,9 @@ var perfCmd = &cobra.Command{
 		//Check prerequisite
 		preReqCheck()
 
-		println("Test name not provided, using random name : ", testName)
+		println("Test name used : ", testName)
 
-		const mesheryURL string = "http://localhost:9081/api/load-test-smps"
+		const mesheryURL string = "http://localhost:9081/api/load-test-smps?"
 		postData := ""
 
 		startTime := time.Now()
@@ -130,7 +130,7 @@ func init() {
 	perfCmd.Flags().StringVar(&qps, "qps", "1", "DESCRIPTION")
 	perfCmd.Flags().StringVar(&parallelRequests, "parallel-requests", "1", "DESCRIPTION")
 	perfCmd.Flags().StringVar(&testDuration, "duration", "10s", "DESCRIPTION")
-	perfCmd.Flags().StringVar(&testCookie, "cookie", "meshery-provider=Default Local Provider", "DESCRIPTION")
+	perfCmd.Flags().StringVar(&testCookie, "cookie", "meshery-provider=Local (ephemeral session) (free use)", "DESCRIPTION")
 	perfCmd.Flags().StringVar(&loadGenerator, "load-generator", "fortio", "DESCRIPTION")
 	rootCmd.AddCommand(perfCmd)
 }


kanishkarj:
@leecalcote Please take a look at this.

mesheryctl/cmd/perf.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Generates random string
+func StringWithCharset(length int) string {
+	const charset = "abcdefghijklmnopqrstuvwxyz" + "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	b := make([]byte, length)
+	for i := range b {
+		b[i] = charset[seededRand.Intn(len(charset))]
+	}
+	return string(b)
+}
```
leecalcote: Will the /load-test-smps API generate a random name for the test if one is not provided?

kanishkarj: I shall look into it.

kanishkarj: No actually, the server responds by saying `Provide a name for the test`.

kanishkarj:
The commit #624da70 updates the cookie value. One of the recent commits has changed the value for meshery-provider cookie from `Default Local Provider` to `Local (ephemeral session) (free use)` .

mesheryctl/cmd/perf.go
leecalcote: I saw the same behavior in my tests earlier today, too. Interesting as I suppose Meshery's frontend generates a unique value when tests are invoked in the UI. That random name generator functionality should ultimately sit server-side with Meshery.

kanishkarj: So shall I add a random name generator in the server side?

mesheryctl/cmd/perf.go
```diff
@@ -0,0 +1,136 @@
...
+
+func init() {
+	perfCmd.Flags().StringVar(&testURL, "url", "", "DESCRIPTION")
+	perfCmd.Flags().StringVar(&testName, "name", StringWithCharset(8), "DESCRIPTION")
+	perfCmd.Flags().StringVar(&testMesh, "mesh", "", "DESCRIPTION")
+	// perfCmd.Flags().StringVar(&testFile, "file", "", "DESCRIPTION")
+	perfCmd.Flags().StringVar(&qps, "qps", "1", "DESCRIPTION")
+	perfCmd.Flags().StringVar(&parallelRequests, "parallel-requests", "1", "DESCRIPTION")
+	perfCmd.Flags().StringVar(&testDuration, "duration", "10s", "DESCRIPTION")
+	perfCmd.Flags().StringVar(&testCookie, "cookie", "meshery-provider=Local (ephemeral session) (free use)", "DESCRIPTION")
```
leecalcote: Ideally, `perf` command might lookup the name of the currently loaded provider.

mesheryctl/cmd/perf.go
```diff
@@ -0,0 +1,136 @@
...
+	perfCmd.Flags().StringVar(&loadGenerator, "load-generator", "fortio", "DESCRIPTION")
```
leecalcote: Might want to lookup the available load-generators (as this is an area of extensibility for Meshery) and then validate that the string provided matched an available generator.

mesheryctl/cmd/perf.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func init() {
+	perfCmd.Flags().StringVar(&testURL, "url", "", "DESCRIPTION")
+	perfCmd.Flags().StringVar(&testName, "name", StringWithCharset(8), "DESCRIPTION")
+	perfCmd.Flags().StringVar(&testMesh, "mesh", "", "DESCRIPTION")
+	// perfCmd.Flags().StringVar(&testFile, "file", "", "DESCRIPTION")
+	perfCmd.Flags().StringVar(&qps, "qps", "1", "DESCRIPTION")
+	perfCmd.Flags().StringVar(&parallelRequests, "parallel-requests", "1", "DESCRIPTION")
+	perfCmd.Flags().StringVar(&testDuration, "duration", "10s", "DESCRIPTION")
```
leecalcote: It'd be good to mimic the same defaults as provided in the UI.

mesheryctl/cmd/perf.go
kushthedude: @kanishkarj Why is there a random name generator? AFAIK, in the UI we can enter the test name, won't this create confusion when the user sees the test results? Or am I getting something over here ?

kanishkarj: oh yes, we wanted to handle the case where the user doesn't provide a test name. let's say the user doesn't provide a name while calling the CLI, then we call the random string generator to get a random name, and use that.

Actually, we still aren't sure if this random generation should happen server side or client-side. I put it here for now.

PR #756, opened 2020-03-24
Fixes #746 

### LINKED ISSUE #746: [Meshery] Unable to provision Citrix Service Mesh
**Description**
Use of ["Citrix" in Navigator.js](https://github.com/layer5io/meshery/blob/master/ui/components/Navigator.js#L266) prevents loading of the adapter page for the Citrix Service Mesh. If `id="Citrix"` is swapped for `id="cpx"` the page loads and users can successfully invoke each of the adapter's features except provisioning of the Citrix Service Mesh.

**Expected Behavior**
Citrix Service Mesh should be provisioned.

dheerajng:
Correcting the id value in Navigator so as to provision Citrix Service Mesh

Signed-off-by: dheerajng <dhiraj.gedam@citrix.com>

#### Description

This PR fixes #746


leecalcote:
@dheerajng thank you! using “cpx” here has the affect of displaying a generic service mesh logo, not that of the Citrix logo, if I recall correctly from my quick tests of making this same change earlier. Hmm.

leecalcote:
@dheerajng thanks for this. There's a disconnect between how the UI is mapping menu items and how we're using labels/descriptions. Here's behavior after the last commit (the same issue I faced when updating descriptions). You can easily get the menu to work, but then lose the proper icon. You can get the proper icon, but then lose the ability to actually deploy CPX. It may be the case the an update will be necessary in the adapter.

Untitled

leecalcote:
I get the following error when choosing to run the install with mTLS:

```
DEBU[0071] Computed Resource: schema.GroupVersionResource{Group:"", Version:"v1", Resource:"secrets"}
WARN[0071] unable to create the requested resource, attempting operation without namespace: the server does not allow this method on the requested resource
ERRO[0071] unable to create the requested resource, attempting to update: the server does not allow this method on the requested resource
WARN[0071] unable to retrieve the resource with a matching name, attempting operation without namespace: the server could not find the requested resource
ERRO[0071] unable to retrieve the resource with a matching name, while attempting to apply the config: the server could not find the requested resource
DEBU[0071] sending event: &meshes.EventsResponse{EventType:2, Summary:"Error while deploying Cpx", Details:"unable to retrieve the resource with a matching name, while attempting to apply the config: the server could not find the requested resource", OperationId:"95c976dd-eb7f-4301-b4e6-8c1aaa79f3f6", XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}```

leecalcote:
@dheerajng With this change, are you able to successfully provision Citrix Service Mesh? I’m not. We might want to jump on a call to troubleshoot this issue.

leecalcote:
Whoo-hoo! Thank you, @dheerajng.

PR #797, opened 2020-04-05
feat: Finalise eslint and all commands

kushthedude:
Eslint is added with all errors removed.
Lot of work is pending but I have managed to suppress the warnings.
CI is integrated for lint too.


kushthedude:
@leecalcote Successfully integrated EsLint for UI:
- Added CI command.
- Added makefile command, need to update documentation that pleases run `make run-ui-lint` before committing changes.
- Fixed all the errors, there were 200 errors remaining which were more of a labour work than actual logic hence I have converted them into warnings which may be solved by new-contributors and I will open a tracking issue for same.

Remaining :
- Completing integration for Provider-UI.
- I will integrate pre-commit-hook which will run lint command as soon as developers do ` git commit _______`

kushthedude:
@leecalcote Is it good to be merged?

.github/workflows/ci.yml
```diff
@@ -36,63 +36,9 @@ jobs:
       with:
         go-version: ${{ secrets.GO_VERSION }}
     - run: GOPROXY=direct GOSUMDB=off GO111MODULE=on go get -u github.com/kisielk/errcheck; /home/runner/go/bin/errcheck -tags draft ./...
-#   static_check:
```
leecalcote, line 39: I'm leery of removing these comments as the intention is uncomment at some point. These serve as a good reminder that they need to be gradually uncommented and included in the build. What do you think? If we have someone ready to take on issues that are found with these checks, we can begin uncommenting.

Nikhil-Ladha:
LGTM, That's quite a bit of cleaning though! Woof.. :+1:

ui/lib/store.js
```diff
@@ -254,7 +254,7 @@ export const makeStore = (initialState, options) => {
 
 export const resultsMerge = (arr1, arr2) => {
   const keys = {}
-  const arr = [];
+  var arr = [];
```
nveenjain, line 257: why are we using ```var``` instead of ```const```?

ui/.eslintignore
```diff
@@ -0,0 +1,7 @@
+node_modules
+out
+.next
+static
+package.json
+package-lock.json
+nodemon.json
```
nveenjain, line 7: It would be great if there was a newline character at end of file and was enforced via the `eol-last` eslint rule.

ui/components/CustomTableFooter.js
```diff
@@ -141,10 +103,9 @@ class CustomTableFooter extends Component {
...
 CustomTableFooter.propTypes = {
-  // classes: PropTypes.object.isRequired,
   changePage: PropTypes.func.isRequired,
   rowsPerPage: PropTypes.number.isRequired,
   page: PropTypes.number.isRequired,
 };
 
-export default withStyles(defaultFooterStyles, { name: "CustomFooter" })(CustomTableFooter);
\ No newline at end of file
+export default withStyles(defaultFooterStyles, { name: 'CustomFooter' })(CustomTableFooter);
```
nveenjain, line 111: It would be great if all the strings were either template literals(using quote:"backtick") or just normal strings with a single-quotation (using quote:"backtick") or quote:"double". (In this file we have all the 3 format of strings).

ui/components/CustomToolbarSelect.js
```diff
@@ -121,51 +118,55 @@ class CustomToolbarSelect extends React.Component {
             <CompareArrowsIcon className={[classes.icon, classes.inverseIcon].join(" ")} />
```
nveenjain, line 118: These comments should also be deleted if we are deleting comments as in ui/components/CustomTableFooter.js

ui/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "dev": "nodemon ui_dev_server.js",
+    "lint": "eslint --fix .",
```
nveenjain, line 8: I guess, it should only be eslint , the ```--fix``` option is not required..

leecalcote:
Thanks for this makeover, @kush. No small bit of work. 🔨

kushthedude:
@nveenjain For your reference
-  --fix will fix up all the lint errors until it is out of bound for
ESLint, hence fix is the best option for it.
- Constant declaration and then changing its value is considered as bad
practice for a good developer.
- A production open source project should be free from commented code. I
still dont understand point of having this much commented code, not only it
decreases code quality but also creates hurdle when new contributors get to
know the codebase.

nveenjain:
@kushthedude, while I agree that --fix will fix all errors, but adding them in lint command is not required, eslint should report these errors and the developer can run eslint --fix on it. lint should not fix any error, it should just check for errors.
Because it may happen that someone pushes the code without running lint command, and in the CI no error will be seen. Since --fix option is applied. However the code would not have been change, it would just change on the CI server.

I agree that constant declarations are good, that's why I think that there we should use const instead of var (which is ES5 already).

Agreed that open source project should be free from irrelevant comments, that's why I think that the comments in ui/components/CustomToolbarSelect.js should also be removed.

kushthedude:
Removing comment is totally labour work, I am working on a script to do the work for me. Meanwhile if you want to remove them from other go+js files, please be my guest.
--fix args is the best option, moreover there are 150+ errors of useless props and unused variable which I have turned them into warnings. Again, that is a pure labour work for which I will open a tracking issue so that all new contributors or existing may take up a file and refine it.

snitin315:
We should use --fix, most of the projects do.

nveenjain:
yes @snitin315 , --fix should be used, but it should not be used while we lint, by lint, I mean run as part of test in CI job for the project, it should be used as a separate command, by the developer.
All the top projects related to JS directly use eslint command, --fix will result in extra errors.
It should be executed by the developer side and not during test.
For reference:-
https://github.com/facebook/react/blob/master/package.json#L105 (its a script that import eslint and doesn't use --fix option)
https://github.com/babel/babel/blob/master/Makefile#L112
https://github.com/webpack/webpack/blob/master/package.json#L134
https://github.com/graphql/graphql-js/blob/master/package.json#L31

 --fix option can be run once offline and it will fix all the errors eslint can fix, and it won't be required again.

snitin315:
Well we can have two different scripts as well. 
lint and lint:fix

nveenjain:
Yes, that would be perfect.
As long as lint:fix doesn't runs as a part of tests, it would work, and in
some cases even help developers who will be able to fix eslint errors using
`npm run lint:fix` instead of specifying eslint path(from node modules or
from global installation) with --fix option.

snitin315:
I'm also thinking to integrate lint-staged and husky. What do you think? 
I can create a PR if you agree.

nveenjain:
Sounds good @snitin315, but I guess @kushthedude maybe already working on it as per https://github.com/layer5io/meshery/pull/797#issuecomment-609471616 please confirm.

.github/workflows/ci.yml
kushthedude: @leecalcote we can restore it later. Also, anyone can again see the commented part by going through history. For now it can be removed imo.

kushthedude:
> I'm also thinking to integrate lint-staged and husky. What do you think?
> I can create a PR if you agree.

You may go ahead once this PR is finalised, also try to integrate go-precommit if possible else no issues.

kushthedude:
> Yes, that would be perfect. As long as lint:fix doesn't runs as a part of tests, it would work, and in some cases even help developers who will be able to fix eslint errors using `npm run lint:fix` instead of specifying eslint path(from node modules or from global installation) with --fix option.
> […](#)

Comment to your point, developers won't be using npm run lint as we already makefile support. Make lint will do the thing for them including fixing the lints issues. Running the linter and then again running the fix command is obviously the pain in the back. Also, most of the packages you mentioned are services which are used by-products. Most end products integrate the fix with linter to save running another command and not to mention save resources. 
A big project usually takes 3-5 minutes to run linter to check the rules, not wasting another 3-5 minutes for linter to fix the obvious indents & literals is not productive in my pov.

kushthedude:
@leecalcote Please check if anything is remaining, I will open the tracking issues for the lint warnings once this gets merged.

snitin315:
I agree with @kushthedude.

nveenjain:
> Most end products integrate the fix with linter to save running another command and not to mention save resources.
@kushthedude, having eslint --fix will result in errors, as CI will never report any error that can be fixed by eslint. CI would fix it's own copy of the repository. If we are saving efforts of running a command that would be run once by developer while that could cause more errors in the repository, that would not be a good design...
All the products (including the tools used in JS) only use eslint.
For reference, all the JS end products in trending section on GitHub right now using eslint without --fix option
https://github.com/trekhleb/javascript-algorithms/blob/master/package.json#L7
https://github.com/RocketChat/Rocket.Chat/blob/develop/package.json#L29
https://github.com/jitsi/jitsi-meet/blob/master/package.json#L141

kushthedude:
I think you don't understand the term.use of resources. At once you are concerned about the lint script and now you are concerened about the lint introduced in CI. Also FYI, there will be no errors as after merging of this @snitin315 will introduce the pre-commit hooks for lint which means before every git commit, Lint script will be run on local . I dont know where do you get yourself confused in this easy chain of command, it would be better if you read the my initial comment on this PR for flow of operations after this. If you still have any doubt, ping me I will help you understand.

kushthedude:
Since this conversation with @nveenjain is going to nowhere and only resulting in more difference arguments from initial one. 
@leecalcote can you please check the above code and flow of operations listed by me in initial comments? Let me know if there is anything missing or needs to be amend.

snitin315:
@nveenjain there will be no chances of having lint errors, if user have any they will be fixed before committing the changes.

nveenjain:
My take is lint command should just check for the lint and it should not fix any errors.

As mentioned in https://github.com/layer5io/meshery/pull/797#issuecomment-610548249 and https://github.com/layer5io/meshery/pull/797#issuecomment-610511259

If we are relying solely on git hooks, to fix the eslint errors, then there should not be any reason to include lint in CI, as no one will be able to commit without eslint errors. git hooks can be skipped, and if they are, it should result in error anyhow if checking on CI,but it wouldn't in current premises. Most projects I pointed out also have pre-commit hooks, but linting step doesn't fix the errors automatically there too. Some do provide a separate command to let eslint auto-fix, but that is a separate command and not used either in git pre-commit or while testing.

snitin315:
You are asking for the following conf.
For CI - lint
For hook - lint:fix

nveenjain:
Yeah, this is the way it should be. CI should never fix any error.

kushthedude:
So what do you want, developers to commit the un-linted code and see the CI
check , oh my code is not linted and push the code again.
CI reinforcements of tests is for double check and maintaining the quality
of delivery to be made from next commit. Going your way, I should remove
the fix script and not setup hooks and let developers wait for CI to fail
and then again push the linted code.

snitin315:
also `eslint --fix` does not work for all rules.

kushthedude:
I will wait for @leecalcote now, it has been converted into an argument
from a PR review which is still ambiguous.

.github/workflows/ci.yml
```diff
@@ -119,11 +64,24 @@ jobs:
...
+      uses: actions/checkout@master
+      with:
+        fetch-depth: 1
+    - uses: actions/setup-node@v1
+      with:
+        node-version: ${{ secrets.NODE_VERSION }}
+    - run: |
+        cd ui
+        npm i
+        npm test
```
nveenjain, line 81: This works! Thanks!

snitin315:
don't worry I will configure such that the changes will be staged.

kushthedude:
@leecalcote Good to be merged ?

leecalcote:
> @leecalcote Good to be merged ?

I think so, yes. I've put this as a topic in today's community meeting. Could you socialize this update and what it means to contributing to the UI?

kushthedude:
@leecalcote update ?

leecalcote:
Given the amount of discussion generated and implications of this change, a verbal discussion is warranted, I think. Can we get this on the agenda of this week's developer meeting? 

What do contributors updating the UI need to be aware of given the changes here? Are those considerations incorporated into CONTRIBUTING docs?

kushthedude:
> Given the amount of discussion generated and implications of this change, a verbal discussion is warranted, I think. Can we get this on the agenda of this week's developer meeting?
> 
> What do contributors updating the UI need to be aware of given the changes here? Are those considerations incorporated into CONTRIBUTING docs?

Sure! Contributors must be aware of the new rules we have incorporated of which one of the biggest & most ignored rule in our codebase in `importing of unused variables` & `blank variable assignments`.
Also, we have to update the docs to let contributors know about lint script so that CI doesn't fail on their Pull Request. 
We can definitely have a discussion for this & then move ahead.

PR #814, opened 2020-04-10
feat: adding security & static checks

kushthedude:
Reference : https://github.com/layer5io/meshery/pull/797#discussion_r404800265


kushthedude:
@leecalcote There is some error with GoTools as StaticCheck is not absorbing staticcheck.config file for loading its configuration. Hence I have disabled some rules manually. I will remove them once this issue is fixed https://github.com/golang/go/issues/36373

staticcheck.conf
```diff
@@ -0,0 +1,10 @@
+checks = ["all", "-ST1000", "-ST1003", "-ST1016"]
+initialisms = ["ACL", "API", "ASCII", "CPU", "CSS", "DNS",
```
leecalcote, line 2: Should we use the word “acronyms” for this variable name instead?

staticcheck.conf
```diff
@@ -0,0 +1,10 @@
+checks = ["all", "-ST1000", "-ST1003", "-ST1016"]
```
leecalcote, line 1: Is it possible to locate this file somewhere other than the root folder in order to today and organize the project folder structure?

models/grafana_helper.go
```diff
@@ -67,6 +67,9 @@ func (g *GrafanaClient) Validate(ctx context.Context, BaseURL, APIKey string) er
 }
 func (g *GrafanaClient) makeRequest(ctx context.Context, queryURL, APIKey string) ([]byte, error) {
 	req, err := http.NewRequest(http.MethodGet, queryURL, nil)
+	if err != nil {
```
leecalcote, line 70: The grafana sdk isn’t great. We’ll want to test to confirm behavior remains the same here.

models/load-test.go
```diff
@@ -125,8 +125,6 @@ func (m *MesheryResult) ConvertToSpec() (*BenchmarkSpec, error) {
 		results = httpResults
 		logrus.Debugf("httpresults: %+v", httpResults)
 		b.EndpointURL = httpResults.URL
-	} else {
```
leecalcote, line 128: We do want to add gRPC support. If this else statement is harmful in the meantime, we can remove it.

handlers/session_sync_handler.go
```diff
@@ -39,30 +29,20 @@ func (h *Handler) SessionSyncHandler(w http.ResponseWriter, req *http.Request, _
...
-			// }
 		}
 
 		if len(prefObj.K8SConfig.Nodes) == 0 {
 			// fetching nodes, if it has not already been
 			prefObj.K8SConfig.Nodes, _ = helpers.FetchKubernetesNodes(prefObj.K8SConfig.Config, prefObj.K8SConfig.ContextName)
-			// if err != nil {
-			// 	http.Error(w, "unable to fetch nodes metadata from the kubernetes server", http.StatusInternalServerError)
-			// 	return
-			// }
```
leecalcote, line 65: Should we uncomment instead of deleting? Good to capture this error?

handlers/session_sync_handler.go
```diff
@@ -39,30 +29,20 @@ func (h *Handler) SessionSyncHandler(w http.ResponseWriter, req *http.Request, _
 	}
 	logrus.Debugf("final list of active adapters: %+v", meshAdapters)
 	prefObj.MeshAdapters = meshAdapters
-	err = provider.RecordPreferences(req, user.UserID, prefObj)
+	err := provider.RecordPreferences(req, user.UserID, prefObj)
 	if err != nil { // ignoring errors in this context
 		logrus.Errorf("unable to save session: %v", err)
-		// http.Error(w, "unable to save session", http.StatusInternalServerError)
-		// return
```
leecalcote, line 46: We have a bug in this area now. @nveenjain and @shivaylamba looking at this.

models/load-test.go
kushthedude: We can add it later onwards. No sense of having a blank else block.

handlers/session_sync_handler.go
kushthedude: Any open issue for this reference ?

handlers/session_sync_handler.go
kushthedude: Error is already captured above. There is no new error assignment for this block, hence this block is redundant.

models/grafana_helper.go
kushthedude: I have just captured the error which was neither captured nor reported. This wont have any effect on the sdk.

staticcheck.conf
kushthedude: initiaslisms is the standard name used in the library.

staticcheck.conf
kushthedude: We can hide the file by adding an `.` before file name if we don't want to show on root folder.

kushthedude:
@leecalcote Security & Static checks are restored, please review & merge!

staticcheck.conf
leecalcote: That's nice, yeah. This frustrating bit about this approach is the marketing aspect of the layer5io/meshery repo. When people view that GitHub page, the hidden files still (appropriately) show in that view. I wonder if there's another "tools" folder we might consolidate this and similar files under. 

Not the end of the world. But, just a consideration as we go to create this file (before it gets semi-cemented in-place).

kushthedude: How about putting all such files in `config folder`? They may support relative imports I suppose.

staticcheck.conf
leecalcote: Oh! Well, in that case... 😄

models/grafana_helper.go
leecalcote: Oops. I was on my mobile phone reviewing this PR and thought that you had removed, not added this catch. 

Ok, 👍

handlers/session_sync_handler.go
leecalcote: The behavior of the issue might have morphed slightly over time, but this is the relevant issue - https://github.com/layer5io/meshery/issues/273

kushthedude: I don't think this is related to the issue. There was old error reporting which was commented and replaced, I have removed that part.

staticcheck.conf
kushthedude: @leecalcote Done reorganised the configuration files into tools folder!

kushthedude:
@leecalcote Please take a look again!

tools-config/staticcheck.conf
```diff
@@ -0,0 +1,10 @@
+checks = ["all", "-ST1000", "-ST1003", "-ST1016"]
```
leecalcote, line 1: Thanks for creating this new folder.

PR #830, opened 2020-04-11
Changes for Grafana and Promotheus with revisions.

vijcheru:
Signed-off-by: vijcheru <vijcheru@gmail.com>

#### Description

This PR fixes #    

#### Checklist

- [] [DCO](https://github.com/layer5io/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin) signed.
- [] Title of the PR starts with area name, e.g. `meshery:`, `mesheryctl:`, `ui:`.

### DIFF
--- a/handlers/prometheus_handlers.go
+++ b/handlers/prometheus_handlers.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 
 	"github.com/gorilla/sessions"
+	"github.com/layer5io/meshery/helpers"
 	"github.com/layer5io/meshery/models"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


nveenjain:
@leecalcote , should we also add IP addresses in the response, if so, which IP address should be put there? ClusterIP (if present, if present but set to none, what should be done or can we assume that it will always be present?)

nveenjain:
Current response:-

```json
{"grafana.istio-system":["3000"]}
```

leecalcote:
@nveenjain you're quite correct. Good foresight. Not only will having an IP address provide a better user experience now, but it will be a critical need in the future as we bring forth the concept of Environments and begin providing features to explicitly manage multiple clusters, multiple data planes under a common control plane, and multiple, integrated control planes.

For now, I'll suggest that we accept @vijcheru's pull request, and move forward with a design spec to identify a robust approach to which will be a complex challenge given the number of supported platforms. In advance of supporting multiple Environments, we'll want to develop routines that when given a service (e.g. Grafana) and port number can investigate the service mesh and suggest a full endpoint (with IP address or FQDN) to access the service external to the service mesh (access could still be limited to the cluster, but off-mesh).

leecalcote:
🤩 @vijcheru

PR #832, opened 2020-04-13
feat: Restore Go_Tests and finalise CI Actions

### LINKED ISSUE #807: Add back the CI jobs e.g `static_check` `vet` `sec_check ` and `tests `
**Current Behavior**

The CI job of `meshery` was commented out. I believe we should add them back. Because of the unit test cases need to trigger by these CI jobs. And the CI config we can reference `meshery-linkerd` project.

* `static_check `
* `vet`
* `sec_check `
* `tests `

**Desired Behavior**

All CI jobs work well

---
**Resources**

**Alternatives / Additional Context**

kushthedude:
Fixes #807

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
         with:
           args: ./... -exclude=G301,G304,G107
   tests:
-    needs: [lint, error_check, static_check, vet, security_check]
+    needs: [lint, error_check, static_check, vet_check, security_check]
     name: Tests
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


kushthedude:
@leecalcote Please review & merge. Also restored the test dependencies on each other for better CI Actions Flow of our tests!

kushthedude:
> I think this looks right. I'm interested to see how the collective set of activities we're incorporating into CI shapes up when all executed.
> 
> Are we able to exclude execution of Netlify build and mesheryctl build if no changes to /docs or /mesheryctl, respectively?

I will check into it and see if we can exclude execution of some of the jobs when the changes are not concerned for them.

Issue #827, opened 2020-04-11
Addition of Spell Test in CI


kushthedude:
Addition Go-SpellTest in the current CI-Workflow.

___

### Resources -:
https://github.com/client9/misspell
https://github.com/client9/misspell#gometalinter

kushthedude:
@leecalcote How does it sounds?

nveenjain:
+1 for the spell test! While misspell is great,  I do have a suggestion:- Can we use https://github.com/streetsidesoftware/cspell/tree/master/packages/cspell, it also has support for almost all programming languages and some common computer science terms and is extensible (just need to add a file in repo). We can also propose our own terms to https://github.com/streetsidesoftware/cspell-dicts

kushthedude:
> https://github.com/streetsidesoftware/cspell/tree/master/packages/cspell

Its npm engine, How will you support go with it? Also, Eslint has spell check plugin.

nveenjain:
@kushthedude , it installs using npm, but it supports every major language that we are using in the project (go, javascript, html,css)

https://github.com/streetsidesoftware/cspell/tree/master/packages/cspell#features

It also has extension for VSCode, and can be easily integrated with the CI pipeline.

kushthedude:
No, thats what not I am getting.
To use in GitHub Actions you will first install node and then npm and then grab the dependency with the npm engine. 
Cool, If this is what we need to do at place of a single line of go get _____.

nveenjain:
I'm suggesting this for entire codebase(JS,go based code, CSS files,etc) not for only go based code. (for the complete file, not only comments, cspell, as I see, checks only for comments). We already install node and can install this globally... 
https://github.com/meshery/meshery/blob/master/.github/workflows/ci.yml#L122

leecalcote:
Stale issue? Please reopen at your leisure.

PR #840, opened 2020-04-13
CI: Added a spell check for go files

nveenjain:
Refers #827, the spell test for Javascript based files needs to be done.
Did the initial setup for spell check.

#### Notes to reviewers:
This is not a complete fix, there needs to be a PR opened for JS based files. And then we can create a pre commit hook
#### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,21 @@ on:
       - master
 
 jobs:
+  spell-check:
+    name: Check spelling in code
+    runs-on: ubuntu-latest
+    # only checks go code for now
+    steps:
+    - name: Check out code
+      uses: actions/checkout@master
+      with:
+        fetch-depth: 1
+    - uses: actions/setup-node@v1
+      with:
+        node-version: ${{ secrets.NODE_VERSION }}
+    - run: |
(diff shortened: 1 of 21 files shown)


leecalcote:
I wonder if spell check makes sense in context of ensuring that this job is only executed when files are changed within the /docs and /ui directories (and exclude the running of this job for changes in all other directories). I'm not sure. 

It is worth noting that we've injected a number of new steps in our CI workflow recently, and in advance of breaking out components in to separate repos (e.g. /docs), would do well to consider if we can invoke CI jobs based on file changes in specific directories.

leecalcote:
After discussion in the Meshery Development Meeting today, we'll go forth with evaluation of this check in `meshery-tmp`. 👍

leecalcote:
@nveenjain irrespective of whether the spell checker is incorporated, there are many misspellings that you're corrected in this PR, which is great. For now, would you like to submit an alternative PR with just the spelling corrections?

nveenjain:
Sure @leecalcote, I'll open another PR and one pr in the meshery-tmp repo

kushthedude:
I think we are good to close this.

Issue #829, opened 2020-04-11
Remove docker-compose check for Windows 


anandrkskd:
**Current Behavior**
Currently we have a wrong check for installed docker-compose on the system.  

**Expected Behavior**
No checks are required for the docker-compose on Windows.

PR #845, opened 2020-04-13
docker-compose check for Windows removed

anandrkskd:
This PR resolves issue #829 by removing check of docker-compose for windows based system.

### DIFF
--- a/mesheryctl/cmd/vars.go
+++ b/mesheryctl/cmd/vars.go
@@ -24,6 +24,7 @@ import (
 	"os/exec"
 	"path"
 	"strings"
+	"runtime"
 
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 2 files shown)


mesheryctl/cmd/vars.go
```diff
@@ -92,6 +92,10 @@ func setFileLocation() {
 
 //Pre-Flight Check
 func preReqCheck() {
+	//No checks for Docker-compose for windows
+	if runtime.GOOS == "windows" {
+		return
+		}
 	//Check for installed docker-compose on client system
```
leecalcote, line 99: @anandrkskd you'll want to double-check on a Windows machine, but I think we will still want to run a check for `docker-compose` as a prerequisite, let the user know if a prerequisite is unmet, but just not try to automatically install that unmet prereq (not run `installprereq()`).

anandrkskd: Understood

PR #851, opened 2020-04-15
[mesheryctl] add `--reset` flag to `mesheryctl stop`

### LINKED ISSUE #850: [mesheryctl] Add `--reset` flag to `mesheryctl stop`
**Current Behavior**
Currently, `mesheryctl stop` stops all Meshery containers, removes their instances, prunes their connected volumes, and resets the meshery.yaml config back to default settings.

**Desired Behavior**
Change mesheryctl stop behavior to not reset meshery.yaml unless explicitly told to do so with an additional `--reset` flag. This flag will invoke the `cleanup` command, which should soon be renamed `reset`.

leecalcote:

fixes #850

### DIFF
--- a/docs/pages/guides/mesheryctl.md
+++ b/docs/pages/guides/mesheryctl.md
@@ -9,34 +9,36 @@ type: guide
 ### Meshery Lifecycle Management
 Installation, troubleshooting and debugging of Meshery and its adapaters.
 
-| command   | flag  | function                  | Usage                     |
-|:----------|:-----:|:--------------------------|:--------------------------|
-|cleanup    |       |Pulls current meshery.yaml from meshery repo. *Warning: Any local changes will be overwritten.* | `mesheryctl cleanup` |
-|log        |       |Starts tailing Meshery debug logs.              | `mesheryctl log` |
-|start      |       |Start all Meshery containers.   | `mesheryctl start` |
-|status     |       |Displays the status of Meshery's containers.       | `mesheryctl status` |
-|stop       |       |Stop all Meshery containers.    | `mesheryctl stop` |
+| command   | flag      | function                  | Usage                     |
+|:----------|:---------:|:--------------------------|:--------------------------|
+|cleanup    |           | Pulls current meshery.yaml from meshery repo. *Warning: Any local changes will be overwritten.* | `mesheryctl cleanup` |
+|log        |           | Starts tailing Meshery debug logs.                    | `mesheryctl log` |
+|start      |           | Start all Meshery containers.                         | `mesheryctl start` |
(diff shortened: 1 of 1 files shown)


docs/pages/overview.md
```diff
@@ -3,12 +3,14 @@ layout: page
...
 permalink: overview
 ---
-# Introducing Meshery 
-The multi-service mesh management plane adopting, operating and developing on different service meshes. 
-Meshery facilitates learning about functionality and performance of service meshes and incorporates the collection and display of metrics from applications running on or across service meshes. 
-Meshery provides this high-level functionality: 
-
-1. Performance Benchmarking
+# Introducing Meshery
+The service mesh management plane adopting, operating and developing on different service meshes.
```
kushthedude, line 7: ~Adopting~ Adapting would suit better here.

mesheryctl/cmd/cleanup.go
```diff
@@ -22,14 +22,14 @@ import (
 // cleanupCmd represents the cleanup command
 var cleanupCmd = &cobra.Command{
 	Use:   "cleanup",
-	Short: "Clean up Meshery",
-	Long:  `Clean up Meshery configuration`,
+	Short: "Cleanup Meshery's configuration",
+	Long:  `Reset Meshery to it's default configuration.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		log.Info("Cleaning Meshery config...")
+		log.Info("Reseting Meshery config...")
```
kushthedude, line 28: Resetting*

leecalcote: Ha! Good eye.

leecalcote:
@anandrkskd yes, it doesn't hurt to provide this `--reset` flag there as well. I suppose folks are about equally as likely to use the flag in either place.

Two other items are needed in this same context:

1) a renaming of `cleanup` to `reset` as the name of the command.
2) confirmation that `--config <file>` works with `--start`. In my tests, it doesn't.

docs/pages/overview.md
leecalcote: Good comment given the pervasiveness of adapters in this project. In this case, we're trying to describe part of the initial value that people get when using the project (certainly adapters to different meshes is one of those), the first of those starting with adoption of a service mesh (facilitating understanding of how they work and selection of which one to use).

leecalcote:
> Two other items are needed in this same context:
> 
> 1. a renaming of `cleanup` to `reset` as the name of the command.
> 2. confirmation that `--config <file>` works with `--start`. In my tests, it doesn't.

@anandrkskd would you be so kind as to file issues on these two items?

anandrkskd:
Sure @leecalcote

PR #857, opened 2020-04-17
fix: User Preferences statistics toggle

### LINKED ISSUE #676: [UI] User Preference selections not remembered
**Description**
User selections are not reflected accurately after the user navigates away and comes back to the Preferences screen.

**Expected Behavior**
Users should see their previous selections when returning to the Preferences screen.

**Screenshots**
Untitled

**Related**
This issue is related to #599

kushthedude:
Fixes #676


kushthedude:
Yes this is just the fix at frontend.
Currently, server endpoint issues null response with hit at api/user/stat .
I will push a fix for the same in another PR or I can merge both.

kushthedude:
Closed in favour of #897

PR #862, opened 2020-04-18
meshery: Fixed unexpected 404 while reloading adapter's page

### LINKED ISSUE #861: Unexpected 404 when reloading adapter page
**Description**

404: Not found is displayed when we reload any adapter's page

**Expected Behavior**

It should reload with adapter's setting if it is available.

**Screenshots**

**Environment:**
 - OS: Any
 - Browser: Any
 - Version: [Any
 - Device: Any

---
[Optional] **To Reproduce**
1. Install meshery
2. Click on any adapter
3. Reload
4. See the unexpected 404

[Optional] **Additional Context**
NA

nveenjain:

#### Description

This PR fixes #861     

#### Checklist

### DIFF
--- a/handlers/ui_handler.go
+++ b/handlers/ui_handler.go
@@ -12,13 +12,13 @@ func ServeUI(w http.ResponseWriter, r *http.Request, reqBasePath, baseFolderPath
 		http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
 		return
 	}
-	reqURL := r.URL.RequestURI()
+	reqURL := r.URL.EscapedPath()
 	reqURL = strings.Replace(reqURL, reqBasePath, "", 1)
 	var filePath strings.Builder
 	filePath.WriteString(reqURL)
 	if reqURL == "/" || reqURL == "" {
 		filePath.WriteString("index.html")
-	} else if filepath.Ext(r.URL.RequestURI()) == "" {
+	} else if filepath.Ext(reqURL) == "" {
 		filePath.WriteString(".html")
 	}
 	finalPath := filepath.Join(baseFolderPath, filePath.String())


leecalcote:
@nveenjain this is an improvement for sure. One item to note is that when the page is refreshed, users don't always land back on the adapter that they have selected. 

Untitled

leecalcote:
I strongly suspect this issue is do to out of order messages received through the go channel used by go routines to query all the adapters for their metadata. Assuming this is the case, this article might be of value - https://medium.com/@fonseka.live/how-to-maintain-the-order-of-go-routines-117a5be86c4f

leecalcote:
https://github.com/layer5io/meshery/issues/871 opened to track this issue. Please self-assign, if interested in fixing.

kushthedude:
@leecalcote @nveenjain Can you please me reproducing steps of the issue this PR has addresed ? 
I have been trying for a while now but I didn't encountered issue #861 . Also I have already reverted this PR on my local. It would be great if you may guide me where is it giving 404 after refreshing?
I tried to refresh page `http://localhost:3000/management?adapter=meshery-adapter_name:adapter_port` but there was no error has stated in the issue .

leecalcote:
@kushthedude, you bet.  I think we're talking about two items here. Repro steps for each:

1. Prior to the fix for https://github.com/layer5io/meshery/issues/861, it should be the case that when you nav to one of the adapter's (e.g. http://localhost:9081/management?adapter=meshery-linkerd:10001), then refresh your page, you would arrive a 404. Reminder: you may need to `make build-ui`, then `make run-ui-dev`.

1. Prior to PR https://github.com/layer5io/meshery/pull/872, it should be the case that while visiting a specific type of service mesh adapter (e.g. http://localhost:9081/management?adapter=meshery-linkerd:10001), when you refresh the page repeatedly, you will see different service meshes as being selected, not the selection you originally made.

kushthedude:
> @kushthedude, you bet. I think we're talking about two items here. Repro steps for each:
> 
> 1. Prior to the fix for #861, it should be the case that when you nav to one of the adapter's (e.g. http://localhost:9081/management?adapter=meshery-linkerd:10001), then refresh your page, you would arrive a 404. Reminder: you may need to `make build-ui`, then `make run-ui-dev`.


I am able to produce the scenario before #872, but alas I am still unable to reproduce the scenario-1 which is before #861 . PS: I ran `make build-ui` before `make run-ui-dev` .

nveenjain:
@kushthedude , it's actually working fine in development mode of ui. Hence working in 3000 port, but doesn't work after site is exported using next.

nveenjain:
You would need to run `make run-local` and access the site from 9081 port...

kushthedude:
> You would need to run `make run-local` and access the site from 9081 port...

I think shouldn't accessing from `make run-ui` also throw the error. Because the behaviour which I am encountering before #862 is similar to the scenario-2 described by @leecalcote above, which makes me more curious about why there is such mis-alignment b/w ui & server rendered ui ?

nveenjain:
Actually management.html is rendered by next js, there is no other file for management page, the server points /management?adapter=________ to management.html rendered in the out folder rendered by next. It should parse the query parameters from window location, which were ignore before #872.

leecalcote:
@kushthedude right - `make run local`. My theory is that part of this misalignment is due to out of order messages received through the go channel used by go routines to query all the adapters for their metadata (adapter information being returned in different order on each page load) from the REST call. Following on this hunch, the UI should be matching on adapter name / menu item name, not an unstable array.

PR #865, opened 2020-04-19
[docs] adapter updates: initial NSM, tweaks on Octarine, rename NSX SM

leecalcote:

Also included:

- screenshot of environment metrics shown with perf metrics.
- recording of multi-cluster demo


leecalcote:
@schillingjerrid you might be interested in the netlify integration here, which we're using to provide a convenient preview build (see the URL in the checks above)

PR #911, opened 2020-04-28
mesheryCTL: Added support for accpeting file as input in perf command

### LINKED ISSUE #898: mesheryctl: Import test config using perf --f
**Current Behavior**

Currently, there is no mechanism to import test config from a file. SMPS also stores the test config. 

**Desired Behavior**

Import test config from SMPS yaml file.

nveenjain:

#### Description

This PR fixes #898    

#### Checklist

### DIFF
--- a/mesheryctl/cmd/perf.go
+++ b/mesheryctl/cmd/perf.go
@@ -51,20 +51,22 @@ Usage:
   mesheryctl perf --[flags]
 
 Available Flags for Performance Command:
-  name[string]                  (optional) Name for the Test, if not provided random name will be used.
-  url[string]                   (required) URL Endpoint at which test is to be performed
-  duration[string]              (required) Duration for which test should be performed. See standard notation https://golang.org/pkg/time/#ParseDuration
-  load-generator[string]        (optional) Load-Generator to be used to perform test.(fortio/wrk2) (Default "fortio")
-  mesh[string]              	(optional) Name of the service mesh to be tested.
-  cookie[string]            	(required) Choice of the cloud server provider (Default "Default Local Provider")
-  concurrent-requests[string]   (required) Number of paraller requests to be used (Default "1")
-  qps[string]                   (required) Queries per second (Default "0")
-  file[string]					(optional) file containing SMPS-compatible test configuration
+  name[string]                  (optional) A short descriptor to serve as reference for this test. If not provided, a random name will be generate.
+  url[string]                   (required) URL endpoint to send requests.
+  duration[string]              (required) Length of time to perform test (e.g 30s, 15m, 1hr). See standard notation https://golang.org/pkg/time/#ParseDuration
+  load-generator[string]        (optional) Name of load generator to be used to perform test (default: "fortio")
+  mesh[string]              	(optional) Name of the service mesh to be tested (default: "None")
+  provider[string]            	(required) Choice of Provider (default: "Meshery")
+  concurrent-requests[string]   (required) Number of parallel requests to be sent (default: "1")
+  qps[string]                   (required) Queries per second (default: "0")
+  file[string]			(optional) file containing SMPS-compatible test configuration. See https://github.com/layer5io/service-mesh-performance-specification
   help                          Help for perf subcommand
 
-url,duration,concurrent-requests,qps are optional as flag inputs if specified through an SMPS compatible yaml file
-Example usage of Performance Sub-command :-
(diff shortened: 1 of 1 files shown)


leecalcote:
yes, yes, yes, yes

mesheryctl/cmd/perf.go
```diff
@@ -86,7 +91,18 @@ var perfCmd = &cobra.Command{
 			log.Print(perfDetails)
 			return
 		}
+		if filePath != "" {
+			var t models.BenchmarkSpec
```
leecalcote, line 95: A nit: where we can, it may serve us well to use the term PerformanceSpec as opposed to BenchmarkSpec. In doing so, we'll expand the implicit description of this Spec's applicability to use cases beyond benchmarking to include a number of other performance-related concerns.

nveenjain: Done!

mesheryctl/cmd/perf.go
```diff
@@ -86,7 +91,18 @@ var perfCmd = &cobra.Command{
 			log.Print(perfDetails)
 			return
 		}
+		if filePath != "" {
+			var t models.PerformanceSpec
+			err := yaml.Unmarshal([]byte(filePath), &t)
 
+			if err != nil {
+				log.Errorf("Error: File Supplied is invalid.\n%v", err)
```
leecalcote, line 99: Question: Would "Error: The file provided contains invalid yaml." be a more descriptive (accurate) error to provide the user?

mesheryctl/cmd/perf.go
nveenjain: This will be breaking change for those importing meshery through go though... But this migration is necessary, as it would be better to move PerformanceSpec go file in SMPS repository and have symlink present here only.

mesheryctl/cmd/perf.go
```diff
@@ -86,7 +91,18 @@ var perfCmd = &cobra.Command{
...
+			}
+			testDuration = fmt.Sprintf("%fs", t.EndTime.Sub(t.StartTime).Seconds())
```
leecalcote, line 101: I suppose we'll want to go through the hassle of verifying the presence of each of these and fail if any required attributes are missing.

mesheryctl/cmd/perf.go
nveenjain: Yeah, but the error might be occurring if the file was not present there or there were no permissions present... Exact reason of error will be printed in the next line... should we write "Invalid yaml file" instead? or keep it as is?

mesheryctl/cmd/perf.go
nveenjain: Nice catch!!! Thanks a lot @leecalcote !!

nveenjain: Actually Lee, it would already be tested in the next few lines, such as to check whether URL exist or not:- https://github.com/layer5io/meshery/pull/911/files#diff-95c2c60d8d4612976d9000d256797085L121-L125

mesheryctl/cmd/perf.go
```diff
@@ -175,5 +190,6 @@ func init() {
 	perfCmd.Flags().StringVar(&testDuration, "duration", "30s", "(optional) Length of test (e.g. 10s, 5m, 2h). For more, see https://golang.org/pkg/time/#ParseDuration")
```
nveenjain, line 190: Note to me:-
Endpoint URL is still required PARAM, it should be optional, if file is supplied.

mesheryctl/cmd/perf.go
leecalcote: Oh, I see. Yes,  "Invalid yaml file" sounds good, if we leave this to only one check. If we were to break into two checks - one for presence of a file and the other for validity of yaml - we could have two errors. I don't know that it's necessary to breakout into two checks. "Invalid yaml file" does cover them both.

mesheryctl/cmd/perf.go
```diff
@@ -54,8 +59,10 @@ Available Flags for Performance Command:
   cookie[string]            	(required) Choice of the cloud server provider (Default "Default Local Provider")
```
leecalcote, line 59: @kanishkarj we need to figure out how users will provide their JWT via CLI.

kushthedude:
CI broke due to this PR. Checking into it

PR #918, opened 2020-04-28
UI: Add deep linking for tabs in Settings page

### LINKED ISSUE #670: UI: Provide anchor on each Settings tab for deep-linking
**Current Behavior**
A user visiting the /settings page in Meshery may want to land within one of the Settings tabs (Environment, Service Meshes, Metrics). 

As an example, currently, if a user navigates to http://meshery/settings#service-meshes, the "Service Meshes" tab is not open.

**Desired Behavior**
Named anchor (<a>) tags should be added to each Settings screen tab, like so:

- `<a name="environment"> </a>`
- `<a name="service-meshes"> </a>`
- `<a name="metrics"> </a>`

**Reference**
```
<button class="MuiButtonBase-root MuiTab-root MuiTab-textColorPrimary Mui-selected MuiTab-labelIcon MuiTab-fullWidth" tabindex="0" type="button" role="tab" aria-selected="true" title="Connect Meshery Adapters">
```
**Screenshot**
Screen Shot 2020-02-12 at 10 40 32 PM

aayushdutt:
Signed-off-by: Aayush Dutt <aayushd100@gmail.com>

#### Description

Add deep linking in Settings page for environtment, service-mesh, and metrics tabs
This PR fixes #670    

#### Notes to reviewers:
Not sure if this is the best way to achieve this. Reviews/suggestions are welcome.

#### Checklist

### DIFF
--- a/ui/components/MesherySettings.js
+++ b/ui/components/MesherySettings.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { withStyles } from '@material-ui/core/styles';
+import { withRouter } from 'next/router';
 import { connect } from 'react-redux';
 import Tabs from '@material-ui/core/Tabs';
 import Tab from '@material-ui/core/Tab';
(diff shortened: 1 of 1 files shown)


leecalcote:
Hello! This is exciting. 😄

leecalcote:
A good way to put this new feature to use it to change from this:

Screen Shot 2020-04-29 at 12 39 45 AM

to this:

Screen Shot 2020-04-29 at 12 39 45 AM

aayushdutt:
> A good way to put this new feature to use it to change from this:
> 
> Screen Shot 2020-04-29 at 12 39 45 AM
> 
> to this:
> 
> Screen Shot 2020-04-29 at 12 39 45 AM
> 
> The component used here, doesn't need to be a button.

Good suggestion. I'll put that up in a new issue

aayushdutt:
> Tested. This works!
> 
> I don't know that it would be an improvement to dynamically name anchors based on the name of the tabs. I don't think doing so makes the code any less fragile or easy to sustain.

@leecalcote I agree that would be better
Here are some considerations though:
* We will need to re-render the page since the names of the tabs will not be known to the constructor.
* Current structure of the page is not using name attributes, instead, it is using the imported component and rendering is based on the tab index in state.

PR #922, opened 2020-04-28
mesheryctl: Feature/cli login

### LINKED ISSUE #913: [mesheryctl] Support use of JWTs instead of Cookies
**Description**
The move away from cookies to jwt needs to be incorporated into `mesheryctl`.

**Expected Behavior**
- Both Local and Remote Providers should be supported.
- "None" provider should implicitly send anonymous test results.

**Future Behavior**
"None" provider should have option to disable this telemetry.

kanishkarj:
#### Description

This PR fixes #913    

#### Notes to reviewers:
To test : 
- rebuild meshery UI
- in the dropdown when you press the user profile pic, there will be an option "Get Token"
- that will raise a popup to download the token, go ahead and download it.
- build mesheryctl and run : `./mesheryctl perf --name "a quick stress test" --url https://www.github.com --qps 1 --concurrent-requests 1 --duration 1m --cred <PATH-TO-AUTH.JSON>`
#### Checklist

- [] [DCO](https://github.com/layer5io/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin) signed.
- [] Title of the PR starts with area name, e.g. `meshery:`, `mesheryctl:`, `ui:`.

### DIFF
--- a/mesheryctl/cmd/perf.go
+++ b/mesheryctl/cmd/perf.go
@@ -85,6 +85,7 @@ func StringWithCharset(length int) string {
 	return string(b)
 }
 
+// AddAuthDetails Adds authentication cookies to the request
 func AddAuthDetails(req *http.Request, filepath string) error {
 	file, err := ioutil.ReadFile(filepath)
 	if err != nil {
@@ -109,6 +110,7 @@ func AddAuthDetails(req *http.Request, filepath string) error {
 	return nil
 }
 
+// UpdateAuthDetails checks gets the token (old/refreshed) from meshery server and writes it back to the config file
 func UpdateAuthDetails(filepath string) error {
 	req, err := http.NewRequest("GET", mesheryAuthToken, bytes.NewBuffer([]byte("")))
 	if err != nil {
@@ -131,10 +133,7 @@ func UpdateAuthDetails(filepath string) error {
 		return err
 	}
 
-	if err := ioutil.WriteFile(filepath, data, os.ModePerm); err != nil {
-		return err
-	}
-	return nil
+	return ioutil.WriteFile(filepath, data, os.ModePerm)
 }
 
 // perfCmd represents the Performance command
@@ -146,10 +145,7 @@ var perfCmd = &cobra.Command{
 		//Check prerequisite
 		preReqCheck()
 
-		// if len(args) == 0 {
-		// log.Print(perfDetails)
-		// return
-		// }
+		// Importing SMPS Configuration from the file
 		if filePath != "" {
 			var t models.PerformanceSpec
 			err := yaml.Unmarshal([]byte(filePath), &t)
@@ -194,7 +190,7 @@ var perfCmd = &cobra.Command{
 		if len(testURL) > 0 {
 			postData = postData + "\nendpoint_url: " + testURL
 		} else {
-			log.Fatal("\nError: Please enter a test URL")
+			log.Fatal("Error: Please enter a test URL")
 			return
 		}
 
@@ -212,7 +208,7 @@ var perfCmd = &cobra.Command{
 
 		req, err := http.NewRequest("POST", mesheryURL, bytes.NewBuffer([]byte(postData)))
 		if err != nil {
-			log.Print("\nError in building the request")
+			log.Print("Error in building the request")
 			log.Fatal("Error Message:\n", err)
 			return
 		}
@@ -244,7 +240,7 @@ var perfCmd = &cobra.Command{
 		defer SafeClose(resp.Body)
 		data, err := ioutil.ReadAll(resp.Body)
 		if err != nil {
-			log.Printf("Erro reading body: %v", err.Error())
+			log.Printf("Error reading body: %v", err.Error())
 			return
 		}
 		log.Print(string(data))
--- a/models/meshery_remote_provider.go
+++ b/models/meshery_remote_provider.go
@@ -469,7 +469,11 @@ func (l *MesheryRemoteProvider) ExtractToken(w http.ResponseWriter, r *http.Requ
 	l.TokenStoreMut.Lock()
 	defer l.TokenStoreMut.Unlock()
 
-	tokenString, _ := l.GetToken(r)
+	tokenString, err := l.GetToken(r)
+	if err != nil {
+		logrus.Errorf("Token not found:  %s", err.Error())
+		return
+	}
 	newts := l.TokenStore[tokenString]
 	if newts != "" {
 		tokenString = newts


leecalcote:
The instructions to reviewers is helpful. 📋 

Will kick the build now...

leecalcote:
This fantastic! 

Output in my terminal is squished to a few columns:

Screen Shot 2020-04-28 at 10 53 59 PM

leecalcote:
To the extent that we use the term "Token" in the UI (which makes sense), we may want to use the same vernacular in `mesheryctl --token <path to auth.json>`.

kanishkarj:
That's something even I noticed, for some reason the Mesheryctl output doesn't seem good at all.. like unnecessary new lines etc.. but if we're good to go I'll do the refactoring.

kanishkarj:
> To the extent that we use the term "Token" in the UI (which makes sense), we may want to use the same vernacular in `mesheryctl --token <path to auth.json>`.

Gotcha!

mesheryctl/cmd/perf.go
```diff
@@ -82,6 +84,52 @@ func StringWithCharset(length int) string {
...
+
+	data, _ := ioutil.ReadAll(resp.Body)
+	log.Printf("%v", string(data))
+	ioutil.WriteFile(filename, data, os.ModePerm)
+	return nil
+}
+
+func defaultAuthConfig() string {
+	// Find home directory.
+	home, err := os.UserHomeDir()
```
leecalcote, line 125: I wonder if we can reuse this function if we have it return the location - [setFileLocation](https://github.com/layer5io/meshery/blob/master/mesheryctl/cmd/vars.go#L41) or if we create a getFileLocation.

mesheryctl/cmd/perf.go
```diff
@@ -82,6 +84,52 @@ func StringWithCharset(length int) string {
...
+	if err != nil {
+		log.Fatal("[ERROR] Cannot determine location of $HOME")
+	}
+	mesheryFolder = path.Join(home, mesheryFolder)
+	return path.Join(mesheryFolder, "auth.json")
```
leecalcote, line 130: Perhaps, we shouldn't mandate that the filename be `auth.json`?

router/server.go
```diff
@@ -96,6 +96,15 @@ func NewRouter(ctx context.Context, h models.HandlerInterface, port int) *Router
 		}
 		h.TokenHandler(w, req, provider, false)
 	})))
+	mux.Handle("/api/gettoken", h.ProviderMiddleware(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
```
leecalcote, line 99: Do we need to be concerned with protecting this endpoint (i.e. `h.AuthMiddleware(h.SessionInjectorMiddleware())`)?

leecalcote:
Tested and confirmed that both the "Meshery" provider and the "None" provider work in this PR.

kanishkarj:
@leecalcote I have made a few more updates and refactored the code. Do let me know if anything is left.

models/default_local_provider.go
```diff
@@ -239,3 +239,16 @@ func (l *DefaultLocalProvider) UpdateToken(http.ResponseWriter, *http.Request) {
...
+// ExtractToken - Returns the auth token and the provider type
+func (l *DefaultLocalProvider) ExtractToken(w http.ResponseWriter, r *http.Request) {
+	resp := map[string]interface{}{
+		"meshery-provider": l.Name(),
+		tokenName:          "",
+	}
+	logrus.Debugf("encoded response : %v", resp)
+	if err := json.NewEncoder(w).Encode(resp); err != nil {
+		logrus.Errorf("Unable to extract auth details: %v", err)
+		http.Error(w, "unable to extract auth details", http.StatusInternalServerError)
```
nveenjain, line 252: I wonder whether it should be `http.StatusInternalServerError` or `http.StatusUnauthorized`

nveenjain: It's opinionated though... Leave it as is if you want...

mesheryctl/cmd/perf.go
```diff
@@ -82,6 +85,58 @@ func StringWithCharset(length int) string {
...
+		return err
+	}
+
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	if err := ioutil.WriteFile(filepath, data, os.ModePerm); err != nil {
+		return err
```
nveenjain, line 135: No need to check whether err!= nil,
Please return error itself in the next line

mesheryctl/cmd/perf.go
```diff
@@ -82,6 +85,58 @@ func StringWithCharset(length int) string {
...
+	})
+	req.AddCookie(&http.Cookie{
+		Name:     providerName,
+		Value:    tokenObj[providerName],
+		HttpOnly: true,
+	})
+	return nil
+}
+
+func UpdateAuthDetails(filepath string) error {
```
nveenjain, line 112: Please add a comment here (all exported functions must have comments starting with function names)

models/meshery_remote_provider.go
```diff
@@ -463,3 +463,25 @@ func (l *MesheryRemoteProvider) UpdateToken(w http.ResponseWriter, r *http.Reque
 		})
 	}
 }
+
+// ExtractToken - Returns the auth token and the provider type
+func (l *MesheryRemoteProvider) ExtractToken(w http.ResponseWriter, r *http.Request) {
+	l.TokenStoreMut.Lock()
+	defer l.TokenStoreMut.Unlock()
+
+	tokenString, _ := l.GetToken(r)
```
nveenjain, line 472: Please don't ignore the error instead, you can log in the next line maybe... (Error check might fail...)

mesheryctl/cmd/helpers.go
```diff
@@ -0,0 +1,14 @@
+package cmd
+
+import (
+	"io"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// SafeClose is a helper function help to close the io
+func SafeClose(co io.Closer) {
```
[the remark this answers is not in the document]
kanishkarj: I agree! I was thinking of this since a while. But should we create another issue for it? in which we will pull off all helper functions into a separate module?

models/default_local_provider.go
kanishkarj: Actually, if the user was unauthorized this function won't be called at all right. The middlewares do check if the users are authorized or not right ?

kanishkarj:
@nveenjain Thanks for the comments, go through the changes!

mesheryctl/cmd/helpers.go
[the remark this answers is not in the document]
nveenjain: Yeah, a separate issue would be great too!

models/default_local_provider.go
leecalcote: @thakursaurabh1998, it would be good to get your eyes here.

nveenjain: Agreed... But if token is invalid, I prefer it to be StatusUnauthorized... Anyways, this is a no-blocker and highly opinionated... This issue shouldn't stop this PR from getting merged.

PR #914, opened 2020-04-28
fix: go-releaser for mesheryctl & model rename

kushthedude:
Fixes #915
/cc @leecalcote

PR #923, opened 2020-04-28
fix: Discrepancies in go.sum after pulling master

kushthedude:


kushthedude:
@leecalcote expect CI to fail as releaser will be fix with #914

Issue #910, opened 2020-04-27
Convert the `no-used-var` & `no-undef` rule to error


kushthedude:
**Description**

We have recently set up ES-Lint rules for our client-side code. Due to poor code quality & to begin with, we had to use these rules as optional so that error does not occur. But, this is one of the most occurring error in our codebase of the client-side.

**Expected Behavior**

Remove all `unused-var` & `no-undef` variables to accommodate to extend the ES-Lint rules.

** Work on Local**

To check this error on local:-

https://github.com/layer5io/meshery/blob/3f21f32ab9008d9a2410221f08fcb0a79ccbcf9e/ui/.eslintrc.js#L50
https://github.com/layer5io/meshery/blob/3f21f32ab9008d9a2410221f08fcb0a79ccbcf9e/ui/.eslintrc.js#L42
Change the `warn` & `off` to `error` respectively and then run `npm test` to see all the places where we have this bug.

**Additional Context**

This will require a lot of files to change. New contributors & old contributors are encouraged to pick-up a file or a component so that the progress can be subsequently increased.

kushthedude:
@aayushdutt Would you like to take care of remaining errors ?

aayushdutt:
@kushthedude  I'd love to. Not my highest priority at the moment though.

PR #936, opened 2020-04-30
fix: Complete the provider-ui eslint rules

kushthedude:
Refers #910 
- Restore all major rules for Provider-UI
- Get rids of half of the warning (total more than 200)
- Half of the work still remaining on UI part which will be done in subsequent PR.

### DIFF
--- a/ui/components/GrafanaComponent.js
+++ b/ui/components/GrafanaComponent.js
@@ -74,7 +74,6 @@ class GrafanaComponent extends Component {
       grafanaBoardSearch: '',
       grafanaURL,
       grafanaAPIKey,
-      grafanaBoardSearch,
       grafanaBoards,
       selectedBoardsConfigs,
       ts: new Date(),


ui/components/GrafanaComponent.js
```diff
@@ -74,7 +74,7 @@ class GrafanaComponent extends Component {
 
       grafanaURL,
       grafanaAPIKey,
-      grafanaBoardSearch: '', // we probably dont need this retrieved from store
+      grafanaBoardSearch,
```
aayushdutt, line 77: removing the default value might cause bugs. Did you absolutely make sure it wasn't required?
I think it is being used in line 177 of the file.
Also, look at line 184 (https://github.com/layer5io/meshery/blob/a9ee4fd229c54c2de7dc0699dcd8501e8d0f2264/ui/components/GrafanaComponent.js#L184), we don't want to send undefined in the URL.

ui/components/GrafanaComponent.js
```diff
@@ -202,7 +200,7 @@ class GrafanaComponent extends Component {
         }, self.handleError('There was an error communicating with Grafana'));
       }
 
-      handleError = (msg) => (error) => {
```
aayushdutt, line 205: Don't think entirely removing the nested function is right. It might be better if we did something like
`handleError = msg => () => {...`

ui/components/GrafanaCustomChart.js
```diff
@@ -425,7 +425,7 @@ class GrafanaCustomChart extends Component {
       const processReceivedData = (result) => {
         self.props.updateProgress({ showProgress: false });
         if (typeof result !== 'undefined') {
-          const fullData = self.transformDataForChart(result, target);
```
aayushdutt, line 428: Why remove the target?

ui/components/GrafanaCustomChart.js
```diff
@@ -493,7 +493,7 @@ class GrafanaCustomChart extends Component {
       }
     }
 
-    transformDataForChart(data, target) {
```
aayushdutt, line 496: Why remove the target? See https://github.com/layer5io/meshery/blob/a9ee4fd229c54c2de7dc0699dcd8501e8d0f2264/ui/components/GrafanaCustomChart.js#L409

ui/components/MesheryAdapterPlayComponent.js
```diff
@@ -406,8 +386,6 @@ class MesheryAdapterPlayComponent extends React.Component {
                     self.setState({ selectionError: false, cmEditorValAddError: false });
                   }
                 }}
-                onChange={(editor, data, value) => {
```
aayushdutt, line 409: Are you sure it doesn't exist to override any existing onChange handler of the component?

aayushdutt:
@kushthedude I have made some comments. Everything except them looks good :+1:

leecalcote:
thank you, @aayushdutt ✔️

ui/components/GrafanaComponent.js
kushthedude: Yes, it was not required, also it was null in the assignment. So I dont think sending empty value is better. Anyways it is assigned its proper value, you may test your local Grafana setup.

ui/components/GrafanaCustomChart.js
kushthedude: It was not getting used, in the `transformDataForChart` I don't think increasing an argument is better.

ui/components/GrafanaCustomChart.js
kushthedude: I think you need to take a better look at code again, giving a unused argument to function is not a good way to write code.

ui/components/GrafanaComponent.js
kushthedude: A single function is better than empty nested arrows.

ui/components/MesheryAdapterPlayComponent.js
kushthedude: It was but that part of code was already commented making this useless.

ui/components/GrafanaComponent.js
aayushdutt: I'm afraid it isn't how it works. I've confirmed it is a higher-order function and needs to be this way. 
See: https://github.com/layer5io/meshery/blob/c28a5355ec372568d72fca3718efd685160191f3/ui/components/GrafanaSelectionComponent.js#L170

ui/components/GrafanaComponent.js
aayushdutt: Maybe you are right but as far as I know, non assigned variables have `undefined` as value. And using `undefined` as a variable in template string causes it to have undefined in string type.

Example:
```js
var aNonAssignedVariable;
var constructedString = `prefix${aNonAssignedVariable}suffix`;
console.log(constructedString) // "prefixundefinedsuffix"
```

aayushdutt: We absolutely don't want this to be sent in the query parameter. Empty string is way better and can be validated in the server side.

ui/components/MesheryAdapterPlayComponent.js
aayushdutt: Can you explain this a bit more, please?
I think the CodeMirror component is imported directly from `node_modules`. How was it's inbuild onChange handler commented?

aayushdutt, line 409: Can you explain this a bit more, please?
I think the CodeMirror component is imported directly from node_modules. How was it's inbuild onChange handler commented?

ui/components/GrafanaComponent.js
kushthedude: https://github.com/layer5io/meshery/blob/c28a5355ec372568d72fca3718efd685160191f3/ui/components/GrafanaSelectionComponent.js#L170
`error` is passed in `msg` arg & it is working as expected. You may even check on your local and confirm there is no need for nested function.
Have a look at ss 👍 
Screenshot 2020-05-01 at 3 49 07 AM

If you think there is some error, please let me know

aayushdutt, line 205: I'm afraid it isn't how it works. I've confirmed it is a higher-order function and needs to be this way. 
See: https://github.com/layer5io/meshery/blob/c28a5355ec372568d72fca3718efd685160191f3/ui/components/GrafanaSelectionComponent.js#L170

ui/components/GrafanaComponent.js
```diff
@@ -74,7 +74,7 @@ class GrafanaComponent extends Component {
 
       grafanaURL,
       grafanaAPIKey,
-      grafanaBoardSearch: '', // we probably dont need this retrieved from store
```
aayushdutt, line 77: Maybe you are right but as far as I know, non assigned variables have `undefined` as value. And using `undefined` as a variable in template string causes it to have undefined in string type.

Example:
```js
var aNonAssignedVariable;
var constructedString = `prefix${aNonAssignedVariable}suffix`;
console.log(constructedString) // "prefixundefinedsuffix"
```

aayushdutt: We absolutely don't want this to be sent in the query parameter. Empty string is way better and can be validated in the server side.

ui/components/GrafanaComponent.js
kushthedude: Then there is no need to fetch BoardSearch prop in constructor from prop if we are ultimately giving it a empty string.

ui/components/GrafanaComponent.js
kushthedude: https://github.com/layer5io/meshery/pull/936#discussion_r418321437

ui/components/GrafanaComponent.js
kushthedude: Done

ui/components/MesheryAdapterPlayComponent.js
kushthedude: An empty onChange handler? Can you let me know more what function does this empty handler is supposed to perform?

aayushdutt:
@kushthedude I think the better way to resolve commits is from the reviewer. This way we will be having a better, unbroken, continuous discussion about the fix/resolution.

ui/components/MesheryAdapterPlayComponent.js
aayushdutt: Not entirely sure, it might be useful for overriding the existing onChange function of the component. However, it is more likely it does nothing, but we need to make sure.

kushthedude:
> @kushthedude I think the better way to resolve change requests is from the reviewer. This way we will be having a better, unbroken, continuous discussion about the fix/resolution since GitHub doesn't allow to re open resolved changes.

You can comment on any change you find absurd or can't understand, I will mark unresolved for you as I have done on the above changes wherever you were not satisfied with the change.

ui/components/MesheryAdapterPlayComponent.js
kushthedude: This function is not called anywhere, whoever wrote this block of code copied the skeleton of the codeMirror component but later forgot to search for empty blocks & remove them. This is not the only place, if you will see #910 you will get to know how un-organised our client-side code is. This PR merely solved 59 errors that too were all because of copying of constructor props, function skeleton, setState function from other components regardless of how all the variables & parameters getting defined are even used or not.

ui/components/GrafanaComponent.js
aayushdutt: You are right, the error will show up. the problem is it will **always** show up. Since you have removed the inner function, it is executed immediately when passed into the callback instead of getting executed in the catch statement.
I'd suggest you revert this change

ui/components/GrafanaComponent.js
aayushdutt: That's exactly what the code comment on that statement says

aayushdutt: that's exactly the comment on the statement says.

ui/components/GrafanaComponent.js
kushthedude: Done

ui/components/MesheryAdapterPlayComponent.js
aayushdutt: Can you make sure by manually testing the CodeMirror component for now? Does it works as expected?

aayushdutt:
@kushthedude That's great. But, I still think it is a better practice to allow the reviewer to decide whether the change is resolved or not. We don't want unresolved changes to remain marked as resolved right?

ui/components/MesheryAdapterPlayComponent.js
kushthedude: Tested both Codemirror & custom config application 
Screenshot 2020-05-01 at 4 47 16 AM

aayushdutt: That's great! Thanks a lot @kushthedude

aayushdutt:
@leecalcote Everything looks nice to me now, we are pretty much good to go. You might want to see the changes at your end.

ui/components/GrafanaComponent.js
```diff
@@ -71,10 +71,10 @@ class GrafanaComponent extends Component {
       urlError: false,
 
       grafanaConfigSuccess,
-
+      grafanaBoardSearch: '',
       grafanaURL,
       grafanaAPIKey,
-      grafanaBoardSearch: '', // we probably dont need this retrieved from store
+      grafanaBoardSearch,
```
aayushdutt, line 77: I think the grafanaBoardSearch variable comes twice in this.state
Please rectify this issue @kushthedude

kushthedude: First one was retrieved from store which is not needed & is unused-var.

aayushdutt: @kushthedude I guess I couldn't explain it better. Actually we don't want `this.state` (or any js object) to have the same key twice.
image
(see line 74 and 77).

This is from the `GrafanaComponent.js` file from this PR: https://github.com/layer5io/meshery/blob/3470113f37609da4b56320cc5cebd59dfe8d1774/ui/components/GrafanaComponent.js

aayushdutt: I'd suggest you to remove the one on line 77

kushthedude: This missed my eye.

aayushdutt: No problem :+1:   That's why we have PR reviewers :smile:

aayushdutt:
@leecalcote  The changes look nice to me! You might want to go through the changes at your end.

leecalcote:
Good to go! Just think of how much fewer check failures there will be after this merge. 😄

PR #949, opened 2020-05-02
feat: UI Server improvisations & commands revamping

### LINKED ISSUE #948: UI: `npm start` doesn't work nor the production build
`npm start` doesn't work nor the production build. Also there is no way to start the production build server and `npm start`

kushthedude:
Fixes #948

### DIFF
--- a/ui/ui_dev_server.js
+++ b/ui/ui_dev_server.js
@@ -26,7 +26,7 @@ app.prepare().then(() => {
     if (pathname.startsWith("/api") || pathname.startsWith("/logout") || pathname.startsWith("/login")){
       proxy.web(req, res, { target: 'http://localhost:9081' });
     } else {
-        handle(req, res)
+      handle(req, res)
     }
 
   }).listen(port, err => {


provider-ui/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "eslint .",
     "next": "next",
     "build": "next build",
-    "start": "next start",
+    "start": "NODE_ENV=production nodemon ui_dev_server.js",
```
aayushdutt, line 12: This might not be the best way to do it.

1. NODE_ENV is the place for Dockerfile/Docker-compose, it shouldn't really be here.
2. We shouldn't be using nodemon in production, it is a development only tool
    * We don't want the app to keep running after an exception occurs, nodemon will keep the server running even if it is intended to crash.
    * There are better ways to do it (something like pm2)
    * nodemon is a dev dependency and won't be installed on the server
3. Having said that, the `ui_dev_server.js` file is specifically written for use during development. Are you sure that the correct host and port for communicating with API in production server is `http://localhost:9081`? ([line28](https://github.com/layer5io/meshery/blob/94a363e1cbb736cb3c4411566f6335c0729e26a6/provider-ui/ui_dev_server.js#L28))

4. Just to make sure, am I right when I say we are using the next.js server, and NOT the static build of the UI? @leecalcote

ui/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "eslint .",
     "next": "next",
     "build": "next build",
-    "start": "next start",
+    "start": "NODE_ENV=production nodemon ui_dev_server.js",
```
aayushdutt, line 12: Same concern as the comment made in `provider-ui/package.json`

provider-ui/package.json
kushthedude: I think you really need to get acquainted with next.js and ssr. Hope this will help you https://nextjs.org/docs/advanced-features/custom-server

aayushdutt: Yes I looked through that page @kushthedude 
I was just making sure we are actually making use of SSR

kushthedude:
Had it been any static template renderer like gatsby, I would have followed
all of your suggestions.

provider-ui/package.json
aayushdutt: Now that I know we are using SSR, maybe the best way to do it will be `"start": "node ui_dev_server.js" `

kushthedude:
No nodemon should be there to monitor for any changes which are done while
server is running. If we use node instead of nodemon then we will have to
restart the server everytime we make some changes.

provider-ui/package.json
nveenjain: We are using next export (https://github.com/layer5io/meshery/blob/master/Dockerfile#L10) . We are using npm start only for development purposes. Then again NODE_ENV should not be production in such case.

aayushdutt: Oh, thanks @nveenjain. 
Since we are not using SSR (yet), and we are using a separate dev script for development, there is no need for a start script for now.
So probably we should best have it as `"start": "node ui_dev_server.js"`

aayushdutt:
> No nodemon should be there to monitor for any changes which are done while
server is running. If we use node instead of nodemon then we will have to
restart the server everytime we make some changes.

@kushthedude You maybe mistook it as the development script? We have the `dev` script for development and we don't change server files in production so we might not need nodemon.

aayushdutt:
Also, since we are not using SSR in production yet, it might be a good idea to list that enhancement in the todo list.

kushthedude:
> @kushthedude You maybe mistook it as the development script? We have the dev script for development and we don't change server files in production so we might not need nodemon.

Agreed!

provider-ui/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "eslint .",
     "next": "next",
     "build": "next build",
-    "start": "NODE_ENV=production nodemon ui_dev_server.js",
+    "start": "NODE_ENV=production node ui_dev_server.js",
```
aayushdutt, line 12: Great to see nodemon was removed.
I think `NODE_ENV=production` shouldn't be there either.
Let's go with: `"start": "node ui_dev_server.js"` @kushthedude

ui/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "eslint .",
     "next": "next",
     "build": "next build",
-    "start": "NODE_ENV=production nodemon ui_dev_server.js",
+    "start": "NODE_ENV=production node ui_dev_server.js",
```
aayushdutt, line 12: Same as above

kushthedude:
Why?

aayushdutt:
> Why?

@kushthedude  Actually, environment variables are a place for the environment, not the script itself. It's better to have either a `.env` file or just add them in `dockerfile` or `docker-compose.yml`.
Also, this variable doesn't serve any purpose at all here as we will be adding that variable in our [docker-compose.yaml](https://github.com/layer5io/meshery/blob/master/docker-compose.yaml)

kushthedude:
What if I want to replicate on my local without contenarisation ? I don't
believe in specifying multiple env variables also if you take a look at
dev_server.js file, it takes input of node_env from run script to identify
the environment.

aayushdutt:
> What if I want to replicate on my local without contenarisation ? I don't
believe in specifying multiple env variables also if you take a look at
dev_server.js file, it takes input of node_env from run script to identify
the environment.

For testing on local without containerisation, you can just do `export NODE_ENV=production` to set the environment variable as production locally.

kushthedude:
Specifying the env variable is lot better than I presume.

aayushdutt:
Hmmm... I agree it is easier to access production variant that way. But I still think the environment variables should not belong to the script. Also, we don't need to access the production server that often really. 
I mean we can have it there in the script, it doesn't break anything or change anything as we use a different script for development. 
What are your thoughts on this @nveenjain @leecalcote?

provider-ui/package.json
nveenjain: In production we use next export... I'm little curious, why NODE_ENV=production is specified though...

provider-ui/ui_dev_server.js
```diff
@@ -23,7 +20,10 @@ proxy.on('error', (err, req, res) => {
 
 app.prepare().then(() => {
   createServer((req, res) => {
-    const { pathname } = parse(req.url, true);
+    // Be sure to pass `true` as the second argument to `url.parse`.
+    // This tells it to parse the query portion of the URL.
+    const parsedUrl = parse(req.url, true)
```
nveenjain, line 25: I guess we can revert these changes (maybe keep comment)? Since we are only using pathName that was already desctructured from parse function...

ui/ui_dev_server.js
```diff
@@ -23,22 +20,15 @@ proxy.on('error', function (err, req, res) {
 
 app.prepare().then(() => {
   createServer((req, res) => {
-    const { pathname } = parse(req.url, true)
+    // Be sure to pass `true` as the second argument to `url.parse`.
+    // This tells it to parse the query portion of the URL.
+    const parsedUrl = parse(req.url, true)
+    const { pathname, query } = parsedUrl
```
nveenjain, line 26: Same as in provider-ui/ui-dev_server.js

provider-ui/ui_dev_server.js
kushthedude: 👍

kushthedude:
@leecalcote It is good to go now.

leecalcote:
@aayushdutt and @nveenjain I think your comments have been addressed. Will you confirm?

leecalcote:
Was awaiting @nveenjain's feedback. It looks like its been addressed above, though.

PR #957, opened 2020-05-04
fix: discrepancy in go.sum after updating the modules

kushthedude:


kushthedude:
> @leecalcote so after updating this module, I still got some unstaged changes on my local. I really think dependabot is misbehaving with go.sum file. See here #957

@leecalcote Ok so I got the issue, whenever dependabot updated a module rather than updating the existing module, it adds another module over its top. And when we set-up on our local, the old module gets removed which shows unstaged changes.
I will see with dependabot what we can do here, meanwhile I think it is good to go.

leecalcote:
Thank you, @kushthedude. It would be nice to avoid the go.sum update with each PR.

@nveenjain, @kanishkarj and I were openly wondering whether running dependencybot on UI packages is more helpful, than hurtful. This would be a good point of discussion in an upcoming call.

PR #974, opened 2020-05-07
feat: Addition of performance defaults in settings

### LINKED ISSUE #213: Performance: M4: Expose Load Generator settings in UI
**Description**
See #126 for full description of Load Generator Integration Epic. Selection of load generator should be made available in two locations:

1. as a general setting under the /settings screen. This setting will be the default for all performance tests.
1. as a per test setting under the /performance screen. This setting will be one-time for _that_ specific test.

**M4: Expose Load Generator settings in UI**
- [x] Per test selection under /performance: Update UI to select from the choice of load generators (fortio and wrk2) and include it as part of the payload to submit a job for load test, fetch response and process it the usual way. See mockup:

Screen Shot 2019-09-24 at 4 23 05 PM

- [ ] Default selection under /settings: Add load generators as a new section in the Settings screen. Fortio will be the default setting for new Meshery installations. Use the performance.svg icon provided by https://github.com/layer5io/meshery/pull/292.

performance

Meshery UI uses Material UI. Radio buttons are here - https://material-components.github.io/material-components-web-catalog/#/component/radio

**Additional context**
See [Load Generator Integration design doc](https://docs.google.com/document/d/1jZMOih3Qn1ixrq1Ve6fkdZZ4TuKNoF_qfClDZTnLNcw/edit#).

kushthedude:
fixes #213

### DIFF
--- a/ui/components/MesherySettings.js
+++ b/ui/components/MesherySettings.js
@@ -77,6 +77,9 @@ class MesherySettings extends React.Component {
         case 'metrics': 
           tabVal = 2;
           break;
+        case 3:
+          newRoute+='#performance'
+          break;
       }
     }
     this.state = {
--- a/ui/components/MesherySettingsPerformanceComponent.js
+++ b/ui/components/MesherySettingsPerformanceComponent.js
@@ -140,6 +140,7 @@ class MesherySettingsPerformanceComponent extends React.Component {
         });
       }
     }, this.handleError('There was an error sending your preference'));
+        window.location.reload(true);
   }
     componentDidMount() {
     this.getLoadTestPrefs();
@@ -172,6 +173,7 @@ class MesherySettingsPerformanceComponent extends React.Component {
           });
         }
       }, self.handleError('There was an error sending your preference'));
+
     }
 
   handleError(msg) {


leecalcote:
💥

ui/components/MesherySettings.js
```diff
@@ -121,6 +118,9 @@ class MesherySettings extends React.Component {
           case 2:
             newRoute+='#metrics'
             break;
+          case 3:
+            newRoute+='#performance'
```
leecalcote, line 122: Nice.

ui/components/MesherySettingsPerformanceComponent.js
```diff
@@ -0,0 +1,301 @@
...
+import { withSnackbar } from 'notistack';
+import { connect } from 'react-redux';
+import CloseIcon from '@material-ui/icons/Close';
+import LoadTestTimerDialog from './load-test-timer-dialog';
+
+
+const loadGenerators = [
+  'fortio',
+  'wrk2',
+];
```
leecalcote, line 19: We're missing nighthawk here. ;)

kushthedude: That won't be for so long 🌟

kushthedude:
Current Flow -:
- This PR only introduces UI Elements & Input Fields related function.
- A subsequent PR will be made which will store this preferences in-store & user will be shown an checkbox 
` Import Defaults from Settings` on the `/performance` page or if the user doesn't choose it, the values will be overridden only for a single test.

leecalcote:
@kushthedude thanks for moving this area forward. Give thought to whether a checkbox on the /performance page is necessary, if you would. Consider that the /performance page will load with the user's defaults (whether those have been personalized or remain the project defaults) and that as the user goes to run a given test that they have the option of overriding those defaults for that specific test. 

Later, we go to provide the ability for users to create performance test profiles, users will then be enabled with a greater set of controls on what settings are used for specific test scenarios they might be running.

With respect to the /settings#performance page, given that each of the fields are marked as required, consider populating the default setting values (or the user's saved, personalized settings) in each of the three text fields on page load.

kushthedude:
> Later, we go to provide the ability for users to create performance test profiles, users will then be enabled with a greater set of controls on what settings are used for specific test scenarios they might be running.

So what I got from your point is, when the user navigates to `/performance`, the settings defaults should populate the current values in the corresponding fields rather than a checkbox which can be overridden if the user wants to.
Seems fine to me!

kushthedude:
@leecalcote Can we move ahead with this ?

leecalcote:
> @leecalcote Can we move ahead with this ?

Good question. If we had multiple release channels, yes. Without additional version controls here, I'm disinclined to "release" a half-functioning feature, so to speak.

kushthedude:
> > @leecalcote Can we move ahead with this ?
> 
> Good question. If we had multiple release channels, yes. Without additional version controls here, I'm disinclined to "release" a half-functioning feature, so to speak.

Ohk, I will push the subsequent commits here rather than another PR :+1:

leecalcote:
@kushthedude still in draft mode here, right?

kushthedude:
> @kushthedude still in draft mode here, right?

This still needs some work, I was testing the `saveLoadPref` endpoint which is already implemented to make sure that the functionality is not broken since Meshery is getting some new powers now-a-days.

kushthedude:
@leecalcote The response of the GET call

```json
{c: 10, qps: 5, t: "12s", loadGenerator: "fortio"}
```

kushthedude:
@leecalcote REady for review, please test on local!

PR #978, opened 2020-05-08
fix: Deprecate the billboard due to invalid function args

### LINKED ISSUE #977: Results: Comparision of two results is not working
**Description**

https://github.com/layer5io/meshery/issues/958#issuecomment-623664832

**Expected Behavior**

Comparision should work fine.

---
[Optional] **To Reproduce**
- Still ambiguous

kushthedude:
Fixes #977

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -37,7 +37,7 @@
     "jsonlint-mod": "^1.7.4",
     "jss": "latest",
     "minimist": ">=1.2.2",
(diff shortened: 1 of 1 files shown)


kushthedude:
**Screenshot**
Screenshot 2020-05-08 at 11 30 41 AM

kushthedude:
@leecalcote With this, I think it is high time that we update dependabot to check for our UI deps too.

PR #983, opened 2020-05-08
Addition of extra space to avoid weird size of description

kushthedude:
@leecalcote A small neat to avoid the weird size of issue description.
As can be seen here 
https://github.com/layer5io/meshery/issues/981

Screenshot 2020-05-08 at 6 41 00 PM


kushthedude:
@leecalcote Please check!

PR #1020, opened 2020-05-18
meshery: Setup EditorConfig

tsingh2k15:
#### Description

- This PR fixes #    (https://github.com/layer5io/meshery/issues/982) 
- Add an EditorConfig file under project root. 
- When opening a file, EditorConfig plugins look for a file named .editorconfig in the directory of the opened file and in every parent directory. A search for .editorconfig files will stop if the root filepath is reached or an EditorConfig file with root=true is found.

#### Checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,17 +1,20 @@
-# EditorConfig helps developers define and maintain consistent
-# coding styles between different editors and IDEs
-# editorconfig.org
-
 root = true
 
-
 [*]
-end_of_line = lf
 charset = utf-8
+end_of_line = crlf
+insert_final_newline = false
+max_line_length = 120
 trim_trailing_whitespace = true
-insert_final_newline = true
-indent_style = space
+
+[*.go]
+indent_size = 4
+indent_style = tab
+
+[*.js]
 indent_size = 2
+indent_style = space
 
 [*.{diff,md}]
 trim_trailing_whitespace = false
+


.editorconfig
```diff
@@ -0,0 +1,17 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+
+[*]
```
kushthedude, line 8: Hi @tsingh2k15 , great work here! I suggest to add the this file only for UI code as the conventions for our js code and golang code are different in terms of spaces and indenting.

tsingh2k15: > Hi @tsingh2k15 , great work here! I suggest to add the this file only for UI code as the conventions for our js code and golang code are different in terms of spaces and indenting.

@kushthedude Thanks for the feedback! I will work on the review comment, can you clarify my below questions?

- Initially I added this file under two UI sub repos under the main repo, but later I thought moving it to root level folder can be better. I can move it back to individual UI folders, but that would require it to be present under two sub repos right? There are two of them - "ui" and "provider-ui". Please share your thoughts if that is okay?
- Another reason I moved it to root directory was my thoughts on having just one file where we can include rules for *.go files and *.js files. Is this feasible to club the rules in one file, that may be easy to maintain if most of the rules are common across?

kushthedude: I agree that we should have single EditorConfig file for ease of workflow & presentation of repository.

> * . Is this feasible to club the rules in one file, that may be easy to maintain if most of the rules are common across?

Even I think we should have the same indents & spaces over the repository. I will open the issue for it, I think only place to make change will be the ESLint rules file.

leecalcote:
Gents, are we settled on an indent of 2 here and the same in the UI (https://github.com/layer5io/meshery/pull/1031)?

// @nveenjain

leecalcote:
@kushthedude any other feedback on this PR?

tsingh2k15:
@kushthedude I think having same rules for Golang and JS files will create problems for developers where rules might be different for Golang and JS files. JS developers may have different preference and Golang developers may have different preference. 

 I would suggest to have just one editorconfig on root project and include different desired rules for *.go file extensions and different rules can be included for *.css, *.html anf *.js files as desired. Any override in the ui and provider-ui can be deleted. This way we can have one editorconfig file for entire monorepo but with in that file we can have rules tailored to different file extensions.

Below is a sample of what I am thinking. I am not sure if consistent styles are currently being used in this project, but below example can be tweaked for this monorepo.

```
root = true

[*]
charset = utf-8
end_of_line = crlf
insert_final_newline = false
max_line_length = 120
indent_style = space
tab_width = 4

[*.go]
indent_size = 4

[*.js]
indent_size = 2
```
Let me know your thoughts. Once we have rules finalized, a PR can be made.

kushthedude:
@tsingh2k15 Looks good to me, just a suggestion can we use this as the config 
```
root = true

[*]
charset = utf-8
end_of_line = crlf
insert_final_newline = false
max_line_length = 120
trim_trailing_whitespace = true

[*.go]
indent_size = 4
indent_style = tab

[*.js]
indent_size = 2
indent_style = space

[*.{diff,md}]
trim_trailing_whitespace = false

```

Since in golang we use tab spacing with tab_width 4 and in js we use key spacing with width 2. We would want to trim the trailing spaces in our code files but we don't want them to mess up the indents on our markdown or documentation files.
Let me know if it seems fine to you.

tsingh2k15:
@kushthedude Yes, this looks fine. I will update the diff with the suggested config.

tsingh2k15:
> @kushthedude any other feedback on this PR?

@leecalcote This is peer reviewed complete. Waiting for your review.

leecalcote:
Excellent.

leecalcote:
Yay! 🎈@tsingh2k15 🥳

PR #1040, opened 2020-05-21
meshery: Support sending additional data during load tests

kanishkarj:
**Description**

This PR allows users to send additional data in the requests while performing load tests. We will need to freshen the UI and add other functionalities.

**Notes to reviewers**
Please rebuild meshery UI to test this PR.
In the meshery UI there will be a small textbox below the URL text field. Enter the headers in JSON format for eg : 

```json
{ 
  "token":"bearer abcdef...", 
  "sampleHeader":"sampleValue"
}
``` 

Point the meshery load test to a local server that prints the headers of the request it receives so that one can verify the headers being sent. The sample Node.js script I used to test it is : 

```javascript
var http = require('http');

http.createServer(function (req, res) {
    console.log(req.headers)
// randomly fails half of the requests
    if (Math.random() >= 0.5) {
        res.writeHead(429, {'Content-Type': 'text/html'});
    }    
res.end();
}).listen(8080); //the server object listens on port 8080 
```

#### Checklist

- [x] Send headers via Fortio
- [ ] Send headers via Wrk2
- [ ] Send body via Fortio
- [ ] Send body via Wrk2
- [ ] Should we also support cookies?

### DIFF
--- a/helpers/fortio.go
+++ b/helpers/fortio.go
@@ -35,7 +35,6 @@ func sharedHTTPOptions(opts *models.LoadTestOptions) *fhttp.HTTPOptions {
 	// }
 	httpOpts.FollowRedirects = true
 	httpOpts.DisableFastClient = true
-
 	for key, val := range *opts.Headers {
 		httpOpts.AddAndValidateExtraHeader(key + ":" + val)
 	}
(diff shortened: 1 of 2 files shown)


kanishkarj:
@leecalcote Please try this out and let me know. Also, it would be great if someone could tidy up the UI part.

leecalcote:
@kanishkarj so far, so good. This is functional and the request headers is as in good a place as any at the moment. 

@shivaylamba some client-side json validation and formatting in this new text field would be nice.

kanishkarj:
I have added support to send cookies and request body with the requst.

PR #1053, opened 2020-05-25
Revert "Add properties file for meshery server"

kushthedude:
Reverts layer5io/meshery#1043

https://github.com/layer5io/meshery/pull/1043#issuecomment-633681566
https://github.com/layer5io/meshery/runs/701315774

The PR broke docker action because the `get previous tag` action associated with the PR only works on releases which have only release version as their tag name. But for meshery release, our tag name is `Meshery v0.x.x` hence the action is failing & docker push is not happening

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -271,22 +271,6 @@ jobs:
       uses: actions/checkout@master
       with:
         fetch-depth: 1
-    - name: Set properties file with release tag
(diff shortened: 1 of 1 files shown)


naturalett:
@kushthedude 
I removed the external action which retrieved me back the previous tag release and I'm using:
git describe --abbrev=0 --tags

**Test:**
https://github.com/layer5io/meshery-tmp/runs/714426397?check_suite_focus=true

PR #1107, opened 2020-06-05
feat: Add pre-commit hook for lint

kushthedude:

PR #1108, opened 2020-06-05
docs: Addition of ESLint in contributing.md

kushthedude:


kushthedude:
@leecalcote Please take a look here!

CONTRIBUTING.md
```diff
@@ -160,6 +160,15 @@ update /etc/hosts
 127.0.0.1 mesherylocal.layer5.io
 ```
 
+### UI Lint Rules 
+
+We are using ES-Lint to maintain code quality & consistency in our UI Code. To make sure your PR passes all the UI & ES-Lint Tests, please see below :
+
+- Remember to run `make run-ui-lint` & `make run-provider-ui-lint` if you are making changes in Meshery-UI & Provider-UI respectively.
```
leecalcote, line 167: Nice. Should we take the next step and include these as subcommands called from the main `make run-ui-dev`, too?

CONTRIBUTING.md
```diff
@@ -160,6 +160,15 @@ update /etc/hosts
...
+- The above commands will only fix some basic indenting rules for you, you will have to manually check your code to ensure there is no duplications, un-used variables or un-declared constants.
```
leecalcote, line 168: Will `eslint` flag and error out on these, too? Do this leave a gap where contributors may be frustrated by not knowing that their changes will fail a build check?

CONTRIBUTING.md
kushthedude: > Nice. Should we take the next step and include these as subcommands called from the main `make run-ui-dev`, too?

That would be great, even I was thinking the same as using a lot of make commands will be a hectic for contributor 😄

CONTRIBUTING.md
```diff
@@ -160,6 +160,15 @@ update /etc/hosts
...
+- We will soon be adding Pre-Commit Hooks to make sure you get to know your errors before you commit the code.
```
leecalcote, line 169: Given this PR is timed with PR #1107, perhaps, we should update this phrasing to say that we _do_ include a pre-commit hook. What do you think?

kushthedude: That PR is incomplete, I am thinking to complete it once we have `golang-ci-lint` as golang-ci provides an option to run on local too.

CONTRIBUTING.md
kushthedude: Yes the flag will point all the errors, its just that some error will be fixed automatically but some won't be.

leecalcote: Good deal.

CONTRIBUTING.md
leecalcote: Ah, got it. 

Sounds like a good plan (`golang-ci-lint`)

CONTRIBUTING.md
leecalcote: @kushthedude are you thinking of adding that Makefile update (of command grouping) in another PR?

kushthedude: Done here https://github.com/layer5io/meshery/pull/1109#pullrequestreview-424965774

leecalcote: Oh, I see. I don't think I understand the connection between and the `package.json` scripts and the makefile commands. Which `package.json` scripts are executed as part of a given makefile command?

I was thinking that `make run-ui-lint` would be available as it's own command, but also implicitly called as part of an overarching command like `run-ui-dev`.

https://github.com/layer5io/meshery/blob/master/Makefile#L87

kushthedude: Actually, `make run-ui-dev` runs `npm run dev` from package.json. What I have done is I have modified `npm run dev` to include a prior run to check for lint errors, as if someone who do not use `make run-ui-dev` and use `npm run dev` will also be able to see the eslint errors if there are any.

leecalcote: Oh, I see. So, whether through `make` or through `npm` directly, linting will be performed. This was sitting so obviously in my face with `npm run dev;` inside of the make commands.

PR #1126, opened 2020-06-10
Fix for 1122

kushthedude:


provider-ui/package.json
```diff
@@ -36,7 +36,7 @@
     "react-jss": "^8.1.0",
     "react-select": "^2.4.3",
     "styled-jsx": "latest",
-    "uuid": "^3.3.2"
+    "uuid": "^8.1.0"
```
leecalcote, line 39: I wonder if uuid is used in provider-ui at all; if it is a candidate for removal.

kushthedude: Yup, it is a candidate for removal

kushthedude:
Good for merge! @leecalcote

PR #1151, opened 2020-06-18
ui: Added toggle Switch to hide Headers

### LINKED ISSUE #1099: [UI] Implement toggle to show/hide header fields
**Current Behavior**

Right now, all the header fields are visible on the performance page which is not looking good on UI.

**Desired Behavior**

Implement a toggle to show/hide the header fields.

**Screenshot**
Screenshot 2020-06-03 at 4 56 56 AM

---

dhruv0000:
Signed-off-by: Dhruv Patel <patel.4@iitj.ac.in>

**Description**
Added the switch button to toggle between labels and placeholders

This PR fixes #1099 

**Notes to reviewers**

#### Checklist

### DIFF
--- a/ui/components/MesheryPerformanceComponent.js
+++ b/ui/components/MesheryPerformanceComponent.js
@@ -1,3 +1,4 @@
+/* eslint-disable */ 
 import React from 'react';
 import PropTypes from 'prop-types';
 import Button from '@material-ui/core/Button';
@@ -14,7 +15,7 @@ import { bindActionCreators } from 'redux';
 import CloseIcon from '@material-ui/icons/Close';
 import GetAppIcon from '@material-ui/icons/GetApp';
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore'
-import { updateLoadTestData, updateStaticPrometheusBoardConfig } from '../lib/store';
+import { updateLoadTestData, updateStaticPrometheusBoardConfig, updateLoadTestPref, updateProgress } from '../lib/store';
 import dataFetch from '../lib/data-fetch';
 import MesheryChart from './MesheryChart';
 import LoadTestTimerDialog from './load-test-timer-dialog';
@@ -81,7 +82,7 @@ class MesheryPerformanceComponent extends React.Component {
   constructor(props) {
     super(props);
     const {
-      testName, meshName, url, qps, c, t, result, staticPrometheusBoardConfig
+      testName, meshName, url, qps, c, t, result, staticPrometheusBoardConfig, k8sConfig, loadTestPrefs,
     } = props;
 
     this.state = {
@@ -121,6 +122,8 @@ class MesheryPerformanceComponent extends React.Component {
 
     this.setState({ [name]: event.target.value });
   };
+
+
 
   handleSubmit = () => {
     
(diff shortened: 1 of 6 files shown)


kushthedude:
Congrats @dhruv0000 on your first PR 🎉

ui/components/MesheryPerformanceComponent.js
```diff
@@ -54,6 +54,10 @@ const styles = (theme) => ({
     marginTop: theme.spacing(3),
     marginLeft: theme.spacing(1),
   },
+  expansionPanel: {
+    boxShadow:'none',
+    border: '1px solid rgb(196,196,196)',
```
kushthedude, line 59: Can we remove borders from the expansion panel ?

ui/components/MesheryPerformanceComponent.js
```diff
@@ -613,6 +558,77 @@ class MesheryPerformanceComponent extends React.Component {
                   </RadioGroup>
                 </FormControl>
               </Grid>
+              <Grid item xs={12} sm={12} gutterBottom>
+                <ExpansionPanel className={classes.expansionPanel}>
+                  <ExpansionPanelSummary expanded={true}>
+                    <Typography align="center" color="textSecondary" varient="h6">More Options</Typography>
```
kushthedude, line 564: Also, can we remove the border structure from the More Options & add an arrow-down/up icon on the basis of the state of the expansion panel?
Screenshot 2020-06-19 at 7 59 17 PM

Somewhat like this with an arrow icon adjacent to it.
Screenshot 2020-06-19 at 8 02 40 PM

kushthedude, line 564: Can we rename `More Options` to `Add Request Data` ?

dhruv0000:
image
How about now?

kushthedude:
Looks great @dhruv0000, but I think the previous bordered design over the options was better. Can you please make it bordered again rather than the shadowed box. It will be good to merge after that.

dhruv0000:
image
@kushthedude

Nikhil-Ladha:
Just a question here @kushthedude , @dhruv0000 and @leecalcote , instead of creating a button of the same width as other fields won't it be better to make a smaller one with a color to do the same, as it looks more like a drop-down to me rather than a toggler. What say?

kushthedude:
> Just a question here @kushthedude , @dhruv0000 and @leecalcote , instead of creating a button of the same width as other fields won't it be better to make a smaller one with a color to do the same, as it looks more like a drop-down to me rather than a toggler. What say?

It is not a dropdown but an expansionPanel which should be of the same width for even UI if we make a dropdown of smaller width which then will reveal the fields of more width than the controller dropdown it will look ugly I think, a dropdown should only the fields with the width same or less then the parent IMO.
What you say @Nikhil-Ladha ?

Nikhil-Ladha:
I just checked out the PR in my local, and it is looking pretty good. Much different than what I was thinking about. 
Another thing, the advanced option isn't appealing or doesn't looks like an expansion panel and more like any other field, just a 'font-weight:bold' may help, like the image I have attached. Otherwise, it is all good.
Screenshot from 2020-06-20 18-28-44

dhruv0000:
> I just checked out the PR in my local, and it is looking pretty good. Much different than what I was thinking about.
> Another thing, the advanced option isn't appealing or doesn't looks like an expansion panel and more like any other field, just a 'font-weight:bold' may help, like the image I have attached. Otherwise, it is all good.
> Screenshot from 2020-06-20 18-28-44

Hey @Nikhil-Ladha, won't making it bold emphasizes more on advanced option than other fields, and as for giving a subtle hint about the expansion, I tried using [a shadow effect](https://github.com/layer5io/meshery/pull/1151#issuecomment-646845911), but as @kushthedude later correctly advised that it was looking a bit off theme. 
Mabey something could be done in that direction?

Nikhil-Ladha:
Well, It didn't affected much and seemed to be just as I hoped. But, if that may happen then we not consider that option.
Not sure about the shadowed one though. Current one looks good.

kushthedude:
I tried your suggestion @Nikhil-Ladha, it was highlighting the expansion-panel heading from the other fields which should not be the case. Coming to the point of letting the user know of the drop-down the arrow icon added in the field is sufficient for now.
If needed we can iterate more into this, as we also need to make changes in our performance UI.

kushthedude:
@leecalcote I think we are good to go with this, please have a look!

leecalcote:
@dhruv0000 ooooh, niiice!

leecalcote:
@dhruv0000 I can see perhaps adding a light shade of gray or some other color to signify that the Advanced Options dropdown is a group of fields. 

Untitled

Other than this, it looks fantastic! You're hired. 😉

PR #1156, opened 2020-06-20
Mesheryctl perf test using SMPS test config

### LINKED ISSUE #596: Performance: mesheryctl-based SMPS Import and Test Execution
**Current Behavior**
Users are able to invoke performance tests via UI.

**Desired Behavior**
Capability: Provide -f and --file import of an SMPS document.
Behavior: Support local filesystem references and HTTP/S references. Accepting a single file reference only, populate SMPS test configuration values into Performance page fields.

See `test` command in [Meshery CLI Commands & Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#).

---
**Resources**
[Meshery Performance Management](https://docs.google.com/document/d/1VEvh4qprHbbI3CjeOf9zU9bRhpmhBAVEaPhIO2XDDlc/edit#heading=h.1tnd1q3bwl7c)

kanishkarj:
Signed-off-by: kanishkarj <kanishkarj@hotmail.com>

**Description**

This PR imports the SMPS PerformanceTestConfig yaml file and performs a load test based on it.

This PR fixes #596  

**Notes to reviewers**

```bash
./mesheryctl perf --file ./smps-test.yaml --token ~/Downloads/auth.json
```
```yaml
# smps-test.yaml 
---
smpsVersion: 'v0.1'
id: 7df8a0ac-1b90-4852-857f-719f75c5cf0e
name: quick-test
labels:
  h1: v1
clients:
- internal: false
  loadGenerator: fortio
  protocol: http
  connections: 10
  rps: 10
  # headers:
  #   header1: val1
  # cookies:
  #   cookie1: val1
  # body: ReqBody
  # contentType: application/txt
  endpointUrl: 
  - https://github.com
duration: 1s
```

#### Checklist

- [] [Signed commits](https://github.com/layer5io/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin).
- [] PR title includes `[<name of area>]:<description>`


handlers/load_test_handler.go
```diff
@@ -39,42 +41,58 @@ func (h *Handler) LoadTestUsingSMPSHandler(w http.ResponseWriter, req *http.Requ
 		http.Error(w, msg, http.StatusInternalServerError)
 		return
 	}
-	benchMark := &models.PerformanceSpec{}
```
leecalcote, line 42: Good. No more “benchmark” references.

handlers/load_test_handler.go
```diff
@@ -39,42 +41,58 @@ func (h *Handler) LoadTestUsingSMPSHandler(w http.ResponseWriter, req *http.Requ
...
-	if err := yaml.Unmarshal(body, benchMark); err != nil {
+	jsonBody, err := yamlj.YAMLToJSON(body)
+	if err != nil {
+		msg := "unable to convert YAML to JSON"
+		err = errors.Wrapf(err, msg)
+		logrus.Error(err)
+		http.Error(w, msg, http.StatusInternalServerError)
+		return
+	}
+	benchMark := &SMPS.PerformanceTestConfig{}
```
leecalcote, line 52: Perhaps, “perfTest” or something similar, so that we might be rid of “benchmark” references?

The logic behind being rid of benchmark references is that Meshery’s performance management capabilities are broader in nature than the specific use case of a benchmark.

handlers/load_test_handler.go
```diff
@@ -92,17 +110,17 @@ func (h *Handler) LoadTestUsingSMPSHandler(w http.ResponseWriter, req *http.Requ
...
+	loadGenerator := testClient.LoadGenerator
 
-	// switch loadGenerator {
-	// case "wrk2":
-	// 	loadTestOptions.LoadGenerator = models.Wrk2LG
-	// default:
-	loadTestOptions.LoadGenerator = models.FortioLG
+	switch loadGenerator {
+	case SMPS.PerformanceTestConfig_Client_wrk2:
+		loadTestOptions.LoadGenerator = models.Wrk2LG
```
leecalcote, line 117: @kushthedude will this change once we have a Load Generator Interface?

mesheryctl/internal/cli/root/perf/perf.go
```diff
@@ -129,79 +127,75 @@ var PerfCmd = &cobra.Command{
...
 			}
-			if testURL == "" {
-				testURL = t.EndpointURL
-			}
-			if concurrentRequests == "" {
-				concurrentRequests = fmt.Sprintf("%d", t.Client.Connections)
+
+			req, err = http.NewRequest("POST", mctlCfg.GetPerf().LoadTestSmpsURI, bytes.NewBuffer(smpsConfig))
+			if err != nil {
+				return errors.Wrapf(err, utils.PerfError(fmt.Sprintf("failed to create new request to %s", mctlCfg.GetBaseMesheryURL())))
```
leecalcote, line 140: “Failed to invoke performance test”?

mesheryctl/internal/cli/root/perf/perf.go
```diff
@@ -129,79 +127,75 @@ var PerfCmd = &cobra.Command{
...
-			testName = StringWithCharset(8)
-			log.Debug("Using random test name: ", testName)
-		}
+			postData := ""
 
-		postData := ""
+			startTime := time.Now()
+			duration, err := time.ParseDuration(testDuration)
+			if err != nil {
+				return errors.Wrapf(err, utils.PerfError(fmt.Sprintf("failed to parse test duration %s", testDuration)))
```
leecalcote, line 155: Can this be “Invalid test duration. Please use the format xs, xm, xh to specify seconds, minutes, hours”?

mesheryctl/internal/cli/root/perf/perf.go
```diff
@@ -129,79 +127,75 @@ var PerfCmd = &cobra.Command{
...
-		}
+			endTime := startTime.Add(duration)
 
-		endTime := startTime.Add(duration)
+			postData = postData + "start_time: " + startTime.Format(time.RFC3339)
+			postData = postData + "\nend_time: " + endTime.Format(time.RFC3339)
 
-		postData = postData + "start_time: " + startTime.Format(time.RFC3339)
-		postData = postData + "\nend_time: " + endTime.Format(time.RFC3339)
+			if testURL != "" {
```
leecalcote, line 163: When supporting gRPC, will the testURL be blank? Or be a function name? I think blank because it will just be a ping test.

kushthedude:
@kanishkarj This gives the power to use SMPS using CLI right, @leecalcote Do we need a similar a sort of feature with UI where I can get option to upload my yaml file & everything will be done on own ?

go.mod
```diff
@@ -39,3 +41,5 @@ require (
 	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
 	k8s.io/utils v0.0.0-20191010214722-8d271d903fe4 // indirect
 )
+
+replace github.com/layer5io/service-mesh-performance-specification v0.0.0-20200617124327-27b13cd1be5e => github.com/kanishkarj/service-mesh-performance-specification v0.0.0-20200620172814-ca391dc349b7
```
kushthedude, line 45: @kanishkarj Is this just for test purpose ?

mesheryctl/internal/cli/root/perf/perf.go
kushthedude: > Can this be “Invalid test duration. Please use the format xs, xm, xh to specify seconds, minutes, hours”?

Or better can we point to the URL, as we have done in perf -help argument ?

mesheryctl/internal/cli/root/perf/perf.go
```diff
@@ -129,79 +127,75 @@ var PerfCmd = &cobra.Command{
...
-		} else {
-			return errors.New(utils.PerfError("please enter a test URL"))
-		}
+			// Method to check if the entered Test URL is valid or not
+			var validURL = govalidator.IsURL(testURL)
 
-		// Method to check if the entered Test URL is valid or not
-		var validURL = govalidator.IsURL(testURL)
+			if !validURL {
+				return errors.New(utils.PerfError("please enter a valid test URL"))
```
kushthedude, line 173: @kanishkarj Can we also log the errors using logrus ?

mesheryctl/internal/cli/root/perf/perf.go
```diff
@@ -129,79 +127,75 @@ var PerfCmd = &cobra.Command{
...
-				testDuration = fmt.Sprintf("%fs", t.EndTime.Sub(t.StartTime).Seconds())
+				return err
 			}
-			if testURL == "" {
-				testURL = t.EndpointURL
-			}
-			if concurrentRequests == "" {
-				concurrentRequests = fmt.Sprintf("%d", t.Client.Connections)
+
+			req, err = http.NewRequest("POST", mctlCfg.GetPerf().LoadTestSmpsURI, bytes.NewBuffer(smpsConfig))
```
kushthedude, line 138: Just a doubt, what is the behaviour when I have given an SMPS file as well as I have entered some arguments while executing the command?

leecalcote:
Yes, sir, we do. We’ll call it “performance test profiles”.

leecalcote:
That’s exactly what I was thinking when typing out this comment on my mobile phone. Sadly, my thumbs hurt too much to backspace. More dictating. less typing.

handlers/load_test_handler.go
kanishkarj: Done

mesheryctl/internal/cli/root/perf/perf.go
kanishkarj: Hmm, interesting point. Another thing to consider is, maybe we might have to restructure the SMPS PerformanceTestConfig a little. The plan is to support 3 types of load test if I am correct: HTTP, TCP and gRPC. And, the three of them might have a different way to invoke the tests right, so a lot of the test configs must end up being specific to the test type? For example, request headers/cookie/body makes sense only in the case of HTTP and not in the other two! Similarly, we might have a specific config to TCP and gRPC too. I guess we should consider this in the SMPS!

@kushthedude any thoughts on this?

go.mod
kanishkarj: Yes, until we merge the changes in the SMPS repo.

mesheryctl/internal/cli/root/perf/perf.go
kanishkarj: I think error logging is handled separately... I am not sure who, But someone who restructured the commands to return errors instead of logging them?

mesheryctl/internal/cli/root/perf/perf.go
kanishkarj: Currently, the other configs are ignored. I am not sure if it should be that way though. Currently there is an API  endpoint `/api/load-test-smps` that we are using that expects and SMPS file in the request. So if we want to override the test details in the SMPS then we will have to modify the SMPS file and then send it. Also, considering we are moving towards multiple clients and multiple endpoints, it might become tedious to specify everything in detail using CLI flags. We should discuss and decide on this. @leecalcote ?

mesheryctl/internal/cli/root/perf/perf.go
kanishkarj: Done

mesheryctl/internal/cli/root/perf/perf.go
kanishkarj: Done

handlers/load_test_handler.go
kushthedude: I guess so

kanishkarj:
What else has to be done in this PR? @leecalcote 
Also, the performance test with CLI flags won't work now. Because earlier it was forwarding requests to /api/load-test-smps, but because we modified it to accept actual SMPS it doesn't work. I haven't fixed this yet because I have a question, in the CLI test should we create an SMPS file and create a request with it?

mesheryctl/internal/cli/root/perf/perf.go
leecalcote: Yes, we will likely need a traffic proto. For now, HTTP GET, gRPC ping, and TCP open/close can be supported by logic in `mesheryctl` only.

mesheryctl/internal/cli/root/perf/perf.go
leecalcote: @nitishm maybe?

nitishm: > @nitishm maybe?

Yup I switched to cobra's RunE function.

kushthedude:
> What else has to be done in this PR? @leecalcote
> Also, the performance test with CLI flags won't work now. Because earlier it was forwarding requests to /api/load-test-smps, but because we modified it to accept actual SMPS it doesn't work. I haven't fixed this yet because I have a question, in the CLI test should we create an SMPS file and create a request with it?

Can't we make load-test endpoint answer the Flags request ? @kanishkarj

mesheryctl/pkg/utils/helpers.go
```diff
@@ -343,3 +343,48 @@ func IsValidSubcommand(available []*cobra.Command, sub string) bool {
 	}
 	return false
 }
+
```
kushthedude, line 346: @kanishkarj this function is already added, I don't know why did the mesheryctl build passed on this. Can you fix this ?

go.mod
```diff
@@ -12,13 +12,15 @@ require (
 	github.com/docker/docker v1.13.1
 	github.com/docker/go-connections v0.4.0 // indirect
 	github.com/docker/go-units v0.4.0 // indirect
+	github.com/ghodss/yaml v1.0.0
```
kushthedude, line 15: I don't know much about marshal/unmarshal, but don't we have go-yaml already added ? Is the following better or both have different use-cases?

kanishkarj:
> > What else has to be done in this PR? @leecalcote
> > Also, the performance test with CLI flags won't work now. Because earlier it was forwarding requests to /api/load-test-smps, but because we modified it to accept actual SMPS it doesn't work. I haven't fixed this yet because I have a question, in the CLI test should we create an SMPS file and create a request with it?
> 
> Can't we make load-test endpoint answer the Flags request ? @kanishkarj 

Yes!

go.mod
kanishkarj: Yes, the purpose of this package is different. This is used to convert YAML obj to JSON obj

mesheryctl/internal/cli/root/perf/perf.go
kanishkarj: I don't exactly know how cobra's RunE fn works, @nitishm If I log it separately will it get redundant?

mesheryctl/pkg/utils/helpers.go
kanishkarj: So do I have to remove it?

kanishkarj:
@leecalcote would you give this a run, and let me know if there is anything to fix?

kushthedude:
@kanishkarj Can you resolve the failing build?

mesheryctl/pkg/utils/helpers.go
kanishkarj: This function is still in use right? I saw in RunE of system subcommand.

kushthedude: yes

kanishkarj:
@kushthedude Done!

leecalcote:
If a user uses an invalid auth token, they receive the following error:
```
 ./mesheryctl perf --file ./smps-test.yaml --token ~/Downloads/auth.json
Error: failed to make request to 
See https://meshery.layer5.io/docs/guides/mesheryctl#performance-management for usage details
: Post /: stopped after 10 redirects
Usage:
  mesheryctl perf [flags]
.
.
.
```

leecalcote:
If a user specifies an invalid `loadGenerator` (e.g. "forti" - missing the "o") or an invalid `protocol` (e.g. "tcp"), these invalid parameters are ignored and the test is executed anyway. It is executed successfully with results persisted.

leecalcote:
@kanishkarj We can document the bugs above and move forward with the merge of this PR if it's helpful to do so.

kanishkarj:
> When trying to use the same token twice, I get the following error:
> 
> ```
> Error: failed to update auth details
> See https://meshery.layer5.io/docs/guides/mesheryctl#performance-management for usage details
> : invalid character '<' looking for beginning of value
> Usage:
>   mesheryctl perf [flags]
> .
> .
> .
> ```
> 
> "use the same token twice" meaning that I'm able to run a test successfully once with an auth token, but not use that same token to run a subsequent test. A new token must be downloaded using "Get Token"

I shall fix this issue, actually what is happening is the the token file is getting overwritten with some HTML response from the first request. Thus, it reads the '<' character

kanishkarj:
> If a user uses an invalid auth token, they receive the following error:
> 
> ```
>  ./mesheryctl perf --file ./smps-test.yaml --token ~/Downloads/auth.json
> Error: failed to make request to 
> See https://meshery.layer5.io/docs/guides/mesheryctl#performance-management for usage details
> : Post /: stopped after 10 redirects
> Usage:
>   mesheryctl perf [flags]
> .
> .
> .
> ```
> 
> "invalid auth token" meaning an auth token from a different user account, not the currently authenticated user.

Oh, I shall look into this.

kanishkarj:
> Inside the SPMS file, if a user specifies an invalid `loadGenerator` (e.g. "forti" - missing the "o") or an invalid `protocol` (e.g. "tcp"), these invalid parameters are ignored and the test is executed anyway. It is executed successfully with results persisted. Is the intention to support these fields in a later PR?
> 
> If a user changes the `duration` inside the smps file, this value is honored.

This issue came up because in the proto file we changed loadGenerator type to string from enum. I can change it back to enum in the proto file, or I can do a switch case check here. Which do you think would be better?

kanishkarj:
I have fixed the auth issues @leecalcote.

leecalcote:
Thank you, @kanishkarj. I see that some of the issues are addressed. 👍 One or two of the minor issues seem to still be around. Let's merge this PR and file separate issues, so that we can progress here.

kanishkarj:
Oh, sure!

PR #1148, opened 2020-06-18
Feature/smps proto

kanishkarj:
**Description**

CRUD operations on test config update using the proto message from SMPS.

Create (POST), returns TEST_ID :
```shell
curl --location --request POST 'localhost:9081/api/load-test-prefs' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--header 'Cookie: meshery-provider=None' \
--data-urlencode 'qps=10' \
--data-urlencode 't=10s' \
--data-urlencode 'c=10' \
--data-urlencode 'gen=fortio' \
--data-urlencode 'protocol=TCP' \
--data-urlencode 'headers={"h1":"v1"}' \
--data-urlencode 'cookies={"h1":"v1"}' \
--data-urlencode 'reqBody=reqBody' \
--data-urlencode 'contentType=type' \
--data-urlencode 'endpoint=http://google.com' \
--data-urlencode 'labels={"h1":"v1"}'
```

Update (PUT):
```shell
curl --location --request PUT 'localhost:9081/api/load-test-prefs' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--header 'Cookie: meshery-provider=None' \
--data-urlencode 'qps=100' \
--data-urlencode 't=10s' \
--data-urlencode 'c=10' \
--data-urlencode 'gen=fortio' \
--data-urlencode 'protocol=TCP' \
--data-urlencode 'headers={"h1":"v1"}' \
--data-urlencode 'cookies={"h1":"v1"}' \
--data-urlencode 'reqBody=reqBody' \
--data-urlencode 'contentType=type' \
--data-urlencode 'endpoint=http://github.com' \
--data-urlencode 'labels={"h1":"v1"}' \
--data-urlencode 'uuid=55dc16e3-a5c1-48e2-a76c-4a8fce309d51'
```

Read (GET):
```shell
curl --location --request GET 'localhost:9081/api/load-test-prefs?uuid=<TEST_ID>' \
--header 'Cookie: meshery-provider=None'
```

Read All (GET):
```shell
curl --location --request GET 'localhost:9081/api/load-test-prefs' \
--header 'Cookie: meshery-provider=None'
```

Remove (DELETE):
```shell
curl --location --request DELETE 'localhost:9081/api/load-test-prefs?uuid=<TEST_ID>' \
--header 'Cookie: meshery-provider=None'
```

**Notes to reviewers**
* This needs a bit more refining before finalizing.
* Lacks frontend.

#### Checklist

[...]

PR #1158, opened 2020-06-21
Create/Delete/Read on user test profiles

kanishkarj:
**Description**

CRUD operations on test config update using the proto message from SMPS.

Create (POST), returns TEST_ID :
```shell
curl --location --request POST 'localhost:9081/api/user/test-prefs' \
--header 'Content-Type: application/json' \
--header 'Cookie: meshery-provider=None' \
--data-raw '{
  "smpsVersion": "v0.1",
  "name": "quick-test",
  "labels": {
    "h1": "v1"
  },
  "clients": [
    {
      "internal": false,
      "loadGenerator": "fortio",
      "protocol": "http",
      "connections": 10,
      "rps": 10,
      "headers": {
        "header1": "val1"
      },
      "cookies": {
        "cookie1": "val1"
      },
      "body": "ReqBody",
      "contentType": "application/txt",
      "endpointUrl": [
        "https://githudfgb.com"
      ]
    }
  ],
  "duration": "1s"
}'
```

Read (GET):
```shell
curl --location --request GET 'localhost:9081/api/user/test-prefs?uuid=<TEST_ID>' \
--header 'Cookie: meshery-provider=None'
```

Read All (GET):
```shell
curl --location --request GET 'localhost:9081/api/user/test-prefs' \
--header 'Cookie: meshery-provider=None'
```

Remove (DELETE):
```shell
curl --location --request DELETE 'localhost:9081/api/user/test-prefs?uuid=<TEST_ID>' \
--header 'Cookie: meshery-provider=None'
```

**Notes to reviewers**
* This needs a bit more refining before finalizing.
* Lacks frontend.
* This is a followup on #1148

#### Checklist

- [] [Signed commits](https://github.com/layer5io/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin).
- [] PR title includes `[<name of area>]:<description>`

### DIFF
--- a/handlers/load_test_preferences_handler.go
+++ b/handlers/load_test_preferences_handler.go
@@ -177,7 +177,7 @@ func (h *Handler) UserTestPreferenceDelete(w http.ResponseWriter, req *http.Requ
 		return
 	}
 	if err := provider.SMPSTestConfigDelete(req, testUUID); err != nil {
-		logrus.Error("error deleting testConfig: %v", err)
+		logrus.Errorf("error deleting testConfig: %v", err)
 		http.Error(w, "error deleting testConfig", http.StatusBadRequest)
 		return
 	}


kanishkarj:
We will have to consider how and where we will be storing the performance test profiles. @leecalcote I am assuming this should be a functionality of the provider, shall I add function definitions for the same in the provider interface?

leecalcote:
Right on. Yes, please, @kanishkarj.

handlers/load_test_preferences_handler.go
```diff
@@ -75,3 +80,197 @@ func (h *Handler) LoadTestPrefencesHandler(w http.ResponseWriter, req *http.Requ
...
+		switch proto {
+		case "HTTP":
+			protocol = SMPS.PerformanceTestConfig_Client_http
+		case "gRPC":
+			protocol = SMPS.PerformanceTestConfig_Client_grpc
+		case "TCP":
+			protocol = SMPS.PerformanceTestConfig_Client_tcp
+		default:
+			logrus.Error("invalid value for protocol")
+			http.Error(w, "please provide a valid value for gen (load generator)", http.StatusBadRequest)
```
kushthedude, line 168: Wrong error message over here

handlers/load_test_preferences_handler.go
```diff
@@ -75,3 +80,197 @@ func (h *Handler) LoadTestPrefencesHandler(w http.ResponseWriter, req *http.Requ
...
+		EmitDefaults: true,
+	}
+	val, err := m.MarshalToString(c.Val)
+	return []byte(val), err
+}
+
+// UserTestPreferenceHandler is used for persisting load test preferences
+func (h *Handler) UserTestPreferenceHandler(w http.ResponseWriter, req *http.Request, prefObj *models.Preference, user *models.User, provider models.Provider) {
+
+	if req.Method == http.MethodPost {
```
kushthedude, line 99: Can we create two different handlers for Post & Delete rather than putting complete boilerplate code in the conditionals?

router/server.go
```diff
@@ -22,9 +22,11 @@ func NewRouter(ctx context.Context, h models.HandlerInterface, port int) *Router
 	mux.HandleFunc("/api/providers", h.ProvidersHandler)
 	mux.HandleFunc("/provider/", h.ProviderUIHandler)
 
+	mux.Handle("/api/config/sync", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.SessionSyncHandler))))
+
 	mux.Handle("/api/user", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.UserHandler))))
 	mux.Handle("/api/user/stats", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.AnonymousStatsHandler))))
-	mux.Handle("/api/config/sync", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.SessionSyncHandler))))
```
kushthedude, line 27: Is this route not being used?

handlers/load_test_preferences_handler.go
kanishkarj: Totally, I was considering this. I was actually thinking we must use gorilla/mux.  That will allow us to build better REST APIs, check [this](https://gist.github.com/ssaumyaranjan7/5732157a21835c9f08a59be6a9317c0b#file-bookstore-routes-go-L10) out. What's your opinion on this?

handlers/load_test_preferences_handler.go
kanishkarj: Instead of reading in these values as form input, I am considering to read it as JSON, what do you suggest?

kushthedude:
> There is an issue with supporting DELETE operation on the test configs. How should we handle the case where the test config has been deleted, but there are perf test results referring to it?
> @leecalcote @kushthedude Thoughts on this?

A doubt here, we are referencing to test-profiles while storing results, like the result & the test-profile should be seperate imo. If I delete my test-profile or update it, this should not have any change on the test-results done using the profile, right ?

kushthedude:
> ```shell
> curl --location --request GET 'localhost:9081/api/load-test-prefs' \
> --header 'Cookie: meshery-provider=None'
> ```

Is this a typo @kanishkarj ?

kanishkarj:
> > There is an issue with supporting DELETE operation on the test configs. How should we handle the case where the test config has been deleted, but there are perf test results referring to it?
> > @leecalcote @kushthedude Thoughts on this?
> 
> A query here, we are referencing to test-profiles while storing results, like the result & the test-profile should be seperate imo. If I delete my test-profile or update it, this should not have any change on the test-results done using the profile, right ?

Yes, exactly. But there are issues of inconsistency here. Let's say a test profile was updated, in a way it invalidates the old test results. Even worse, if a profile was deleted, we'd have dangling references from test results to non existent profiles.

leecalcote:
Gorilla/mix sounds good. @nveenjain

kanishkarj:
I have added gorilla/mux but not have restructured the code to exploit its features (might be a separate PR on its own). Also the Store SMPS takes in JSON data instead of application/x-www-form-urlencoded.

go.mod
```diff
@@ -15,10 +15,12 @@ require (
 	github.com/gofrs/uuid v3.3.0+incompatible
 	github.com/gogo/protobuf v1.3.1 // indirect
 	github.com/golang/protobuf v1.4.2
+	github.com/gorilla/mux v1.7.4
 	github.com/gosimple/slug v1.9.0
 	github.com/grafana-tools/sdk v0.0.0-20200411085644-f7626bef00b3
 	github.com/jinzhu/copier v0.0.0-20190924061706-b57f9002281a
 	github.com/layer5io/gowrk2 v0.0.0-20191111234958-a4c9071c0f87
+	github.com/layer5io/service-mesh-performance-specification v0.0.0-20200622030323-136a17f58ad9
```
kushthedude, line 23: @kanishkarj Can we refer to the SMPS v0.1.0 here?

kushthedude:
> I have added gorilla/mux but not have restructured the code to exploit its features (might be a separate PR on its own). Also the Store SMPS takes in JSON data instead of application/x-www-form-urlencoded.

So the curl request in mentioned above needs to be changed into JSON format right ?

leecalcote:
Resolving merge conflicts...

kanishkarj:
@leecalcote regarding storing user test profiles on a local database in case of default provider: 
* we can modify meshery/models/bitcask_results_persister.go to be more general and support storing performance test configs too.
* We can create a copy of meshery/models/bitcask_results_persister.go specific to storing performance test configs.

which do you think we should proceed with?

models/meshery_remote_provider.go
```diff
@@ -489,3 +490,16 @@ func (l *MesheryRemoteProvider) ExtractToken(w http.ResponseWriter, r *http.Requ
 		http.Error(w, "unable to extract auth details", http.StatusInternalServerError)
 	}
 }
+
+func (l *MesheryRemoteProvider) SMPSTestConfigStore(req *http.Request, perfConfig *SMPS.PerformanceTestConfig) (string, error) {
```
leecalcote, line 494: Will we need to account for this in Meshery Cloud?

models/providers.go
```diff
@@ -58,4 +59,9 @@ type Provider interface {
 	PublishMetrics(tokenVal string, data *MesheryResult) error
 	GetResult(*http.Request, uuid.UUID) (*MesheryResult, error)
 	RecordPreferences(req *http.Request, userID string, data *Preference) error
+
+	SMPSTestConfigStore(req *http.Request, perfConfig *SMPS.PerformanceTestConfig) (string, error)
```
leecalcote, line 63: Given the short-lived nature of the local provider, I wonder if its best to include here or not. Hmm.

handlers/load_test_preferences_handler.go
leecalcote: Use of gorilla mux is a great improvement. 👍

handlers/load_test_preferences_handler.go
```diff
@@ -2,13 +2,18 @@
...
 	"net/http"
 	"strconv"
 	"time"
 
+	"github.com/golang/protobuf/jsonpb"
+	"google.golang.org/protobuf/encoding/protojson"
 	"encoding/json"
 
 	"github.com/layer5io/meshery/models"
+	SMPS "github.com/layer5io/service-mesh-performance-specification/spec"
```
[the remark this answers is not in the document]
kanishkarj: I think so ..

router/server.go
```diff
@@ -5,26 +5,32 @@ import (
...
+
+	mux.Handle("/api/user/test-prefs", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.UserTestPreferenceStore)))).Methods(http.MethodPost)
+	mux.Handle("/api/user/test-prefs", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.UserTestPreferenceGet)))).Methods(http.MethodGet)
+	mux.Handle("/api/user/test-prefs", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.UserTestPreferenceDelete)))).Methods(http.MethodDelete)
```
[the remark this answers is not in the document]
leecalcote: Never mind. I get it now.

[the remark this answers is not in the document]
kanishkarj: You can have one handler per endpoint. But the advantage with having one per verb is that you won't have to do an if statement for each http verb inside a single handler. You can have multiple handlers, one for each verb.

models/validators.go
```diff
@@ -0,0 +1,40 @@
...
+	SMPS "github.com/layer5io/service-mesh-performance-specification/spec"
+	"github.com/pkg/errors"
+)
+
+func SMPSPerformanceTestConfigValidator(perfTest *SMPS.PerformanceTestConfig) error {
+	if perfTest.Name == "" {
+		return errors.Errorf("Error: name field is blank")
+	}
+	if _, err := time.ParseDuration(perfTest.Duration); err != nil {
+		return errors.Wrapf(err, "error parsing test duration, please refer to: https://meshery.layer5.io/docs/guides/mesheryctl#performance-management")
```
leecalcote, line 16: Can we make a new global const variable for DOCS_BASEURL?

leecalcote:
> @leecalcote regarding storing user test profiles on a local database in case of default provider:
> 
> * we can modify meshery/models/bitcask_results_persister.go to be more general and support storing performance test configs too.
> * We can create a copy of meshery/models/bitcask_results_persister.go specific to storing performance test configs.
> 
> which do you think we should proceed with?

I'm inclined to the latter. While related, Test Profiles and Test Results have independent lifecycles; one can exist without the other. Given that each are their own construct; that each have full CRUD operations, keeping clean separation of their data models makes sense to me.

leecalcote:
I'm not sure if there is an open question about the Delete operation for either the test **profiles** or the test **results**. But, I'll note that for those persisted by the Meshery remote provider, we'll want to support a soft delete (to tombstone the records and exclude them from resultsets, but not actually delete them) for both test **profiles** and test **results**.

Hard deletes are fine for the Meshery local provider.

router/server.go
```diff
@@ -5,26 +5,32 @@ import (
...
 type Router struct {
-	s    *http.ServeMux
+	s    *mux.Router
 	port int
 }
 
 // NewRouter returns a new ServeMux with app routes.
 func NewRouter(ctx context.Context, h models.HandlerInterface, port int) *Router {
-	mux := http.NewServeMux()
+	mux := mux.NewRouter()
```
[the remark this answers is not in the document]
kanishkarj: Yes!

models/providers.go
kanishkarj: Actually the provider object would hold a bitcast storage. So we would be using that to CRUD these things. That should be fine right?

kushthedude:
@kanishkarj What is the test URL for the same? The above request seems not to be working

leecalcote:
@kanishkarj is this still working for you? I'm not able to test. I'm not sure if it's my issue or not.

kanishkarj:
I shall test this rightaway

leecalcote:
Thank you, sir. I’m having a side issue with running the provider-UI, which may be causing my issue in terms of being able to initiate a none provider session.

kanishkarj:
Oh yes, I was not able to get the Provider UI too. I think if I merge to the latest master branch the issue must be solved. Currently you can still test through the requests from curl.

kanishkarj:
There is some issue loading the front-end. @kushthedude could you help me out with it?

kanishkarj:
For some reason in my PR, the content-types aren't being set right by meshery server. Let's say the app requests a js file, then meshery doesn't return it with content-type of application/javascript, rather it sends it as text/plain. Thus the browser blocks these responses (security reasons).

kushthedude:
> Oh yes, I was not able to get the Provider UI too. I think if I merge to the latest master branch the issue must be solved. Currently you can still test through the requests from curl.

@kanishkarj, I & @leecalcote were not able to test the CURL too, the post request given above was failing.

I am not sure but can the problem be due to intermixing of routers, as we have used gorilla-mux for the new routes but we have not configured above old routes according to how gorilla works?

kanishkarj:
Interesting, it worked on my machine. What error did you get? Also, did you make requests with the updated curl comand above? Because, we have moved to json from form.

kushthedude:
> Interesting, it worked on my machine. What error did you get? Also, did you make requests with the updated curl comand above? Because, we have moved to json from form.

Oh I don't guess I sent JSON object, let me quickly retry

kanishkarj:
> > Interesting, it worked on my machine. What error did you get? Also, did you make requests with the updated curl comand above? Because, we have moved to json from form.
> 
> Oh I don't guess I sent JSON object, let me quickly retry

Also when you try, try it out with curl -v ...   This will give a detailed reason why the request failed

kushthedude:
@kanishkarj I tried it out, POST & GET for a single test object are working but there is some error with getting for all test configs.
```
DEBU[0059] page , pageSize:
ERRO[0059] unable to parse page number: strconv.ParseUint: parsing "": invalid syntax
ERRO[0059] error fetching test configs
```
This was the error I got

kanishkarj:
Oh.. yes.. the curl command for that is wrong. Basically, to the request could you add 2 http url query parameters:
page=1 and pageSize=10? Also could you check the same for remote provider?

kushthedude:
> Oh.. yes.. the curl command for that is wrong. Basically, to the request could you add 2 http url query parameters:
> page=1 and pageSize=10? Also could you check the same for remote provider?

I tried adding the query params to URL but it is still not working 
command used
` curl -v --location --request GET 'localhost:9081/api/user/test-prefs?page=1&pageSize=10' \
--header 'Cookie: meshery-provider=None'`

Error logs:
```
DEBU[0818] page 1, pageSize: 10
DEBU[0818] received page: 1, page size: 10, total: 1
DEBU[0818] computed start index: 10, end index: 19
ERRO[0818] error fetching test configs
```

kanishkarj:
the ui does not work Because of some request type errors. I feel it's something to do with the go code. So, let's say the ui loads up and it has to download some js files. What happens is, the meshery front-end requests for the js file with conten-type as plaintext (weird). And the browser blocks this, saying that should have been application/javascript. Check the network tab in the page.

@kushthedude could you look into this?

kushthedude:
I will check into it, thanks for the detailed description @kanishkarj

PR #1175, opened 2020-06-25
chore: Use pinned versions cache from gomod rather than updating

kushthedude:


kushthedude:
This will save us from breaking changes made in the dependencies without infecting our CI which has been affected multiple times in past.

Issue #211, opened 2019-08-20
M2: General Interface to Load Generators


leecalcote:
**Description**
See #126 for full description of Load Generator Integration Epic.

**Milestone 2: Build and Packaging**

- [ ] Create an alias type “SupportedLoadTestMethods” of type int. 
- [ ] Define constants for each of the supported load test methods.
  - [ ] HTTP
  - [ ] gRPC
  - [ ] TCP
- [ ] Create an interface with common methods for all the load generators “LoadGeneratorInterface”. Methods we need in the interface at the moment:
  - [ ] SupportedMethods() []
  - [ ] SupportedLoadTestMethods
RunLoadTest(*models.LoadTestOptions) (map[string]interface{}, *periodic.RunnerResults, error)

- [ ] Move the fortio code, which is directly accessed by meshery today, under an implementation of “LoadGeneratorInterface”, “FortioLoadGenerator”.
- [ ] Update the existing handlers to work with instances of the interface to remove the tight coupling (the instances can be created in main and propagated into the handler instance).

**Additional context**
See [Load Generator Integration design doc](https://docs.google.com/document/d/1jZMOih3Qn1ixrq1Ve6fkdZZ4TuKNoF_qfClDZTnLNcw/edit#).

leecalcote:
@girishranganathan in consideration of the [Meshery Extension Points](https://docs.google.com/document/d/1E7NWFy5ZVEPX2W892mfVG-3yku1uci2KRDU42tNRZwA/edit) doc, do we consider this general interface complete? 

// @raghavaggarwal99

leecalcote:
@kushthedude given PR #1180, will you update the tasks in this issue to reflect current status?

kushthedude:
@leecalcote The last part should fix when we have gRPC test functionality also.

PR #1180, opened 2020-07-01
feat: Initiate loadGeneratorInterface

kushthedude:
Refers #211


leecalcote:
https://github.com/layer5io/meshery/pull/1156#discussion_r443153953

kushthedude:
> #1156 (comment)

No it should not change, the only change will happen due to the SupportedLoadTestMethods which we will invoke through SMPS.

models/load-test.go
```diff
@@ -28,6 +28,20 @@ func (l LoadGenerator) Name() string {
 	return string(l)
 }
 
+// SupportedLoadTestMethods - represents the load test method type
+type SupportedLoadTestMethods int
```
leecalcote, line 32: What factors weigh on a decision to or to not use the SMPS traffic.proto? Would that be a good or bad thing?

leecalcote:
@kushthedude this works for me in fortio tests. There are a couple merge conflicts to address in load_test_handler.go.

kushthedude:
@leecalcote conflicts resolved

leecalcote:
@kushthedude I wasn't able to test out the wrk2 (build in docker), but I assume that we're good to go with this first portion of the interface.

kushthedude:
> @kushthedude I wasn't able to test out the wrk2 (build in docker), but I assume that we're good to go with this first portion of the interface.

It is working too, I had tested it before changes.

PR #1244, opened 2020-07-08
[UI]Feat: Performance Tests: A drop-down list for the `Duration` field

### LINKED ISSUE #1149: [UI] Performance Tests: A drop-down list for the `Duration` field
**Current Behavior**
The Duration field is current a text field. While the tooltip provides some reference, it has been noted that users may not be readily aware of the full set of acceptable values for the Duration field are.

See this discussion of the issue: https://www.youtube.com/watch?v=G6wDVdTXwcY&feature=youtu.be&t=2013

**Desired Behavior**
Use of an autocomplete field in combination with the reference information in the tooltip will help guide users to both successfully filling in the field and to understanding the full set of possible acceptable values.

---
**Resources**
Thanks to @MikeBorovik:

- https://material-ui.com/components/autocomplete/ - an autocomplete reference
- https://www.npmjs.com/package/react-autocomplete - a ready-to-use package for react

PmcFizz:
Signed-off-by: fizz <pangmengchen@deepexi.com>

**Description**

This PR fixes #1149 

**Notes to reviewers**

#### Checklist


PmcFizz:
what allow users to enter custom values outside of the confined list of presets  is a little hard , but I will try my best.

yes  i will remove `2.5h` this value

Users are given the option to adjust the default values used in performance tests,  this feat could be next PR

I studuy that good prodeuct need strict stander  Thanks
@leecalcote

ui/public/static/style/index.css
```diff
@@ -0,0 +1,6 @@
+.error-autocomplete-value .MuiOutlinedInput-notchedOutline {
```
kushthedude, line 1: Why custom css?

ui/components/MesheryPerformanceComponent.js
```diff
@@ -115,18 +119,21 @@ class MesheryPerformanceComponent extends React.Component {
     if (name === 'url' && event.target.value !== '') {
       this.setState({ urlError: false });
     }
-    if (name === 't' && (event.target.value.toLowerCase().endsWith('h')
```
kushthedude, line 118: Why this check is not needed 🤔 ?

ui/components/MesheryPerformanceComponent.js
```diff
@@ -78,6 +79,8 @@ const styles = (theme) => ({
   },
 });
 
+const prePopulatedOptions = ['15s', '30s', '1m', '3m', '5m', '10m', '30m', '1h', '2.5h', '5h', '10h', '1d']
```
kushthedude, line 82: These are a lot to options we just want to make users know how we support entering the values.

ui/components/MesheryPerformanceComponent.js
PmcFizz: because when you change this drop-down will execute other function handleDurationChange

ui/components/MesheryPerformanceComponent.js
PmcFizz: I don't understand you mean. just a review msg ?

kushthedude: Hi @PmcFizz , We want the dropdown to just let users know what sort of string or what sort of input they want to give. But here we have given 10 pre-populated options which I don't think is mandatory, I think just 4 options like 10s, 1m, 1h should be sufficient to let the user know that we accept the following input. I guess we can cut out some options just to maintain it as a guide of how you should enter the input in following field.

leecalcote: Gents, to get this PR over the line, for now, let's stick with the set of values outlined in the issue - https://github.com/layer5io/meshery/issues/1149#issuecomment-654844521.

ui/public/static/style/index.css
PmcFizz: when input error , the border will be red to remind people should input a current value

PmcFizz:
I have solve  the problem that don‘’t allow users to enter custom values outside of the confined list of presets
but I don't sure  other place is affected ， if there is some problem  you can @me
Thanks

leecalcote:
> hi @leecalcote
> I have solve the problem that don‘’t allow users to enter custom values outside of the confined list of presets
> but I don't sure other place is affected ， if there is some problem you can @me
> Thanks

@PmcFizz, yes, good request for an example. @kushthedude would you mind highlighting an example of this behavior in the drop-down component used in the Settings page for populating service mesh adapters?

leecalcote:
@PmcFizz, I believe [this component](https://github.com/layer5io/meshery/blob/master/ui/components/MeshAdapterConfigComponent.js) is the one that I highlighted in [this screen recording](https://github.com/layer5io/meshery/issues/1149#issuecomment-654844521).

leecalcote:
> @PmcFizz, I believe [this component](https://github.com/layer5io/meshery/blob/master/ui/components/MeshAdapterConfigComponent.js) is the one that I highlighted in [this screen recording](https://github.com/layer5io/meshery/issues/1149#issuecomment-654844521), which can be used as an example of a component providing both a predefined set of values, and also, allowing users to type in a custom value.

@kushthedude, when you have a moment, will you confirm that I'm highlighting the right example for @PmcFizz?

leecalcote:
@PmcFizz you're so close on this PR. I'm hoping you can have another look. We're good to tie-off only a portion of it and open another PR for some of the rest of the functionality.

PmcFizz:
> @PmcFizz you're so close on this PR. I'm hoping you can have another look. We're good to tie-off only a portion of it and open another PR for some of the rest of the functionality.

sorry  I am a little busy recently, I will look at your require

PmcFizz:
> > @PmcFizz, I believe [this component](https://github.com/layer5io/meshery/blob/master/ui/components/MeshAdapterConfigComponent.js) is the one that I highlighted in [this screen recording](https://github.com/layer5io/meshery/issues/1149#issuecomment-654844521), which can be used as an example of a component providing both a predefined set of values, and also, allowing users to type in a custom value.
> 
> @kushthedude, when you have a moment, will you confirm that I'm highlighting the right example for @PmcFizz?

I careful read your require,  Maybe I make a mistake.  at first  I ues  Autocomplete of @material-ui/lab to realize the feature what user can select and customer value to Duration. but today I see you provide this example the [components](https://github.com/layer5io/meshery/blob/master/ui/components/MeshAdapterConfigComponent.js) . this component use custome component named **ReactSelectWrapper**.   and if user input a custome value in this component ,they have to click Create. 

So what you want me to do ?   use **ReactSelectWrapper** to replace **Autocomplete**  on [MesheryPerformanceComponent.js](https://github.com/layer5io/meshery/blob/master/ui/components/MesheryPerformanceComponent.js) ? 
or use **Autocomplete** to replace **ReactSelectWrapper**  on  [MeshAdapterConfigComponent](https://github.com/layer5io/meshery/blob/master/ui/components/MeshAdapterConfigComponent.js) 

or let me on [MesheryPerformanceComponent.js](https://github.com/layer5io/meshery/blob/master/ui/components/MesheryPerformanceComponent.js)   add create option like on [MeshAdapterConfigComponent](https://github.com/layer5io/meshery/blob/master/ui/components/MeshAdapterConfigComponent.js)

Aisuko:
@leecalcote Any feedback?

leecalcote:
Yay!! @PmcFizz 🎈 😎

PR #1258, opened 2020-07-12
chore: remove call to git at every submethod

kushthedude:


leecalcote:
Let's catch up with @kumarabd on this. The `git` command is being invoked to facilitate realistic responses to the `mesheryctl version` command while developing. Before removing entirely, a short discussion on alternative approaches would be good.

kushthedude:
This runs the git command at every make command from which it is removed, I
dont suppose we need to show the version details unless user wants to know.

leecalcote:
This may be a conversation we should pick up on in Slack or on Wednesday’s dev call, as I suspect, a verbal discussion may help.

In essence, the `git` routine was included by Abishek to improve the developer experience in particular for the `mesheryctl version` command. This was a recent need for Abishek when he was creating the server-side version lookup.

kumarabd:
@kushthedude  In which scenario do you not want the version to be displayed? Because I had added the git command to inject client version for every build, on the idea that any binary that is been built should be associated with a version so that it will be easier to figure out which one we are using.

kushthedude:
> @kushthedude  In which scenario do you not want the version to be displayed? Because I had added the git command to inject client version for every build, on the idea that any binary that is been built should be associated with a version so that it will be easier to figure out which one we are using.

Does this mean that call on every make action like make run-local, make run-cloud etc is just added to inject the client version?
We would just want to to show these details when mesheryctl -v is invoked.

kumarabd:
Yes, this needs to be done during the build time because once the binary is out we won't be able to fetch the git details. Also this needs to run on every make command because we consider each of them as an independent build, meaning each make produces it's own binary version of it. @kushthedude  if you have a better suggestion on handling this please pitch in.

kushthedude:
Can we invoke it in the dockerfile ? As for Meshery Users, the version output will not give server-version.

kumarabd:
@kushthedude I can put it in the dockerfile however it will be needed for non-docker builds. Let me make some quick changes for this.

kushthedude:
> @kushthedude I can put it in the dockerfile however it will be needed for non-docker builds. Let me make some quick changes for this.

non-docker as if development binary created for development or testing purpose on local, right @kumarabd ?

kumarabd:
@kushthedude  yes that's correct. Or if someone doesn't want a  docker build instead they are using a different container orchestration.

kushthedude:
> @kushthedude yes that's correct. Or if someone doesn't want a docker build instead they are using a different container orchestration.

Then we need to have a discussion on this rather than moving forward with this PR

kumarabd:
> Then we need to have a discussion on this rather than moving forward with this PR
Sure, Lets catchup sometime to discuss on this.

kushthedude:
Not required now!

PR #1307, opened 2020-07-23
[CI] Add PR auto labeler action

kushthedude:
/cc @leecalcote


leecalcote:
Next up.... filepath mapping? Then ci.yml conditional logic?

kushthedude:
> Next up.... filepath mapping? Then ci.yml conditional logic?

Yes correct 🚀

PR #1310, opened 2020-07-24
chore: Change the slider toggle color scheme

kushthedude:
The previous toggle didn't have colour change between true/false state.


kushthedude:
We need to fix the security tests as soon as possible!

kushthedude:
To be merged after CI gets fixed

PR #1364, opened 2020-08-05
fix: Notification error toastr do not disappear 

kushthedude:
- The banner doesn't disappear & doesn't have any cross button
- It appears when an adapter is already configured
- The settings link looks too much odd.
- It does not differentiate b/w configures & missing adapter

here is the screenshot when I ping my istio adapter which is already configured from the dashboard
Screenshot 2020-08-05 at 5 45 43 PM

### DIFF
--- a/ui/components/DashboardComponent.js
+++ b/ui/components/DashboardComponent.js
@@ -185,7 +185,7 @@ class DashboardComponent extends React.Component {
     const self = this;
     this.props.enqueueSnackbar(`${msg}. To configure an adapter, visit`, {
       variant: 'error',
-      autoHideDuration: 5000,
+      autoHideDuration: 2000,
       action: (key) => (
         <>
           <Button


kushthedude:
Due to time limitations, I have just reverted the fix rather than enhancing the toastr

Nikhil-Ladha:
I found the errors that are in the code, the points 1, 2 can be taken care of. The 3rd point for the Settings button, I think the button was made that way intentionally. 4. The function only checks whether the adapter can be ping led or not, so it will not differentiate between missing and configuring the adapter.

kushthedude:
> The function only checks whether the adapter can be ping led or not, so it will not differentiate between missing and configuring the adapter.

So you mean whether my adapter is configured or not, I need to see this error message

kushthedude:
> The 3rd point for the Settings button, I think the button was made that way intentionally

That's too much odd, a simple anchor tag would be way more beautiful

Nikhil-Ladha:
> > The function only checks whether the adapter can be ping led or not, so it will not differentiate between missing and configuring the adapter.
> 
> So you mean whether my adapter is configured or not, I need to see this error message

If your adapter in configured, but it's not reachable only then you will see this message.

Nikhil-Ladha:
> > The 3rd point for the Settings button, I think the button was made that way intentionally
> 
> That's too much odd, a simple anchor tag would be way more beautiful

Agreed, will make some changes to that.

Nikhil-Ladha:
Don't merge this PR, I will push changes here itself.

kushthedude:
> Don't merge this PR, I will push changes here itself.

Sure, Please

Nikhil-Ladha:
Now, it's good to go.

kushthedude:
> Now, it's good to go.

Just reduced the timeout, else looks good!

ui/components/DashboardComponent.js
```diff
@@ -81,7 +81,6 @@ const styles = (theme) => ({
   },
   redirectButton: {
     marginLeft: '-.5em',
-    backgroundColor: '#ccc',
```
leecalcote, line 84: A slightly off color from the exist red notification will draw attention, but perhaps, avoid the jarring affect that this light gray induced.

What will the "Settings" link/button look like now?

Nikhil-Ladha: It will be in ' Black' color, and on hovering a bit of dark red background color will come up.

PR #1374, opened 2020-08-06
chore: Remove redundant step from docker push

kushthedude:


kushthedude:
@leecalcote Did I just see auto-label ?

leecalcote:
This LGTM. I think that we are good to go in terms of removing the properties.yaml altogether.

PR #1389, opened 2020-08-12
Fix for the results page

### LINKED ISSUE #1004: [UI] Performance test results with MeshSync information failing to display
**Description**
Performance test results that include environment information from MeshSync do not display in the /results page.

**Expected Behavior**
The modal window should show the perf test results and environment information gathered by MeshSync (Kubernetes and node information).

**Screenshots**

1. The first results shown here are from a test run when Meshery **is not** connected to Kubernetes (does not run MeshSync). 
1. The second results shown here (in error) are from a test run when Meshery **is** connected to Kubernetes (does run MeshSync).

Untitled

Nikhil-Ladha:

**Description**

This PR fixes #1004 

**Notes for Reviewers**

### DIFF
--- a/ui/components/MesheryResultDialog.js
+++ b/ui/components/MesheryResultDialog.js
@@ -91,19 +91,20 @@ class MesheryResultDialog extends React.Component {
 
     renderMeshesInfo(detectedMeshes) {
       const { classes } = this.props;
+      //const meshes = Object.entries(detectedMeshes);
       const meshes = Object.keys(detectedMeshes);
       return (
         <NoSsr>
           <Typography className={classes.title} variant="h6" id="tableTitle">
             Service Mesh
-            {meshes.length > 1 ? 'es' : ''}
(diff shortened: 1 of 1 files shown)


kushthedude:
Can we create an issue for this as to see why was this happening in the first place?

Nikhil-Ladha:
There is an issue mentioned above, and I think we know already why was the issue coming up.

kushthedude:
> There is an issue mentioned above, and I think we know already why was the issue coming up.

I don't guess we know it yet, or else this would not happen in some occasional cases.

Nikhil-Ladha:
I said that yesterday, why was it happening. As, when we have a mesh configured at that time when we were fetching the results at that time for the "Service Mesh" part of the Results dialog, we were passing a key-value pair to the function which was rendered to the page directly. But, actually, the value was an array of objects as a result the error was occurring. So, instead of passing the value to the dialogue, we are only passing the key to it for the time being.

PR #1392, opened 2020-08-13
Fix for unwanted snackbar in perf preferences tab

Nikhil-Ladha:

- Added the null condition before showing the snackbar.

- Also, added the default values of the fields for the preferences page.

### DIFF
--- a/ui/components/MesherySettingsPerformanceComponent.js
+++ b/ui/components/MesherySettingsPerformanceComponent.js
@@ -42,13 +42,13 @@ class MesherySettingsPerformanceComponent extends React.Component {
   constructor(props) {
     super(props);
     const {
-      qps, c, t
+      qps, c, t, gen
     } = props;
     this.state = {
       qps,
       c,
       t,
-      gen: '',
+      gen,
       blockRunTest: false,
       tError: false,
     };
--- a/ui/lib/store.js
+++ b/ui/lib/store.js
@@ -30,7 +30,7 @@ const initialState = fromJS({
     qps: 0,
     t: '30s',
     c: 0,
-    gen: '',
+    gen: 'fortio',
     ts: new Date(),
   },
   meshAdapters: [],


Nikhil-Ladha:
> We discussed to populate the defaults from the perf page. It would be odd to have values populated at the perf page with preference being empty.

I have added that, and also mentioned it in the PR description.

kushthedude:
> > We discussed to populate the defaults from the perf page. It would be odd to have values populated at the perf page with preference being empty.
> 
> I have added that, and also mentioned it in the PR description.

I dont see where are they added?

kushthedude:
> > We discussed to populate the defaults from the perf page. It would be odd to have values populated at the perf page with preference being empty.
> 
> I have added that, and also mentioned it in the PR description.

Is the generator getting updated?

Nikhil-Ladha:
> > > We discussed to populate the defaults from the perf page. It would be odd to have values populated at the perf page with preference being empty.
> > 
> > 
> > I have added that, and also mentioned it in the PR description.
> 
> Is the generator getting updated?

Yes, it is.

kushthedude:
> > > > We discussed to populate the defaults from the perf page. It would be odd to have values populated at the perf page with preference being empty.
> > > 
> > > 
> > > I have added that, and also mentioned it in the PR description.
> > 
> > 
> > Is the generator getting updated?
> 
> Yes, it is.

Can you recheck it, gen is not populated to fortio by default as it doesn't fetch the value from store.

Nikhil-Ladha:
I am updating the MesheryResultsDialogue.js file with the indentation to fix the failing of UI test.

Nikhil-Ladha:
> > > > > We discussed to populate the defaults from the perf page. It would be odd to have values populated at the perf page with preference being empty.
> > > > 
> > > > 
> > > > I have added that, and also mentioned it in the PR description.
> > > 
> > > 
> > > Is the generator getting updated?
> > 
> > 
> > Yes, it is.
> 
> Can you recheck it, gen is not populated to fortio by default as it doesn't fetch the value from store.

Yes, the load-generators are not getting populated but only the QPS, Concurrent Rqsts, and the Duration fields are.

kushthedude:
> Yes, the load-generators are not getting populated but only the QPS, Concurrent Rqsts, and the Duration fields are.

We would this to also be set to its default value -> fortio

Nikhil-Ladha:
> > Yes, the load-generators are not getting populated but only the QPS, Concurrent Rqsts, and the Duration fields are.
> 
> We would this to also be set to its default value -> fortio

Ohh, okay will add that.

PR #1351, opened 2020-08-02
Fixed the issue with objects in results page

Nikhil-Ladha:
Signed-off-by: Nikhil-Ladha <nikhilladha1999@gmail.com>

This PR fixes #1004 

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #1393, opened 2020-08-13
fix: broken ci for ui tests

kushthedude:
Tests were broken in #1351 , this PR corrects them


Nikhil-Ladha:
Actually, I kept the code commented out instead of removing it, so that we remember to do something about it in the future. Anyways, I was just gonna push the CI fix in my open PR😂

kushthedude:
> Actually, I kept the code commented out instead of removing it, so that we remember to do something about it in the future.

We can reimplement it when we get MeshSync in place 😄 , commented code without reference just increase confusion tbh 😓

PR #1394, opened 2020-08-13
Added the validation for protocol in URL

### LINKED ISSUE #1252: [performance] Improve handing of missing protocol in Performance Test form
**Current Behavior**
If a user enters a domain name (e.g., `google.com`) without specifying the protocol to be used for the endpoint (e.g., `https://`), the user receives this error message:

Screen Shot 2020-07-10 at 12 51 52 PM

This error message is inaccurate. Meshery fails to generate load against the endpoint, because the endpoint string is missing definition of which protocol to use.

**Desired Behavior**
- Client-side form validation should address this issue before the user is allowed to submit the form.

Nikhil-Ladha:

**Description**

This PR fixes #1252 

**Notes for Reviewers**

### DIFF
--- a/ui/components/MesheryPerformanceComponent.js
+++ b/ui/components/MesheryPerformanceComponent.js
@@ -107,6 +107,7 @@ class MesheryPerformanceComponent extends React.Component {
       blockRunTest: false,
       urlError: false,
       tError: '',
+      disableTest : true,
 
       testUUID: this.generateUUID(),
       staticPrometheusBoardConfig,
@@ -119,10 +120,14 @@ class MesheryPerformanceComponent extends React.Component {
     if (name === 'url' && event.target.value !== '') {
       let urlPattern = event.target.value;
       let val = urlPattern.match(/(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g);
-      if ( !val )
+      if ( !val ){
+        this.setState({ disableTest: true });
         this.setState({ urlError: true });
-      else
+      }
+      else{
+        this.setState({ disableTest: false });
         this.setState({ urlError: false });
+      }
     }
     else
       this.setState({ urlError: false });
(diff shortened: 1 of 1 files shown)


ui/components/MesheryPerformanceComponent.js
```diff
@@ -117,8 +117,15 @@ class MesheryPerformanceComponent extends React.Component {
 
   handleChange = (name) => (event) => {
     if (name === 'url' && event.target.value !== '') {
-      this.setState({ urlError: false });
+      let urlPattern = event.target.value;
+      let val = urlPattern.match(/(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g);
+      if ( !val )
+        this.setState({ urlError: true });
+      else
```
kushthedude, line 124: urlError is already `false` in the state, how about we just set `urlError:true` & remove two conditional branches?

Nikhil-Ladha: I added them as these changes are done on the handleChange(), in which the data change is dynamic. So, as soon as the URL is authenticated the error is gone and as the URL is false the error is highlighted. If I remove the else parts then, the above error will occur. 
Also, the outer else is there, so that when the user wipes out the complete URL after typing something wrong in there, at that time the error should be gone. So, the state should be back to false.

ui/components/MesheryPerformanceComponent.js
```diff
@@ -117,8 +117,15 @@ class MesheryPerformanceComponent extends React.Component {
 
   handleChange = (name) => (event) => {
     if (name === 'url' && event.target.value !== '') {
-      this.setState({ urlError: false });
+      let urlPattern = event.target.value;
+      let val = urlPattern.match(/(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g);
+      if ( !val )
+        this.setState({ urlError: true });
```
kushthedude, line 123: You would want to set `blockRunTest:true` when URL is invalid

Nikhil-Ladha: Oh yes, but will have to create a new state as using that one would also cause the circular wheel to come up on the button :')

kushthedude: the button will be disabled in `BlockRunTest:True` so circular wheel won't come.

Nikhil-Ladha: And that's exactly when the circular wheel comes up.. When the blockruntest is true

Nikhil-Ladha: Instead of adding new state, I think just adding the check with urlError state is sufficient.

kushthedude: I tried it on local, UrlError doesn't disable the button rather the server is pinged/

Nikhil-Ladha: Aree, I meant I will be adding that. It is not there currently :').....But, that also is not good to use as initially the urlError will be false and the button won't be disabled. But, we want the button to be disabled until the URL is fine. So, I will add a new state and use that to check it.

leecalcote:
Thanks, gents. Yes, disabled makes sense. in this way, we don't allow the user to shoot themselves in the foot.

PR #1450, opened 2020-08-25
feat: Initialise nighthawk load generator interface

kushthedude:

### DIFF
--- a/helpers/load_test_interface.go
+++ b/helpers/load_test_interface.go
@@ -227,7 +227,7 @@ func NighthawkLoadTest(opts *models.LoadTestOptions) (map[string]interface{}, *p
 	bd, err := json.Marshal(res)
 
 	if err != nil {
-		err = errors.Wrap(err, "error while converting results to map")
+		err = errors.Wrap(err, "Error while converting  Nighthawk results to map")
 		logrus.Error(err)
 		return nil, nil, err
 	}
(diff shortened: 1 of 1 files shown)


helpers/load_test_interface.go
```diff
@@ -168,6 +169,90 @@ func WRK2LoadTest(opts *models.LoadTestOptions) (map[string]interface{}, *period
...
+		err = errors.Wrap(err, "error while unmarshaling data to the FortioHTTPRunner")
+		logrus.Error(err)
+		return nil, nil, err
+	}
+
+	resultsMap := map[string]interface{}{}
+	err = json.Unmarshal(bd, &resultsMap)
+
+	if err != nil {
+		err = errors.Wrap(err, "error while unmarshaling data to map")
```
leecalcote, line 247: Are these errors displayed as toaster notifications? If so, please cleanup.

kushthedude:
These are logs and displayed on stdout, this PR is for server side changes.

models/load-test.go
```diff
@@ -21,6 +21,9 @@ const (
 
 	// Wrk2LG - represents the wrk2 load generator
 	Wrk2LG LoadGenerator = "wrk2"
+
+	// NighthawkLG - represents the nighthawk load generator
+	NighthawkLG LoadGenerator = "nighthawk"
```
leecalcote, line 26: Performance preferences for this load generator are currently the same as for the other two load generators, right?

kushthedude: I didn't get this, but if it is about the preferences set up from settings then yes they will be same for each load generator.

helpers/load_test_interface.go
kushthedude: These are logs and displayed on stdout, this PR is for server side changes.

models/load-test.go
leecalcote: Roger. SGTM

PR #1466, opened 2020-08-26
enh: add timeout in golang-ci lint

kushthedude:


kushthedude:
> If after 5m we cancel this step, does that mean the build check will fail overall?

The flake which you encountered yesterday was because that there was no timeout & due to setting the cache it may had took longer & it failed. So it is just a precaution, hoping the flake won't occur again.

leecalcote:
Makes sense

PR #1520, opened 2020-09-03
Kumarabd/bug/smi chip

### LINKED ISSUE #1519: Modifications in smi conformance chip
**Description**
Modify the smi conformance chip name and remove notifications

**Expected Behavior**
The chip should say `View SMI Conformance results` and also the notifications must be removed.

kumarabd:
**Description**

This PR fixes #1519 

**Notes for Reviewers**


leecalcote (approved):
Thank you, @kumarabd. This is an improvement.

kushthedude:
Results are not persisted now, how would this behaviour be achieved then?

kumarabd:
> Results are not persisted now, how would this behaviour be achieved then?
> The concept of chip was to trigger the test and abstract away the functionality from the validate service mesh button

@kushthedude  we have another plan for the persistence here 🥇 . A cloud solution basically.

kushthedude:
Isn't it better to first implement the models and tables in the remote provider and then make changes in the UI?

leecalcote:
This is an intermediate step to be able to transiently tie together the results with the ability to view them in the UI before implementing a Validation menu item with persisted results similar to the Performance—>Results model.

kushthedude:
Why do display the dummy chip then?

leecalcote:
Much of this is driven by our ability to convince others of the project being “real”. Point in time decisions/code that need iterated as we go. For many project features, I have laid out extensive designs only to find that available contributors are incapable or unavailable to implement them. So, in general, we’re forced to take incremental steps to delivery. We need to inspire confidence in adopters along the way. This step is refactoring the intermediary UI chip to invoke a test and show results. We use UI chips as status indicators and connectivity testers today. We don’t want to break this UX paradigm with the user. The validation test is invocable from the Validation group of operations. Today, it’s these operations groups that are used to invoke mutable (and non-mutable) operations against a service mesh. 

We need to persist the conformance results in both the local and the Meshery remote providers. As an intermediary step to doing so, we’re changing the name from “run” to “view”, since users run the operation from the Validate operations group (box with plus sign). After we design tables for persistence, it’s likely that this UI chip will be displaced by a new Validation (or some similar name) as a menu item for holding Conformance results and Configuration best practice results.

PR #1562, opened 2020-09-08
Update main.go

kushthedude:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/cmd/mesheryctl/main.go
+++ b/mesheryctl/cmd/mesheryctl/main.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+
 package main
 
 import (


leecalcote:
Niiice. So long as this action completes before the others, we can use labeling in the build’s conditional logic. Does it make sense to incorporate PR Labeler into the builds managed by golangci?

kushthedude:
> Niiice. So long as this action completes before the others, we can use labeling in the build’s conditional logic. Does it make sense to incorporate PR Labeler into the builds managed by golangci?

Did you see the discrepancy above, it marked the PR with language/markdown label? Moreover, the runner gives out an error over here https://github.com/layer5io/meshery/pull/1505/checks?check_run_id=1085836565 . This needs to be observed for a few more PRs before getting incorporated to build pipelines.

PR #1571, opened 2020-09-10
Fetch SMI data

### LINKED ISSUE #1566: SMI chip integration with the test
**Description**
Parse the SMI results to the chip

dhruv0000:
Signed-off-by: Dhruv Patel <patel.4@iitj.ac.in>

**Description**
Not the final way of fetching the data, or the smi_result structure. Just a hack to get things going.
This PR fixes #1566

**Notes for Reviewers**

### DIFF
--- a/ui/components/MesheryResults.js
+++ b/ui/components/MesheryResults.js
@@ -25,6 +25,9 @@ const styles = (theme) => ({
   grid: {
     padding: theme.spacing(2),
   },
+  table: {
+    margin: theme.spacing(10),
+  },
   chartContent: {
     // minHeight: window.innerHeight * 0.7,
   },
@@ -49,6 +52,8 @@ class MesheryResults extends Component {
       // results_selection,
 
       selectedRowData: null,
+
+      smi_results: [],
     };
   }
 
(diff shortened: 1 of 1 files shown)


leecalcote:
Great to see progress on this project!

kumarabd:
Nice, lets test out the persistence flow tomorrow. Im seeing this happening by this weekend.

dhruv0000:
Currently, the PR getting our data from the event stream of the notifications and adding it to the redux-store. Would it be better to have a separate API from where we can fetch the SMI results data @kumarabd?

ui/components/MesheryNotification.js
```diff
@@ -122,6 +124,11 @@ class MesheryNotification extends React.Component {
       const { events } = this.state;
       const data = JSON.parse(e.data);
       events.push(data);
+      //Temperory Hack
+      if(data.summary==="Smi conformance test completed successfully"){
```
kushthedude, line 128: The response should be consumed from the trigger call, this is really not good 😕 .

What you can do is check for the SMI click in adapter play component and rather than having a notification stream for it, just display `Test completed` and stream the results into store

kumarabd: Yes you're right. Although this us just a draft. We are phasing out the changes. We'll be removing these eh the pr is ready to merge

kumarabd:
@dhruv0000 yes. I am drafting an endpoint with persistence now. Your can call those directly to fetch results.

ui/components/MesheryNotification.js
dhruv0000: Yeah. I know. This is not at all the way we should or are going to do it! Notification Component should not be involved in any way.

leecalcote: Resolving comment.

Signal when ready to test, @dhruv0000

PR #1583, opened 2020-09-13
Improve URL matching regex pattern.

### LINKED ISSUE #1528: The validation for protocol in URL donst allow urls without top level domain.
**Description**

The protocol wont consider `http://172.17.0.2:31329/productpage` as a vaild URL as it lacks a top level domain.
Rather it would accept `http://172.17.0.2:31329.com/productpage`
**Expected Behavior**

**Screenshots**

image

wryonik:
Signed-off-by: shubhamgupta2956 <shubham.gupta2956@gmail.com>

**Description**
The current URL matching regex fails for many cases including IP addresses, URLs without `www`, etc. This PR improves the URL matching regex.

**Note for reviewers**
Currently, I had put the regex pattern in the same file as I could not find any constants/helper files. Is there any file like this where we can put all the reusable things like this expression or it is ok to put it here?

This PR fixes #1528 

### DIFF
--- a/ui/components/MesheryPerformanceComponent.js
+++ b/ui/components/MesheryPerformanceComponent.js
@@ -79,6 +79,7 @@ const styles = (theme) => ({
     width: '100%',
   },
 });
+
 class MesheryPerformanceComponent extends React.Component {
   constructor(props) {
     super(props);
@@ -116,7 +117,7 @@ class MesheryPerformanceComponent extends React.Component {
 
   handleChange = (name) => (event) => {
     if (name === 'url' && event.target.value !== '') {
-      const compulsoryProtocolValidUrlPattern = new RegExp('^(https?:\\/\\/)' // compulsory protocol
+      const compulsoryProtocolValidUrlPattern = new RegExp('(^(http|https|nats|tcp):\\/\\/)' // compulsory protocol
       + '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' // domain name
       + '((\\d{1,3}\.){3}\\d{1,3}))' // OR ip (v4) address
       + '(\\:\\d+)?(\/[-a-z\\d%_.~+]*)*' // port and path


ui/components/MesheryPerformanceComponent.js
```diff
@@ -80,6 +80,13 @@ const styles = (theme) => ({
   },
 });
 
+const compulsoryProtocolValidUrlPattern = new RegExp('^(https?:\\/\\/)' // compulsory protocol
+  + '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' // domain name
+  + '((\\d{1,3}\.){3}\\d{1,3}))' // OR ip (v4) address
+  + '(\\:\\d+)?(\/[-a-z\\d%_.~+]*)*' // port and path
+  + '(\\?[;&a-z\\d%_.~+=-]*)?' // query string
+  + '(\\#[-a-z\\d_]*)?$', 'i'); // fragment locator
+
```
dhruv0000, line 89: Great work @shubhamgupta2956 . Can you define the RegExp inside the handle change function, since it is only used there? Would be easier for future contributors to read.

dhruv0000 (changes requested):
Looks good.

ui/components/MesheryPerformanceComponent.js
shubhamgupta2956: Sure @dhruv0000. But I had a suggestion. It may be possible that we may encounter other places in the future also where we may need this type of regex. Should we make a separate `constant.js/helper.js` file in which we can put these in one place? This will help when we are using the same function/constants in multiple files

dhruv0000: hmm. I don't think its needed for regex right now. But if ever needed, we have a lib folder for this kind of stuff.

dhruv0000 (dismissed):
Nice work @shubhamgupta2956

Neilblaze (dismissed):
Nice work @shubhamgupta2956 :100:

ui/components/MesheryPerformanceComponent.js
```diff
@@ -117,8 +116,15 @@ class MesheryPerformanceComponent extends React.Component {
 
   handleChange = (name) => (event) => {
     if (name === 'url' && event.target.value !== '') {
+      const compulsoryProtocolValidUrlPattern = new RegExp('^(https?:\\/\\/)' // compulsory protocol
```
leecalcote, line 119: This is an improvement. Thank you, @shubhamgupta2956. At some point, Meshery will support generating load for other protocols like gRPC, TCP and potentially NATS.

leecalcote: While we are updating this expression, let's go ahead and include the following protocols as valid:

`nats://`
`tcp://`

Do you mind making this update, @shubhamgupta2956?

shubhamgupta2956: Yeah sure. I will update the PR in a moment.

ui/components/MesheryPerformanceComponent.js
```diff
@@ -117,8 +117,15 @@ class MesheryPerformanceComponent extends React.Component {
 
   handleChange = (name) => (event) => {
     if (name === 'url' && event.target.value !== '') {
+      const compulsoryProtocolValidUrlPattern = new RegExp('(^(http|https|nats|tcp)?:\\/\\/)' // compulsory protocol
```
leecalcote, line 120: thanks, @shubhamgupta2956. Have a quick check on the use of `?`. It might produce a positive match in the event the user doesn't enter a protocol, like this: `://example.com/endpoint`

shubhamgupta2956: Yes, it is allowing this. I have updated the commit and removed the `?`.

leecalcote (approved):
Looks great, @shubhamgupta2956. Thank you. 👍

KarenEfereyan:
Amazing

PR #1601, opened 2020-09-22
feat: package nighthawk with meshery

### LINKED ISSUE #1599: Package nighthawk with meshery production bundle
**Current Behavior**

Nighthawk needs to build seperate from meshery

**Desired Behavior**

Package nighthawk with meshery production bundle

kushthedude:

**Description**

This PR fixes #1599

**Notes for Reviewers**

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -60,6 +60,8 @@ services:
       - "com.centurylinklabs.watchtower.enable=true"
     ports:
       - "10009:10009"
+  nighthawk-lg:
+    image: envoyproxy/nighthawk-dev:latest
   watchtower:
     image: v2tec/watchtower
     labels:


docker-compose.yaml
```diff
@@ -60,6 +60,8 @@ services:
       - "com.centurylinklabs.watchtower.enable=true"
     ports:
       - "10009:10009"
+  nighthawk-lg:
+    image: envoyproxy/nighthawk-dev:latest
```
leecalcote, line 64: Curious, what's the size of this image?

leecalcote: Use of the `-dev` is temporary, until when?

After `-dev` is gone, what is our strategy for pinning to a specific version? This question is the same for other generators.

kushthedude: 42 mb

kushthedude: > Use of the -dev is temporary, until when?

Until nighthawk makes its v1.0 debut.

> After -dev is gone, what is our strategy for pinning to a specific version? This question is the same for other generators.

This may get addressed when we deduce the strategy of versioning the adapters and pinning. Most probably, keep up with the latest as Load Generators may only get better with their new versions with very few breaking changes but this may not be same when we address the adapters.

leecalcote (approved):
Sounds good!

kushthedude:
As expected, mesheryctl build has errored out. @leecalcote Please refrain from making any stable releases right now

Issue #1651, opened 2020-10-05
[mesheryctl] mesh validate command to support non local meshery server deployments


anirudhjain75:
**Description**

Handling of non-local Meshery server deployments by way of supporting a user-provided URL to Meshery server (not hardcoding localhost).

**Expected Behavior**

Ability to work with custom server URL for mesh validate

PR #1652, opened 2020-10-05
[mesheryctl] Fix for #1651: Allow users to pass in custom meshery server url instead of using the hardcoded localhost

anirudhjain75:


anirudhjain75:
@leecalcote one of the two suggestion you had on improving the cmd

anirudhjain75:
The Issue page is  #1651

leecalcote (commented):
This is a good flag in concept. Ultimately, mesheryctl needs a global flag and/or a global configuration for the CLI.

We have a global configuration file option for `mesheryctlconfig` today, which is an ideal way to address this holistically.

This global configuration file is an incomplete implementation in that we need to define what the actual structure of this config file looks like.

Can we direct our efforts into this global view before implementing a flag for a specific command?

leecalcote:
Reference on `mesheryctlconfig` - https://github.com/layer5io/meshery/issues/1083

PR #1684, opened 2020-10-16
Add adapter flag to meshery start

sladyn98:
**Description**

This pull request adds the adapter flag to the `mesheryctl system start ` command.
Use Case: 
[ ] I want to just choose one of the services and I would not like all of them to startup.
[ ] Ability to spin up maybe one or two and not all. 

Things to consider:
a) Transient start to meshery for a certain set of adapters
b) Persistent start to meshery where the user configures a set of adapters using a yaml file. 
 
We could have a set of profiles such as default / a more persistent write to yaml files. 

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/internal/cli/root/system/start.go
+++ b/mesheryctl/internal/cli/root/system/start.go
@@ -34,6 +34,7 @@ import (
 
 var (
 	skipUpdateFlag bool
+	adapter string
 )
 
 // startCmd represents the start command
@@ -87,7 +88,11 @@ func start() error {
 	}
 
 	log.Info("Starting Meshery...")
+	log.Info("Adapter chosen: " + adapter)
 	start := exec.Command("docker-compose", "-f", utils.DockerComposeFile, "up", "-d")
+	if adapter != "" {
+		start = exec.Command("docker-compose", "-f", utils.DockerComposeFile, "up", "-d", "meshery", adapter)
+	}
 	start.Stdout = os.Stdout
 	start.Stderr = os.Stderr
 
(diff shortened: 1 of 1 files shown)


sladyn98:
@anirudhjain75 @NupurThakur27 One thing I noticed the update Containers function is called before the service can start, so I guess I would need to pass the adapter as a parameter there so that the appropriate containers are being updated.

leecalcote:
@clementkng, would you like to review?

leecalcote:
@sladyn98 we may need some discussion on this design. Starting Meshery and a select set of adapter using this method will have a transient effect. Onto its own, that’s not a bad thing. But, we should design this functionality in coordination with the same behavior that will be delivered in a persistent way.

sladyn98:
> @sladyn98 we may need some discussion on this design. Starting Meshery and a select set of adapters using this method will have a transient effect. Onto its own, that’s not a bad thing. But, we should design this functionality in coordination with the same behavior that will be delivered in a persistent way.

Agreed let me know the feedback on this @leecalcote

leecalcote (commented):
@sladyn98 this is most helpful. Can you move this to `mesheryctl exp system start`? We can move forward with a merge there to start.

leecalcote:
@sladyn98, we finally have requirements written up for this functional area. Would you like to take a crack at #2105?

anirudhjain75:
Hey @sladyn98 , thanks for the ton of work here. We are planning to take on this in the future once we move on to user experience and final touches for meshery with context part. There are a lot of changes introduced to mesheryctl since this PR. The idea of context simply allows users to configure the adapters the user required to boot up with. Also feel free to test out `mesheryctl system context`. Let me know if you have any comments or face any issue.
Thanks :)

leecalcote:
@sladyn98 come jump into `mesheryctl system context` - - your help is much needed: https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#

PR #1691, opened 2020-10-17
fix(UI): table filters not rendering due to custom header

kushthedude:

### DIFF
--- a/ui/components/MesheryResults.js
+++ b/ui/components/MesheryResults.js
@@ -25,42 +25,22 @@ const styles = (theme) => ({
   grid: {
     padding: theme.spacing(2),
   },
-  tableHeader: {
-    fontWeight: 'bolder',
-    fontSize: 18,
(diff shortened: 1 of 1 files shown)


kushthedude:
@dhruv0000 This PR mostly reverts the changes made in the Table Tweaks, alas but seems that custom header components are causing issues with the sort filters. So I suggest you re-open the issue concerning the tweaks you had made in the results table.

dhruv0000:
Hmm. The custom headers were used to make the fonts bolder. Let me look into it.

kushthedude:
> Hmm. The custom headers were used to make the fonts bolder. Let me look into it.

If that was all the function the custom renders were to provide, then we are good without them ig.

dhruv0000:
#1694 Should fix the sorting option.

kushthedude:
Duplicate of #1694

PR #1715, opened 2020-10-21
Add experimental subcommand to mesheryctl

anirudhjain75:


mesheryctl/internal/cli/root/mesh/validate.go
```diff
@@ -102,11 +102,11 @@ var validateCmd = &cobra.Command{
...
-	_ = validateCmd.MarkFlagRequired("adapter")
-	validateCmd.Flags().StringVarP(&namespace, "namespace", "n", "default", "Kubernetes namespace to be used for deploying the validation tests and sample workload")
-	validateCmd.Flags().StringVarP(&tokenPath, "tokenPath", "t", "", "Path to token for authenticating to Meshery API")
-	_ = validateCmd.MarkFlagRequired("tokenPath")
+	ValidateCmd.Flags().StringVarP(&spec, "spec", "s", "smi", "specification to be used for conformance test")
+	_ = ValidateCmd.MarkFlagRequired("spec")
+	ValidateCmd.Flags().StringVarP(&adapterURL, "adapter", "a", "meshery-osm:10010", "Adapter to use for validation")
+	_ = ValidateCmd.MarkFlagRequired("adapter")
+	ValidateCmd.Flags().StringVarP(&namespace, "namespace", "n", "default", "Kubernetes namespace to be used for deploying the validation tests and sample workload")
+	ValidateCmd.Flags().StringVarP(&tokenPath, "tokenPath", "t", "", "Path to token for authenticating to Meshery API")
```
leecalcote, line 110: Will this support for `--token` and `-t`?

mesheryctl/internal/cli/root/root.go
```diff
@@ -110,7 +110,7 @@ func init() {
 		versionCmd,
 		system.SystemCmd,
 		perf.PerfCmd,
-		mesh.MeshCmd,
```
leecalcote, line 113: Is this removing `mesheryctl mesh` and replacing it with `exp`? Ideally, any new command to be exposed under `exp` will retain its intended syntax structure; e.g. `mesheryctl exp mesh validate`.

anirudhjain75: Hmm Okay

mesheryctl/internal/cli/root/mesh/validate.go
anirudhjain75: Yea It would

leecalcote, line 110: Use of `utils.AuthConfigFile` (example here - https://github.com/layer5io/meshery/blob/49819bd9438140cc7ad5572716c1ae2c158e511c/mesheryctl/internal/cli/root/perf/perf.go#L161) seems to provide CLI users with the convenience of assuming a default location for the token file. Is that something we might incorporate here? If so, consistency of UX is ideal.

anirudhjain75: Yea I think so

anirudhjain75: Although the change for this would be done in another PR where I've made changes

mesheryctl/internal/cli/root/root.go
anirudhjain75: No actually this was adding the entire mesh subcmd inside exp

mesheryctl/internal/cli/root/mesh/validate.go
leecalcote: Ok, sounds good.

mesheryctl/internal/cli/root/root.go
leecalcote: Nice. ✔️

leecalcote (approved):
A couple lint checks to cleanup, but LGTM. 👍

anirudhjain75:
Done

anirudhjain75:
I've kept the original structure as well as added to exp subcmd.

leecalcote (approved):
Wham, bam. Done!

leecalcote:
Good stuff.

PR #1721, opened 2020-10-21
fix(UI): Persist Grafana charts after a Reloading

dhruv0000:
**Description**

This PR fixes #

**Notes for Reviewers**


ui/lib/store.js
```diff
@@ -46,12 +46,12 @@ const initialState = fromJS({
     grafanaBoardSearch: '',
     grafanaBoards: [],
     selectedBoardsConfigs: [],
-    ts: new Date(),
+    ts: new Date(-8640000000000000),
```
leecalcote, line 49: @dhruv0000 what is the logic here? Is this to invalidate any cached info?

dhruv0000: To not let the initialized empty data be seen as the latest data.

leecalcote (approved):
Demo in today's Dev call, if you would.

PR #1761, opened 2020-10-25
Add latest available version in the API and in the dashboard

### LINKED ISSUE #1752: [UI] Add Meshery Version to Release card on System Dashboard
### Current Behavior
Meshery has a System Dashboard showing a disabled set of radio buttons to change between release channels, but does not show the running version of Meshery.

### Desired Behavior
Users should be shown their currently running version. Latest available version can be addressed by another issue/PR.

### Implementation
1. Retrieve Meshery server version from http://<meshery-server>/api/server/version
1. Add "Channel" and "Version" headings in the **Release** card.
1. Display Meshery server version

### Acceptance Tests
1. If Meshery version is unretrievable, show an empty value or "Unknown".

### Mockups
Frame 1-2

### LINKED ISSUE #1753: [UI] Add Latest Available Version to Release card on System Dashboard
**Dependency:** #1752 should be implemented first or should be implemented in combination with this issue.

### Current Behavior
Meshery has a System Dashboard showing a disabled set of radio buttons to change between release channels, the currently running version of Meshery, but does not show the user the latest available version.

### Desired Behavior
Users should be shown the latest available version. 

### Implementation
_Meshery Server_
1. Augment the existing endpoint `/api/server/version` with new functionality to retrieve the latest available release from GitHub API
1. Use `github.com/tcnksm/go-latest` library to retrieve latest release. Reference [implementation in `mesheryctl`](https://github.com/layer5io/meshery/blob/master/mesheryctl/internal/cli/root/version.go#L50-L73)

_Meshery UI_
1. Retrieve Meshery server version and latest available version from http://<meshery-server>/api/server/version
1. Display Latest Available version as shown in mockup.

### Acceptance Tests
1. If the user is running the latest available version, show text "Meshery is up to date".
1. If Meshery version is unretrievable, show an empty value or "Unknown".

### Mockups
Frame 1-2

tangledbytes:
**Description**
Latest available version was added to the the server spec handler response and the same was added to the dashboard component.

***Relevant screenshots***

Running latest version:

meshery-latest

Running outdated version:

meshery-outdated

Version fetch error:

meshery-failed

This PR fixes #1752 and fixes #1753 

**Notes for Reviewers**
This is my first pull request please let me know if everything is alright.


ui/components/DashboardComponent.js
```diff
@@ -1,597 +1,658 @@
...
+    );
+
+    /**
+     * getMesheryVersionText returs a well formatted version text
+     * @param {string} type type of version could be "latest" or "current"
+     */
+    const getMesheryVersionText = (type) => {
+      const {build, latest, outdated} = this.state.versionDetail
+      if (type === "current") {
+        if (outdated) return `Running Meshery: stable-${build}`
```
Nikhil-Ladha, line 565: If the current Meshery version is an edge release then we should display that too, right?

Nikhil-Ladha:
Also, I guess some golang test is failing can you run the golang lint command in local and then check that?

ui/components/DashboardComponent.js
leecalcote: @Nikhil-Ladha, good call, yes. @utkarsh-pro, fyi, an Edge release will be in a similar format (e.g. `edge-latest`).

utkarsh-pro: Can you please tell me that how an edge release and a stable release is determined? Will it depend on the selected channel?

leecalcote: @utkarsh-pro, yes, that's right. Although, currently, the ability to switch between `stable` and `edge` releases is disabled in the UI and must be done separately. We need a doc on this.

Briefly, you can edit your `~/.meshery/meshery.yaml` and on the fourth line update it to `edge-latest`. Then `mesheryctl system restart`.

leecalcote: @kumarabd will you speak to how to use environment variables to manipulate stable, edge, version numbers when compiling meshery locally?

utkarsh-pro: Should I convert Version struct to this?
```diff
type Version struct {
    Build     string `json:"build,omitempty"`
    Latest   string `json:"latest,omitempty"`
    Outdated  *bool  `json:"outdated,omitempty"`
    CommitSHA string `json:"commitsha,omitempty"`
+   Release string `json:"release,omitempty"
}
```

But I have no idea from where we I can get this `release` info on the server.

utkarsh-pro:
> Also, I guess some golang test is failing can you run the golang lint command in local and then check that?
> The command can be found in the makefile.

I have fixed that particular issue now.

ui/components/DashboardComponent.js
leecalcote: @kushthedude, can you highlight what changes will need to be done in the build system in order to provide release channel name "stable" or "edge" as an environment variable, so that the `/api/server/version` endpoint can be enhanced to include the channel name?

- Dockerfile: https://github.com/layer5io/meshery/blob/3abaa633b14e331187396f32d36c7d5f1f55b989/Dockerfile#L3-L4
- GitHub action: https://github.com/layer5io/meshery/blob/3abaa633b14e331187396f32d36c7d5f1f55b989/.github/workflows/build-and-release.yml#L63

ui/components/DashboardComponent.js
```diff
@@ -1,597 +1,658 @@
...
-    k8sconfig,
-    grafana,
-    prometheus,
-  };
-};
-
-export default withStyles(styles)(
-  connect(mapStateToProps, mapDispatchToProps)(withRouter(withSnackbar(DashboardComponent)))
-);
+import React from "react";
```
kushthedude, line 1: Can you fix this, why is there change at everyline?

handlers/server_spec_handler.go
```diff
@@ -23,14 +28,44 @@ func (h *Handler) ServerVersionHandler(w http.ResponseWriter, r *http.Request) {
...
+// checkLatestVersion takes in the current server version compares it with the target
+// and returns the result (latest.CheckResponse)
+func (h *Handler) checkLatestVersion(serverVersion string) (*latest.CheckResponse, error) {
+	githubTag := &latest.GithubTag{
+		Owner:      mesheryGitHubOrg,
+		Repository: mesheryGitHubRepo,
+	}
+
+	// Compare current running Meshery server version to the latest available Meshery release on GitHub.
+	res, err := latest.Check(githubTag, serverVersion)
```
kushthedude, line 62: Not needed, we already have a server version endpoint, please use that.

kushthedude (changes requested):

ui/components/DashboardComponent.js
utkarsh-pro: Actually I couldn't understand this behavior either as when I see diff in my system it correctly points out the changes while here it doesn't. I have tried running `eslint --fix` as well as have editorconfig plugin installed. Can you please suggest a solution for the same?

handlers/server_spec_handler.go
utkarsh-pro: I searched through the repository and I could only find something related to latest version checking in the mershyctl package as it was pointed out by leecalcote in the description of issue #1753 . Can you please point out the endpoint, actually I am still exploring the repository :smile: .

ui/components/DashboardComponent.js
kumarabd: > @kumarabd will you speak to how to use environment variables to manipulate stable, edge, version numbers when compiling meshery locally?

@leecalcote It is controlled by the ld-flags during compile time of the Golang backend server. The flags are available in here - https://github.com/layer5io/meshery/blob/master/Dockerfile#L9. If we have to build multiple images with the different version tags on each build, one way of doing it is to pass docker build argument in the CI itself (Thus not having dependency on the project). This can be done by passing these arguments - https://github.com/layer5io/meshery/blob/master/Dockerfile#L2-L4

handlers/server_spec_handler.go
leecalcote: @kushthedude, the task at-hand here is to take the currently running server version from the `/api/server/version` endpoint, retrieve the version number of the latest release from GitHub and compare them. This functionality was recently incorporated into `mesheryctl version`. It would be nice to let users of the UI know if they are a version behind.

ui/components/DashboardComponent.js
leecalcote: @utkarsh-pro, yes, adding a new attribute to the Version struct makes sense to me. "Channel" is an ideal name for the attribute.

handlers/vars.go
```diff
@@ -1,5 +1,7 @@
 package handlers
 
 const (
-	cookieSuffix = "_referrer"
+	cookieSuffix      = "_referrer"
+	mesheryGitHubOrg  = "layer5io"
```
leecalcote, line 5: Hard to believe that we didn't have these defined already. 🙄 Wow.

handlers/server_spec_handler.go
kushthedude: Then we can improve the version endpoint to show this, rather than having additional code in both mesheryctl and UI to do this task.

ui/components/DashboardComponent.js
kushthedude: Try correcting your editor configuration, this has been done by your editor/IDE. This makes it very difficult to review as we don't know which lines have been added or delete, or amended.

utkarsh-pro: I have fixed the diff issue. The problem was apparently in .editorconfig's `end_of_line: crlf`. I disabled editorconfig plugin on my text editor(vscodium) and that kind of fixed the issue for me. I assume it could be due to the way windows handles end of lines (CRLF) and the way linux handles it (LF).

ui/components/DashboardComponent.js
```diff
@@ -526,17 +555,49 @@ class DashboardComponent extends React.Component {
       </div>
     );
 
+    /**
+     * getMesheryVersionText returs a well formatted version text
+     * @param {string} type type of version could be "latest" or "current"
+     */
+    const getMesheryVersionText = (type) => {
```
kushthedude, line 562: Why is this function needed? All of this can be handled in showRelease itself.
We need to show `running meshery` for both cases, and for the next line use
 ```js
{outdated ? 'Latest available' : 'up-to-date'}
```

kushthedude: Any way we can handle this later if you want to keep this PR simple. LGTM! 🙂

leecalcote (approved):
Thank you, @utkarsh-pro. 💪

PR #1767, opened 2020-10-27
fix: remove the duplicate function required for version checking

kushthedude:


leecalcote (approved):
LGTM. The CLI command still executes properly even when Meshery server is not available.

Issue #1037, opened 2020-05-20
[CI] Addition of tests for Meshery UI (Umbrella issue)


kushthedude:
**Description**

We have already integrated the Cypress testing framework. Let's start adding more & more tests for our UI.

#### Integration Tests(Routes):

- [x] ~`provider` (#1778)~
- [ ] `/management`
- [ ] `/performance`
- [x] ~`/userpreference` (#1853)~ 
- [ ] `/smi_results`
- [ ] `/results`
- [ ] `/settings` (Environment (#2316), Service Meshes (#2317), Metrics, Performance)
- [ ] `/index` (#2315)
- [ ] `/configuration` (#4582)

#### E2E Tests(Routes):
- [ ] `/management`
- [ ] `/performance`
- [x] ~`/userpreference` (#1862)~
- [ ] `/smi_results`
- [ ] `/results`
- [ ] `/settings` (~Environment (#2139)~, ~Service Meshes (#2218)~, Metrics, Performance)
- [ ] `/index`
- [ ] - [ ] `/configuration` (#4582)
- [ ] Add integration tests for /configuration/applications route
- [ ] Add integration tests for /configuration/filters route
- [ ] Add integration tests for /configuration/patterns route
**E2E Tests (these are more usecase or scenario wise as they'll touch multiple routes and multiple pages in each child issue with actual server responses):**
- [ ] Add E2E tests to Deploy Applications/Filters/Patterns Configuration through Meshery UI (#4826)
- [ ] Add E2E tests to Edit Applications/Filters/Patterns Configuration through Meshery UI
- [ ] Add E2E tests to Delete Applications/Filters/Patterns Configuration through Meshery UI"

**Additional Context**

tsingh2k15:
@kushthedude I can contribute. I will take a look and share the component which I can cover.

kushthedude:
> @kushthedude I can contribute. I will take a look and share the component which I can cover.

Great! AFAIK provider-ui can work without server, performance page, settings page & results page can render if the server is not working.

ramrodo:
@kushthedude I want to contribute to creating tests for the `provider-ui`. Should I create an issue for this specific route or how can we coordinate this?

kushthedude:
> @kushthedude I want to contribute to creating tests for the `provider-ui`. Should I create an issue for this specific route or how can we coordinate this?

Hi @ramrodo , Provider-UI only has single route so I guess we don't need to create a different issue and you can just refer your PR with this issue 😄

[...]

PR #1853, opened 2020-11-02
[UI] Add test integration for /userpreference

ramrodo:
**Description**

This change:
- Upgrades Cypress to `5.5.0` for stub fetch requests
- Adds integration tests for `/userpreference`
- Updates integration test for `provider-ui`
- Fixes a typo in the confirmation message for User Preferences

This PR is related to #1037

**Notes for Reviewers**

### DIFF
--- a/ui/cypress/fixtures/stats.json
+++ b/ui/cypress/fixtures/stats.json
@@ -1,0 +1,5 @@
+{
+  "anonymousUsageStats":true,
+  "anonymousPerfResults":true,
+  "updated_at":""
+}
--- a/ui/components/UserPreference.js
+++ b/ui/components/UserPreference.js
@@ -83,12 +83,12 @@ class UserPreference extends React.Component {
     if (name == 'anonymousUsageStats') {
       val = anonymousStats;
       msg = !val ? "Sending anonymous usage statistics was enabled"
-        : "Sending anonymous usage statistics was diabled";
+        : "Sending anonymous usage statistics was disabled";
 
     } else {
       val = perfResultStats;
       msg = !val ? "Sending anonymous performance results was enabled"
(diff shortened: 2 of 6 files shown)


ui/cypress/fixtures/stats.json
```diff
@@ -0,0 +1,5 @@
+{
+  "anonymousUsageStats":true,
```
kushthedude, line 2: How about using mirage for this?

ramrodo: I didn't know about MirageJS. I did a research and I found that there are two popular API mocking tools:
- Mirage JS
- Mock Service Worker (msw)

Here is a comparison of both: https://miragejs.com/docs/comparison-with-other-tools/#msw

And after reading different opinions about these two...

_"the biggest difference today is that MSW runs in a Service Worker which enables real-enough network requests that show up in the Network tab of the Devtools"_,

**but**

_"in Mirage JS, the in-memory database + ORM help maintain your data's relational integrity across reads and writes to all endpoints"_

(extracted from [here](https://github.com/miragejs/miragejs/issues/449#issuecomment-622519413))

So, I'm not sure which one could be the right one for the meshery project. What do you think?

kushthedude: I guess we will move to MirageJS, but considering we have seldom of UI tests it will be worth to move away when Cypress fixes becomes redundant and less efficient. For now, we can use the cypress stubs & fixes as is you have implemented in this PR. 

PS: That's quite neat and beautiful work you have done here 😄

kushthedude (approved):
Great work here with tests, stubs & fixes @ramrodo, feels good to have server independent tests for Meshery 🎉 !

PR #1858, opened 2020-11-04
Add mesh names derived from SMP spec

tangledbytes:
**Description**
This PR has changes both in the server and in the UI.

Changes in the server:
1. Added an endpoint to retrieve service meshes names from SMP.  

Changes in the UI:
1. Replaced the static list with the call to the above mentioned endpoint.
2. Render the sorted list of the retrieved meshes.

This PR fixes layer5io/service-mesh-performance#150

**Notes for Reviewers**

### DIFF
--- a/router/server.go
+++ b/router/server.go
@@ -61,7 +61,7 @@ func NewRouter(ctx context.Context, h models.HandlerInterface, port int) *Router
 		Methods("GET")
 	gMux.Handle("/api/perf/result", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.GetResultHandler)))).
 		Methods("GET")
-	gMux.Handle("/api/perf/mesh/list", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.GetSMPServiceMeshes)))).
+	gMux.Handle("/api/mesh", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.GetSMPServiceMeshes)))).
 		Methods("GET")
 
 	gMux.Handle("/api/smi/results", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.FetchSmiResultsHandler)))).
--- a/ui/components/MesheryPerformanceComponent.js
+++ b/ui/components/MesheryPerformanceComponent.js
@@ -377,7 +377,7 @@ class MesheryPerformanceComponent extends React.Component {
 
   getSMPMeshes = () => {
     const self = this
-    dataFetch('/api/perf/mesh/list', {
+    dataFetch('/api/mesh', {
       credentials: 'same-origin',
       credentials: 'include',
       method: 'GET'


ui/components/MesheryPerformanceComponent.js
```diff
@@ -391,6 +375,21 @@ class MesheryPerformanceComponent extends React.Component {
     }, () => {});
   }
 
+  getSMPMeshes = () => {
+    const self = this
+    dataFetch('/api/perf/meshes', {
+      credentials: 'same-origin',
```
kushthedude, line 381: Specify method here

handlers/load_test_meshes_handler.go
```diff
@@ -0,0 +1,35 @@
...
+		AvailableMeshes: make([]string, 0, len(SMP.ServiceMeshes_Type_name)),
+	}
+
+	for _, v := range SMP.ServiceMeshes_Type_name {
+		if v != SMP.ServiceMeshes_INVALID_MESH.String() {
+			meshes.AvailableMeshes = append(meshes.AvailableMeshes, strings.Title(strings.ToLower(strings.ReplaceAll(v, "_", " "))))
+		}
+	}
+
+	w.Header().Set("Content-Type", "application/json")
```
kushthedude, line 30: Not required

handlers/load_test_meshes_handler.go
```diff
@@ -0,0 +1,35 @@
...
+
+	if err := json.NewEncoder(w).Encode(meshes); err != nil {
+		logrus.Errorf("unable to send data: %v", err)
```
kushthedude, line 33:
```suggestion
		logrus.Errorf("error encoding meshList object: %v", err)
		http.Error(w, "Error encoding meshList object", http.StatusInternalServerError)
		return
```

kushthedude (changes requested):
nice work @utkarsh-pro , some minor changes and better error handling need to be done

ui/components/MesheryPerformanceComponent.js
utkarsh-pro: Just to clarify, should I covert the dataFetch function options to:
```diff
credentials: 'same-origin',
credentials: 'include',
+ method: 'GET'
```

kushthedude: Yes

router/server.go
```diff
@@ -61,6 +61,8 @@ func NewRouter(ctx context.Context, h models.HandlerInterface, port int) *Router
 		Methods("GET")
 	gMux.Handle("/api/perf/result", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.GetResultHandler)))).
 		Methods("GET")
+	gMux.Handle("/api/perf/meshes", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.GetSMPServiceMeshes)))).
```
kushthedude, line 64: Can we rename it to something more descriptive `api/perf/mesh/list`?

utkarsh-pro: Oh yeah, that makes more sense. Thanks!

handlers/load_test_meshes_handler.go
```diff
@@ -0,0 +1,35 @@
...
+// SMPMeshes defines the JSON payload structure for available meshes api
+type SMPMeshes struct {
+	AvailableMeshes []string `json:"available_meshes,omitempty"`
+}
+
+// GetSMPServiceMeshes handles the available meshes request
+func (h *Handler) GetSMPServiceMeshes(w http.ResponseWriter, r *http.Request, prefObj *models.Preference, user *models.User, provider models.Provider) {
+	meshes := SMPMeshes{
+		AvailableMeshes: make([]string, 0, len(SMP.ServiceMeshes_Type_name)),
+	}
```
kushthedude, line 22: Why is this logic needed? Can't we use 
```suggestion
	var meshes SMPMeshes
```

kushthedude (approved):
Nice work, everything looks in place. do look at the last review comment else LGTM

handlers/load_test_meshes_handler.go
utkarsh-pro: This was just to set initial capacity of the slice or else while appending the elements to the slice, the capacity would double (albeit till a certain threshold) which could make it a bit slower. Here we already have the information of the maximum capacity for the slice that we may need so I just utilized it here :sweat_smile:

kushthedude: This is a very minor optimization which makes code rather difficult to read. But still, let's move ahead with this.

router/server.go
```diff
@@ -61,6 +61,8 @@ func NewRouter(ctx context.Context, h models.HandlerInterface, port int) *Router
 		Methods("GET")
 	gMux.Handle("/api/perf/result", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.GetResultHandler)))).
 		Methods("GET")
+	gMux.Handle("/api/perf/mesh/list", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.GetSMPServiceMeshes)))).
```
leecalcote, line 64: Let's give holistic consideration to the endpoints that we are exposing. A list of possible service meshes is interesting under a number of uses, not strictly in context of performance.

We need to get together of set of API docs (@anirudhjain75 or @chandrashritii  do you have this?) - or simply a set of endpoints into a categorized list.

utkarsh-pro: @leecalcote does `/api/mesh/list` look better?

leecalcote: @utkarsh-pro, yes, or simply a `/api/mesh/` a GET on this resource should return this list. Later, it should also return information on installed and actively running service meshes.

leecalcote (approved):
Looking good, @utkarsh-pro

kushthedude (approved):
nice work here @utkarsh-pro

PR #1862, opened 2020-11-05
[UI] Add e2e test for /userpreference

ramrodo:
**Description**

This change:
- Adds e2e test for `/userpreference`
- Updates `provider-ui` integration tests to follow best practices (changes attribute to `data-cy` & replaces the URL to visit)
- Changes Cypress config to not store videos by default (in order to not generate them in CI)
- Configures e2e tests to be run in Github Actions
- Configures cache for `node_modules` and dependencies related to 

This PR is related to #1037

**Notes for Reviewers**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,10 +101,10 @@ jobs:
     - name: Install dependencies
       run: |
         cd ui
-        npm ci
+        npm i
         cd ..
         cd provider-ui
-        npm ci
+        npm i
   tests-ui:
     needs: [build-ui]
     name: UI integration tests
@@ -131,10 +131,10 @@ jobs:
       if: steps.node-cache.outputs.cache-hit != 'true'
       run: |
         cd ui
-        npm ci
+        npm i
         cd ..
         cd provider-ui
-        npm ci
+        npm i
     - name: Run tests
       run: |
         cd ui
@@ -180,10 +180,10 @@ jobs:
         if: steps.node-cache.outputs.cache-hit != 'true'
         run: |
           cd ui
-          npm ci
+          npm i
           cd ..
           cd provider-ui
-          npm ci
+          npm i
       - name: Build Next App
         run: cd ui && npm run build && npm run export
       - name: Run backend


ramrodo:
Currently, I'm having problems with caching `node_modules`. But, I'm working on solving this.

ramrodo:
After several commits and reading about GitHub Actions and Cypress cache, the tests passed!

.github/workflows/ci.yml
```diff
@@ -106,21 +117,90 @@ jobs:
...
+          cd ..
+          cd provider-ui
+          npm ci
+      - name: Build Next App
+        run: cd ui && npm run build && npm run export
+      - name: Run backend
+        env:
+          SAAS_BASE_URL: http://localhost:9876
+          PORT: 9081
+          ADAPTER_URLS: "mesherylocal.layer5.io:10000 mesherylocal.layer5.io:10001 mesherylocal.layer5.io:10002 mesherylocal.layer5.io:10003 mesherylocal.layer5.io:10004 mesherylocal.layer5.io:10008 mesherylocal.layer5.io:10009"
```
leecalcote, line 193: I wonder if there's an alternative way of loading these from the `Makefile` or `docker-compose.yaml` such that this list doesn't need to be maintained (new adapters are not added frequently, but are added occasionally).

If not, then, we may want to expand this list to include ports 10007 (kuma) - it's ready for testing. Almost ready to move from beta to stable.

ramrodo: Ok, I see. Let me see if we could extract them from those files. If not, I'll add 10007 as you mention.

ui/cypress/integration/e2e/userpreference_spec.js
```diff
@@ -0,0 +1,45 @@
...
+      cy.wait('@postUserStats')
+      cy.get('.MuiSnackbarContent-message').should('have.text', 'Sending anonymous usage statistics was disabled')
+    })
+
+    it('activates "Send Anonymous Usage Statistics"', () => {
+      cy.route2('POST', '/api/user/stats').as('postUserStats')
+
+      cy.get('fieldset > div > label:nth-child(1)').click()
+      cy.wait('@postUserStats')
+      cy.get('.MuiSnackbarContent-message').should('have.text', 'Sending anonymous usage statistics was enabled')
```
leecalcote, line 26: @ramrodo looking good. Two thoughts:

1) The text used in these messages might be updated over time. Is there any other approach to validating the test passed or failed? If there is no less fragile way to design this test, then we'll just be sure to update these tests as we go.

2) We have overused (abused) snackbars in Meshery's current user experience. Hopefully, soon the UX will improve to reduce the noise here and start to show other visual indicators of a switch being set to on or off.

leecalcote:
Haha! No stinking cache for node_modules will stand in your way!

ui/cypress/integration/e2e/userpreference_spec.js
ramrodo: Yes, there is another way to validate if that test pass. We can check if the switch button is disabled or if the CSS style changed. Let me update it and remove that message validation. It's challenging to maintain these successful messages as you mention.

ramrodo:
> Haha! No stinking cache for node_modules will stand in your way!

Hahaha right! It will very useful in the future when we add more and more tests

.github/workflows/ci.yml
```diff
@@ -70,6 +70,11 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-go-        
     - run: GOPROXY=direct GOSUMDB=off GO111MODULE=on go build -tags draft ./cmd/main.go
+    - name: Upload artifacts
```
kushthedude, line 73: I don't think this is needed 🤔

ui/cypress/support/commands.js
```diff
@@ -1,25 +1,6 @@
...
-//
-//
-// -- This is a dual command --
-// Cypress.Commands.add("dismiss", { prevSubject: 'optional'}, (subject, options) => { ... })
-//
-//
-// -- This will overwrite an existing command --
-// Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+Cypress.Commands.add('selectProviderNone', () => {
+  cy.request({
```
kushthedude, line 2: Nice ⭐

ui/cypress.json
```diff
@@ -1,4 +1,5 @@
 {
   "baseUrl": "http://localhost:3000",
-  "experimentalNetworkStubbing": true
+  "experimentalNetworkStubbing": true,
+  "video": false
```
kushthedude, line 4: Let's have it to true, we can see the test video. Moreover, we can apply for cypress dashboard too as we are open-source 😎

.github/workflows/ci.yml
kushthedude: You can just use `make run-local` over here, which will save us from adding artifacts action.

kushthedude (changes requested):
Nice work @ramrodo 👍 , some minor neats if you will look into them

ui/cypress.json
ramrodo: Great! Have the Cypress Dashboard, sounds really good

.github/workflows/ci.yml
ramrodo: I didn't use `make run-local` because that command [builds the _meshery_ binary too](https://github.com/layer5io/meshery/blob/master/Makefile#L63); and we already building _meshery_ in a previous step ([build-backend](https://github.com/layer5io/meshery/blob/master/.github/workflows/ci.yml#L72)). Using `artifacts` - _a way to share files between jobs_ -  will allow us to not build _meshery_ twice. If we rebuild the _meshery_ again, it will take **50s** instead of **1s** downloading the artifact. You can see that in [this Github Action](https://github.com/layer5io/meshery/pull/1862/checks?check_run_id=1361845268). This is like using caching node dependencies to avoid install them twice.

.github/workflows/ci.yml
ramrodo: I explained the reasons to use it in [this comment below](https://github.com/layer5io/meshery/pull/1862/files#r518824113)

.github/workflows/ci.yml
ramrodo: I added the port directly _(10007)_ because actually, we don't have tests related to test those Adapters yet. I think that if there is a new Adapter, it'll be included as part of writting the test. What do you' all think?

.github/workflows/ci.yml
```diff
@@ -81,22 +86,28 @@ jobs:
...
+        key: ${{ runner.os }}-node-${{ hashFiles('**/lockfiles') }}
         restore-keys: |
           ${{ runner.os }}-node-
-    - run: |
+    - name: Install dependencies
+      run: |
         cd ui
-        npm i
-        npm run build && npm run export
+        npm ci
```
kushthedude, line 104: Can we not use this, as npm ci removes the node_modules which will nullify the cache action.

.github/workflows/ci.yml
```diff
@@ -81,22 +86,28 @@ jobs:
...
         cd ..
         cd provider-ui
-        npm i
-        npm run build && npm run export
+        npm ci
```
kushthedude, line 107: This is to check the build and export of Meshery bundle, I don't think npm ci will execute export 🤔 ?

.github/workflows/ci.yml
```diff
@@ -81,22 +86,28 @@ jobs:
...
+        path: |
+          ~/cache
+          ~/.cache
+          !~/cache/exclude
+          **/node_modules
+        key: ${{ runner.os }}-node-${{ hashFiles('**/lockfiles') }}
         restore-keys: |
           ${{ runner.os }}-node-
-    - run: |
+    - name: Install dependencies
```
kushthedude, line 101: Why is the rename required ?

kushthedude (commented):
Everything looks nice @ramrodo , just some small queries else its lgtm

.github/workflows/ci.yml
ramrodo: Ok, let me get back that

.github/workflows/ci.yml
ramrodo: Ok, let me get back that to `npm i`. I'm building and exporting Next App [here](https://github.com/ramrodo/meshery/blob/gh-1037-add-e2e-test-for-userpreference/.github/workflows/ci.yml#L188)

kushthedude: Got it, then I guess let's get rid of this action @ramrodo . With e2e in place I don't think we need additional UI build check

.github/workflows/ci.yml
ramrodo: That was only to have a better descriptive name for that step:

Before:

image


After:

image

leecalcote (approved):
Keep 'em coming, @ramrodo. This is great. :

PR #1868, opened 2020-11-06
[UI] Istio control plane addition to the dashboard

### LINKED ISSUE #1856: [UI] Istio: Initial Control Plane Dashboard
### Current Behavior
MeshSync v1 collection information about Istio service mesh deployments.

### Desired Behavior
Istio control plane information should be exposed in the Meshery UI.

### Implementation
- use `/api/mesh/scan` endpoint to retrieve Istio service mesh information.
- create two "flat" paper elements to separate Service Mesh and Connection Status columns (sections).
- component versions can be retrieved from "operator.istio.io/version" attribute.

### Acceptance Tests
- namespace dropdown will only list a single namespace currently (istio-system)

### Mockups
Screen Shot 2020-11-03 at 4 15 22 PM

tangledbytes:
**Description**
This PR adds Istio control plane to the dashboard. It adds a generic control plane component to ease out the addition of control planes to the dashboard.
This PR fixes #1856

TODO:
1. [x] Add a generic control plane component
2. [x] Redesign the dashboard
3. [x] Identify all the available namespaces and filter the data accordingly
3. [x] Extract `config` from the MeshScan v1 response object.

Relevant Screenshots:
- When the service mesh is detected
meshery_sm1
- When the service mesh is not detected
meshery_sm0

**Notes for Reviewers**
1. As @kumarabd informed that the MeshScan api response object isn't finalized yet so the extraction of `config` is temporary at the moment.
2. Initially a common `paper` component was being used to wrap all of the component in the `_app.js` file which I had to remove to accommodate the new dashboard design. Now each component adds there own `Paper` component, giving more flexibility but adding a bit of code repetition, suggestions are welcome!

### DIFF
--- a/ui/components/DashboardComponent.js
+++ b/ui/components/DashboardComponent.js
@@ -255,26 +255,14 @@ class DashboardComponent extends React.Component {
           // Check if Istio data is present in the scan
           if (Array.isArray(result.Istio)) {
             const istioData = result.Istio.map(comp => {
-              const compData = {}
-              if(comp.metadata?.name === "istio-egressgateway") {
-                compData.name = "Egress Gateway";
-                compData.version = `v${comp.metadata?.labels["operator.istio.io/version"]}`;
-                compData.config = `${comp.spec?.replicas} Gateway`;
-                compData.namespace = comp.metadata?.namespace;
-              } else if(comp.metadata?.name === "istio-ingressgateway") {
-                compData.name = "Ingress Gateway";
-                compData.version = `v${comp.metadata?.labels["operator.istio.io/version"]}`;
-                compData.config = `${comp.spec?.replicas} Gateway`;
-                compData.namespace = comp.metadata?.namespace;
-              } else if (comp.metadata?.name === "istiod") {
-                compData.name = "Istiod";
-                compData.version = `v${comp.metadata?.labels["operator.istio.io/version"]}`;
-                compData.config = `${comp.metadata?.labels["operator.istio.io/component"]}`;
-                compData.namespace = comp.metadata?.namespace;
+              const compData = {
+                name: comp.metadata.name,
+                component: comp.metadata.labels["operator.istio.io/component"],
+                version: `v${comp.metadata?.labels["operator.istio.io/version"]}`,
+                namespace: comp.metadata.namespace
               }
               return compData;
             })
-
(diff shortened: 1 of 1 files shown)


kushthedude:
A screenshot or gif for the change would be appreciable

ui/components/DashboardComponent.js
```diff
@@ -195,6 +225,64 @@ class DashboardComponent extends React.Component {
...
+          self.setState({ 
+            meshScanNamespaces: [...namespaces], 
+            activeMeshScanNamespace: [...namespaces][0] || "" 
+          });
+
+          // Check if Istio data is present in the scan
+          if (Array.isArray(result.Istio)) {
+            const istioData = result.Istio.map(comp => {
+              const compData = {}
+              if(comp.metadata?.name === "istio-egressgateway") {
```
kushthedude, line 259: Why the hardcoding? By this you are limiting the data being display, query over the array, and display metadata.name for each service detected.

kushthedude (changes requested):
I suggest make a table with headers, `Name`, `Component`, `Version`.
For name query over `metadata.name`, for component query over `metadata.labels. operator.istio.io/component `, and for Version query over `metadata.labels. operator.istio.io/component`.

With the above suggestion, you will be saved from querying over hardcoded components + this will also let you know about add-on components like Kiali, Prometheus installed in the namespace.

ui/components/DashboardComponent.js
utkarsh-pro: Actually this was kind of hardcoded because I couldn't figure out a way to convert `istio-egressgateway` to `Istio Egress` as it was in the mockup and same goes for ingress gateway. Should I add a default case where the `comp.metadata.name` is displayed when moment of the cases match?

kushthedude: See the above review

leecalcote:
Understanding that support for additional service meshes is forthcoming, we will want to change this message to be generic: "No service meshes detected in the $cluster_name cluster."

Screen Shot 2020-11-08 at 10 02 50 AM

leecalcote (approved):
looking good, @utkarsh-pro! Nice. :)

We're ready for the same thing for `linkerd`

Screen Shot 2020-11-08 at 10 02 50 AM

PR #1870, opened 2020-11-06
fix: meshery helm chart notes are invalid

### LINKED ISSUE #1257: [Install]Meshery helm chart installation is not working
**Description**

Meshery helm chart installation is not working. The deployment applications are fired up but meshery is not recognised as a service which makes the installation fail at port forwarding step.

**Expected Behavior**

Helm installation should work fine

kushthedude:
**Description**

fixes #1257 
- Meshery pod spins up on 8080 port whereas notes expect it on 80 which is wrong hence results in port-forward refused.


Aisuko (approved):
LGTM

Issue #1872, opened 2020-11-06
Add AKS support in mesheryctl


kushthedude:
**Current Behavior**

Mesheryctl only supports GKE & Minikube

**Desired Behavior**

Add AKS support in mesheryctl

PR #1875, opened 2020-11-07
fix: bash script to utilise dynamic input

kushthedude:

**Description**

This PR refers #1872

### DIFF
--- a/mesheryctl/internal/cli/root/system/config.go
+++ b/mesheryctl/internal/cli/root/system/config.go
@@ -173,14 +173,16 @@ var configCmd = &cobra.Command{
 
 			// Write AKS compatible config to the filesystem
 			if err := utils.GenerateConfigAKS(resourceGroup, aksName); err != nil {
-				log.Fatal("Error generating config:", err)
+				log.Fatal("Error generating kubeconfig: ", err)
 				return
 			}
 		default:
 			log.Fatal("The argument has to be one of gke | minikube | aks")
 		}
 
-		configPath := "/tmp/meshery/kubeconfig.yaml"
+		// TODO: Assumes Mac or Linux. Make arch-specific
+		// Issue: https://github.com/layer5io/meshery/issues/1894
+		configPath := "$HOME/.meshery/aks-kubeconfig.yaml"
 
 		log.Info(tokenPath)
 		contexts, err := getContexts(configPath, tokenPath)


kushthedude:
With this, we can use AKS cluster in mesheryctl 🎉

mesheryctl/pkg/utils/scripts.go
```diff
@@ -141,20 +145,16 @@ func GenerateConfigAKS() error {
...
 	
 	fetch_aks_script() {
-		echo -n "enter your resource group"
-		printf "\n"
-		read -r resource_group
-		echo -n "enter your aks cluster name"
 		printf "\n"
-		read -r aks_name
-		az aks get-credentials --resource-group "${resource_group}" --name "${aks_name}" --file "${TARGET_FILE}"
+		az aks get-credentials --resource-group "${Resource_Group}" --name "${Cluster_Name}" --file "${TARGET_FILE}"
```
leecalcote, line 149: Where are we checking for the `az` dependency?

leecalcote (changes requested):

mesheryctl/pkg/utils/scripts.go
kushthedude: We are not as it is a prereq, or do we want to make script fetch the az too if not found?

leecalcote: Great question. For now, if we just gracefully handle the potentially missing prerequisite error, we can merge this PR and put out a nice-to-have PR for a contributor to consider picking up two auto install `az`￼￼￼.

kushthedude: Done!

leecalcote: I wonder if this is the same formatting issue being run into by Prateek T.?

kushthedude: Yes, the formatting issue is due to the presence of editor config for golang files ig. Every IDE formats according to golang standards but after seeing editorconfig the presets are overridden resulting in failure of gofmt. 😓

mesheryctl/pkg/utils/scripts.go
```diff
@@ -1,172 +1,172 @@
...
+	set -e
+	set -o pipefail
+
+	TARGET_FOLDER="/tmp/meshery"
+	TARGET_FILE="$TARGET_FOLDER/kubeconfig.yaml"
+
+	Resource_Group="%s"
+	Cluster_Name="%s"
+
+	check_for_az() {
```
leecalcote, line 141: Do we exit if not found? or is this handled elsewhere?

mesheryctl/pkg/utils/scripts.go
leecalcote: Got it, ok. Prateek's PR was held up on account of this perhaps for longer than it should have been in retrospect.

mesheryctl/pkg/utils/scripts.go
leecalcote: `az version` will error with a "command not found", I assume, but no instructions to users to install the prerequisite.

kushthedude: We exit if the command is not found.

kushthedude: > `az version` will error with a "command not found", I assume, but no instructions to users to install the prerequisite.

I thought of handling this later by providing an option to install `az` in the mesheryctl utility itself

leecalcote: Makes sense, however, in effect, this check doesn't do anything different than if the check wasn't included.

kushthedude: done the changes, see if it looks alright!

leecalcote:
Bugs to fix:

- https://github.com/layer5io/meshery/issues/1896
- https://github.com/layer5io/meshery/issues/1895

Issue #1782, opened 2020-10-29
[Helm] Add OSM, Kuma, NSM adapters to the Meshery Helm Chart


leecalcote:
**Current Behavior**
The Meshery Helm chart is missing adapters.

**Desired Behavior**
Meshery Helm chart should have all adapters.

Aisuko:
This issue on my schedule

Aisuko:
We can close this issue because of the referenced issues above.

PR #1900, opened 2020-11-10
Add meshery adaptor chart for networking service mesh

Aisuko:
Signed-off-by: aisuko <urakiny@gmail.com>

**Description**

This PR related to #1782

**Notes for Reviewers**

### DIFF
--- a/install/kubernetes/helm/meshery/Chart.yaml
+++ b/install/kubernetes/helm/meshery/Chart.yaml
@@ -29,7 +29,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 1.2.0
+version: 1.3.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
@@ -39,20 +39,23 @@ appVersion: stable-latest
 
 dependencies:
 - name: mesheryistio
-  version: 0.1.0
+  version: 1.0.0
   condition: mesheryistio.enabled
 - name: mesherylinkerd
-  version: 0.1.0
+  version: 1.0.0
   condition: mesherylinkerd.enabled
 - name: mesheryconsul
-  version: 0.1.0
+  version: 1.0.0
   condition: mesheryconsul.enabled
 - name: meshery-kuma
-  version: 0.1.0
+  version: 1.0.0
   condition: meshery-kuma.enabled
 - name: meshery-osm
-  version: 0.1.0
+  version: 1.0.0
   condition: meshery-osm.enabled
+- name: meshery-nsm
+  version: 1.0.0
+  condition: meshery-nsm.enabled
 
 maintainers:
   - name: Layer5 Authors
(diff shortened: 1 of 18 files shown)


leecalcote:
@edwarnicke fyi

install/kubernetes/helm/meshery/Chart.yaml
```diff
@@ -39,20 +39,23 @@ appVersion: stable-latest
...
+  version: 1.0.0
   condition: mesheryistio.enabled
 - name: mesherylinkerd
-  version: 0.1.0
+  version: 1.0.0
   condition: mesherylinkerd.enabled
 - name: mesheryconsul
-  version: 0.1.0
+  version: 1.0.0
   condition: mesheryconsul.enabled
```
leecalcote, line 49: @Aisuko thanks for this. It would be great if we can revisit the other charts and update their names to include consistent hyphenation (e.g. meshery-consul)

leecalcote (approved):

install/kubernetes/helm/meshery/Chart.yaml
Aisuko: No problems. I will fix that in next coming PR

install/kubernetes/helm/meshery/charts/meshery-osm/templates/NOTES.txt
```diff
@@ -17,5 +17,5 @@
 {{- else if contains "ClusterIP" .Values.service.type }}
   export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "meshery-osm.name" . }},app.kubernetes.io/instance={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
   echo "Visit http://127.0.0.1:8080 to use your application"
-  kubectl --namespace {{ .Release.Namespace }} port-forward $POD_NAME 8080:80
+  kubectl --namespace {{ .Release.Namespace }} port-forward $POD_NAME 8080:10009
```
rrachev, line 20: Why we not use {{ .Values.service.port }} to populate value of port "10009". This could be used for all port and in this way we will have only one place where they have to be changed (values.yaml), because from what I noticed we changed them manually here in NOTES.txt

Aisuko: `port-forward` command forward the service which in the pod, we should not use the `service` port and we need to use the `service` port.

Aisuko: That's may good point is here that we can change the service port which in the pod into `values.yaml`. I will do this in next PR.

rrachev: Yes, my point is to point it to the correct port from the values.yaml

Aisuko: I mean we add the adapter pod port into values.yaml, not mean use the service port of the pod. That's two different things.

PR #1923, opened 2020-11-12
Proposal for replacing kubescan with new implementation

tangledbytes:
**Description**
This PR replaces the previous implementation of the `ScanKuberenetes` function. 
Previous implementation used to first find all the available namespaces and then look for the deployments in those deployments. This implementation used to miss some of the pods.
The replacement is a single step process where it queries for all of the pods and filters out **kube-system** namespace pods. 

This change had to be rippled down to the dashboard control pane component to avoid breaking things.
Relevant screenshots:
meshery_sm1

**Notes for Reviewers**
1. The `name` column in the dashboard pane is showing the name of the running pods as it is. Please suggest if we should clean up that name by removing `pod-template-hash`.

### DIFF
--- a/ui/components/DashboardComponent.js
+++ b/ui/components/DashboardComponent.js
@@ -265,8 +265,9 @@ class DashboardComponent extends React.Component {
             const istioData = result.Istio.map(comp => {
               const compData = {
                 name: comp.metadata.name,
-                component: comp.metadata.labels["operator.istio.io/component"],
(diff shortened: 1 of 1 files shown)


leecalcote (approved):
This looks great, @utkarsh-pro. Very nice.

leecalcote:
@utkarsh-pro, yes, if you can remove the `pod-template-hash` from the main view, but add those details (the full name and hash) as a tooltip, that would be great.

Also, let's rename the `name` column to `Control Plane Pods`.

PR #1928, opened 2020-11-14
Utkarsh pro/meshery/kubernetes service scan

### LINKED ISSUE #772: [UI] Display discovered Prometheus and Grafana instances
**Current Behavior**
Based on #770, Meshery is cognizant of whether Prometheus and Grafana instances are deploying in the currently connected Kubernetes cluster.

However, Meshery does not automatically expose the Prometheus or Grafana instances in the UI. 

**Desired Behavior**
If only one instances of Prometheus or Grafana or one of both is available, set these as the active Prometheus and Grafana connections. User can change these connections in the /settings screen. Users can see these connections in the Dashboard screen. Users can also click to perform an ad-hoc "ping" (connectivity test) between Meshery and the Prom and Graf instances.

**Example**
Ideally, this field is pre-populated with the Grafana endpoint.

Screen Shot 2020-03-27 at 10 57 00 PM

---
**Future**
- support more a than one connection to Prometheus
- support more than one connection to Grafana

tangledbytes:
**Description**
This PR adds dropdown select for both grafana and prometheus configuration menu. Along with that it adds minor bug fixes for proper filtering of pods in kubernetes_scan.go. 

Following is the comparison for the proposed change in `ScanKubernetes` function:

| Pros                                                                                                                                                                                    |                                                                            Cons                                                                            |
|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:----------------------------------------------------------------------------------------------------------------------------------------------------------:|
| Better filtering of pods, in the previous approach  identifying pods based on container images created  ambiguity as some pods have containers from different meshes (eg. Linkerd pods) | Matching on names can be problematic if someone deploys a pod with name similar to a pod created by a service mesh. (eg. someone creates a pod istio-test) |

PR also adds a function to scan services based on names which is used to scan prometheus and grafana but **does not** replaces  `detectServiceForDeployementImage`. Although `detectServiceForDeployementImage` is no longer being used anywhere else.

Relevant screenshots:
meshery_grafana
meshery_prometheus

Additional note: With the current implementation, the drop down displays the urls in the dropdown iff prometheus and grafana services have an allocated external ip (ie external ip must not be in *pending* state). This was tested on KinD with metallb.

This PR fixes #772 

**Notes for Reviewers**
@leecalcote following issues were addressed:
1. Add "https" protocol for port 443
2. Show "NA" if we fail to extract version from the mesh scan rather than "vundefined"
However, error notification needs a few clarifications on my part and could be addressed in a subsequent PRs?


ui/components/MesherySettings.js
```diff
@@ -129,6 +139,64 @@ class MesherySettings extends React.Component {
...
+              self.setState(state => ({ scannedPrometheus: [...state.scannedPrometheus, ...urls] }))
+            }
+          })
+        }
+
+        if (Array.isArray(result.grafana)) {
+          result.grafana.forEach(graf => {
+            if (Array.isArray(graf.status?.loadBalancer?.ingress)) {
+              const urls = graf.status.loadBalancer.ingress.map((ip, i) => {
+                return `http://${ip.ip}:${graf.spec?.ports[i]?.port}`
```
leecalcote, line 174: A potential issue with secure endpoints (https)

kushthedude:
Why we need to scan based on names? Scanning on images is more reliable and extensible.

ui/components/MesherySettings.js
utkarsh-pro: @leecalcote, Actually I couldn't find a reliable way to determine "http" and "https". Should I assign "https" if the port is 443 and http for the rest? But I think that other ports can also be used other than 443 for https, not sure about that though.

utkarsh-pro:
> Why we need to scan based on names? Scanning on images is more reliable and extensible.

@kushthedude 
1. In the initial  implementation we were looking for all of the pods inside the deployments and filtering them based on the container images. But this approach used to miss out any pod that was not a part of a deployment (as far as I remember, it was significant in case of nsm).
2. I replaced the previous implementation with an implementation which is analogous to `kubectl get pods -A` by which we could access all of the pods and I was filtering based on container images inside the pod. The problem with filtering based on image names emerged when I realized that some of the linkerd pods have istio based containers too which was increasing ambiguity.
3. In the final code that I have just pushed, I am matching/filtering based on the names of the pods this removes the container images ambiguity and returns all of the pods **but** will be problematic if someone manually deploys a pod with similar names to the pods deployed by a service mesh, for eg `linkerd-test`.

I couldn't figure out a better alternative :sweat_smile:

kushthedude:
> > Why we need to scan based on names? Scanning on images is more reliable and extensible.
> 
> @kushthedude
> 
> 1. In the initial  implementation we were looking for all of the pods inside the deployments and filtering them based on the container images. But this approach used to miss out any pod that was not a part of a deployment (as far as I remember, it was significant in case of nsm).

Then NSM may move away from this approach soon, pods and its count are controller by the deploymentSets. Sp querying the deployment images is not a bad approach but yet more efficient than kgp -A .

> 2. I replaced the previous implementation with an implementation which is analogous to `kubectl get pods -A` by which we could access all of the pods and I was filtering based on container images inside the pod. The problem with filtering based on image names emerged when I realized that some of the linkerd pods have istio based containers too which was increasing ambiguity.

Can you share the containerImages names of linkerd which were creating issues? Also if few linkerd deployments have istio containerImages then what is the ambiguity here since they are part of different mesh deployments?

> 3. In the final code that I have just pushed, I am matching/filtering based on the names of the pods this removes the container images ambiguity and returns all of the pods **but** will be problematic if someone manually deploys a pod with similar names to the pods deployed by a service mesh, for eg `linkerd-test`.

Wont be needed if the above points are addressed.

utkarsh-pro:
> > > Why we need to scan based on names? Scanning on images is more reliable and extensible.
> > 
> > 
> > @kushthedude
> > 
> > 1. In the initial  implementation we were looking for all of the pods inside the deployments and filtering them based on the container images. But this approach used to miss out any pod that was not a part of a deployment (as far as I remember, it was significant in case of nsm).
> 
> Then NSM may move away from this approach soon, pods and its count are controller by the deploymentSets. I dont know if maintaing bare pods without deployments has any advantage, So querying the deployment images is not a bad approach but yet more efficient than kgp -A .

Oh no, I did not mean bare pods. I meant pods created by daemonset, which both consul and nsm do. Kind of new to this so I am not sure if we should be showing them.
Also, I thought that `kgp -A` would be more efficient than the previous implementation as the previous implementation used to first query all the namespaces then all the deployments and then used to get all the pods. Where as right now it's a single query to `/api/v1/pods`. Again, I am not sure if internally this is an expensive call, I considered lowering number of calls and nested loops only.
> 
> > 2. I replaced the previous implementation with an implementation which is analogous to `kubectl get pods -A` by which we could access all of the pods and I was filtering based on container images inside the pod. The problem with filtering based on image names emerged when I realized that some of the linkerd pods have istio based containers too which was increasing ambiguity.
> 
> Can you share the containerImages names of linkerd which were creating issues? Also if few linkerd deployments have istio containerImages then what is the ambiguity here since they are part of different mesh deployments?

I think I couldn't clarify in my previous comment, I meant that I was already querying only for the pods (`kgp -A`) so when I was comparing the container images in those **pods**, some of the linekrd pods had istio images. Example is linkerd-controller (It took me several hours to reproduce this issue though, steps followed: Deploy istio, prometheus, grafana, linkerd latest).

Sample json response of kubernetes api server to `/api/v1/pods`
[sample_response.zip](https://github.com/layer5io/meshery/files/5541086/sample_response.zip)

ui/components/MesherySettings.js
leecalcote: @utkarsh-pro you're right. There really isn't a definitive way to identify this protocol… Or at least not an easy one. Yes, the logic that you highlighted regarding HTTPS for 443 and HTTP for everything else will work for the vast majority of users. Refining this overtime is something that we can do with user feedback.

leecalcote:
I think you gents hashed through the logic as to why changing to a pod name approach solves some problems. Efficiency isn't a concern in MeshSync v1, but is addressed in the v2 architecture instead.

leecalcote:
I think the only outstanding item on this PR is that of including logic for HTTPS when port 443 is present.

leecalcote:
This error is one that we will want to avoid using a toaster notification for.

Screen Shot 2020-11-14 at 11 56 02 PM

While we should show the user this error, this error is one that we will want to avoid using a toaster notification for. Instead what we should aim for is incorporating badges into Meshery's use of chips (https://material-ui.com/components/badges/).

Screen Shot 2020-11-14 at 11 54 55 PM

leecalcote:
Let's avoid including the hardcoded "v" when MeshSync is unable to retrieve a version number.

Screen Shot 2020-11-14 at 9 27 07 PM

leecalcote (dismissed):
The comments that I've noted on this PR can either be addressed in this PR or turned into issues to be addressed in subsequent PR(s).

utkarsh-pro:
> The comments that I've noted on this PR can either be addressed in this PR or turned into issues to be addressed in subsequent PR(s).

@leecalcote I'll address them in my subsequent commits in the same PR.

leecalcote:
@utkarsh-pro the new changes look good. I came across another issue:

Screen Shot 2020-11-15 at 8 35 47 PM

Expect to see `http://localhost:9000`.

Output from `/api/prometheus/scan`:

```
{
   "prometheus":[
      {
         "metadata":{
            "name":"prometheus",
            "namespace":"istio-system",
            "selfLink":"/api/v1/namespaces/istio-system/services/prometheus",
            "uid":"69388fc2-4ed4-4494-a747-a674892814d4",
            "resourceVersion":"473127",
            "creationTimestamp":"2020-11-16T02:34:02Z",
            "labels":{
               "app":"prometheus",
               "chart":"prometheus-11.7.0",
               "component":"server",
               "heritage":"Helm",
               "release":"prometheus"
            },
            "managedFields":[
               {
                  "manager":"kube-vpnkit-forwarder",
                  "operation":"Update",
                  "apiVersion":"v1",
                  "time":"2020-11-16T02:34:02Z"
               },
               {
                  "manager":"meshery-istio",
                  "operation":"Update",
                  "apiVersion":"v1",
                  "time":"2020-11-16T02:34:02Z"
               }
            ]
         },
         "spec":{
            "ports":[
               {
                  "name":"http",
                  "protocol":"TCP",
                  "port":9090,
                  "targetPort":9090,
                  "nodePort":32603
               }
            ],
            "selector":{
               "app":"prometheus",
               "component":"server",
               "release":"prometheus"
            },
            "clusterIP":"10.105.224.124",
            "type":"LoadBalancer",
            "sessionAffinity":"None",
            "externalTrafficPolicy":"Cluster"
         },
         "status":{
            "loadBalancer":{
               "ingress":[
                  {
                     "hostname":"localhost"
                  }
               ]
            }
         }
      },
      {
         "metadata":{
            "name":"osm-prometheus",
            "namespace":"osm-system",
            "selfLink":"/api/v1/namespaces/osm-system/services/osm-prometheus",
            "uid":"bc1e4605-f38f-4f99-b85a-1ad16ea191c2",
            "resourceVersion":"14009",
            "creationTimestamp":"2020-11-13T01:49:49Z",
            "labels":{
               "app.kubernetes.io/managed-by":"Helm"
            },
            "annotations":{
               "meta.helm.sh/release-name":"osm",
               "meta.helm.sh/release-namespace":"osm-system",
               "prometheus.io/port":"7070",
               "prometheus.io/scrape":"true"
            },
            "managedFields":[
               {
                  "manager":"Go-http-client",
                  "operation":"Update",
                  "apiVersion":"v1",
                  "time":"2020-11-13T01:49:49Z"
               }
            ]
         },
         "spec":{
            "ports":[
               {
                  "protocol":"TCP",
                  "port":7070,
                  "targetPort":7070
               }
            ],
            "selector":{
               "app":"osm-prometheus"
            },
            "clusterIP":"10.106.115.124",
            "type":"ClusterIP",
            "sessionAffinity":"None"
         },
         "status":{
            "loadBalancer":{
               
            }
         }
      }
   ]
}
```

leecalcote (changes requested):
Additional details: 

`http://undefined:9090` also shows when only the Istio instance of Prometheus is deployed as type LoadBalancer with a value of localhost.

utkarsh-pro:
> Additional details:
> 
> `http://undefined:9090` also shows when only the Istio instance of Prometheus is deployed as type LoadBalancer with a value of localhost.

@leecalcote, unfortunately initially I didn't consider `hostname` while creating the urls. I think the recent push should have fixed that. Can you please test the latest push as I don't have a system to test `hostname` on my system (kind + metallb)?

I have also now included `clusterIP` based urls now,
meshery_sm1

leecalcote:
Looking good, @utkarsh-pro. I’m kicking off the build again with @kushthedude having fixed it.

utkarsh-pro:
@leecalcote, I am not sure why golang lint is failing although nothing in golang has really changed in this commit. I have also tested it locally by running `golangci-lint run` and it seemed to work fine.

leecalcote (approved):

leecalcote:
Will publish into Meshery Edge release channel now...

PR #1949, opened 2020-11-18
Add redirectErrorToConsole method to let mesh sync call fail silently

### LINKED ISSUE #1948: [UI] MeshSync v1 error is being pushed in a toaster notification
**Description**
This error of failing to match a string should not be included in a toaster notification.

Screen Shot 2020-11-17 at 2 07 10 PM

tangledbytes:

**Description**
This PR adds a redirectErrorToConsole method to let errors produced by meshsync v1 scan call fail silently (i.e the user won't get a toaster notification for them) instead the errors will be redirected to the browser console under a relevant group.

This PR also attempts to fix the cause of the popup mentioned in issue #1948 .

This PR fixes #1948

**Notes for Reviewers**

### DIFF
--- a/ui/components/DashboardComponent.js
+++ b/ui/components/DashboardComponent.js
@@ -365,7 +365,7 @@ class DashboardComponent extends React.Component {
           }
         }
       },
-      self.handleError("Unable to fetch mesh scan data.")
+      self.redirectErrorToConsole("Unable to fetch mesh scan data.")
     );
   }
 
@@ -393,11 +393,13 @@ class DashboardComponent extends React.Component {
    * @returns {string}
    */
   generateMeshScanVersion = (versionStr) => {
-    const matchResult = versionStr?.match(/\d+(\.\d+){2,}/g)[0];
+    if (typeof versionStr !== "string") return "NA";
+
+    const matchResult = versionStr.match(/\d+(\.\d+){2,}/g);
     if (!matchResult) return "NA";
     
     // Add "v" iff we have a valid match result
-    return `v${matchResult}`;
+    return `v${matchResult[0]}`;
   }
 
   handleError = (msg) => (error) => {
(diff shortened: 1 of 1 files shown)


ui/components/DashboardComponent.js
```diff
@@ -414,6 +416,19 @@ class DashboardComponent extends React.Component {
     });
   };
 
+  /**
+   * redirectErrorToConsole returns a function which redirects
+   * ther error to the console under the group labelled by the "msg"
+   * param
+   * @param {string} msg
```
kushthedude, line 423: Have you configured your editor as typescript one? Why such annotations get added everytime?

ui/components/DashboardComponent.js
```diff
@@ -365,7 +365,7 @@ class DashboardComponent extends React.Component {
           }
         }
       },
-      self.handleError("Unable to fetch mesh scan data.")
+      self.redirectErrorToConsole("Unable to fetch mesh scan data.")
```
kushthedude, line 368: No need for a new function if not widely used, change this to
```js
    this.props.enqueueSnackbar("Unable to fetch mesh scan data", {
      variant: "error",
      action: (key) => (
        <IconButton key="close" aria-label="Close" color="inherit" onClick={() => self.props.closeSnackbar(key)}>
          <CloseIcon />
        </IconButton>
      ),
      autoHideDuration: 7000,
    });
```
As we need to show a toaster for failed scan

ui/components/DashboardComponent.js
utkarsh-pro: No, actually they are [JSDoc](https://jsdoc.app/),  I add them manually as they assist:
1. In understanding what code does and helps other developers to understand what parameters a function can take in and what it can return, `any` isn't very helpful.
2. It helps code editor/IDE to facilitate code completion.
3. I use `// @ts-check` to check for type issues before pushing the code to ensure that I am not invoking any invalid method. By default editors use type inference to show such errors but usually they fallback to infer `any` which again isn't very helpful.

ui/components/DashboardComponent.js
utkarsh-pro: So I should not let the meshsync fail silently and instead show "Unable to fetch mesh scan data" for every error?

ui/components/DashboardComponent.js
kushthedude: Nice tool, thanks for info!

ui/components/DashboardComponent.js
kushthedude: Pardon for my wrong, just do console.log(err) and nothing else.

leecalcote (approved):
Thanks!

PR #1987, opened 2020-11-25
Added reset function call in update

### LINKED ISSUE #1190: [mesheryctl] add `system reset` function to `system update`
**Current Behavior**
As the "stable" release channel is created, users will not naturally upgrade to this channel. Their ~./meshery/meshery.yaml file will need to be updated. This update is only made when performing a `system reset`. 

**Desired Behavior**
Users upgrading to newer versions of Meshery should be updated to the stable release channel, unless they have otherwise specified the "edge" channel, which we don't have yet. Essentially, all new and existing users should be migrated to the "stable" channel. Adding a call to the `reset` function inside of the `update` command will cause users to upgrade to the stable channel the next time they start Meshery (the `start` command implicitly calls `update`) unless the `--skip-update` flag is present.

**Implementation**
- add a call to the `reset` function from within the `update` command.

**Roadmap**
In the future, additional controls will be provided to allow users to control which version of Meshery they want to run.

Sudhish2101:
Signed-off-by: sudhish <sudhishmishra824@gmail.com>

**Description**
Added a call to the system reset function from within the update command.

This PR fixes #1190 

**Notes for Reviewers**

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -34,8 +34,6 @@ linters-settings:
       - wrapperFunc
   gocyclo:
     min-complexity: 15
-  gofmt:
-    simplify: false
   goimports:
     local-prefixes: github.com/golangci/golangci-lint
   golint:


.golangci.yml
```diff
@@ -34,6 +34,8 @@ linters-settings:
       - wrapperFunc
   gocyclo:
     min-complexity: 15
+  gofmt:
```
kushthedude, line 37: Why this change?

mesheryctl/internal/cli/root/system/update.go
```diff
@@ -44,8 +44,12 @@ var updateCmd = &cobra.Command{
 				return errors.Wrapf(err, utils.SystemError(fmt.Sprintf("failed to download %s file from %s", utils.DockerComposeFile, fileURL)))
 			}
 		}
+		err := resetMesheryConfig()
+		if err != nil {
+			return errors.Wrap(err, utils.SystemError("failed to update meshery containers"))
```
kushthedude, line 49: Can this error be changed specific to the function which is called?
```suggestion
			return errors.Wrap(err, utils.SystemError("failed to reset meshery configuration"))
```

kushthedude (changes requested):
looks good to me! few neats and doubts 😄

.golangci.yml
Sudhish2101: The first commit which I made for the PR did not pass the golangci-lint check with an error msg: "File is not `gofmt`-ed with `-s` (gofmt)". Then I found out that this is an error that occurs mainly on Windows (and since I'm also on windows) with correctly formatted .go files that have CRLF line endings. To solve this I had to use the -s flag which stands for 'simplify' inside the gofmt.

mesheryctl/internal/cli/root/system/update.go
Sudhish2101: @kushthedude I also thought the same at first, but since this function call is inside the update command if a user hits update and an error msg pops up related to resetting meshery configuration it might confuse the user. So maybe the same error message as the update function would be better. Correct me if I'm wrong.

.golangci.yml
kushthedude: This is not an error, this means you may need to run `gofmt -s -w` on your files to fix it. We should fix the errors in files rather than amending the CI rules until and unless the rule is buggy 🙂

Sudhish2101: Ohhh. Shall I delete that line and make the PR again after fixing it?

kushthedude: Yes, please revert the change and update PR accordingly.

leecalcote:
Thanks for this, @Sudhish2101 👍

kushthedude (approved):
looks great @Sudhish2101 !

PR #1990, opened 2020-11-27
feat: add channel switch in mesheryctl

### LINKED ISSUE #1986: [CI] Allow users to switch release channels with mesheryctl
**Current Behavior**
Users must manually switch between release channels.

**Desired Behavior**
Create a convenient sub-command, `mesheryctl system channel [stable|edge]`

kushthedude:
**Description**

This PR fixes #1986 

**Test Usage**

```sh
./mesheryctl system channel --set edge|stable
```


kushthedude:
@leecalcote Not to be merged until all adapters have got edge release channels.

leecalcote:
@kushthedude if you would like to get this into a release quickly in order to test out, you can land it under the `exp` command.

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,133 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var channelName string
+
+// configCmd represents the config command
+var channelCmd = &cobra.Command{
+	Use:   "channel",
+	Short: "Switch the release channel",
```
leecalcote, line 32:
```suggestion
	Short: "Switch between release channels",
```

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,133 @@
...
+	Long:  `Switch the available release channels 'Stable' & 'Edge'`,
```
leecalcote, line 33:
```suggestion
	Long:  `Subscribe to a release channel. Choose between either 'stable' or 'edge' channels.`,
```

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,133 @@
...
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+
+		if len(channelName) < 0 {
+			log.Fatal("Give a valid channel name")
```
leecalcote, line 38:
```suggestion
			log.Fatal("Provide 'stable' or 'edge' as the release channel name.")
```

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,133 @@
...
+
+		}
+		if channelName == "" {
+			log.Fatal("Invalid channel name")
```
leecalcote, line 42:
```suggestion
			log.Fatal("Invalid channel name. Provide 'stable' or 'edge' as the release channel name.")
```

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,133 @@
...
+				log.Fatal("Error generating config:", err)
+				return
+			}
+		case "edge":
+			if err := useEdgeChannel(); err != nil {
+				log.Fatal("Error generating config:", err)
+				return
+			}
+		default:
+			log.Fatal("Channel name has to be Stable | Edge.")
```
leecalcote, line 57:
```suggestion
			log.Fatal("Channel name has to be 'stable' or 'edge'.")
```

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		if utils.IsMesheryRunning() {
+			if err := stop(); err != nil {
+				log.Fatal("Failed to stop meshery:", err)
```
leecalcote, line 62:
```suggestion
				log.Fatal("Failed to stop Meshery:", err)
```

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,133 @@
...
+				return
+			}
+		}
+
+		if err := start(); err != nil {
+			log.Fatal("Failed to start meshery:", err)
```
leecalcote, line 68:
```suggestion
			log.Fatal("Failed to start Meshery:", err)
```

leecalcote (approved):
Ideally, move to `exp` given use of a new prerequisite `sed`

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,78 @@
...
+
+		}
+		if channelName == "" {
+			log.Fatal("Invalid channel name. Provide 'stable' or 'edge' as the release channel name.")
+		}
+
+		switch channelName {
+		case "stable":
+			if err := utils.SearchAndReplace(utils.DockerComposeFile, "edge", "stable"); err != nil {
+				log.Fatal("Error switching channel:", err)
```
leecalcote, line 44:
```suggestion
				log.Fatal("Error subscribing to release channel:", err)
```

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,78 @@
...
+				return
+			}
+		case "edge":
+			if err := utils.SearchAndReplace(utils.DockerComposeFile, "stable", "edge"); err != nil {
+				log.Fatal("Error switching channel:", err)
```
leecalcote, line 49:
```suggestion
				log.Fatal("Error subscribing to release channel:", err)
```

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,78 @@
...
+				return
+			}
+		default:
+			log.Fatal("Channel name has to be 'stable' or 'edge'.")
```
leecalcote, line 53:
```suggestion
			log.Fatal("Please subscribe to either the 'stable' or 'edge' release channel.")
```

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,78 @@
...
+		log.Info("Successfully switched channel...")
+
+		if utils.IsMesheryRunning() {
+			if err := stop(); err != nil {
+				log.Fatal("Failed to stop Meshery:", err)
+				return
+			}
+		}
+
+		if err := start(); err != nil {
```
leecalcote, line 65: This assumes that the user desires to run Meshery. We shouldn't make that assumption.

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,78 @@
...
+				log.Fatal("Error switching channel:", err)
+				return
+			}
+		default:
+			log.Fatal("Channel name has to be 'stable' or 'edge'.")
+		}
+
+		log.Info("Successfully switched channel...")
+
+		if utils.IsMesheryRunning() {
```
leecalcote, line 58: Subscribing to a release channel should not cause Meshery to stop. To make it easy for people to switch, we could offer a `-r` for restart or prompt them with a [y/n] to restart or not. From there, we should invoke the `restart` function.

Alternative to offering a `-r`, we can present the user with two different flags:

- `--set` which doesn't start, stop, restart or prompt to restart.
- `--switch` which restarts Meshery.

Let's reuse the `restart()` function. If that function is missing logic necessary for reuse here, let's consider adding that logic to the `restart()` function.

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -0,0 +1,78 @@
...
+			return
+		}
+	},
+}
+
+func init() {
+	channelCmd.Flags().StringVarP(&channelName, "set", "s", "stable", "Release channel to be used for Meshery and its adapters.")
+	err := channelCmd.MarkFlagRequired("set")
+	if err != nil {
+		log.Fatal("Failed to mark set as required flag")
```
leecalcote, line 76: Ideally, we are denoted `set` with a special marking like "--set" or "set" in this string.

mesheryctl/pkg/utils/helpers.go
```diff
@@ -450,3 +450,26 @@ func ContentTypeIsHTML(resp *http.Response) bool {
 	}
 	return false
 }
+
+func SearchAndReplace(path, oldString, newString string) error {
```
leecalcote, line 454: Can use use more descriptive parameters? e.g. 

"currentChannel" and "newChannel"

"path" is the Meshery Configuration file, correct? If so, could we rename to "mesheryConfig" or similar variable name?

mesheryctl/internal/cli/root/system/channel.go
kushthedude: If user will have meshery running then only we will do the start else we won't

mesheryctl/pkg/utils/helpers.go
```diff
@@ -450,3 +450,26 @@ func ContentTypeIsHTML(resp *http.Response) bool {
...
+	fileData, err := ioutil.ReadFile(path)
+	if err != nil {
+		log.Errorf("unable to read file: %v", err)
```
leecalcote, line 457: "unable to read Meshery configuration file"

mesheryctl/pkg/utils/helpers.go
```diff
@@ -450,3 +450,26 @@ func ContentTypeIsHTML(resp *http.Response) bool {
 	}
 	return false
 }
+
+func SearchAndReplace(path, oldString, newString string) error {
+	fileData, err := ioutil.ReadFile(path)
```
leecalcote, line 455: "fileData"... perhaps, better named "mesheryConfig" or similar?

mesheryctl/pkg/utils/helpers.go
```diff
@@ -450,3 +450,26 @@ func ContentTypeIsHTML(resp *http.Response) bool {
...
+	if err != nil {
+		log.Errorf("unable to read file: %v", err)
+		return err
+	}
+
+	if !strings.Contains(string(fileData), oldString) {
+		log.Errorf("unable to replace with %s", newString)
+		return errors.New("unable to find requested string")
```
leecalcote, line 463: Can we update this logic to consider for the case that someone may be "setting" or "switching" to the same channel that they are already subscribed to? In this case, we wouldn't error, but report success and a note that the setting remains unchanged.

As the return message is updated, can we move away from "string" and replace with "release channel"?

mesheryctl/pkg/utils/helpers.go
```diff
@@ -450,3 +450,26 @@ func ContentTypeIsHTML(resp *http.Response) bool {
...
+	}
+
+	outputFile := strings.ReplaceAll(string(fileData), oldString, newString)
+
+	err = ioutil.WriteFile(path, []byte(outputFile), 0644)
+	if err != nil {
+		log.Errorf("unable to write into file: %v", err)
```
leecalcote, line 470: Perhaps, we could change this to: "Subscription to $releaseChannel" not successful. Could not write to Meshery configuration file: err"

mesheryctl/pkg/utils/helpers.go
kushthedude: I was thinking of handling this in the channel command, as I wanted to be the following function be a general utility not specific to channel command/

mesheryctl/pkg/utils/helpers.go
kushthedude: I was thinking of handling this in the channel command, as I wanted to be the following function be a general utility not specific to channel command/

mesheryctl/pkg/utils/helpers.go
leecalcote: Output from a test using the current logic in which the user is already subscribed to `stable`:

```
$ ./mesheryctl system channel --set stable
unable to replace with stable
Error switching channel:unable to find requested string
```

leecalcote:
Re-running failed job...

anirudhjain75:
There are so many changes. What am I to look into here ?

leecalcote (approved):
I believe everything is in order here. Let's merge and test in an edge release. 

Switching over the edge release will be its first test! 😆

leecalcote (approved):

leecalcote:
@anirudhjain75 thank you for looking at this. Let's (all) do some user acceptance testing with this in an edge release. Thank you, big @kushthedude!

PR #2114, opened 2020-12-19
Docs: Modified release page to use release collections

SudhanAruna:
**Description**
In this PR release page is modified to use release collection to display the release versions with separate markdown file 
for each release.
This PR fixes #

**Notes for Reviewers**


docs/_config.yml
```diff
@@ -90,6 +90,8 @@ collections:
   es:
     output: true
     permalink: /:collection/:path
+  releases:
```
leecalcote, line 93: Looks good

docs/_layouts/releasepage.html
```diff
@@ -0,0 +1,42 @@
+<!DOCTYPE html>
+<html lang="en" class="no-js">
+  <script src="https://code.jquery.com/jquery-3.2.1.min.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/clipboard@1/dist/clipboard.min.js"></script>
```
leecalcote, line 4: you will want to include head.html instead of redefining it here. Better yet, using the existing default layout but including the releases collection  will work better. There shouldn’t be a need to create a new layout.

leecalcote (changes requested):

docs/_layouts/releasepage.html
SudhanAruna: Sorry, I have remove it.

leecalcote:
@SudhanAruna, well done! One small item to deal with here: the descending order of releases by version number is being itemized slightly out of order with the string “v0.4.9” coming before “v0.4.25", because the “9” is greater than “2", as opposed to “25” being greater than “9".

leecalcote:
https://github.com/layer5io/meshery/issues/2126 has been opened to track the ordering of releases.

Super fantastic work here, @SudhanAruna. It's been 16 months since this challenge was first presented to the community. Way to tackle it!

leecalcote (approved):

PR #2227, opened 2021-01-07
remotecomponent for userpref

vineethvanga18:
**Description**

**Notes for Reviewers**

### DIFF
--- a/ui/components/RemoteUserPref.js
+++ b/ui/components/RemoteUserPref.js
@@ -23,8 +23,7 @@ const RemoteUserPref = ({ startOnZoom, handleToggle, url }) => {
   return (
     <div>
       <RemoteComponent
-        injectProps={{ handleToggle, startOnZoom }}
-        handleToggle={handleToggle}
+        injectProps={{ startOnZoom, handleToggle }}
       />
     </div>
   );


leecalcote:
@vineethvanga18 will you update the Meshery Extensibility document accordingly?

leecalcote:
@vineethvanga18 I get a blank screen. 
Screen Shot 2021-01-07 at 7 36 09 PM

vineethvanga18:
That is because we still don't get user_pref component from the extensions end point

hwpplayer1:
> That is because we still don't get user_pref component from the extensions end point

Could you please point the lines between the working code and "breakline"

leecalcote:
> That is because we still don't get user_pref component from the extensions end point

@vineethvanga18 the two preferences for anonymous test results and statistics should be part of Meshery's core, while MeshMap preferences should be part of the extension.

ui/components/RemoteUserPref.js
```diff
@@ -0,0 +1,33 @@
...
+    );
+  }
+  if (err != null) {
+    return <div>Unknown Error: {err.toString()}</div>;
+  }
+
+  return (
+    <div>
+      <RemoteComponent
+        startOnZoom={startOnZoom}
```
utkarsh-pro, line 26: @vineethvanga18 you called it `injectProps` in the extension so here it would be like:
```suggestion
        injectProps={{ handleToggle, startOnZoom }}
```

vineethvanga18: Yeah I used it, not sure why I removed it.

vineethvanga18:
> > That is because we still don't get user_pref component from the extensions end point
> 
> @vineethvanga18 the two preferences for anonymous test results and statistics should be part of Meshery's core, while MeshMap preferences should be part of the extension.

I just realized you were on ui-dev-server. Those preferences are fetched from "/api/user/stats" and for them to show up you got to run meshery server too. @leecalcote

leecalcote (approved):
Oh, I see now! Thanks, @vineethvanga18

PR #2236, opened 2021-01-09
[mesheryctl] Dynamically modified docker-compose based on current-context

### LINKED ISSUE #2224: [mesheryctl] config: “platform” awareness for system commands for Kubernetes
### Current Behavior:
A context unaware static config file (meshery.yaml) which is used for deploying meshery in kubernetes.  Meshery will support three types of platforms: docker, kubernetes, podman. 

### Desired Behavior:
All system commands should be cognizant of the specific platform used by each context and interface with that platform accordingly.

### Future: 
Meshery may differentiate between different types of Kubernetes systems.

### Implementation:
1. Use viper to read, modify, write (rewrite) the meshery.yaml.
2. Viper should update the meshery.yaml according to the context configuration specified in config.yaml.
3. Only one `meshery.yaml` will be managed by `mesheryctl`.
4. `docker-compose` is being wrapped for start, stop, logs, restart

### Acceptance Tests:
1. All `system` commands are aware of the `platform` attribute.
2. Only one `meshery.yaml` will be present in the `~/.meshery` app folder.
3. Each system command will consider the platform type prior to executing.

### Out of Scope:
1. Users will not be provided the ability to bring their own manifest.yaml as opposed to using meshery.yaml.

anirudhjain75:
Fixes #2224 ##2207

### DIFF
--- a/mesheryctl/internal/cli/root/system/system.go
+++ b/mesheryctl/internal/cli/root/system/system.go
@@ -17,7 +17,7 @@ package system
 import (
 	"fmt"
 
-	config2 "github.com/layer5io/meshery/mesheryctl/internal/cli/root/config"
+	config "github.com/layer5io/meshery/mesheryctl/internal/cli/root/config"
 	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/system/context"
 
 	"github.com/layer5io/meshery/mesheryctl/pkg/utils"
@@ -45,7 +45,7 @@ var SystemCmd = &cobra.Command{
 		if ok := utils.IsValidSubcommand(availableSubcommands, args[0]); !ok {
 			return errors.New(utils.SystemError(fmt.Sprintf("invalid command: \"%s\"", args[0])))
 		}
-		mctlCfg, err := config2.GetMesheryCtl(viper.GetViper())
+		mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
 		if err != nil {
 			return errors.Wrap(err, "error processing config")
 		}


mesheryctl/internal/cli/root/system/config.go
```diff
@@ -49,7 +50,7 @@ type k8sContext struct {
 func getContexts(configFile, tokenPath string) ([]string, error) {
 	client := &http.Client{}
 
-	mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
+	mctlCfg, err := config2.GetMesheryCtl(viper.GetViper())
```
kumarabd, line 53: Why `config2`? I think there is no other package or variable with the name `config` - @anirudhjain75

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -42,10 +45,10 @@ var startCmd = &cobra.Command{
 	Short: "Start Meshery",
 	Long:  `Run 'docker-compose' to start Meshery and each of its service mesh adapters.`,
 	Args:  cobra.NoArgs,
-	PreRunE: func(cmd *cobra.Command, args []string) error {
-		//Check prerequisite
-		return utils.PreReqCheck(cmd.Use)
-	},
+	//PreRunE: func(cmd *cobra.Command, args []string) error {
```
kumarabd, line 48: Why is Pre-check removed? @anirudhjain75

mesheryctl/internal/cli/root/system/system.go
```diff
@@ -45,7 +45,7 @@ var SystemCmd = &cobra.Command{
 		if ok := utils.IsValidSubcommand(availableSubcommands, args[0]); !ok {
 			return errors.New(utils.SystemError(fmt.Sprintf("invalid command: \"%s\"", args[0])))
 		}
-		mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
+		mctlCfg, err := config2.GetMesheryCtl(viper.GetViper())
```
kumarabd, line 48: Same here on why make it `config2`

kumarabd (changes requested):

mesheryctl/internal/cli/root/system/config.go
anirudhjain75: There is one inside system.go. Do have a look you'll see

mesheryctl/internal/cli/root/system/start.go
anirudhjain75: Because moving forward we are looking for multiple platform support @leecalcote can brief you more on why I removed it

mesheryctl/internal/cli/root/system/system.go
anirudhjain75: There is one inside system.go. Do have a look you'll see

mesheryctl/internal/cli/root/system/config.go
kumarabd: yes but why would that matter here?

mesheryctl/internal/cli/root/system/start.go
kumarabd: okay, @leecalcote please intervene. I think the PreCheck is necessary to check for docker-compose availability.

kumarabd:
@leecalcote A little involvement on the PreReq() removal in the above.

mesheryctl/internal/cli/root/system/start.go
leecalcote: As we build in support for `platform`, we'll want to ensure that a prerequisite check is being run for docker-compose/docker anytime that `mesheryctl` is invoking docker-compose/docker. 


We'll need a switch/case statement to ensure that "docker" is being handled for contexts of those platform types.

anirudhjain75: Sure. I'll make the change and push it in about 10 hours from now

anirudhjain75: Made the changes

mesheryctl/internal/cli/root/system/system.go
```diff
@@ -17,7 +17,7 @@ package system
 import (
 	"fmt"
 
-	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/config"
+	config2 "github.com/layer5io/meshery/mesheryctl/internal/cli/root/config"
```
leecalcote, line 20: "config2"... is there a conflict with using "config" as the package name?

anirudhjain75: Nah it was fixed . This must be a leftover

anirudhjain75: Just committed to fix this

anirudhjain75:
@leecalcote @abhishekkumkar This should be ready to merge. Please go ahead with it

leecalcote (approved):

Issue #2158, opened 2020-12-30
[Platform] Meshery support for Podman


tripathyprateek:
#### Current Behavior
Meshery deploys as a set of containers. Deployment manifests/packages/tooling has been provided for Docker and Kubernetes-based platforms - https://docs.meshery.io/installation/platforms

RedHat based Linux distribution have shifted to using Podman instead of Docker. Running Meshery on one of these distros means installing Docker. 

#### Desired Behavior

Make it more convenient to tun Meshery on podman systems. Support podman as a platform.

#### Implementation
1. Use [podman-compose](https://github.com/containers/podman-compose) which has been created to mirror `docker-compose`, which is what `mesheryctl` uses to deploy Meshery locally.
2. Add "podman" as a platform type to Meshery config file (`config.yaml`) (using `mesheryctl system context`)

#### Acceptance Tests
1. All `mesheryctl system` commands should behave in the same way they do against `docker-compose` as the platform.

jaybatra26:
@leecalcote  Please shed some light on point 2. What is the config.yaml here you are referring to?

jaybatra26:
Hi @leecalcote Do I have to make changes in a this file most likely. https://github.com/layer5io/meshery/blob/master/mesheryctl/pkg/utils/helpers.go#L208

leecalcote:
@jaybatra26 hopefully, this helps - https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.fpuuy11gk85g

jaybatra26:
Thanks @leecalcote

PR #2256, opened 2021-01-16
[mesheryctl] Rework mesheryctl system channel to confirm to latest design and make it context and compose aware. Fixes #2246

anirudhjain75:
**Description**

Modify channel command to use context instead of directly modifying the file, allow system start to also build changes from the version and channel of current context allowing reuse of code for switch in channel command. In struct change from yaml to mapstructure to conform to viper usage

This PR fixes #2246 #2128

PR #2257, opened 2021-01-17
WIP: Make use of Podman-compose

jaybatra26:
Since Docker-compose is going to be decommissioned.
This PR replaces docker-compose with podman-compose.

Fixes: #2158
Signed-off-by: Jay Batra <jaybatra73@gmail.com>

**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/internal/cli/root/system/start.go
+++ b/mesheryctl/internal/cli/root/system/start.go
@@ -96,159 +96,159 @@ func start() error {
 	if err != nil {
 		return errors.Wrap(err, "error processing config")
 	}
-	if(mctlCfg.Contexts["local"].Platform== "docker") 
-	{
+	ctxPlatform := mctlCfg.GetContextContent().Platform
+	if ctxPlatform == "docker" || ctxPlatform == "" {
 		// Viper instance used for docker compose
-	utils.ViperCompose.SetConfigFile(utils.DockerComposeFile)
-	err = utils.ViperCompose.ReadInConfig()
-	if err != nil {
-		return err
-	}
-
-	compose := &utils.DockerCompose{}
-	err = utils.ViperCompose.Unmarshal(&compose)
-	if err != nil {
-		return err
-	}
-
-	services := compose.Services                                        // Current Services
-	RequestedAdapters := mctlCfg.GetContextContent().Adapters           // Requested Adapters / Services
-	err = ValidateComposeFileForRecreation(services, RequestedAdapters) // Validates docker-compose file and recreates and sync's if required
-	if err != nil {
-		return err
-	}
-
-	err = utils.ViperCompose.Unmarshal(&compose)
-	if err != nil {
-		return err
-	}
-	services = compose.Services // Current Services
-	RequiredService := []string{"meshery", "watchtower"}
-
-	AllowedServices := map[string]utils.Service{}
-	for _, v := range RequestedAdapters {
-		if services[v].Image == "" {
-			log.Fatalf("Invalid adapter specified %s", v)
-		}
-		AllowedServices[v] = services[v]
-	}
-	for _, v := range RequiredService {
-		AllowedServices[v] = services[v]
-	}
-
-	utils.ViperCompose.Set("services", AllowedServices)
-	err = utils.ViperCompose.WriteConfig()
-	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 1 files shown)


leecalcote:
@ishankhare07, would you like to review this PR?

leecalcote (changes requested):
@jaybatra26, `docker-compose` will still be a supported platform for Meshery (the default platform). `podman-compose` will be an additionally supported platform. You'll want to add support for `podman-compose` to most of the `mesheryctl system` commands (start, stop, restart, update, logs, channel...) 

See the `platform` attribute in `~/.meshery/config.yaml` as the key to use when adding "podman" as an option. `system context` is a new comment, not yet released into a stable Meshery release. `system context` creates and controls `config.yaml` and currently only supports "docker" as a value. It needs to support "podman" and "kubernetes" as values, too.

leecalcote:
Thank you for taking this on, @jaybatra26 👍

jaybatra26:
@leecalcote  It's my pleasure. :)

jaybatra26:
Hi @leecalcote Do I have to make changes in a this file most likely. https://github.com/layer5io/meshery/blob/master/mesheryctl/pkg/utils/helpers.go#L208

tripathyprateek:
Yes we need to try and test how podman-compose behaves with Meshery. I guess it will be a good enough issue to tackle for some time. And we need to keep support for Docker as default platform. Also there seems some more development around Podman to support Docker-compose with Podman 3.0 [here](https://www.redhat.com/sysadmin/podman-docker-compose)

anirudhjain75:
Hi @jaybatra26 , Do look into the golint error caused in the CI

jaybatra26:
Thanks @tripathyprateek  and @anirudhjain75 sure.

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -96,159 +96,159 @@ func start() error {
 	if err != nil {
 		return errors.Wrap(err, "error processing config")
 	}
-	if(mctlCfg.Contexts["local"].Platform== "docker") 
-	{
+	ctxPlatform := mctlCfg.GetContextContent().Platform
+	if ctxPlatform == "docker" || ctxPlatform == "" {
```
anirudhjain75, line 100: We might not want to default to docker. @leecalcote shed some light if possible. I dont think we should do anything or maybe give an error if platform is not set.

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -96,159 +96,159 @@ func start() error {
 	if err != nil {
 		return errors.Wrap(err, "error processing config")
 	}
-	if(mctlCfg.Contexts["local"].Platform== "docker") 
-	{
+	ctxPlatform := mctlCfg.GetContextContent().Platform
```
anirudhjain75, line 99: These is an addon that can be done here. from #2256 I am planning to add a context flag. Taking that into consideration for which context to focus on would be great. Another issue could be created for this.

anirudhjain75 (commented):
The changes look fine to be. ignoring the docker as default for now. @leecalcote might be able to shed some light

leecalcote:
> Yes we need to try and test how podman-compose behaves with Meshery. I guess it will be a good enough issue to tackle for some time. And we need to keep support for Docker as default platform. Also there seems some more development around Podman to support Docker-compose with Podman 3.0 [here](https://www.redhat.com/sysadmin/podman-docker-compose)

@tripathyprateek I would love to see results from your compatibility tests here. 👍

leecalcote:
@tripathyprateek have you been successful in running Meshery on Podman? Any luck progressing on the mesheryctl command here? 

Assuming we're in about the same spot, I'll suggest that we push this feature to post v0.5.0. What do you think?

tripathyprateek:
> @tripathyprateek have you been successful in running Meshery on Podman? Any luck progressing on the mesheryctl command here?
> 
> Assuming we're in about the same spot, I'll suggest that we push this feature to post v0.5.0. What do you think?

@leecalcote I have been trying to run Meshery on Podman. However I am  running into a lot of errors with Podman and Buildah which I am trying to figure out. I would suggest we can wait to add support for Podman for a little while as Podman is not fully working on some platforms (Ubuntu 20.10 and earlier) I guess.

PR #2287, opened 2021-01-22
[mesheryctl] Feature: `system reset` should ask confirmation from the user

### LINKED ISSUE #2283: [mesheryctl] `system reset` should ask confirmation from the user
#### Current Behavior

Running `mesheryctl system reset` will reset meshery to its default setting. As of PR #2265 (soon to be merged)

```
user@computer:$ mesheryctl system reset
Meshery resetting...

Current Context: local
Channel: stable
Version: v0.4.20

Fetching default docker-compose file at version: v0.4.20...

...Meshery config (/home/user/.meshery/meshery.yaml) now reset to default settings.
```

#### Desired Behavior

`system reset` should ask the user whether they are sure about resetting to default.

#### Implementation

Instead of initiating reset, the `system reset` command could ask the user something like this(open for discussion) as shown and then continue the reset process.

```
user@computer:$ mesheryctl system reset
Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]? y

Meshery resetting...

Current Context: local
Channel: stable
Version: v0.4.20

Fetching default docker-compose file at version: v0.4.20...

...Meshery config (/home/user/.meshery/meshery.yaml) now reset to default settings.
```

pottekkat:

**Description**

This PR fixes #2283

`mesheryctl system reset` command will now ask the user `Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]?` and the user can either continue with the reset or abort it. Example-

```
user@computer: $ mesheryctl system reset
Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]? y
Meshery resetting...

Current Context: local
Channel: stable
Version: v0.4.20

Fetching default docker-compose file at version: v0.4.20...

...Meshery config (/home/user/.meshery/meshery.yaml) now reset to default settings.
user@computer: $
```

And if the user responds 'n'-

```
user@computer: $ mesheryctl system reset
Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]? n
Aborting reset...
user@computer: $
```

**New-** When user enters an invalid response, it asks again until it gets a valid response-

```
user@computer: $ mesheryctl system reset
Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]? d
Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]? f
Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]? g
Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]? h
Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]? j
Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]? no
Aborting reset...
user@computer: $ 
```

**New-** Works if the input is "y,Y,yes,YES,Yes" and "n,N,no,NO,No".
**New-** Now uses the `AskForConfirmation` function that is implemented in `helpers.go`.

**Notes for Reviewers**

Please suggest for any changes to the log messages.
**Slight insignificant change made in the `root.go` file to work with the small change in the `AskForConfirmation` function. Only a single character change.**

### DIFF
--- a/mesheryctl/internal/cli/root/system/reset.go
+++ b/mesheryctl/internal/cli/root/system/reset.go
@@ -43,7 +43,7 @@ func resetMesheryConfig() error {
 	userResponse := utils.AskForConfirmation("Meshery config file will be reset to system defaults. Are you sure you want to continue")
 
 	if !userResponse {
-		log.Info("Aborting reset...")
+		log.Info("Reset aborted.")
 	} else {
 		log.Info("Meshery resetting...\n")
 


ramrodo:
@navendu-pottekkat I found a possible implementation of how to manage okay or not okay responses:

https://gist.github.com/albrow/5882501
```go
func askForConfirmation() bool {
	var response string
	_, err := fmt.Scanln(&response)
	if err != nil {
		log.Fatal(err)
	}
	okayResponses := []string{"y", "Y", "yes", "Yes", "YES"}
	nokayResponses := []string{"n", "N", "no", "No", "NO"}
	if containsString(okayResponses, response) {
		return true
	} else if containsString(nokayResponses, response) {
		return false
	} else {
		fmt.Println("Please type yes or no and then press enter:")
		return askForConfirmation()
	}
}
```

navendu-pottekkat:
@ramrodo This looks good. I thought about going for the "infinite loop" until you get a correct response but then went with just exiting the command.

This seems like a better way to approach it. Thank you for suggesting this. We could use a function like this for every confirmation question we need to handle in mesheryctl. I think for now we only have `system reset` but it maybe useful in the future.

navendu-pottekkat:
@ramrodo It seems like we already have such a function. I think that this approach is more simple and readable but it essentially does the same function. I have implemented it and made a new commit.

```
func AskForConfirmation(s string) bool {
	reader := bufio.NewReader(os.Stdin)

	for {
		fmt.Printf("%s [y/n]? ", s)

		response, err := reader.ReadString('\n')
		if err != nil {
			log.Fatal(err)
		}

		response = strings.ToLower(strings.TrimSpace(response))

		if response == "y" || response == "yes" {
			return true
		} else if response == "n" || response == "no" {
			return false
		}
	}
}
```

leecalcote:
Good work on both of your accounts. 👌

leecalcote:
Using the phrase: “Reset aborted.” might be a slight improvement.

leecalcote:
Overall, this looks great.

leecalcote (approved):

navendu-pottekkat:
> Using the phrase: “Reset aborted.” might be a slight improvement.

Making the change now.

navendu-pottekkat:
@leecalcote Made the changes, ready for merge

leecalcote (approved):
Perfect. Thank you, @navendu-pottekkat 👍

PR #2299, opened 2021-01-26
[mesheryctl] Feature: `--verbose` `-v` global flags for verbose logs

pottekkat:
Signed-off-by: navendu-pottekkat navendupottekkat@gmail.com

**Description**

This PR fixes #2295 

Added `--verbose` `-v` global flag for verbose logs. It basically sets the log level to debug.

This will(in addition to normal logs) print the `debug()` logs.

```go
func setVerbose() {
	log.SetLevel(log.InfoLevel)

	if verbose {
		log.SetLevel(log.DebugLevel)
	}
}
```

Example usage-

`mesheryctl system start -v`

Also removed the `--version` flag. And `--debug` flag was modified to `--verbose` flag.

**Notes for Reviewers**

This global flag was implemented with similar logic to the already existing `--config` flag.

So the `setVerbose` function is passed to the `cobra.OnInitialize()` function.

Please let me know if this is not the right approach for this. I am not really sure about it. But everything seems to work fine and nothing breaks.

- [x] Yes, I signed my commits.

Issue #2295, opened 2021-01-25
[mesheryctl] Add persistent `--verbose` `-v` flag for all commands


navendu-pottekkat:
#### Current Behaviour

Currently, there is a `--debug` `-d` flag which sets output to verbose mode. _It seems to have some error in its implementation._

There is a top level `version` command and a `--version` `-v` flag which is obsolete given the current usage. 

#### Desired Behaviour

1. No need for `--version` `-v` flag. 
2. `--verbose` `-v` for verbose logs.

#### Implementation

Change the flag to `--verbose` `-v` and fix it if there are any errors.

Remove the `--version` `-v` flag.

#### Future Scope

Add more debug logs to commands that does not have any currently.

leecalcote:
Separate from the docs update, it would be good to include this function in a `mesheryctl` Contributor's Guide - https://github.com/layer5io/meshery/issues/2303

PR #2306, opened 2021-01-27
[Docs] Add doc for --verbose flag

pottekkat:

**Description**

Added documentation for the newly added `--verbose` flag. Please refer this PR #2299 and Issue #2295 

This change is made in the https://docs.meshery.io/reference/mesheryctl page and under the "Global Commands and Flags" section.


leecalcote:
In terms of usage, I wonder if this isn’t a slightly better example: `mesheryctl <command> --verbose`, so that uses don’t think this only works for system commands.

not a big deal. It is under the global flags section.

navendu-pottekkat:
> In terms of usage, I wonder if this isn’t a slightly better example: `mesheryctl <command> --verbose`, so that uses don’t think this only works for system commands.
> 
> not a big deal. It is under the global flags section.

Yes it is. I will change it

navendu-pottekkat:
@leecalcote Made the changes. Can merge now...

leecalcote (approved):
Good work, @navendu-pottekkat. 👍 

I tossed in a short description of the command categories.

PR #2368, opened 2021-02-05
[mesheryctl] fix Bug: `system reset` on `stable` channel breaks if no version is specified

### LINKED ISSUE #2366: [mesheryctl] Bug: `system reset` on `stable` channel breaks if no version is specified
#### Description

When `mesheryctl system reset` is run it will fetch the `docker-compose.yaml` file from GitHub corresponding to the `channel` and `version` specified in the `current-context`. 

If the `channel` is `stable` and the `version` is not specified in the `config.yaml` file in the `~/.meshery` folder, it will fail to pick up a valid file.

`version` _is not specified by default when Meshery is installed._

Example `config.yaml` where `system reset` fails-

```
(base) user@computer:~/.meshery$ cat config.yaml 
contexts:
  local:
    token:
      name: Default
      location: auth.json
    platform: kubernetes
    adapters:
    - meshery-istio
    - meshery-linkerd
    - meshery-consul
    - meshery-octarine
    - meshery-nsm
    - meshery-kuma
    - meshery-cpx
    - meshery-osm
    - meshery-nginx-sm
    channel: stable #stable channel but version is not specified
current-context: local
tokens: {}
```

Here, `system reset` will try to pick up a file from `fileURL = "https://raw.githubusercontent.com/layer5io/meshery/" + currVersion + "/docker-compose.yaml"` (line 83 in reset.go).

Since there is no current version specified, the URL will be `https://raw.githubusercontent.com/layer5io/meshery/docker-compose.yaml` which is invalid.

#### Expected Behavior

Instead if the `channel` is `stable` and the `version` is not specified in the `config.yaml` file, it should pick up the `docker-compose.yaml` file from the latest release(_not the edge release but the latest stable release_).

_May need to use GitHub API for this._

aneeshnema:
**Description**

This PR fixes #2366 
- Added a helper function `GetLatestStableReleaseTag` in `mesheryctl/pkg/utils/helpers.go` to fetch latest stable release tag using GitHub API
- This is later used by `resetMesheryConfig` in case no version is specified while using `stable` channel

**Notes for Reviewers**
Should I rather move the `GetLatestStableReleaseTag` logic inside the `resetMesheryConfig` itself? (will have to import additional packages in `reset.go` which are already imported in `helpers.go`)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/layer5io/meshkit v0.2.0
 	github.com/layer5io/meshsync v0.1.5
 	github.com/layer5io/nighthawk-go v0.1.5
-	github.com/layer5io/service-mesh-performance v0.3.2
+	github.com/layer5io/service-mesh-performance v0.3.3
 	github.com/pkg/errors v0.9.1
 	github.com/prologic/bitcask v0.3.9
 	github.com/prometheus/client_golang v1.9.0
--- a/go.sum
+++ b/go.sum
@@ -623,8 +623,8 @@ github.com/layer5io/meshsync v0.1.5 h1:OOQ8DmrjeE2QKQ08ngx9qUCpOoWz0O8xykkRtcszB
 github.com/layer5io/meshsync v0.1.5/go.mod h1:Lla/67ISTsJyiliUSoWt+44dZvR5EoLiFwLLfzg63NQ=
 github.com/layer5io/nighthawk-go v0.1.5 h1:edygERCTGc4Nqxtqr7dSxwgv9iH3+hPNciGl8T4faTY=
 github.com/layer5io/nighthawk-go v0.1.5/go.mod h1:gbi0RbmV9VElcQaJlMAAjEwtMTBzptn7i737/MidKCM=
-github.com/layer5io/service-mesh-performance v0.3.2 h1:bkzKFlgIMtiszq6dIKf5Sjy2Yx2VTFVd5oxuYv5Shcs=
-github.com/layer5io/service-mesh-performance v0.3.2/go.mod h1:W153amv8aHAeIWxO7b7d7Vibt9RhaEVh4Uh+RG+BumQ=
+github.com/layer5io/service-mesh-performance v0.3.3 h1:KtouYXg64y+G0soPJwDeB0sM6PXolBpkV6Ke15aqwmk=
+github.com/layer5io/service-mesh-performance v0.3.3/go.mod h1:W153amv8aHAeIWxO7b7d7Vibt9RhaEVh4Uh+RG+BumQ=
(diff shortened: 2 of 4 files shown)


mesheryctl/internal/cli/root/system/reset.go
```diff
@@ -80,6 +80,13 @@ func resetMesheryConfig() error {
 		if currChannel == "edge" {
 			fileURL = "https://raw.githubusercontent.com/layer5io/meshery/master/docker-compose.yaml"
 		} else if currChannel == "stable" {
+			if currVersion == "" {
+				log.Info("Version unspecified, fetching latest stable release...\n")
```
navendu-pottekkat, line 84: We may not need to log this to the user. It would be confusing.

Since we output the version at the end, I don't think we need this. If this looks good to everyone, then don't mind me.

mesheryctl/pkg/utils/helpers.go
```diff
@@ -591,3 +591,24 @@ func ValidateURL(URL string) error {
 	}
 	return nil
 }
+
+// GetLatestStableReleaseTag fetch and return latest stable release tag
+func GetLatestStableReleaseTag() (string, error) {
```
navendu-pottekkat, line 596: This function is neat! It would be best to have it in the helpers.go file as this would be reused in the future.

Maybe this small change in the function doc.
```suggestion
// GetLatestStableReleaseTag fetches and returns the latest release tag from GitHub
func GetLatestStableReleaseTag() (string, error) {
```

navendu-pottekkat (approved):
That was fast and neat @aneeshnema . LGTM. 💯

I have _suggested_ some small changes.

The rest looks good and you can merge.

navendu-pottekkat:
> Should I rather move the GetLatestStableReleaseTag logic inside the resetMesheryConfig itself? (will have to import additional packages in reset.go which are already imported in helpers.go)

We can have this function in the `helpers.go` file itself. We will need to add the `context` support and `platform` support for some more commands and it would be reused a lot.

mesheryctl/internal/cli/root/system/reset.go
aneeshnema: I was also a little confused about this while adding it. Seemed a little verbose. Let's see what others suggest 👀.

navendu-pottekkat: Yes, that is what I thought. Maybe this could be a debug(verbose) log?

anirudhjain75:
Nice work @aneeshnema . The changes are looking great.

anirudhjain75 (dismissed):
I still see the logs in the file, I agree that removing it would be better

mesheryctl/internal/cli/root/system/reset.go
aneeshnema: removed it ✔

navendu-pottekkat (approved):
LGTM

leecalcote (approved):

PR #2397, opened 2021-02-10
[mesheryctl] Feature: Added helper function for current-context support 

### LINKED ISSUE #2393: [mesheryctl] Respect current-context, fetch correct docker-compose.yaml
#### Description

Whenever `mesheryctl` cannot find `docker-compose.yaml` it downloads it from Github, but all it does not take into account the `correct-context` and always download the latest file from the master branch.
[Several instances](https://github.com/layer5io/meshery/search?q=fileURL) of this behaviour needs to be corrected.

`mesheryctl system reset` has been updated to fetch correct `docker-compose.yaml`  file from GitHub corresponding to the `channel` and `version` specified in the `current-context`. 

The logic inside `resetMesheryConfig` can be extracted to create a separate helper function that fetches correct file.
https://github.com/layer5io/meshery/blob/1a91051afe17d82a668c7429c735bf7ed94970c7/mesheryctl/internal/cli/root/system/reset.go#L80-L90

 

#### Expected Behavior

`mesheryctl` should fetch correct `docker-compose.yaml` based on `current-context`

#### Child Issues

Issue #2394 [mesheryctl] Feature: Add current-context support to `system update`
Issue #2395 [mesheryctl] Feature: Add current-context support to `system log`
Issye #2396 [mesheryctl] Feature: Add current-context support to `system start`

### LINKED ISSUE #2394: [mesheryctl] Feature: Add current-context support to `system update`
#### Description

#### Current Behavior

https://github.com/layer5io/meshery/blob/1a91051afe17d82a668c7429c735bf7ed94970c7/mesheryctl/internal/cli/root/system/update.go#L44-L48

#### Expected Behavior

`mesheryctl system update` should fetch correct `docker-compose.yaml` as per `version` and `channel` specified in `current-context`

aneeshnema:
**Description**

Added helper function to resolve child issues of #2393 
Refactored `resetMesheryConfig` to use the new helper functions `DownloadDockerComposeFile` and `GetCurrentContext`

### DIFF
--- a/mesheryctl/pkg/utils/helpers.go
+++ b/mesheryctl/pkg/utils/helpers.go
@@ -694,36 +694,39 @@ func IsAdapterValid(manifestArr []Manifest, adapterManifest string) bool {
 	return false
 }
 
-// GetCurrentContext returns the current context name and context struct
+// GetCurrentContext returns the current context name and context struct.
+// If the user mentions a temporary context(tempCtxName) with -c flag, change the current-context and proceed to temporary-context
 func GetCurrentContext(tempCtxName string) (string, config.Context, error) {
-	// if the user has mentioned a temporary context in the -c flag, change the context and proceed to reset
 	mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
 	if err != nil {
 		return "", config.Context{}, errors.Wrap(err, "error processing config")
 	}
 
-	if tempCtxName == "" {
-		return mctlCfg.CurrentContext, mctlCfg.Contexts[mctlCfg.CurrentContext], nil
+	if tempCtxName != "" {
+		mctlCfg.CurrentContext = tempCtxName
 	}
-	if tempCtxtStruct, val := mctlCfg.Contexts[tempCtxName]; val {
-		return tempCtxName, tempCtxtStruct, nil
+
+	currCtx, err := mctlCfg.CheckIfCurrentContextIsValid()
+	if err != nil {
+		// if the user specifies a context that is not in the config.yaml file, throw an error and show the available contexts
+		log.Errorf("\n\"%s\" context does not exist. The available contexts are:", mctlCfg.CurrentContext)
+		for context := range mctlCfg.Contexts {
+			log.Errorf("%s", context)
+		}
+		return "", config.Context{}, errors.New("context does not exist")
 	}
-	// if the user specifies a context that is not in the config.yaml file, throw an error and show the available contexts
-	log.Errorf("\n\"%s\" context does not exist. The available contexts are:", tempCtxName)
-	for context := range mctlCfg.Contexts {
-		log.Errorf("%s", context)
-	}
-	return "", config.Context{}, errors.Errorf("\n\"%s\" context does not exist", tempCtxName)
+	return mctlCfg.CurrentContext, currCtx, nil
 }
 
-// DownloadDockerComposeFile fetches docker-compose.yaml based on passed context if it does not exists
+// DownloadDockerComposeFile fetches docker-compose.yaml based on passed context if it does not exists.
+// Use force to override download anyway
 func DownloadDockerComposeFile(ctx config.Context, force bool) (string, error) {
 	if _, err := os.Stat(DockerComposeFile); os.IsNotExist(err) || force {
 		fileURL := ""
 
 		if ctx.Channel == "edge" {
 			fileURL = "https://raw.githubusercontent.com/layer5io/meshery/master/docker-compose.yaml"
-			ctx.Version = "edge"
+			ctx.Version = "latest"
 		} else if ctx.Channel == "stable" {
 			if ctx.Version == "" {
 				ctx.Version, err = GetLatestStableReleaseTag()
--- a/mesheryctl/internal/cli/root/config/config.go
+++ b/mesheryctl/internal/cli/root/config/config.go
@@ -50,12 +50,13 @@ func GetMesheryCtl(v *viper.Viper) (*MesheryCtlConfig, error) {
 
 // CheckIfCurrentContextIsValid checks if current context is valid
 func (mc *MesheryCtlConfig) CheckIfCurrentContextIsValid() (Context, error) {
-	currentContext := mc.CurrentContext
-	if currentContext == "" {
+	if mc.CurrentContext == "" {
 		return Context{}, errors.New("current context not set")
 	}
-
-	return mc.Contexts[currentContext], nil
+	if ctx, exists := mc.Contexts[mc.CurrentContext]; exists {
+		return ctx, nil
+	}
+	return Context{}, errors.New("current context:" + mc.CurrentContext + "does not exist")
 }
 
 // GetBaseMesheryURL returns the base meshery server URL


navendu-pottekkat (commented):
Shouldn't we also consider for the temporary changes in the context? This is implemented in the `system reset` command right now and will also be added to the other commands. 

```
currentContext := mctlCfg.CurrentContext

		// if the user has mentioned a temporary context in the -c flag, change the context and proceed to reset
		if tempContext != "" {
			if _, val := mctlCfg.Contexts[tempContext]; val {
				currentContext = tempContext
			} else {
				// if the user specifies a context that is not in the config.yaml file, throw an error and show the available contexts
				log.Errorf("\n\"%s\" context does not exist. The available contexts are:", tempContext)
				for context := range mctlCfg.Contexts {
					log.Errorf("%s", context)
				}
				return nil
			}
		}
```

One suggestion for this would be to take in the current context as an argument in the function.

navendu-pottekkat:
@aneeshnema merging this PR may close the epic issue since you have mentioned "this PR fixes"...

mesheryctl/internal/cli/root/system/reset.go
```diff
@@ -48,55 +44,22 @@ func resetMesheryConfig() error {
...
-			if currVersion == "" {
-				currVersion, err = utils.GetLatestStableReleaseTag()
-				if err != nil {
-					return errors.Wrapf(err, utils.SystemError(fmt.Sprintf("failed to fetch latest stable release tag")))
-				}
-			}
-			fileURL = "https://raw.githubusercontent.com/layer5io/meshery/" + currVersion + "/docker-compose.yaml"
+		err = utils.DownloadDockerComposeFile(tempContext, true)
+		if err != nil {
+			return errors.Wrap(err, "failed to fetch docker-compose file")
```
navendu-pottekkat, line 60: Why do we need to pass in the tempContext? We are already getting the current context in line 47. Once we have the current-context or temp context, we just need to pass that to the `DownloadDockerComposeFile()`function.

mesheryctl/pkg/utils/helpers.go
```diff
@@ -693,3 +693,58 @@ func IsAdapterValid(manifestArr []Manifest, adapterManifest string) bool {
...
+	}
+	return "", config.Context{}, errors.Errorf("\n\"%s\" context does not exist", tempContext)
+}
+
+// DownloadDockerComposeFile check if docker-compose.yaml exists, if not fetches the correct file based on current-context
+func DownloadDockerComposeFile(tempContext string, force bool) error {
+	if _, err := os.Stat(DockerComposeFile); os.IsNotExist(err) || force {
+		_, currCtx, err := GetCurrentContext(tempContext)
+		if err != nil {
+			return errors.Wrap(err, "failed to retrieve current-context")
```
navendu-pottekkat, line 724: Why do we need to make this call again since we have already picked the current context or the temporary context before?

navendu-pottekkat (commented):
@aneeshnema This may require some changes. I have made some comments on the code and on the first run, I got an error where in my `config.yaml` file, I have not specified the version and I am getting an output like this-

Screenshot from 2021-02-11 17-25-19

Could you please check these?

mesheryctl/internal/cli/root/system/reset.go
aneeshnema: The current context in line 47 was just for getting the name, channel and version of the context. These values are later logged in the function see line 53-56.

mesheryctl/pkg/utils/helpers.go
aneeshnema: Yes we did already get the context, but that is only in the case of `resetMesheryConfig`, it was done for logging purpose as mentioned in the other comment. For the rest of the areas where `DownloadDockerComposeFile` will be used, we will not find the `current-context` beforehand.

aneeshnema:
@navendu-pottekkat its' currently logging the values specified current-context, previously if the version was unspecified the latest version would be downloaded. Now after the refactoring, that logic has been moved to `DownloadDockerComposeFile`. Now to solve this problem, we have two ways:

1. Add that logic back to `resetMesheryConfig` additionally(redundant).
2. Add the logging to `DownloadDockerComposeFile`, but this will log anytime if the docker-compose.yaml file will be downloaded (for all commands)

I personally prefer the 2nd option as it would be good to show the user the configuration they are using, and this would only be one time (since you wouldn't download the file again).

mesheryctl/pkg/utils/helpers.go
navendu-pottekkat: In those cases too, can we use the same logic as we used here? One function which takes in the temporary context and returns either the temporary context or the current context. Another function which takes in the current/temporary context (whichever is set) and then downloads the corresponding version file. I think that approach would be neat and readable.

Adding @anirudhjain75 to this thread.

navendu-pottekkat:
@aneeshnema I think we need the log only for the reset command, and maybe the function can return the version if the version is unspecified.

mesheryctl/pkg/utils/helpers.go
aneeshnema: If we are planning to add temporary-context to other commands, then I think you are correct. Currently, all the other commands except `resetMesheryConfig` download docker-compose file only if it is not found. Ideally, it should download the file if it does not exist or **if the docker-compose file does not match the current-context** (especially needed for temporary-context). The second part is not handled at the moment. Though making mesheryctl always download docker-compose file will be a pain for users not connected to the internet.

navendu-pottekkat: The `--context` flag would be ideally a global flag that is available in all of the mesheryctl commands. This has yet to be implemented but it will be down the road. So we should be adding it in such a way that we won't have to rewrite it in the future.

mesheryctl/pkg/utils/helpers.go
```diff
@@ -693,3 +693,52 @@ func IsAdapterValid(manifestArr []Manifest, adapterManifest string) bool {
...
+}
+
+// DownloadDockerComposeFile fetches docker-compose.yaml based on passed context if it does not exists
+func DownloadDockerComposeFile(ctx config.Context, force bool) (string, error) {
+	if _, err := os.Stat(DockerComposeFile); os.IsNotExist(err) || force {
+		fileURL := ""
+
+		if ctx.Channel == "edge" {
+			fileURL = "https://raw.githubusercontent.com/layer5io/meshery/master/docker-compose.yaml"
+			ctx.Version = "edge"
```
navendu-pottekkat, line 726: Should the version also be "edge"? Or should it be "latest"? The channel and version being "edge" feels a bit off. What does everyone suggest?

navendu-pottekkat (approved):
This looks good. We can use the `GetCurrentContext` and the `DownloadDockerComposeFile` throughout the other commands.

Screenshot from 2021-02-12 09-00-39

One small change I would suggest here is to replace the second _"dev" context does not exist_ with just "context does not exist" since we are already mentioning about the "dev" context in the line above.

Adding comments on the `reset.go` file and a comment on using the `force` bool in the function in `helpers.go` would be helpful in the future.

Rest LGTM and works as expected when testing. Good work!

navendu-pottekkat:
@anirudhjain75 Could you also please check this out before we merge?

anirudhjain75:
Sure. Give me sometime

mesheryctl/pkg/utils/helpers.go
anirudhjain75: Yea this might raise an issue.. Currently the versions are edge-latest and stable-latest. so the version should either be set as latest or be leftout. first alternative is better.

anirudhjain75: Also you dont need a GetCurrentContext function... The mesheryctl config class already has that inplace. A good place to look into is config.go

anirudhjain75 (changes requested):
Hey @aneeshnema great stuff here. Just have a few things you might want to address :).

mesheryctl/pkg/utils/helpers.go
anirudhjain75: if a change is required, feel free to edit config.go instead

aneeshnema:
@leecalcote some of your commits are not signed-off, hence my DCO has failed. The instructions on the DCO page will signs-off with my identity but the author remains the original person. Thus it will fail again due to mismatch.

mesheryctl/pkg/utils/helpers.go
```diff
@@ -693,3 +693,52 @@ func IsAdapterValid(manifestArr []Manifest, adapterManifest string) bool {
 
 	return false
 }
+
+// GetCurrentContext returns the current context name and context struct
+func GetCurrentContext(tempCtxName string) (string, config.Context, error) {
+	// if the user has mentioned a temporary context in the -c flag, change the context and proceed to reset
```
navendu-pottekkat, line 699:
```suggestion
	// if the user mentions a temporary context with the -c flag, change the current-context to temporary context
```

navendu-pottekkat (approved):
Looks good. Thanks for making the changes suggested. @leecalcote  @anirudhjain75  I think we can merge this.

mesheryctl/pkg/utils/helpers.go
```diff
@@ -693,3 +693,52 @@ func IsAdapterValid(manifestArr []Manifest, adapterManifest string) bool {
...
+	}
+	if tempCtxtStruct, val := mctlCfg.Contexts[tempCtxName]; val {
+		return tempCtxName, tempCtxtStruct, nil
+	}
+	// if the user specifies a context that is not in the config.yaml file, throw an error and show the available contexts
+	log.Errorf("\n\"%s\" context does not exist. The available contexts are:", tempCtxName)
+	for context := range mctlCfg.Contexts {
+		log.Errorf("%s", context)
+	}
+	return "", config.Context{}, errors.Errorf("\n\"%s\" context does not exist", tempCtxName)
```
navendu-pottekkat, line 716:
```suggestion
	return "", config.Context{}, errors.Errorf("\n context does not exist")
```

mesheryctl/pkg/utils/helpers.go
aneeshnema: check out the updated code, I don't think we should get rid of GetCurrentContext. If we remove it, we will need to add duplicate code at many places

leecalcote:
@aneeshnema it looks like I'm holding you back. 😉 It's frustrating that commits performed through GitHub's web interface don't include sign-offs despite the user being authenticated. Fixing DCO...

mesheryctl/pkg/utils/helpers.go
navendu-pottekkat: @aneeshnema Yes, that would be best

navendu-pottekkat (approved):
@anirudhjain75 @leecalcote LGTM. We can merge the changes and let @aneeshnema carry on with the awesome work he has been doing.

anirudhjain75:
Merging conflict I guess

aneeshnema:
resolved the conflict ✔ @anirudhjain75 
but now I have DCO problems again. Can someone help? @leecalcote

anirudhjain75:
You can simple foolow steps from dco

aneeshnema:
> @leecalcote some of your commits are not signed-off, hence my DCO has failed. The instructions on the DCO page will signs-off with my identity but the author remains the original person. Thus it will fail again due to mismatch.

@anirudhjain75

leecalcote:
@aneeshnema the likely culprit here is a branch being out of date. You’ll want to sync your fork and branch with the latest upstream. Ideally, those commits are not included in your PR. This section will help - https://github.com/layer5io/meshery/blob/master/CONTRIBUTING-gitflow.md#verify-the-new-

anirudhjain75:
@aneeshnema are the issues with end to end test on ci because of your changes ?

aneeshnema:
@anirudhjain75 nope, I think it's a time out error
https://github.com/layer5io/meshery/pull/2397/checks?check_run_id=1890292464
> upstream request timeoutupstream connect error or disconnect/reset before headers. reset reason: local resetupstream connect error or disconnect/reset before headers. reset reason: local resetupstream connect error or disconnect/reset before headers. reset reason: local resetupstream connect error or disconnect/reset before headers. reset reason: connection failureupstream connect error or disconnect/reset before headers. reset reason: local reset

anirudhjain75:
@leecalcote can you confirm this if not from this PR.

leecalcote:
You're right, gents. This failure isn't do to the code in this PR.

PR #2407, opened 2021-02-14
[mesheryctl] Feature: "context support" for system start command

### LINKED ISSUE #2396: [mesheryctl] Feature: Add current-context support to `system start`
#### Description

#### Current Behavior

Inside function `ValidateComposeFileForRecreation` that is used by `start`
https://github.com/layer5io/meshery/blob/03a4d251a2e0f437e43ed91b91e0e108ef8b0723/mesheryctl/internal/cli/root/system/start.go#L80-L82

#### Expected Behavior

`mesheryctl system start` should fetch correct `docker-compose.yaml` as per `version` and `channel` specified in `current-context`

pottekkat:
**Description**

This PR fixes #2396 

`system start` will now check for the `current-context` in the `config.yaml` file and pulls the `docker-compose.yaml` file according to the `channel` and `version` specified in the current context.

Also added a `--context` `-c` flag to allow users to temporarily change context when the `system start` command is run.

**Notes for Reviewers**

"context support" for `platform: kubernetes` is not added in this PR. The current logic on how the manifest files needs to be pulled from GitHub needs to be changed. This will be done after adding "context support" to other commands in mesheryctl. 

### DIFF
--- a/mesheryctl/internal/cli/root/system/start.go
+++ b/mesheryctl/internal/cli/root/system/start.go
@@ -184,6 +184,9 @@ func start() error {
 	// get the platform, channel and the version of the current context
 	// if a temp context is set using the -c flag, use it as the current context
 	currCtxName, currCtx, err := utils.GetCurrentContext(tempContext)
+	if err != nil {
+		return err
+	}
 	currPlatform := mctlCfg.Contexts[currCtxName].Platform
 	RequestedAdapters := mctlCfg.Contexts[currCtxName].Adapters // Requested Adapters / Services
 


mesheryctl/internal/cli/root/system/start.go
```diff
@@ -250,23 +250,26 @@ func start() error {
...
-					ContextContent.Version = "latest"
-				} else if ContextContent.Channel == "stable" {
-					if ContextContent.Version == "" {
-						ContextContent.Version, err = utils.GetLatestStableReleaseTag()
+				// if channel is edge, pick the edge-latest images
+				if currCtx.Channel == "edge" {
+					currCtx.Version = "latest"
+				} else if currCtx.Channel == "stable" {
+					// if the channel is stable, pick the image corresponding to version
+					// if no version is specified, pick the version of the latest stable release
```
leecalcote, line 261: I love seeing these comments. 😀

leecalcote (approved):
Good!

mesheryctl/internal/cli/root/system/start.go
navendu-pottekkat: Future @navendu-pottekkat thanks present @navendu-pottekkat

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -181,22 +181,23 @@ func start() error {
...
 	case "docker":
 
 		if _, err := os.Stat(utils.MesheryFolder); os.IsNotExist(err) {
-			if err := utils.DownloadFile(utils.DockerComposeFile, fileURL); err != nil {
-				return errors.Wrapf(err, utils.SystemError(fmt.Sprintf("failed to download %s file from %s", utils.DockerComposeFile, fileURL)))
+			// download the docker-compose.yaml file corresponding to the current version
+			if _, err := utils.DownloadDockerComposeFile(currCtx, true); err != nil {
+				return errors.Wrapf(err, utils.SystemError(fmt.Sprintf("failed to download %s file", utils.DockerComposeFile)))
 			}
 		}
```
aneeshnema, line 202: @navendu-pottekkat ` utils.DownloadDockerComposeFile` already handles checking if file exists, that's why we had added `force` argument.
The above code currently skips downloading the file if the folder exists, thus skipping our forceful download. (It also skips the case when `.meshery` folder exists but no `meshery.yaml` file in that folder.
Should I fix this change when working on log and update command child issues?

navendu-pottekkat: @aneeshnema The force argument is set to true right? So it should download the files, shouldn't it?

aneeshnema: @navendu-pottekkat  yes, it will.
I am talking about this,
`		if _, err := os.Stat(utils.MesheryFolder); os.IsNotExist(err) {`
This would skip the inside code block altogether if the `.meshery` folder exists.

navendu-pottekkat: Yep, you could do that. This was done since the meshery.yaml file was recreated in the system start command. Changing that should not break anything. But lets make sure that is the case. Please add it in the PR.

PR #2412, opened 2021-02-14
[mesheryctl] fix context support for all commands, clean-up

### LINKED ISSUE #2393: [mesheryctl] Respect current-context, fetch correct docker-compose.yaml

### LINKED ISSUE #2394: [mesheryctl] Feature: Add current-context support to `system update`

### LINKED ISSUE #2395: [mesheryctl] Feature: Add current-context support to `system log`
#### Description

#### Current Behavior

https://github.com/layer5io/meshery/blob/f8227b1978de965ecad238529a60fae35af61a99/mesheryctl/internal/cli/root/system/logs.go#L45-L49

#### Expected Behavior

`mesheryctl system log` should fetch correct `docker-compose.yaml` as per `version` and `channel` specified in `current-context`

aneeshnema:
**Description**

This PR fixes #2393, fixes #2394 , fixes #2395 

**Notes for Reviewers**

1.  ~Added `skipConfirmation` flag to `resetMesheryConfig`~
2.  Added user confirmation ⚠ to update command 
3.  ~`GetCurrentContext` now also sets the correct `Version` instead of leaving it empty~ 
    ❌function removed, use newly added setter `SetCurrentContext` for similar functionality
4.  Renamed ~`GetContextContent`~ `GetCurrentContext`✔ (getter)
5.  Removed `ValidateComposeFileForRecreation` from start.go because it is not needed now as we are always downloading the correct docker-compose file at every start.
6.  Updated all files based on the above changes.

### DIFF
--- a/mesheryctl/internal/cli/root/config/config.go
+++ b/mesheryctl/internal/cli/root/config/config.go
@@ -1,11 +1,6 @@
 package config
 
 import (
-	"encoding/json"
-	"fmt"
-	"io/ioutil"
-	"net/http"
-
 	log "github.com/sirupsen/logrus"
 
 	"github.com/pkg/errors"
@@ -59,8 +54,8 @@ func (mc *MesheryCtlConfig) CheckIfCurrentContextIsValid() (Context, error) {
 		return Context{}, errors.New("current context not set")
 	}
 	if ctx, exists := mc.Contexts[mc.CurrentContext]; exists {
-		if err := ctx.SetDefaultVersion(); err != nil {
-			return ctx, err
+		if ctx.Version == "" {
+			ctx.Version = "latest"
 		}
 		return ctx, nil
 	}
@@ -111,46 +106,4 @@ func (v *Version) GetBuild() string {
 // GetCommitSHA returns the commit sha for the binary
 func (v *Version) GetCommitSHA() string {
 	return v.CommitSHA
-}
-
-// SetDefaultVersion sets version of config if not specified
-func (ctx *Context) SetDefaultVersion() error {
-	if ctx.Version == "" {
-		// if no version is specified, pick latest imgaes based on channel
-		if ctx.Channel == "edge" {
-			// if channel is edge, pick the edge-latest images
-			ctx.Version = "latest"
-		} else if ctx.Channel == "stable" {
-			// if the channel is stable, pick the version of the latest stable release
-			version, err := GetLatestStableReleaseTag()
-			ctx.Version = version
-			if err != nil {
-				return errors.Wrapf(err, fmt.Sprintf("failed to fetch latest stable release tag"))
-			}
-		} else {
-			return errors.Errorf("unknown channel %s", ctx.Channel)
-		}
-	}
-	return nil
-}
-
-// GetLatestStableReleaseTag fetches and returns the latest release tag from GitHub
-func GetLatestStableReleaseTag() (string, error) {
-	url := "https://api.github.com/repos/layer5io/meshery/releases/latest"
-	resp, err := http.Get(url)
-	if err != nil {
-		return "", errors.Wrapf(err, "failed to make GET request to %s", url)
-	}
-	defer resp.Body.Close()
-
-	var dat map[string]interface{}
-	body, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return "", errors.Wrap(err, "failed to read response body")
-	}
-	if err := json.Unmarshal(body, &dat); err != nil {
-		return "", errors.Wrap(err, "failed to unmarshal json into object")
-	}
-
-	return dat["tag_name"].(string), nil
 }
--- a/mesheryctl/internal/cli/root/system/reset.go
+++ b/mesheryctl/internal/cli/root/system/reset.go
@@ -32,19 +32,17 @@ var resetCmd = &cobra.Command{
 	Long:  `Reset Meshery to it's default configuration.`,
 	Args:  cobra.NoArgs,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return resetMesheryConfig(false)
+		return resetMesheryConfig()
 	},
 }
 
 // resets meshery config, skips conirmation if skipConfirmation is true
-func resetMesheryConfig(skipConfirmation bool) error {
+func resetMesheryConfig() error {
 	// ask user for confirmation
-	if !skipConfirmation {
-		userResponse := utils.AskForConfirmation("Meshery config file will be reset to system defaults. Are you sure you want to continue")
-		if !userResponse {
-			log.Info("Reset aborted.")
-			return nil
-		}
+	userResponse := utils.AskForConfirmation("Meshery config file will be reset to system defaults. Are you sure you want to continue")
+	if !userResponse {
+		log.Info("Reset aborted.")
+		return nil
 	}
 
 	// Get viper instance used for context
--- a/mesheryctl/internal/cli/root/system/start.go
+++ b/mesheryctl/internal/cli/root/system/start.go
@@ -246,7 +246,7 @@ func start() error {
 
 		// Reset Meshery config file to default settings
 		if utils.ResetFlag {
-			err := resetMesheryConfig(false)
+			err := resetMesheryConfig()
 			if err != nil {
 				return errors.Wrap(err, utils.SystemError("failed to reset meshery config"))
 			}
--- a/mesheryctl/internal/cli/root/system/update.go
+++ b/mesheryctl/internal/cli/root/system/update.go
@@ -17,10 +17,12 @@ package system
 import (
 	"github.com/pkg/errors"
 
+	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/config"
 	"github.com/layer5io/meshery/mesheryctl/pkg/utils"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 // updateCmd represents the update command
@@ -37,14 +39,31 @@ var updateCmd = &cobra.Command{
 		return utils.PreReqCheck(cmd.Use, "")
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
-		userResponse := utils.AskForConfirmation("Meshery container images are going to be updated. Are you sure you want to continue")
-		if !userResponse {
-			log.Info("Update aborted.")
-			return nil
-		}
-		err := resetMesheryConfig(true)
+		// Get viper instance used for context
+		mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
 		if err != nil {
-			return errors.Wrap(err, utils.SystemError("failed to update meshery containers"))
+			return errors.Wrap(err, "error processing config")
+		}
+		// get the platform, channel and the version of the current context
+		// if a temp context is set using the -c flag, use it as the current context
+		currCtx, err := mctlCfg.SetCurrentContext(tempContext)
+		if err != nil {
+			return err
+		}
+		if currCtx.Version != "latest" {
+			// ask confirmation if user has pinned the version in config
+			log.Infof("You have pinned version: %s in your current conext", currCtx.Version)
+			userResponse := utils.AskForConfirmation("Updating Meshery container images will supersede the version to latest. Are you sure you want to continue")
+			if !userResponse {
+				log.Info("Update aborted.")
+				return nil
+			}
+			currCtx.Version = "latest"
+		}
+		log.Printf("Fetching latest docker-compose file for channel: %s...\n", currCtx.Channel)
+		err = utils.DownloadDockerComposeFile(currCtx, true)
+		if err != nil {
+			return errors.Wrap(err, "failed to fetch docker-compose file")
 		}
 
 		err = utils.UpdateMesheryContainers()
--- a/mesheryctl/pkg/utils/helpers.go
+++ b/mesheryctl/pkg/utils/helpers.go
@@ -705,6 +705,12 @@ func DownloadDockerComposeFile(ctx config.Context, force bool) error {
 		if ctx.Channel == "edge" {
 			fileURL = "https://raw.githubusercontent.com/layer5io/meshery/master/docker-compose.yaml"
 		} else if ctx.Channel == "stable" {
+			if ctx.Version == "latest" {
+				ctx.Version, err = GetLatestStableReleaseTag()
+				if err != nil {
+					return errors.Wrapf(err, fmt.Sprintf("failed to fetch latest stable release tag"))
+				}
+			}
 			fileURL = "https://raw.githubusercontent.com/layer5io/meshery/" + ctx.Version + "/docker-compose.yaml"
 		} else {
 			return errors.Errorf("unknown channel %s", ctx.Channel)
--- a/mesheryctl/internal/cli/root/system/stop.go
+++ b/mesheryctl/internal/cli/root/system/stop.go
@@ -86,7 +86,7 @@ func stop() error {
 
 	// Reset Meshery config file to default settings
 	if utils.ResetFlag {
-		err := resetMesheryConfig(false)
+		err := resetMesheryConfig()
 		if err != nil {
 			return errors.Wrap(err, utils.SystemError("failed to reset meshery config"))
 		}


mesheryctl/internal/cli/root/system/update.go
```diff
@@ -41,11 +38,6 @@ var updateCmd = &cobra.Command{
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
 
-		if _, err := os.Stat(utils.DockerComposeFile); os.IsNotExist(err) {
-			if err := utils.DownloadFile(utils.DockerComposeFile, fileURL); err != nil {
-				return errors.Wrapf(err, utils.SystemError(fmt.Sprintf("failed to download %s file from %s", utils.DockerComposeFile, fileURL)))
-			}
-		}
 		err := resetMesheryConfig()
 		if err != nil {
```
aneeshnema, line 42: `resetMesheryConfig` already handles downloading the docker-compose file
removed redundant code

TODO: fixing the update command to abort update when reset is aborted. Currently `resetMesheryConfig` does not return anything that can signal reset aborted.

mesheryctl/internal/cli/root/system/logs.go
```diff
@@ -42,10 +41,10 @@ var logsCmd = &cobra.Command{
 
 		log.Info("Starting Meshery logging...")
 
-		if _, err := os.Stat(utils.DockerComposeFile); os.IsNotExist(err) {
-			if err := utils.DownloadFile(utils.DockerComposeFile, fileURL); err != nil {
-				return errors.Wrapf(err, utils.SystemError(fmt.Sprintf("failed to download %s file from %s", utils.DockerComposeFile, fileURL)))
-			}
+		// download(don't force) the docker-compose.yaml file corresponding to the current version
+		_, currCtx, err := utils.GetCurrentContext(tempContext)
+		if _, err := utils.DownloadDockerComposeFile(currCtx, false); err != nil {
```
aneeshnema, line 46: skip the download of docker-compose file if it already exists (since meshery must be already running, the file should be correct?)

navendu-pottekkat: It should be correct. @anirudhjain75 @leecalcote This can be done right?

mesheryctl/internal/cli/root/system/update.go
navendu-pottekkat: When running reset() from the update command, we can maybe pass in an argument that would suppress the confirmation question. If we still need to confirm with the user, we can have a confirmation question in the update command.

navendu-pottekkat (commented):
Looks good. I have left some suggestions for handling reset() in the update command. Please check if that would work.

mesheryctl/internal/cli/root/system/update.go
leecalcote: Let's take note of this in consideration for a future, global `--silent` flag.

leecalcote: Note that the `system update` command sets precedent with `--skip` flag. Different, but somewhat related in concept.

mesheryctl/internal/cli/root/system/logs.go
leecalcote: Right. If the user wants to see Meshery logs and there is no docker-compose file, then we should invoke a central function that does something like:

1. Inform the user of the missing deployment file.
2. Ask if they would like mesheryctl to retrieve a new deployment file (or...)
3. Ask if they would like mesheryctl to generate a new deployment file (assuming that mesheryctl has this ability at some point).

mesheryctl/internal/cli/root/system/update.go
navendu-pottekkat: Got it. @leecalcote but how should we handle it now? When the user runs `system update`, it will call the `reset()` function which will ask for confirmation from the user. If the user responds "no", the `system update` command would not consider it and still continue with the update process. This should be changed.

leecalcote: Currently, (in the released versions), `mesheryctl system update` does not call `reset()`. I haven't read through this whole issue. Is there a need to change this behavior?

navendu-pottekkat: It resets first and then proceeds to update doesn't it. We do not need to change this... The issue is that since we added a user confirmation in the reset function, we need to also add handling for it in the update command.

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -194,11 +194,9 @@ func start() error {
...
 	case "docker":
 
-		if _, err := os.Stat(utils.MesheryFolder); os.IsNotExist(err) {
-			// download the docker-compose.yaml file corresponding to the current version
-			if _, err := utils.DownloadDockerComposeFile(currCtx, true); err != nil {
-				return errors.Wrapf(err, utils.SystemError(fmt.Sprintf("failed to download %s file", utils.DockerComposeFile)))
-			}
+		// download the docker-compose.yaml file corresponding to the current version
+		if _, err := utils.DownloadDockerComposeFile(currCtx, true); err != nil {
+			return errors.Wrapf(err, utils.SystemError(fmt.Sprintf("failed to download %s file", utils.DockerComposeFile)))
```
aneeshnema, line 199: This update was done to ensure the correct version of the docker-compose file is downloaded based on context.
For example:
suppose I already have a file `meshery.yaml` that was downloaded when my context used `channel: edge`,
now I change my context such that it uses `channel: stable` . 
There is no indicator in `meshery.yaml`  that can identify the version of the file.

If this is acceptable, then we can also remove `ValidateComposeFileForRecreation` function as it becomes redundant.

@anirudhjain75 @navendu-pottekkat

anirudhjain75: So the meshery.yaml is to be fetched only when there is a change in version. The reasonable way would be to compare the latest releasr version and see it there is a change... The compose is to be built using context, a function can be created so as to update the meshery.yaml based off on context

leecalcote:
@aneeshnema is this PR ready for review? It is currently marked as Draft.

mesheryctl/internal/cli/root/system/start.go
leecalcote: Is this an optimization that can be addressed later?

navendu-pottekkat:
@leecalcote 

@aneeshnema and I will discuss with @anirudhjain75 on this on a call tomorrow and we would hopefully be able to clear things up and @aneeshnema will probably be able to get it done tomorrow itself. There is just some confusion about refactoring things.

aneeshnema:
@leecalcote no, there are few more changes requires as per suggestions above. I was busy travelling/shifting back to my college campus so couldn't work on the PR. Will update tomorrow.

leecalcote:
What?! In-person schooling? So fancy. What a refreshing change, I'm sure.

navendu-pottekkat (approved):
@aneeshnema Looks good to me! 

Great work with the refactoring. It may be best if we test it a bit more since we have made a few changes and make sure that nothing is breaking.

So `mesheryctl` is fully context supported?

aneeshnema:
@navendu-pottekkat yes, mesheryctl is fully context supported now (for docker platform). I have tried all the commands on my local machine, everything seems to work fine.  Still, do test out yourself, just in case.

mesheryctl/internal/cli/root/system/reset.go
```diff
@@ -30,35 +30,37 @@ var resetCmd = &cobra.Command{
...
+		if !userResponse {
+			log.Info("Reset aborted.")
+			return nil
 		}
+	}
 
-		log.Printf("Current Context: %s", currCtxName)
-		log.Printf("Channel: %s", currCtx.Channel)
-		log.Printf("Version: %s\n", currVersion)
+	currCtxName, currCtx, err := utils.GetCurrentContext(tempContext)
```
anirudhjain75, line 48: Why are you still using a seperate function for getting current context ?... You should create / modify new function in config.go

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -175,30 +150,22 @@ func start() error {
...
-	// Get viper instance used for context
-	mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
-	if err != nil {
-		return errors.Wrap(err, "error processing config")
-	}
-
 	// get the platform, channel and the version of the current context
 	// if a temp context is set using the -c flag, use it as the current context
-	currCtxName, currCtx, err := utils.GetCurrentContext(tempContext)
+	_, currCtx, err := utils.GetCurrentContext(tempContext)
```
anirudhjain75, line 155: same here

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -175,30 +150,22 @@ func start() error {
...
 	switch currPlatform {
 	case "docker":
 
-		if _, err := os.Stat(utils.MesheryFolder); os.IsNotExist(err) {
-			// download the docker-compose.yaml file corresponding to the current version
-			if _, err := utils.DownloadDockerComposeFile(currCtx, true); err != nil {
-				return errors.Wrapf(err, utils.SystemError(fmt.Sprintf("failed to download %s file", utils.DockerComposeFile)))
-			}
+		// download the docker-compose.yaml file corresponding to the current version
+		if err := utils.DownloadDockerComposeFile(currCtx, true); err != nil {
```
anirudhjain75, line 167: here too

mesheryctl/pkg/utils/helpers.go
```diff
@@ -715,33 +715,41 @@ func GetCurrentContext(tempCtxName string) (string, config.Context, error) {
...
 		return "", config.Context{}, errors.New("context does not exist")
 	}
+	if currCtx.Version == "" {
+		// if no version is specified, pick latest imgaes based on channel
+		if currCtx.Channel == "edge" {
+			// if channel is edge, pick the edge-latest images
+			currCtx.Version = "latest"
+		} else if currCtx.Channel == "stable" {
+			// if the channel is stable, pick the version of the latest stable release
+			currCtx.Version, err = GetLatestStableReleaseTag()
```
anirudhjain75, line 725: this can be put inside compose.go as a function of struct

anirudhjain75 (changes requested):
Here are the comments I can make after looking into the changes... Do address these, @navendu-pottekkat @leecalcote do let me know if these suggestions are wrong. I'm trying to keep functionality of compose and context in their respective class objects.

anirudhjain75:
Thanks for refractoring @aneeshnema :).

navendu-pottekkat:
@anirudhjain75 I agree.

Maybe moving the `GetCurrentContext()` function inside `config.go` file and replacing the existing `GetContextContent()` function would work. Or we can make this a separate function in config.go

mesheryctl/internal/cli/root/system/reset.go
aneeshnema: moved `GetCurrentContext` to config.go

navendu-pottekkat (approved):
This looks good. We are fully "context" supported!

mesheryctl/internal/cli/root/config/config.go
```diff
@@ -88,3 +96,69 @@ func (v *Version) GetBuild() string {
...
+			}
+		} else {
+			return errors.Errorf("unknown channel %s", ctx.Channel)
+		}
+	}
+	return nil
+}
+
+// GetLatestStableReleaseTag fetches and returns the latest release tag from GitHub
+func GetLatestStableReleaseTag() (string, error) {
```
anirudhjain75, line 122: This should probably be a part of compose.go and part of the docker compose struct

mesheryctl/internal/cli/root/config/config.go
```diff
@@ -88,3 +96,69 @@ func (v *Version) GetBuild() string {
...
+	if err := json.Unmarshal(body, &dat); err != nil {
+		return "", errors.Wrap(err, "failed to unmarshal json into object")
+	}
+
+	return dat["tag_name"].(string), nil
+}
+
+// GetCurrentContext returns the current context name and context struct.
+// If the user mentions a temporary context(tempCtxName) with -c flag, change the current-context and proceed to temporary-context
+func GetCurrentContext(tempCtxName string) (string, Context, error) {
```
anirudhjain75, line 144: This should be a part of context struct

aneeshnema: It cannot be part of the context struct because that would be a paradox. The function is there to return context(i.e. the user using it does not have access to context)
it can be part of `MesheryCtlConfig`, but then we will have to create `mctlCfg` everywhere and handle its error. Then we will also have to call the member function and handle its error too. All this will be added everywhere where the only purpose is to get context.

mesheryctl/internal/cli/root/config/config.go
aneeshnema: `GetLatestStableReleaseTag` is used to get the latest release tag of meshery, which is used in creating the docker-compose file. This tag is not directly linked with docker-compose. In future, it will also be used for Kubernetes platform to fetch correct manifest files based on version.

Moving only `GetLatestStableReleaseTag` to `helpers.go` will create cyclic dependency.

anirudhjain75 (approved):

anirudhjain75:
LGTM

mesheryctl/internal/cli/root/system/update.go
```diff
@@ -40,13 +37,12 @@ var updateCmd = &cobra.Command{
 		return utils.PreReqCheck(cmd.Use, "")
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
-
-		if _, err := os.Stat(utils.DockerComposeFile); os.IsNotExist(err) {
-			if err := utils.DownloadFile(utils.DockerComposeFile, fileURL); err != nil {
-				return errors.Wrapf(err, utils.SystemError(fmt.Sprintf("failed to download %s file from %s", utils.DockerComposeFile, fileURL)))
-			}
+		userResponse := utils.AskForConfirmation("Meshery container images are going to be updated. Are you sure you want to continue")
```
leecalcote, line 40: This is desired behavior only if users are also notified that they are running an old version and asked to update. Otherwise, we have intentionally set the behavior here to autoupdate to ensure that users are always using the latest and greatest.

A subscription to a release channel is .... just that - a _subscription_. If they are on `stable-latest`, then they should receive _latest_. Unless the user has pinned their version in their meshconfig, autoupdates should be happening under their subscription.

leecalcote (changes requested):

mesheryctl/internal/cli/root/system/update.go
navendu-pottekkat: i.e for implementing this, 

if the user is on the stable-latest channel, if they run `mesheryctl system update`, it should just go and update without asking the user for confirmation.

If the user is on a specific version, then we should ask for confirmation.

Is that the case?

leecalcote: Perfect, yes.

leecalcote: We may get feedback in the future that asking the user to update when they have intentionally pinned to a specific version is not ideal. Let's wait for people to give that feedback, though.

navendu-pottekkat: Yep. @aneeshnema could you please check this out?

aneeshnema: fixed ✔

leecalcote (approved):
Good work!

PR #2430, opened 2021-02-21
controlplane queries

vineethvanga18:
Signed-off-by: vineethvanga18 <reddy.8@iitj.ac.in>

**Description**

**Notes for Reviewers**

### DIFF
--- a/ui/components/DashboardComponent.js
+++ b/ui/components/DashboardComponent.js
@@ -188,6 +188,145 @@ class DashboardComponent extends React.Component {
     return st;
   }
 
+  // using '/api/mesh/scan' as of now. To be removed after control plane resolvers are in place.
+  fetchMeshScanData = () => {
+    const self = this;
+    self.props.updateProgress({ showProgress: true });
+    dataFetch(
+      "/api/mesh/scan",
+      {
+        credentials: "same-origin",
+        method: "GET",
+        credentials: "include",
+      },
+      (result) => {
+        self.props.updateProgress({ showProgress: false });
+        if (result) {
+          // Extract all the unique namespaces in the mesh scan
+          const namespaces = {};
+          const activeNamespaces = {};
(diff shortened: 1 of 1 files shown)


leecalcote:
Thank you, Vineeth.

kumarabd:
@vineethvanga18 There is one that is already merged for this query. can you look that up once?

vineethvanga18:
@kumarabd I dont think there is one for this(control plane). Although there are queries written for addons and operator status.

ui/schema.graphql
```diff
@@ -0,0 +1,90 @@
+enum AddonSelector {
```
dhruv0000, line 1: Why do we need one more schema file here @vineethvanga18 ? 
`    "relay": "relay-compiler --src ./components --schema ../internal/graphql/schema/schema.graphqls" ` uses the schema in the server itself.

vineethvanga18: It was already present. Though you are right.

leecalcote: So... what do we want to do here?

vineethvanga18: Just delete this file and use what is there at 'internal/graphql/schema' directory

leecalcote: Call for a re-review when this is done, please.

dhruv0000:
image
@vineethvanga18 I think the resolves are in place, but the logic is still to be implemented. Maybe also use the http://localhost:9081/api/mesh/scan endpoint as an initial query for now?

vineethvanga18:
@leecalcote we will use the old end point 'api/mesh/scan' for now until the resolvers are written for control plane. Although the query and subscription are done.

dhruv0000 (approved):

leecalcote:
Ok, gents. Let's give this a go.

PR #2452, opened 2021-02-24
Improve provider capabilities loading

tangledbytes:

**Description**

This PR improves loading provider capabilities mechanism. In the previous approach, meshery server used to query all of the remote providers and download their extension packages before booting up. However this approach was limited to be able to load to only one type of capability per provider (i.e. a provider cannot change capabilities based on the current active user).

This PR replaces previous approach in the following ways:
1. It splits previous "initialization" process into 2 individual processes
    1. Loading provider capabilities
    2. Downloading the remote provider extension package
2. During bootup meshery server will only do 1.1, i.e will request only for provider's generic capabilities. This is important for meshery to be able to serve login page with appropriate provider info (like description, name, etc.)
3. At the end of user authentication flow, meshery will again query remote provider for capabilities (this time with the user token) so that a remote provider can choose to send either the same generic capabilies or it may send a user specific capabilities. Meshery server upon loading the fresh capabilities will then proceed to download the remote provider extension package (iff package doesn't already exists in the file system).

**Notes for Reviewers**

### DIFF
--- a/models/remote_provider.go
+++ b/models/remote_provider.go
@@ -69,9 +69,34 @@ type UserPref struct {
 // Initialize function will initialize the RemoteProvider instance with the metadata
 // fetched from the remote providers capabilities endpoint
 func (l *RemoteProvider) Initialize() {
-	// Get the capabilities
+	// Get the capabilities with no token
+	// assuming that this will help get basic info
+	// of the provider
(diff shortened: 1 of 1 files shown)


leecalcote (approved):
A thing of beauty.

PR #2454, opened 2021-02-24
[meshery] Extend meshops.proto to provide adapter version, and other info

### LINKED ISSUE #2449: [meshery] Extend meshops.proto to provide adapter version, and other info
#### Current Behavior

The adapter gRPC service does not provide a function to return the version  of an adapter, and other information of interest.

#### Desired Behavior

Create an function `getInfo` or similar to return adapter version and other information. 

This issue is related to https://github.com/layer5io/meshery/issues/1530, but handles only one aspect of it.

mgfeller:
**Description**

This PR fixes #2449

Proposal to extend meshops.proto to provide information about an adapter, as ComponentInfo. An adapter is one of many Meshery components, and exposing the same function/resource for all components (that is for instance ComponentInfo gRPC rpc or equivalent REST etc) will provide consistent and unified approach. 

Also included a function to return a list of supported mesh versions, so that we can handle 2 API changes in one go (if that's ok).

**Notes for Reviewers**

`pb.go` files created as well.


leecalcote (dismissed):
LGTM

ComponentHealth, OperationStatus... are other potential additions later.

leecalcote:
This looks good to me. It's related, but orthogonal to the messaging framework and notification center. Every MeshKit component should be able to pass along its details like version number. @kumarabd what is the sequence of steps needed to implement and release this (breaking?) change to the proto?

meshes/meshops.proto
```diff
@@ -7,11 +7,14 @@ package meshes;
 service MeshService {
     rpc CreateMeshInstance(CreateMeshInstanceRequest) returns (CreateMeshInstanceResponse) {}
     rpc MeshName(MeshNameRequest) returns (MeshNameResponse) {}
+    rpc MeshVersions(MeshVersionsRequest) returns (MeshVersionsResponse) {}
```
Aisuko, line 10: Should we merge Mesh information to just one gRPC service? Even though we need to bring the breaking change. And I'd like to say that we need to follow the OOP message struct that would be more standard and easy to extend, right?

kumarabd: MeshVersions is actually the versions of the service mesh available in the adapter, where as ComponentInfo gives the adapter version @Aisuko

kumarabd:
@mgfeller Will you be able to rebase this branch?

meshes/meshops.proto
Aisuko: OK, I see and thanks.

leecalcote:
@kumarabd what is the sequence of steps that we need to take to rollout this update?

mgfeller:
> @kumarabd what is the sequence of steps that we need to take to rollout this update?

@leecalcote @kumarabd imho: implement it in `meshery-adapter-library` in such a way that adapters can override the return values, but nothing breaks if they don't but still upgrade to the corresponding new version of the library.

PR #2492, opened 2021-03-02
reverted to graphql resolvers

kumarabd:
Signed-off-by: kumarabd <abishekkumar92@gmail.com>

**Description**

- Fixed operator flow
- bumped meshkit
- reverted to resolvers with MeshSync data
- Fixed mesheryCtl for backward compatibility

**Notes for Reviewers**


cmd/main.go
```diff
@@ -44,6 +46,15 @@ func main() {
 		models.GlobalTokenForAnonymousResults = globalTokenForAnonymousResults
 	}
 
+	// Initialize Logger instance
+	log, err := logger.New("meshery", logger.Options{
+		Format: logger.SyslogLogFormat,
+	})
+	if err != nil {
+		logrus.Error(err)
```
leecalcote, line 54: MeshKit is imported, new logger initialized, but logrus used for logging?

cmd/main.go
```diff
@@ -155,9 +168,9 @@ func main() {
...
 		GenericPersister:       dbHandler,
 		GraphqlHandler: graphql.New(graphql.Options{
-			DBHandler:        &dbHandler,
-			GetKubeClient:    helpers.NewKubeClientGenerator(viper.GetString("KUBECONFIG_FOLDER")),
-			GetDynamicClient: helpers.NewDynamicClientGenerator(viper.GetString("KUBECONFIG_FOLDER")),
+			Logger:     log,
+			DBHandler:  &dbHandler,
+			KubeClient: &kubeclient,
 		}),
 		GraphqlPlayground: graphql.NewPlayground(graphql.Options{
```
leecalcote, line 175: I still haven't seen documentation on why this is included; how it is protected.

internal/graphql/resolver/error.go
```diff
@@ -7,8 +7,10 @@ import (
 const (
 	errCode               = "test_code"
 	ErrInvalidRequestCode = "test_code"
+	ErrNilClientCode      = "test_code"
```
leecalcote, line 10: Have we made enough initial decisions on the messaging framework to start assigning error codes?

internal/graphql/resolver/meshsync.go
```diff
@@ -4,30 +4,33 @@ import (
...
 	mesherykube "github.com/layer5io/meshkit/utils/kubernetes"
-	"github.com/layer5io/meshsync/pkg/broker"
 	meshsyncmodel "github.com/layer5io/meshsync/pkg/model"
 )
 
 var (
-	meshsyncName = "meshsync"
-	meshsyncYaml = "https://raw.githubusercontent.com/layer5io/meshery-operator/master/config/samples/meshery_v1alpha1_meshsync.yaml"
+	meshsyncName    = "meshsync"
+	meshsyncSubject = "meshery.>"
```
leecalcote, line 16: Educate me, if you would: why the `.>`?

internal/graphql/resolver/meshsync.go
```diff
@@ -4,30 +4,33 @@ import (
...
+	meshsyncQueue   = "meshery"
+	meshsyncYaml    = "https://raw.githubusercontent.com/layer5io/meshery-operator/master/config/samples/meshery_v1alpha1_meshsync.yaml"
```
leecalcote, line 18: Hmm. If we don't externalize now, it'd be good to create an issue to externalize this later.

internal/graphql/resolver/operator.go
```diff
@@ -25,78 +28,85 @@ const (
...
-		err = runMeshSync(kubeClient, del)
+		err = runMeshSync(kubeclient, del)
 		if err != nil {
+			r.Log.Error(err)
 			r.operatorChannel <- &model.OperatorStatus{
 				Status: status,
 				Error: &model.Error{
-					Code:        errCode,
+					Code:        "",
 					Description: err.Error(),
```
leecalcote, line 81: "MeshSync failed to start"?

internal/graphql/resolver/addons.go
```diff
@@ -6,163 +6,114 @@ import (
 	"strings"
 
 	"github.com/layer5io/meshery/internal/graphql/model"
+	"github.com/layer5io/meshkit/utils"
```
leecalcote, line 9: I didn't notice this filename previously. Calling Prometheus and Grafana "add-ons" is a carry-over from Istio and not entirely how Meshery views such integrations. Meshery's use of either the of the two of these spans well past Istio - past any service mesh for that matter.

internal/graphql/resolver/schema.resolvers.go
```diff
@@ -46,28 +46,28 @@ func (r *queryResolver) GetOperatorStatus(ctx context.Context) (*model.OperatorS
...
 	if selector != nil {
-		return r.listenToAddonEvents(ctx)
+		return r.listenToAddonState(ctx)
 	}
 
 	return nil, ErrInvalidRequest
 }
 
-func (r *subscriptionResolver) ListenToControlPlaneEvents(ctx context.Context, filter *model.ControlPlaneFilter) (<-chan []*model.ControlPlane, error) {
+func (r *subscriptionResolver) ListenToControlPlaneState(ctx context.Context, filter *model.ControlPlaneFilter) (<-chan []*model.ControlPlane, error) {
```
leecalcote, line 57: Is there a meaningful difference between control plane state and add-on state? Why are the two tracked independently?

internal/graphql/schema/schema.graphqls
```diff
@@ -18,6 +18,7 @@ enum MeshType {
 enum Status {
 	ENABLED
 	DISABLED
+	PROCESSING
```
leecalcote, line 21: Will these states be used in the Messaging System; used universally across all operations performed by Meshery Adapters?

Why are these not centrally defined?

handlers/operator.go
```diff
@@ -159,7 +159,14 @@ func (h *Handler) subscribeToBroker(datach chan *broker.Message) error {
 	}
 
 	// subscribing to nats
-	natsClient, err := nats.New(broker.Status.Endpoint.External)
+	natsClient, err := nats.New(nats.Options{
+		URLS:           []string{broker.Status.Endpoint.External},
+		ConnectionName: "meshery",
+		Username:       "",
```
leecalcote, line 165: Where are these values stored?

cmd/main.go
kumarabd: Protection - They're done the usual way, how we do it for REST handlers.
What exactly to document in here @leecalcote ?
There are included here because resolver implementations have dependency on these.

internal/graphql/resolver/error.go
kumarabd: Not yet, we will need to start writing the script for autogeneration as an initial step. Before that we shud stick to the current practice.

internal/graphql/resolver/meshsync.go
kumarabd: yes, rather actually move it as a `config`. Meshery is the only repo currently without a config.

internal/graphql/resolver/meshsync.go
kumarabd: it means any subject that matches meshery.xxx or meshery.xxx.xxx.xx ... and so on. Currently MeshSync publishes to subject name - `meshery.meshsync.discovery`. There could be many more, just namespaced it for meshery in general.

internal/graphql/resolver/operator.go
kumarabd: Oh actually the same case as above, before starting with messaging system implementation, its better to have it blank. To avoid confusion. wt say? @leecalcote

internal/graphql/resolver/schema.resolvers.go
kumarabd: They are on different pages, thus different subscriptions (websocket. connections). so had to segregate them.

internal/graphql/schema/schema.graphqls
kumarabd: We cannot separate and import schemas from different sources. A graphql limitation of the relay plugin.

handlers/operator.go
kumarabd: No there are no creds atm.

leecalcote (approved):

leecalcote (changes requested):
A merge conflict to investigate...

PR #2549, opened 2021-03-12
Add support for performance profiles

tangledbytes:
**Description**

This PR adds Performance Profile feature in meshery. Following are the APIs added in the meshery server:

1. CRUD for Performance profiles
2. CRUD for schedules (this is not yet exposed in the UI)
3. CRUD for performance profiles test results

Following is a short demo of performance profle UI:
Kazam-screencast-00001

**Notes for Reviewers**
APIs for schedules are in place but they aren't yet exposed in the UI. So the "next run" column in the profiles table is missing (code is commented) and form for the same is also missing.

### DIFF
--- a/ui/components/MesheryPerformance/PerformanceCard.js
+++ b/ui/components/MesheryPerformance/PerformanceCard.js
@@ -13,7 +13,6 @@ function PerformanceCard({
   endpoints,
   loadGenerators,
   testRunDuration,
-  updatedAt,
   lastRun,
   reqHeaders,
   results,
--- a/ui/components/MesheryPerformance/PerformanceProfileGrid.js
+++ b/ui/components/MesheryPerformance/PerformanceProfileGrid.js
@@ -14,7 +14,6 @@ function PerformanceCardGridItem({ profile, deleteHandler, setProfileForModal })
       <PerformanceCard
         id={profile.id}
         name={profile.name}
-        updatedAt={profile.updated_at}
         endpoints={profile.endpoints}
         loadGenerators={profile.load_generators}
         reqHeaders={profile.request_headers}


cmd/main.go
```diff
@@ -163,12 +169,13 @@ func main() {
...
-		ResultPersister:        resultPersister,
-		SmiResultPersister:     smiResultPersister,
-		TestProfilesPersister:  testConfigPersister,
-		GenericPersister:       dbHandler,
+		ProviderBaseURL:              DefaultProviderURL,
+		MapPreferencePersister:       preferencePersister,
+		ResultPersister:              resultPersister,
+		SmiResultPersister:           smiResultPersister,
+		TestProfilesPersister:        testConfigPersister,
+		PerformanceProfilesPersister: performanceProfilePersister,
```
leecalcote, line 177: Educate me: why is performanceProfilePersister separate from resultPersister or testPersister? Does each persister represent a file on disk?

utkarsh-pro: @leecalcote, performanceProfilePersister only concerns itself with persisting the performance profiles while the resultPersister concerns itself with persisting only the performance results. However I am not too sure about the testConfigPersister which was already present in the meshery code base.

And yes, each persister gets its own file on the user's file system.

leecalcote: What factor most greatly influences the division between persisters? 

- logical grouping, which might mean that these two logically-related constructs should be persisted in the same data file. ￼

- Timing of when data needs to be written to disk. 


Considering that a test result only exists in context of a profile, should these two persisters be combined?￼

// @kumarabd

utkarsh-pro: @leecalcote profiles and results are still on different persister as all though results are associated with the profiles and it's existence is now only valid in context of a performance profile, they are two different entities from the implementation point of view.

There are still APIs that perform action on results in bulk (like date range queries or getting total results count) which benefit from keeping these two entities separate internally. The files in someway represent two different SQL DB tables, one which stores all of the results while holding a reference to the performance profile it belongs to.

Another reason to keep the persister is to let the older APIs work as intended as they rely on this persister. These older APIs may still be getting used in mesheryctl which is another meshery server client.

leecalcote: Thanks for this explanation @utkarsh-pro

leecalcote (approved):

leecalcote:
Ah, geez. I thought only two files had been updated. GH can be deceiving when you have already viewed other changes.

leecalcote:
Hope you were ready, @utkarsh-pro 😉

PR #2554, opened 2021-03-15
[meshery] Changes to meshery bash script for better user experience

### LINKED ISSUE #2547: Install meshery to a local binary directory
#### Current Behavior
Currently, the installation for meshery on Linux system is  `curl -L https://git.io/meshery | bash -`
This fails when the permission to write to /usr/local/bin is not given.

```bash
Installing mesheryctl in /usr/local/bin.
mv: inter-device move failed: meshery/mesheryctl' to '/usr/local/bin/mesheryctl'; unable to remove target: Permission denied
Permission Resolved: mesheryctl installed using sudo permissions.
chown: changing ownership of '/usr/local/bin/mesheryctl': Operation not permitted
Unable to write to /usr/local/bin/mesheryctl. Please verify write permission for fruffy.
```
For now, I just download the install file, edit it manually, and then run the installation.

#### Desired Behavior
It would be nice to have the option to install to a local directory. For example, `~/.local/bin/`. I am not sure if that is already possible with the current install script. 
Also if there is any file the current working directory that has the same name as one of the files that is being unpacked (for example, LICENSE, README.md), the installation will fail with a confusing error message. 

---
[Optional]

hexxdump:
**Description**
1. unzip command fails when the files are already present in PWD, use -o option to overwrite existing files
2. Ask user for install path to avoid failure in case of user does not have write permissions to /usr/local/bin

This PR fixes #2547

**Notes for Reviewers**
I assumed the expectation w.r.t. unzip is to overwrite with latest files(LICENCE, README.md and mesheryctl).

### DIFF
--- a/meshery
+++ b/meshery
@@ -58,17 +58,25 @@ printf "\nDownloading %s for %s...\n\n" "$NAME" "$OSEXT"
 curl -L ${URL} -o ${PWD}/meshery.zip
 
 printf "\nExtracting %s...\n" "$NAME"
-unzip ${PWD}/meshery.zip 
+unzip -o ${PWD}/meshery.zip
 
-printf "\nInstalling mesheryctl in /usr/local/bin.\n"
+# Use user provided installation path from env variable(MESHERY_INSTALL_DIR)
+# by default its /usr/local/bin
+install_dir=${MESHERY_INSTALL_DIR}
+if [ -z "$install_dir" ]
+then
+    install_dir="/usr/local/bin"
+fi
+
+printf "\nInstalling mesheryctl in $install_dir.\n"
 WHOAMI=$(whoami)
-if mv ${PWD}/mesheryctl /usr/local/bin/mesheryctl ; then		
+if mv ${PWD}/mesheryctl $install_dir/mesheryctl ; then
   echo "mesheryctl installed."
 else
-  if sudo mv ${PWD}/mesheryctl /usr/local/bin/mesheryctl ; then
+  if sudo mv ${PWD}/mesheryctl $install_dir/mesheryctl ; then
     echo "Permission Resolved: mesheryctl installed using sudo permissions."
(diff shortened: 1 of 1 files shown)


navendu-pottekkat (commented):
Looks great. 

I guess unzipping should overwrite the files. So this would be the correct way.

leecalcote (commented):
@hexxdump thank you for this. The `-o` option is great. We're good to merge that portion.

The other portion about prompting the user interrupts the installation flow and breaks our one-step to managing service meshes promise. I'm not sure that we want to satisfy the original feature request here. 

If we do, we can provide an option to override the default install location, which will satisfies the original feature request and also keep sane defaults and a quick installation process for the other 90% of our users.

I should have been clear in my response on the original feature request. I was trying to be positive and encouraging of the user providing feedback, but in doing so, I also insinuated that their suggestion should be executed verbatim (it shouldn't).

hexxdump:
> @hexxdump thank you for this. The `-o` option is great. We're good to merge that portion.
> 
> The other portion about prompting the user interrupts the installation flow and breaks our one-step to managing service meshes promise. I'm not sure that we want to satisfy the original feature request here.
> 
> If we do, we can provide an option to override the default install location, which will satisfies the original feature request and also keep sane defaults and a quick installation process for the other 90% of our users.
> 
> I should have been clear in my response on the original feature request. I was trying to be positive and encouraging of the user providing feedback, but in doing so, I also insinuated that their suggestion should be executed verbatim (it shouldn't).

@leecalcote, Introduced an environment variable MESHERY_INSTALL_DIR, which users can export before running the meshery script, so that the 90% users need not worry about the interactive prompt.

leecalcote:
Excellent. 😄 This looks great, @hexxdump.

leecalcote (approved):

PR #2560, opened 2021-03-16
[mesheryctl] system reset platform(kubernetes) aware, refactoring

### LINKED ISSUE #2310: [mesheryctl] [child] make `system reset` platform aware(`platform: kubernetes`)
See Epic #2307 for context and further description of this issue.

**Current Behavior**

`mesheryctl system reset` will pull the context specific `docker-compose.yaml` files and does not consider for the current platform. Only the channel and version are considered.

**Desired Behavior**

`mesheryctl system reset` should also consider for the `platform`. Specifically, if the platform is Kubernetes, retrieve the specific Kubernetes manifest files as it currently does for `platform: docker`.

aneeshnema:
**Description**

This PR fixes #2310 
`system reset` now supports the Kubernetes platform. The reset command will also fetch manifest files based on the version specified in the current context.

`GetManifestTreeURL` will return the tree link to the `install/deployment_yamls/k8s` folder with its state representing the version, i.e., the tree link will contain the names of only those files that were there at the release of the specified version.
Later the `resetMesheryConfig` function passes the above list of files to download and the root raw github folder URL with the branch as specified by version to the `utils.DownloadManifests` for download.

Update:
Moved platform related functions to `platform.go`
Cleaned up some code(as picked up by golang static analysis)


mesheryctl/internal/cli/root/system/reset.go
```diff
@@ -58,16 +62,73 @@ func resetMesheryConfig() error {
...
+				version, err = utils.GetLatestStableReleaseTag()
+				if err != nil {
+					return err
+				}
+			}
+		}
+
+		log.Debug("fetching required Kubernetes manifest files...")
+		// get correct minfestsURL based on version
+		manifestsURL, err := utils.GetManifestTreeURL(version)
```
navendu-pottekkat, line 96: Could we add this to the start command too? I think by current logic, it picks up only the latest version.

leecalcote:
@aneeshnema! Looking good here!

mesheryctl/internal/cli/root/system/reset.go
aneeshnema: Yes, the start command picks up only the latest version.
Should I update this PR or do it in a new one?

navendu-pottekkat: You can update this PR to add that and mention it in the description.

leecalcote: In part, the question here becomes that of centralizing this functionality.

The helper script(s) have become too bloated and need breaking out into different files.

navendu-pottekkat: @leecalcote I agree. There are many platform/ component specific functions that are in the helpers.go file and it is difficult to maintain.

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -118,7 +118,7 @@ func ApplyManifestFiles(manifestArr []utils.Manifest, requestedAdapters []string
 		adapterDeployment := adapterFile + "-deployment.yaml"
 		adapterService := adapterFile + "-service.yaml"
 
-		if utils.IsAdapterValid(manifestArr, adapter+"-deployment.yaml") == false {
+		if !utils.IsAdapterValid(manifestArr, adapter+"-deployment.yaml") {
```
navendu-pottekkat, line 121: Neat 👍🏼 

IDK why I felt `== false` would be more readable. Maybe because I always overlook the `!` and I got into a habit of making it obvious by adding the `==`

navendu-pottekkat (approved):
Looks good @aneeshnema . Thanks for refactoring things and tidying things up.

@leecalcote could you please merge this PR?

leecalcote (approved):
Very good. 😄

PR #2583, opened 2021-03-22
Added version check for config.yaml

### LINKED ISSUE #2294: [mesheryctl] Bug: version in config.yaml file is not validated
#### Description

You can have invalid versions as shown below(`version: bob`) in the `config.yaml` file. 

```
user@computer:~/.meshery$ cat config.yaml
contexts:
  local:
    token:
      name: Default
      location: auth.json
    platform: docker
    adapters:
    - meshery-istio
    - meshery-linkerd
    - meshery-consul
    - meshery-octarine
    - meshery-nsm
    - meshery-kuma
    - meshery-cpx
    - meshery-osm
    - meshery-nginx-sm
    channel: stable
    version: bob

current-context: local
tokens: {}
```

This can break commands like `mesheryctl system reset` which will try to pull the `docker-compose.yaml` file from the specified `version` in GitHub-

```
user@computer:$ mesheryctl system reset -c local
Meshery config file will be reset to system defaults. Are you sure you want to continue [y/n]? y
Meshery resetting...

Current Context: local
Channel: stable
Version: bob

Fetching default docker-compose file at version: bob...

...Meshery config (/home/user/.meshery/meshery.yaml) now reset to default settings.
user@computer:$ 
```

That version does not exist and [GitHub throws a 404 error](https://raw.githubusercontent.com/layer5io/meshery/bob/docker-compose.yaml). Which is missed by our current error handling function and the meshery.yaml file will be like this-

Screenshot from 2021-01-25 14-37-09

This is because it actually tries to pull the file https://raw.githubusercontent.com/layer5io/meshery/bob/docker-compose.yaml and does not check if the version is invalid. 

This error needs to be handled. For ex, here it should throw an error saying the version is invalid and should **not** try to reset with an invalid version.

#### Possible fixes

A quick fix would be to check if the response body is `404: Not Found`. 

But, a better and robust fix would be to have a central validator with the `MesheryCtlConfig` struct in the `config.go` file. This would be helpful if we require version validation for other commands.

The validation could be done using Regex.

metonymic-smokey:

**Description**
Checks the validity of the current context's version, using a regex.

This PR fixes #2294 

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/internal/cli/root/system/logs.go
+++ b/mesheryctl/internal/cli/root/system/logs.go
@@ -16,16 +16,26 @@ package system
 
 import (
 	"bufio"
+	"bytes"
+	"context"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
+	"strings"
 
 	"github.com/pkg/errors"
 
+	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/config"
 	"github.com/layer5io/meshery/mesheryctl/pkg/utils"
 
+	meshkitkube "github.com/layer5io/meshkit/utils/kubernetes"
 	log "github.com/sirupsen/logrus"
+	apiCorev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 // logsCmd represents the logs command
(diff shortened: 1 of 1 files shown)


mesheryctl/internal/cli/root/config/config.go
```diff
@@ -53,15 +54,44 @@ func (mc *MesheryCtlConfig) CheckIfCurrentContextIsValid() (Context, error) {
...
+		fmt.Println("The version for this context is: ", ctx.Version)
+
 		return ctx, nil
 	}
+
 	return Context{}, errors.New("current context:" + mc.CurrentContext + " does not exist")
 }
 
+// CheckIfCurrentVersionIsValid checks if current version is valid, return context as a string
+func CheckIfCurrentVersionIsValid(version string) (string, error) {
```
aneeshnema, line 77: I think we don't need to return the version again, just an `error` would do
We can make this a member function of the `Context` struct

aneeshnema (changes requested):
@metonymic-smokey the CI is failing because of lint errors.

metonymic-smokey:
@aneeshnema  thanks!
I've made the necessary changes, let me know if this is okay.

mesheryctl/internal/cli/root/config/config.go
```diff
@@ -74,22 +77,22 @@ func (mc *MesheryCtlConfig) CheckIfCurrentContextIsValid() (Context, error) {
 }
 
 // CheckIfCurrentVersionIsValid checks if current version is valid, return context as a string
-func CheckIfCurrentVersionIsValid(version string) (string, error) {
-
+func (mc *MesheryCtlConfig) CheckIfCurrentVersionIsValid() error {
```
aneeshnema, line 80:
```suggestion
func (ctx *Context) ValidateVersion() error {
```

@metonymic-smokey I think you misread my previous suggestion, I meant `Context` struct not `MesheryCtlConfig` struct

mesheryctl/internal/cli/root/config/config.go
```diff
@@ -60,11 +62,12 @@ func (mc *MesheryCtlConfig) CheckIfCurrentContextIsValid() (Context, error) {
 			ctx.Version = "latest"
 		}
```
aneeshnema, line 63: you can also move this assignment of the version to "latest" if it's empty to that member function of `Context`

mesheryctl/internal/cli/root/config/config.go
metonymic-smokey: right, totally my bad. So sorry!

mesheryctl/internal/cli/root/config/config.go
```diff
@@ -58,28 +56,26 @@ func (mc *MesheryCtlConfig) CheckIfCurrentContextIsValid() (Context, error) {
...
 
 	return Context{}, errors.New("current context:" + mc.CurrentContext + " does not exist")
 }
 
-// CheckIfCurrentVersionIsValid checks if current version is valid, return context as a string
-func (mc *MesheryCtlConfig) CheckIfCurrentVersionIsValid() error {
-	version := mc.Contexts[mc.CurrentContext].Version
-	if version == "" || version == "latest" {
+// CheckIfCurrentVersionIsValid checks if current version is valid, returns error accordingly
+func (ctx *Context) CheckIfCurrentVersionIsValid() error {
```
aneeshnema, line 71:
```suggestion
// ValidateVersion checks if the version is valid, if empty sets it to default value latest. Returns an error if the version is invalid.
func (ctx *Context) ValidateVersion() error {
```

mesheryctl/internal/cli/root/config/config.go
```diff
@@ -58,28 +56,26 @@ func (mc *MesheryCtlConfig) CheckIfCurrentContextIsValid() (Context, error) {
 	}
 
 	if ctx, exists := mc.Contexts[mc.CurrentContext]; exists {
-		if ctx.Version == "" {
-			ctx.Version = "latest"
-		}
-
-		err := mc.CheckIfCurrentVersionIsValid()
+		err := ctx.CheckIfCurrentVersionIsValid()
```
aneeshnema, line 59:
```suggestion
		err := ctx.ValidateVersion()
```

mesheryctl/internal/cli/root/config/config.go
```diff
@@ -58,28 +56,26 @@ func (mc *MesheryCtlConfig) CheckIfCurrentContextIsValid() (Context, error) {
...
 	return Context{}, errors.New("current context:" + mc.CurrentContext + " does not exist")
 }
 
-// CheckIfCurrentVersionIsValid checks if current version is valid, return context as a string
-func (mc *MesheryCtlConfig) CheckIfCurrentVersionIsValid() error {
-	version := mc.Contexts[mc.CurrentContext].Version
-	if version == "" || version == "latest" {
+// CheckIfCurrentVersionIsValid checks if current version is valid, returns error accordingly
+func (ctx *Context) CheckIfCurrentVersionIsValid() error {
+	version := ctx.Version
```
aneeshnema, line 72:
```suggestion
```

we don't need another variable, can reuse `ctx.Version`

aneeshnema (changes requested):
`CheckIfCurrentVersionIsValid` does not suit as correct function name as the version is singular for a given context. (there are not multiple versions in a Context to choose current version from)

We named `CheckIfCurrentContextIsValid` that way it is because config can contain multiple contexts as defined by the user. But when executing mesheryctl commands, these commands only work on a specific context, i.e, `current-context` defined in config or as passed by user with `-c` flag. Hence comes the word _Current_.

aneeshnema (dismissed):
LGMT 💯

leecalcote:
Very nice teamwork, @metonymic-smokey and @aneeshnema. 😎

leecalcote (approved):

mesheryctl/internal/cli/root/config/config.go
```diff
@@ -53,15 +54,42 @@ func (mc *MesheryCtlConfig) CheckIfCurrentContextIsValid() (Context, error) {
 	if mc.CurrentContext == "" {
 		return Context{}, errors.New("current context not set")
 	}
+
 	if ctx, exists := mc.Contexts[mc.CurrentContext]; exists {
-		if ctx.Version == "" {
-			ctx.Version = "latest"
+		err := ctx.ValidateVersion()
+		if err != nil {
+			return Context{}, errors.New("invalid version " + ctx.Version + " specified")
```
navendu-pottekkat, line 61:
```suggestion
			return Context{}, err
```

Can't we just propagate back the error? @metonymic-smokey @aneeshnema

navendu-pottekkat (dismissed):
Looks good! Nice work @metonymic-smokey @aneeshnema . 

I have left one comment could you please check that out?

The rest looks good.

mesheryctl/internal/cli/root/config/config.go
metonymic-smokey: @navendu-pottekkat isn't specifying with an error message clearer?

This is what I received when I tried this change:
image

aneeshnema: Yes, specifying with an error message is clearer but we are already returning the same error message from the ValidateVersion function. So we don't need to print out the same error twice.

I don't think @navendu-pottekkat' s suggestion is causing the above error, can you please check again? (your validate version is accepting bob as a valid version)

metonymic-smokey: @aneeshnema 
If I just return `err`, it is always `<nil>` irrespective of whether the string matches the regex or not(the boolean `matched` however, reflects correctly). If I return a string using `errors.New`, I'm avoiding that bug.  

I ran the `ValidateVersion` function separately to observe it:
image

image

If we want to avoid double error messages, maybe I can change the return type to boolean returning `matched` and modify accordingly?

navendu-pottekkat: @metonymic-smokey That should not be the case, if the version is not valid, it would return an error in line 59. Instead of again creating a new error, we can just pass back the error which is exactly the same. This would avoid making two instances of the same error.

navendu-pottekkat: I checked this out and it works as expected. We can just pass the error instead of creating a new error.

navendu-pottekkat (commented):
@metonymic-smokey The changes you made look good. Small suggestion to pass back the `err` instead of creating a new `error`.

Could you also modify the `SetCurrentContext` and `CheckIfCurrentContextIsValid` function so that if the version is wrong, it does not throw "Context does not exist" error?

Screenshot from 2021-03-24 21-31-55

This should be a quick fix and it would make the error more readable and obvious.

mesheryctl/internal/cli/root/config/config.go
```diff
@@ -53,13 +54,43 @@ func (mc *MesheryCtlConfig) CheckIfCurrentContextIsValid() (Context, error) {
...
+	matched, err := regexp.MatchString(`v[0-9]+.[0-9]+.[0-9]+[-a-z0-9]*`, ctx.Version)
+	if err != nil || !matched {
+		return errors.New("invalid version " + ctx.Version + " specified")
+	}
+
+	if matched {
+		return nil
+	}
+
+	return errors.New("invalid version " + ctx.Version + " specified")
```
navendu-pottekkat, line 93: This errors are propagated to where the `ValidateVersion` is called right? Which is line 59. So in line 62 we can just pass the err instead of a new error. This would not change any output. But it would be cleaner.

navendu-pottekkat (commented):
@metonymic-smokey  The refactored error handling looks good. Could you consider the other change with creating the new error that I mention in the review below?

mesheryctl/internal/cli/root/config/config.go
metonymic-smokey: @navendu-pottekkat understood!
Just check these changes.
Thanks!

navendu-pottekkat (approved):
Thank you @metonymic-smokey !

PR #2599, opened 2021-03-25
Move from Bitcask to Sqlite

tangledbytes:
**Description**

This PR:
1. Moves the performance profiles and meshery results persistence handler from bitcask to Sqlite
2. Fixes the total results and total profiles text size in the dashboard

Relevant screen shots:

Screenshot_20210325_222224

**Notes for Reviewers**

### DIFF
--- a/models/meshery_results_persister.go
+++ b/models/meshery_results_persister.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/gofrs/uuid"
 	"github.com/layer5io/meshkit/database"
+	"github.com/sirupsen/logrus"
 )
 
 type MesheryResultsPersister struct {
@@ -86,7 +87,9 @@ func (mrp *MesheryResultsPersister) GetResult(key uuid.UUID) (*MesheryResult, er
 
 func (mrp *MesheryResultsPersister) WriteResult(key uuid.UUID, result []byte) error {
 	var data MesheryResult
-	json.Unmarshal(result, &data)
+	if err := json.Unmarshal(result, &data); err != nil {
+		return err
+	}
 
 	t := time.Now()
 	data.TestStartTime = &t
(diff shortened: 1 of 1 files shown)


leecalcote:
The user prefs still exist in bitcask and nothing else, is that right? // @kumarabd

kumarabd:
> The user prefs still exist in bitcask and nothing else, is that right? // @kumarabd

yes correct

models/performance_profile_persister.go
```diff
@@ -0,0 +1,94 @@
...
+	profiles := []*PerformanceProfile{}
+
+	query := ppp.DB.
+		Select(`
+		id, name, load_generators, 
+		endpoints, qps, service_mesh, 
+		duration, request_headers, request_cookies, 
+		request_body, content_type, created_at,
+		updated_at, (?) as last_run, (?) as total_results`,
+			ppp.DB.Table("meshery_results").Select("MAX(meshery_results.test_start_time)").Where("performance_profile = performance_profiles.id"),
```
kumarabd, line 41: @utkarsh-pro The subquery works then?

kumarabd (approved):

models/performance_profile_persister.go
utkarsh-pro: @kumarabd to some extent it does but we lose the type conversion for some reason. That's why I had to change the type to `string` rather than `time.Time`.

But yeah, it finally did :tada:

leecalcote:
Fantastic. 😄

models/performance_profile_persister.go
kumarabd: yes a custom type needs to be defined for time.Time as well. Basically a separate valuer/scanner interface

PR #2643, opened 2021-04-01
[docs] add `minikube tunnel` to minikube installation docs

pottekkat:

**Description**

To run Meshery locally on minikube, we also need to run minikube tunnel to expose the services to the host machine-

Screenshot from 2021-04-01 11-06-14

### DIFF
--- a/docs/pages/installation/minikube.md
+++ b/docs/pages/installation/minikube.md
@@ -109,7 +109,8 @@ To establish connectivity between a particular Meshery Adapter and Kubernetes se
  </div></div>
  </pre>
 <br/>
- Run the command below to expose the services to the host machine:
+
+ **Optionally run the command below to expose the LoadBalancer services to the host machine** :
   
  <pre class="codeblock-pre"><div class="codeblock">
  <div class="clipboardjs">


leecalcote:
This one has me scratching my head a little. I might be missing the need for this. @kumarabd?

kumarabd:
@leecalcote  it's not a necessity. Running it will expose the loadbalancer type service endpoints. Not doing so will expose nodeport endpoints, it's the same for a user.

leecalcote:
Thanks. This optional step needs clarified as such.

docs/pages/installation/minikube.md
```diff
@@ -108,6 +108,15 @@ To establish connectivity between a particular Meshery Adapter and Kubernetes se
  mesheryctl system config minikube -t ~/Downloads/auth.json
  </div></div>
  </pre>
+<br/>
```
leecalcote, line 111: This optional, but helpful step should be clearly marked as such and presented in such a way that the lazy reader doesn't assume it is a mandatory step.

leecalcote (changes requested):

navendu-pottekkat:
@leecalcote I have mentioned the optional use case. I think this would be helpful 🤔

leecalcote (approved):
SGTM

PR #2650, opened 2021-04-02
[Security] Static analysis using CodeQL

leecalcote:
**Description**

This PR adds code analysis for potential security issues.

**Notes for Reviewers**
The intended initial focus of this analysis is that of the golang, but could also include the javascript.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,7 +13,7 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ * ]
+    branches: [ ** ]
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ master ]


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
```
mgfeller, line 16: Wouldn't we want to run this on all branches, i.e. also in forks?

leecalcote: I think that's right. Although, when a PR is open from a fork to push to master, this analysis will run. 

Curious, are actions (workflows) currently running in your forks?

leecalcote: Checking on this, I think the workflows are executing your meshery fork today and that it might be a similar configuration to invoke on push to master is running in the remote fork (here's an example - https://github.com/layer5io/meshery/blob/master/.github/workflows/build-and-release.yml#L7).

Yes, having this build on any branch sounds good. My ignorance gets in the way of being able to intelligible supporting statements to this suggestion as well as intelligible detractors. @ramrodo, thoughts?

Nikhil-Ladha: Yes, when we update our fork from the latest code of upstream. The actions do run on our forks as well.

mgfeller: Here's another example: https://github.com/layer5io/meshery/blob/074493c57e07f8f98fa11664f1873c6264b96889/.github/workflows/build-and-release.yml#L7 
It runs on all pushes to all branches in all forks (e.g. https://github.com/mgfeller/meshery-consul/actions), also upstream of course. The advantage is that it runs before submitting a PR, allowing to catch issues early.

leecalcote: That makes much sense. 👍

leecalcote: Updated to include * branches on the push event. I can't muster the brainpower to determine whether `master` remains appropriate for pull_request events or if this needs expanded, too. The comment from CodeQL says that branches for pull_request events should be a subset, so it's not clear to me. I've little to no opinion here (on the PR at-large). @mgfeller, please feel at leisure to take this over and get us going in the best direction.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ * ]
```
mgfeller, line 16: @leecalcote it should be `**`, as `*` does not include branches with  `/` in their name.
if this is changed, we're good to go :+1: 

`master` in `pull_request` is fine, it is a subset and afaik we don't create PRs to other branches.
(can't seem to edit this file or push to this branch, otherwise I would have)

mgfeller (changes requested):

.github/workflows/codeql-analysis.yml
leecalcote: @mgfeller I think this was the last excuse we needed to graduate you from Meshery Adapter Maintainer to Meshery Maintainer. 💪

leecalcote, line 16:
```suggestion
    branches: [ ** ]
```

leecalcote: Second `*` added. Let's see how she works...

mgfeller: > @mgfeller I think this was the last excuse we needed to graduate you from Meshery Adapter Maintainer to Meshery Maintainer. muscle

Cool :smile: Graduation party?! :tada:

PR #2664, opened 2021-04-04
Add customize way to choice deploy the meshery adapters

### LINKED ISSUE #2600: Helm installation for individual adapters
#### Implementation
If a user insists on running meshery inside the Kubernetes cluster using helm charts, currently all of the adapters get deployed by default. We would need a provision in the helm `values.yml` to specify the list of adapters to run before installing inside the cluster.

Ref: https://github.com/layer5io/meshery/tree/master/install/kubernetes/helm/meshery

#### Mockups

---

Aisuko:
Signed-off-by: aisuko <urakiny@gmail.com>

**Description**

This PR fixes #2600 

**Notes for Reviewers**

### DIFF
--- a/docs/pages/installation/kubernetes.md
+++ b/docs/pages/installation/kubernetes.md
@@ -47,9 +47,9 @@ Deploy Meshery to your Kubernetes cluster by executing:
 </pre>
 
 
-### **Using Helm**
+### **Using Helm Charts**
 
-Run the following:
+Run the following for default:
 
  <pre class="codeblock-pre">
  <div class="codeblock"><div class="clipboardjs">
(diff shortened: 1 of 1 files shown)


docs/pages/installation/kubernetes.md
```diff
@@ -58,6 +58,21 @@ Run the following:
...
  </pre>
 
+Customize of deployment the Meshery adapters:
+
+ <pre class="codeblock-pre">
+ <div class="codeblock"><div class="clipboardjs">
+ $ helm repo add meshery https://meshery.io/charts/
+
+ # Manually disable the adapters you do not want to deploy
+ $ helm install --set key=value --set key=value my-meshery meshery/meshery --version 2.1.2
```
leecalcote, line 68: Thanks for this, @Aisuko. For as much as we can, we encourage Meshery deployments to subscribe to either the `stable-latest` or `edge-latest` release channels. Is it possible to have `stable-latest` as the default version in the Helm chart?

Aisuko: Did you mean the Docker image of Meshery, I believe we have https://github.com/layer5io/meshery/blob/ecceb1a296c48205c3b8759bebe27fd7d58f1d11/install/kubernetes/helm/meshery/values.yaml#L9

And you mean that we can support `helm install` support set `statble` and `edge` ?

leecalcote: @Aisuko, thanks for clarifying. Having reflected on this a little bit more (in today's Meshery Development meeting), the suggestion coming out of the meeting was to create two independent versions of the chart, resulting in two entries for Meshery on Artifact Hub: stable and edge (see Linkerd project as an example).

The other consensus formed was that of consolidating the Helm chart version to be the same as that of Meshery version that releases it, so that there's no confusion for users between what Meshery version they are running and what Helm chart version they are running: both would be the same. Can we move to this going forward?

Do both of these make sense / sound good?

Aisuko: Thank you told me this, sir. That sounds good. I believe that makes sense. I will do PR for this, but this PR only includes en/disable meshery adapters,we can merge at first.

leecalcote: Ok, sounds real good. 👍

PR #2705, opened 2021-04-11
Add API for importing pattern files from github

tangledbytes:
**Description**

This PR adds API for importing pattern files from github. The support has been added to both remote provider as well as local provider, however remote provider (like meshery-cloud) may be able to support both public as well as private repositories but local provider will be able to support only public repositories.

API endpoint added:
1. `GET` - `/patterns/import/github.com/:owner/:repo?path=path/to/the/pattern/file` 

**Notes for Reviewers**
The UI for remote imports needs some discussion. Hence the PR only includes the APIs.


leecalcote:
Whoa!

And here, I meant that it would be nice to review the design spec for this feature this weekend. 😉

leecalcote:
Whoa!

And here, I meant that it would be nice to review the design spec for this feature this weekend. 😉

leecalcote (approved):

PR #2724, opened 2021-04-18
Remote Pattern Imports

tangledbytes:
**Description**

This PR adds the following:
- Features:
  1. Generic HTTP pattern import
  2. Recursive Github import (A generic Github Walker has been added)
  3. More efficient format conversion
  4. Remote imports can now be combined directly with format change endpoint by specifying `?output=cytoscape` in the endpoint.
  5. `Location` of imported pattern is also now tracked
- Others:
  1. All of the imports, exports and manual uploads APIs have been consolidated in this PR. This means that the following components **may break**.
      1. mesheryctly commands which relied on the older APIs // @Kn99HN @navendu-pottekkat 
      2. meshmap wherever it leverages exports APIs // @dhruv0000 
  2. Updates `mesheryctl` pattern commands endpoints
  3. Meshery Pattern UI reflects the changes made in this PR, so the UI should not break.

For more details regarding the changes please refer the [design spec](https://docs.google.com/document/d/1B2N78EdRiZF-yVo1-HY3syppwBBDumgMuYg6seD-AJ4) which this PR conforms to.

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/internal/cli/root/pattern/apply.go
+++ b/mesheryctl/internal/cli/root/pattern/apply.go
@@ -32,7 +32,7 @@ var applyCmd = &cobra.Command{
 		}
 
 		client := &http.Client{}
-		req, err := http.NewRequest("POST", mctlCfg.GetBaseMesheryURL()+"/api/experimental/patternfile/deploy", fileReader)
+		req, err := http.NewRequest("POST", mctlCfg.GetBaseMesheryURL()+"/api/experimental/pattern/deploy", fileReader)
 		if err != nil {
 			return err
 		}
--- a/mesheryctl/internal/cli/root/pattern/delete.go
+++ b/mesheryctl/internal/cli/root/pattern/delete.go
@@ -32,7 +32,7 @@ var deleteCmd = &cobra.Command{
 		}
 
 		client := &http.Client{}
-		req, err := http.NewRequest("DELETE", mctlCfg.GetBaseMesheryURL()+"/api/experimental/patternfile/deploy", fileReader)
+		req, err := http.NewRequest("DELETE", mctlCfg.GetBaseMesheryURL()+"/api/experimental/pattern/deploy", fileReader)
 		if err != nil {
 			return err
 		}
--- a/mesheryctl/internal/cli/root/pattern/view.go
+++ b/mesheryctl/internal/cli/root/pattern/view.go
@@ -47,16 +47,16 @@ var viewCmd = &cobra.Command{
 		url := mctlCfg.GetBaseMesheryURL()
 		if len(pattern) == 0 {
 			if viewAllFlag {
-				url += "/api/experimental/patternfile?page_size=10000"
+				url += "/api/experimental/pattern?page_size=10000"
 			} else {
 				return errors.New("[pattern-name|pattern-id] not specified, use -a to view all patterns")
 			}
 		} else if isID {
 			// if pattern is a valid uuid, then directly fetch the pattern
-			url += "/api/experimental/patternfile/" + pattern
+			url += "/api/experimental/pattern/" + pattern
 		} else {
 			// else search pattern by name
-			url += "/api/experimental/patternfile?search=" + pattern
+			url += "/api/experimental/pattern?search=" + pattern
 		}
 
 		client := &http.Client{}
--- a/router/server.go
+++ b/router/server.go
@@ -118,15 +118,13 @@ func NewRouter(ctx context.Context, h models.HandlerInterface, port int) *Router
 
 	gMux.Handle("/api/promGrafana/scan", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.ScanPromGrafanaHandler))))
 
-	gMux.Handle("/api/experimental/patternfile/deploy", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.PatternFileHandler)))).
+	gMux.Handle("/api/experimental/pattern/deploy", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.PatternFileHandler)))).
 		Methods("POST", "DELETE")
-	gMux.Handle("/api/experimental/patternfile", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.PatternFileRequestHandler)))).
-		Methods("POST", "GET")
 	gMux.Handle("/api/experimental/pattern", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.PatternFileRequestHandler)))).
-		Methods("POST")
-	gMux.Handle("/api/experimental/patternfile/{id}", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.GetMesheryPatternHandler)))).
+		Methods("POST", "GET")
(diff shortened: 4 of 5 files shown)


models/meshery_pattern.go
```diff
@@ -16,6 +19,26 @@ type MesheryPattern struct {
...
 	UpdatedAt *time.Time `json:"updated_at,omitempty"`
 	CreatedAt *time.Time `json:"created_at,omitempty"`
 }
+
+// GetPatternName takes in a stringified patternfile and extracts the name from it
+func GetPatternName(stringifiedFile string) (name string, err error) {
+	out := map[string]interface{}{}
+
+	if err = yaml.Unmarshal([]byte(stringifiedFile), &out); err != nil {
+		return
```
subhamkrai, line 33:
```suggestion
		return " ",err
```

models/meshery_pattern.go
```diff
@@ -16,6 +19,26 @@ type MesheryPattern struct {
...
+	}
+
+	// Get Name from the file
+	name, ok := out["name"].(string)
+	if !ok {
+		err = fmt.Errorf("invalid patternfile - name field is either not present or is not valid")
+		return
```
subhamkrai, line 40:
```suggestion
		return " ", err
```

models/meshery_pattern.go
```diff
@@ -16,6 +19,26 @@ type MesheryPattern struct {
...
+	}
+
+	return
```
subhamkrai, line 43:
```suggestion
	return name, nil
```

subhamkrai (changes requested):

handlers/meshery_pattern_handler.go
```diff
@@ -67,6 +67,24 @@ func (h *Handler) ImportPatternFileGithub(
 	fmt.Fprintf(rw, "%s", cont)
 }
 
+// ImportPatternFileGithub will import patternfile file from github
```
subhamkrai, line 70:
```suggestion
// ImportPatternFileHTTP will import patternfile file from HTTP endpoint
```

subhamkrai (changes requested):

models/meshery_pattern.go
utkarsh-pro: Aren't both the same? Or the suggestion is from code readability point of view? :thinking:

handlers/meshery_pattern_handler.go
utkarsh-pro: Thanks @subhamkrai! :smile:

models/meshery_pattern.go
subhamkrai: they are same but I believe naked return is good in case of short programs not in big projects and also it better code readability(which will help other contributors) which I think we should follow. @leecalcote  any thoughts?
Thanks

utkarsh-pro: Ah, I see. On the contrary, I believe that the functions which are public facing, (i.e are not in `internal` directory or not being exported) and are short in their size (so that chances of shadowing are minimal) then named returns helps in "self-documenting". 

However, I do agree that this maybe less readable in some senses :thinking:

utkarsh-pro: Updated the PR. Replaced named and naked returns with explicit returns. Thanks for reviewing @subhamkrai! :smile:

subhamkrai: @utkarsh-pro thanks😀

leecalcote:
I wonder if we need a `/import` of if `/pattern` as the endpoint with POST / PUT as the verb is a more restful approach. Considering the answer to this holistically and consistently across all endpoints is something we should give thought to.

kumarabd:
@utkarsh-pro yes sql/ to be moved to meshkit. Its definitely useful in there and also removes version conflicts of the package when used elsewhere.

kumarabd (changes requested):
sql/ to be moved to meshkit inside database/

utkarsh-pro:
> I wonder if we need a `/import` of if `/pattern` as the endpoint with POST / PUT as the verb is a more restful approach. Considering the answer to this holistically and consistently across all endpoints is something we should give thought to.

@leecalcote, we are not updating the resources so I am not sure if `PUT` would be appropriate. I did consider using `POST` but wasn't sure as "resource creation" won't always happen. Moreover, a user doesn't posts anything either. 

Would you suggest making the change?

utkarsh-pro:
> sql/ to be moved to meshkit inside database/

Done, kindly take a look at https://github.com/layer5io/meshkit/pull/47

utkarsh-pro:
> A small conflict in `models/providers.go` to review...

@leecalcote, the PR needs to be updated a bit, please do not merge it until it's updated. I'll resolve the merge conflicts too :sweat_smile:

leecalcote:
> > A small conflict in `models/providers.go` to review...
> 
> @leecalcote, the PR needs to be updated a bit, please do not merge it until it's updated. I'll resolve the merge conflicts too 😅

I appreciate this note, @utkarsh-pro. This helps. :D

leecalcote:
Some good looking commit messages.

leecalcote (approved):

Kn99HN:
Hi @utkarsh-pro 

I was trying to using `api/experimental/pattern` to import file remotely using this URL: `https://github.com/layer5io/meshery/blob/master/.goreleaser.yml`

The request body looks like this:
```
{
 "url": "https://github.com/layer5io/meshery/blob/master",
  "path": ".goreleaser.yml",
  "save": true
}
```
but an error happens `failed to import pattern: file not found`

Can you let me know if what I was trying to do correct?

handlers/meshery_pattern_handler.go
```diff
@@ -148,14 +64,115 @@ func (h *Handler) SavePatternFile(
...
 	}
 
-	resp, err := provider.SaveMesheryPattern(token, parsedBody)
-	if err != nil {
-		http.Error(rw, fmt.Sprintf("failed to save the pattern: %s", err), http.StatusInternalServerError)
+	format := r.URL.Query().Get("output")
+
+	// If Content is not empty then assume it's a local upload
+	if parsedBody.PatternData != nil {
+		patternName, err := models.GetPatternName(parsedBody.PatternData.PatternFile)
```
killbotXD, line 71: @utkarsh-pro CMIIW but I suppose this line should be 
`patternName, err := models.GetPatternName(parsedBody.PatternData.Name)` as we are trying to get the name of the file from model.MesheryPattern.

models/default_local_provider.go
```diff
@@ -453,46 +457,50 @@ func (l *DefaultLocalProvider) DeleteMesheryPattern(req *http.Request, patternID
...
+		repo := parsedPath[2]
+		branch := "main"
 
-	// Get the name of the file
-	name, ok := respJSON["name"].(string)
-	if !ok {
-		return nil, fmt.Errorf("failed to get filename from github")
-	}
+		if len(parsedPath) == 4 {
+			branch = parsedPath[3]
```
killbotXD, line 479: hey @utkarsh-pro 
I think should be 
```
if len(parsedPath) >= 4 {
			branch = parsedPath[len(parsedPath)-1]
		}
```
to facilitate requests that contain `blob` 
```
{
 "url": "https://github.com/layer5io/meshery/blob/master",
  "path": ".goreleaser.yml",
  "save": true
}
```

utkarsh-pro: That's not the format. Format is `https://github.com/<org>/<repo>/[branch]`. There is no need for blob.

handlers/meshery_pattern_handler.go
utkarsh-pro: The code is correct. GetPatternName will read the YAML to extract the name.

PR #2736, opened 2021-04-22
Meshery Operator - Lifecycle Management

metonymic-smokey:
**Description**
PR for Meshery operator lifecycle management.
Updates operator manifests for the `system start`, `system restart` and `system channel switch` commands.


metonymic-smokey:
// adding @kumarabd to this thread

leecalcote:
Yay!

kumarabd (approved):

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -301,6 +302,43 @@ func start() error {
 		if err != nil {
 			return err
 		}
+
+		operatorURL := "https://raw.githubusercontent.com/layer5io/meshery-operator/master/config/manifests/default.yaml"
```
leecalcote, line 306: Thank you, @metonymic-smokey 🥇 

@navendu-pottekkat, these manifests should really be place into a central location.where do you suggest?

@hexxdump, should we update the Meshery bash installation script to include these manifests for Meshery Operator, which is essentially mandatory for running Meshery? @kumarabd, should these manifests be added to the artifacts for the brew and scoop packages?

@aneeshnema, are we caching manifest files locally? When `meshery system update` is invoked, should these Meshery Operator manifests be refreshed as well?

@Aisuko, these manifests should be accounted for in the air-gapped instructions.

Please create separate issues to track each of these considerations save for the first question of centralization, which can be addressed in this PR.

leecalcote (changes requested):

mesheryctl/internal/cli/root/system/start.go
navendu-pottekkat: @metonymic-smokey @leecalcote This could be placed here along with some related URLs.

https://github.com/layer5io/meshery/blob/bec7f0737a36ad1358d484ff749885eb1f30d4d7/mesheryctl/pkg/utils/helpers.go#L37

navendu-pottekkat: Yes, mesheryctl does store the manifest files locally in `.meshery/manifests` folder.

So, from what I understand from what we are trying to do here, these manifest files should be updated and applied to the cluster.

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -301,6 +302,43 @@ func start() error {
...
+		meshsyncManifest, err := meshkitutils.ReadFileSource(meshsyncURL)
+		if err != nil {
+			return err
+		}
+
+		err = utils.ApplyManifest([]byte(meshsyncManifest), client, true, false)
+		if err != nil {
+			return err
+		}
+
```
navendu-pottekkat, line 341: We can move the `fetch- download- apply` steps of the manifest files to the already existing functions. 

That is, the level of abstraction in `start.go` should remain the same as before and all the changes made here would be additions to the existing functions in `platform.go`.

I am suggesting this because the only change here is that we have more manifest files and re-writing some stuff could be avoided and the abstraction in the `start.go` would be helpful.

navendu-pottekkat (changes requested):
@metonymic-smokey Left some suggestions to look into before you move further.

mesheryctl/internal/cli/root/system/start.go
kumarabd: @leecalcote No these manifests are going to be dynamic, based on the operator version. So the current way seems better.

metonymic-smokey: So I'll undo my latest commit @kumarabd and keep it as is WRT URLs?

mesheryctl/internal/cli/root/system/start.go
metonymic-smokey: @navendu-pottekkat made these changes in my latest commit. Please review them.

mesheryctl/internal/cli/root/system/start.go
leecalcote: @kumarabd, open an issue to have these manifests be added to the artifacts for the brew and scoop packages, please. List the issue number here.

leecalcote:
@aneeshnema, will you open an issue to ensure that when meshery system update is invoked, should these Meshery Operator manifests are refreshed as well?

leecalcote:
@Aisuko, these manifests should be accounted for in the air-gapped instructions. Will you crate an issue for this?

mesheryctl/internal/cli/root/system/start.go
kumarabd: @leecalcote Wt I meant was to not add it to brew/scoop but continue pulling from the Meshery-operator repository. This is to maintain consistency.

kumarabd: @leecalcote Wt I meant was to not add it to brew/scoop but continue pulling from the Meshery-operator repository. This is to maintain consistency.

leecalcote:
Air-gapped support

mesheryctl/pkg/utils/platform.go
```diff
@@ -96,6 +96,25 @@ func DownloadManifests(manifestArr []Manifest, rawManifestsURL string) error {
 			}
 		}
 	}
+
+	operatorFilepath := filepath.Join(MesheryFolder, ManifestsFolder, MesheryOperator)
+	err := DownloadFile(operatorFilepath, OperatorURL)
```
navendu-pottekkat, line 101: These files will be downloaded from the same release regardless of the version specified in the Meshery config file.

navendu-pottekkat (changes requested):
@metonymic-smokey A slight issue here is that the new manifest files downloaded does not consider the version specified in the Meshery config file.

These files should also be downloaded from the corresponding release of Meshery as the other manifest files are downloaded.

That implementation could be modified to include these files too.

mesheryctl/pkg/utils/platform.go
leecalcote: @navendu-pottekkat, which is not good, correct? We should be downloading the version in accordance with what is specified in the respective `context`.

navendu-pottekkat: Yes @leecalcote . We should fetch the file from the particular release specified in the context. 

@metonymic-smokey This should be a simple fix as the other files are already being downloaded in this way.

leecalcote (approved):
@metonymic-smokey other than the need to identify the specific manifest version prior to download, I think this PR ready for merge. Will you open a separate issue to track  the need for version-specific download?

navendu-pottekkat:
@leecalcote @metonymic-smokey Related concern- https://github.com/layer5io/meshery/issues/2758#issuecomment-828151529

PR #2745, opened 2021-04-24
Add "order by" query sanitization

tangledbytes:

**Description**

This PR adds "order by" query sanitization for preventing SQL injection attacks

**Notes for Reviewers**

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -55,10 +55,10 @@ func main() {
 		os.Exit(1)
 	}
 
-// 	operatingSystem, err := exec.Command("uname", "-s").Output()
-// 	if err != nil {
-// 		logrus.Error(err)
-// 	}
+	// operatingSystem, err := exec.Command("uname", "-s").Output()
+	// if err != nil {
+	// 	logrus.Error(err)
+	// }
 
 	ctx := context.Background()
 


leecalcote:
Do we need to be concerned with `group by`?

utkarsh-pro:
> Do we need to be concerned with `group by`?

We don't take `group by` queries from users at the moment. However, in future if our APIs do start supporting `group by` clause then probably we will need to sanitize that too.

models/sql-utils.go
```diff
@@ -0,0 +1,36 @@
...
+		if col == inputCol {
+			if typ == "desc" {
+				return fmt.Sprintf("%s desc", col)
+			}
+
+			return fmt.Sprintf("%s asc", col)
+		}
+	}
+
+	return ""
```
leecalcote, line 35: I wonder if we should be logging an error or if that leaves Meshery open to a potential overflow of logs.

models/meshery_pattern_persister.go
```diff
@@ -25,6 +25,8 @@ type MesheryPatternPage struct {
 
 // GetMesheryPatterns returns all of the performance profiles
 func (mpp *MesheryPatternPersister) GetMesheryPatterns(search, order string, page, pageSize uint64) ([]byte, error) {
+	order = sanitizeOrderInput(order, []string{"created_at", "updated_at", "name"})
```
leecalcote, line 28: How do we make this new function compulsory for all Get() functions that invoke SQL queries?

Do we need to consider Update() and Delete() functions as receiving the sanitizeOrderInput() as well?

models/sql-utils.go
utkarsh-pro: I am not sure if logging an error that an invalid "order by" query was passed will be helpful as if for some reason the query was actually malformed then it is probably an attack. However, there could be a case when a developer may have written a malformed query while developing, so for this use case adding a debug log in here maybe helpful in debugging.

models/meshery_pattern_persister.go
utkarsh-pro: This function is now invoked before every `SELECT` type query where we expose a portion of the query in the API. I am not sure if can get any more stricter here as the function needs to know the "whitelisted" column names and they differ for each table.

We won't need the custom sanitization in the functions like `Update` or `Delete` as they don't use any of the SQL injection unsafe method listed here: https://gorm.io/docs/security.html

ssst0n3:
wonderful fix, LGTM

models/meshery_pattern_persister.go
leecalcote: Thank you, @utkarsh-pro :thu

PR #2700, opened 2021-04-09
[CI] Testing Helm Chart Release workflow

leecalcote:
**Description**

Just a test of the new workflow.

PR #2757, opened 2021-04-27
[CI] Helm Chart Releaser Workflow: update helm chart location

leecalcote:
**Description**

This PR fixes #2700

**Notes for Reviewers**
I think that the TARGET environment reference will probably need updated as well.


leecalcote:
@Vi1234sh12 first step taken...

PR #2761, opened 2021-04-28
Display endpoint based on platform

### LINKED ISSUE #2752: [Mesheryctl] Print meshery endpoint after deployment
#### Description
Meshery endpoint information needs to be printed in the output log, where meshery is available after deployment.

#### Implementation
Use meshkit service discovery in case of Kubernetes deployment.

commands:
mesheryctl system start

---

metonymic-smokey:

**Description**
This PR displays endpoint on `mesheryctl system start` for both Docker and Kubernetes.

This PR fixes #2752 

**Screenshots**
1. Docker:
image

2. Kubernetes
image

### DIFF
--- a/mesheryctl/internal/cli/root/system/status.go
+++ b/mesheryctl/internal/cli/root/system/status.go
@@ -40,8 +40,6 @@ var statusCmd = &cobra.Command{
 	Args:  cobra.NoArgs,
 	Long:  `Check status of Meshery and Meshery adapters.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		log.Info("Meshery status... \n")
-
 		// Get viper instance used for context
 		mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
 		if err != nil {


metonymic-smokey:
@kumarabd this displays the endpoint URL on `system start`.   
Anything more I could add to this?

Thanks!

kumarabd:
> @kumarabd this displays the endpoint URL on `system start`.   
> Anything more I could add to this?
> 
> Thanks!  

This is it for kubernetes, for docker.. how is the endpoint discovered currently?

leecalcote:
@metonymic-smokey, there’s unwanted duplicity in the messaged for Docker-based endpoint.

metonymic-smokey:
> This is it for kubernetes, for docker.. how is the endpoint discovered currently?
@kumarabd , for Docker I use the endpoint directly from the `config.yaml`, as seen in the screenshot.

metonymic-smokey:
> @metonymic-smokey, there’s unwanted duplicity in the messaged for Docker-based endpoint.

good point @leecalcote 
I'll change the earlier deployment message accordingly. It has `localhost:9081` hardcoded so that should be changed anyways.

metonymic-smokey:
image

@leecalcote I have removed the duplicity here.  
Is this alright?

Thanks!

leecalcote:
Thanks for updating. No, not quite. Let’s avoid two sentences and have just one. Ideally, their is not change in phrasing of what we have today, but the change is that the endpoint displayed has been vetted and is potentially updated based on endpoint discovery.

leecalcote:
@metonymic-smokey, let's be as consistent as possible between the two outputs, ideally, using the same phrasing for both.

metonymic-smokey:
@leecalcote @kumarabd I just discovered that changing the endpoint in `config.yaml` when the platform is docker does not necessarily result in the endpoint getting shifted from `localhost:9081` to the changed endpoint. I'll look into this and try to fix it as part of this PR.
Thanks!

leecalcote:
@metonymic-smokey _Gasp!_

navendu-pottekkat:
Related issue- https://github.com/layer5io/meshery/issues/2737

metonymic-smokey:
@kumarabd @leecalcote changing the port in `meshery.yaml` does the trick and I can now access Meshery UI on port 6081.  
My approach to fixing this is:
1. Extract port from the endpoint URL in `config.yaml`.  
2. Modify `meshery.yaml` to replace [9081](https://github.com/layer5io/meshery/blob/386fa6913e7dedc615e308c53fb5babce5872418/docker-compose.yaml#L15) with the custom port.  
3. Run as usual.  

Please let me know if this approach is in the right direction. 
Thanks!

metonymic-smokey:
@leecalcote with regard to consistent messaging, we could use this message for both platforms: "Opening Meshery in your browser. If Meshery does not open, please point your browser to `$endpoint_URL` to access Meshery".  

To make this consistent, Kubernetes deployments will also have to navigate to the browser once Meshery is deployed, similar to  docker deployments This will also result in a consistent user experience for deployments across platforms.

Let me know if this is in the right direction.
Thanks!

kumarabd:
@metonymic-smokey sounds good!

metonymic-smokey:
@kumarabd do I open the default endpoint for docker(`localhost:9081`) in the browser, in this PR? Or can I go ahead and try to modify the docker compose files?

Thanks!

metonymic-smokey:
@kumarabd I have implemented a draft version of the approach in my latest commit. Let me know what you think!

@leecalcote I have synchronised  the endpoint display messages and the user experience for both platforms. Let me know if there are any other changes needed!

kumarabd:
> @kumarabd I have implemented a draft version of the approach in my latest commit and Meshery UI can now be accessed on a custom port, based on `config.yaml`. Let me know what you think!
> 
> @leecalcote I have synchronised  the endpoint display messages and the user experience for both platforms. Let me know if there are any other changes needed!

Awesome , I'll check this out!

metonymic-smokey:
@leecalcote @kumarabd I have moved browser navigation to a function in `helpers.go`, which helps in abstracting all of that from `start.go`.   
In the earlier format, there used to be a `break` if `exec.LookPath` failed in Linux. I'm not sure how to handle that here, so I've just returned the error there for the time being.  How should I handle that here? (I've marked it with a comment)

Thanks!

navendu-pottekkat (commented):
So, if user changes the endpoint in `config.yaml` it will change the deployed endpoint for docker. But it does not do the same for Kubernetes- which is an inconsistent user experience.

We will add a feature after this PR to automatically update the endpoint `config.yaml`. (#2752)

This will ensure that `mesheryctl` calls the correct API endpoints when trying to reach Meshery server. For example in `pattern apply`, we use the `GetBaseMesheryURL()` function to get the `endpoint` from `config.yaml`.

So, the workflow in my opinion should be that Meshery should have the default endpoint at `http://localhost:9081` and if the user changes this endpoint, it will only change the base URL used by `mesheryctl`.

And if the user wants to deploy Meshery to some other endpoint, then they have to manually do it in the docker-compose / manifest files.

For that, we need to stop overwriting the docker-compose with every `start` command and preserve the changes added there. We will need to check how that would be feasible.

But if this is the behaviour we want, then the only thing missing here is that we update the endpoint in `config.yaml` for the Kubernetes platform. (#2752)

metonymic-smokey:
@navendu-pottekkat thanks for the review.
agreed it will be an inconsistent UX. 
If we follow your approach, to decide whether the docker-compose file should be downloaded, would checking if it is present in the `.meshery` folder be a viable solution? Maybe also add a flag to download the latest docker-compose file?

@kumarabd @leecalcote which approach do you think would be more suited here?

kumarabd:
@metonymic-smokey my call is, always pull from GitHub. Because in cases when we introduce new adapters or functionalities in docker-compose, it won't get updated because the older version already exists in `.meshery`. Although another approach, a better one would be to compare hash of downloaded and localfile and proceed futher, but I think it will introduce complexity. Your thoughts? @navendu-pottekkat @leecalcote

leecalcote:
I think that we're not really ready for "cached" manifests and all of the considerations that come with them. We'll need a design specification to really address all of the concerns (e.g. provide user control over a mode: air-gapped vs. internet-connected, allow user to customize deployment manifests, but don't overwrite their customizations - the concept of extension points is missing (and we probably shouldn't have it, but implicitly do today), and a number of other concerns).

Such a design spec and consideration is not a priority now and won't be for the next 6 or more months. Our more immediate concerns are:

1. providing consistent UX
2. not allowing changes in one mesheryctl context to adversely affect another (constraining endpoint update to the current-context or to the context define in the global flag `--config`.
3. verifying that in fact Meshery endpoint is accessible.
4. that mesheryctl can both reach _and_ authenticate to that endpoint
5. documenting in user-facing docs (docs.meshery.io in one of the Guides) when and what of users' customizations of manifests will be overwritten for both the docker and kubernetes platform.
    - example customization to docker manifest: the pinning of a giving Meshery Adapter to a specific image tag.

metonymic-smokey:
thanks @leecalcote @kumarabd !
so for this PR, should I hardcode `localhost:9081` as the endpoint for docker and just display the endpoints, or should I change the baseURL of the endpoint and change `config.yaml` accordingly?

kumarabd (approved):
LGTM now!

leecalcote:
Thank you for committing to docs in 5. above, @metonymic-smokey. Is there logic included here that addresses 1. to 4.?

metonymic-smokey:
@leecalcote 
1. I added the `NavigateToBroswer` function to provide consistent UX.
2. The only changes made in the `config.yaml`/`meshery.yaml` files are made for the `current-context`.
3. and 4. I believe this is done for Kubernetes, internally by the `meshkitkube` functions I'm using. 
I'm not sure what these two points entail when docker is used. 
I think I could use `TcpCheck` from meshkit utils once I'm a bit clearer on it.

Thanks!

metonymic-smokey:
@leecalcote added a draft of the endpoint accessibility check for docker deployments. I'm a bit unsure about it since I'm not too familiar with this utility. Could you please review it?
Much thanks!

leecalcote:
Oh, very good. I'm tagging @piyushsingariya here to have a look as well.

metonymic-smokey:
*sighs in relief* I was so unsure of this last commit :P

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -173,6 +179,22 @@ func start() error {
...
+		endpoint.Port = int32(tempPort)
+
+		var mockEndpoint *meshkitutils.MockOptions
+		mockEndpoint = nil
+
+		res := meshkitutils.TcpCheck(&endpoint, mockEndpoint)
+		if res {
+			return errors.New("the endpoint is not accessible")
+		}
+
```
piyushsingariya, line 197: LGTM, we're checking if any other services aren't already running on the defined port, right?

leecalcote: Ideally, we are also invoking an open (unauthenticated) API on Meshery to verify that the endpoint is in fact Meshery and is healthy; e.g. `/api/server/version`. @mgfeller @kumarabd @Aisuko recommendations on:

1. a "/healthz" endpoint in Meshery Server?
2. a reusable approach in MeshKit for verifying an application response beyond an open TCP connection/endpoint?

piyushsingariya: Ohh I have one tiny idea, can we add it as `runPortAvailabilityHealthCheck()` in `check.go`??

Thoughts?? @navendu-pottekkat @leecalcote

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -173,6 +179,22 @@ func start() error {
...
+		tempPort, err := strconv.Atoi(userPort[len(userPort)-1])
+		if err != nil {
+			return err
+		}
+		endpoint.Port = int32(tempPort)
+
+		var mockEndpoint *meshkitutils.MockOptions
+		mockEndpoint = nil
+
+		res := meshkitutils.TcpCheck(&endpoint, mockEndpoint)
```
navendu-pottekkat, line 193: Shouldn't we check this after Meshery is deployed?

navendu-pottekkat (commented):
Shouldn't we also do `TcpCheck` for `platform: kubernetes`?

Tested out with Kubernetes platform, getting an error opening the browser to the URL I guess-

Screenshot from 2021-05-05 22-54-52

mesheryctl/internal/cli/root/system/start.go
metonymic-smokey: TcpCheck is done by the `meshkit` functions I'm using, `GetServiceEndpoint`.

metonymic-smokey: Also, the browser navigation code is the same as the one used for Docker so IDK how it's not working for you here. It definitely works for me for both platforms.

metonymic-smokey: I'll change the TcpCheck placement for docker to after the deployment. Thanks!

metonymic-smokey: @navendu-pottekkat made the change in my latest commit.

It's failing the `golangci-lint` test despite showing no errors locally. Is there something to fix that?
image

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -108,32 +115,42 @@ func start() error {
 		if err != nil {
 			return err
 		}
+
 		services := compose.Services // Current Services
+		//extracting the custom user port from config.yaml
+		userPort := strings.Split(currCtx.Endpoint, ":")
+		userPortMapping := userPort[len(userPort)-1] + ":8080"
```
leecalcote, line 122: The hardcoded port 8080 here is a dangerous assumption.

@kumarabd have you updated the helm chart to use 9081, yet?

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -304,7 +318,60 @@ func start() error {
...
+			} else {
+				time.Sleep(1 * time.Second)
+			}
+		}
+
+		if err != nil {
+			return err
+		}
+
+		currCtx.Endpoint = "http://" + endpoint.External.Address + ":" + strconv.Itoa(int(endpoint.External.Port))
```
leecalcote, line 346: The assumption that SSL (https) is not being used is dangerous. Can this be centralized as a property?

mesheryctl/internal/cli/root/system/start.go
metonymic-smokey: Agreed, @leecalcote
I have updated it to use the container port mentioned in the `meshery.yaml` port mapping for `meshery`. 
But as long as `system start` downloads `meshery.yaml` on start everytime, it will remain 8080. That's why I had hardcoded it earlier.

mesheryctl/internal/cli/root/system/start.go
metonymic-smokey: @leecalcote I tried removing it and the default is `https`, which resulted in the error: `Kubernetes unreachable`. 
I was thinking this could be centralised in `helpers.go`?
Thanks!

leecalcote: @navendu-pottekkat @kumarabd comments on the central location? // @s1ntaxe770r

navendu-pottekkat: The central location is in the utils package...

metonymic-smokey: Yes @navendu-pottekkat , `helpers.go` is part of the `utils` package.
I have pushed a commit centralising it in `utils` package.

mesheryctl/internal/cli/root/system/start.go
kumarabd: @lee yes the helm chart is updated for 9081.

mesheryctl/internal/cli/root/system/start.go
navendu-pottekkat: Yes, I missed your comment

leecalcote:
@metonymic-smokey a small merge conflict here - https://github.com/layer5io/meshery/pull/2761/conflicts

metonymic-smokey:
@leecalcote @navendu-pottekkat 
I have resolved the conflict and am displaying endpoint for both the platforms in `system status` too now. Let me know if this is okay.
1. docker
image

2. kubernetes
image

Thanks!

leecalcote:
Good deal, @metonymic-smokey. I know this one has been a marathon, so I hate to suggest more changes. 

- Does this display the Operator status for Docker deployments?
- the “Meshery Status...” line seems like an unnecessary line that can be removed.

leecalcote:
Will this show status of the broker and MeshSync as well?

metonymic-smokey:
> Good deal, @metonymic-smokey. I know this one has been a marathon, so I hate to suggest more changes.
> 
> * Does this display the Operator status for Docker deployments?
> * the “Meshery Status...” line seems like an unnecessary line that can be removed.

@leecalcote 
1. No, it does not since we are currently not applying operators when the platform is docker. There was a separate issue opened for this and I was assigned. Once a PR for that is merged, we could add this.
2. Alright, I have removed that line in my next commit.  
3. Right now it does not, just displays pod status. I will add broker and meshsync status in my next commit. 

Thanks!

mesheryctl/internal/cli/root/system/start.go
leecalcote: Do we have such a health endpoint defined for Meshery components?

metonymic-smokey:
@leecalcote removed the extra line.
The remaining points were part of a [closed issue](https://github.com/layer5io/meshery/issues/2853). 

Let me know if there's anything more I can do @leecalcote @navendu-pottekkat .

Thanks!

leecalcote:
Wow! What a PR this is. 😄 Way to stick with this one and see it through, @metonymic-smokey. Just fantastic.

PR #2927, opened 2021-05-13
[Meshery UI] - Updated Meshery Provider UI

### LINKED ISSUE #2925: [Meshery UI] - The Meshery Provider UI Needs Some Changes - Button, Headings, Modal
#### Current Behavior
- There are two buttons, which should be made as a single one.

image

-  The Dividers needs to have it's color changed and add margins at the end along with Disabled text above it (Marked in Red Below)

image

- The modal, shown in below image, the Provider text should be made bold, also the OK button needs to have a container (shown in blue)
image

#### Desired Behavior
 - All of the above concerns to be addressed :)
 
#### Implementation
- All of the respective components are here, https://github.com/layer5io/meshery/blob/master/provider-ui/components/ProviderComponent.js

adithyaakrishna:
Signed-off-by: Adithya Krishna <aadithya794@gmail.com>

**Description**

This PR fixes #2925 

**Screenshots**

**Provider Button:**
image

**Provider Modal:**
image

### DIFF
--- a/.github/workflows/chart-releaser.yaml
+++ b/.github/workflows/chart-releaser.yaml
@@ -1,0 +1,25 @@
+name: Release Charts
+on:
+  push:
+    branches:
+      - master
+      
+jobs:
+  release-chart:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Publish Helm chart
+        uses: stefanprodan/helm-gh-pages@master
+        with:
+          token: ${{ secrets.GH_ACCESS_TOKEN }}
+          charts_dir: install/kubernetes/helm/meshery/charts
+          charts_url: https://meshery.io
+          owner: layer5io
+          repository: meshery.io
+          branch: master
+          target_dir: charts
+          commit_username: l5io
+          commit_email: l5io@layer5.io
+          #if: github.event_name != 'pull_request' && startsWith(github.ref, 'refs/tags/') && success()
+          # chart_version: ${GITHUB_REF/refs\/tags\//}
--- a/.github/workflows/helm-chart-releaser.yml
+++ b/.github/workflows/helm-chart-releaser.yml
@@ -60,20 +60,20 @@ jobs:
       - name: Helm Chart Releaser For Remote
         uses: layer5io/helm-chart-releaser@v0.1.0
         with:
-          charts_dir: install/kubernetes/helm/meshery
+          charts_dir: charts
           owner: layer5io
-          repo: layer5io/meshery.io
+          repo: meshery.io
           # For helm charts index
           charts_repo_url: https://meshery.io/charts/
           pages_branch: master
-          activity: none
+          activity: push
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
       - name: Checkout target repo
         uses: actions/checkout@v2
         env:
           REPO: layer5io/meshery.io
-          REPO_PATH: meshery.io
+          REPO_PATH: charts
         with:
           fetch-depth: 0
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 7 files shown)


adithyaakrishna:
@leecalcote would a disabled text like this be okay?
**Option 1:**
image

**or**

**Option 2:**
image

leecalcote:
Thank you. Option 2 is better. Even then, too much attention is called to these disabled options. They need to fade into the background, not have attention called to them.

adithyaakrishna:
> Thank you. Option 2 is better. Even then, too much attention is called to these disabled options. They need to fade into the background, not have attention called to them.

Done 💯 
image

leecalcote:
Haha! Nice. 🙂

leecalcote:
Looks like we might have an end-to-end test failing here - https://github.com/layer5io/meshery/pull/2927/checks?check_run_id=2578845607

adithyaakrishna:
@leecalcote I'm not able to figure why it's failing :(
Is it because I used `&nbsp;` ?

provider-ui/components/ProviderComponent.js
```diff
@@ -256,7 +274,7 @@ class ProviderComponent extends React.Component {
                       onClick={self.handleToggle()}
                     >
                       <ArrowDropDownIcon />
-                    </Button>
+                    </Button> */}
```
ramrodo, line 277: The CI error is because you commented/deleted the button that we're using for selecting a Provider. You may want to move this `data-cy="select_provider"` property to your new button.

ramrodo (changes requested):

provider-ui/components/ProviderComponent.js
adithyaakrishna: Tried it, still facing the same error 😶

ramrodo: The error is different now :)

The previous error was:

```bash
AssertionError: Timed out retrying after 4000ms: Expected to find element: `[data-cy="select_provider"]`, but never found it.
```

Now, the problem is that, by default, the **Meshery** provider is auto-selected. But, if you want to change the Provider (after selecting one or if it is already selected by default in this test), you're not able to change it anymore.

I think we have two options:
1. We should allow changing the Provider after selecting one or;
2. Send the request immediately after selecting one option in the Dropdown (and not require to click again in the button to send the request)

@leecalcote what do you think about these two options?

leecalcote:
Thank you for laying this out so clearly.This makes sense. My vote is for option two.

ramrodo:
> Thank you for laying this out so clearly. This makes sense. My vote is for option two.

@adithyaakrishna please make the changes according to this. Then, we can update the test together.

leecalcote:
This is very close to completion. 👍

adithyaakrishna:
@leecalcote Yesss, I'm kinda stuck up on this itself, but will fix this soon and make it ready for Merging 💯

leecalcote:
@Tanuj22 @Nikhil-Ladha do either of you know how to set the auto-select and submit behavior as described in #2 here?

Nikhil-Ladha:
> @Tanuj22 @Nikhil-Ladha do either of you know how to set the auto-select and submit behavior as described in #2 here?

I think this might be a bit difficult to have. But, the former should be possible. Anyways, I would check them and see what can be done.

adithyaakrishna:
Thanks a lot @Nikhil-Ladha for helping me out with this 🎉

provider-ui/components/ProviderComponent.js
adithyaakrishna: Fixed this 🎉

adithyaakrishna:
@ramrodo The error in ProviderComponent was fixed, but now the UI Integration test is failing because of some cypress error,

image

I found a solution for this locally, we can fix it by adding this line,

image

ramrodo:
> I found a solution for this locally, we can fix it by adding this line,

Yeah! That solution is ok. It's because we are now submitting the provider chosen and we're hitting the API. Disabling the `chromeWebSecurity` allows this request to be done to localhost (CORS problem).

adithyaakrishna:
Thanks a lot @ramrodo and @Nikhil-Ladha 🥳
Finally this PR works and is ready to merge @leecalcote 🎉

ramrodo (approved):
Great! :)

leecalcote (approved):
Way to see this through, Adi.

PR #2807, opened 2021-05-03
Cached Manifests during Kubernetes deployment

Maxgit3:
**Description**

This PR fixes #2580 

**Notes for Reviewers**
Added a yaml sruct and a new function for getting cached local manifests

**[Signed commits](https://github.com/layer5io/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #2929, opened 2021-05-14
[mesheryctl] Add skip-update flag on `system restart` cmd

### LINKED ISSUE #2906: [mesheryctl] Add `--skip-update` flag to `system restart` command

#### Current Behavior
`system start` offers user the ability the flag `--skip-update`, which has the effect of allowing them to bypass checking for and downloading new versions of Meshery's container images.

#### Desired Behavior
Ideally, `system restart` supports the same `--skip-update` flag with the same behavior.

---

#### Contributor Resources

alphaX86:

**Description**

This PR fixes #2906 

**Notes for Reviewers**
Tested locally and it worked. Maintainers do check if it needs any changes!


alphaX86:
Ugh... Need to lint the files...

metonymic-smokey:
@alphaX86 I tried running this without starting Meshery in the first place. It says Meshery is restarted but running `system status`/`system stop` shows Meshery is not running.    
image

image

Maybe this could be a condition you could check for? Check if Meshery is not running and start it in that case.

alphaX86:
@metonymic-smokey 😮 I didn't encounter this actually yet. Anyway, I'll check it though! 👍

alphaX86:
@metonymic-smokey I tried your way
image

I didn't encounter error actually. I stopped in middle to show the output

alphaX86:
And this is when meshery is running...
image

alphaX86:
IDK what went wrong then... 🤔

alphaX86:
@leecalcote what do you think about this? Do I need to dig further or ask any comments from reviewers?

mesheryctl/internal/cli/root/system/restart.go
```diff
@@ -66,6 +66,14 @@ func restart() error {
...
 		}
 
+		if skipUpdateFlag {
+			log.Info("Skipping Meshery Update...")
+		} else {
+			if err := start(); err != nil {
+				return errors.Wrap(err, utils.SystemError("Failed to restart Meshery"))
+			}
+		}
+
```
navendu-pottekkat, line 76: In the restart command, I guess we need to only set the `skipUpdateFlag` as the actual function is carried out by `start()`. 

There is already a logging in `start.go` that does this, so no additional logic needs to be written here.

Please check line 178 in `start.go` and confirm-

```go
		if skipUpdateFlag {
			log.Info("Skipping Meshery update...")
		} else {
			err := utils.UpdateMesheryContainers()
			if err != nil {
				return errors.Wrap(err, utils.SystemError("failed to update Meshery containers"))
			}
		}
```

mesheryctl/internal/cli/root/system/restart.go
```diff
@@ -97,3 +105,7 @@ func restart() error {
 	}
 	return nil
 }
+
+func init() {
+	restartCmd.Flags().BoolVarP(&skipUpdateFlag, "skip-update", "", false, "(optional) skip checking for new Meshery's container images.")
```
navendu-pottekkat, line 110: We would only need to register the flag here I guess and it is centrally controlled in `start.go` for all the system commands.

navendu-pottekkat (changes requested):
@alphaX86 Please check the suggestions.

On a related note, for `system start --skip-update` `platform: kubernetes` does not take the flag into account. Need to work out how it will handle that.

// @metonymic-smokey We are finding bugs already

navendu-pottekkat:
@metonymic-smokey Yes, this command is also missing a `isMesheryRunning` check.

navendu-pottekkat:
Or, if Meshery is not running, it can automatically just start it. Maybe providing that as a log to the user

mesheryctl/internal/cli/root/system/restart.go
alphaX86: Got it. But won't we need to print during help flag used?

mesheryctl/internal/cli/root/system/restart.go
alphaX86: Alright

alphaX86: So, instead of re-writing the whole fn again, what I need to do in such case?

alphaX86: @navendu-pottekkat oh yeah. I removed the fn just now and it worked! 😨

mesheryctl/internal/cli/root/system/restart.go
alphaX86: I mean when --help is used....

alphaX86:
> @alphaX86 Please check the suggestions.
> 
> On a related note, for `system start --skip-update` `platform: kubernetes` does not take the flag into account. Need to work out how it will handle that.
> 
> // @metonymic-smokey We are finding bugs already

I didn't consider `kubernetes` for now. I also saw the source code `start.go` seems it applies only on Docker for now

mesheryctl/internal/cli/root/system/restart.go
piyushsingariya: @alphaX86 What Navendu is saying here is to only set the value of `skipUpdateFlag` to be true with the help of flag, That's all. That'll handle the whole logic.

mesheryctl/internal/cli/root/system/restart.go
piyushsingariya: That's all. Only this was needed.

mesheryctl/internal/cli/root/system/restart.go
alphaX86: Oooh OK ok

mesheryctl/internal/cli/root/system/restart.go
alphaX86: Interesting....

mesheryctl/internal/cli/root/system/restart.go
alphaX86: I'll change it after lunch

mesheryctl/internal/cli/root/system/restart.go
```diff
@@ -97,3 +97,7 @@ func restart() error {
 	}
 	return nil
 }
+
```
piyushsingariya, line 100: It works correctly on the Docker platform, there's not a need for `skip-update` for Kubernetes in `system restart`. 
Nice work, @alphaX86

mesheryctl/internal/cli/root/system/restart.go
```diff
@@ -97,3 +97,7 @@ func restart() error {
...
+func init() {
+	restartCmd.Flags().BoolVarP(&skipUpdateFlag, "skip-update", "", false, "(optional) skip checking for new Meshery's container images.")
```
piyushsingariya, line 102: It works correctly on the Docker platform, there's not a need for `skip-update` for Kubernetes in `system restart`. 
LGTM, Nice work! @alphaX86

piyushsingariya (approved):

metonymic-smokey:
> Or, if Meshery is not running, it can automatically just start it. Maybe providing that as a log to the user

Yes, that's what I was thinking too.

For Kubernetes, we'll have to find some way to download the latest manifests as updates, maybe by comparing file hashes or something similar.

piyushsingariya:
> For Kubernetes, we'll have to find some way to download the latest manifests as updates, maybe by comparing file hashes or something similar.

@metonymic-smokey We just need to trigger `start` if meshery isn't running, right?

metonymic-smokey:
> > For Kubernetes, we'll have to find some way to download the latest manifests as updates, maybe by comparing file hashes or something similar.
> 
> @metonymic-smokey We just need to trigger `start` if meshery isn't running, right?

If the `skip-update` flag has been used, yes.  
If the flag has been used, we don't download the manifests again.

alphaX86:
@metonymic-smokey @piyushsingariya IG we can talk about this during the bug hunt meet today. What do you say guys?

alphaX86:
Meanwhile, I'll make some minor changes in my PR, test them, lint the file and I'll send it in upstream

piyushsingariya:
> I was thinking of a way to avoid overwriting and downloading manifests every time we need to update.

@metonymic-smokey I guess a PR is in progress to use cached manifests, we might be able to use that for this. #2807

metonymic-smokey:
Yes, let's do that.
I'll add it to the test plan.

alphaX86:
> Meanwhile, I'll make some minor changes in my PR, test them, lint the file and I'll send it in upstream

Rechecked again, seems I need not change again. Reviewers test the change and see if it works good

alphaX86:
Any reviewer here? 😅

piyushsingariya:
Yeah, me!

leecalcote (approved):
Looks good. Works good.

PR #2951, opened 2021-05-16
Change endpoint in meshconfig(platform: Docker)

metonymic-smokey:
**Description**
This PR changes the endpoint in the meshconfig when `platform: Docker` to reflect the port specified by the user(PORT) in meshconfig's endpoint. Changes the endpoint to `http://localhost:PORT` based on the user's confirmation.
Abstracts this into a function in `platform.go` to change based on the platform.

**Screenshots**
1. Before start:
image

2. After start:
image

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,7 +1,7 @@
 jekyll=bundle exec jekyll
 
 site: install
-	$(jekyll) serve --drafts --livereload --config _config.yml,_config_dev.yml
+	$(jekyll) serve --drafts --livereload --config _config_dev.yml
 
 build: install
 	$(jekyll) build --drafts
--- a/docs/_config_dev.yml
+++ b/docs/_config_dev.yml
@@ -68,9 +68,6 @@ permalink: /:year/:title/
 markdown: kramdown
 exclude: [_site, CHANGELOG.md, LICENSE, README.md, vendor, Makefile]
 
-# plugins:
-#   - jekyll-redirect-from
-
 sass:
     sass_dir: _sass
     style: compact
--- a/docs/pages/extensibility/extensibility.md
+++ b/docs/pages/extensibility/extensibility.md
@@ -28,9 +28,10 @@ The following points of extension are currently incorporated into Meshery.
 
 ## Types of Extension Points
 
-1. [Providers]({{site.baseurl}}/extensibility/providers)
-1. [Load Generators]({{site.baseurl}}/extensibility/load-generators)
 1. [Adapters]({{site.baseurl}}/extensibility/adapters)
-1. [REST API](/extensibility/api#rest)
+   - Messsaging Framework (CloudEvents and NATS)
 1. [GraphQL API](/extensibility/api#graphql)
-1. CloudEvents Integration
+1. [Load Generators]({{site.baseurl}}/extensibility/load-generators)
+1. Patterns (Open Application Model (OAM))
+1. [Providers]({{site.baseurl}}/extensibility/providers)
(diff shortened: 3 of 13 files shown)


navendu-pottekkat (commented):
Does this mean that a user can only configure the port to which Meshery is deployed to?

metonymic-smokey:
Yes, that's how it was earlier as well(for docker).

metonymic-smokey:
@leecalcote could you review this pls?
Thanks!

leecalcote:
Yes, I'm having a look now... Docker as a platform is a special case as we think of it today. If we hardcode `localhost`, we're assuming that all Docker-based deployments are to be done locally, which for now, is ok, and time will tell if any user ever has the need for remote administration of a Docker-based, Meshery deployment.

leecalcote:
While our use case for Docker-based deployments is local-only, I think we want to avoid rewriting the hostname/IP address in the endpoint URL. As a user, I know that such an update would come as a surprise to me. If we really consider that the endpoint the user has listed should be updated, what do you think of prompting the user to confirm the change? Doing so hopefully has two positive effects:

1. Meshery gets to show off how intelligent it is to the user.
2. The user appreciates Meshery's intelligence and remains comfortable in the thought they are in control; that Meshery is there to help them, offering helpful suggestions. Trust is built with the user.

metonymic-smokey:
@leecalcote yes, that's a good point with regard to the user experience.   
To that end, how about using the `-y` flag and asking for a confirmation before changing the endpoint to localhost? That way, a user has a choice to deny the change.

leecalcote:
> @leecalcote yes, that's a good point with regard to the user experience.
> To that end, how about using the `-y` flag and asking for a confirmation before changing the endpoint to localhost? That way, a user has a choice to deny the change and this can be automated too, if required. It will also ensure a uniform UX across commands if we use the same flag.

This sounds spot on. 👍

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -193,7 +195,27 @@ func start() error {
...
+		userResponse := false
+		if utils.SilentFlag {
+			userResponse = true
+		} else {
+			// ask user for confirmation
+			userResponse = utils.AskForConfirmation("The endpoint address will be changed to localhost. Are you sure you want to continue?")
+		}
+
+		if userResponse {
+			endpoint.Address = utils.EndpointProtocol + "://localhost"
```
kumarabd, line 208: @metonymic-smokey does this mean that the endpoint is localhost always in a docker deployment?

metonymic-smokey: @kumarabd no, not always.
It is changed to `localhost` based on the user's confirmation.

kumarabd (approved):
LGTM

leecalcote:
Just a couple failing lint checks to be addressed here...

kumarabd:
yes I didn't realise this, ill fix this one.

metonymic-smokey:
> Just a couple failing lint checks to be addressed here...

@leecalcote I had run `golangci-lint` locally and it did not throw up any lint errors.
I remember the lint check failed because some package was missing...

metonymic-smokey:
> Just a couple failing lint checks to be addressed here...

@leecalcote the lint checks passed for my commits till yesterday...

leecalcote:
No sweat, @metonymic-smokey. They were a very simple `gofmt -s`  check. This note was in part for Abishek to consider as the PR is/was merged.

PR #2992, opened 2021-05-21
[CI] Helm: meshery server chart

leecalcote:
**Description**

Creates a new chart for Meshery server.

**Notes for Reviewers**

This seems to be necessary based on the [chart-releaser.yaml workflow.](https://github.com/layer5io/meshery/blob/master/.github/workflows/chart-releaser.yaml)


Aisuko:
This may not make sense, meshery chart can work well. Maybe we can debug on the action

PR #2997, opened 2021-05-22
Add platform support to bash script

### LINKED ISSUE #2309: [mesheryctl] [child] platform support during “bash script” installation(`platform: kubernetes`)
See Epic #2307 for context and further description of this issue.

**Current Behavior**
When a user tries to install Meshery through bash,

```
$ curl -L https://git.io/meshery | bash -
```

the [bash script](https://github.com/layer5io/meshery/blob/master/meshery) _optionally_ runs [`mesheryctl system start`](https://github.com/layer5io/meshery/blob/566261f9f9618af1200a16ec593b327644e98d67/meshery#L99) at the end of the install procedure, and assumes a Docker-host installation is desired. 

i.e. It doesn’t not provide consideration for Kubernetes as a target platform. 

**Desired Behavior**
Identify Kubernetes as a target platform. 

**Implementation**
Create a separate flag to identify target platform type, supporting either “docker” or "kubernetes”.

Assignee: @hexxdump

pottekkat:

**Description**

This PR fixes #2309 

Add a `-p` flag to the bash script which takes in a `PLATFORM`(defaults to "docker" if no platform is specified) and uses this flag to start Meshery with `mesheryctl system start -y -p $PLATFORM`.

A new global `--platform` `-p` flag has been added which will take in a platform and initialises the `config.yaml` file with it when it is first being created.

**Notes for Reviewers**

I have tested the changes in mesheryctl and it works fine.

I tested the bash script by copying the changes to a test.sh file and executing `bash test.sh -p <platform name>` pointing to the locally created `mesheryctl` binary.

Is there any other way to test this? I don't why it should not work when we deploy this to the stable release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,15 +141,15 @@ setup-nighthawk:
 #Incorporating Make docs commands from the Docs Makefile
 jekyll=bundle exec jekyll
 
-docs:
-	$(jekyll) serve --drafts --livereload
+site:
+	cd docs; bundle install; $(jekyll) serve --drafts --livereload
+
 
 build-docs:
-	$(jekyll) build --drafts --livereload
+	cd docs; $(jekyll) build --drafts
 
 docker-docs:
-	docker run --name meshery-docs --rm -p 4000:4000 -v `pwd`:"/srv/jekyll" jekyll/jekyll:3.8.5 bash -c "bundle install; jekyll serve --drafts --livereload"
-
+	cd docs; docker run --name meshery-docs --rm -p 4000:4000 -v `pwd`:"/srv/jekyll" jekyll/jekyll:3.8.5 bash -c "bundle install; jekyll serve --drafts --livereload"
 
 .PHONY: chart-readme
 chart-readme:
--- a/mesheryctl/internal/cli/root/system/context/list.go
+++ b/mesheryctl/internal/cli/root/system/context/list.go
@@ -1,7 +1,6 @@
 package context
 
 import (
-	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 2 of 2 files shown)


navendu-pottekkat:
@leecalcote I have tested the changes script with a local `mesheryctl` binary(not sure how to test e2e) but I do not see why it would behave differently when deployed.

If this is okay, I will also add the update to the documentation adding new commands.

leecalcote:
Our goal here being to eventually move to Kubernetes as the default platform or even better prompting the user if `-p` isn’t specified.

navendu-pottekkat:
@leecalcote Yes, changing the default platform here and on mesheryctl is straightforward. Once we have Kubernetes ready, we can make that change.

Should I add the prompt user in the bash script?

leecalcote:
Thank you. Yes, both in the bash script and in mesheryctl when a new meshconfig is being created, if the platform isn't set, please do prompt the user.

navendu-pottekkat:
Sounds good. Will add the change

navendu-pottekkat:
This should work for the bash script right?

The first one is the input and I echo the platform at the end of the script

Screenshot from 2021-05-22 22-42-54

navendu-pottekkat:
@leecalcote Made the changes as you mentioned. (Check the above comment for bash script changes)

1. start without specifying platform and then specifying a valid platform
2. start without specifying platform and then specifying an invalid platform
3. start specifying a platform

Screenshot from 2021-05-22 23-13-23

Please ignore the errors as they are unrelated.

leecalcote (approved):
This is great, @navendu-pottekkat. The only tweak I can suggest is that echoing the name of the platform when `-p` is present with docker or kubernetes isn't necessary. It's just debug.

If echo'ed, a message like "Deploying Meshery to <platform>..." is more useful to the user.

leecalcote (approved):

navendu-pottekkat:
> This is great, @navendu-pottekkat. The only tweak I can suggest is that echoing the name of the platform when `-p` is present with docker or kubernetes isn't necessary. It's just debug.
> 
> If echo'ed, a message like `Deploying Meshery to <platform>...` is more useful to the user.

I do not think I echo, I used that for a demo script to demonstrate the change I made rather than testing with the whole bash script.

PR #3001, opened 2021-05-22
Fix operator lifecycle for update and reset

### LINKED ISSUE #2958: [mesheryctl] `system update` does not update operator manifests for Docker platform

#### Current Behavior

`mesheryctl system update` for Docker platform will only update the running containers and not the Meshery operator pods deployed on the Kubernetes cluster.

#### Desired Behavior

`mesheryctl system update` should also update the meshery-operator similar to how this command function for the Kubernetes platform.

---

#### Contributor Resources

metonymic-smokey:
**Description**
This PR fixes the operator lifecycle for `system update` and `system reset` commands.  It also adds the `skip-reset` flag for docker in `update`.
It also adds a few messages for a better user experience.  

This PR fixes #2958. 

### DIFF
--- a/ui/components/Filters.js
+++ b/ui/components/Filters.js
@@ -13,7 +13,6 @@ import {
 } from "@material-ui/core";
 import { UnControlled as CodeMirror } from "react-codemirror2";
 import DeleteIcon from "@material-ui/icons/Delete";
-import UpdateIcon from "@material-ui/icons/Update";
 import UploadIcon from "@material-ui/icons/Publish";
 import { connect } from "react-redux";
 import { bindActionCreators } from "redux";
@@ -71,20 +70,11 @@ function YAMLEditor({ filter, onClose, onSubmit }) {
             lint: true,
             mode: "text/x-yaml",
           }}
-          onChange={(_, data, val) => setYaml(val)}
+          onChange={(val) => setYaml(val)}
         />
       </DialogContent>
       <Divider variant="fullWidth" light />
       <DialogActions>
-        <Tooltip title="Update Filter">
-          <IconButton
-            aria-label="Update"
-            color="primary"
-            onClick={() => onSubmit(yaml, filter.id, filter.name, "update")}
-          >
-            <UpdateIcon />
-          </IconButton>
-        </Tooltip>
         <Tooltip title="Delete Filter">
           <IconButton
             aria-label="Delete"
--- a/ui/pages/configuration/filters.js
+++ b/ui/pages/configuration/filters.js
@@ -5,8 +5,6 @@ import { connect } from "react-redux";
 import { bindActionCreators } from 'redux'
 import Head from 'next/head';
 import { getPath } from "../../lib/path";
-// import { resolveHref } from "next/dist/next-server/lib/router/router";
-// import dataFetch from "../../lib/data-fetch";
 
 const styles = {
   paper: {
@@ -21,28 +19,6 @@ class NewFilters extends React.Component {
     console.log(`path: ${getPath()}`);
     this.props.updatepagepath({ path: getPath() });
   }
-
-  // async componentsDidMount() {
-  //   await new Promise(resolve => {
-  //     dataFetch(
-  //       `/api/experimental/filter${query}`,
-  //       {
-  //         credentials: "include",
-  //       },
-  //       (result) => {
-  //         console.log("FilterFile API", `/api/experimental/filter${query}`);
-  //         updateProgress({ showProgress: false });
-  //         if (result) {
-  //           setFilters(result.filters || []);
-  //           setPage(result.page || 0);
-  //           setPageSize(result.page_size || 0);
-  //           setCount(result.total_count || 0);
-  //         }
-  //       },
-  //       handleError
-  //     );
-  //   })
-  // }
 
   render () {
     return (


metonymic-smokey:
I think the lifecycle part is fixed here but I have a doubt: in `update.go`, for docker, should I create manifests folder assuming cases where user runs this before `start`?

Also, let me know if any of the messages logged need to be improved. Just added them for a more consistent UX.

Thanks!

leecalcote:
@metonymic-smokey, you're difficult to keep up with... and I love it. 😄

navendu-pottekkat:
@metonymic-smokey Ideally yes. We should always handle such errors. So even if the user runs `update` as their first command, we should handle it gracefully.

metonymic-smokey:
Pushed an extra commit since `golangci-lint` was failing on a merge commit.

metonymic-smokey:
> @metonymic-smokey Ideally yes. We should always handle such errors. So even if the user runs `update` as their first command, we should handle it gracefully.

@navendu-pottekkat for that situation, I'm creating the manifests folder only if Meshery is not running.

metonymic-smokey:
Some changes I thought of:  
1. Set skipUpdate flag to true in update.go - since an update has just been performed, there is no need to repeat it before restart; can save a few seconds - a minute.
2. Consider overwriting the version of the current context in meshconfig to ‘latest’.  

What do you think?
@navendu-pottekkat @leecalcote

metonymic-smokey:
Made the changes.  
Could you pls test these out locally @navendu-pottekkat ?  I'm not entirely sure of the changes for the kubernetes platform.  

Also, the command description and the opening confirmation question need to be updated to sound more generic(right now, they are very docker specific).  

Thanks!

metonymic-smokey:
Any more changes that I could add @navendu-pottekkat @leecalcote ?
Updated the logging messages, command descriptions and confirmatory question in the last 2-3 commits.  

Thanks!

navendu-pottekkat:
Running update without starting Meshery results in all adapters being pulled-

Screenshot from 2021-05-28 10-38-31

This is undesired right? Not related to your work but a general observation while testing.

navendu-pottekkat:
Since we are pulling a new compose file, the existing one where the adapters have been overwritten are not used. Hence causes this issue

metonymic-smokey:
Yep, this is definitely undesirable.    
The general idea of the update flag is to update to the newest version of _everything_ by default, unless otherwise specified.  
If we have to change this behaviour, we might have to overhaul the update command by making skip reset the default option and reseting files to be chosen using flags.

mesheryctl/internal/cli/root/system/update.go
```diff
@@ -122,14 +184,42 @@ var updateCmd = &cobra.Command{
...
 				}
+
+				err = utils.ApplyOperatorManifest(client, true, false)
+
+				if err != nil {
+					return err
+				}
 			}
+
+			skipUpdateFlag = true
```
navendu-pottekkat, line 195: Why do we need this flag here?

mesheryctl/internal/cli/root/system/update.go
```diff
@@ -67,15 +67,77 @@ var updateCmd = &cobra.Command{
...
+				}
+
+				err = utils.ApplyOperatorManifest(kubeClient, true, false)
+
+				if err != nil {
+					return err
+				}
+			}
+
+			skipUpdateFlag = true
```
navendu-pottekkat, line 117: Why skipUpdate?

metonymic-smokey:
If we want to fix it without overhauling the update command, we could try something like this:
1. Check if meshery is running.
2. If not, update based on existing `meshery.yaml`.
3. If it is running and `skip-reset` is not used, re-download `meshery.yaml`.

mesheryctl/internal/cli/root/system/update.go
```diff
@@ -67,15 +67,77 @@ var updateCmd = &cobra.Command{
...
+			err = utils.ViperDocker.Unmarshal(&dockerCompose)
+			if err != nil {
+				return err
+			}
+
+			utils.ViperDocker.Set("contexts."+mctlCfg.CurrentContext, currCtx)
+
+			err = utils.ViperDocker.WriteConfig()
+			if err != nil {
+				return err
```
navendu-pottekkat, line 140: What is the purpose of this block?

mesheryctl/internal/cli/root/system/update.go
```diff
@@ -122,14 +184,42 @@ var updateCmd = &cobra.Command{
...
+			}
+
+			utils.ViperK8s.Set("contexts."+mctlCfg.CurrentContext, currCtx)
+
+			err = utils.ViperK8s.WriteConfig()
+			if err != nil {
+				return err
+			}
+
```
navendu-pottekkat, line 222: Same here. I am not sure if the intention is clear here.

navendu-pottekkat (commented):
@metonymic-smokey I need some clarifications to make sure we are doing this right. Could you check it out?

mesheryctl/internal/cli/root/system/update.go
metonymic-smokey: This is to avoid redundant updates on `start`(when restarted) since it has already been updated.

mesheryctl/internal/cli/root/system/update.go
metonymic-smokey: This is to change the version of the current context to `latest` in `meshconfig`, since the version is effectively changed while updating.

mesheryctl/internal/cli/root/system/update.go
metonymic-smokey: This is for the same purpose as the previous, when platform: kubernetes.

mesheryctl/internal/cli/root/system/update.go
navendu-pottekkat: Okay, then why do we have multiple instances of viper for the same config file?

mesheryctl/internal/cli/root/system/update.go
navendu-pottekkat: But system update does not call system start right?

metonymic-smokey: Same changes that I proposed here - https://github.com/layer5io/meshery/pull/3001#issuecomment-848857022

metonymic-smokey: `update` does call `restart`, which calls `start`(for docker).  
Maybe for kubernetes, I could remove it.

mesheryctl/internal/cli/root/system/update.go
metonymic-smokey: I'll try to fix that @navendu-pottekkat 
Thanks!

mesheryctl/internal/cli/root/system/update.go
navendu-pottekkat: We do not need to restart for docker right? Just updating the container images should work

metonymic-smokey: I called `restart` in docker, seeing how it was called in kubernetes.  
Should I remove it?

metonymic-smokey: @navendu-pottekkat I have re worked the behaviour when platform: docker and removed `restart` from it.

mesheryctl/internal/cli/root/system/update.go
metonymic-smokey: @navendu-pottekkat done this in my latest commit.

navendu-pottekkat: Thank you. I will check it out.

mesheryctl/internal/cli/root/system/update.go
```diff
@@ -29,8 +29,8 @@ import (
 // updateCmd represents the update command
 var updateCmd = &cobra.Command{
 	Use:   "update",
-	Short: "Pull new Meshery images from Docker Hub.",
-	Long:  `Pull Docker Hub for new Meshery container images and pulls if new image version(s) are available.`,
+	Short: "Pull new Meshery images/manifest files and operator manifests.",
+	Long:  `Pull new Meshery container images/manifests and operator manifests from Github/Docker Hub and pull if new version(s) are available.`,
```
navendu-pottekkat, line 33: Should we need to specify about operator manifests? I think just specifying images and manifests should be enough. Keep the user above a level of abstraction.

mesheryctl/internal/cli/root/system/update.go
```diff
@@ -50,12 +50,12 @@ var updateCmd = &cobra.Command{
...
 		if currCtx.Version != "latest" {
 			// ask confirmation if user has pinned the version in config
-			log.Infof("You have pinned version: %s in your current conext", currCtx.Version)
+			log.Infof("You have pinned version: %s in your current context", currCtx.Version)
 			userResponse := false
 			if utils.SilentFlag {
 				userResponse = true
 			} else {
-				userResponse = utils.AskForConfirmation("Updating Meshery container images will supersede the version to latest. Are you sure you want to continue")
+				userResponse = utils.AskForConfirmation("Updating Meshery container images/manifest files and Meshery operator manifests will supersede the version to latest. Are you sure you want to continue")
```
navendu-pottekkat, line 58: Same as above

navendu-pottekkat (approved):
Looks neat @metonymic-smokey . Thank you 💯

mesheryctl/internal/cli/root/system/update.go
metonymic-smokey: agreed @navendu-pottekkat 
Done that in the latest commit.

navendu-pottekkat (approved):
Looks Good!

leecalcote:
I could be thinking of this incorrectly, but it seem that a fresh manifest for the operator is not being retrieved. I checked out branch at commit 697f2233, built mesheryctl, and executed `updated` in a `platform: docker` deployment.


```
➜  mesheryctl git:(update-reset-fix) ✗ ./mesheryctl system update
Fetching latest docker-compose file for channel: stable...

Updating Meshery...
Updating Meshery now...
Pulling meshery              ... done
Pulling meshery-istio        ... done
Pulling meshery-linkerd      ... done
Pulling meshery-consul       ... done
Pulling meshery-nsm          ... done
Pulling meshery-traefik-mesh ... done
Pulling meshery-kuma         ... done
Pulling meshery-cpx          ... done
Pulling meshery-osm          ... done
Pulling watchtower           ... done
Meshery is now up-to-date
```

metonymic-smokey:
@leecalcote it assumes the manifests folder is already created when Meshery is running, hence the check using `isMesheryRunning`.  
Do you think that check should be removed and the manifests folder should be created in all cases?

metonymic-smokey:
@leecalcote this was the output for me(I had deleted the `~/.meshery/manifests` folder prior to running start): 
1.  Here the folder is shown as being created:  
image

2. 
image

kumarabd (approved):
LGTM

leecalcote (approved):
Nice!!!

leecalcote:
Thank you for pointing out the difference in behavior, @metonymic-smokey

PR #3037, opened 2021-05-28
[Patterns] Add support for deploying applications using meshery

### LINKED ISSUE #2901: Add local "Application" Pattern capability
#### Current Behavior
Right now meshery server has only one trait. 

#### Desired Behavior
"Application" specific OAM traits and workload definition needs to be added in meshery server in order to add those capabilities.

An `application` pattern may look like:
```yaml
id: f18a7e16-9324-407a-9ab6-858c56b8ad5e # Unique ID for the pattern
name: ApplicationPattern # Name of the pattern file
version: 0.0.1-alpha # Pattern file version
services:
  myapp: # Application name (Not neccesarily a Kubernetes service)
    type: Application
    namespace: linkerd-test
    mesh: linkerd # Required only if multiple meshes are in the cluster
    settings:
      replicas: 1 # Number of replicas to run 
      # (Will be important if cluster has no service mesh installed)
      containers:
        - name: myapp # Name of the container
          image: utkarsh23/meshy:v5 # Image (only docker hub images for now)
          ports:
          - name: http
            containerPort: 8080
          resources: # Optional
            requests:
              memory: 32Mi
              cpu: 5m
    traits:
      strategy: # Deployment strategy - Can be canary or blue green
        canary:
          steps:
          - setWeight: 5 # Redirect 5% traffic to the canary
          - pause: {duration: 60} # Wait for 60s before promoting to next step'
          
          - setWeight: 40
          - pause: {duration: 10}
```

#### Acceptance Tests
1. New Trait Definition must be added along with their JSON schema
2. New Workload Definition must be added along with their JSON schema

#### Mockups
Please refer the `oam` directory for examples.

tangledbytes:
**Description**

This PR fixes #2901 

This PR adds support for new entity type in Patterns, called `Application`. This lets a user easily deploy there application on the k8s cluster.

Logical flow:
1. User runs `mesheryctl pattern apply -f application.yaml`.
2. Meshery checks k8s for available rollout engines using k8s API (this will be changed once meshsync is capable of detecting these).
3. If no engine is found then it defaults to argo and argo rollouts is installed in `argo-rollouts` namespace.
4. Rollout Engine manager exposes the requested engine via an interface. This is done in order to extend support for other engines like flagger.
5. Engine will create a rollout and will create a loadbalancer type service for it.

If a user attempts to re-apply the same pattern then K8s Merge Patch strategy is used (Refer: [RFC 7386](https://datatracker.ietf.org/doc/html/rfc7386))

**Sample pattern**
```yaml
name: ApplicationPattern
services:
  myapp:
    type: Application
    namespace: test
    settings:
      replicas: 1
      containers:
        - name: meshy
          image: utkarsh23/meshy:v5
          ports:
            - name: http
              containerPort: 8080
```

**Notes for Reviewers**
The code has some references to `canary` however it is either commented out or the code segment is unused.


leecalcote:
@utkarsh-pro can users simply deploy an application without having to use rollouts from Argo?

leecalcote:
I can't tell where the golinter is failing.

utkarsh-pro:
> @utkarsh-pro can users simply deploy an application without having to use rollouts from Argo?

Not quite. Right now, engine manager exposes an API for deploying applications. 

**Why do we have this behaviour?**
Scenario: User deploys v1 of the application without any deployment strategies, so technically we can do that without a rollout engine. That's good. However the issue would come when the user updates there pattern to v2 of there application and now they want to use canary deployment strategy. This would cause issues. Because although we can just delete the simple k8s deployment and then create a rollout resource, argo won't actually perform a canary on it. Why is that? It is so because technically it is the first version of the application that argo encountered, hence argo would directly promote 100% traffic to the application. 

**Is there a workaround to the problem stated above?**
Sort of. We can delete the k8s native deployment. Then proceed to create a rollout resource using argo and then delete that one (version 1) and then perform an upgrade. It would work, not so clean though.

utkarsh-pro:
> I can't tell where the golinter is failing.

Yeah, that's quite strange. I ran `golangci-lint run` on my system and it gave no error. I then proceeded to run golangci-lint v1.32 (running in github actions) and it also worked without any issues.

Not sure where exactly it is failing.

navendu-pottekkat:
@leecalcote @utkarsh-pro The lint error is fixed in one of @piyushsingariya 's PR. I will check and merge

models/pattern/utils/k8s.go
```diff
@@ -0,0 +1,175 @@
...
+//
+// If a resource already exists and has the label controller set to "meshery"
+// then it will try to update the resource.
+//
+// If a resource exists but does not have the label then the function will return an error
+// however if force parameter is set to true then it will replace that resource with a new
+// resource with the appropriate labels
+//
+// It will add labels to the resources:
+//	"controller": "meshery"
```
leecalcote, line 29: Potentially, we should namespace Meshery's labels.

models/pattern/patterns/application/argo/v1alpha1/types.go
```diff
@@ -0,0 +1,770 @@
+package v1alpha1
```
leecalcote, line 1: Ideally, we are importing these as a package.

models/pattern/patterns/application/argo.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/layer5io/meshery/models/pattern/patterns/application/argo/v1alpha1"
+	patternUtils "github.com/layer5io/meshery/models/pattern/utils"
+	"github.com/layer5io/meshkit/utils"
+	meshkube "github.com/layer5io/meshkit/utils/kubernetes"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const argoManifest = "https://raw.githubusercontent.com/argoproj/argo-rollouts/stable/manifests/install.yaml"
```
leecalcote, line 15: Needs to be noted in the air-gapped deployment manual.

leecalcote: Reference needs to be externalized to a Meshery Cloud as an artifact repository.

models/pattern/patterns/application/application.go
```diff
@@ -0,0 +1,229 @@
...
+// detectRolloutEngine will detect available rollout engine in the k8s cluster
+func detectRolloutEngine(kclient *meshkube.Client) rolloutEngine {
+	if detectArgoRollout(kclient) {
+		return argo
+	}
+
+	return ""
+}
+
+// detectIstioServiceMesh returns true if k8s cluster has istio service mesh installed
```
leecalcote, line 162: File an issue about MeshSync not working in your environment.

leecalcote:
Need to expose service type options to the user: ingress, loadbalancer service type and so on.

leecalcote (approved):

models/pattern/patterns/application/application.go
leecalcote: Issue #3121 opened.

models/pattern/patterns/application/argo.go
leecalcote: Issue opened https://github.com/layer5io/meshery/issues/3085

models/pattern/patterns/application/argo/v1alpha1/types.go
leecalcote: Issue opened  #3122

models/pattern/patterns/application/application.go
```diff
@@ -0,0 +1,198 @@
...
+func detectRolloutEngine(kclient *meshkube.Client) rolloutEngine {
+	if detectArgoRollout(kclient) {
+		return argo
+	}
+
+	return ""
+}
+
+// detectArgoRollout returns true if argo rollout is present in the cluster
+func detectArgoRollout(kclient *meshkube.Client) bool {
```
leecalcote, line 159: Ultimately, this should be read from MeshSync.

PR #3052, opened 2021-05-28
[Docs] Release Notes for Meshery v0.5.18

l5io:
An auto-generated pull request to document release notes for Meshery v0.5.18

### DIFF
--- a/docs/_releases/v0.5.18.md
+++ b/docs/_releases/v0.5.18.md
@@ -1,0 +1,37 @@
+---
+layout: release
+date: 2021-05-30
+tag: v0.5.18
+---
+
+## What's New
+**General**
+- Revert [Docs]
+mesheryctl:
+Create
+a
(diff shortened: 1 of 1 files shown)


leecalcote:
Interesting, it looks like the we're accidentally grabbing the next release's notes as opposed to the current release's notes. Could be a simple change in sequencing that is needed.

ramrodo:
> Interesting, it looks like the we're accidentally grabbing the next release's notes as opposed to the current release's notes. Could be a simple change in sequencing that is needed.

Right, I'll dig more into it. We need to take the current tag number that is already created

leecalcote:
I think we're good to close out this PR. Is that right, @ramrodo?

ramrodo:
Yeah, we can close this PR.

leecalcote:
Good deal.

PR #3140, opened 2021-06-05
Bug: system start does not change platform if config file is already created

pottekkat:

**Description**

When `system start` is run and the user already has a config file, then the `--platform` flag is ignored.

Hence running `mesheryctl system start --kubernetes` will not change the platform in the config file if the file exists.

This PR fixes it.

Moved the `--platform` flag to `system start` and then setting the flag will overwrite the meshconfig file. 


metonymic-smokey:
@navendu-pottekkat works for me locally. When I use the `--platform` flag on `start` with a pre-existing `meshconfig`, it changes the platform in the file for me.

LGTM!

PR #3193, opened 2021-06-11
Fix #3192: Fix installation steps in mesheryctl bash file

### LINKED ISSUE #3192: [Mesheryctl] Not able to install mesheryctl when doing from meshery repo
#### Description

Was following guide to install meshery from [here](https://docs.meshery.io/installation/quick-start#install-meshery)
When we are checkout out to meshery source code and try to follow the first step of installation, the mesheryctl is not installed.
Why? 
Because when we unzip the folder (unzipping code :arrow_down: )
https://github.com/layer5io/meshery/blob/1b567f7e11cd0aa6015d586f060cce0173db066e/meshery#L110
the folder named mesheryctl is already there and that is not replaced because it is a directory. 
Actually, you get an intermediate error which is:
```bash
error:  cannot delete old mesheryctl
        Is a directory
```
still it continue it's execution and do the remaining step and end up moving mesheryctl directory to `/usr/local/bin` which is not the required behaviour.

#### Expected Behavior

1. The bash file should be reselient enough to handle situations for the mesheryctl filenames/folders already present in the directory. Since mesheryctl is a common name, someone could have stored their important configuration or data with naming the file as meshery ctl, the `-o` option in `unzip` will override those things and there is no going back. 

     I think it is better to isolate temporary installation files from the environment and delete it once work is done.

2. If there are intermediate command failures in the bash files, there is no point of moving forward since ultimately the result will be broken or bad installation of mesheryctl.

#### Environment:
- Meshery Version: 
- Kubernetes Version: 
- Host OS: Ubuntu 20.04
- Browser: Firefox

---

**Contributor Resources:**

Abhishek-kumar09:
**Description**

- Fixes some problems in meshery installation bash file.
- Added intermediate ephemeral directories to install mesheryctl
- Stop the further command execution on failure at process of unzipping the installation files

This PR fixes #3192 

**Notes for Reviewers**: The changes are related to mesheryctl installation


navendu-pottekkat (approved):
The changes looks good. Usually, a user will not run this command inside the Meshery source code where we already have a mesheryctl folder but makes sense to unzip this to a temporary directory.

hexxdump (approved):
LGTM

leecalcote (approved):
Nice! Good work, @Abhishek-kumar09 👍

PR #3198, opened 2021-06-13
Gracefully handling invalid meshconfig errors

### LINKED ISSUE #3189: [mesheryctl] Old meshconfig causes parsing failure

#### Current Behavior
```
➜  ~ mesheryctl system start
Error: error processing config: 2 error(s) decoding:

* 'contexts[bob].token' expected type 'string', got unconvertible type 'map[string]interface {}'
* 'contexts[local].token' expected type 'string', got unconvertible type 'map[string]interface {}'
```

#### Meshconfig

- [config.yaml.zip](https://github.com/layer5io/meshery/files/6635530/config.yaml.zip)

#### Desired Behavior

Failure to parse is caught and gracefully handled by prompting the user to create a new config.

#### Environment

- Host OS: Mac
- Meshery Client Version: stable-v0.5.23

---

#### Contributor Resources

metonymic-smokey:

**Description**
This PR attempts to gracefully handle viper errors when parsing an outdated config.   
When faced with an outdated config, a user has the choice to let Meshery create a default config.   
If the user chooses to accept, the old config is backed up and the default config is created in place of the original.   The user is informed and it terminates after creation.   

This PR fixes #3189 

**Screenshots**
image

image


metonymic-smokey:
@navendu-pottekkat I sometimes get errors like this for token/context but I'm not sure how to fix it:  

image  

Could you tell me where to start so I can try fixing this?   
Thanks!

mesheryctl/internal/cli/root/system/check.go
```diff
@@ -42,7 +44,50 @@ var checkCmd = &cobra.Command{
...
+		}
+
+		cfgFile := utils.CfgFile
+		log.Println("Backing up " + cfgFile + " to original_config.yaml")
+		err := os.Rename(cfgFile, "original_config.yaml")
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		_, err = os.Create(cfgFile)
```
leecalcote, line 68: Should we be using https://github.com/layer5io/meshery/blob/3538340cf267bc92c60eed749c857d5d28c5f89e/mesheryctl/pkg/utils/helpers.go#L588 instead?

mesheryctl/internal/cli/root/root.go
```diff
@@ -115,6 +115,7 @@ func init() {
 
 // initConfig reads in config file and ENV variables if set.
 func initConfig() {
+	utils.CfgFile = cfgFile
```
leecalcote, line 118: @navendu-pottekkat, @hexxdump, have we solidified plans to move us out of an 800 line utils file and toward a centralized struct that contains such references?

leecalcote:
Thank you for this, @metonymic-smokey 👍

mesheryctl/internal/cli/root/system/check.go
metonymic-smokey: @Lee I have a doubt.  
 I think `CreateConfigFile` creates a file in `~/.meshery`, whereas I think it would be better to create a file in the user's original directory, with the same name, since there might already be a `meshconfig` in `~/.meshery`.  
What do you think?

navendu-pottekkat, line 68: Should we be doing this in root.go? i.e when the config file is being created? It will then run for each command and would not require check being invoked.

leecalcote: Good consideration... this is where the .bak comes in.

leecalcote: Unless root.go is setup to run through a number of checks already, which are defined in a single location - `system check`....

metonymic-smokey: Alright @leecalcote!  
So just to confirm before the commit, if a user has a buggy meshconfig:  
1. Create a `.bak` copy in their current directory.    
2. Create a new config(use the existing function) and populate it with defaults, but this time in `~/.meshery`, overwriting their existing `meshconfig,` if any, in `~/.meshery`.  

Is that the final plan then?

mesheryctl/internal/cli/root/root.go
anirudhjain75: There are some functions in utils.go that are expected to be moved into the mesheryctl config struct. Moved them should be a good step towards it.

anirudhjain75:
@metonymic-smokey A tip is to use a branch in your repo for development instead of directly making changes to your master branch.

metonymic-smokey:
> @metonymic-smokey A tip is to use a branch in your repo for development instead of directly making changes to your master branch.

@anirudhjain75 yep, agree with you. I usually do that, just slacked off this time :P 
Thanks for pointing it out!

metonymic-smokey:
Current result:   
image

Let me know if there is anything I could change here.  
Thanks!

mesheryctl/internal/cli/root/system/check.go
metonymic-smokey: @leecalcote I've used `.bak.yaml` since viper recognises only `YAML` in Meshery.

navendu-pottekkat (commented):
Shouldn't this work for any commands? For example, I have an invalid config file and I run `mesheryctl system status` but still the command runs instead of backing up the file.

Also, it throws an error when I modified the default config file and made it invalid-

Screenshot_20210622_093630

metonymic-smokey:
@navendu-pottekkat thanks for the review!
Fixed the second issue. This is the result for my default config:  
image

Thanks!

metonymic-smokey:
Added a fix for `channel switch` here. This is the result:  
image

image

Let me know if there is something to change here.

leecalcote:
Way to keep at this, @metonymic-smokey 💪🏽

metonymic-smokey:
@navendu-pottekkat @leecalcote one form of invalid `config.yaml` could be an empty one.  Currently, in that case, this is the result, not a very `meshconfig` specific error message:  
image

I was considering adding a specific check for an empty file(i.e a file with size 0) and producing a specific error message. Something like this:  

image

(pls excuse the rough error message, was done just to demo)

Let me know what you think of this.  
Thanks!

navendu-pottekkat:
@metonymic-smokey Yes, always yes if you are finding edge cases that everyone else misses. Thank you for being thorough 🚀

leecalcote:
The mesheryctl build check failed on its last run. I'm rerunning the build jobs now.

metonymic-smokey:
@leecalcote lint is surprisingly failing on one of the merge commits...

PR #3232, opened 2021-06-18
Changed the port , added prefix and the namespace into the istio yml

### LINKED ISSUE #3223: [Install] Change Meshery Istio Ingress Gateway port
#### Current Behavior

Meshery ingress deployment yaml currently points to the wrong service port 
#### Desired Behavior

Change the port so it points to 9081 
#### Implementation

see https://github.com/layer5io/meshery/blob/106faddab40332c0fbe6a9f55ac17784849c5a96/install/deployment_yamls/istio.yaml#L33
#### Mockups

---

samirparhi-dev:
Signed-off-by: Samir Parhi <samirparhi@gmail.com>

**Description**
Changed the port , added prefix and the namespace into the istio yml
This PR fixes #3223

**Notes for Reviewers**

### DIFF
--- a/install/deployment_yamls/istio.yaml
+++ b/install/deployment_yamls/istio.yaml
@@ -27,7 +27,7 @@ spec:
   http:
   - match:
     - uri:
-        prefix: /meshery
+        prefix: /
     route:
     - destination:
         host: meshery


install/deployment_yamls/istio.yaml
```diff
@@ -25,10 +27,10 @@ spec:
   http:
   - match:
     - uri:
-        prefix: /
+        prefix: /meshery
```
kumarabd, line 30: @samirparhi-dev Any reason why prefix is set to `/meshery`

samirparhi-dev: This is because, if my ingress gateway has the FQDN ( lestsay : myapplication.com), which serves all the service from my k8s ( which is a best practice in production grade cluster). I am supposed to access meshery with. myapplication.com/meshery. Insted exposing individual loadbalancer ip to outside world .

leecalcote: Gents, was there any further discussion had here?

samirparhi-dev: @leecalcote : I was waiting for the review and response.

kumarabd: @samirparhi-dev I think this is something specific to your usecase, there could be more others who will be using multiple endpoints with separate FQDNs. Because meshery is going to manage multiple service meshes, I think a separate FQDN is an ideal design. // @leecalcote thoughts?

leecalcote: @kumarabd waiting for your response here.

leecalcote: Err... no, we're not waiting on a response from @kumarabd. 😆 I missed your last statement and question, @kumarabd.

I agree: `prefix` is an environment-specific concern that we would do well not to predefine. @samirparhi-dev, assuming you're ok with that perspective, will you update this line? And we'll proceed with the other changes.

samirparhi-dev: Yes , I will change that

samirparhi-dev:
@kumarabd @leecalcote  : I have corrected according to your Comments. Kindly review it.

leecalcote (approved):
LGTM

PR #3283, opened 2021-06-26
Updated the meshkit error for user handler

sayantan1413:
Signed-off-by: sayantan1413 <sayantanbose2001@gmail.com>

**Description**

Updated the meshkit error for User Handler

This PR fixes #

**Notes for Reviewers**


handlers/user_handler.go
```diff
@@ -19,8 +17,8 @@ func (h *Handler) UserHandler(w http.ResponseWriter, req *http.Request, _ *model
 	// }
 
 	if err := json.NewEncoder(w).Encode(user); err != nil {
-		logrus.Errorf("error getting user data: %v", err)
-		http.Error(w, "unable to get session", http.StatusInternalServerError)
+		h.log.Error(ErrUserData(err))
```
dhruv0000, line 20: For common errors, like encoding, unmarshalling, it would be more viable to create a generic function to use everywhere, instead of creating individual functions everytime, which are already there in `error.go`.

```
func ErrMarshal(err error) error {
	return utils.ErrMarshal(err)
}
func ErrUnmarshal(err error) error {
	return utils.ErrUnmarshal(err)
}
```

I would suggest few improvements on those functions.
1. Switch from `util` package to a `errors.New(..)` (So it can generate the error while using the Error Code Utility Tool)
2. Include one more string parameter in the function to specify what it was encoding or unmarshalling.

Something like this.
```
func ErrMarshal(err error, obj string) error {
	return errors.New(...)
}
```


This will save a lot of effort in other handlers :wink:

handlers/user_handler.go
```diff
@@ -43,8 +41,8 @@ func (h *Handler) AnonymousStatsHandler(w http.ResponseWriter, req *http.Request
 	if usageStats != "" {
 		aUsageStats, err := strconv.ParseBool(usageStats)
 		if err != nil {
-			err = errors.Wrap(err, "unable to parse anonymousUsageStats")
-			logrus.Error(err)
+			h.log.Error(err)
```
dhruv0000, line 44: A similar abstract functiuon for `PraseBool` error?

handlers/user_handler.go
```diff
@@ -56,23 +54,23 @@ func (h *Handler) AnonymousStatsHandler(w http.ResponseWriter, req *http.Request
...
+
 			http.Error(w, "please provide a valid value for anonymousPerfResults", http.StatusBadRequest)
 			return
 		}
 		prefObj.AnonymousPerfResults = aPerfStats
 		if err = provider.RecordPreferences(req, user.UserID, prefObj); err != nil {
-			logrus.Errorf("unable to save user preferences: %v", err)
-			http.Error(w, "unable to save user preferences", http.StatusInternalServerError)
+			h.log.Error(ErrSaveUserPreference(err))
```
dhruv0000, line 64: Can reuse `ErrRecordPreferences` here.

dhruv0000 (changes requested):

handlers/user_handler.go
sayantan1413: Hey @dhruv0000 there are 4 parameters in the error.New so in which parameter should I put this one **Include one more string parameter in the function to specify what it was encoding or unmarshalling.** ?

handlers/user_handler.go
sayantan1413: So for this `ParseBool` error I need to create a new function right?

sayantan1413: I got a little confused in this `ParseBool` error can you tell me what I need to do here?

handlers/user_handler.go
sayantan1413: Done.

mgfeller (changes requested):
Good work :+1: 

As part of work done on errors, the errorutil should be run, and the updated files (`errorutil_*.json` and `component_info.json`) committed as well. It will also output warnings of duplicate error names etc. (`make error`).
The contributor notes are probably not yet updated with this info, it's all quite fresh. 

Also, the design is that new error codes get text placeholders, like `ErrUserDataCode = "replace_me"` then the utility replaces this with int codes. Which int code is the next one is determined by `next_error_code` in `component_info.json`. The utility then updates `component_info.json`. If you set int-codes manually, you have to change the `next_error_code` in `component_info.json` accordingly. 

Alternatively, all codes could be updated by the utility, thus renumbered.

dhruv0000 (dismissed):

dhruv0000:
@sayantan1413 Upload the JSON files with updated `next_error_code` and we are good to go.

sayantan1413:
Hey @dhruv0000 I have updated all PR with the JSON files and the "next_error_code".

dhruv0000 (approved):

handlers/prometheus_handlers.go
```diff
@@ -34,8 +34,9 @@ func (h *Handler) ScanPromGrafanaHandler(w http.ResponseWriter, req *http.Reques
 		return
 	}
 	if err = json.NewEncoder(w).Encode(availablePromGrafana); err != nil {
-		h.log.Error(ErrMarshal(err))
-		http.Error(w, ErrMarshal(err).Error(), http.StatusInternalServerError)
+		obj := "payloads"
```
kumarabd, line 37: shouldn't this be specific? Like prometheus payload or grafana payload?

kumarabd (approved):

handlers/prometheus_handlers.go
sayantan1413: Okay @kumarabd I will update this in my next pr

leecalcote:
> Alternatively, all codes could be updated by the utility, thus renumbered.

@mgfeller Is there any situation in which this happens today? Ideally, error code assignments are permanent such that accuracy of the docs:

1. doesn’t have to rely on the freshness of Google’s indexing of them
1. user’s can become familiar with a code and not be confused by presentation of a different code for the same error in a later version. 
1. user’s running an older version of Meshery don’t see inaccurate codes listed in docs that are showing them a later version of refactored codes.

errorutil_analyze_errors.json
```diff
@@ -702,24 +774,35 @@
...
       {
         "name": "ErrGrafanaBoardsCode",
         "code": "",
         "severity": "Alert",
         "long_description": "",
-        "short_description": "",
+        "short_description": "unable to get Prometheus boards",
         "probable_cause": "",
-        "suggested_remediation": "Check if your Prometheus endpoint is correct.\nConnect to Prometheus from the settings page in the UI"
+        "suggested_remediation": "Check if your Prometheus endpoint is correct\nConnect to Prometheus from the settings page in the UI"
```
leecalcote, line 805: Why is punctuation removed? This is user-facing documentation.

mgfeller:
> > Alternatively, all codes could be updated by the utility, thus renumbered.
> 
> @mgfeller Is there any situation in which this happens today? Ideally, error code assignments are permanent such that accuracy of the docs:
> 
>     1. doesn’t have to rely on the freshness of Google’s indexing of them
> 
>     2. user’s can become familiar with a code and not be confused by presentation of a different code for the same error in a later version.
> 
>     3. user’s running an older version of Meshery don’t see inaccurate codes listed in docs that are showing them a later version of refactored codes.

@kumarabd this is just an option, no-one has to use it. It is intended as a possibility if one would like to tidy up a legacy sequence, e.g. starting at 600 up to 11000 with lots of holes when introducing the new regime. Can't remember in which repository I saw that, but I did. My sense of order would urge med to clean that up. So far, non of these codes were in any reference document, i.e. your valid points do not yet apply. 

Also, in general, I think that before a 1.0.0 release, it should be ok to tidy up, change things and even (intentionally) implement breaking changes, to be able to move forward and change code to the better. That appears to be common practice  :smile:. 

Besides that, I agree with your points, of course.

errorutil_analyze_errors.json
sayantan1413: Hey @leecalcote so should I keep all the punctuation?

PR #3285, opened 2021-06-27
[WIP] Reworking workflow for start and restart

metonymic-smokey:

**Description**
This PR attempts to fix:
1. workflow in `start` for docker when endpoint is not discovered. 
2. Reworking `restart`.  
3.  Reworking updating in `start` for docker and kubernetes to include operator manifest updates in addition to container updates. 
 


metonymic-smokey:
@ reviewers, let me know if the commits fixing `restart` and `status` work as expected for you.  

The changes to `status` avoid the panic situation when pods are starting and instead displays an output like the one below:  
image

After ~1 min. this is the status: 
image

leecalcote:
@anirudhjain75 will you review?

anirudhjain75:
A bit out of context here. @leecalcote Is there someone else who can review. ?

navendu-pottekkat:
@piyushsingariya has made a recent PR that fixes the panic in `system status` when we restart Meshery. It is a simple check to see if the containers are ready. That was the thing that caused the panic.

mesheryctl/internal/cli/root/system/restart.go
```diff
@@ -94,6 +94,16 @@ func restart() error {
...
+
+		ok, err := utils.IsMesheryRunning(currPlatform)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			if err := start(); err != nil {
+				return errors.Wrap(err, utils.SystemError("Failed to restart Meshery"))
+			}
+		}
```
navendu-pottekkat, line 106: can this be an if...else? If Meshery is running then great, restart the pods. If Meshery is not running, let's start Meshery and don't attempt to restart the pods.

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -372,14 +372,14 @@ func start() error {
 		}
 
 		if err != nil {
-			return err
-		}
-
-		currCtx.Endpoint = utils.EndpointProtocol + "://" + endpoint.External.Address + ":" + strconv.Itoa(int(endpoint.External.Port))
+			log.Info("service endpoint not discovered. Meshery deployed on default endpoint")
+		} else {
+			currCtx.Endpoint = utils.EndpointProtocol + "://" + endpoint.External.Address + ":" + strconv.Itoa(int(endpoint.External.Port))
```
navendu-pottekkat, line 377: To a new user, what is a "default endpoint"? It would be confusing and the user may not know what to do next.

// @hexxdump Suggestions?

mesheryctl/internal/cli/root/system/status.go
```diff
@@ -93,6 +93,10 @@ var statusCmd = &cobra.Command{
 			// List the pods in the MesheryNamespace
 			podList, err := utils.GetPods(client, utils.MesheryNamespace)
 
+			if len(podList.Items) == 0 {
+				log.Fatal("no pods up yet...try again in a while")
+			}
+
```
navendu-pottekkat, line 99: If @piyushsingariya 's fixes work, then let's stick with that now unless something else is still breaking. If it is not, let's revert these changes.

navendu-pottekkat (commented):
@metonymic-smokey Thank you for taking this on. 

Couple of suggestions to consider.

metonymic-smokey:
Yep, I'll change that commit.

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -372,14 +372,14 @@ func start() error {
 		}
 
 		if err != nil {
-			return err
-		}
-
-		currCtx.Endpoint = utils.EndpointProtocol + "://" + endpoint.External.Address + ":" + strconv.Itoa(int(endpoint.External.Port))
+			log.Info("service endpoint not discovered. Meshery deployed on default endpoint")
```
hexxdump, line 375: @metonymic-smokey, Why should we display this message when we do not display the endpoint in the else case? I am not sure how this information is useful to the user on system start?

mesheryctl/internal/cli/root/system/status.go
metonymic-smokey: Yes done!

mesheryctl/internal/cli/root/system/start.go
metonymic-smokey: Agreed @hexxdump. I've removed that message entirely.

mesheryctl/internal/cli/root/system/restart.go
metonymic-smokey: @navendu-pottekkat yes, good idea. Just check my latest commit once to see if we're on the same page...

mesheryctl/internal/cli/root/system/status.go
metonymic-smokey: Alright, reverted the change.

navendu-pottekkat (approved):
Looks good @metonymic-smokey 💯 

On a related note, we might consider moving the existing logs in `system restart` to debug logs and use the spinner that we use in `system start` and `stop` here.

I will open up an issue for this.

PR #3296, opened 2021-06-30
[CI][WIP] integrating error codes utility into CI

DelusionalOptimist:

**Description**
Integrating the utility into CI to update the error reference at https://docs.meshery.io/reference/error-codes and running checks in PRs.
This PR fixes #

**Notes for Reviewers**
TODOs:
- [x] Support for meshery server.
- [ ] Support for mesheryctl. (Once mesheryctl uses new error codes?).
- [ ] Support for other components (To be done in their repos)
- [ ] Running checks in PRs (keeping out of scope of this PR)

### DIFF
--- a/.github/workflows/error-ref-publisher.yaml
+++ b/.github/workflows/error-ref-publisher.yaml
@@ -1,0 +1,41 @@
+name: Error Code Reference Publisher
+on:
+  push:
+    branches:
+      - 'master'
+    paths:
+      - '**.go'
+
+jobs:
+  Update-error-ref:
+    name: Error codes utility
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with: ${{ secrets.GH_ACCESS_TOKEN }}
+
+      - name: Setup Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: ${{ secrets.GO_VERSION }}
+
+      - name: Update error reference for Meshery Server
+        run: |
+          go run github.com/layer5io/meshkit/cmd/errorutil -d . analyze --skip-dirs mesheryctl
+          echo '{ "errors_export": "" }' | jq --slurpfile export ./errorutil_errors_export.json '.errors_export = $export[0]' > ./docs/_data/errorref/meshery-server_errors_export.json
+
+      - name: Update error reference for Mesheryctl
+        run: |
+          cd ./mesheryctl
+          go run github.com/layer5io/meshkit/cmd/errorutil -d . analyze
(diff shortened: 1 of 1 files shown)


.github/workflows/error-ref-publisher.yaml
```diff
@@ -0,0 +1,34 @@
+name: Error Reference Publisher
```
leecalcote, line 1:
```suggestion
name: Error Code Reference Publisher
```

.github/workflows/error-ref-publisher.yaml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  push:
+    branches:
+      - 'master'
+    paths-ignore:
+      - 'docs/**'
+      - 'mesheryctl/**'
```
leecalcote, line 8: We will probably want to run this workflow on change to mesheryctl as well.

.github/workflows/error-ref-publisher.yaml
```diff
@@ -0,0 +1,34 @@
...
+          echo '{ "errors_export": '$(cat errorutil_errors_export.json)' }' | jq > docs/_data/errorref/meshery-server_errors_export.json
+
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_user_name: l5io
+          commit_user_email: ci@layer5.io
+          commit_author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          commit_options: '--signoff'
+          commit_message: "[Docs] update meshery server error reference"
```
leecalcote, line 34:
```suggestion
          commit_message: "[Docs] Error Code Reference: update Meshery server error codes"
```

.github/workflows/error-ref-publisher.yaml
DelusionalOptimist: I think we should do this once mesheryctl gets updated with new error codes :thinking:

leecalcote: Ok, that's fine. 

For clarity: the error code utility should have idempotent behavior. What harm does enabling it on mesheryctl now cause?

DelusionalOptimist: Hmm yes, no harm. I mixed things up a bit. We can enable this :P
The utility will just throw `ERRO[0000] Unable to execute root command (open component_info.json: no such file or directory)` and exit with status 0. So the workflow will still continue.

leecalcote:
@DelusionalOptimist ready to merge current progress?

PR #3297, opened 2021-06-30
ui: disable auth token fetch without any auth providers

### LINKED ISSUE #3293: `Get Token` cannot return auth token when the user is not signed in with any providers
#### Description

The `Get Token` menuItem on the meshery user interface returns the auth token provided by the signed in providers like google, github etc. But in case the user is not logged in with any provider but only as a none user it can't actually return any token.
Currently it is returning a file with empty json fields.
What if that option can be disabled and a tooltip can be inserted that shows `login with providers to get auth token`.

#### Screenshots/Logs

image
Downloaded Token (Which is empty and may not serve the purpose that is thought before clicking the getToken button)
```
{"meshery-provider":"None","token":""}
```

#### Environment:
- Meshery Version: edge-v0.5.23
- Kubernetes Version: 1.20.7
- Host OS: Ubuntu 20.04
- Browser: Firefox

---

**Contributor Resources:**

Abhishek-kumar09:

**Description**
The Get Token menuItem on the meshery user interface returns the auth token provided by the signed in providers like google, github etc.
Currently if user is not signed in with any provider then it can't get any provider token too.
When *hovered* on the `Get Token` the tooltip shows that you need to be signed in with provider to use download token.
| Before | After |
--- | --- |
 | Screenshot from 2021-06-29 20-53-31 

This PR fixes #3293 

**Notes for Reviewers**
The ui remains unchanged if the user is signed in with any provider. It only adds additional tooltip if  the provider is none.

### DIFF
--- a/ui/components/User.js
+++ b/ui/components/User.js
@@ -5,6 +5,7 @@ import { bindActionCreators } from 'redux';
 import { withStyles } from '@material-ui/core/styles';
 import MenuList from '@material-ui/core/MenuList';
 import Grow from '@material-ui/core/Grow';
+import Tooltip from '@material-ui/core/Tooltip';
 import MenuItem from '@material-ui/core/MenuItem';
 import Popper from '@material-ui/core/Popper';
 import Paper from '@material-ui/core/Paper';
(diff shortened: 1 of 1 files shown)


Abhishek-kumar09:
@utkarsh-pro This can be closed right? or should I start writing coverage tests.

leecalcote:
@Abhishek-kumar09, yes, this can be closed, but an issue with the security label can be opened to track the need for ensuring that deployments with Local Provider enabled are secure.

@utkarsh-pro that might be the thing to do: allow users to enable or disable certain providers.

leecalcote:
Great work here, @Abhishek-kumar09. Let’s catch up when you’re free.

PR #3375, opened 2021-07-02
[mesheryctl]Adding dashboard command to mesheryctl system

### LINKED ISSUE #3035: [mesheryctl] `system dashboard` to open up Meshery UI
#### Current Behavior

Users are presented with the endpoint to reach the Meshery UI when the initially run `system start` but it is not re-callable in any other subsequent commands.

#### Desired Behavior

A new subcommand `system dashboard` which will test the connection with the Meshery UI and if available, open it on the user's default browser.

If the UI is not accessible and the platform is Kubernetes, then initiate a kube proxy tunnel.

#### Acceptance Test

Running `mesheryctl system dashboard` should open up the Meshery UI providing a consistent UX for both Docker and Kubernetes platforms.

DantrazTrev:
Signed-off-by: Ayush PS <ayushpratap16@gmail.com>

**Description**

This PR aims to fix #3035 

Tasks
- [x] Base command to display the endpoint in the terminal 
- [x] Base command to open the endpoint in the browser
- [x] Testing connection to UI
- [ ] For Kubernetes if not accessible initial kube proxy tunnel

**Notes for Reviewers**

**Not Ready for review yet**

### DIFF
--- a/mesheryctl/internal/cli/root/system/dashboard.go
+++ b/mesheryctl/internal/cli/root/system/dashboard.go
@@ -15,6 +15,8 @@
 package system
 
 import (
+	"os/exec"
+
 	"github.com/pkg/errors"
 
 	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/config"
@@ -67,11 +69,32 @@ func dashboard() error {
 	if err != nil {
 		return err
 	}
+
+	currPlatform := currCtx.Platform
+
+	err = utils.NavigateToBrowser(currCtx.Endpoint)
+
+	if err != nil {
+
+		switch currPlatform {
+		case "docker":
+			//Nothing can be done ?
+			break
+		case "kubernetes":
+			// Probably works
+			exec.Command("kubectl proxy")
+
+		}
+
+		err = utils.NavigateToBrowser(currCtx.Endpoint)
+		if err != nil {
+			return err
+		}
+
+	}
+
 	log.Info("Meshery UI is accesible at " + currCtx.Endpoint + "status:" + string(rune(resp.StatusCode)))
 	log.Info("Opening Meshery in your browser. If Meshery does not open, please point your browser to " + currCtx.Endpoint + " to access Meshery.")
-	err = utils.NavigateToBrowser(currCtx.Endpoint)
-	if err != nil {
-		return err
-	}
+
 	return nil
 }


piyushsingariya:
@DantrazTrev This issue got to be platform-dependent. We need to implement a Kube proxy tunnel in case of Kubernetes as a platform.

DantrazTrev:
@piyushsingariya I'm waiting to hear something about the bug related to UI endpoints in kubernetes env https://layer5io.slack.com/archives/C019426UBNY/p1625136547045900?thread_ts=1625136547.045900&cid=C019426UBNY

But yeah I would try to implement that meanwhile

DantrazTrev:
> @DantrazTrev This issue got to be platform-dependent. We need to implement a Kube proxy tunnel in case of Kubernetes as a platform.

Hei @piyushsingariya can you help me with the kube-proxy part , I'm not sure what's the exact command to use ?

leecalcote:
@DantrazTrev, I get excited each time I look at the subject line of this PR. Don't tease me like this. 😉 

In all seriousness, it would be great to have this command. I believe that implementation here will need to use or enhance MeshKit's capabilities. // @Aisuko

DantrazTrev:
Hei @leecalcote I believe there is just one major step left which is setting up a kube proxy tunnel for k8s when needed.  However this part is the one where I've been facing a few issues since simple kube proxy tunnels don't really work and also the endpoint for k8s is often incorrect so the tunneling would happen either way . 

I'll try my best to solve it although some dependent issues might need to be figured out first.

Aisuko:
Thanks for ping me, sir @leecalcote . If I understand right by the context, we want to achieve the feature is localhost open an proxy the dashboard. I believe that similar with the `osm dashboard`  command.
https://github.com/openservicemesh/osm/blob/7ef0c4723cc52a17feb8b8b744960a497e067d59/cmd/cli/dashboard.go#L65

leecalcote:
Right. Or another example is `linkerd dashboard`

mesheryctl/internal/cli/root/system/dashboard.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		return nil
+
+	},
+}
+
+func dashboard() error {
+
+	// Get viper instance used for context
+	mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
```
Aisuko, line 52: We provide the reformat tools under the project, https://github.com/meshery/meshery/blob/1ff876f10e0a9a0387bb84f96ca0a2edb1add6cd/Makefile#L90-L92 , please kindly check how to use this tool under localhost.

Aisuko (changes requested):
Please reformat the code style, and the tool below will help you.

mesheryctl/internal/cli/root/system/dashboard.go
Aisuko: If this does not work for you, please modify the environment `GO111MODULE` and `GOPROXY`

leecalcote:
Thanks for this review, @Aisuko. What do you think, @DantrazTrev?

leecalcote:
@Aisuko, perhaps, you'd like to carry this forward?

Aisuko:
@leecalcote No problem sir, I will do. Mid month I will move the home between two cities. I would schedule this after that.

navendu-pottekkat:
Thank you @Aisuko This needs some thought put into it to make sure that we get the correct endpoint. We would have to work to enhance the function in MeshKit.

navendu-pottekkat:
@Aisuko @leecalcote I'm closing this PR as this would be reworked.

Aisuko:
@navendu-pottekkat No problem.

Issue #3407, opened 2021-07-08
[mesheryctl] (child) `mesheryctl system` unit testing


piyushsingariya:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

check Epic here #3016 

#### Current Behavior
Currently, we don't have any unit tests setup to test `mesheryctl system` commands and functions

#### Desired Behavior
Required to enable unit testing support for `mesheryctl system` commands and functions, to ensure the accuracy and robustness of each mesheryctl release.

**mesheryctl Unit Testing**
- Mark all possible situations for each `mesheryctl system` commands and functions.
- Write tests using Golang’s standard library.
- A combination of CodeCov and GitHub Actions are to be used as mainstays in the approach to unit testing - https://github.com/codecov/codecov-action.

#### Subcommands to cover

- [x] mesheryctl system start
- [x] mesheryctl system stop
- [x] mesheryctl system context
- [x] mesheryctl system channel
- [x] mesheryctl system token
- [x] mesheryctl system logs
- [x] mesheryctl system reset
- [x] mesheryctl system restart
- [x] mesheryctl system status
- [x] mesheryctl system update
- [x] mesheryctl system config

---

#### Contributor Resources

- [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- [Layer5 Community Slack](http://slack.layer5.io)

piyushsingariya:
This issue can be split-up further for sub-commands under `mesheryctl system` subcommands and can be picked up by multiple devs for each file/command.

chetak123:
Hey @piyushsingariya can I take unit-test for system reset and restart please ?

sayantan1413:
Hey @piyushsingariya can I take the system start one?

piyushsingariya:
Yeah sure @chetak123, go ahead!
[...]

PR #3408, opened 2021-07-08
Issue #3401/Added field token-location in meshery system context view

### LINKED ISSUE #3401: [mesheryctl] Enhance `context view` to display context-specific token

This child issue is related to epic #3286 for new token command.

#### Current Behavior
Tokens are in process of being written into meshconfig (~/.meshery/config.yaml). See #3396.

#### Desired Behavior
Nearly all of the mesheryctl commands will need to account for and use the active context's token. `mesheryctl system context` is one of those. `mesheryctl system context view` should include all details the associated token.

#### Screenshots/Logs

Current implementation is missing token location.

```
➜  ~ mesheryctl system context view

Current Context: local

token: Default
platform: docker
adapters:
- meshery-istio
- meshery-linkerd
- meshery-consul
- meshery-nsm
- meshery-kuma
- meshery-cpx
- meshery-osm
- meshery-traefik-mesh
channel: stable
version: latest

➜  ~ cat ~/.meshery/config.yaml
contexts:
  local:
    token: Default
    platform: docker
    adapters:
    - meshery-istio
    - meshery-linkerd
    - meshery-consul
    - meshery-nsm
    - meshery-kuma
    - meshery-cpx
    - meshery-osm
    - meshery-traefik-mesh
    channel: stable
    version: latest
current-context: local
tokens:
- location: auth.json
  name: Default
```

#### Acceptance Tests

- [ ] User sees token name and location of active context when executing `mesheryctl system context view`.
  - [ ] Unit test should cover this scenario.
- [ ] User sees token name and location of specified context when executing `mesheryctl system context view --context <name>`.
  - [ ] Unit test should cover this scenario.
- [ ] Presence of a token in the context should be verified. If token name is not specified or is missing from context, then the user should be shown the contents of the context with a warning that token could not be found. 
  - [ ] Unit test should cover this scenario.
- [ ] Full definition of token in meshconfig file should be verified. If token name is provided in the context, but a token of that same name is not available in the `tokens:` stanza, then the user should be shown the contents of the context with a warning that token could not be found. 
  - [ ] Unit test should cover this scenario.

---

#### Contributor Resources

- [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)

Revolyssup:

**Description**
## Incase token field is missing altogether inside context

This PR fixes #3401 

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/internal/cli/root/system/context/view.go
+++ b/mesheryctl/internal/cli/root/system/context/view.go
@@ -48,10 +48,10 @@ var viewContextCmd = &cobra.Command{
 			for k, v := range configuration.Contexts {
 				if v.Token == "" {
 					log.Warnf("[Warning]: Token not specified/empty for context \"%s\"", k)
-					temp, _ := addTokenLocation(&v)
+					temp, _ := getContextWithTokenLocation(&v)
 					tempcontexts[k] = *temp
 				} else {
-					temp, ok := addTokenLocation(&v)
+					temp, ok := getContextWithTokenLocation(&v)
 					tempcontexts[k] = *temp
 					if !ok {
 						log.Warnf("[Warning]: Token \"%s\" could not be found! for context \"%s\"", tempcontexts[k].Token, k)
@@ -86,7 +86,7 @@ var viewContextCmd = &cobra.Command{
 			log.Printf("\nCurrent Context: %s\n", context)
 			log.Print(getYAML(contextData))
 		} else {
-			temp, ok := addTokenLocation(&contextData)
+			temp, ok := getContextWithTokenLocation(&contextData)
 			log.Printf("\nCurrent Context: %s\n", context)
 			if !ok {
 				log.Warnf("[Warning]: Token \"%s\" could not be found! for context \"%s\"", temp.Token, context)
@@ -109,7 +109,7 @@ func getYAML(strct interface{}) string {
 	return string(out)
 }
 
-func addTokenLocation(c *config.Context) (*contextWithLocation, bool) {
+func getContextWithTokenLocation(c *config.Context) (*contextWithLocation, bool) {
 	temp := contextWithLocation{
 		Endpoint:      c.Endpoint,
 		Token:         c.Token,


mesheryctl/internal/cli/root/config/config.go
```diff
@@ -34,12 +34,13 @@ type Token struct {
...
 type Context struct {
-	Endpoint string   `mapstructure:"endpoint,omitempty"`
-	Token    string   `mapstructure:"token,omitempty"`
-	Platform string   `mapstructure:"platform"`
-	Adapters []string `mapstructure:"adapters,omitempty"`
-	Channel  string   `mapstructure:"channel,omitempty"`
-	Version  string   `mapstructure:"version,omitempty"`
+	Endpoint      string   `mapstructure:"endpoint,omitempty"`
+	Token         string   `mapstructure:"token,omitempty"`
+	TokenLocation string   `mapstructure:"platform" yaml:"token-location,omitempty"`
```
piyushsingariya, line 39: @Revolyssup I don't this needs to be added here. Because we do not have this entry in config.yaml

piyushsingariya: The goal is to only print the Token Name and the Location so there's no point in adding this field and eventually this will break a lot of commands as Config.yaml won't be able to Unmarshal into this struct as we don't have this entry originally.

mesheryctl/internal/cli/root/system/context/view.go
```diff
@@ -23,8 +25,22 @@ var viewContextCmd = &cobra.Command{
...
+			tokenNameLocation[tok.Name] = tok.Location
+		}
 
 		if allContext {
+			for k, v := range configuration.Contexts {
+				if v.Token == "" {
+					log.Warnf("[Warning]: Token not specified/empty for context \"%s\"", k)
+					v.TokenLocation = ""
+				} else {
+					addTokenLocation(&v, k)
```
piyushsingariya, line 39: So instead of adding the location to the struct, we can directly print the config.Token with the name that has been stored in a map[string][config.Token]

See, the point is why tokens(name, and location) are separated from a context is that are reusable for multiple contexts what you implemented here doesn't align with that.

piyushsingariya: or you could do the same but with a temp struct not with editing the Config.Context struct.

piyushsingariya (changes requested):

mesheryctl/internal/cli/root/system/context/view.go
Revolyssup: @piyushsingariya  # I added an auxilliary struct instead of modifying the original context as you mentioned. 


## Incase the token field is left empty/or not mentioned in a specific context:


## Incase the token is not present inside the tokens: array


## When both cases individually in different contexts:

mesheryctl/internal/cli/root/system/context/view.go
```diff
@@ -60,3 +108,19 @@ func getYAML(strct interface{}) string {
 	out, _ := yaml.Marshal(strct)
 	return string(out)
 }
+
+func addTokenLocation(c *config.Context) (*contextWithLocation, bool) {
```
piyushsingariya, line 112:
```suggestion
func getContextWithTokenLocation(c *config.Context) (*contextWithLocation, bool) {
```
Can we rename it like this?

Revolyssup: @piyushsingariya  Changed the name as you suggested.

mesheryctl/internal/cli/root/system/context/view.go
leecalcote: I love seeing these test cases. Are these covered by unit tests, too?

Revolyssup: I don't think there are any test cases for any of the functions inside of context sub-command or even "meshery system" command in general as of now.  It has been pointed out in issue #3407

leecalcote: Thank you, @revolyssup. Batter-up, in that case. 🙂 

Note the Acceptance Tests for the https://github.com/meshery/meshery/issues/3401 issue.

Revolyssup: @leecalcote  For unit tests, currently there is a single test yaml inside pkg/utils, that is being used in "system channel" test.
I can write unit tests for my specific changes in "system context view", but I would also need some test yaml files for covering different cases.
How do you propose I move forward?
1. Concatenate all 4 test cases in a single new test yaml in pkg/utils and use that for my tests. (Cannot concatenate in already existing yaml as that yaml is being parsed in existing test as a struct and not slice of structs)
2. Create a sub directory inside pkg/utils and use some descriptive naming convention to create separate yaml files, as there can be a need of more test YAMLs to cover different cases in different commands and subcommands.

leecalcote: + @hexxdump

piyushsingariya: Wait for this weekend, I'll raise a PR that would help you set up the context with custom path. and you should cover that in a separate PR tbh.

leecalcote:
@Revolyssup let's chat through this one and get some more perspectives going. Will you present this at the Meshery Development Meeting this week?  // @Anita-ihuman

Revolyssup:
> @Revolyssup let's chat through this one and get some more perspectives going. Will you present this at the Meshery Development Meeting this week? // @Anita-ihuman
Yeah, sure.

navendu-pottekkat (approved):
LGTM, Thank you @Revolyssup 🚀

PR #3426, opened 2021-07-09
fix: Use `COPY` instead of `ADD` for files and folders

iamrajiv:
## Description

Although `ADD` and `COPY` are functionally similar, generally speaking, `COPY` is preferred. That’s because it’s more transparent than `ADD`. `COPY` only supports the basic copying of local files into the container, while `ADD` has some features (like local-only tar extraction and remote URL support) that are not immediately obvious.


leecalcote:
Thanks, @iamrajiv. It's good to have you here and contributing.

I wonder what value this change truly provides. Do we need to be able to extract tars or retrieve from remote URLs?

What isn't transparent about the ADD of files from `ui` to `ui` (for example)?

Not strictly required, but as a practice opening an issue first and having a suggested change/enhancement/bug confirmed before submitting a PR will spare you effort to the extent changes in the PR aren't warranted.

iamrajiv:
Hi, @leecalcote. No `COPY` can't handle remote URLs and can't auto-extract tar files. I thought we do not need it.

leecalcote:
@iamrajiv this is a good PR to review in the Meshery CI call this week (on Thursday) - https://meet.layer5.io. // @ramrodo

ramrodo:
> @iamrajiv this is a good PR to review in the Meshery CI call this week (on Thursday) - https://meet.layer5.io. // @ramrodo

Added to the [Agenda (July 22th)](https://docs.google.com/document/d/1GrVdGHZAYeu6wHNLLoiaKNqBtk7enXE9XeDRCvdA4bY/edit#)

bariqhibat (dismissed):
Yep this looks good, COPY function is more obvious than ADD, it is a better Docker practices.

bariqhibat:
Hey @iamrajiv, could you attend to the [DCO check](https://github.com/meshery/meshery/pull/3426/checks?check_run_id=3213215134) as well?

iamrajiv:
@bariqhibat I tried to sign in DCO but it is not passing I don't know why. For the last commit I did this `git commit -s -m"formatted readme"` but DCO failed.

bariqhibat:
> @bariqhibat I tried to sign in DCO but it is not passing I don't know why. For the last commit I did this `git commit -s -m"formatted readme"` but DCO failed.

Hi @iamrajiv, you need to make sure every single commit must be signed. In your case, the older commits are not signed so it's still showing the fail.

PR #3495, opened 2021-07-13
[mesheryctl] Add GET request step in HealthCheck

### LINKED ISSUE #3434: [mesheryctl] Fix `IsMesheryRunning()` for locally running meshery

#### Current Behavior

Currently many commands fail to execute because the function fails to detect if meshery is running locally or not.

#### Desired Behavior

It should detect, ofcourse!
Send a GET request to http://localhost:9081/api/version ? If the response is 200 then it is running and reachable or else it is not. and that means Meshery is might be running local

---

#### Contributor Resources

- [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)

alphaX86:

**Description**
This PR adds a GET request step in process to check if Meshery is running or not

This PR fixes #3434 

**Notes for Reviewers**
Tested on Linux (Ubuntu) no errors caught so far. If any changes needed, report them so that I can fix it

### DIFF
--- a/mesheryctl/pkg/utils/healthcheck.go
+++ b/mesheryctl/pkg/utils/healthcheck.go
@@ -125,7 +125,7 @@ func IsMesheryRunning(currPlatform string) (bool, error) {
 
 	urlEndpoint := currCtx.GetEndpoint()
 
-	urlTest := urlEndpoint + "/api/server/version"
+	urlTest := urlEndpoint + "/api/system/version"
 
 	// Checking if Meshery is running with the URL obtained
 	resp, _ := http.Get(urlTest)


alphaX86:
Oh my bad.... :) fixing lint issues

alphaX86:
@piyushsingariya this works only if port is 9081, so if a user has other port assigned by changing config, then how we can check and change the port?

piyushsingariya:
@alphaX86 `make run-local` and `make run-fast` doesn't fetch the port from config.yaml, in case of these commands it is hardcoded to run meshery on port 9081.

alphaX86:
Agreed, I'm asking in general though actually

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -114,6 +115,12 @@ func parseKubectlShortVersion(version string) ([3]int, error) {
 
 // IsMesheryRunning checks if the meshery server containers are up and running
 func IsMesheryRunning(currPlatform string) (bool, error) {
+	resp, err := http.Get("http://localhost:9081/api/server/version")
+	// Port number is subject to change depending on user's configuration on config.yaml
+
+	if resp != nil {
```
piyushsingariya, line 121: Better to check resp.StatusCode==200!!

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -155,6 +162,10 @@ func IsMesheryRunning(currPlatform string) (bool, error) {
 		}
 	}
 
+	if err != nil {
+		return false, errors.Wrap(err, "Meshery not running locally")
+	}
+
```
piyushsingariya, line 168: No need to add this thing, we are returning False status in the next line.

piyushsingariya (changes requested):

mesheryctl/pkg/utils/healthcheck.go
alphaX86: Agreed but that Golang rules... 🙂 Once declared, the variable has to be used right?

mesheryctl/pkg/utils/healthcheck.go
alphaX86: Got it 👍

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -114,6 +115,12 @@ func parseKubectlShortVersion(version string) ([3]int, error) {
 
 // IsMesheryRunning checks if the meshery server containers are up and running
 func IsMesheryRunning(currPlatform string) (bool, error) {
+	resp, err := http.Get("http://localhost:9081/api/server/version")
```
piyushsingariya, line 118:
```suggestion
	resp, _ := http.Get("http://localhost:9081/api/server/version")
```
Mark it like this.

piyushsingariya: Also, add a comment above this
// Checking if Meshery is running with "make run-local" or "make run-fast"

piyushsingariya (changes requested):

mesheryctl/pkg/utils/healthcheck.go
piyushsingariya: add a comment over this,
// checking if meshery is running with "make run-local" or "make run-fast"

alphaX86: Hmm alright! That will do 👍

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -115,10 +115,10 @@ func parseKubectlShortVersion(version string) ([3]int, error) {
 
 // IsMesheryRunning checks if the meshery server containers are up and running
 func IsMesheryRunning(currPlatform string) (bool, error) {
-	resp, err := http.Get("http://localhost:9081/api/server/version")
-	// Port number is subject to change depending on user's configuration on config.yaml
+	resp, _ := http.Get("http://localhost:9081/api/server/version")
+	// Checking if Meshery is running with "make run-local" or "make run-fast"
```
piyushsingariya, line 119: shift it above http.GET

alphaX86:
This is new! 👀 @piyushsingariya and @navendu-pottekkat the CI action reports as invalid memory reference... Any ideas on how to fix it?

mesheryctl/pkg/utils/healthcheck.go
alphaX86: Done

alphaX86:
Totally weird 👀 even after changes... seems I need to test it again

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -114,6 +115,12 @@ func parseKubectlShortVersion(version string) ([3]int, error) {
 
 // IsMesheryRunning checks if the meshery server containers are up and running
 func IsMesheryRunning(currPlatform string) (bool, error) {
+	// Checking if Meshery is running with "make run-local" or "make run-fast"
+	resp, _ := http.Get("http://localhost:9081/api/server/version")
+
+	if resp.StatusCode == 200 {
```
piyushsingariya, line 121:
```suggestion
	if resp != nil && resp.StatusCode == 200 {
```
This is the nil reference thing.

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -114,6 +115,12 @@ func parseKubectlShortVersion(version string) ([3]int, error) {
 
 // IsMesheryRunning checks if the meshery server containers are up and running
 func IsMesheryRunning(currPlatform string) (bool, error) {
+	// Checking if Meshery is running with "make run-local" or "make run-fast"
+	resp, _ := http.Get("http://localhost:9081/api/server/version")
```
leecalcote, line 119: The hostname and port cannot be hardcoded, but must use a central reference for these values.

leecalcote (changes requested):

mesheryctl/pkg/utils/healthcheck.go
piyushsingariya: @leecalcote Check here https://github.com/meshery/meshery/blob/master/Makefile#L66

alphaX86: @leecalcote that's what I felt when I saw this, on public version, the port number may tend to change depending on user config. But here we check locally using `make run-local` or `make run-fast` (which works only on port 9081 or desired port)

IDK if it's ok for now

leecalcote: Users don't deploy using the Makefile. The hostname and port cannot be hardcoded, but must use a central reference for these values.

alphaX86: Agreed, so à placeholder variable then in this case?

piyushsingariya: @leecalcote this is being added for a specific case i.e. Development Mode of Meshery Server.

If some developer made some changes in server-side and is testing out a mesheryctl command (e.g. `mesheryctl system login`) which needs some confirmation that IsMesheryRunning(), which will eventually fail because of the only platform based checking and it won't possible to test the command with new changes in server. that is why the hardcoded part.

alphaX86: So..... are we good to go or still needs to be looked on this?

alphaX86:
It's been a week... Can anyone say if this PR is good or still has a conflict?

piyushsingariya:
> It's been a week... Can anyone say if this PR is good or still has a conflict?

@leecalcote @navendu-pottekkat

alphaX86:
Rebasing is really a problem kind :/

mesheryctl/pkg/utils/healthcheck.go
leecalcote: Unless the hardcoded value used in this check only applies to developers who are compiling `mesheryctl` and is not enforced for users who are running Meshery at an unknown endpoint (a dynamically defined endpoint), no, we cannot proceed until hardcoded `localhost` is removed.

alphaX86: OK........... Makes sense at first look at the problem. So the dynamic way then :)

alphaX86:
I have added the variables which will extract the URL endpoint from config file and tested it. Worked pretty much

alphaX86:
Is this PR good to go? Or still needs some changes?

piyushsingariya (approved):

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -114,6 +118,23 @@ func parseKubectlShortVersion(version string) ([3]int, error) {
...
 // IsMesheryRunning checks if the meshery server containers are up and running
 func IsMesheryRunning(currPlatform string) (bool, error) {
+	// Get viper instance used for context to extract the endpoint from config file
+	mctlCfg, _ := config.GetMesheryCtl(viper.GetViper())
+
+	currCtx, _ := mctlCfg.GetCurrentContext()
+
+	urlEndpoint := currCtx.GetEndpoint()
+
+	urlTest := urlEndpoint + "/api/server/version"
```
piyushsingariya, line 128:
```suggestion
	urlTest := urlEndpoint + "/api/system/version"
```
Switch it to system, we are switching the URL in another PR.

piyushsingariya (changes requested):

mesheryctl/pkg/utils/healthcheck.go
alphaX86: Oh ok then! 👍

alphaX86: It's resolved

piyushsingariya (approved):

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -114,6 +118,23 @@ func parseKubectlShortVersion(version string) ([3]int, error) {
...
 // IsMesheryRunning checks if the meshery server containers are up and running
 func IsMesheryRunning(currPlatform string) (bool, error) {
+	// Get viper instance used for context to extract the endpoint from config file
+	mctlCfg, _ := config.GetMesheryCtl(viper.GetViper())
+
+	currCtx, _ := mctlCfg.GetCurrentContext()
+
+	urlEndpoint := currCtx.GetEndpoint()
+
+	urlTest := urlEndpoint + "/api/system/version"
```
leecalcote, line 128: Looking real good, @alphaX86!

leecalcote (approved):

Issue #3487, opened 2021-07-12
remote_provider.go


kumarabd:

Shreyas220:
@sayantan1413 can i take this

Issue #3486, opened 2021-07-12
remote_auth.go


kumarabd:

Shreyas220:
@sayantan1413 can i take this

PR #3528, opened 2021-07-16
Migrated errors for models package #2

dhruv0000:
**Description**

This PR fixes:
#3485 
#3486 
#3487 

**Notes for Reviewers**


models/error.go
```diff
@@ -76,6 +107,42 @@ func ErrUnmarshal(err error, obj string) error {
...
+
+func ErrDelete(err error, obj string, statusCode int) error {
+	return errors.New(ErrDeleteCode, errors.Alert, []string{"Unable to delete data from the Provider", obj}, []string{"Status Code: " + fmt.Sprint(statusCode), err.Error()}, []string{}, []string{})
+}
+
+func ErrDecodeBase64(err error, obj string) error {
+	return errors.New(ErrDecodeBase64Code, errors.Alert, []string{"Error occurred while decoding base65 string", obj}, []string{err.Error()}, []string{}, []string{})
+}
+
+func ErrMarshalPKIX(err error) error {
```
Shreyas220, line 130: Everything looks great, but just one question we already have ErrMarshal() 
Was there a specific reason to create this function ?

kumarabd: @Shreyas220 great catch. @dhruv0000 We will want to reuse the Marshal function.

dhruv0000: yes. I did thought of it, but this error is specifically for converting a public key to PKIX (X.509) format and not JSON, so I thought it deserved its own error type.

kumarabd: If the statement and everything is the same, then irrespective of the use case we could re-use.

PR #3544, opened 2021-07-17
Add support for custom callback URL in meshery server

tangledbytes:
**Description**

This PR fixes https://github.com/layer5labs/meshery-extensions-packages/issues/8

This package adds support for an environmental variable `MESHERY_SERVER_CALLBACK_URL` which will allow the user to customize the provider callback URL (eg if meshery is running behind  a proxy).

The support has been added in two places:
1. Meshery Server if gets `MESHERY_SERVER_CALLBACK_URL` env var then it will use this as callback URL. The default is `{MESHERY_HOST}/api/user/token
2. `mesheryctl` also supports the same environment variable in the following commands:
    1. `MESHERY_SERVER_CALLBACK_URL=http://something.com/meshery/api/user/token mesheryctl system start` this will modify the k8s manifest (if context platform is k8s) or compose file (if context platform is docker) to pass this env var to the meshery server.
    2. `mesheryctl system reset` will **not** preserve the callback URL, however if `MESHERY_SERVER_CALLBACK_URL` is provided here then that will be respected.
    3. `mesheryctl system restart` will respect the callback URL set in the `start` command.

**Notes for Reviewers**
New utility functions related to YAML transformation and nested map transformation are added in this PR along with unit tests. These utilities can be used in the mesheryctl in general as a lot of mesheryctl performs only YAML transformations making the code too verbose (whenever the maps are nested).

### DIFF
--- a/mesheryctl/internal/cli/root/system/start.go
+++ b/mesheryctl/internal/cli/root/system/start.go
@@ -173,9 +173,7 @@ func start() error {
 			spliter := strings.Split(temp.Image, ":")
 			temp.Image = fmt.Sprintf("%s:%s-%s", spliter[0], currCtx.GetChannel(), "latest")
 			if v == "meshery" {
-				mesheryServerCallbackURL := "MESHERY_SERVER_CALLBACK_URL"
-
-				temp.Environment = append(temp.Environment, fmt.Sprintf("%s=%s", mesheryServerCallbackURL, viper.GetString(mesheryServerCallbackURL)))
+				temp.Environment = append(temp.Environment, fmt.Sprintf("%s=%s", "MESHERY_SERVER_CALLBACK_URL", viper.GetString("MESHERY_SERVER_CALLBACK_URL")))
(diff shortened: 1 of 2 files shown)


leecalcote:
Viper doesn’t cover the yaml transformation use cases, huh?

utkarsh-pro:
> Viper doesn’t cover the yaml transformation use cases, huh?

No. Viper can "get" and "set" only the root level keys, not the nested keys. But a lot mesheryctl actually does that, hence the code is too verbose.

This new utility function was created to allow mutation at any level in the YAML without manually dealing with go maps.

kumarabd (approved):

leecalcote:
@utkarsh-pro conflict in start.go - https://github.com/meshery/meshery/pull/3544/conflicts

leecalcote:
The latest build failures seem to be because of our upgrade to golang 1.116

mgfeller:
> The latest build failures seem to be because of our upgrade to golang 1.116

@leecalcote https://github.com/meshery/meshery/pull/3545#issuecomment-884006803

utkarsh-pro:
Merging this PR...

PR #3548, opened 2021-07-17
Add settings page path to exportPathMap

### LINKED ISSUE #3547: [UI] Settings page returns 404 while reloading
#### Description
Settings page works fine when we use the **gear** button in navigation menu, but explicitly going to 'localhost:9081/settings' returns 404. Also, after going to the settings page, reloading it returns 404 as well.

#### Expected Behavior
It should not return 404 in any case, be it, going to 'localhost:9081/settings' or reloading.

**Contributor Resources:**

humblenginr:
**Description**

Added settings page path to exportPathMap in next config file so that, while exporting, it will build settings page as well so that meshery server will be able to serve the settings page when navigating to '/settings'.

This PR fixes #3547 

### DIFF
--- a/ui/next.config.js
+++ b/ui/next.config.js
@@ -12,6 +12,7 @@ module.exports = compose([
         '/configure': { page: '/configure' },
         '/playground': { page: '/playground' },
         '/provider/*': { page: '/provider' },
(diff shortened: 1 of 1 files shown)


leecalcote (approved):
Looks good.

PR #3669, opened 2021-07-27
[CI] Fixed the lint failure

sayantan1413:
Signed-off-by: sayantan1413 <sayantanbose2001@gmail.com>

**Description**

Fixed the lint failure

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/internal/cli/root/system/start.go
+++ b/mesheryctl/internal/cli/root/system/start.go
@@ -109,9 +109,9 @@ func start() error {
 		currCtx.SetPlatform(utils.PlatformFlag)
 	} else {
 		if utils.CfgFile != utils.DefaultConfigPath {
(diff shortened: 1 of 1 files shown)


leecalcote:
This silly lint issue is going to cause our error output to be harder to read. @sayantan1413 do you mind looking into how we might set certain lint checks to be ignored?

sayantan1413:
Yup @leecalcote sure I will look into this.

mgfeller:
> Yup @leecalcote sure I will look into this.

I think it would be good to have a linter rule / config file template, e.g. in this repo, that can be used in other repos as well (or at least form the basis). That way, we don't have to consider and discuss silly and not-so-silly rules for each repo. I thought the previous set was quite good and appropriate, but as `revive` should up the config has to be adjusted. Also, as `revive` has more rules than `golint`, there might be overlap between `revive` and other linters in use. 

Similar work is done in meshkit. https://github.com/layer5io/meshkit/pull/89

// @sayantan1413 @leecalcote @iamrajiv

PR #3562, opened 2021-07-19
added check for docker or kubernetes platform

nehaLohia27:
Signed-off-by: Neha Lohia <nehapithadiya444@gmail.com>

**Description**

This PR fixes #3444 

**Notes for Reviewers**

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [ ] Yes, I signed my commits.

PR #3675, opened 2021-07-28
system start fails with empty platform flag

### LINKED ISSUE #3673: [mesheryctl] `system start` fails
#### Expected Behavior

`system start` fails when not passing -p flag. If flag hasn't passed platform should be fetched from config.

#### Screenshots/Logs

```
❯ ./mesheryctl system start
Error: failed to start Meshery
See https://docs.meshery.io/reference/mesheryctl/system for usage details
: the platform '' is not supported. Supported platforms are:

- docker
- kubernetes

Verify this setting in your meshconfig at /home/nectar/.meshery/config.yaml or verify by executing `mesheryctl system context view`
```

#### Environment

- Host OS: Mac Linux Windows

---

**Contributor Resources:**

nehaLohia27:
Signed-off-by: Neha Lohia <nehapithadiya444@gmail.com>

**Description**

This PR fixes #3673 

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/internal/cli/root/system/context/context.go
+++ b/mesheryctl/internal/cli/root/system/context/context.go
@@ -26,10 +26,10 @@ import (
 	"github.com/spf13/viper"
 )
 
-var configuration *config.MesheryCtlConfig
-var tempContext string = "local"
 var (
 	availableSubcommands []*cobra.Command
+	configuration        *config.MesheryCtlConfig
+	tempContext          = "local"
 )
 
 // ContextCmd represents the context command


mesheryctl/internal/cli/root/system/start.go
```diff
@@ -105,6 +105,9 @@ func start() error {
 		return err
 	}
 
+	if utils.PlatformFlag == "" {
+		utils.PlatformFlag = currCtx.GetPlatform()
+	}
 	if utils.PlatformFlag == "docker" || utils.PlatformFlag == "kubernetes" {
 		currCtx.SetPlatform(utils.PlatformFlag)
 	} else {
```
piyushsingariya, line 113: better to cover these if-else, in another if statement
 `if utils.PlatformFlag != "" {}`

navendu-pottekkat:
@nehaLohia27 There is lint warning on the lines surrounding this change. See your [earlier PR](https://github.com/meshery/meshery/pull/3562/files), please fix that as well.

mesheryctl/internal/cli/root/system/start.go
leecalcote: Piyush, let’s get this fixed urgently. Will you “suggest” this fix with CTRL+G?

nehaLohia27: @leecalcote @piyushsingariya I made the final changes. Can u check once.

nehaLohia27:
@navendu-pottekkat The lint errors are already fixed and merged in earlier PR. This PR does not have lint issues.

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -105,13 +105,18 @@ func start() error {
 		return err
 	}
 
-	if utils.PlatformFlag == "docker" || utils.PlatformFlag == "kubernetes" {
-		currCtx.SetPlatform(utils.PlatformFlag)
-	} else {
-		if utils.CfgFile != utils.DefaultConfigPath {
+	if utils.PlatformFlag == "" {
+		utils.PlatformFlag = currCtx.GetPlatform()
+	}
```
piyushsingariya, line 110: this is not required now. as we are already checking if the platform flag is empty or not ahead .

piyushsingariya (changes requested):

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -108,13 +108,15 @@ func start() error {
 	if utils.PlatformFlag == "" {
 		utils.PlatformFlag = currCtx.GetPlatform()
 	}
-	if utils.PlatformFlag == "docker" || utils.PlatformFlag == "kubernetes" {
-		currCtx.SetPlatform(utils.PlatformFlag)
-	} else {
-		if utils.CfgFile != utils.DefaultConfigPath {
+	if utils.PlatformFlag != "" {
```
hexxdump, line 111: So currCtx.GetPlatform() failed and the `mesheryctl system start` command was run without the -p flag. 
This  might not be the very first instance of `mesheryctl system start` as bash (one-liner) script takes platform argument. Is the installer script run in a different way? Or
Someone intentionally emptied out the platform value from current context?

mesheryctl/internal/cli/root/system/start.go
```diff
@@ -105,13 +105,18 @@ func start() error {
...
+		if utils.PlatformFlag == "docker" || utils.PlatformFlag == "kubernetes" {
+			currCtx.SetPlatform(utils.PlatformFlag)
+		} else {
+			if utils.CfgFile != utils.DefaultConfigPath {
+				return fmt.Errorf("the platform '%s' is not supported. Supported platforms are:\n\n- docker\n- kubernetes\n\nVerify this setting in your meshconfig at %s or verify by executing `mesheryctl system context view`", utils.PlatformFlag, utils.CfgFile)
+			}
 			return fmt.Errorf("the platform '%s' is not supported. Supported platforms are:\n\n- docker\n- kubernetes\n\nVerify this setting in your meshconfig at %s or verify by executing `mesheryctl system context view`", utils.PlatformFlag, utils.CfgFile)
 		}
-		return fmt.Errorf("the platform '%s' is not supported. Supported platforms are:\n\n- docker\n- kubernetes\n\nVerify this setting in your meshconfig at %s or verify by executing `mesheryctl system context view`", utils.PlatformFlag, utils.CfgFile)
 	}
```
hexxdump, line 120: we still need to handle the case of utils.PlatformFlag == "" case, right?

mesheryctl/internal/cli/root/system/start.go
piyushsingariya, line 110: @nehaLohia27  This part is unnecessary, as we have already maintained this thing later ahead.

mesheryctl/internal/cli/root/system/start.go
piyushsingariya: actually we are reversing the case we check if the flag's value != "". then we can check if it is docker or kubernetes and if it is none of them we return err.

piyushsingariya: and default case for the flag value would be "" empty string in that case we will fetch from the config file.

mesheryctl/internal/cli/root/system/start.go
nehaLohia27: @piyushsingariya I removed this code and pushed changes

nehaLohia27: @piyushsingariya one lint error is coming.   
Warning: var-declaration: should omit type string from declaration of var tempContext; it will be inferred from the right-hand side (revive)

this is the code snippet. 
	if tempContext != "" {
		err = mctlCfg.SetCurrentContext(tempContext)
		if err != nil {
			return errors.Wrap(err, "failed to set temporary context")
		}
	}
any idea?

navendu-pottekkat: `tempContext` we do not need to explicitly define the data type of this variable as it will be figured out from its initial assignment.

piyushsingariya (approved):

PR #3681, opened 2021-07-28
meshery: Fix read operation not waiting for user input

### LINKED ISSUE #3680: Meshery:  `curl -L https://git.io/meshery | bash -` doesn't wait for input
#### Description

When we are installing or updating meshery with command ` curl -L https://git.io/meshery | bash -` the terminal is not waiting for  input but fills it with empty value (because of use of curl command) and fails at the end.
Screenshot below :point_down: 

image

#### Expected Behavior

It should prompt me for the platform input. It was also pointed out by @samirparhi-dev in today's meeting, i think.

#### Environment

- Host OS:  Linux 
- Platform: Docker and Kubernetes

---

Abhishek-kumar09:

**Description**
When we are installing or updating meshery with command curl -L https://git.io/meshery | bash - the terminal is not waiting for input but fills it with empty value (because of use of curl command) and fails at the end.

The possible reason is using `pipe` command here due to which the newline character is automatically added from stdIn, I think.
The operation is modified to take input from active terminal device. i.e. `/dev/tty`
 

**This PR fixes #3680**

After change: 
image

**Notes for Reviewers**

### DIFF
--- a/meshery
+++ b/meshery
@@ -45,7 +45,7 @@ done
 # If no flag is present, prompt the user for a platform
 if [ ! "$PLATFORM" ]; then
     echo Enter a platform to deploy Meshery. Available platforms [docker, kubernetes]:
-    read PLATFORM
+    read PLATFORM < /dev/tty
 fi
 
 case "$PLATFORM" in


leecalcote (approved):

leecalcote:
@hexxdump fyi

PR #3744, opened 2021-08-02
fix: Use `COPY` instead of `ADD` for files and folders

iamrajiv:
## Description

Although `ADD` and `COPY` are functionally similar, generally speaking, `COPY` is preferred. That’s because it’s more transparent than `ADD`. `COPY` only supports the basic copying of local files into the container, while `ADD` has some features (like local-only tar extraction and remote URL support) that are not immediately obvious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ See the [getting started](https://meshery.io/#getting-started) section to quickl
 [Meshery documentation](https://docs.meshery.io/installation) offers thorough installation guides for your platform of choice.
 
  </details>
- 
+
 <h2><a name="service-meshes"></a>Supported Service Meshes</h2>
 <p>Meshery supports 10 diffferent service meshes.</p>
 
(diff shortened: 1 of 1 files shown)


iamrajiv:
@bariqhibat Please check this PR the last PR was close automatically when I did force push.

bariqhibat (approved):
Looks good to me!

iamrajiv:
@leecalcote

leecalcote:
I'm still looking for a compelling reason to switch. Any change to the project is a risk. Without any stated benefit other than this is a best practice and provides additional transparency.... we haven't hit the threshold for risk vs reward here.

Instead, I would love to see PRs that are dropping the image size of Meshery Server. _That_ is very much needed.

leecalcote:
The addition of swagger / api docs just added another ~200MB, I think.

PR #3748, opened 2021-08-02
[GraphQL] Added Token to Req context

dhruv0000:
Signed-off-by: dhruv0000 <patel.4@iitj.ac.in>

**Description**

GraphQL resolvers should reuse the methods defined in `provider` interface. To do that, they need access to the token. This PR:
1. Addes the token to the req context (which resolver and handler both can access).
2. Updates `GetResults` method of `provider` interface as an example on how to make the methods generic for GraphQL to use.

**Notes for Reviewers**

### DIFF
--- a/handlers/fetch_results_handler.go
+++ b/handlers/fetch_results_handler.go
@@ -103,7 +103,9 @@ func (h *Handler) GetResultHandler(w http.ResponseWriter, req *http.Request, _ *
 		return
 	}
 
-	bdr, err := p.GetResult(req, key)
+	tokenString := req.Context().Value("token").(string)
+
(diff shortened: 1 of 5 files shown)


dhruv0000:
@bariqhibat ^^

leecalcote:
In context of the Making GraphQL Solid plan, where does this land us?

dhruv0000:
> In context of the Making GraphQL Solid plan, where does this land us?

On track. The Provider interface needs to become more generic in order for it to be used by both resolvers and handlers. This is just an example on how to do it.

utkarsh-pro (approved):
LGTM

leecalcote (approved):

PR #3750, opened 2021-08-03
[UI] Change the dev server script to use node rather than nodemon

humblenginr:
**Changes being introduced:**

Previously, for starting the development server for meshery ui, the script that was used was `eslint . && nodemon ui_dev_server.js`

Now, i have changed it to `eslint . && node ui_dev_server.js`

**Why?**

`ui_dev_server.js` introduces a proxy on top of Nextjs development server so that the requests that are to be handled by the backend are forwarded to `localhost:9081`.

Now, Nextjs development server by default uses _Hot Module Replacement(HMR)_ to provide amazing developer experience by giving _almost_ instantaneous feedback on the changes that they have made.

But in our case, since we are using `nodemon`, it _restarts_ the server each and every time changes has been made(it basically rebuilds the UI, whereas, in case of HMR, it only changes the modules that have changed), so it creates a very bad developer experience that the developers have to wait for around 2 mins to get the feedback on the changes that they have made, even if the changes are very small like changing a colour or style, which, can sometimes be frustating.

So, i have changed the script to use `node` rather than `nodemon`, so that we are able to get very fast feedback on the changes that we have made

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -4,7 +4,7 @@
   "description": "the service mesh management plane",
   "main": "index.js",
   "scripts": {
-    "dev": "eslint . && nodemon ui_dev_server.js",
+    "dev": "eslint . && node ui_dev_server.js",
     "lint": "eslint --fix .",
     "next": "next",
     "build": "next build",


leecalcote:
Makes sense to me. This sounds like a significant improvement.

leecalcote (approved):

sudo-NithishKarthik:
Yes it is. It would drastically improve productivity 🚀 
It has been bugging me for quite sometime and i was not able to tolerate it lately 😫  , so, i thought i have to fix it somehow.

So, finally, it's just _3 characters_ that has been giving us a hard time 😂

PR #3805, opened 2021-08-06
Refactor Patterns and add some features

tangledbytes:
**Description**

This PR adds the following:
1. Support for `fillers`, i.e runtime expression evaluation in the patterns. These expressions can be used for:
    1. Embedding resources IDs which meshery server generates in the pattern while keeping the patterns free if infrastructure specific info.
    2. Reduce the risk of typos by referencing once declared values throughout the file.
2. Patterns are now evaluated in stages. The following stages are added:
    1. Service Identifier - This stage identifies services and add IDs to them.
    2. Filler - Expressions are evaluated at this stage.
    3. Validation - Patterns are evaluated for their validity and correctness. This stage leverages a new construct called `"selector"` which helps selecting appropriate workload/trait definitions in case of conflicts. The selector is in its initial stages.
    4. Provision - Provision stage will provision or de-provision the resources requested in the pattern. This step relies on DAG that is generated internally and hence is concurrent in nature. Inversion of plans are also introduced to serve more reliable "deletion" of the resources.
    5. Persist - Persist stage will persist the state of the resources which were successfully processed in the previous stage.
3. Capabilities store has been enhanced to support multiple capabilities if they differ in any aspect. `md5` hash is computed to compare the capabilities. This change adds support for:
    1. Registering resources for multiple service mesh version with same name
    2. Registering resources for multiple adapters (the support is limited at the moment)
4. Support for `Application` has been enhanced to now support:
    1. Passing environmental variables
    2. Advanced settings which allow users to opt out from auto service creation, placing custom labels and annotations to the resources. This is under "advanced" because tinkering with these setting may result in undesirable behaviour unless done properly.
5. Support for native k8s service has been added.
6. Pattern files to cytoscape and vice versa conversions now support persisting the edges data.

**Notes for Reviewers**
Apologies for 2k+ LOC PR :sweat_smile: 


leecalcote:
@sudo-NithishKarthik what do you think about https://github.com/meshery/meshery/pull/3805/checks?check_run_id=3264826056#step:5:274?

sudo-NithishKarthik:
#3803 will fix it @leecalcote , it's because we removed the dashboard icon in the header, this test spec is checking for the visibility of that button, that's why it fails.

leecalcote:
Thank you, @sudo-NithishKarthik. I've sync'ed this branch with master.

leecalcote:
@utkarsh-pro a couple items to look at - https://github.com/meshery/meshery/pull/3805/checks?check_run_id=3266038615#step:3:73

utkarsh-pro:
@leecalcote, fixed the Unit Test issue. It was a flaky test, fixed it. Addressed the golangci-lint issue too

leecalcote:
Adding the notes from this PR to the architecture deck....

PR #3819, opened 2021-08-08
[UI]] [Graphql] Use a single web socket connection for all graphql subscriptions

humblenginr:
**Description**
Meshery UI was using different websocket connections for each graphql subscription which is not effective since it will have to establish a websocket connection every time it has to subscribe to an event.

This pr changes that behaviour so that it now creates a single websocket connection and uses it for all graphql subscriptions.

This PR fixes #3782 

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #3822, opened 2021-08-09
[UI] Implement proxying for websocket requests in development server

humblenginr:
**Description**
This pr implements proxying for the websocket connections while running development server

**Notes for Reviewers**

### DIFF
--- a/ui/ui_dev_server.js
+++ b/ui/ui_dev_server.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
 const { createServer } = require('http')
 const { parse } = require('url')
 const next = require('next')
@@ -40,7 +41,10 @@ app.prepare().then(() => {
 
   })
   server.on("upgrade", (req,socket,head) => {
-    proxy.ws(req,socket, head)
+    proxy.ws(req,socket, head, err => {
+      socket.write("HTTP/" + req.httpVersion + " 500 Connection error\r\n\r\n");
+      socket.end();
+    })
   })
 
   server.listen(port, err => {


leecalcote (approved):

leecalcote:
@dhruv0000 @bariqhibat what do you think?

bariqhibat:
> @dhruv0000 @bariqhibat what do you think?

I'm not exactly sure about what value it adds to the system. From what I read, it seems like websocket is unpopular alright, but using GraphQL, it is completely abstracted away and we don't actually feel any difference between websocket and http protocol. 
And, from what I read (again, haha this is not something that I know much), the http proxy seems to be using a load-balancing  software (think NGINX) instead of handling it on application level

sudo-NithishKarthik:
> I'm not exactly sure about what value it adds to the system

ui/ui_dev_server.js
```diff
@@ -8,7 +9,14 @@ const app = next({ dev })
 const handle = app.getRequestHandler()
 var httpProxy = require('http-proxy');
 
-var proxy = httpProxy.createProxyServer({});
+var proxy = httpProxy.createProxyServer({
+  target: {
+    host: "localhost",
```
leecalcote, line 14: Avoid this hardcoding, if we can.

leecalcote:
> > @dhruv0000 @bariqhibat what do you think?
> 
> I'm not exactly sure about what value it adds to the system. From what I read, it seems like websocket is unpopular alright, but using GraphQL, it is completely abstracted away and we don't actually feel any difference between websocket and http protocol.
> And, from what I read (again, haha this is not something that I know much), the http proxy seems to be using a load-balancing software (think NGINX) instead of handling it on application level

Nginx isn’t involved here. Graphql does abstract away websockets (websockets are what Graphql is using underneath).

leecalcote:
Does our chosen graphql server library not provide socket reuse and reconnect functions?

sudo-NithishKarthik:
> I'm not exactly sure about what value it adds to the system 

So, right now, when we run `npm run dev` we are using a proxy on top of next js development server that listens on port `3000` which proxies the appropriate requests to the backend server which would be running in port `9081` .

In the present state of our code in `ui_dev_server.js`, we have the logic to proxy _http_ requests but not for ws connections, as a result of that, all requests for graphql subscription initiation will not be responded with.

Screenshot 2021-08-09 at 4 41 27 PM

As a result of which, it will keep retrying to establish a connection but will never be able to succeed.

So, in this pr, i have added the logic for proxying websocket connections to `ws://localhost:9081`

Screenshot 2021-08-09 at 4 45 48 PM

So that the subscriptions are successfully initialised.

sudo-NithishKarthik:
> > > @dhruv0000 @bariqhibat what do you think?
> > 
> > 
> > I'm not exactly sure about what value it adds to the system. From what I read, it seems like websocket is unpopular alright, but using GraphQL, it is completely abstracted away and we don't actually feel any difference between websocket and http protocol.
> > And, from what I read (again, haha this is not something that I know much), the http proxy seems to be using a load-balancing software (think NGINX) instead of handling it on application level
> 
> Nginx isn’t involved here. Graphql does abstract away websockets (websockets are what Graphql is using underneath).

yup

sudo-NithishKarthik:
> Does our chosen graphql client library not provide socket reuse and reconnect functions?

It does, that is why it keeps reconnecting when it fails and as of #3819 , it will reuse the same socket for all subscriptions.

sudo-NithishKarthik:
**NOTE**
 This does not in any way influence the behaviour of meshery server or ui, it just fixes an issue with development server

ui/ui_dev_server.js
sudo-NithishKarthik: The only scenario i can think of where it might change is when we do not run meshery server in `localhost:9081` and i don't think someone who wants to just contribute will do that 🤔 
What do you think ?

bariqhibat:
> So that the subscriptions are successfully initialised ☝🏽

I see, I can see the problem now. This is a good work.

ui/ui_dev_server.js
dhruv0000: `ui_dev_server` is only used for development purposes, and anyways the UI dev server has no way to know the address of the Meshery server, so no way around it.

Looks good Karthik.

dhruv0000 (approved):

PR #3788, opened 2021-08-05
Changed docker compose default location

metonymic-smokey:
**Description**
This PR changes the default location of the docker compose file and also create a manifests folder similar to the .meshery folder to accomadate it. 

This PR fixes #3755 

**Notes for Reviewers**

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #3824, opened 2021-08-09
Revert "Changed docker compose default location"

pottekkat:
Reverts meshery/meshery#3788

Causes the names of the docker-compose file to change causing the streaming logs. We need to remove hardcoding and prevent docker from changing the container names.

### DIFF
--- a/mesheryctl/internal/cli/root/system/start.go
+++ b/mesheryctl/internal/cli/root/system/start.go
@@ -86,12 +86,6 @@ func start() error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


navendu-pottekkat:
@metonymic-smokey FYI^

navendu-pottekkat:
This was not an issue from your end but it was an unseen consequence of that change. We need to approach it differently.

metonymic-smokey:
alright @navendu-pottekkat!
Which hardcoding are you referring to though?

navendu-pottekkat:
@metonymic-smokey where we check the meshery_meshery_1 container in start.go and decide whether or not to show the logs.

PR #3844, opened 2021-08-11
[WIP][CI/CD] change the mesheryctl from zip package to binary

### LINKED ISSUE #3756: [mesheryctl] bash installer script fails when unzip command is missing on system
#### Description
mesheryctl installer script fails when the system does not have unzip command installed.

#### Expected Behavior
Possible solutions:
1. move away from zip file format packaging - either just download mesheryctl command or move to tar package format. [Requires changes to CI/CD pipelines]
2. throw an user friendly error log to "install unzip command and rerun the installer script"

#### Environment

- Host OS: Ubuntu

---

**Contributor Resources:**
- [Contributing](https://docs.meshery.io/project/contributing)

hexxdump:

**Description**
instead of packaging mesheryctl, README.md and LICENSE, package the binary alone.

This PR fixes #3756

**Notes for Reviewers**
Need help with testing the changes. Will my changes be effective on triggering a release?

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -53,7 +53,9 @@ archives:
     windows: Windows
     386: i386
     amd64: x86_64
-  format: zip
+  format: binary
+  files:
+    - none*
 #   format_overrides:
 #     - goos: windows
 #       format: zip


hexxdump:
> @hexxdump We would also need to change the bash script right?

yes @navendu-pottekkat, once we confirm that the mesheryctl binary is pushed properly on release page, we need to change the installer script to get rid of unzip related code.

navendu-pottekkat:
@hexxdump Okay sure, we should be able to test this on a fork right? Otherwise making a release would break mesheryctl installation.

DelusionalOptimist:
> Need help with testing the changes. Will my changes be effective on triggering a release?

Yes @hexxdump, you'll be able to test this if you do a tagged release.
You can remove https://github.com/meshery/meshery/blob/0f78883c6ffa65a3ac8879e53be82d64b1d23235/.github/workflows/build-and-release.yml#L42
make appropriate secrets https://github.com/meshery/meshery/blob/0f78883c6ffa65a3ac8879e53be82d64b1d23235/.github/workflows/build-and-release.yml#L56
and then trigger this by creating a release from the releases page (or simply pushing a new git tag).

leecalcote:
@hexxdump that's right, yes. These changes will not going to affect upon merge. You're correct in thinking that MesheryCTL is The only component that does￼ not get released in edge releases￼.

leecalcote:
> @hexxdump We would also need to change the bash script right?

Yes, indeed.

DelusionalOptimist:
@hexxdump @leecalcote @navendu-pottekkat If we have a purpose of having releases shipped in archives (we might ship additional artifacts in future?) should we move to a tar based release for *nix based platforms instead and continue with the zip based release for windows?
I believe every linux distro as well as MacOS comes with `tar` pre-installed :thinking:.
This is however just a suggestion in case we have something in mind for releasing stuff other than `mesheryctl`

leecalcote:
Good suggestion, @DelusionalOptimist. As we ponder air-gapped deployments, bundling of Meshery deployments in the mesheryctl package might be a very sensible approach. Considering that we would like to support air-gapped deployments now... @hexxdump, perhaps, I have misled us and we should shift to Rudraksh's tar approach, and bundle the respective release's manifests (ideally, helm charts).

navendu-pottekkat:
For air-gapped deployments, we can always have the user download the manifests manually and use these downloaded manifests in an air-gapped deployment. I do not see the necessity of having them bundled with mesheryctl. If we have the possibility of shipping other artefacts, then using tar makes sense.

leecalcote:
@navendu-pottekkat, you're suggesting that we should have two installation steps instead of one? Bundling helps avoid GH rate limiting and helps seed installation files so that if someone isn't connected to the Internet later (after they install with brew and have no install files on their system), then they aren't stuck not being able to deploy and having to take an extra, manual step. Is there are negative consequence of bundling the installation files that you're trying to avoid?

navendu-pottekkat:
@leecalcote My suggestion was still 1 step but would involve users having to connect to the Internet do download the manifests but they could also download the manifests to their machines for air-gapped environments.

But bundling of the manifests would give out-of-the-box installation without having the user download anything else. So, it would be better.

DelusionalOptimist:
Yes @hexxdump I think this is what we're doing :white_check_mark:.
Also, you tagged the wrong @leecalcote :sweat_smile:

navendu-pottekkat:
@hexxdump @DelusionalOptimist Final verdicts here?

DelusionalOptimist:
@hexxdump I think we would should remove `zip` from here:
https://github.com/meshery/meshery/blob/986937b1a897ca4d7af2159ba22a3e14ff926e89/.goreleaser.yml#L56
It by-default packages it in a `tar.gz` if nothing is specified according to https://goreleaser.com/customization/archive/.


After that we'll need to uncomment this:
https://github.com/meshery/meshery/blob/986937b1a897ca4d7af2159ba22a3e14ff926e89/.goreleaser.yml#L57-L59
It would ensure that when packing the binary, it uses a `zip` when platform is windows.

PR #3845, opened 2021-08-11
Add more eslint rules to enforce strict formatting 

### LINKED ISSUE #3701: [UI][lint] Add more rules to eslint config
#### Current Behavior
At present, we are using eslint for linting and formatting. And since we don't have enough rules specified in the `.eslintrc.js` file, code formatting is not consistent. It becomes very evident when a contributor is using a different formatter and it formats the code according to its rules and our eslint doesn't do anything about it because we haven't asked it to.

#### Desired Behavior
More rules should be added to the eslint config to main consistency over the codebase. So, this way, even when a contributor uses a different formatter, eslint should jump in and enforce the set of rules we provided.

---

humblenginr:
This PR fixes #3701 

**Note for reviewers:**
All the files are reformatted using the newly added eslint rules, that is why, this is a 5000 lines change, but those are just formatting ✌🏽 


leecalcote:
A pre-commit hook will keep everyone inline, yeah?

An ignorant question￼: how do we disseminate and propagate the settings to contributor's IDEs?￼

sudo-NithishKarthik:
> how do we disseminate and propagate the settings to contributor's IDEs?

I am not sure I understand what you mean 🤔 
As soon as the user runs `npm install` , it will install the pre-commit hook which will automatically lint the files before commit.
For the IDE, if the user has `eslint` plugin installed for vscode, it will automatically detect the rules and do it's job

leecalcote:
Sounds good. Thank you.

leecalcote (approved):

PR #3595, opened 2021-07-22
[UI] block fetch to grafana or prometheus if not configured

bariqhibat:
Signed-off-by: Bariq <bariqhibat@gmail.com>

**Description**

This PR fixes #3566

**Notes for Reviewers**

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #3847, opened 2021-08-11
[UI] block fetch to grafana or prometheus if not configured

### LINKED ISSUE #3566: [UI] Fetch grafana and prometheus scan data only if they are configured
#### Description
Right now, Meshery UI client attempts to fetch grafana and prometheus data even when the user has not configured them.

For instance, 

Screenshot 2021-07-19 at 8 23 47 PM

^^ In this case, the user has not configured granafa and prometheus, but still, on every load, it attempts to connect to them, which may not be compelling for the user if he consciously retracted himself from configuring them.

Screenshot 2021-07-19 at 8 27 22 PM

^^ When the user navigates to the **Settings** page, meshery UI client automatically tries to fetch grafana and prometheus scan data even when they are not configured.

#### Expected Behavior

It would be better if meshery UI client only tries to connect to grafana and prometheus only when the user has configured it and not show the errors that are not reasonable. 

Also, it should only try to fetch the grafana and prometheus scan data only if the user has configured them.

---

**Contributor Resources:**

bariqhibat:

**Description**

This PR fixes #3566 from this [PR](https://github.com/meshery/meshery/pull/3595#event-5077626917)

**Notes for Reviewers**

### DIFF
--- a/ui/components/DashboardComponent.js
+++ b/ui/components/DashboardComponent.js
@@ -178,6 +178,8 @@ class DashboardComponent extends React.Component {
       meshScan: [],
       activeMeshScanNamespace: {},
       meshScanNamespaces: {},
+
+      isMetricsConfigured: grafana.grafanaURL !== '' && prometheus.prometheusURL !== '' && k8sconfig.clusterConfigured
     };
   }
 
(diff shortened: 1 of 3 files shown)


sudo-NithishKarthik (approved):
Thank you @bariqhibat 🎉

leecalcote:
The ol' block and tackle. Nice.

PR #3869, opened 2021-08-13
Support minikube out of the box and multi k8s context support

### LINKED ISSUE #3759: [server] Mount Minikube Config in Meshery Container
#### Current Behavior
Minikube users often stumble upon the issue that minikube config is not flattened and minified. 

#### Desired Behavior
`~/.minikube` path must be mounted along with the `~/.kube` directory so that certificate data can be resolved inside meshery server as well.

This partly solves the problem but meshery server also need to be able to flatten and minify the config so that flattened configuration can be sent to the adapters.

#### Implementation
1. Mount `~/.minikube` directory in the `docker-compose.yaml`
2. Write functions for flattening config in meshery server
3. Overwrite the internal config with the flattened config to support reuse by adapters

#### Acceptance Tests
1. Minikube users should be able to reach kubernetes without manually uploading configuration
2. Adapters should be reachable

---

tangledbytes:
**Description**

This PR fixes #3759 
This PR adds the following:
1. Auto flattening of file references in the kubeconfig. `~/.minikube` directory is mounted into the container now.
2. Multiple k8s contexts are supported now. 2 new API endpoints are introduced:
    1. `PUT /api/system/kubernetes/contexts` - Allows to switch context without interfering with file system kubeconfig.
    2. `GET /api/system/kubernetes/contexts` - Returns a list of contexts

**Notes for Reviewers**

### DIFF
--- a/handlers/k8sconfig_handler.go
+++ b/handlers/k8sconfig_handler.go
@@ -179,21 +179,6 @@ func (h *Handler) GetContextsFromK8SConfig(w http.ResponseWriter, req *http.Requ
 	}
 }
 
-// swagger:route POST /api/system/kubernetes/contexts SystemAPI idGetKubernetesContexts
-// Handle GET requests for kubernetes contexts
-//
-// Get list of kubernetes contexts
-// responses:
-// 	200:
-
-func (h *Handler) GetContexts(w http.ResponseWriter, req *http.Request, prefObj *models.Preference, user *models.User, provider models.Provider) {
-	if err := json.NewEncoder(w).Encode(prefObj.K8SConfig.Contexts); err != nil {
-		err = errors.Wrap(err, "unable to marshal the payload")
-		h.log.Error(ErrMarshal(err, "contexts"))
-		http.Error(w, ErrMarshal(err, "contexts").Error(), http.StatusInternalServerError)
-	}
-}
-
 func (h *Handler) loadInClusterK8SConfig() (*models.K8SConfig, error) {
 	return h.setupK8sConfig(true, nil, "")
 }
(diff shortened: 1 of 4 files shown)


handlers/k8sconfig_handler.go
```diff
@@ -295,3 +328,65 @@ func (h *Handler) setupK8sConfig(inClusterConfig bool, k8sConfigBytes []byte, co
...
+func (h *Handler) ChangeK8sContext(w http.ResponseWriter, req *http.Request, prefObj *models.Preference, user *models.User, provider models.Provider) {
+	contextReqObj := map[string]interface{}{}
+
+	defer func() {
+		_ = req.Body.Close()
+	}()
+
+	if err := json.NewDecoder(req.Body).Decode(&contextReqObj); err != nil {
+		err = errors.Wrap(err, "unable to marshal the payload")
+		logrus.Error(ErrMarshal(err, "context-name"))
```
sayantan1413, line 349: Hey @utkarsh-pro I was thinking that we can be consistent with the `h.log.Error`.  Though I don't know the requirement here but I think if the usage is consistent then it would have been better I guess😃.

utkarsh-pro: Yep, you are right. Missed this out.

utkarsh-pro:
@piyushsingariya any idea why the swagger workflow is failing? // @DelusionalOptimist

DelusionalOptimist:
hmm...
I might have missed it. @piyushsingariya @sayantan1413 enlighten me,  why do we need to run this in PRs? :thinking:

sayantan1413:
Hey @DelusionalOptimist we need to run this in the PR because if some error is there while writing the swagger then this workflow will fail giving what is the error about. And if everything is okay the it will update the swagger.yml after the PR is merged.

DelusionalOptimist:
I see. So this here is failing because the token which we're using in this is not available to forks. We'll need to split the part for performing checks in PRs to https://github.com/meshery/meshery/blob/master/.github/workflows/ci.yml. 
@utkarsh-pro for now, if you're able to successfully run the swagger tests offline (using one of the make targets) then this is good to go.
I'll help @sayantan1413 address the changes to the workflow in a new PR. :+1:

router/server.go
```diff
@@ -59,6 +59,10 @@ func NewRouter(ctx context.Context, h models.HandlerInterface, port int, g http.
 		Methods("POST", "DELETE")
 	gMux.Handle("/api/system/kubernetes/contexts", h.ProviderMiddleware(h.AuthMiddleware(http.HandlerFunc(h.GetContextsFromK8SConfig)))).
 		Methods("POST")
+	gMux.Handle("/api/system/kubernetes/contexts", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.ChangeK8sContext)))).
+		Methods("PUT")
```
leecalcote, line 63: @utkarsh-pro how do we reconcile this against API design principle #1? https://docs.google.com/spreadsheets/d/1ABJCbfQRi0uN_YoP2kmHZ-lTI4S4QuvRk1_unomKNRE/edit#gid=1241442399

models/preference.go
```diff
@@ -9,12 +9,13 @@ import (
...
-	K8Sfile           string `json:"k8sfile,omitempty"`
-	Config            []byte `json:"config,omitempty"`
-	Server            string `json:"configuredServer,omitempty"`
-	ContextName       string `json:"contextName,omitempty"`
-	ClusterConfigured bool   `json:"clusterConfigured,omitempty"`
+	InClusterConfig   bool         `json:"inClusterConfig,omitempty"`
+	K8Sfile           string       `json:"k8sfile,omitempty"`
+	Config            []byte       `json:"config,omitempty"`
+	Server            string       `json:"configuredServer,omitempty"`
+	ContextName       string       `json:"contextName,omitempty"`
```
leecalcote, line 16: Is CurrentContext a more accurate descriptor?

utkarsh-pro: Yes it is. I would have called it the same but as this was already there so I kept the changes to the minimum. Should I change the name now?

router/server.go
utkarsh-pro: We already had a `POST` method defined for the endpoint and it is also being used at some places. I had two options
1. Change endpoint: Didn't do so because that would mean that endpoint should be pointing to a new "resource", which it does not.
2. Add another method: Had to resort to this because the couldn't think of something better,

Any suggestions here would be helpful :sweat_smile:

docker-compose.yaml
```diff
@@ -11,6 +11,7 @@ services:
       - "EVENT=mesheryLocal"
     volumes:
       - $HOME/.kube:/home/appuser/.kube:ro
+      - $HOME/.minikube:$HOME/.minikube:ro
```
leecalcote, line 14: Looking good.

handlers/error.go
```diff
@@ -312,3 +314,11 @@ func ErrSaveSession(err error) error {
 func ErrCreateDir(err error, obj string) error {
 	return errors.New(ErrCreateDirCode, errors.Alert, []string{"Error creating directory ", obj}, []string{err.Error()}, []string{}, []string{})
 }
+
+func ErrInvalidRequestObject(fields ...string) error {
+	return errors.New(ErrCreateDirCode, errors.Alert, append([]string{"Error invalid request object:"}, fields...), []string{}, []string{}, []string{})
+}
+
+func ErrChangeK8sContext(err error) error {
+	return errors.New(ErrCreateDirCode, errors.Alert, []string{"Error changing context"}, []string{err.Error()}, []string{}, []string{})
```
leecalcote, line 323: Probable cause? Suggested Remediation?

router/server.go
leecalcote: We might be looking at a /id

handlers/error.go
utkarsh-pro: Added

router/server.go
utkarsh-pro: `POST` to `/api/system/kubernetes/contexts/{name}`. Is this the suggestion?

Example:
`POST` to `/api/system/kubernetes/contexts/minikube`

PR #3928, opened 2021-08-18
[UI] Fix  'Error communicating with Prometheus' error 

humblenginr:
**Description**

A function definition is supposed to be passed to the `datafetch` method as an argument to handle error and success, but a function call was being passed, so, every time the compiler hit that line, it executed that function, which resulted in error notification.
Now that the function call is changed to a definition, it will behave as expected

This PR fixes the issue of `Error communicating with Prometheus` even when Prometheus is configured properly.

**Notes for Reviewers**

### DIFF
--- a/ui/components/PrometheusComponent.js
+++ b/ui/components/PrometheusComponent.js
@@ -130,7 +130,7 @@ class PrometheusComponent extends Component {
             });
           }
         },
-        self.handleError("There was an error communicating with Grafana")
+        self.handleError
       )
   }
 


leecalcote:
Great write up and explanation, @sudo-NithishKarthik

leecalcote (approved):

leecalcote:
@samirparhi-dev FYI - this fix along with a few other bug fixes will be released in the stable release channel later today.

PR #3943, opened 2021-08-19
Fix user pref tabs UI

### LINKED ISSUE #3936: [UI] User Prefs: Only display remote provider tab when user signs into a remote provider.
#### Current Behavior

This screenshot is of a user using the Local Provider (None). They should not see the Remote Provider tab.
Screen Shot 2021-08-18 at 7 30 22 PM

The labels on the tabs should be responsive and should not be cut off.

Screen Shot 2021-08-18 at 7 34 53 PM

#### Desired Behavior
In `/user/preferences`, the Remote Provider tab should only be shown when the user is connected to a Remote Provider.

#### Environment

- Browser: Safari
- Host OS: Mac
- Meshery Server Version: stable-v0.5.48

---

ghost:
**Description**
- Show remote provider only when remote provider is used otherwise hide it.
- Fix tab and their labels to be responsive.

This PR fixes #3936

### DIFF
--- a/ui/components/UserPreferences.js
+++ b/ui/components/UserPreferences.js
@@ -35,7 +35,7 @@ const styles = (theme) => ({
   },
   paperRoot : {
     flexGrow : 1,
-    maxWidth : "90%",
+    maxWidth : "100%",
     marginLeft : 0,
     borderTopLeftRadius : 3,
     borderTopRightRadius : 3,
@@ -231,6 +231,16 @@ class UserPreference extends React.Component {
                 label={<span className={classes.tabLabel}>General</span>}
               />
             </Tooltip>
+            <Tooltip title="Choose Performance Test Defaults" placement="top">
+              <Tab
+                className={classes.tab}
+                icon={
+                  <FontAwesomeIcon icon={faTachometerAlt} transform={mainIconScale} fixedWidth />
+                }
+                label={<span className={classes.tabLabel}>Performance</span>}
+              />
+            </Tooltip>
+            {/* NOTE: This tab's appearance is logical hence it must be put at last here! Otherwise added logic will need to be added for tab numbers!*/}
             {userPrefs && providerType != 'local' &&
               <Tooltip title="Remote Provider preferences" placement="top">
                 <Tab
@@ -242,15 +252,6 @@ class UserPreference extends React.Component {
                 />
               </Tooltip>
             }
-            <Tooltip title="Choose Performance Test Defaults" placement="top">
-              <Tab
-                className={classes.tab}
-                icon={
-                  <FontAwesomeIcon icon={faTachometerAlt} transform={mainIconScale} fixedWidth />
-                }
-                label={<span className={classes.tabLabel}>Performance</span>}
-              />
-            </Tooltip>
           </Tabs>
         </Paper>
         <Paper className={classes.root}>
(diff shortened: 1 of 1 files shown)


sudo-NithishKarthik:
https://user-images.githubusercontent.com/75248557/130205945-a79b0e46-b416-4a0f-8bb3-7eb43dd57a2d.mov

I still see the `Remote Provider` tab 🤔 , @yagarwal1307

sudo-NithishKarthik (commented):
When using the `None` provider, `Performance` tab does not have any content in it.

Screenshot 2021-08-20 at 2 26 51 PM

sudo-NithishKarthik:
@leecalcote , should we actually show the `performance` tab in case of local provider, considering the fact that we would be persisting it in the remote provider in the future ?

warunicorn19 (changes requested):
Screenshot 2021-08-20 at 2 58 04 PM

While @sudo-NithishKarthik points out the important stuff, i'll nitpick a little, is there a reason why theres this line? if its insignificant please remove it.😃

yagarwal1307:
@sudo-NithishKarthik Ahhhhh, stupid mistake! Nice catch btw! Fixing it just 2 minutes.

@warunicorn19 I don't know about the line, this was already there, must be the border. Looked cool to me so I just left it there :P. For now I am setting the width of enclosing container 100% that would fix it!

warunicorn19:
For some reason there is a little lag when remote provider is clicked, the lag was longer when i opened it the first time.


https://user-images.githubusercontent.com/53532328/130230999-982accd3-f5a6-431c-9838-7c8535419301.mov

yagarwal1307:
@warunicorn19 I don't know in my case remote provider tab is empty! Also I am pretty sure it is not because of the changes I made or is it? (If you think it is please tell me the potential reason, I dont think its because of me)

But I think I can guess the reason!

According to me it is happening because of the nature of how react works.
If you go check ExtensionSandbox component you will find that everytime this tab is rendered useEffect is called and useEffect is called once the code has already been rendered. So what might be happening is that before data is fetched since you have nothing in data nothing is shown. But once useEffect is called, you get data and it gets re rendered. Hence the case!

warunicorn19 (approved):
No no, even i dont think its because of the changes you've done😅. i was just wondering. i guess what you said makes sense and that's probably the cause. Thanks!

All looks good to me🎉, thanks for this @yagarwal1307 .
@sudo-NithishKarthik would you like to add anything else?

yagarwal1307:
@warunicorn19 the issue was the same as I was referring to. In the extensionSandbox.js when no data is available we are rendering null. I replaced that null with 'hello' and you can see the result. Please check the video below.
https://user-images.githubusercontent.com/49173018/130235508-7c46f114-51f7-4113-904d-3c09d64c9f15.mp4

Also I have observed this same thing in a lot of places in the code base and we should use some placeholders in all those places. It will improve the ui significantly.

yagarwal1307:
Also one more thing making the request for same data again and again does not make any sense in the first place in this case. For this I think we should rather use redux or something and just fetch the data once!

warunicorn19:
@yagarwal1307 good catch. Something to think about. If you'd like you can add it to the agenda of our community meeting and we can discuss this at length there.

yagarwal1307:
sure @warunicorn19 I will do that.

sudo-NithishKarthik:
@yagarwal1307 , Thank you for fixing it.

> For some reason there is a little lag when remote provider is clicked, the lag was longer when i opened it the first time.

Yes, @yagarwal1307 is right, the reason for the delay in the content to appear is because of the fact that `null` is being rendered until the data is received from the network request which fetches provider capabilities and also partly because of the fact that it is an extension point of Meshery, which loads the component each time you go to the tab. We should definitely use a progress bar or something instead. Can any of you guys open an issue for it ? @yagarwal1307  or @warunicorn19 .

> Also one more thing making the request for same data again and again does not make any sense in the first place in this case. For this I think we should rather use redux or something and just fetch the data once!

There is an open proposal for this which addresses this issue among the others, we will be switching to Redux based architecture real soon 🚀 .

warunicorn19:
@yagarwal1307 please go ahead and create the issue😃.

yagarwal1307:
cool! Doing that

sudo-NithishKarthik (approved):

sudo-NithishKarthik:
Thank you @yagarwal1307 , for you valuable contributions 🎉

PR #3960, opened 2021-08-23
[Performance] Performance REST API migration to GraphQL

bariqhibat:
**Description**

**Notes for Reviewers**
GraphQL performance subscription. When a new result comes, it'll publish the result to the channels, then the listeners would receive the data.

To reproduce:
1. Run the below query on GraphQL playground
Query
```graph
subscription{
  subscribePerfProfile(profileID: "44693737-3c17-485e-bdd2-c2f9be9c5dda"){
    meshery_id
    name
    mesh
  }
}
```
2. Go to [Performance profiles](http://localhost:9081/performance/profiles)
3. Run performance test for this specific profile (since the corresponding ID on the query above is for this profile)
image
4. Once test finishes, you'll see a new performance result on the GraphQL playground
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,6 @@ require (
 	fortio.org/fortio v1.17.0
 	git.mills.io/prologic/bitcask v0.3.14
 	github.com/99designs/gqlgen v0.13.0
-	github.com/agnivade/levenshtein v1.1.1 // indirect
 	github.com/asaskevich/govalidator v0.0.0-20200907205600-7a23bdc65eef
 	github.com/briandowns/spinner v1.16.0
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
@@ -58,7 +57,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.8.1
 	github.com/tcnksm/go-latest v0.0.0-20170313132115-e3007ae9052e
-	github.com/vektah/gqlparser/v2 v2.1.0
+	github.com/vektah/gqlparser/v2 v2.2.0
 	github.com/vmihailenco/taskq/v3 v3.0.0-beta.9.0.20200519124923-d9823546b85a
 	golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a // indirect
 	golang.org/x/oauth2 v0.0.0-20210514164344-f6687ab2804c
--- a/go.sum
+++ b/go.sum
@@ -117,8 +117,6 @@ github.com/afex/hystrix-go v0.0.0-20180502004556-fa1af6a1f4f5/go.mod h1:SkGFH1ia
 github.com/agnivade/levenshtein v1.0.1/go.mod h1:CURSv5d9Uaml+FovSIICkLbAUZ9S4RqaHDIsdSBg7lM=
 github.com/agnivade/levenshtein v1.0.3 h1:M5ZnqLOoZR8ygVq0FfkXsNOKzMCk0xRiow0R5+5VkQ0=
 github.com/agnivade/levenshtein v1.0.3/go.mod h1:4SFRZbbXWLF4MU1T9Qg0pGgH3Pjs+t6ie5efyrwRJXs=
-github.com/agnivade/levenshtein v1.1.1 h1:QY8M92nrzkmr798gCo3kmMyqXFzdQVpxLlGPRBij0P8=
-github.com/agnivade/levenshtein v1.1.1/go.mod h1:veldBMzWxcCG2ZvUTKD2kJNRdCk5hVbJomOvKkmgYbo=
 github.com/ajstarks/svgo v0.0.0-20180226025133-644b8db467af/go.mod h1:K08gAheRH3/J6wwsYMMT4xOr94bZjxIelGM0+d/wbFw=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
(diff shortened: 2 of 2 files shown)


leecalcote:
👀

ui/components/MesheryPerformance/Dashboard.js
```diff
@@ -43,18 +44,28 @@ function Dashboard({
...
+        page : `0`,
+        search : ``,
+        order : ``,
+      },
+    }).subscribe({
+      next : (res) => {
+        // @ts-ignore
+        let result = res?.getPerformanceProfiles;
+        if (typeof result !== "undefined") {
+          updateProgress({ showProgress : false });
```
sudo-NithishKarthik, line 60: This can be moved outside of the if block ig, because, if the `result` is undefined, then the progress bar will not stop, which can be frustrating at times 😅

ui/components/MesheryPerformance/PerformanceProfiles.js
```diff
@@ -86,26 +87,30 @@ function PerformanceProfile({ updateProgress, enqueueSnackbar, closeSnackbar })
...
+        page : `${page}`,
+        search : `${encodeURIComponent(search)}`,
+        order : `${encodeURIComponent(sortOrder)}`,
+      },
+    }).subscribe({
+      next : (res) => {
+        // @ts-ignore
+        let result = res?.getPerformanceProfiles;
+        if (typeof result !== "undefined") {
+          updateProgress({ showProgress : false });
```
sudo-NithishKarthik, line 103: same here

bariqhibat:
Hey @sudo-NithishKarthik, thanks for the feedback. 

There's one more step to this actually, it is to update the cache store in our Relay. I've looked into it for a while but it's not really as intuitive as I thought, could you hop in to help me out with this?

The idea is to update by appending the newly-came data from subscription to the original query.

sudo-NithishKarthik:
Hey @bariqhibat , 

To answer your question, Seems like Relay [handles that automatically](https://relay.dev/docs/guided-tour/updating-data/graphql-subscriptions/), we just have to make sure we update the data in the store, for which we can use the methods that [they provide](https://relay.dev/docs/api-reference/store/#recordsourceselectorproxy) to work with Relay store, but, yes, I don't find it very intuitive as well 😕 .

And it makes sense because Relay basically wants to have full control over the store.

But, with our current implementation, we are not actually making use of relay's cache, yes, we are using `fetchQuery` which internally stores the data in Relay's store, but we are updating the React local state whenever we get a payload from subscriptions, this way, we do not have to fight with relay for control over the store 😝 .

And, this is only till we setup Redux, after that, we will only have to update the global store.

internal/graphql/resolver/addons.go
```diff
@@ -21,7 +22,7 @@ func (r *Resolver) getAvailableAddons(ctx context.Context, provider models.Provi
 		selectors = append(selectors, *selector)
 	}
 
-	addonlist, err := model.GetAddonsState(selectors, r.Config.KubeClient, provider)
+	addonlist, err := services.GetAddonsState(selectors, r.Config.KubeClient, provider)
```
dhruv0000, line 25: Why chance from `model` to `service` package name? We have been using `model` convention in Meshery.

bariqhibat: Since I need to decouple "model"-definition generated by gqlgen and the "models"-function like

dhruv0000: Why the need to decouple? The resolvers are both generated and hand written, and we still use the same package for it.

bariqhibat:
Putting this to draft since need to wait for state management for the subscription update

internal/graphql/resolver/addons.go
bariqhibat: For me, it doesn't make sense to put the type definitions for the resolvers to be in the same package as functions for talking to the database. The other thing is that it's difficult to connect the packages due to cyclic imports.

bariqhibat:
Migrated all REST API to GraphQL resolvers on `/performance`

image

bariqhibat:
Updated the sheet on 
https://docs.google.com/spreadsheets/d/1ABJCbfQRi0uN_YoP2kmHZ-lTI4S4QuvRk1_unomKNRE/edit#gid=1241442399

Added new Get Resolver for fetchAllResults, getPerformanceProfiles

leecalcote:
> Updated the sheet on https://docs.google.com/spreadsheets/d/1ABJCbfQRi0uN_YoP2kmHZ-lTI4S4QuvRk1_unomKNRE/edit#gid=1241442399
> 
> Added new Get Resolver for `fetchAllResults`, `getPerformanceProfiles`

Thank you, @bariqhibat. Do you consider this PR ready for merge?

handlers/load_test_handler.go
```diff
@@ -456,6 +458,27 @@ func (h *Handler) executeLoadTest(ctx context.Context, req *http.Request, testNa
...
+
+	// h.config.PerformanceChannels[performanceProfileID] <- &graphqlModel.MesheryResult{
+	// 	MesheryID:          &resultID,
+	// 	Name:               &result.Name,
+	// 	Mesh:               &result.Mesh,
+	// 	PerformanceProfile: &performanceProfile,
+	// 	TestID:             &result.TestID,
+	// 	RunnerResults:      result.Result,
+	// 	ServerMetrics:      &serverMetrics,
+	// 	ServerBoardConfig:  &serverBoardConfig,
```
leecalcote, line 476: @dhruv0000 ServerBoard synonymous with StaticBoard? // @nehaLohia27

bariqhibat:
> Thank you, @bariqhibat. Do you consider this PR ready for merge?

Yes, I've tested this manually and it's working fine.

leecalcote:
Thank you, @bariqhibat. Moving forward...

PR #4085, opened 2021-09-04
Fix Pattern syncup with rjsf form

Abhishek-kumar09:
**Description**
1. Patterns was not in sync because earlier the patterns was searched on the basis of workload key. The workload key can have any different value according to user will that in turns breaks the sync between rjsf and actual pattern yaml.

Now the patterns are matched according to type that fixes the problem

2. The pattern form was sharing code between `mesherymeshInterface` so if developers make changes in that portion of code then some unwanted changes were reflected on pattern's page. Extracted out the code that is used by only pattern forms.

**Notes for Reviewers**

### DIFF
--- a/ui/components/MesheryPatternServiceForm.js
+++ b/ui/components/MesheryPatternServiceForm.js
@@ -87,7 +87,7 @@ function createPatternFromConfig(config, namespace, partialClean = false) {
 
   Object.keys(pattern.services).forEach((key) => {
     // Delete the settings attribute/field if it is set to "true"
-    if (pattern.services[key].settings === true) delete pattern.services[key].settings;
+    if (pattern.services[key].settings) delete pattern.services[key].settings;
   });
 
   return pattern;
--- a/ui/components/MesheryPatterns.js
+++ b/ui/components/MesheryPatterns.js
@@ -785,9 +785,10 @@ function PatternForm({ pattern, onSubmit, show }) {
   function insertPattern(workload) {
     const attrName = getPatternAttributeName(workload);
     var returnValue = {}
-    Object.keys(deployServiceConfig).forEach(key => {
+    Object.keys(deployServiceConfig).find(key => {
       if (deployServiceConfig[key]['type'] === attrName) {
         returnValue = deployServiceConfig[key]
+        return true
       }
     })
 


ui/components/MesheryPatterns.js
```diff
@@ -779,6 +782,18 @@ function PatternForm({ pattern, onSubmit, show }) {
     setYaml(data.valueOf().getValue())
   }
 
+  function insertPattern(workload) {
+    const attrName = getPatternAttributeName(workload);
+    var returnValue = {}
+    Object.keys(deployServiceConfig).forEach(key => {
+      if (deployServiceConfig[key]['type'] === attrName) {
+        returnValue = deployServiceConfig[key]
```
bariqhibat, line 790: arent you trying to find here? can we use the `Object.keys(object).find` instead?

ui/components/MesheryPatterns.js
```diff
@@ -720,7 +723,7 @@ function PatternForm({ pattern, onSubmit, show }) {
   }
 
   function getPatternAttributeName(jsonSchema) {
-    return PascalCaseToKebab(jsonSchema?._internal?.patternAttributeName || "NA");
+    return jsonSchema?._internal?.patternAttributeName || "NA";
```
bariqhibat, line 726: just curious, is there a reason why we're not using the PascalToKebab function?

ui/components/MesheryPatternServiceForm.js
```diff
@@ -0,0 +1,206 @@
...
+}
+
+function a11yProps(index) {
+  return {
+    id : `simple-tab-${index}`,
+    "aria-controls" : `simple-tabpanel-${index}`,
+  };
+}
+
+function getPatternAttributeName(jsonSchema) {
```
bariqhibat, line 36: I saw this function on the other file, can we make it reusable instead of this duplicate definitions?

ui/components/MesheryPatternServiceForm.js
```diff
@@ -0,0 +1,206 @@
...
+      pattern.services[name] = config[key];
+
+      pattern.services[name].type = key;
+      pattern.services[name].namespace = namespace;
+    }
+  });
+
+  Object.keys(pattern.services).forEach((key) => {
+    // Delete the settings attribute/field if it is set to "true"
+    if (pattern.services[key].settings === true) delete pattern.services[key].settings;
```
bariqhibat, line 90: nitpick: remove the === true

bariqhibat: maybe just to `pattern.services[key].settings` or `!!pattern.services[key].settings`

bariqhibat (commented):
Just made some comments, thanks for the work Abhi!

ui/components/MesheryPatternServiceForm.js
leecalcote: @bariqhibat that is a great suggestion. Yes, we need to avoid repetition between patterns, application, and filters. Will you `gh pr checkout 4085` and assist? // @Chadha93 @sudo-NithishKarthik

ui/components/MesheryPatterns.js
Abhishek-kumar09: Yeah, sounds good.

ui/components/MesheryPatterns.js
Abhishek-kumar09: Actually, pascalToKebab is to search for the keys as a name of service. But that was not serving the purpose since it the name of service may be chosen to be different in most of the ideal scenarios. so now the patterns are matched according to their type, so now type doesns't need this type of transformation.

ui/components/MesheryPatternServiceForm.js
Abhishek-kumar09: Actually, it is intended.
 80% of the file content inside `MesheryPatternServiceForm.js` is same as `MesheryMeshInterface/PatternServiceForm.js`.
 But this is done on purpose, since earlier I was actually sharing all the code with the `MesheryMeshInterface`, so both of the components are pre-mature and they are under rapid development. So whenever there is a change done thinking to make` MeshInterface` better, there is a side effect on the `Patterns page`. So earlier we had seen such conflicts, like the rjsf form styling issues inside patterns, the accordion expansion , and the syncup of contents. 
 
 That's why just tried to extract out the functionalities we can, and once both components become mature we can try and merge the two codes.
 What do you say? @bariqhibat @leecalcote

ui/components/MesheryPatterns.js
Abhishek-kumar09: Done :smile:

ui/components/MesheryPatternServiceForm.js
Abhishek-kumar09: Done.

ui/components/MesheryPatterns.js
```diff
@@ -640,10 +639,14 @@ const mapStateToProps = (state) => {
...
 
+
+
 // --------------------------------------------------------------------------------------------------
 // -------------------------------------------- PATTERNS FORM ---------------------------------------
 // --------------------------------------------------------------------------------------------------
 
+
+
 function PatternForm({ pattern, onSubmit, show }) {
```
Chadha93, line 650: @Abhishek-kumar09 @bariqhibat @leecalcote We are using this function `in` app`s as well but the main issue that arises is passing the values we need to assign values in every function accordingly to the file, for eg here we are using `patterns, pattern_file` so I think some of the function need to be replicated in both, please do correct my opinion if required.

ui/components/MesheryPatternServiceForm.js
bariqhibat: I see, if both configurations are independent to each other, then it makes sense to separate them. And eventually both need to be merged. Could you see into it once they're at that point? @Abhishek-kumar09

Abhishek-kumar09: yeah, for sure :smile:

ui/components/MesheryPatterns.js
Abhishek-kumar09: Yeah, in the current point of time we can change few things in the the components that will made this work as a single reusable component for both of the the pages. But if Application and pattern and filters has too much architecture difference, then need to separate out the things.

ui/components/MesheryPatternServiceForm.js
leecalcote: Thank you, gents. Let's proceed according to your plan.

leecalcote (approved):

PR #4153, opened 2021-09-10
Fix prometheus and grafana discovery

tangledbytes:

**Description**

This PR fixes prometheus and grafana discovery in meshery UI.

Screenshot_20210911_011934

**Notes for Reviewers**

### DIFF
--- a/ui/components/MesherySettings.js
+++ b/ui/components/MesherySettings.js
@@ -108,7 +108,7 @@ class MesherySettings extends React.Component {
       tabVal,
       subTabVal,
 
-      isMeshConfigured : k8sconfig.clusterConfigured && grafana.grafanaURL !== '' && prometheus.prometheusURL !== '',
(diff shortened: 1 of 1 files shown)


leecalcote (approved):
It seems unintuitive that we wouldn't check for an empty endpoint url.

utkarsh-pro:
> It seems unintuitive that we wouldn't check for an empty endpoint url.

@leecalcote, here is my understanding: previous code was checking if prometheus URL and grafana URL has been configured or not, if it is then scan or else skip. The change just removes the check for previously configured prometheus and grafana and performs discovery if cluster is configured.

leecalcote:
Thanks for this explanation. Our next steps here involve introduction of states. The workflow engine and implementation of a finite state machine through that engine to facilitate "discovered", "registered", "active", "ignored" and so on is our likely approach.

PR #4173, opened 2021-09-12
Added support for seeded content in local provider

Revolyssup:
**Description**

This PR adds out of the box support for filters and patterns built into the meshery server's file system, for local provider.

**Notes for Reviewers**


navendu-pottekkat:
@Revolyssup Heads up! Lint errors

leecalcote:
Reminder: add labels to PRs when you can.

helpers/seed.go
```diff
@@ -0,0 +1,114 @@
+package helpers
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/gofrs/uuid"
+	models "github.com/layer5io/meshery/models"
+	"github.com/sirupsen/logrus"
```
DelusionalOptimist, line 10: If we can, let's use meshkit logger here?
https://github.com/layer5io/meshkit/blob/master/logger/logger.go

helpers/seed.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+type contentandName struct {
```
DelusionalOptimist, line 13: Is this new struct needed? Can we use the ones for patterns and filters already defined in `models` package? Helps us keep uniformity and maintainability.
You can also then assign UUIDs while reading seed content from the filesystem.

helpers/seed.go
```diff
@@ -0,0 +1,114 @@
...
+}
+type Component string
+
+const (
+	Filter  Component = "FILTER"
+	Pattern Component = "PATTERN"
+)
+
+var seededUUIDs []uuid.UUID
+var localProvider *models.DefaultLocalProvider
```
DelusionalOptimist, line 25: Any reason for not using the original reference to `DefaultLocalProvider` and creating this global variable? I think we can pass on that reference to other functions too and avoid this new var. :thinking:

DelusionalOptimist (changes requested):

provider-ui/package-lock.json
```diff
@@ -5,6 +5,7 @@
   "requires": true,
```
DelusionalOptimist, line 5: unintended changes here :smile:

helpers/seed.go
```diff
@@ -0,0 +1,114 @@
+package helpers
```
utkarsh-pro, line 1: @Revolyssup why do any of this here? If we already are being explicit that this is a functionality of none provider then add a new method to `none` provider? This approach will result in the following:
1. You don't to keep a global variable around to keep track of the provider
2. You can use the meshkit logger

helpers/seed.go
Revolyssup: this is a redundant struct, removing it anyways

helpers/seed.go
Revolyssup: it was a mistake

helpers/seed.go
Revolyssup: @utkarsh-pro  Can I leave the helper function here? which fetched the files from file system and returns them as a string.
The function which is called can then be a method of local provider.

Revolyssup: @utkarsh-pro  Then again, how do I keep track of the uuids of seeded components?

Revolyssup: Actually cant leave helper functions there due to import cycle.

Revolyssup: Can I add the SeededUUID's field in localProvider struct?

utkarsh-pro: You don't have to keep track of the generated UUIDs in the provider. Here are rough function signatures:
```go
// cmd/main.go
// ...
seededUUIDs, err := provider.SeedContent()
// ...
<- c 
// Clean up
provider.CleanupSeeded(seededUUIDs)
```

Revolyssup:
@utkarsh-pro  can you review the pr

cmd/main.go
```diff
@@ -268,5 +269,7 @@ func main() {
 		}
 	}()
 	<-c
+	logrus.Info("Doing seeded content cleanup...")
+	lProv.CleanupSeeded(seededUUIDs)
```
utkarsh-pro, line 273: Do we actually need this cleanup?

Revolyssup: The end note was to keep it for now.
@leecalcote  can you confirm?

models/default_local_provider.go
```diff
@@ -1104,3 +1157,37 @@ func genericHTTPApplicationFile(fileURL string) ([]MesheryApplication, error) {
 
 	return []MesheryApplication{af}, nil
 }
+func getSeededComponents(comp string, log logger.Handler) ([]string, []string, error) {
+	wd, err := os.UserHomeDir()
```
utkarsh-pro, line 1161: Use meshkit's HomeDir method?

Revolyssup: cool

cmd/main.go
utkarsh-pro: The advantage to have this is:
If the volume attached to meshery server is persisted across meshery restarts in those case we will end up with duplicate in the database.

Revolyssup:
@utkarsh-pro  is this ready to merge?

utkarsh-pro:
> @utkarsh-pro is this ready to merge?

Yep :+1:

utkarsh-pro (approved):
LGT

Issue #3712, opened 2021-07-31
[UX] GraphQL: Single Subscription


leecalcote:
#### Description
Meshery Operator slow to display as connected in Settings page. Meshery deployments with connected Meshery Operator, MeshSync, and Meshery Broker instances are slow to report the connected status. 

#### Expected Behavior
Meshery Operator's connected status (as connected) should be shown immediately.

Meshery UI should be using a single, pervasive subscription shared across pages.

---

**Contributor Resources:**
- [Contributing](https://docs.meshery.io/project/contributing)

Meghana-12:
I'm not sure if this works but since we are using NextJS, can we do a server side rendering first and then display?

sudo-NithishKarthik:
Our current approach makes sure that the ws connection is not busy when the data is not needed by a component(when we are in a page that does not need subscriptions i.e `filters` page)

 Whereas, if we go about initialising the subscriptions on  the startup of our webapp and keep it alive till the webapp is closed, we will get the desired behaviour.

This removes the overhead of having to subscribe to events on every mount of a component, which means we get the data needed by the component instantaneously when we navigate to that component, but this in case, ws connection will be busy until the webapp is closed even when the user is on a page that does not consume the data from sub.


So, we will have to takes this into consideration before implementing it.

@dhruv0000 suggests we stick with the same approach that we have implemented right now as he stated,

> And anyways for small subs, why make clot the ws with data that it is not using currently

// @dhruv0000  @leecalcote

leecalcote:
@MUzairS15 this has been completed, correct?

ishwar170695:
I believe this issue is obsolete. It proposes a single shared GraphQL subscription across the UI, but GraphQL subscriptions (`Subscription`/`graphql-ws`) have since been removed from Meshery in favor of SSE and REST/RTK Query. If the original delay in showing the operator's connected status still exists, it should probably be tracked as a new issue against the current architecture rather than the retired GraphQL subscription model.

PR #4189, opened 2021-09-15
[Broadcast] migrate all subscriptions to broadcaster-based subscription

bariqhibat:

**Description**

Fixes https://github.com/meshery/meshery/issues/3712

Migrate subscription for dataplane to broadcast-based subscription.

**Notes for Reviewers**
Turns out I don't need to add a type in the Meshkit.

### DIFF
--- a/internal/graphql/resolver/addons.go
+++ b/internal/graphql/resolver/addons.go
@@ -50,7 +50,7 @@ func (r *Resolver) listenToAddonState(ctx context.Context, provider models.Provi
 			case message := <-broadcastChannel:
 				// filtering based on source
 				// TODO: change to meshsync source
-				if message.Source == "meshsync" {
+				if message.Type == "meshsync" {
 					status, err := r.getAvailableAddons(ctx, provider, selector)
 					if err != nil {
 						r.Log.Error(ErrAddonSubscription(err))


leecalcote (dismissed):

internal/graphql/resolver/data-plane.go
```diff
@@ -41,18 +43,23 @@ func (r *Resolver) listenToDataPlaneState(ctx context.Context, provider models.P
 
 		for {
 			select {
-			case <-r.MeshSyncChannel:
+			case <-dataPlaneSyncChannel:
+				r.Log.Info("Dataplane sync channel called")
```
dhruv0000, line 47: Need to do a type check here, or else it will be triggered for other types too.

dhruv0000: We were using `MeshSyncChannel` for trigger, right. We should use it here too.

Also, you might need to update other places too that uses the `MeshSyncChannel` channel to use broadcast instead. (including handler ([here](https://github.com/meshery/meshery/blob/master/handlers/handler_instance.go#L29)) and Meshery-extension ([here](https://github.com/meshery/meshery/blob/de283e40e3b1477c79e64fdac7a4d1b7de8ad082/handlers/extensions.go#L44)) too).

dhruv0000 (changes requested):

internal/graphql/resolver/data-plane.go
bariqhibat: Yeah, do we want to move this meshsync channel to the broadcaster? If so, there'll be a lot of changes for this @dhruv0000

bariqhibat:
I'll migrate the rest of the subscriptions in this PR. And will make another PR for migrating MeshSyncChannel to broadcast-based system

bariqhibat:
Hey @leecalcote, I might need to do a change in the meshkit for the MeshSyncChannel that are not found in the enums. 

This might be an issue in the future as I might need to do a lot of small changes like this, should I just convert the type to be string instead of type `BroadcastSource`? 
Ref: https://github.com/layer5io/meshkit/blob/master/utils/broadcast/broadcaster.go#L26

// @dhruv0000

bariqhibat:
Well for now, I'll just use the Source of `OperatorSyncChannel` and filter the broadcast message according to Type instead.

leecalcote:
Feedback @dhruv0000?

dhruv0000:
> Well for now, I'll just use the Source of `OperatorSyncChannel` and filter the broadcast message according to Type instead.

We can also define the `BroadcastSource type` in Meshery repo itself too if that makes thing easier (package like `helper` might be a good place).

Using Type currently will just increase your work later on. in my opinion. :sweat_smile:

dhruv0000:
> I'll migrate the rest of the subscriptions in this PR. And will make another PR for migrating MeshSyncChannel to broadcast-based system

Sure. Makes sense, since it would involve migrating meshery-extensions to broadcaster too.

dhruv0000:
Great work btw @bariqhibat :fist_raised: .

bariqhibat:
Hey @utkarsh-pro, could you help reviewing this as well? If possible, this [PR](https://github.com/layer5labs/meshery-extensions/pull/355) as well, since it is dependent to this. If we're good to go, we'll merge them both

PR #4190, opened 2021-09-15
support eks authentication in out of cluster deployments

### LINKED ISSUE #4184: [meshery-server] supporting auth protocols not supported by `k8s.io/client-go` natively
#### Description
`k8s.io/client-go` and other tools (kubectl) need to be authenticated first using the cloud provider's auth methods before they can interact with the cluster.
For doing this, they generally depend on external CLIs given by the providers. For more information on this, lookup [this](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins).

When the application using the k8s client library is installed in the cluster, it doesn't need this authentication. However, out-of-cluster applications need to authenticate first.
Thus, to work with an environment in which Meshery is deployed out of cluster and connected to a managed cluster like EKS, Meshery needs to support this authentication.

For EKS, this is handled by the `aws` cli. Particularly the `aws eks get-token` command.
a Kubeconfig for EKS looks like the one below implying every time any `kubectl` command is executed, the `aws eks get-token` command is run for generating a token. The token also has an expiration time after which no further requests can be made using it.  :
```
kind: Config
preferences: {}
users:
- name: <context>
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args:
      - --region
      - <region>
      - eks
      - get-token
      - --cluster-name
      - <cluster-name>
      command: aws
``` 

When the client libraries can't find the `aws` CLI in PATH, the requests fail and Meshery is unable to connect with the cluster. The error below is actually from Meshery server:
```
time="2021-09-14T10:42:58Z" level=error msg="failed to get K8s Config: Get \"https://<cluster-url>.eks.amazonaws.com/version?timeout=32s\": getting credentials: exec: exec: \"aws\": executable file not found in $PATH"
```

#### Implementation
* **Immediate solution**: Ship `aws` cli in Meshery's docker image so that the client libraries can handle authentication.
* **Possibly better solution**: Use [aws-iam-authenticator](https://github.com/kubernetes-sigs/aws-iam-authenticator) internally. (I'm not entirely sure about this yet)

DelusionalOptimist:

**Description**

This PR fixes #4184 

**Notes for Reviewers**
This PR makes it possible to connect your out of cluster deployments with EKS
* when users call `mesheryctl system config eks`, their user aws token will be obtained and sent to the meshery's docker container. It tries to get tokens using three ways:
```
// Firstly looks for them in environment variables AWS_ACCESS_KEY_ID
// and AWS_SECRET_ACCESS_KEY
// Secondly in the directory $HOME/.aws/
// Lastly tries to take user input and generates a file in ~/.meshery/aws
```
* these auth tokens are then sent to the meshery container so that the client APIs can use the `aws` CLI and interact with the cluster on behalf of the user.

* This PR also adds the `aws` CLI in Meshery server's container. This is very wild but is the truth at least for now. I've written down the reason behind this in https://github.com/meshery/meshery/issues/4184
* Ironically, mesheryctl won't use the `aws` binary and everything is done using aws sdk and docker client.

### DIFF
--- a/mesheryctl/internal/cli/root/system/config.go
+++ b/mesheryctl/internal/cli/root/system/config.go
@@ -24,9 +24,12 @@ import (
 	"os/user"
 	"path/filepath"
 
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/config"
 
 	"github.com/layer5io/meshery/mesheryctl/pkg/utils"
+	meshkitutils "github.com/layer5io/meshkit/utils"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
@@ -39,8 +42,20 @@ import (
 
 // TODO: https://github.com/layer5io/me shery/issues/1022
 
-const paramName = "k8sfile"
-const kubeConfigYaml = "kubeconfig.yaml"
+var (
+	home       = meshkitutils.GetHome()
+	awsDir     = filepath.Join(home, ".aws")
+	credPath   = filepath.Join(awsDir, "credentials")
+	mesheryAWS = filepath.Join(home, ".meshery", "aws")
+)
+
+const (
+	staticProvider       = "StaticProvider"
+	envConfigProvider    = "EnvConfigCredentials"
+	sharedConfigProvider = "SharedConfigCredentials"
+	paramName            = "k8sfile"
+	kubeConfigYaml       = "kubeconfig.yaml"
+)
 
 func getContexts(configFile, tokenPath string) ([]string, error) {
 	client := &http.Client{}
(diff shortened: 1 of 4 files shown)


leecalcote (changes requested):
@DelusionalOptimist a couple lint issues to look at. Will you also include a docs update (with the description you have included in this PR)?

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -85,6 +100,95 @@ func getContexts(configFile, tokenPath string) ([]string, error) {
...
+			return err
+		}
+	}
+
+	// since there is no way yet to check expired creds for StaticProvider
+	// we only check creds obtained from shared directories or environment
+	// variables
+	if credValue.ProviderName == envConfigProvider || credValue.ProviderName == sharedConfigProvider {
+		if creds.IsExpired() {
+			// credentials have expired. generate new
```
hexxdump, line 147: Should we add some log here to capture what exactly failed?

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -85,6 +100,95 @@ func getContexts(configFile, tokenPath string) ([]string, error) {
...
+		log.Println("Enter `aws_secret_access_key`: ")
+		_, err = fmt.Scanf("%s", &secretAccessKey)
+		if err != nil {
+			// err reading entered cred
+			return err
+		}
+
+		creds = credentials.NewStaticCredentials(accessKeyID, secretAccessKey, "")
+		credValue, err = creds.Get()
+		if err != nil {
```
hexxdump, line 136: Should we add some log here to capture what exactly failed?

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -85,6 +100,95 @@ func getContexts(configFile, tokenPath string) ([]string, error) {
...
+
+	err := os.MkdirAll(mesheryAWS, 0755)
+
+	err = os.WriteFile(filepath.Join(mesheryAWS, "credentials"), []byte(fileString), 0644)
+	if err != nil {
+		return err
+	}
+
+	_, err = os.Stat(filepath.Join(mesheryAWS, "credentials"))
+	if err != nil {
```
hexxdump, line 186: simple "return err" would be enough here.

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -85,6 +100,95 @@ func getContexts(configFile, tokenPath string) ([]string, error) {
...
+			return err
+		}
+		err = utils.DockerCp("/meshery_meshery_1", mesheryAWS, "/home/appuser/.aws")
+		if err != nil {
+			return err
+		}
+	}
+
+	err = utils.DockerCp("/meshery_meshery_1", awsDir, "/home/appuser/.aws")
+	if err != nil {
```
hexxdump, line 166: 166-170 can be replaced with "return err" statement

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -85,6 +100,95 @@ func getContexts(configFile, tokenPath string) ([]string, error) {
...
+	}
+
+	// write the tokens to ~/.aws/credentials in case of static and env credentials
+	// providers so that it can be copied into the docker container
+	if credValue.ProviderName == envConfigProvider || credValue.ProviderName == staticProvider {
+		err = createSharedProviderFile(credValue)
+		if err != nil {
+			return err
+		}
+		err = utils.DockerCp("/meshery_meshery_1", mesheryAWS, "/home/appuser/.aws")
```
hexxdump, line 159: Will a non-root user executing mesheryctl have access to copy files into the meshery docker image?

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -85,6 +100,95 @@ func getContexts(configFile, tokenPath string) ([]string, error) {
...
+// TODO: Support EC2 instance role credentials
+// https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.40.43#readme-configuring-credentials
+func awsCredentialHelper() error {
+	log.Info("Looking for aws user credentials in $AWS_ACCESS_KEY_ID and $AWS_SECRET_ACCESS_KEY OR ~/.aws/credentials...")
+	sess := session.Must(session.NewSession())
+	creds := sess.Config.Credentials
+	credValue, err := creds.Get()
+
+	if err != nil {
+		log.Info("Credentials not found. Generating static credentials now...")
```
hexxdump, line 118: Would it be a better to elaborate? i.e.,
Credentials not found in environment variables. Generating static credentials now..

DelusionalOptimist: yes :heavy_check_mark:

mesheryctl/internal/cli/root/system/config.go
DelusionalOptimist: Hmm... only if they're in the `docker` user group which is a basic requirement to use docker :thinking:

navendu-pottekkat:
@DelusionalOptimist With @hexxdump recent refactor, you might need to move this changes to the new subcommand. I missed to give you a heads up before we merged that PR.

navendu-pottekkat:
@DelusionalOptimist Should we close this PR now?

PR #4191, opened 2021-09-15
Added Connectivity check for NATS and MeshSync + Minor Refactor

dhruv0000:
**Description**

This PR fixes:
- Change MeshSync chip to ad hoc connectivity test.
- Change NATS chip to ad hoc connectivity test.

**Notes for Reviewers**

### DIFF
--- a/ui/components/MesheryMeshInterface/PatternService/RJSF.js
+++ b/ui/components/MesheryMeshInterface/PatternService/RJSF.js
@@ -1,8 +1,7 @@
 import React from "react";
-import { utils } from "@rjsf/core";
 import { withTheme } from "@rjsf/core";
 import { Theme as MaterialUITheme } from "@rjsf/material-ui";
-import { Button, Grid, TextField } from "@material-ui/core";
+import { Button, Grid } from "@material-ui/core";
 import { MuiThemeProvider, createTheme } from '@material-ui/core/styles';
 import JS4 from "../../../assets/jsonschema/schema-04.json";
 import ArrayFieldTemplate from "./RJSFCustomComponents/ArrayFieldTemlate"
--- a/ui/components/MesheryMeshInterface/PatternService/RJSFCustomComponents/ObjectFieldTemplate.js
+++ b/ui/components/MesheryMeshInterface/PatternService/RJSFCustomComponents/ObjectFieldTemplate.js
@@ -21,9 +21,7 @@ const useStyles = makeStyles({
 });
 
 const ObjectFieldTemplate = ({
-  DescriptionField,
   description,
-  TitleField,
   title,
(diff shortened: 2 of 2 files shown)


l5io:
Task linked: [CU-1fzt52f Change NATS chip to ad hoc connectivity test.](https://app.clickup.com/t/1fzt52f)

l5io:
Task linked: [CU-1fzt529 Change MeshSync chip to ad hoc connectivity test.](https://app.clickup.com/t/1fzt529)

internal/graphql/resolver/operator.go
```diff
@@ -80,10 +77,9 @@ func (r *Resolver) changeOperatorStatus(ctx context.Context, provider models.Pro
 		err = model.RunMeshSync(kubeclient, del)
 		if err != nil {
 			r.Log.Error(err)
-			// r.operatorSyncChannel <- false
 			r.Broadcast.Submit(broadcast.BroadcastMessage{
 				Type:    broadcast.OperatorSyncChannel,
-				Message: false,
+				Message: err,
```
bariqhibat, line 82: I saw that this Message field can be a boolean or an error. This will pose a problem in the future if we cast the variabel just like what we are doing, for example boolean/error but it turns out to be the other way around. Could we just define a new field just in case, like ok: boolean (and is always there true/false), then error is a null / error. What do you think? Hence, if we find ok = false -> we get the error field.

bariqhibat: this would also make it clearer if the response is bad or not in the switch case in your other file

bariqhibat (commented):
A few comments

internal/graphql/resolver/operator.go
dhruv0000: I kept message as `interface{}` to keep it abstract, since it might be vary from different type of the broadcast. Not sure if it would make sense for other Broadcast type. Said that, I also dont like the idea of not knowing what the message type is. 
One way might be to define individual stuct for each type?

leecalcote:
@dhruv0000 before we create message types, which we should, please see the Messaging Framework and Notification Center design spec. // @mgfeller

internal/graphql/resolver/operator.go
dhruv0000: Although I dont want the overhead of defining a struct type in meshkit once again, in case we are using it to send data locally defined in meshery.

I think adding comments when defining broadcast type about what to expect in message should do for now.

dhruv0000:
Lets merge this till we decide if / what kind of Message Type we need for Broadcaster.

Issue tracked at #4195

utkarsh-pro (approved):
CITY

PR #4267, opened 2021-09-28
New modal

warunicorn19:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -19,12 +19,10 @@
         "classnames": "^2.2.6",
         "eslint-plugin-jest": "^24.4.2",
         "isomorphic-unfetch": "^3.1.0",
-        "moment": "^2.29.1",
         "msw": "^0.35.0",
         "next": "11.1.0",
         "notistack": "^1.0.10",
         "react": "^17.0.2",
-        "react-big-calendar": "^0.35.0",
         "react-dom": "^17.0.2",
         "react-redux": "^7.2.5",
         "react-relay": "^11.0.2",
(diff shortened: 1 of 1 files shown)


warunicorn19:
Aah i guess we wont be able to create new PR's on this branch unless and until the previous is merged.

warunicorn19:
@sudo-NithishKarthik i think the modal here is okay, could we merge this if everything else seems fine? I'll do the `styled` API in a separate PR.

ui/components/Modal.js
```diff
@@ -0,0 +1,41 @@
...
+ * }} props
+ * @returns 
+ */
+
+export const CustomModal = ({
+    open, Content, handleClose, container
+  }) =>  { const theme = useTheme();
+    return (
+      <CustomModalWrapper
+        sx={{theme}}
```
sudo-NithishKarthik, line 27: what's the purpose of this ?

warunicorn19: This styled api thing is confusing the hell out of me, I thought this is how the style is called🤦, but now that i use my small little brain I understand that <CustomModalWrapper /> tag would do that, right?

sudo-NithishKarthik: yes, you actually need not use that `sx` prop here

ui/components/Modal.js
```diff
@@ -0,0 +1,41 @@
...
+        open={open}
+        onClose={handleClose}
+        closeAfterTransition
+        BackdropComponent={Backdrop}
+        BackdropProps={{ timeout : 200, }}
+        container={container}
+      >
+        <Fade in={open} style={{ maxHeight : "90vh", overflow : "auto" }} >{Content}</Fade>
```
sudo-NithishKarthik, line 35: we can use `sx` instead of `style` to keep it consistent.

ui/components/Modal.js
sudo-NithishKarthik: other than that, everything else is fine...

warunicorn19: Okay this is good?

ui/components/Modal.js
warunicorn19: Cool.

ui/components/Modal.js
```diff
@@ -0,0 +1,40 @@
...
+ * @param {{
+ *  open?: boolean,
+ *  Content?: JSX.Element,
+ *  handleClose?: (event: {}, reason: "backdropClick" | "escapeKeyDown") => void,
+ *  container?: React.ReactInstance | (() => React.ReactInstance)
+ * }} props
+ * @returns 
+ */
+
+export const CustomModal = ({
```
sudo-NithishKarthik, line 22: Need not export here since you are exporting it using default export at the end.

sudo-NithishKarthik:
```suggestion
const CustomModal = ({
```

warunicorn19: committed the suggestion @sudo-NithishKarthik

PR #4273, opened 2021-09-29
[Extension] Meshery sidebar collapse onClick plugin menu item

bariqhibat:
**Description**

Collapse Meshery sidebar onclick of a plugin menu item

**Notes for Reviewers**

https://user-images.githubusercontent.com/54849895/135280640-de9ea211-54b3-4dde-b2e1-ddcce43c706d.mp4


l5io:
Task linked: [CU-1hk0c8j [MeshMap] Minimize Meshery menu upon click of MeshMap menu item](https://app.clickup.com/t/1hk0c8j)

bariqhibat:
> @bariqhibat on clicking Meshmap when the navigator is opened does it just close the navigator or does it also take us to MeshMap?

Close and take to Meshmap. In the video, it takes a while since it's in the UI dev or needs time to render

warunicorn19:
Yeah i got confused by the video😅, Cool good stuff.

warunicorn19 (approved):

leecalcote:
Let's ensure that this behavior and the ability to control it is treated as an extension point. Control should be exposed to the plugin to determine on an individual menu item basis what behavior that plugin desires.

leecalcote:
...that ideally, the plugin is provided the option of specifying behavior of what occurs onclick. Each of the menu items defined in Meshery UI or defined in an extensions will then ideally have the option of specifying this behavior: to collapse or not.

bariqhibat:
> ...that ideally, the plugin is provided the option of specifying behavior of what occurs onclick. Each of the menu items defined in Meshery UI or defined in an extensions will then ideally have the option of specifying this behavior: to collapse or not.

I looked into this, but it seems like it requires change from the Meshery Cloud. For now, I implemented generalized way of callbacks. By default , all other tabs are having the tab opened. 

```javascript
  onClickCallback(href){
    if (typeof(href) === "string")
      switch (href){
        case '/extension/meshmap':
          return this.toggleMiniDrawer(true)
        default:
          // by default, sidebar is opened
          return this.toggleMiniDrawer(false)
      }
  }
```

https://user-images.githubusercontent.com/54849895/136195764-609e116e-e443-46cc-916a-3444a42d902b.mp4

utkarsh-pro:
@bariqhibat How about adding a field in the capabilities which can determine the behaviour when we click on the sidebar button?

This will not be as flexible as loading dynamic code for that callback handler (which is totally possible, just a bit harder) as above mentioned approach is "configuration" driven rather "code" driven.

// @leecalcote

bariqhibat:
Hey @utkarsh-pro, yes I wanted to do that as well, but it might require a change in the Meshery Cloud, as I've outlined before. Do correct me if I'm wrong though.

Would `OnClickCollapse (on_click_collapse)` sufficient for this usecase? By design, it's not quite scalable. Do you have any other suggestion?

utkarsh-pro:
@bariqhibat yes you are right, it will require an additional field in the capabilities but isn't that quite desirable considering that we do want the providers to be able to control how the extension is loaded.

However, if this is an urgent thing to be included, we can probably proceed with your suggestion but as you mentioned it may come back to bite us someday :stuck_out_tongue: .

bariqhibat:
Is this good to go? @utkarsh-pro

bariqhibat:
> This should be good now, I made the default of the `on_click_callback` field on the UI to be 0. We just need to change the MeshMap part in the Meshery Cloud to be 1 for this to start working, which can be done after this PR is merged.

Can I check whether this is applied yet in the Meshery Cloud? @utkarsh-pro @leecalcote

l5io:
Task linked: [CU-1mzrywu Use Extension Point to Collapse Meshery UI Menubar onClick of plugin menu item](https://app.clickup.com/t/1mzrywu)

leecalcote:
Gents, I'm not experiencing the behavior of the Meshery menu collapsing onclick of MeshMap. Should I be?

bariqhibat:
> Gents, I'm not experiencing the behavior of the Meshery menu collapsing onclick of MeshMap. Should I be?

Can I check whether the action above has been done on the Meshery Cloud? @utkarsh-pro

utkarsh-pro:
@bariqhibat gotcha

PR #4285, opened 2021-10-01
UI info display fix

### LINKED ISSUE #4265: [UI] Nighthawk test results don't contain test title and URL
#### Current Behavior
The results displayed after running a performance test have empty "title" and "URL" fields.

#### Desired Behavior
The results should have both of these fields.

#### Screenshots/Logs
This can't be a back end issue as the results returned by Meshery server do contain these. See in the below ss:

#### Environment

- Browser: Firefox
- Host OS: Linux
- Meshery Server Version: stable-v0.5.61
- Meshery Client Version: stable-v0.5.61
- Platform: Kubernetes

---

sudhay23:
Signed-off-by: sudhay23 <sudhay2001@gmail.com>

Signed off now

**Description**
The fields in response have now been modified appropriately 

This PR fixes #4265 

**Notes for Reviewers**

### DIFF
--- a/ui/components/MesheryResults.js
+++ b/ui/components/MesheryResults.js
@@ -398,7 +398,7 @@ class MesheryResults extends Component {
             <TableRow>
               <TableCell colSpan={colSpan}>
                 <div className={classes.chartContent}>
-                  <MesheryChart data={[row]} hideTitle />
+                  <MesheryChart rawdata={[self.state.results[rowMeta.dataIndex]]} data={[row]} hideTitle />
                 </div>
                 {boardConfig && boardConfig !== null && Object.keys(boardConfig).length > 0 && (
                   <div>


sudhay23:
@DelusionalOptimist Please verify the changes. Thank you

DelusionalOptimist (changes requested):
@sudhay23 seems like some auto-formatter went over the code and made some unneeded changes. Please revert those and try to use `make run-ui-lint` for linting your changes instead :slightly_smiling_face:

warunicorn19:
Yeah @sudhay23 I can't really find the lines you've made changes to😅, could you please revert the unwanted changes.

sudhay23:
@DelusionalOptimist @warunicorn19 Hope this is fine

DelusionalOptimist:
This is from fortio. URL is fine. Title is missing now. It was working earlier. Could you take a look into it?



This is from nighthawk. URL shows up now. Nice! : )

sudhay23:
@DelusionalOptimist Looks like the 'res' object doesn't have sufficient info.

leecalcote:
@DelusionalOptimist can we provide ourselves named keys and not rely on an index number?

sudhay23:
@warunicorn19 @DelusionalOptimist Can you have a look now and see if the changes are right?

helpers/load_test_interface.go
```diff
@@ -290,7 +290,8 @@ func NighthawkLoadTest(opts *models.LoadTestOptions) (map[string]interface{}, *p
 		OpenLoop:                             &wrappers.BoolValue{Value: false},
 		JitterUniform:                        durationpb.New(0 * time.Second),
 		ExperimentalH2UseMultipleConnections: &wrappers.BoolValue{Value: false},
-		Labels:                               make([]string, 0),
+		// Labels:                               make([]string, 0),
+		Labels:                               labels,
```
warunicorn19, line 294: The golang lint test is failing here because you're using a using string in place of a struct literal, could you please take a look.

DelusionalOptimist:
@sudhay23 @warunicorn19 do we need this change in Meshery server? JSON returned by the server already contains two separate `URL` and `Title` fields. I'm curious, can't we read the entire JSON in the `res` object? So that we don't need to read these from the `labels`? This seems hacky to me:
https://github.com/meshery/meshery/blob/df44458404f7456b8d5e15434dcce13e912f06f2/ui/lib/chartjs-formatter.js#L60-L70

warunicorn19:
🤔you're right @DelusionalOptimist, since the JSON already contains separate fields for the `Title` and the `URL` there's no need for this change in the Meshery server.

sudhay23:
> 🤔you're right @DelusionalOptimist, since the JSON already contains separate fields for the `Title` and the `URL` there's no need for this change in the Meshery server.

Oh yeah. But I still believe that only the sub-property of the whole object that excludes URL and Title are being returned though they are logged. Correct me if I'm wrong 😅 @warunicorn19 @DelusionalOptimist

sudhay23:
This change was done as I see that the other tests like Fortio are having the Labels field properly filled unlike Nighthawk which lacks it

helpers/load_test_interface.go
sudhay23: @leecalcote @warunicorn19 Thanks for pointing it out. I've made the appropriate fix locally but haven't committed yet. I think I'm close to a potential solution

PR #4327, opened 2021-10-04
Seeding Applications for local provider

Revolyssup:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/handlers/oam_handler.go
+++ b/handlers/oam_handler.go
@@ -127,7 +127,7 @@ func (h *Handler) PatternFileHandler(
 // These operations can be:
 // 1. Adding a workload/trait/scope
 // 2. Getting list of workloads/traits/scopes
-func (h *Handler) OAMRegisterHandler(rw http.ResponseWriter, r *http.Request) {
+func (h *Handler) OAMRegisterHandler(rw http.ResponseWriter, r *http.Request, pref *models.Preference, u *models.User, prov models.Provider) {
 	typ := mux.Vars(r)["type"]
 
 	if !(typ == "workload" || typ == "trait" || typ == "scope") {
@@ -137,7 +137,7 @@ func (h *Handler) OAMRegisterHandler(rw http.ResponseWriter, r *http.Request) {
 
 	method := r.Method
 	if method == "POST" {
-		if err := h.POSTOAMRegisterHandler(typ, r); err != nil {
+		if err := h.POSTOAMRegisterHandler(typ, r, prov); err != nil {
 			rw.WriteHeader(http.StatusInternalServerError)
 			h.log.Debug(err)
 			_, _ = rw.Write([]byte(err.Error()))
@@ -160,7 +160,7 @@ func (h *Handler) OAMRegisterHandler(rw http.ResponseWriter, r *http.Request) {
 // 	200:
 
 // POSTOAMRegisterHandler handles registering OMA objects
-func (h *Handler) POSTOAMRegisterHandler(typ string, r *http.Request) error {
+func (h *Handler) POSTOAMRegisterHandler(typ string, r *http.Request, prov models.Provider) error {
 	// Get the body
 	body, err := ioutil.ReadAll(r.Body)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


handlers/k8sconfig_handler.go
```diff
@@ -325,22 +325,20 @@ func (h *Handler) setupK8sConfig(inClusterConfig bool, k8sConfigBytes []byte, co
...
 	if man == nil {
-		l.Error(errors.New("Could not get k8s components"))
+		l.log.Error(errors.New("Could not get k8s components"))
 		return err
 	}
 	for i, def := range man.Definitions {
 		var ord core.WorkloadCapability
 		ord.Metadata = make(map[string]string)
 		ord.Metadata["io.meshery.ctxid"] = ctx
-		ord.Metadata["adapter.meshery.io/name"] = "kubernetes"
```
utkarsh-pro, line 342: @Revolyssup why are we getting rid of these?

handlers/k8sconfig_handler.go
```diff
@@ -325,22 +325,20 @@ func (h *Handler) setupK8sConfig(inClusterConfig bool, k8sConfigBytes []byte, co
 	kc.ClusterConfigured = true
 	return kc, nil
 }
-func registerK8sComponents(l logger.Handler, config []byte, ctx string) error {
-	l.Info("Starting to register k8s native components")
+func registerK8sComponents(l *Handler, config []byte, ctx string) error {
```
utkarsh-pro, line 328: Rename `l` to something more intuitive.

utkarsh-pro, line 328: Replace `l` with something more intuitive

handlers/k8sconfig_handler.go
```diff
@@ -354,10 +352,16 @@ func registerK8sComponents(l logger.Handler, config []byte, ctx string) error {
 			return err
 		}
 		err = core.RegisterWorkload(content)
+		// Everytime a worload is registered successfully, the SeedApplication function will run and seed all the applications that can be handled with local provider.
+		if err == nil {
+			if l.config.Providers["None"] != nil {
```
utkarsh-pro, line 357: Instead receive `prov models.Provider` from the calling function and check `if prov.Name() == "None" { ...` as this can break somewhere, however interface won't.

handlers/oam_handler.go
```diff
@@ -168,10 +168,22 @@ func (h *Handler) POSTOAMRegisterHandler(typ string, r *http.Request) error {
 	}
 
 	if typ == "workload" {
-		return core.RegisterWorkload(body)
+		err := core.RegisterWorkload(body)
```
utkarsh-pro, line 171: Why duplicate if you want to run it in every case? I can see it is not being replicated when registering the scopes, that is not desirable though.

handlers/seed.go
```diff
@@ -0,0 +1,83 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/layer5io/meshery/models"
+	"github.com/layer5io/meshery/models/pattern/core"
+	"github.com/layer5io/meshery/models/pattern/stages"
+	"github.com/layer5io/meshkit/utils"
+)
+
+func SeedApplications(l models.Provider) {
```
leecalcote, line 14: Individually registering different types of seed content seems unnecessary. Is there value in treating content separately by type or is it more efficient to `RegisterSeedContent()`?

utkarsh-pro (commented):
@Revolyssup did you happen to accidentally commit meshery binary?

leecalcote:
@utkarsh-pro, given that we need seed content included in the v0.6.0 release, please collaborate in changing code as-needed.

handlers/k8sconfig_handler.go
Revolyssup: @utkarsh-pro  this PR is older than the recent changes in master that were made when these fields were added, originally these fields weren't there. That is why they are missing here, adding those fields accordingly now

handlers/k8sconfig_handler.go
Revolyssup: roger that

handlers/oam_handler.go
Revolyssup: was confused here , about if we wanted it to trigger for scopes too. Will make the change

Revolyssup:
> @Revolyssup did you happen to accidentally commit meshery binary?

I guess so, removing it now

handlers/seed.go
leecalcote: I'll suggest that we're trying to be too smart here, and will only end up refactoring this logic later.

Those smarts are necessary and should be defined and executed in one location: the pattern engine (we need to give this a name).

Issue #4103, opened 2021-09-07
[mesheryctl] Migrate from Manifests to Helm for Meshery Server/Adapter deployment


leecalcote:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

#### Current Behavior
1. Server and Adapters are provisioned with manifests.
1. Meshery Operator are  provisioned with helm charts, which uses `ServiceAccountName` and that dependencies are provisioned in correct order of operation. 

#### Desired Behavior
1. Use MeshKit helm chart deployment functionality using https://meshery.io/charts.

---
#### Contributor Resources
- [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- 🙋🏾🙋🏼 Questions: [Layer5 Discussion Forum](https://discuss.layer5.io) and [Layer5 Community Slack](http://slack.layer5.io)

SirZeck:
Can I work on this?

warunicorn19:
Sure @SirZeck

SirZeck:
Umm, can someone help me with where should I start with?




























-09

pottekkat:
@SirZeck Currently we use Kubernetes manifests to deploy Meshery in mesheryctl. This is done using the Kubernetes Go-Client.
What we have to do here is to replace this with deploying Meshery via Helm. This functionality is available in [MeshKit](https://github.com/meshery/meshkit) and the Helm charts are present [here](https://github.com/meshery/meshery/tree/master/install/kubernetes/helm/meshery).

panyuenlau:
Please assign this to me

navendu-pottekkat:
Assigned! @panyuenlau

pottekkat:
Helm chart package in MeshKit: https://github.com/meshery/meshkit/blob/master/utils/kubernetes/apply-helm-chart.go

[...]

PR #4389, opened 2021-10-11
Migrate from plain manifests to helm charts for meshery installation

panyuenlau:
Signed-off-by: Darren Lau <panyuenlau@gmail.com>

**Description**

This PR fixes #https://github.com/meshery/meshery/issues/4103

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/pkg/utils/helpers_test.go
+++ b/mesheryctl/pkg/utils/helpers_test.go
@@ -513,8 +513,8 @@ func TestSetOverrideValues(t *testing.T) {
 	testChannel := "testChannel"
 
 	tests := []struct {
-		name                   string
-		ctx                    *config.Context
+		name                string
+		ctx                 *config.Context
 		mesheryImageVersion string
 		want                map[string]interface{}
 	}{
@@ -522,7 +522,7 @@ func TestSetOverrideValues(t *testing.T) {
 			name: "Context contains no adapters and no meshery image version",
 			ctx: &config.Context{
 				Adapters: nil,
-				Channel: testChannel,
+				Channel:  testChannel,
 			},
 			mesheryImageVersion: "",
 			want: map[string]interface{}{
@@ -562,7 +562,7 @@ func TestSetOverrideValues(t *testing.T) {
 			name: "Context contains part of all available adapters and meshery image version",
 			ctx: &config.Context{
 				Adapters: []string{"meshery-istio", "meshery-osm", "meshery-nsm"},
-				Channel: testChannel,
+				Channel:  testChannel,
 			},
 			mesheryImageVersion: "testImageVersion",
 			want: map[string]interface{}{


panyuenlau:
> @panyuenlau Thank you for taking this up.
> 
> Heads up, there are some lint errors. We use [golangci-lint](https://github.com/golangci/golangci-lint).
> 
> Couple of things:
> 
> 1. We should be downloading the charts to system when we `system start`.
> 2. The downloaded charts should ideally be "cached" in the sense that if there is no need to update, we should not try to fetch the latest Helm chart. (We could discuss the implementation of this)
> 3. `system start` with Meshery running throws an error. Ideally `system start` should be idempotent. We might need to update the helper function in MeshKit to take this into consideration. I'm not quite sure to the extent in which Helm supports this cleanly. See: [helm/helm#3134](https://github.com/helm/helm/issues/3134)
> 
> Screenshot 2021-10-11 at 11 39 36 PM
> 
> 1. Same goes for `system update`. It fails to execute because of the same reason.
> 
> Screenshot 2021-10-11 at 11 47 53 PM

Hey @navendu-pottekkat, thank you for your review
1. The `ApplyHelmChart()` actually downloads the helm chart to the `/tmp` folder by calling the `fetchHelmChart()` function( reference: https://github.com/panyuenlau/meshkit/blob/f0a31a1b873effe10f500c5cb23d08177cbb977c/utils/kubernetes/apply-helm-chart.go#L307), which is different from where we have the manifests downloaded: `~/.meshery`
2. The `fetchHelmChart()` checks if the required helm chart is already present in the user's file system, reference: https://github.com/panyuenlau/meshkit/blob/f0a31a1b873effe10f500c5cb23d08177cbb977c/utils/kubernetes/apply-helm-chart.go#L330-L332 
3. `helm install` is indeed not idempotent
4. For `system update`, the meshkit helper function (`ApplyHelmChart`) needs to be updated to have the option to do `helm upgrade` instead of `helm install`

panyuenlau:
@navendu-pottekkat The `system update` command now works with this PR (https://github.com/layer5io/meshkit/pull/118) in the meshkit, the PR will also resolve linters issues identified by the CI

panyuenlau:
> @panyuenlau I'm arriving late to the party sweat_smile Seems like we still have the support for meshconfig (`~/.meshery/config.yaml`) under works. This was the one I used:
> 
> ```
> contexts:
>   local:
>     endpoint: http://192.168.49.2:9081
>     token: Default
>     platform: kubernetes
>     adapters:
>     - meshery-istio
>     channel: edge
>     version: latest
> current-context: local
> tokens:
> - location: auth.json
>   name: Default
> ```
> 
> * We might wanna have control over which adapters to install. Right now, it installs all of them regardless the ones specified.

The implementation now only installs the adapters that are specified in the context

panyuenlau:
@navendu-pottekkat right we don't need to delete all pods and re-deploy them, `helm upgrade` detects and restarts the pods that need an update. It's fixed now

> Tested it with the latest release of MeshKit. Everything looks and works good.
> 
> Screenshot 2021-10-16 at 3 44 25 AM Screenshot 2021-10-16 at 3 44 39 AM
> 
> One thing to add here is some logs when `mesheryctl system update` is run. It shows only logs from resetting as shown in the screenshot above.
> 
> Related to that, do we need to delete and redeploy Meshery on `system update`? `update` should fetch updated Helm charts (if `skipResetFlag` is not set) and pull the latest images from docker hub. Is that functionality achieved without having to restart?

@navendu-pottekkat Thank you for helping to test that! 

And right we don't need to delete all pods and re-deploy them, `helm upgrade` detects and restarts the pods that need an update. It's fixed now

PR #4438, opened 2021-10-17
[meshery server] switch to helm based deployments for meshery operator

DelusionalOptimist:

**Description**

This PR switches the deployment method of operator from manifests to helm in meshery server.

**Notes for Reviewers**
* Install the operator and CRs using helm.
* Fetches the tar containing helm charts and installs subcharts by referencing the `LocalPath` of respective components.
* Bundles charts into the docker container.

Some considerations to think upon:
* Stopping meshery server shouldn't uninstall the CRs as it might not be desired in real deployments. Thus created a new make target `cleanup-cluster` for this.
* The download location of charts should align with those of mesheryctl (`~/.meshery/manifests`) but this needs an unpacked tar to install the sub charts and mesheryctl stores a gzipped tar. Though no conflict arises due to this.

### DIFF
--- a/install/kubernetes/helm/meshery-operator/Chart.lock
+++ b/install/kubernetes/helm/meshery-operator/Chart.lock
@@ -1,36 +1,9 @@
 dependencies:
-- name: meshery-istio
+- name: meshery-broker
   repository: ""
-  version: 2.1.2
-- name: meshery-linkerd
+  version: 0.1.0
+- name: meshery-meshsync
   repository: ""
-  version: 2.1.2
-- name: meshery-consul
-  repository: ""
-  version: 2.1.2
-- name: meshery-kuma
-  repository: ""
-  version: 2.1.2
-- name: meshery-osm
-  repository: ""
-  version: 2.1.2
-- name: meshery-nsm
-  repository: ""
-  version: 2.1.2
-- name: meshery-traefik-mesh
-  repository: ""
-  version: 2.1.2
-- name: meshery-cpx
-  repository: ""
-  version: 2.1.2
-- name: meshery-nginx-sm
-  repository: ""
-  version: 2.1.2
-- name: meshery-app-mesh
-  repository: ""
-  version: 2.1.2
-- name: meshery
-  repository: ""
-  version: 2.1.2
-digest: sha256:b1230f6f0c6dd0017f96599eaaf38d8dc6d4127dc285a06464d5e04608754abf
-generated: "2021-10-29T15:10:39.691795+08:00"
+  version: 0.1.0
+digest: sha256:276d1539ccf683459c4212861d3b4640d5d718ae877ebc3404c47a3a224dfb7c
+generated: "2021-11-10T16:45:44.540012823+05:30"
--- a/install/kubernetes/helm/meshery-operator/Chart.yaml
+++ b/install/kubernetes/helm/meshery-operator/Chart.yaml
@@ -29,6 +29,14 @@ version: 0.6.0
 # TODO: Need the confirm the image tag.
 appVersion: 0.6.0
 
+dependencies:
+- name: meshery-broker
+  version: 0.1.0
+  condition: meshery-broker.enabled
+- name: meshery-meshsync
+  version: 0.1.0
+  condition: meshery-meshsync.enabled
+
 maintainers:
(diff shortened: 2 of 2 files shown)


leecalcote:
Ultimately, we'll want to keep the helm charts / manifests / compose files under a directory-named by release version.

navendu-pottekkat:
@DelusionalOptimist Are we ready for a review? Does this PR take into account of the updated Helm charts?

DelusionalOptimist:
> @DelusionalOptimist Are we ready for a review? Does this PR take into account of the updated Helm charts?

No @navendu-pottekkat, it doesn't, yet. Yes, it will have some changes. Updating here :+1:

DelusionalOptimist:
I tried using the new charts so as to see what happens if meshery server tries to (un)install the operator now.
Scenarios:
* Meshery out of cluster:
	- [x] Installs the operator and the CR(D)s in cluster when toggled OR kubecontext is added.
	- [x] Uninstalls the operator and the CR(D)s from cluster when toggled OR kubecontext is removed.
	- [x] Cleans up everything on shutdown.

* Meshery in cluster:
	- [ ] Uninstalls the operator and the CR(D)s from cluster when toggled or kubecontext is removed.
		* This doesn't work out. Reason - meshery server is a dependency in the new charts so it uninstalls itself when it tries to uninstall the operator.
	- [ ] Installs the operator and CR(D)s in cluster when toggled or kubecontext is removed.
		* Couldn't test this because of the previous point.

DelusionalOptimist:
@Aisuko in the new charts for meshery-operator - is it possible to avoid having meshery server as a dependency?

Aisuko:
> @Aisuko in the new charts for meshery-operator - is it possible to avoid having meshery server as a dependency?
> I mean, can we remove meshery server as a dependency and keep that chart solely for installing the operator?

Hi @DelusionalOptimist, yes. We have the logic in dependency, please replace the `enabled: true` to `false`.

https://github.com/meshery/meshery/blob/d43bd16ddabb0f650434e3021324e0f712e61134/install/kubernetes/helm/meshery-operator/values.yaml#L157-L160

internal/graphql/model/operator_helper.go
```diff
@@ -25,21 +25,21 @@ const (
...
-	brokerYaml   = "https://raw.githubusercontent.com/layer5io/meshery-operator/master/config/samples/meshery_v1alpha1_broker.yaml"
+	operatorHelmChart = "meshery-operator"
+	brokerHelmChart   = "meshery-broker"
 )
 
 func Initialize(client *mesherykube.Client, delete bool) error {
-	// installOperator
-	err := applyYaml(client, delete, operatorYaml)
+	// install CRDs and operaotr
+	err := installUsingHelm(client, delete, operatorHelmChart)
```
Aisuko, line 34: Here we can use the `meshery-operator` chart as the mainifest though the set all the sub-charts to false. We can set the default value `false`  to the `meshery-operator` sub-charts, and dynamic set the value though `mesheryctl` and internal function.

https://github.com/meshery/meshery/blob/d43bd16ddabb0f650434e3021324e0f712e61134/install/kubernetes/helm/meshery-operator/values.yaml#L157-L160

internal/graphql/model/helpers.go
```diff
@@ -114,3 +140,182 @@ func applyYaml(client *mesherykube.Client, delete bool, file string) error {
...
+
+	return nil
+}
+
+// SetOverrideValues detects the currently insalled adapters and sets appropriate
+// overrides so as to not uninstall them. It also sets overrides for operator to
+// false or true if it is a delete operation
+func SetOverrideValues(delete bool) (map[string]interface{}, error) {
+	installedAdapters := make([]string, 0)
+	adapters := helpers.NewAdaptersTracker(viper.GetStringSlice("ADAPTER_URLS")).GetAdapters(context.TODO())
```
DelusionalOptimist, line 226: @utkarsh-pro @Revolyssup I want to check what all adapters are installed and apply overrides accordingly so that upgrading the helm release doesn't uninstall any of them.
`GetAdapters()` should return all the data about adapters which can then be used to check which one is installed or not, right? Basically, this is the one behind `/api/system/adapters` endpoint's response.
However, it doesn't seem to have all the data. Do you know why?

PR #4515, opened 2021-10-28
[Extension] Added `Hide label` btn in user preferences

NikhilSharma03:
**Description**

- Added `Hide label` in user preferences

### DIFF
--- a/ui/components/UserPreferences.js
+++ b/ui/components/UserPreferences.js
@@ -26,7 +26,7 @@ import Tab from '@material-ui/core/Tab';
 import { Paper, Tooltip } from '@material-ui/core';
 import SettingsRemoteIcon from '@material-ui/icons/SettingsRemote';
 import SettingsCellIcon from '@material-ui/icons/SettingsCell';
-import ExtensionSandbox from "./ExtensionSandbox";
+import ExtensionPoint from "./ExtensionPoint";
 import RemoteUserPref from "./RemoteUserPref";
 import ExtensionPointSchemaValidator from "../utils/ExtensionPointSchemaValidator";
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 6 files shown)


ui/components/UserPreferences.js
```diff
@@ -95,10 +108,17 @@ class UserPreference extends React.Component {
       startOnZoom : props.startOnZoom,
       tabVal : 0,
       userPrefs : ExtensionPointSchemaValidator("user_prefs")(),
-      providerType : ''
+      providerType : '',
+      showBtnLabels : true
     };
   }
 
+  showBtnLabelHandler = () => {
```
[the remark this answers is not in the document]
leecalcote: @NikhilSharma03 what do you think?

ui/components/UserPreferences.js
```diff
@@ -300,7 +324,38 @@ class UserPreference extends React.Component {
             <MesherySettingsPerformanceComponent />
           }
           {tabVal == 2 && userPrefs && providerType != 'local' &&
-            <ExtensionSandbox type="user_prefs" Extension={(url) => RemoteUserPref({ startOnZoom, handleToggle, url })} />
+            <div>
```
leecalcote, line 327: @utkarsh-pro this should be an empty `div`, yes? The contents of this extension point should come from the connected remote provider, if that remote provider offers this capability.

utkarsh-pro: @leecalcote, yes. @NikhilSharma03 I think the changes that are done here should rather be done on the provider side.

utkarsh-pro: I see that a table container is added, is there some issue that you faced while adding this to the extension @NikhilSharma03?

NikhilSharma03: > @leecalcote, yes. @NikhilSharma03 I think the changes that are done here should rather be done on the provider side.

Does provider side mean here :sweat_smile:  https://github.com/meshery/meshery/blob/a541cda0a171c470a75f78ef736f7427f5c6c8f8/ui/components/NavigatorExtension.js

leecalcote:
@NikhilSharma03 please close when this has transitioned to the remote provider.

NikhilSharma03:
@leecalcote I think I am a bit confused here. That is if we want to land that `Hide Label Button` here in user preferences 

Screenshot from 2021-11-04 20-03-58

Then we have to made those changes here in meshery repo where UserPreference.js file resides.
If we want all changes in remote provider then the `Hide Label Button` button should reside in MeshMap then.

leecalcote:
@NikhilSharma03 be sure that you follow up with @Abhishek-kumar09

PR #4577, opened 2021-11-07
[Lint] Change eslint configuration for new styles

Abhishek-kumar09:
**Description**
The bots (sonatype)  and other ci are throwing semicolon errors which is not fixed by `eslint --fix`, Either we should choose to remove the use of semicolons completely, or to use semicolons everywhere required. We can't stay in between since it will create linting issues for other CIs. 
Next, moving to good practices. There should be no spacing required before colons when declaring objects or writing `key: value` pairs. Infact, when we autoformat the code within the IDE(vscode/intelliJ) it formats the code by removing all the spaces before colons(":").
Thus made two changes in the eslint config file to
- Add semicolon as required (We can also choose to remove semicolons from all places as the other alternative)
- Remove whitespace before colons (To get in sync with latest practices) 

Now when you run `eslint --fix` then it will autoformat the code to follow the rules and will not throw error or become burden on developer to add semicolons or remove extra whitespaces.


leecalcote:
@Abhishek-kumar09 @sudo-NithishKarthik, what do we want to do about this bot? Keep it or remove it?

sudo-NithishKarthik:
This bot is not helping us. It is better to be rid of it.

Abhishek-kumar09:
> This bot is not helping us. It is better to be rid of it.

Agree with removing it since it doesn't follow react syntax. maybe it need more customization.

leecalcote:
Thank you, gents. I gave this bot the 👢

sudo-NithishKarthik:
Thank you @Abhishek-kumar09 .
Merge conflicts cropped up ^^

Abhishek-kumar09:
@sudo-NithishKarthik 
I have created a new PR for this, Actually, there were merge conflicts raised and there were problems at some points choosing between the changes, so decided not to take risk.  https://github.com/meshery/meshery/pull/4627

PR #4598, opened 2021-11-10
[Security] fix CVE-2020-26160 with dgrijalva/jwt-go

s1ntaxe770r:
**Description**
jwt-go allows attackers to bypass intended access restrictions in situations with []string{} for m["aud"] (which is allowed by the specification). Because the type assertion fails, "" is the value of aud. This is a security problem if the JWT token is presented to a service that lacks its own audience check. There is no patch available and users of jwt-go are advised to migrate to golang-jwt at version 3.2.1
This PR fixes #
CVE-2020-26160

**Notes for Reviewers**
see https://nvd.nist.gov/vuln/detail/CVE-2020-26160
this vulnerability is rated high severity

### DIFF
--- a/models/remote_auth.go
+++ b/models/remote_auth.go
@@ -14,7 +14,7 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/dgrijalva/jwt-go"
+	"github.com/golang-jwt/jwt"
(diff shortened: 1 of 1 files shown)


leecalcote (approved):
LGTM

PR #4626, opened 2021-11-12
Add a flag and make recipe to skip k8s component generation in dev env

Revolyssup:

**Description**
Currently k8s component generation takes place in a ubiquitous way. But a lot of times, its not really required but on subsequent restarts in the dev environment, it keeps on re-registering as it loses context especially with local provider, bogging down cpu. So it would be nice to have a flag, which can be set to true when we want to skip the k8s component generation. 
The default behaviour remains the same. It is only an addition for dev environment. 

**Notes for Reviewers**


leecalcote (approved):
Thank you. ✔️

PR #4627, opened 2021-11-12
[Refractor]  Change eslint configuration for new styles

Abhishek-kumar09:
From https://github.com/meshery/meshery/pull/4577
**Description**
The bots (sonatype) and other ci are throwing semicolon errors which is not fixed by eslint --fix, Either we should choose to remove the use of semicolons completely, or to use semicolons everywhere required. We can't stay in between since it will create linting issues for other CIs.
Next, moving to good practices. There should be no spacing required before colons when declaring objects or writing key: value pairs. Infact, when we autoformat the code within the IDE(vscode/intelliJ) it formats the code by removing all the spaces before colons(":").
Thus made two changes in the eslint config file to

    Add semicolon as required (We can also choose to remove semicolons from all places as the other alternative)
    Remove whitespace before colons (To get in sync with latest practices)

Now when you run eslint --fix then it will autoformat the code to follow the rules and will not throw error or become burden on developer to add semicolons or remove extra whitespaces.


leecalcote:
Merge conflicts...

leecalcote:
Oh, I see. So long none of the @meshery/meshery-ui-maintainers have dissenting opinions, for my part, I'm good with either lint conventions. The important bit here being that we're enforcing consistency here and in other repos.

leecalcote (approved):
Lgtm

sudo-NithishKarthik:
Yes @Abhishek-kumar09 , thanks for this. 
I would say we go with keeping the semicolons because I have personally ran into some ambiguous situations where the compiler failed to interpret some statements properly because of the absence of semicolons.

sudo-NithishKarthik (dismissed):
Totally makes sense 👍

Abhishek-kumar09:
@leecalcote Ping!

leecalcote:
Merge conflicts...

leecalcote:
merge conflict, @Abhishek-kumar09

PR #4641, opened 2021-11-18
switch to helm based operator deployments in meshery server

DelusionalOptimist:
**Description**

This PR fixes #

**Notes for Reviewers**
* https://github.com/meshery/meshery/commit/4612faa3b46897f151f19676b99127624f5fb1ec just replaces the existing mechanism from install kubernetes manifests to installing helm charts
* https://github.com/meshery/meshery/commit/8854f0bf0135443522a0900acc1acdd3952f581c addresses these:
	* Adding appropriate overrides so that adapters don't get uninstalled when a release is upgraded. Only happens when meshery in cluster
	* Uses meshery-operator chart when meshery is not in cluster.
* https://github.com/meshery/meshery/commit/49694703b5e2aee1c2583a625186275d282f4dc8 uninstalls the meshery-operator helm chart when meshery server shuts down. Helm here takes care of uninstalling clusterroles and clusterrolebindings.

### DIFF
--- a/internal/graphql/model/error.go
+++ b/internal/graphql/model/error.go
@@ -62,5 +62,5 @@ func ErrMesheryClient(err error) error {
 
 // ErrApplyHelmChart is the error which occurs while applying helm chart
 func ErrApplyHelmChart(err error) error {
-	return errors.New(ErrApplyHelmChartCode, errors.Alert, []string{"Error occured while applying Helm Chart"}, []string{err.Error()}, []string{"Kubernetes cluster might not be connected", "Leftover resources from partial install"}, []string{"Try reinstalling", "Try reconnecting your kubernetes cluster", "Clean up artifacts from preinstalled helm release manually"})
+	return errors.New(ErrApplyHelmChartCode, errors.Alert, []string{"Error occurred while applying Helm Chart"}, []string{err.Error()}, []string{"Kubernetes cluster might not be connected", "Leftover resources from partial install"}, []string{"Try reinstalling", "Try reconnecting your kubernetes cluster", "Clean up artifacts from preinstalled helm release manually"})
 }


DelusionalOptimist:
docker img: https://hub.docker.com/layers/luislopez3601/meshery/latest/images/sha256-e4528b448662f4b66c7523bba2279e97a3f9898822bcef31880b1aefb58fa8dc?context=explore
install using:
```
helm install meshery meshery/meshery -n meshery --create-namespace --set image.repository=luislopez3601/meshery,image.tag=latest,image.pullPolicy=IfNotPresent
```

leecalcote:
@100mik it might interest you to review this PR.

leecalcote:
@panyuenlau, your review here would be great.

internal/graphql/model/helpers.go
```diff
@@ -114,3 +123,121 @@ func applyYaml(client *mesherykube.Client, delete bool, file string) error {
...
+	}
+
+	err = client.ApplyHelmChart(mesherykube.ApplyHelmChartConfig{
+		Namespace: "meshery",
+		ChartLocation: mesherykube.HelmChartLocation{
+			Repository: chartRepo,
+			Chart:      chart,
+			Version:    mesheryReleaseVersion,
+		},
+		CreateNamespace: true,
```
hexxdump, line 164: @DelusionalOptimist, should this flag be set to true, even when the action is UNINSTALL ?

internal/graphql/model/operator_helper.go
```diff
@@ -25,22 +25,23 @@ const (
...
 		return err
 	}
 
+	// not needed
 	// installBroker
-	err = applyYaml(client, delete, brokerYaml)
-	if err != nil {
-		return err
-	}
+	// err = applyYaml(client, delete, brokerYaml)
```
hexxdump, line 41: we can remove the commented code, if there is no reason to use yamls going forward.

internal/graphql/resolver/meshsync.go
```diff
@@ -112,23 +112,23 @@ func (r *Resolver) connectToBroker(ctx context.Context, provider models.Provider
...
-	if err != nil {
-		r.Log.Error(err)
-		r.Broadcast.Submit(broadcast.BroadcastMessage{
-			Source: broadcast.OperatorSyncChannel,
-			Data:   err,
-			Type:   "error",
-		})
-		return model.StatusDisabled, err
-	}
+	//	if err != nil {
```
hexxdump, line 123: We can remove this as well.

internal/graphql/model/helpers.go
DelusionalOptimist: @hexxdump this flag is actually meaningless when the action is UNINSTALL. It doesn't have any effect on the ns. But it sure is confusing... I'll add a comment :+1:

leecalcote:
with the urgency high here, I'm merging on what seems to be a good solution. Users are running into conflicts between manifest and helm-based deployments...

leecalcote (approved):

internal/graphql/model/helpers.go
hexxdump: thanks for clarifying that @DelusionalOptimist !

PR #4689, opened 2021-11-30
[Broker] Broker connection GraphQL subscription

bariqhibat:

**Description**

This PR fixes #

**Notes for Reviewers**
A subscription for getting latest broker connection through default checks in the resolver. State management will really complements this as we keep track of state of broker connection.

### DIFF
--- a/internal/graphql/resolver/broker.go
+++ b/internal/graphql/resolver/broker.go
@@ -1,0 +1,51 @@
+package resolver
+
+import (
+	"context"
+	"time"
+)
+
+func (r *Resolver) subscribeBrokerConnection(ctx context.Context) (<-chan bool, error) {
+	ch := make(chan bool)
+	previousIsConnected := r.getBrokerConnection()
+	go func(outCh chan bool) {
+		outCh <- r.getBrokerConnection()
+		for {
+			connected := r.getBrokerConnection()
+
+			// only send message to the client if current state is different from last state
+			if connected != previousIsConnected {
+				// send message
+				outCh <- connected
+
+				// change reference of last to the current one for next iterations
+				previousIsConnected = connected
+			}
+			select {
+			// flush subscription
+			case <-ctx.Done():
+				close(outCh)
+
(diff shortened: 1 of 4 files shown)


leecalcote:
@utkarsh-pro you're up.

utkarsh-pro (approved):
LGTM 🎉

internal/graphql/resolver/broker.go
```diff
@@ -0,0 +1,51 @@
...
+	}(ch)
+
+	return ch, nil
+}
+
+func (r *Resolver) getBrokerConnection() bool {
+	connected := !r.BrokerConn.IsEmpty()
+
+	if connected {
+		err := r.BrokerConn.Publish("connection.test", nil)
```
leecalcote, line 44: Are we able to use MeshKit errors here?

bariqhibat: Can you clarify more on this? Since we wouldn't be using this errors anywhere. Here we're just basically saying that if there's any error, then the connected variable is false

leecalcote: When generating errors using Golang, the Meshery project uses a standard mechanism using a function in MeshKit. This framework facilitates the partially, self-documenting error code system.

bariqhibat: This is already documented by [Meshkit](https://github.com/layer5io/meshkit/blob/master/broker/nats/nats.go#L76)
```go
// Publish - to publish messages
func (n *Nats) Publish(subject string, message *broker.Message) error {
	err := n.ec.Publish(subject, message)
	if err != nil {
		return ErrPublish(err)
	}
	return nil
}
```
Do you mean we'd like this to be logged to the Meshery Server everytime we catch this error? @leecalcote

bariqhibat: Can I check if this is still an issue? @leecalcote

leecalcote (approved):

PR #4710, opened 2021-12-03
Run tests under same performance profile if it is already present

pottekkat:
Signed-off-by: Navendu Pottekkat <navendupottekkat@gmail.com>

**Description**

Updates `mesheryctl perf apply` to first check if a performance profile is already created before running a test under a new performance profile with the same name.

Runs tests under the same profile if it is already present.

This works for tests run with a configuration file, flags or a profile name.

**Notes for Reviewers**

A new release is needed after merging this PR to use this change in the [meshery-smp-action](https://github.com/layer5io/meshery-smp-action/).

### DIFF
--- a/docs/_data/mesheryctlcommands/cmds.yml
+++ b/docs/_data/mesheryctlcommands/cmds.yml
@@ -382,7 +382,7 @@ perf:
           mesheryctl perf apply [profile-name] --flags
 
           # Execute a Performance test without a specified performance profile
-          mesheryctl perf apply --profile [profile-name] --url [url]
+          mesheryctl perf apply [profile-name] --url [url]
 
           # Run Performance test using SMP compatible test configuration
           mesheryctl perf apply -f [filepath]
@@ -439,14 +439,6 @@ perf:
             mesheryctl perf apply [profile-name] --url [URL] --name "name"
           example:
             mesheryctl perf apply local-perf --url https://192.168.1.15/productpage --name "local-test"
-        profile:
-          name: --profile 
-          arg: apply
-          description: '(required/optional) Name for the new Performance Profile.'
-          usage:
-            mesheryctl perf apply --profile [profile-name] --url [URL]
-          example:
-            mesheryctl perf apply --profile local-perf --url https://192.168.1.15/productpage
         qps:
           name: --qps 
           arg: apply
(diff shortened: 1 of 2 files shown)


navendu-pottekkat:
There shouldn't be a reason for failing tests. I'm not sure why...

piyushsingariya:
> There shouldn't be a reason for failing tests. I'm not sure why...

You have made quite a significant amount of changes, tests are supposed to fail, I feel!

navendu-pottekkat:
> > There shouldn't be a reason for failing tests. I'm not sure why...
> 
> You have made quite a significant amount of changes, tests are supposed to fail, I feel!

It shouldn't change the behaviour though. I could not figure it out instantly. Looking into it in detail.

navendu-pottekkat:
@piyushsingariya Removed the `--perf` flag and is now only using the argument with the profile name.

Rest of the functionality is the same with the ability to override the values in a test configuration with flags.

I will update the unit tests if this is okay.

mesheryctl/internal/cli/root/perf/apply.go
```diff
@@ -86,6 +88,7 @@ mesheryctl perf apply local-perf --url https://192.168.1.15/productpage --mesh i
 		}
 
 		// Importing SMP Configuration from the file
```
piyushsingariya, line 90: Just above this line token is being set twice. Please remove one instance.

piyushsingariya:
@navendu-pottekkat can you update the use-cases here with examples+explaination. Just to be on the same page while reviewing.

navendu-pottekkat:
The only additional use case here is to run tests on the same performance profile.

If a performance profile exists, run tests under it.
Else check with the user and create a new performance profile.

Earlier we used to default to a new performance profile with a profile name as same as the name provided in the test configuration file. This will be used in the [meshery-smp-action](https://github.com/layer5io/meshery-smp-action/pull/30) to run similar tests and group them under the same performance profile.

piyushsingariya (changes requested):
Documentation would require a few updates, @navendu-pottekkat please open an issue for this. "good-first-issue" can be added.

navendu-pottekkat:
@piyushsingariya Updated the inline docs as well as the docs pages by removing references of the `--profile` flag.

Also removed the redundant code for setting token path.

l5io:
🚀 Preview for commit ca4fed71ee9dd413b8f0d6762ee3be2ac1366df0 at: https://61ae01eacefb1175c10f30e9--meshery-docs-preview.netlify.app

piyushsingariya (approved):

PR #4730, opened 2021-12-06
Fixpossiblepanic

Revolyssup:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/handlers/extensions.go
+++ b/handlers/extensions.go
@@ -11,6 +11,7 @@ import (
 )
 
 var (
+	//USE WITH CAUTION: Wherever read/write is performed, use this in a thread safe way, using the global mutex
 	extendedEndpoints = make(map[string]*models.Router)
 	mx                sync.Mutex
 )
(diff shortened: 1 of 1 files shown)


leecalcote (dismissed):

handlers/extensions.go
```diff
@@ -19,6 +21,8 @@ type ExtensionVersion struct {
 }
 
 func (h *Handler) ExtensionsEndpointHandler(w http.ResponseWriter, req *http.Request, prefObj *models.Preference, user *models.User, provider models.Provider) {
+	mx.Lock()
+	defer mx.Unlock()
```
utkarsh-pro, line 25: @Revolyssup are we holding for a longer than required? I am not sure but it seems like that we won't release the lock unless the request context dies, right?

Revolyssup: I did this because the variable is still global. I can use the lock just around where the variable is accessed but if someone else tomorrow edits the code inside this function, maybe they dont keep that in mind so I did this.
If you say so, I will lock just around the line variable is read/write. And I will add the comment above the variable to use it in a thread safe way. Sound right?

leecalcote:
@utkarsh-pro how do these changes look?

utkarsh-pro:
@leecalcote this looks good. This is needed as the package that we are using is thread unsafe.

utkarsh-pro (approved):
LGTM. @leecalcote let's get this merged?

leecalcote:
Sounds great, yes.

sayantan1413:
There was a docker build failure in this PR don't know if we should be concern about that. It is a npm error.

PR #4737, opened 2021-12-07
update user preferences handler to accept JSON instead of form data

DelusionalOptimist:
**Description**

This PR adds ability to persist user preferences for extensions.

**Notes for Reviewers**
* Modifies how POST `/api/user/prefs`
* Example JSON object:
```json
{
    "loadTestPrefs": {
        "c": 1,
        "t": "200s",
        "gen": "fortio"
    },
    "anonymousUsageStats": false,
    "anonymousPerfResults": false,
    "usersExtensionPreferences": {
        "hideToolbar": true,
        "showOnZoom": false
    }
}
```

### DIFF
--- a/handlers/error.go
+++ b/handlers/error.go
@@ -79,6 +79,7 @@ const (
 	ErrCreateDirCode            = "2150"
 	ErrInvalidRequestObjectCode = "2151"
 	ErrChangeK8sContextCode     = "2152"
+	ErrSavingUserPreferenceCode = "some_code"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


leecalcote (approved):

leecalcote:
@DelusionalOptimist, @MarioArriaga92, with this change, we'll need to update the end-to-end user preferences test - https://github.com/meshery/meshery/runs/4448911173?check_suite_focus=true

Revolyssup (approved):

utkarsh-pro:
@DelusionalOptimist will there be any changes required in the UI now that we have essentially changed the expected `Content-Type`?

leecalcote:
@utkarsh-pro I suspect this is why the ui integration test is failing. // @Abhishek-kumar09

DelusionalOptimist:
Yeah UI changes would be needed. The UI probably has some values hard coded which are sent as form data. Some changes such that it can parse and send json would be needed. I have attached a sample object in the PR description.

leecalcote:
> Yeah UI changes would be needed. The UI probably has some values hard coded which are sent as form data. Some changes such that it can parse and send json would be needed. I have attached a sample object in the PR description.

@Abhishek-kumar09, tag. You're it.

leecalcote:
@utkarsh-pro there are default values for user preferences set by the Local Provider. This endpoint should be able to respond with those default values. Is this something that you can look into?

PR #4742, opened 2021-12-08
[UI] Pattern GitHub Import with DEMO

### LINKED ISSUE #4157: [UI] Import pattern files from GitHub
#### Current Behavior

Users can import pattern/application/filter files via a URL through Meshery UI.

#### Desired Behavior

Users should be able to import pattern files from GitHub as well when logged in to Meshery Remote Provider via GitHub.

The new UI should reflect the following features:

- User clicks import and is presented with 3 options
  - "browse" for filesystem import
  - "url" for generic, public, HTTP endpoint
  - "GitHub" which will contain a two-level pick list of 1. the user's orgs and 2. the repos in that chosen org. The user will be allowed to select only one repo to import at a time.

---

Chadha93:
Signed-off-by: Gaurav Chadha <gauravchadha1676@gmail.com>

**Description**
The functionality needs to be configured.

This PR fixes #4157 

References: [Docs
](https://docs.google.com/document/d/1B2N78EdRiZF-yVo1-HY3syppwBBDumgMuYg6seD-AJ4/edit#)

**Notes for Reviewers**

https://user-images.githubusercontent.com/65453826/147207529-b90d4ea6-6793-4ff8-b163-f6e564a9e658.mp4


ui/components/GHImport.js
```diff
@@ -0,0 +1,155 @@
...
+                    <h2 id="simple-modal-title">Import using Github</h2>
+                  </Grid>
+                  <div className={classes.ghGroups}>
+                    <FormControl className={classes.formControl}>
+                      <InputLabel htmlFor="grouped-native-select">Repository</InputLabel>
+                      <Select defaultValue="" id="grouped-select">
+                        <MenuItem value="">
+                          <em>None</em>
+                        </MenuItem>
+                        <MenuItem value={1}>service-mesh-patterns</MenuItem>
```
Chadha93, line 116: @utkarsh-pro Is there a way I can directly target the repository and the branch(on select), or it should be done with the function handler for GitHub import?

Chadha93: + what all other things for now I can add as per the UX in the UI ?

Meghana-12: Why are we using dropdown here? Are there limited repos that we can import the files from?
From what I understood from the call, we need to take owner, repo and branch as inputs, dir path and one check box as inputs. So like 4 input fields and one check box, right?

Chadha93: We have currently two repos that we will be using to import the patterns and filters provided by Meshery for now.
since the repos which Meshery provides can remain in the dropdown, and to take the owner repo we may use an input URL option.  yes we could replace the branch dropdown selection with the checkbox.

Meghana-12: I'm sorry I didn't get why we need to replace branch dropdown with check box. We just need to use `TextField` instead of dropdown right? 

Owner: TextField (just take the owner github id) 
Repo: Dropdown 
Branch: TextField 
Dir Path: TextField 
Recurrsive? : Checkbox
This will work, right?

Meghana-12 (changes requested):

ui/components/GHImport.js
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+}));
+import GenericModal from "./GenericModal";
+
+
+const GHImport = ({ onSubmit }) => {
+  const classes = styles();
+  const [open, setOpen] = React.useState(false);
+  const [input, setInput] = React.useState('');
```
Meghana-12, line 65: We need different states here to store `owner, repo, branch, dirPath` and a boolean `recursive` to check if recursion is needed. 
Once we have that,
You should create a new state/ just a const `finalURL` using string interpolation,
set `finalURL` as       
```
`https://github.com/${owner}/${repository}/${branch}/`
```
And another `path` as 
```
`${dirPath}/${ recursive ? '**' / ''}`
```
Something like this.

And then body will be :
```
{ 
  "url": finalURL,
  “path”: path,
  “save”: false 
}
```
This should be passed as `body` parameter `onSubmit` Function from what I understand.

ui/components/MesheryPatterns.js
```diff
@@ -430,6 +441,10 @@ function MesheryPatterns({
     handleSubmit(link, "", "meshery_" + Math.floor(trueRandom() * 100), FILE_OPS.URL_UPLOAD);
     // console.log(link, "valid");
   }
+
+  function ghImportHandler(gh) {
+    handleSubmit(gh, "",  "meshery_" + Math.floor(trueRandom() * 100), FILE_OPS.GH_IMPORT);
```
Meghana-12, line 446: And the POST request endpoint is mentioned as `/api/experimental/pattern` [here](https://docs.google.com/document/d/1B2N78EdRiZF-yVo1-HY3syppwBBDumgMuYg6seD-AJ4/edit#). So I'm not sure if you can use the same `handleSubmit` as it makes request to `/api/pattern`

Meghana-12 (changes requested):

ui/components/MesheryPatterns.js
```diff
@@ -384,14 +389,20 @@ function MesheryPatterns({
...
-    if (type === FILE_OPS.FILE_UPLOAD || type=== FILE_OPS.URL_UPLOAD) {
+    if (type === FILE_OPS.FILE_UPLOAD || type=== FILE_OPS.URL_UPLOAD || type === FILE_OPS.GH_IMPORT) {
       let body;
       if (type === FILE_OPS.FILE_UPLOAD) {
         body = JSON.stringify({  pattern_data : { pattern_file : data }, save : true })
       }
       if (type === FILE_OPS.URL_UPLOAD) {
         body = JSON.stringify({ url : data, save : true })
       }
+      if (type === FILE_OPS.GH_IMPORT) {
```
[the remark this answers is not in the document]
Meghana-12: From the construction of `body` I mentioned, we wouldn't have to build the `body` here again, the `data` parameter itself will be the `body`. No need to change anything, we can pass it directly.

ui/components/MesheryPatterns.js
Chadha93: yes we can use `/api/pattern` it's the new one, it was earlier `/api/experimental/pattern` confirmed.

ui/components/GHImport.js
Chadha93: okay, I'll update the pr accordingly, thanks.

ui/components/GHImport.js
Chadha93: yes, you are right @Meghana-12 , but I think the branch will always, remain master or main only, yes it may change when it's users' own repository(preference). I did the previous one based on repos branches provided by meshery. I'll update up in the UI

ui/components/MesheryPatterns.js
Meghana-12: Great!

ui/components/GHImport.js
utkarsh-pro: @Chadha93 a user can import patterns from any repository and any branch and I think we should give them a text box rather than a drop down.

Chadha93: Gotcha!

Chadha93 (commented):
@Meghana-12 , @utkarsh-pro I have updated the UI and fixed the suggested. 
How is recursive getting used from the endpoints that we have?
Please do suggest some more changes required for this. Thanks : )

Chadha93:
Screenshot from 2021-12-12 14-08-28

ui/components/GHImport.js
```diff
@@ -119,6 +128,12 @@ const GHImport = ({ onSubmit }) => {
                     <FormControl className={classes.formControl}>
                       <TextField id="outlined-basic" label="Path" variant="outlined"  onChange={() => setPath(`${path}/${ recursive ? '**' : ''}`)}/>
```
Meghana-12, line 129: Here, what we are trying to do is if the user wants us to recursively search then we add `**` to the directory path else we don't. And since state `check` tells us if user wants the recursive search or not, we need to use `check` here instead of `recursive`. Since you have declared recursive as `const` and had set its value as `false`, its always going to be false so the current code is not dynamic. You can make it work by using `check `: 
```onChange={() => setPath(`${path}/${ check ? '**' : ''}`)}```

File: ui/components/GHImport.js [hunk 6]
```diff
@@ -65,6 +67,13 @@ const GHImport = ({ onSubmit }) => {
   const [path, setPath] = React.useState('');
   const [repository, setRepository] = React.useState('');
   const [branch, setBranch] = React.useState('');
+  const [check, setCheck] = React.useState({
+    checked1 : true,
+  });
+  const handleCheckCall = (event) => {
+    `${ recursive ? '**' : ''}`
```
Meghana-12, line 74: This line is not needed.

Meghana-12 (changes requested):

ui/components/GHImport.js
```diff
@@ -0,0 +1,164 @@
...
+    setOpen(false);
+  };
+  const validURL = (str) => {
+    return URLValidator(str);
+  }
+  const handleError = (input) => {
+    console.log(input + ' is not valid a valid repository')
+  }
+  const handleSubmit = () => {
+    validURL(input) ? onSubmit(finalURL) : handleError(input);
```
Meghana-12, line 95: Here you need to pass : 
```
{ "url": finalURL, “path”: `${path}/${ check ? '**' : ''}`, “save”: false }
```

ui/components/MesheryPatterns.js
```diff
@@ -384,14 +389,17 @@ function MesheryPatterns({
...
+        body = JSON.stringify({ url : data, })
```
Meghana-12, line 401: As we are passing the object : 
```{ "url": finalURL, “path”: `${path}/${ check ? '**' : ''}```, “save”: false ` as parameter for the onSubmit function before, we need to construct body like this here as far as I know. You can set 
`body=data` and continue.

**Comment by leecalcote** [comment]
@Chadha93 will you resolve any comment that you consider to be addressed? I wonder if another quick recording is needed to help get this merged.

**Comment by Chadha93** [comment]
> @Chadha93 will you resolve any comment that you consider to be addressed? I wonder if another quick recording is needed to help get this merged.

@leecalcote All the above comments are addressed and implemented in this Pr, I'll resolve them. 
The video I shared on slack Is updated in the PR description, 
We can go with the new UI that we discussed in the Figma in the next version of this, 
The only thing that will help in this pr to get merged is storing all the patterns at once in the table, (currently, it's making a perfect request).
I'll discuss this more with @Abhishek-kumar09 to achieve this.

File: ui/components/GHImport.js [hunk 9]
```diff
@@ -0,0 +1,163 @@
...
+    }
+  },
+  overrides : {
+    MuiGrid : {
+      input : {
+        color : '#607d8b'
+      }
+    },
+  }
+})
```
Abhishek-kumar09, line 24: I won't recommend using it. 
you have already used makeStyles and the same can be achieved with that.

Chadha93: Oh yes, I was focussing on it's working much, Thanks for reminding, I'll clean the code styles using makeStyles before final commit.

MarioArriaga92:
@Chadha93 @Abhishek-kumar09  

 I see some failed checks about 13 days old for cypress UI end-to-end tests (its due to fixed cypress caching issue). Maybe someone with permissions should re-run the job (think it'll re-run once more commits are pushed, but not sure tbh)

leecalcote:
Re-running build jobs...

leecalcote:
Yes, a new commit to this PR/branch will have the same affect of rerunning the build jobs. 👍

warunicorn19:
@Chadha93 some conflicts and build fails to take care here.

Chadha93:
> @Chadha93 some conflicts and build fails to take care here.

@warunicorn19 This issue is on hold for some time.

leecalcote:
Rerunning...

There is a merge conflict to address, too.

leecalcote:
Actually, @Chadha93, @warunicorn19 should we close this PR considering recent work in this area?

Chadha93:
> Actually, @Chadha93, @warunicorn19 should we close this PR considering recent work in this area?

Yes, if we are not going with this UX approach and we need some more info to load/store multiple patterns with the same request from a repo.
Let's get the approach(backend+frontend) written and defined UX and get back to this again.
Sounds good?

leecalcote:
@Chadha93 sounds real good. Closing this PR until we circle back...

PR #4748, opened 2021-12-09
[UI] Added Big Number component

meghanacosmos:
**Description**
Add Big Number Component. 

Screenshot 2021-12-10 at 3 31 05 AM

**Notes for Reviewers**

### DIFF
--- a/ui/components/BigNumber/index.js
+++ b/ui/components/BigNumber/index.js
@@ -1,14 +1,7 @@
 import * as React from "react";
 import { Grid, Box, Typography, Button, CardContent, CardActions, Card } from "@mui/material/";
 
-import PropTypes from "prop-types";
-
-BigNumber.propTypes = {
-  title: PropTypes.number,
-  description: PropTypes.description,
-  buttonText: PropTypes.buttonText,
-  handleClick: PropTypes.func,
-};
+// eslint-disable-next-line react/prop-types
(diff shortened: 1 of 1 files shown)


ui/components/BigNumber/index.js
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+import { Grid, Box, Typography, Button, CardContent, CardActions, Card } from "@mui/material/";
+
+import PropTypes from "prop-types";
```
sudo-NithishKarthik, line 4: It is really good of you to use prop-types @Meghana-12 , but we have not yet settled on a solution for types. Our intention was to wait for sometime and then, in the future, move to typescript completely. So, I think its better not to waste time incorporating a different style when it is obvious that we have to ditch it.

sudo-NithishKarthik (changes requested):
Just the prop-types can be reverted. Rest looks fine. Thank you @Meghana-12 🎉

ui/components/BigNumber/index.js
Meghana-12: Oh okay! I was getting eslint errors so I added it. Will remove it then!

sudo-NithishKarthik: I see. We have to change the eslint config then.

sudo-NithishKarthik (approved):
LGTM 🎉

Issue #4731, opened 2021-12-07
[mesheryctl] System Start gives unfriendly message when K8s is not available


leecalcote:
#### Current Behavior
Can we go back to the drawing board on this:
```
$ mesheryctl system start
Error: !! cannot query the Kubernetes API. See https://docs.meshery.io/reference/error-codes
```

It is terribly unfriendly and not obvious what needs to be done.  Also, there is no error code given.

#### Desired Behavior
Provide user with a message that says what the issue is and what to do about it.

```
Your `$contextname` context is configured to run Meshery on Kubernetes using the `$tokenname` token. The Kubernetes cluster is not accessible. Please confirm that your cluster available and that the token is valid. See https://docs.meshery.io/installation/quick-start for additional instructions.
```

---

#### Contributor [Guides](https://docs.meshery.io/project/contributing) and Resources

- 🎨 Wireframes and designs for Meshery UI in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Layer5 Discussion Forum](https://discuss.layer5.io) and [Layer5 Community Slack](http://slack.layer5.io)

ibnjunaid:
@leecalcote I want to work on this issue can you assign it to me.

sayantan1413:
Yeah sure go ahead @ibnjunaid 🚀

leecalcote:
Gents, please be aware that this issue needs to move quickly. It's a poor user experience for new users. @ibnjunaid message quickly if you get stuck, please.

ibnjunaid:
@leecalcote Can you  review  the PR #4751 and please let me know if any changes are needed.

PR #4751, opened 2021-12-10
Improve error message when kubernetes cluster is not available or the token is invalid.

ibnjunaid:
**Description**
Improve error message when kubernetes cluster is not available or the token is invalid.

This PR fixes #

This PR improves the error message when kubernetes cluster is not available or the token is invalid.
It adds contextname as well as tokenname to the error message for better debugging and user experience.

**Notes for Reviewers**

- [Y ] Yes, I signed my commits.

### DIFF
--- a/mesheryctl/internal/cli/root/system/error.go
+++ b/mesheryctl/internal/cli/root/system/error.go
@@ -23,7 +23,7 @@ const (
 	ErrProcessingMctlConfigCode     = "1025"
 	ErrRestartMesheryCode           = "1026"
 	ErrK8sQueryCode                 = "1041"
-	ErrK8sConfigCode                = "1045" // please check this error code
+	ErrK8sConfigCode                = "not set"
 )
 
 func ErrHealthCheckFailed(err error) error {


navendu-pottekkat:
@ibnjunaid We should use meshkit's error logger here. There are other instances where we use this in mesheryctl.

ibnjunaid:
> @ibnjunaid We should use meshkit's error logger here. There are other instances where we use this in mesheryctl.

@navendu-pottekkat I have changed the error logger function to use meshkit's method. Can you please provide feedback

navendu-pottekkat:
Could you move the errors to [this file](https://github.com/navendu-pottekkat/meshery/blob/ab49fb36ea7f98db78949d7eb2a75973a156dfc1/mesheryctl/internal/cli/root/system/error.go#L1)?

Shouldn't the invalid token error be a different error code?

@piyushsingariya can chime in on any other changes needed.

mesheryctl/internal/cli/root/system/check.go
```diff
@@ -276,7 +277,15 @@ func (hc *HealthChecker) runKubernetesAPIHealthCheck() error {
...
-		return errors.New("ctlK8sClient1000: !! cannot initialize a Kubernetes client. See https://docs.meshery.io/reference/error-codes")
+		sdescription := []string{"The Kubernetes cluster is not accessible."}
+		ldescription := []string{fmt.Sprintf("Your %s context is configured to run Meshery on Kubernetes using the %s token", hc.mctlCfg.CurrentContext, hc.context.Token),
+			"The Kubernetes cluster is not accessible",
+			"Please confirm that your cluster is available and that the token is valid",
+			"See https://docs.meshery.io/installation/quick-start for additional instructions",
+		}
+		probablecause := []string{"Kubernetes cluster is unavailable and that the token is invalid"}
+		remedy := []string{"Please confirm that your cluster is available and that the token is valid. See https://docs.meshery.io/installation/quick-start for additional instructions"}
+		return meshkiterrors.New(ErrK8sQueryCode, meshkiterrors.Critical, sdescription, ldescription, probablecause, remedy)
```
piyushsingariya, line 288: Can you move the logic to separate function?

piyushsingariya: Also it would be much helpful if you could do this to the all errors present in file.

ibnjunaid: sure I'll move it into a separate function inside [this file ](https://github.com/navendu-pottekkat/meshery/blob/ab49fb36ea7f98db78949d7eb2a75973a156dfc1/mesheryctl/internal/cli/root/system/error.go#L1)

mesheryctl/internal/cli/root/system/check.go
```diff
@@ -303,7 +296,9 @@ func (hc *HealthChecker) runKubernetesAPIHealthCheck() error {
 			log.Warn("!! cannot query the Kubernetes API")
 			return nil
 		}
-		return ErrK8SQuery(err)
+		errMsg := fmt.Errorf("%s. Your %s context is configured to run Meshery on Kubernetes using the %s token",
```
piyushsingariya, line 299: really not sure if this line should be added in errors from health checks. It shouldn't be added I feel because it's kinda irrelevant to healthchecks.

// @navendu-pottekkat

ibnjunaid:
@piyushsingariya @navendu-pottekkat  Can you please review the PR. I have made the changes you suggested.

mesheryctl/internal/cli/root/system/check.go
ibnjunaid: Its a preflight check so I think it will help in debugging. In the [Issue](https://github.com/meshery/meshery/issues/4731) @leecalcote mentions the error message fromat to be `Your `$contextname` context is configured to run Meshery on Kubernetes using the `$tokenname` token. The Kubernetes cluster is not accessible. Please confirm that your cluster available and that the token is valid. See https://docs.meshery.io/installation/quick-start for additional instructions.`

piyushsingariya: The error is for k8s availability. better to remove the part of it that concerns the token thing? i.e. `using the$tokenname token` and `and that the token is valid.`

ibnjunaid: Yeah. I'll remove it

mesheryctl/internal/cli/root/system/error.go
```diff
@@ -23,6 +23,7 @@ const (
 	ErrProcessingMctlConfigCode     = "1025"
 	ErrRestartMesheryCode           = "1026"
 	ErrK8sQueryCode                 = "1041"
+	ErrK8sConfigCode                = "1045" // please check this error code
```
ibnjunaid, line 26: @piyushsingariya How are we calculating an error code. Do we have any convention

piyushsingariya: just set the value `"not set"` and it would be automatically taken care of. A Layer5 toolkit takes care of this during the CI workflows.

piyushsingariya (commented):
Ready to be merged after resetting the error-code value

ibnjunaid:
@piyushsingariya Can you please approve it. If its all good.

piyushsingariya (approved):
Nice work @ibnjunaid. Contributions.

PR #4786, opened 2021-12-18
Change core workflow for e2e adapter tests

Revolyssup:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/.github/workflows/testadapters.yaml
+++ b/.github/workflows/testadapters.yaml
@@ -102,8 +102,7 @@ jobs:
       # Start tunnelling to allocate external IP to LoadBalancer type service(meshery)(SO that we can access meshery from outside)
       - name: Start tunelling
         run: minikube tunnel &> /dev/null &
-      - name: wait 10 sec
-        run: sleep 10s
+
 
       # Checking meshery server logs
       - name: Check Meshery pod logs
@@ -135,15 +134,12 @@ jobs:
         run: |
             cat ./${{ inputs.patternfile_name }}
             mesheryctl pattern apply -f ./${{ inputs.patternfile_name }}
-      # Wait for sometime to get some logs
-      - name: Sleep for 10 sec
-        run: sleep 10s
-   
+
       - name: Check adapter logs
         run: |
             podname=$(kubectl get pods -n meshery | grep ${{ inputs.adapter_name }} | tr -s ' ' | cut -d " " -f 1)
             kubectl logs pod/$podname -n meshery
-      - name: Check meshery logs
+      - name: Check Meshery logs
         run: |
           podName=$(kubectl get pods -o wide -n meshery | sed -n 2p | tr -s ' ' | cut -d " " -f 1)
           kubectl logs pod/$podName -n meshery


.github/workflows/testadapters.yaml
```diff
@@ -66,13 +82,12 @@ jobs:
...
             cat ~/.meshery/auth.json
-
         
       # Start meshery server (Assuming that this step will create the meshery namespace)
       - name: Start meshery server using mesheryctl
         run: |
-           curl -L https://git.io/meshery | PLATFORM=kubernetes bash -
-           sleep 20s
+              curl -L https://git.io/meshery | PLATFORM=kubernetes bash -
+              sleep 20s
```
leecalcote, line 90: No sleeping.

leecalcote (changes requested):

.github/workflows/testadapters.yaml
```diff
@@ -48,7 +63,8 @@ jobs:
         run: |
           eval $(minikube -p minikube docker-env) 
           docker build  -t test:test .
-
+      - name: What is kube config
```
leecalcote, line 66: kubeconfig is one word. 

It would be nice if this said Retrieve kubeconfig

.github/workflows/testadapters.yaml
Revolyssup: I will remove this one. But some of the sleeping is required at least 30 second total

.github/workflows/testadapters.yaml
```diff
@@ -48,7 +63,8 @@ jobs:
...
+        run:  cat ~/.kube/config     
       # Setup meshery config.yaml and auth.json
       - name: Setup meshery config and infinite token for use
```
leecalcote, line 53: Always capitalize Meshery

.github/workflows/testadapters.yaml
```diff
@@ -66,13 +82,11 @@ jobs:
               echo '{ "meshery-provider": "None", "token": "" }'  > ~/.meshery/auth.json
             fi
             cat ~/.meshery/auth.json
-
         
       # Start meshery server (Assuming that this step will create the meshery namespace)
       - name: Start meshery server using mesheryctl
```
leecalcote, line 72: Capitalize Meshery please. Capitalize Meshery component names, too.

.github/workflows/testadapters.yaml
```diff
@@ -22,6 +22,18 @@ on:
       patternfile_name:
         required: true
         type: string   
+      k8s_version:  
+        required: true
+        type: string   
+      sm_version:
```
leecalcote, line 28: How do we guarantee the the pattern file and these separate version inputs will be the same?

I assume the answer is that we are not. And that the individual configuring this workflow has to account for that.

.github/workflows/testadapters.yaml
```diff
@@ -105,7 +118,6 @@ jobs:
               if [ "$svcstatus" = "<pending>" ];then
```
leecalcote, line 105: @piyushsingariya where is mesheryctl system dashboard in this?

.github/workflows/testadapters.yaml
Revolyssup: In the workflow that is used to call this one. Automatically the latest service mesh version is set ,both in patternfile as well as in parameter. No manual stuff

.github/workflows/testadapters.yaml
Revolyssup: This service status is the information about whether an external IP has been allocated or not by minikube tunnel. I dont think mesheryctl can get us the external IP of a service

.github/workflows/testadapters.yaml
leecalcote: That's great. This leads me to wonder why these parameters are being specified twice then.

.github/workflows/testadapters.yaml
```diff
@@ -66,13 +82,11 @@ jobs:
               echo '{ "meshery-provider": "None", "token": "" }'  > ~/.meshery/auth.json
             fi
             cat ~/.meshery/auth.json
-
         
-      # Start meshery server (Assuming that this step will create the meshery namespace)
-      - name: Start meshery server using mesheryctl
+      # Start Meshery server (Assuming that this step will create the meshery namespace)
+      - name: Start Meshery server using mesheryctl
```
leecalcote, line 87: Meshery Server

.github/workflows/testadapters.yaml
leecalcote: It should be. @hexxdump, will you chime in here?

.github/workflows/testadapters.yaml
```diff
@@ -92,11 +106,10 @@ jobs:
         run: sleep 10s
```
leecalcote, line 92: What can be done other than sleeping? 😴 is not good.

.github/workflows/testadapters.yaml
```diff
@@ -105,26 +118,23 @@ jobs:
...
           name: patternfile
       # This pattern will try to install  service mesh
       - name: Pattern apply (Install service mesh)
         run: |
+            cat ./${{ inputs.patternfile_name }}
             mesheryctl pattern apply -f ./${{ inputs.patternfile_name }}
-
       # Wait for sometime to get some logs
       - name: Sleep for 30 sec
         run: sleep 30s
```
leecalcote, line 130: Ugh

.github/workflows/testadapters.yaml
Revolyssup: We have to pass the sm version so that we can reflect that onto the test results. If not for specifying twice, the other way would have been to extract directly from patternfile, which will not be a good idea considering we want this to be as general as possible.

.github/workflows/testadapters.yaml
Revolyssup: Least I can do is reduce them to 10 seconds. I want to make sure, I get all the logs properly and it takes time for all the logs to come up.

.github/workflows/testadapters.yaml
piyushsingariya: It can get external IP and PORT if minikube tunnel is active in other terminal.

.github/workflows/testadapters.yaml
leecalcote: The problem with sleeping isn't time itself, but know how much time to sleep. Yes, it's inefficient and not ideal. The real issue is that this approach inevitably leads to bugs.  

find something else to poll; something else to check to determine when to move forward instead of sleeping.

leecalcote:
Were you able to identify another method other than sleep?

Revolyssup:
> Were you able to identify another method other than sleep?

I will remove all sleeps

leecalcote:
> > Were you able to identify another method other than sleep?
> 
> 
> 
> I will remove all sleeps

Removing the sleeps without inserting some other logic to wait on a dependency is also likely to cause bugs.

Revolyssup:
> > > Were you able to identify another method other than sleep?
> > 
> > 
> > I will remove all sleeps
> 
> Removing the sleeps without inserting some other logic to wait on a dependency is also likely to cause bugs.

No it shouldn't, I thought about it. At the end where we actually need to wait, I am polling continually for 150 seconds

Revolyssup:
> > > > Were you able to identify another method other than sleep?
> > > 
> > > 
> > > I will remove all sleeps
> > 
> > 
> > Removing the sleeps without inserting some other logic to wait on a dependency is also likely to cause bugs.
> 
> No it shouldn't, I thought about it. At the end where we actually need to wait, I am polling continually for 150 seconds
The only other reason to wait was to wait for logs, but I think we get enough logs almost instantaneously when we do Pattern apply.

leecalcote (approved):
Lgtm

PR #4798, opened 2021-12-20
Default to provided adapter without prompting the user

pottekkat:
Signed-off-by: Navendu Pottekkat <navendupottekkat@gmail.com>

**Description**

When running `mesheryctl mesh deploy --adapter meshery-linkerd --namespace linkerd-ns LINKERD` we assume `mesheryctl` to use the provided adapter (`meshery-linkerd`) to deploy the provided service mesh (`LINKERD`) without having to ask any more questions to the user.

But this is not the case for the adapter before as shown below. It still shows a prompt even if there is only a single adapter discovered.

```
➜  ✗ mesheryctl mesh deploy --adapter meshery-linkerd --namespace linkerd-ns LINKERD  
Verifying prerequisites...
Use the arrow keys to navigate: ↓ ↑ → ← 
? Select an Adapter from the list: 
  ▸ meshery-linkerd:10001
```

This has been fixed by introducing a simple check before prompting the user.

This also allows users to script around `mesheryctl` without having to account for the prompt.

This is also propagated to the `mesh remove` command.

**Notes for Reviewers**

A new release of Meshery need to be made to use this change in the GitHub workflows.

### DIFF
--- a/mesheryctl/internal/cli/root/mesh/deploy.go
+++ b/mesheryctl/internal/cli/root/mesh/deploy.go
@@ -206,6 +206,11 @@ func validateAdapter(mctlCfg *config.MesheryCtlConfig, tokenPath string, name st
 		return ErrNoAdapters
 	}
 
+	if len(adapterNames) == 1 {
+		adapterURL = adapterNames[0]
+		return nil
+	}
+
 	prompt := promptui.Select{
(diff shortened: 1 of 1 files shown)


piyushsingariya (approved):
LGTM

PR #4809, opened 2021-12-22
Fix incorrect endpoint in system status

Revolyssup:

**Description**

This PR fixes #

**Notes for Reviewers**


leecalcote (approved):

navendu-pottekkat:
@Revolyssup What was the issue here? It should be the case that both the old line and the line you changed should return the same output. If that is not the case, the central struct we are referring to is not the source of truth. Which shouldn't be the case.

PR #4822, opened 2021-12-23
[fix/ci]: cypress binary unavailable

Abhishek-kumar09:

**Description**
A test PR to fix github ci problem with cypress tests.
This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,7 @@ jobs:
           !~/cache/exclude
           **/node_modules
           /home/runner/.cache/Cypress
-        key: ${{ runner.os }}-node-${{ hashFiles(''**/lockfiles'') }}
+        key: ${{ runner.os }}-node-${{ hashFiles('**/lockfiles') }}
         restore-keys: |
           ${{ runner.os }}-node-
     - name: Install dependencies
@@ -131,7 +131,7 @@ jobs:
           !~/cache/exclude
           **/node_modules
           /home/runner/.cache
-        key: ${{ runner.os }}-node-${{ hashFiles(''**/lockfiles'') }}
+        key: ${{ runner.os }}-node-${{ hashFiles('**/lockfiles') }}
     - name: Run tests
       run: |
         cd ui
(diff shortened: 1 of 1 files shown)


MarioArriaga92:
Looking at this right now @leecalcote @Abhishek-kumar09, what's the actual issue this is fixing?

Abhishek-kumar09:
Issue: 
`Cypress is installed but binary is not available`

MarioArriaga92:
Ohh, this one @Abhishek-kumar09, right?
Screen Shot 2021-12-23 at 13 06 04
https://github.com/meshery/meshery/runs/4619150486?check_suite_focus=true

MarioArriaga92:
@Abhishek-kumar09 if one pays close attention to the output, cypress version is stated to be '9.0.0' but expected binary location is for '8.1.0'

Abhishek-kumar09:
yep,
Looks like it is fixed. Cleaning up the garbage.
image

MarioArriaga92:
@Abhishek-kumar09 I just don't get why we need to run 'npx cypress install', this should be done by 'npm i' internally in prior steps

Abhishek-kumar09:
That was caching error. Adding one more step to install cypress and cache the location.
Adding more steps doesn't add build minutes since  if cypress is already installed and cached, then the installing and unzipping operations won't be performed, if not installed then the binaries will be installed and cached.

MarioArriaga92:
Especially because we are not doing a production build, so devDependencies should be installed as any npm hooks should run as well
Screen Shot 2021-12-23 at 13 14 20

MarioArriaga92:
@Abhishek-kumar09 sounds good let me know when code clean up is done and I'll take a look, thanks!

MarioArriaga92:
It' just seems weird this started happening for no apparent reason, it'd be interesting to find if there's any underlying root cause.

Abhishek-kumar09:
Yeah, it was caching error at it's root.
We were caching node modules but not cypress binary path. 
Adding `/runner/.cache/Cypress` helped.
Then latest cypress was installed which is `9.2.0` and the code is executed with `8.x.x`.
I don't know why it downloaded latest binaries of cypress, although it is provided that it has to install `8.x.x` cypress version in package.json.

Issue #3743, opened 2021-08-02
[Meshery Server] use `localhost` in Makefile for adapter urls


Abhishek-kumar09:
#### Description

On starting meshery server from `make run-fast`,
Currently all the default urls for adapters in Meshery is pointed to `mesherylocal.layer5.io`. I think when we run adapters outside the meshery server and they are available through `localhost` only.  

This was originally suggested by @utkarsh-pro on [discuss page](https://discuss.layer5.io/t/service-mesh-adapters-are-not-found-on-running-meshery-server-locally/151).

The adapter connection is unsuccessful when going with the legacy makefile while trying to adapter ports on `mesherylocal.layer5.io`.

#### Environment

- Host OS: Linux
- Meshery Server Version: latest
- Meshery Client Version: latest
- Platform: Docker

---

SirZeck:
Is this issue open? Can I work on it and can you guide me to where /etc/hosts file is?

PR #4871, opened 2022-01-01
shift to localhost ref for adapters

leecalcote:
\*\*Description\*\*

This PR fixes \#3743

\*\*Notes for Reviewers\*\*
Old versions of docs left as-is.


Makefile
```diff
@@ -1,6 +1,6 @@
-ADAPTER_URLS := "mesherylocal.layer5.io:10000 mesherylocal.layer5.io:10001 mesherylocal.layer5.io:10002 mesherylocal.layer5.io:10003 mesherylocal.layer5.io:10004 mesherylocal.layer5.io:10005 mesherylocal.layer5.io:10006 mesherylocal.layer5.io:10007 mesherylocal.layer5.io:10008 mesherylocal.layer5.io:10009 meshery-nginx-sm:10010"
+ADAPTER_URLS := "localhost:10000 localhost:10001 localhost:10002 localhost:10003 localhost:10004 localhost:10005 localhost:10006 localhost:10007 localhost:10008 localhost:10009 meshery-nginx-sm:10010"
 
-MESHERY_CLOUD_LOCAL="http://mesherylocal.layer5.io:9876"
+MESHERY_CLOUD_LOCAL="http://localhost:9876"
```
warunicorn19, line 3: is this a intended change?

l5io:
🚀 Preview for commit e6cb663ac2e838de804aad5dd30b95f655ea6b10 at: https://61d0db4bbd572384262dae28--meshery-docs-preview.netlify.app

Makefile
leecalcote: Yes, if you've locally built Meshery Cloud, its reference can be localhost... right? I don't think I'm missing something here, but that said, I can't remember why we had mesherylocal in the first place.

leecalcote: Oh, I see that you might be thrown off by the PR title (that was auto-created by VS Code since this is fixing an issue with the same title).

warunicorn19: ohh yes okay, I get it now.

PR #4887, opened 2022-01-04
[mesheryctl] Chaining token functions for authentication

### LINKED ISSUE #4805: mesheryctl system config aks: Error getting context: could not read token:: open /Users/me/.meshery/auth.json: no such file or directory
mesheryctl  `v0.6.0-rc-2`

Following this guide: https://docs.meshery.io/installation/platforms/aks

```
13:34 $ mesheryctl system config aks
Missing Meshery config file.
Create default config now [y/n]? y
Default config file created at /Users/me/.meshery/config.yaml
{
  "azure-cli": "2.24.1",
  "azure-cli-core": "2.24.1",
  "azure-cli-telemetry": "1.0.6",
  "extensions": {
    "azure-batch-cli-extensions": "6.0.0",
    "storage-preview": "0.7.2"
  }
}
Configuring Meshery to access AKS...
Please enter the Azure resource group name:
me-rg
Please enter the AKS cluster name:
me-test-001-aks
Merged “me-test-001-aks" as current context in /Users/me/.meshery/kubeconfig.yaml
Error getting context: could not read token:: open /Users/me/.meshery/auth.json: no such file or directory
```

hexxdump:
**Description**
`mesheryctl system config ...` command throws an error message, when incorrect/no token file is provided.
Ideally, the config command should be run only after mesheryctl is authenticated to meshery server using `mesheryctl system login` command.

This PR fixes #4805 

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/internal/cli/root/app/view_test.go
+++ b/mesheryctl/internal/cli/root/app/view_test.go
@@ -7,7 +7,6 @@ import (
 	"testing"
 
 	"github.com/jarcoal/httpmock"
-	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/constants"
 	"github.com/layer5io/meshery/mesheryctl/pkg/utils"
 )
 
@@ -93,7 +92,7 @@ func TestAppView(t *testing.T) {
 			apiResponse := utils.NewGoldenFile(t, tt.Fixture, fixturesDir).Load()
 
 			// set token
-			constants.TokenFlag = tt.Token
+			utils.TokenFlag = tt.Token
 
 			// mock response
 			httpmock.RegisterResponder("GET", tt.URL,
--- a/mesheryctl/pkg/utils/helpers.go
+++ b/mesheryctl/pkg/utils/helpers.go
@@ -124,6 +124,8 @@ var (
 	KubeConfig string
 	// KeepNamespace indicates if the namespace should be kept when Meshery is uninstalled
 	KeepNamespace bool
+	// TokenFlag sets token location passed by user with --token
+	TokenFlag = "Not Set"
 )
 
 var CfgFile string
--- a/mesheryctl/internal/cli/root/app/app.go
+++ b/mesheryctl/internal/cli/root/app/app.go
@@ -3,7 +3,6 @@ package app
 import (
 	"fmt"
 
-	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/constants"
 	"github.com/layer5io/meshery/mesheryctl/pkg/utils"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -29,7 +28,7 @@ var AppCmd = &cobra.Command{
 }
 
 func init() {
-	AppCmd.PersistentFlags().StringVarP(&constants.TokenFlag, "token", "t", "", "Path to token file default from current context")
+	AppCmd.PersistentFlags().StringVarP(&utils.TokenFlag, "token", "t", "", "Path to token file default from current context")
 
 	availableSubcommands = []*cobra.Command{onboardCmd, viewCmd, offboardCmd, listCmd}
 	AppCmd.AddCommand(availableSubcommands...)
--- a/mesheryctl/internal/cli/root/app/list.go
+++ b/mesheryctl/internal/cli/root/app/list.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/config"
-	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/constants"
 	"github.com/layer5io/meshery/mesheryctl/pkg/utils"
 	"github.com/layer5io/meshery/models"
 	"github.com/pkg/errors"
@@ -56,7 +55,7 @@ var listCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		tokenObj, err := utils.ReadToken(constants.TokenFlag)
+		tokenObj, err := utils.ReadToken(utils.TokenFlag)
 		if err != nil {
 			return err
 		}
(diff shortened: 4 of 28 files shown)


piyushsingariya:
@hexxdump we can skip that prereqcheck for now, and think back at that when the refactoring of mesheryctl comes into action.

hexxdump:
> @hexxdump we can skip that prereqcheck for now, and roll back to that when the refactoring of mesheryctl comes into action.

sure @piyushsingariya. We can go ahead with these changes for now.

leecalcote:
Can we make sure that we have centralized onto the same token check mechanism and same user-facing message? Do we have need for more than one token validation check? Here's a sample from `perf`:

```
➜  mesheryctl git:(master) ./mesheryctl perf profile sam
Error: authentication token not found. please supply a valid user token with the --token (or -t) flag. or login with `mesheryctl system login`.
See https://docs.meshery.io/reference/mesheryctl/perf/profile for usage details
```

This message used here is different than that proposed in this PR. What's the plan to getting to a single function for token validation?

piyushsingariya:
> Can we make sure that we have centralized onto the same token check mechanism and same user-facing message? Do we have need for more than one token validation check? Here's a sample from `perf`:
> 
> ```
> ➜  mesheryctl git:(master) ./mesheryctl perf profile sam
> Error: authentication token not found. please supply a valid user token with the --token (or -t) flag. or login with `mesheryctl system login`.
> See https://docs.meshery.io/reference/mesheryctl/perf/profile for usage details
> ```
> 
> This message used here is different than that proposed in this PR. What's the plan to getting to a single function for token validation?

This error isn't from a token function, it's from the part where the request made to meshery-server fails.

leecalcote:
@piyushsingariya thanks for confirming. Will the change in this PR elicit the same error message from the server? Can we make sure that we have centralized onto the same token check mechanism and same user-facing message? Do we have need for more than one token validation check?

piyushsingariya:
> @piyushsingariya thanks for confirming. Will the change in this PR elicit the same error message from the server? Can we make sure that we have centralized onto the same token check mechanism and same user-facing message? Do we have need for more than one token validation check?

@leecalcote and @hexxdump I'm taking over this PR, and implementing an idea that will chain all the token functions and will decrease repetitive code all over the place(I guess).

piyushsingariya:
```
➜  mesheryctl git:(master) ./mesheryctl perf profile sam
Error: authentication token not found. please supply a valid user token with the --token (or -t) flag. or login with `mesheryctl system login`.
See https://docs.meshery.io/reference/mesheryctl/perf/profile for usage details
```

@leecalcote now this error is centralized and will be thrown at all places.

and damn, what have I done to this PR.

@hexxdump This PR is now one step closer to centralizing and chaining mesheryctl functions.

piyushsingariya:
@hexxdump time to review your own PR :joy:

mesheryctl/pkg/utils/helpers.go
```diff
@@ -827,3 +806,15 @@ func SetOverrideValues(ctx *config.Context, mesheryImageVersion string) map[stri
...
+// CheckFileExists checks if the given file exists in system or not
+func CheckFileExists(name string) (bool, error) {
+	_, err := os.Stat(name)
+	if err == nil {
+		return true, nil
+	}
+	if errors.Is(err, os.ErrNotExist) {
+		return false, fmt.Errorf("%s does not exist", name)
+	}
+	return false, errors.Wrap(err, fmt.Sprintf("%s does not exist", name))
```
hexxdump, line 819: should this error statement be something generic, as the ErrNotExist case is already handled in line:817
For e.g. Failed to read/fetch the file?

mesheryctl/internal/cli/root/app/view_test.go
```diff
@@ -92,7 +93,7 @@ func TestAppView(t *testing.T) {
 			apiResponse := utils.NewGoldenFile(t, tt.Fixture, fixturesDir).Load()
 
 			// set token
-			tokenPath = tt.Token
+			constants.TokenFlag = tt.Token
```
hexxdump, line 96: Does it make sense to add this flag under a different package or import the constant package with some other alias? This sounds like we are modifying the constants.

piyushsingariya: So the new logic of creating a http request does two things
1. search Token passed by `--token` that has been assigned to var constants.TokenFlag
2. If found ""; i.e. no token was passed with flag, location is fetched from meshconfig
3. Checks if the file exists
4. Try adding the file to request

returns the http request.

During unit tests, we mock server response but token needs to be attached to due to the requirement of the logic, so we save a sample token but it doesn't exists at the location defined in the meshconfig, so two things could be done here; pass the token location with "--token" flag or set the value of `constants.TokenFlag` manually, and passing the flag does the same things so chose wrote it manually.

mesheryctl/pkg/utils/helpers.go
piyushsingariya: Good point, will make the changes tomorrow.

mesheryctl/internal/cli/root/app/view_test.go
hexxdump: @piyushsingariya, I understand the reasons behind, manually storing tt.Token value in constants.TokenFlag
The variables/flags defined in constants.go must be used for read only purpose. 
Ideally, any flag/field which can be modified later, should belong to a different package. 
May be the TokenFlag field should move to either internal/cli/root/config package or pkg/utils package.

piyushsingariya: makes sense, moved to utils package

hexxdump (commented):
LGTM

piyushsingariya (approved):
Let's hope this doesn't break anything. I tried one subcommand under each command to test if the token flag is working correctly.

piyushsingariya:
jk, I have tested the commands myself.

hexxdump:
> jk, I have tested the commands myself.

CITY @piyushsingariya :-)

PR #4941, opened 2022-01-13
[Fix] Remove Redundant NATS Connections in Subscriptions

humblenginr:
**Description**

This PR removes the call to `connectToBroker` function on dataPlane and controlPlane subscriptions. At the startup, when the UI connects to the server, it creates multiple subscriptions including controlPlane, dataPlane and operatorStatus subscriptions. In the current code, we have kept these subscriptions as a trigger to initiate a new connection with NATS which is flawed since this is now creating three connections whereas only one is required.

By removing the function call that initiates the connection to NATS, we now have only a single connection to NATS that's being reused by the server. 

Ideally, we would want to move this process of initiating the NATS connection to the startup of Meshery Server, which would mitigate a lot of issues that we currently have with managing go channels. This is just a temporary fix to minimise the impact of this issue. 

**Notes for Reviewers**

### DIFF
--- a/internal/graphql/resolver/control-plane.go
+++ b/internal/graphql/resolver/control-plane.go
@@ -33,11 +33,6 @@ func (r *Resolver) listenToControlPlaneState(ctx context.Context, provider model
 
 	go func() {
 		r.Log.Info("Initializing ControlPlane subscription")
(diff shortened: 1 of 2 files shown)


leecalcote (approved):
LGTM

Issue #3919, opened 2021-08-18
[UI] Update the user on "Token" expire


Chadha93:
#### Current Behavior
When tokens expire, we show things like “Unexpected error occurred” on the /performance page

#### Desired Behavior
The user needs to be informed that their token is expired and to log in again.

#### Screenshots/Logs
<img width="2520" alt="Screen Shot 2021-08-17 at 4 27 59 PM" src="https://user-images.githubusercontent.com/65453826/129847614-094dd5d2-7616-4413-95d1-591aefb14b4c.png">

#### Environment

- Browser: Chrome Safari Firefox
- Host OS: Mac Linux Windows
- Meshery Server Version: stable-v
- Meshery Client Version: stable-v
- Platform: Docker Kubernetes

---

#### Contributor [Guides](https://docs.meshery.io/project/contributing) and Resources

- 🎨 Wireframes and designs for Meshery UI in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Layer5 Discussion Forum](https://discuss.layer5.io) and [Layer5 Community Slack](http://slack.layer5.io)

yagarwal1307:
if no one's working on this one I can pick it up.

sayantan1413:
Yeah sure @yagarwal1307 go ahead!

yagarwal1307:
I am stuck on this issue because I could not reproduce the error!

warunicorn19:
hey @yagarwal1307 any updates on this?

yagarwal1307:
@warunicorn19 no i couldn't reproduce this error on my system. So I have no idea regarding this issue

Chadha93:
> Hi @Chadha93 and @leecalcote I would like to take this up.

@RajGM are you able to reproduce this issue? What will be your approach for this issue?

RajGM:
Hi @Chadha93 
So far I am unable to reproduce the issue but I am working on that.
There can be multiple ways to solve this but I am assuming that with each request we are sending the token to server with each request (private routes) and then on server side we are verifying the token.
So once the token expires while the user is on the /performance page I can redirect him to login page.
On the login page we can have a logic which check if the user is being redirected from performance page, if yes then after successful login we can redirect him to performance page.

RajGM:
[...]

PR #4956, opened 2022-01-18
Fix capture and handle expired token 

### LINKED ISSUE #3935: [UI] Capture and Handle Expired Token
#### Current Behavior
Users with expired tokens will see the following on the performance management page:

Screen Shot 2021-08-18 at 6 49 50 PM

#### Desired Behavior
Irrespective of the page, users should universally see the same message - a modal informing the user that they their session has expired and that they will be directed to login in 3... 2... 1..

---

MUzairS15:
Signed-off-by: MUzairS15 <muzair.shaikh810@gmail.com>

**Description**

This PR fixes #3935 

**Notes for Reviewers**
This PR also addresses #3919 

https://user-images.githubusercontent.com/79860504/151032929-79e5b5c3-94c9-4d6f-a7d5-ff90eac44b1e.mov

### DIFF
--- a/models/remote_provider.go
+++ b/models/remote_provider.go
@@ -374,14 +374,14 @@ func (l *RemoteProvider) Logout(w http.ResponseWriter, req *http.Request) {
 // Redirects to alert user of expired sesion
 func (l *RemoteProvider) HandleUnAuthenticated(w http.ResponseWriter, req *http.Request) {
 	_, err := req.Cookie("meshery-provider")
-	if (err == nil) {
+	if err == nil {
 		ck, err := req.Cookie(tokenName)
 		if err == nil {
 			ck.MaxAge = -1
 			ck.Path = "/"
 			http.SetCookie(w, ck)
 		}
-		
+
 		http.Redirect(w, req, "/auth/login", http.StatusFound)
 		return
 	}


MUzairS15:
I am not able to see the full coverage report.
Can anyone help me to increase coverage?

ui/components/MesheryPerformance/PerformanceProfiles.js
```diff
@@ -156,19 +208,27 @@ function PerformanceProfile({ updateProgress, enqueueSnackbar, closeSnackbar })
 
   function handleError(msg) {
     return function (error) {
-      updateProgress({ showProgress : false });
+      if (error.source.errors[0].message != undefined) {
```
bariqhibat, line 211: Can we do a type check here? And additionally use optional chaining

bariqhibat (commented):
A comment! Thanks for the contribution! @MUzairS15

leecalcote:
[Grover] "Cookie Monster, come out and play."

leecalcote (changes requested):
@MUzairS15, thanks for this PR! 

This PR was discussed on today's Meshery meeting (we missed you ;) ). A couple comments were given such as "Can the cookie be used to determine session state / token expiry?", "How is this going to be applied to each UI / endpoint? It could be easy to miss an endpoint.", "Can this be achieved in NextJS with one central configuration?", "The golang middleware attached to each endpoint verifies token expiration and can redirect the user. This would be a central approach, too."

MUzairS15:
>  "The golang middleware attached to each endpoint verifies token expiration and can redirect the user. This would be a central approach, too."

Current implementation redirects user to login but it fails. 
Screenshot 2022-01-20 at 8 53 01 AM
Similar to #3919 
Even if, we happen to solve this, without any alert user will be redirected.

> "Can the cookie be used to determine session state/token expiry?"

In the UI itself? Yes, we can do this.I thought about this but later resolved to rely on the server as in `AuthMiddleware` we were doing the same, 

> "How is this going to be applied to each UI / endpoint? It could be easy to miss an endpoint."

In data-fetch.js ` const dataFetch = (url, options = {}, successFn, errorFn) => {`, we can handle the error here itself, (this will secure all the endpoint making a query),

Pls correct me if I am wrong.

MUzairS15:
@leecalcote Can you suggest, should I handle error or add check for whether token is still valid or not in `const dataFetch = (url, options = {}, successFn, errorFn) => {`

(Now, I regret not being able to join the meeting🥲, things have been clear)

Abhishek-kumar09:
> Current implementation redirects user to login but it fails.

Are you referring to the current implementation (on master branch)?
It seems like the login request can't be sent directly from the server to the cloud, just by clients like mesheryctl or UI which is the required behaviour. Youy should try for UI redirection to login page, instead. Is it what you are doing now??

Abhishek-kumar09:
> t, should I handle error or add check for whether token is still valid or not in const dataFetch = (url, options = {}, successFn, errorFn) => {

Token validation can only be done on the server/cloud. Rather you can check for the cookie key (token) be present on all requests that includes credential(also called cookies).

MUzairS15:
> > Current implementation redirects user to login but it fails.
> 
> Are you referring to the current implementation (on master branch)? It seems like the login request can't be sent directly from the server to the cloud, just by clients like mesheryctl or UI which is the required behaviour. Youy should try for UI redirection to login page, instead. Is it what you are doing now??

Yes, on master branch (server code, redirect fails, see screenshot above). And I have added logic to redirect from UI.
Based on https://github.com/meshery/meshery/pull/4956#pullrequestreview-857294128, concern is to apply to each endpoint. So, if I instead of checking for errors in individual query call, what if I move this to `dataFetch`?

MUzairS15:
> > t, should I handle error or add check for whether token is still valid or not in const dataFetch = (url, options = {}, successFn, errorFn) => {
> 
> Token validation can only be done on the server/cloud. Rather you can check for the cookie key (token) be present on all requests that includes credential(also called cookies).

When token expires, cookie is not cleared, so cookie (token) will be present.

Abhishek-kumar09:
> Based on #4956 (review), concern is to apply to each endpoint. So, if I instead of checking for errors in individual query call, what if I move this to dataFetch?

Yeah that's what you should do, intercept every request that passes cookie to server. Do it in data fetch itself.

> When token expires, cookie is not cleared, so cookie (token) will be present.

That's a good thought, but token is set to session expire policy, i.e, it will not expire frequently and only in two conditions: 
1. The user closes the browser (session is lost and so is the token)
2. Token has timeout for 1 day or probably more (@leecalcote could confirm). 

SO the case is extremely rare, meanwhile in case it is not correct (i.e., the auth token expired) the server will automatically redirect to the provider UI.

MUzairS15:
> SO the case is extremely rare, meanwhile in case it is not correct (i.e., the auth token expired) the server will automatically redirect to the provider UI.

I tried with an expired token but was not redirected to provider UI.

Abhishek-kumar09:
Okay let's have a check at two levels: 
1. the client level for auth token present which is just one liner.
2. the server for auth token expiry, and redirection to the provider UI instead of handling the login request itself.

What do you say @MUzairS15 ? What are the things needed for the 2 to be done, if you have any idea.

MUzairS15:
> Okay let's have a check at two levels:
> 
> 1. the client level for auth token present which is just one liner.
> 2. the server for auth token expiry, and redirection to the provider UI instead of handling the login request itself.
> 
> What do you say @MUzairS15 ? What are the things needed for the 2 to be done, if you have any idea.

In server `AuthMiddleware` checks for session validity, I can use the error to handle the redirect. But, then it will not be graceful handling 
> Irrespective of the page, users should universally see the same message - a modal informing the user that they their session has expired and that they will be directed to login in 3... 2... 1...

handlers/middlewares.go
```diff
@@ -56,7 +65,26 @@ func (h *Handler) AuthMiddleware(next http.Handler) http.Handler {
 			// }
 			// return
 			if provider.GetProviderType() == models.RemoteProviderType {
-				provider.Logout(w, req)
+				if err.Error() == "http: named cookie not present" {
+					provider.Logout(w, req)
+				}
+
+				graphQlError := []Body{{"expired"}}
```
Abhishek-kumar09, line 72: What is this doing?

handlers/middlewares.go
```diff
@@ -56,7 +65,26 @@ func (h *Handler) AuthMiddleware(next http.Handler) http.Handler {
 			// }
 			// return
 			if provider.GetProviderType() == models.RemoteProviderType {
-				provider.Logout(w, req)
+				if err.Error() == "http: named cookie not present" {
```
Abhishek-kumar09, line 68: use contains function for this. https://pkg.go.dev/strings#Contains

handlers/middlewares.go
```diff
@@ -67,13 +95,14 @@ func (h *Handler) AuthMiddleware(next http.Handler) http.Handler {
...
+func (h *Handler) validateAuth(provider models.Provider, req *http.Request) (bool, error) {
+	err := provider.GetSession(req)
+	if err == nil {
 		// logrus.Debugf("session: %v", sess)
-		return true
+		return true, err
 	}
 	// logrus.Errorf("session invalid, error: %v", err)
-	return false
+	return false, err
```
Abhishek-kumar09, line 105: You can delete the unusable comments like seen here.

ui/components/MesheryPerformance/PerformanceProfiles.js
```diff
@@ -299,9 +359,25 @@ function PerformanceProfile({ updateProgress, enqueueSnackbar, closeSnackbar })
...
+            <Typography id="transition-modal-title" variant="h6" component="h2" className={classes.modal}>
+              <span className = {classes.span}>&#9432;</span>Session Expired
+            </Typography>
+            { counter && <Typography id="transition-modal-description" className = {classes.modalText} >
+               Your session has expired.
+               You will now be redirected to login... {counter}
+            </Typography>
+            }
+          </Box>
+        } />
```
Abhishek-kumar09, line 380: Please extract this dialog and it's style in a separate component. They have been used twice on different pages.

ui/components/MesheryPerformance/Dashboard.js
```diff
@@ -140,17 +184,25 @@ function Dashboard({ updateProgress, enqueueSnackbar, closeSnackbar, grafana, ro
...
-      enqueueSnackbar(`${msg}: ${error}`, { variant : "error",
-        action : function Action(key) {
-          return (
-            <IconButton key="close" aria-label="Close" color="inherit" onClick={() => closeSnackbar(key)}>
-              <CloseIcon />
-            </IconButton>
-          );
-        },
-        autoHideDuration : 8000, });
+      if (error.source.errors[0].message != undefined) {
```
Abhishek-kumar09, line 187:
```suggestion
      if (error?.source?.errors?.[0]?.message) {
```

ui/components/MesheryPerformance/Dashboard.js
```diff
@@ -82,6 +120,12 @@ function Dashboard({ updateProgress, enqueueSnackbar, closeSnackbar, grafana, ro
     fetchTests();
   }, []);
 
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      setCounter(counter - 1)
+    }, 1000)
+    return () => clearTimeout(timer)
+  })
```
Abhishek-kumar09, line 128: The useEffect Body runs on every state change which is not desired. Actually this modal can be placed in it's own functional component for the purpose of re-usability and then this implementation will look good.

handlers/middlewares.go
```diff
@@ -56,7 +65,26 @@ func (h *Handler) AuthMiddleware(next http.Handler) http.Handler {
 			// }
 			// return
 			if provider.GetProviderType() == models.RemoteProviderType {
-				provider.Logout(w, req)
+				if err.Error() == "http: named cookie not present" {
+					provider.Logout(w, req)
```
Abhishek-kumar09, line 69: Earlier we had the problem for wildcard response set, 

Are we solving it now?

Abhishek-kumar09 (changes requested):
Thanks for the PR @MUzairS15 
So what you have done on the server is:
1. check for token cookie being present 
2. If not present redirect to provider UI 
3. else marshal the cookie token and authenticate the user.

Just one concern, WHy you have added graphql in between? Is there any specific usecase?
I think this is not touching any graphql domain. Do we?

handlers/middlewares.go
MUzairS15: No, we are not touching graphQL, (just for sake of the name `graphQlError`) I will change the name.
I am sending `{
          "errors": {[
                "error" : "expired"
            ]}

}
` as json to get it when we do a query when token becomes invalid.

handlers/middlewares.go
MUzairS15: Sure

handlers/middlewares.go
MUzairS15: Yes, will remove.

handlers/middlewares.go
MUzairS15: Yes, this is solved with the proposed changes, as we will not redirect from the server in case of an invalid session, instead query will fail, and UI will handle it by alerting the user.

handlers/middlewares.go
Abhishek-kumar09: I think giving more verbose error will help, like `auth token expired`

MUzairS15: done

handlers/middlewares.go
MUzairS15: done

ui/components/MesheryPerformance/PerformanceProfiles.js
MUzairS15: done

ui/components/MesheryPerformance/Dashboard.js
MUzairS15: done

ui/components/MesheryPerformance/Dashboard.js
MUzairS15: done

handlers/middlewares.go
```diff
@@ -56,7 +66,26 @@ func (h *Handler) AuthMiddleware(next http.Handler) http.Handler {
...
+					return
+				}
+				authError := []Body{{"auth-token expired"}}
+				errors := Error{authError}
+				jsonMsg, err := json.Marshal(errors)
+				if err != nil {
+					logrus.Errorf("Error marshaling response")
+					http.Error(w, "unable to authenticate user", http.StatusInternalServerError)
+					return
+				}
```
Abhishek-kumar09, line 80:
```suggestion
```

Abhishek-kumar09 (approved):
After 30 mins long iteration, we got the root of the problem. 

SO the `/user/login` is redirecting to the meshery cloud for the user login, but since the redirection is from the server the cloud takes server on localhost:9081 as client and further failing because server can't act as a client and this is the reason behind the cors error. 

Cross origin request from the server is denied on the cloud for the Login and that's how it should behave. You can't **instantiate** a login from the server. 


Now since we are redirecting to the provider UI now, the actual selection of login can be done through client. 

Cookie Policy: 
1. The cookie expiration policy is set to session, i.e., the cookie invalidates on the new session.
2. It has its own expiry time more than 24 hours in case the session is not closed,

We are touching both the cases and all the cases the user can open meshery and cannot use any meshery service without being authenticated. 

There can be a need of slight UX improvemnt in terms of the first load with invalid token sent to meshery and being validated, but that's a seperate concern and is out of the scope of this PR. 

I think this PR can be merged, and you can see the changes with your eyes, the all functions being working the way they should. 

UI improvements (minor flash between meshery homepage to going to provider login page) can be done seprately.

Abhishek-kumar09:
// @leecalcote

leecalcote:
@MUzairS15, it'd be great for you to quick a quick demo of this work on a Meshery Development Meeting. There's one today in 30 minutes, if you're free. Details here: https://meet.layer5.io

leecalcote:
@sudo-NithishKarthik what do you think? https://github.com/meshery/meshery/pull/4956#pullrequestreview-864057301

MUzairS15:
@sudo-NithishKarthik ping

leecalcote:
@MUzairS15 will you put this on the agenda for next week's Meshery Dev meeting?

MUzairS15:
Yes @leecalcote

leecalcote:
Discussed again today. The current server-side redirect is good and the return of the user-facing modal is also desirable.

handlers/provider_handler.go
```diff
@@ -24,7 +23,6 @@ func (h *Handler) ProviderHandler(w http.ResponseWriter, r *http.Request) {
 			http.SetCookie(w, &http.Cookie{
 				Name:     h.config.ProviderCookieName,
 				Value:    p.Name(),
-				Expires:  time.Now().Add(h.config.ProviderCookieDuration),
```
MUzairS15, line 27: Changes `meshery-provider` cookie as Session cookie.

leecalcote: ✔️

PR #4961, opened 2022-01-19
changed dev server port

warunicorn19:
Signed-off-by: Aditya Chatterjee <speak2adi@gmail.com>

**Description**
Changed dev server port from 3000 to 4001. 
If 4001 is occupied port no. increases by one. 
This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/ui/ui_dev_server.js
+++ b/ui/ui_dev_server.js
@@ -4,7 +4,7 @@ const { parse } = require("url");
 const next = require("next");
 const detect = require('detect-port');
 
-var port = parseInt(process.env.PORT, 10) || 4000
+var port = parseInt(process.env.PORT, 10) || 4001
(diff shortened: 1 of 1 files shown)


sayantan1413 (commented):
The changes are ok and also do the same change in the provider UI also. But just a question is it necessary to make this change in this branch because I found that it will cause issues in the UI tests and also in some GitHub workflows as well as in some other places where the port 3000 is hardcoded and will break in some places. So I think if we are trying to implement this method we should also find some way to use this dynamic approach where the port 3000 is hardcoded.

warunicorn19:
Aah now I remember why we didnt change it here before, it's probably not worth the hassle. I'll close the PR. Thanks for the review @sayantan1413

PR #4972, opened 2022-01-21
[Refactor] Improve the way Meshery Server handles Meshsync data

humblenginr:
**Description**

This PR introduces some changes in the startup behaviour of Meshery Server. From now on, whenever Meshery Server is started, it would 

1. Dump the stale MeshSync data in the database 
2. Ask Meshsync to send the data in its informer store and persist them in the database 
3. Subscribe to the events sent by MeshSync and configure what should happen when an event is received, which at the moment is to update the database using the information from the event.

**Notes for Reviewers**

I haven't pushed the changes to the GraphQL subscriptions because, the subscriptions that we have now are very rigid, that is, it is difficult to change something in one subscription without affecting the other, so we have to change this in a separate PR. 

In this PR, the case when the operator is not deployed or not functional when Meshery Server starts is not being handled, it is to be done in a separate PR.


cmd/main.go
```diff
@@ -257,6 +261,45 @@ func main() {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
 
+	go func() {
+		kclient, err := mesherykube.New(nil)
```
utkarsh-pro, line 265: Pardon me, I am a bit rusty but are we assuming here that meshery is deployed within kubernetes?

cmd/main.go
```diff
@@ -257,6 +261,45 @@ func main() {
...
+		if err != nil {
+			log.Error(err)
+		}
+		mesheryclient, err := operatorClient.New(&kclient.RestConfig)
```
utkarsh-pro, line 269: If `err == true` will `kclient == nil`? In that case this line may panic?

cmd/main.go
```diff
@@ -257,6 +261,45 @@ func main() {
...
+		if err != nil {
+			if mesheryclient == nil {
+				log.Error(model.ErrMesheryClient(nil))
+			}
+			log.Error(model.ErrMesheryClient(err))
+		}
+		broker, err := mesheryclient.CoreV1Alpha1().Brokers(MesheryNamespace).Get(context.Background(), "meshery-broker", metav1.GetOptions{})
```
utkarsh-pro, line 276: `context.Background()` would be almost pointless I think? Let's either propagate from parent or let's use `context.TODO()`?

cmd/main.go
```diff
@@ -257,6 +261,45 @@ func main() {
...
+		kclient, err := mesherykube.New(nil)
+		if err != nil {
+			log.Error(err)
+		}
+		mesheryclient, err := operatorClient.New(&kclient.RestConfig)
+		if err != nil {
+			if mesheryclient == nil {
+				log.Error(model.ErrMesheryClient(nil))
+			}
+			log.Error(model.ErrMesheryClient(err))
```
utkarsh-pro, line 274: Do we want to proceed ahead despite the errors?

cmd/main.go
```diff
@@ -257,6 +261,45 @@ func main() {
...
+		brkrConn, err := nats.New(nats.Options{
+			URLS:           []string{brokerEndpoint},
+			ConnectionName: MesherySeverBrokerConnection,
+			Username:       "",
+			Password:       "",
+			ReconnectWait:  2 * time.Second,
+			MaxReconnect:   60,
+		})
+		if err != nil {
+			log.Error(err)
```
utkarsh-pro, line 293: I think we should return...

cmd/main.go
sudo-NithishKarthik: Yep. Makes sense 👍

cmd/main.go
sudo-NithishKarthik: Thanks for pointing that out @utkarsh-pro . This implementation is not something that I intend to keep as is. There are also some other things apart from what you said such as ensuring that the operator is deployed at the startup (which also ensures the availability of MeshSync and broker) that we should consider. But, I haven't started working on those yet. So, I thought, let's introduce this change now, then put in a PR addressing these issues in a more organised manner.

cmd/main.go
sudo-NithishKarthik: ahh.. this was not inside a goroutine before, cuz I wanted it to be synchronous, that's why there are no return statements. But then I decided to put them in a goroutine but failed to add return statements 😅 . Thanks for pointing these @utkarsh-pro , will make the necessary changes.

models/meshsync_events.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+}
+
+func (mh *MeshsyncDataHandler) Run() error {
+	storeSubscriptionStatusChan := make(chan bool)
+
+	go mh.subsribeToStoreUpdates(storeSubscriptionStatusChan)
+	go mh.subscribeToMeshsyncEvents()
+	err := mh.removeStaleObjects()
+	if err != nil {
```
utkarsh-pro, line 39: If there is an error here then we will listen on the `storeSubscriptionStatusChan` causing the entire `subsribeToStoreUpdates` routine to be blocked.

cmd/main.go
sudo-NithishKarthik: @utkarsh-pro , the constructor for the Kubernetes client (`mesherykube.New`) takes care of handling in-cluster vs out of cluster cases. It returns the client accordingly. For reference - meshkit [client.go](https://github.com/layer5io/meshkit/blob/42abdc0df310af02ec4f2bc380964a9fba4e4880/utils/kubernetes/client.go#L28)

bariqhibat (commented):
Is this the one that we talked during the call few days ago?

sudo-NithishKarthik:
Yes @bariqhibat

models/error.go
```diff
@@ -104,6 +105,10 @@ var (
 	ErrIndexOutOfRange      = errors.New(ErrIndexOutOfRangeCode, errors.Alert, []string{"Error: index out of range"}, []string{}, []string{}, []string{})
 )
 
+func ErrBrokerNotFound(err error) error {
+	return errors.New(ErrBrokerNotFoundCode, errors.Alert, []string{"Meshery broker not found"}, []string{"Unable to find meshery broker in the cluster", err.Error()}, []string{"Invalid Grafana Endpoint or API-Key"}, []string{"Update your Grafana URL and API-Key from the settings page in the UI"})
```
leecalcote, line 109:
```suggestion
	return errors.New(ErrBrokerNotFoundCode, errors.Alert, []string{"Meshery Broker not found"}, []string{"Unable to find Meshery Broker in the cluster", err.Error()}, []string{"Dysfunctional Meshery Broker and/or Meshery Operator deployment"}, []string{"Verify that 'brokers.meshery' custom resource is present. Confirm healthy deployment of Meshery Operator."})
```

cmd/main.go
```diff
@@ -257,6 +261,51 @@ func main() {
...
+			return
+		}
+		log.Info("Connected to broker at: ", brokerEndpoint)
+		meshsyncDataHandler := helpers.NewMeshsyncDataHandler(brkrConn, dbHandler, log)
+		err = meshsyncDataHandler.Run()
+		if err != nil {
+			log.Info(err.Error())
+		}
+
```
leecalcote, line 306:
```suggestion
```

leecalcote:
@sudo-NithishKarthik there are a couple merge conflicts to look at.

leecalcote:
Syncing with `master`

leecalcote:
Pending an update to support multiple Kubernetes contexts.

leecalcote:
Refreshing with `master` again.

PR #4979, opened 2022-01-24
Reduced code duplication in handling API calls

MUzairS15:
**Description**
Reduced code duplication in handling failed API calls

**Notes for Reviewers**
Dependent PR for #4956 

### DIFF
--- a/ui/components/MeshAdapterConfigComponent.js
+++ b/ui/components/MeshAdapterConfigComponent.js
@@ -67,7 +67,6 @@ class MeshAdapterConfigComponent extends React.Component {
   }
 
   fetchAvailableAdapters = () => {
-    // const self = this;
     this.props.updateProgress({ showProgress : true });
     dataFetch(
       "/api/system/adapters",
--- a/ui/components/MesherySMIResults.js
+++ b/ui/components/MesherySMIResults.js
@@ -61,24 +61,6 @@ class MesherySMIResults extends Component {
       ))
     }
 
-    // handleError = (error) => {
-    //   this.props.updateProgress({ showProgress : false });
-    //   // console.log(`error fetching results: ${error}`);
-    //   const self = this;
-    //   this.props.enqueueSnackbar(`There was an error fetching results: ${error}`, { variant : 'error',
-    //     action : (key) => (
-    //       <IconButton
-    //         key="close"
-    //         aria-label="Close"
-    //         color="inherit"
-    //         onClick={() => self.props.closeSnackbar(key)}
-    //       >
-    //         <CloseIcon />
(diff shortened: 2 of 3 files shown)


MUzairS15:
@leecalcote , @Abhishek-kumar09 , @sudo-NithishKarthik 
Pls review

bariqhibat (commented):
I don't think this is a better approach, but it's good that we centralize the error handling. Instead, it'd be good if we centralize the API calls and handle error there

Abhishek-kumar09 (commented):
This is not looking good, considering the fact that there has been increased number of functional arguments and hence less readable code. HOF can be a choice, you can also use a custom hook comprised of the `useReducer` and `useContext` hooks. 

I think we can hold off this for some time. I will be providing you more information soon.

MUzairS15:
Okay, then for #4956, to check for server response should I add check logic in every error handler or wait for https://github.com/meshery/meshery/pull/4979#pullrequestreview-860952725.

@bariqhibat @Abhishek-kumar09

Abhishek-kumar09:
> Okay, then for #4956, to check for server response should I add check logic in every error handler or wait for #4979 (review).

Add it at one or two places per say it will act as a place for the experimentation of the behaviour. we can takeover from there after that.

MUzairS15:
Ok, have already added for performance page, will do the requested changes.
 let's keep it for the Performance page only, will work?

leecalcote:
heads-up: @sudo-NithishKarthik is out sick for a couple days.

leecalcote:
@MUzairS15 a couple of merge conflicts have arisen.

MUzairS15:
> @MUzairS15 a couple of merge conflicts have arisen.

Yes, will fix those, but since a different approach is considered and as told in 
> I think we can hold off this for some time. I will be providing you more information soon.

Will update this branch when finally refactoring starts.

sudo-NithishKarthik:
Let us not spend too much in here now. We have all this set up properly in our [ui-refactor branch](https://github.com/meshery/meshery/tree/meshery-ui-restructuring/ui). All the API calls for each feature are put in [helpers.js](https://github.com/meshery/meshery/blob/meshery-ui-restructuring/ui/features/mesheryEnvironment/helpers.js).  
It is not worth our time to try and make the current code better (which is going to be a huge task) as we are just going to switch to the new ui later.

sudo-NithishKarthik:
It is really good that you have taken this initiative @MUzairS15 , but to ensure that your efforts don't go waste, let's not think about improving it further. Once add the changes that are necessary for #4956 to get through. I would encourage you to look into the ui-refactoring branch and explore ways to improve it.

PR #4984, opened 2022-01-25
Fix: Visualizer crashes on first try | CU-1x74e8a

manav1403:
Signed-off-by: manav1403 <manavavni@gmail.com>

**Description**

The visualizer was crashing with an error of invalid route as logged in the console

**Notes for Reviewers**
The condition of error was arising because the routes `extendedEndpoints` were being called before they were initialized by the `LoadExtensionFromPackage` hence there was a race condition, to tackle that I shifted the Mutex lock to the starting of the function.

### DIFF
--- a/handlers/extensions.go
+++ b/handlers/extensions.go
@@ -35,6 +35,7 @@ func (h *Handler) ExtensionsEndpointHandler(w http.ResponseWriter, req *http.Req
 
 func (h *Handler) LoadExtensionFromPackage(w http.ResponseWriter, req *http.Request, provider models.Provider) error {
 	mx.Lock()
+	defer mx.Unlock()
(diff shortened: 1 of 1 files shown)


l5io:
Task linked: [CU-1x74e8a Visualizer crashes on first try](https://app.clickup.com/t/1x74e8a)

warunicorn19 (dismissed):
LGTM! @Revolyssup good to go right?

Revolyssup (changes requested):
@manav1403  instead of doing it again and again Can you defer the unlock of mutex just after we lock it

Revolyssup (approved):

leecalcote:
Nice work.

PR #4989, opened 2022-01-25
[mesheryctl] Check for expired tokens/expired token error message

### LINKED ISSUE #4604: [mesheryctl] UX: Need more informative inactive token message

#### Current Behavior
Users of `mesheryctl perf profile`, who don't have an active token receive the following message:

```
➜  ~ mesheryctl perf profile
Error: failed to make a request.Get "/": stopped after 10 redirects.
See https://docs.meshery.io/reference/mesheryctl/perf/profile for usage details
```

This behavior is true of a number of other commands. Each of the `app`, `pattern`, `mesh`, perf`.... should all apply the same logic and UX to the user - https://github.com/meshery/meshery/blob/2bf19d36cfd924c9c2e07fbb8686c975689309c5/mesheryctl/internal/cli/root/system/check.go#L489

#### Desired Behavior
Users should be informed of their inactive token and informed to run `mesheryctl system login`.

#### Acceptance Tests
- [ ] 1. All `perf` commands apply this same check and UX.
- [ ] 2. All `mesh` commands apply this same check and UX.
- [ ] 3. All `app` commands apply this same check and UX.
- [ ] 4. All `pattern` commands apply this same check and UX.

---

#### Contributor Resources

cpepper96:
Added error message for expired token and check for expired tokens.

**Description**

This PR related to #4604 

**Notes for Reviewers**
Added a new error message for an expired token. I was able to check if the token was expired by checking for 302 status code as described [here](https://github.com/meshery/meshery/issues/4604#issuecomment-992450811). At the moment the check and UX have only been implemented for the perf commands. I wanted to go ahead and submit these changes and then create new issues/pr's for the mesh/app/pattern commands as well. The new output for the commands, when supplied with an expired token, is below.

perf

apply

result

### DIFF
--- a/mesheryctl/internal/cli/root/perf/profile_test.go
+++ b/mesheryctl/internal/cli/root/perf/profile_test.go
@@ -95,8 +95,8 @@ func TestProfileCmd(t *testing.T) {
 			{Method: "GET", URL: profileURL, Response: profile1005, ResponseCode: 200},
 		}, profile1010output, testToken, true},
 		{"failing add authentication test", []string{"profile"}, []utils.MockURL{}, profile1011output, testToken + "invalid-path", true},
-		{"Server Error 500", []string{"profile"}, []utils.MockURL{
-			{Method: "GET", URL: profileURL, Response: profile1006, ResponseCode: 500},
+		{"Server Error 400", []string{"profile"}, []utils.MockURL{
+			{Method: "GET", URL: profileURL, Response: profile1006, ResponseCode: 400},
 		}, profile1009output, testToken, true},
 	}
 
--- a/mesheryctl/internal/cli/root/perf/result_test.go
+++ b/mesheryctl/internal/cli/root/perf/result_test.go
@@ -79,9 +79,9 @@ func TestResultCmd(t *testing.T) {
 			{Method: "GET", URL: resultURL, Response: result1005, ResponseCode: 200},
 		}, result1010output, testToken, true},
 		{"failing add authentication test", []string{"result", "abhishek"}, []utils.MockURL{}, result1011output, testToken + "invalid-path", true},
-		{"Server Error 500", []string{"result", "abhishek"}, []utils.MockURL{
+		{"Server Error 400", []string{"result", "abhishek"}, []utils.MockURL{
 			{Method: "GET", URL: profileURL, Response: result1000, ResponseCode: 200},
-			{Method: "GET", URL: resultURL, Response: result1006, ResponseCode: 500},
+			{Method: "GET", URL: resultURL, Response: result1006, ResponseCode: 400},
 		}, result1009output, testToken, true},
 		{"No profile passed", []string{"result"}, []utils.MockURL{}, result1012output, testToken, true},
 	}
--- a/mesheryctl/internal/cli/root/perf/testdata/apply/1002.golden
+++ b/mesheryctl/internal/cli/root/perf/testdata/apply/1002.golden
@@ -1,2 +1,1 @@
-failed to run test.
-See https://docs.meshery.io/reference/mesheryctl/perf/apply for usage details
+Response Status Code 400, possible Server Error
--- a/mesheryctl/internal/cli/root/perf/testdata/profile/1009.golden
+++ b/mesheryctl/internal/cli/root/perf/testdata/profile/1009.golden
@@ -1,2 +1,1 @@
-Response Status Code 500, possible Server Error.
-See https://docs.meshery.io/reference/mesheryctl/perf/profile for usage details
+Response Status Code 400, possible Server Error
--- a/mesheryctl/internal/cli/root/perf/testdata/profile/1011.golden
+++ b/mesheryctl/internal/cli/root/perf/testdata/profile/1011.golden
@@ -1,1 +1,1 @@
-authentication token not found. please supply a valid user token with the --token (or -t) flag. or login with `mesheryctl system login`
+Authentication token not found. please supply a valid user token with the --token (or -t) flag. or login with `mesheryctl system login`
--- a/mesheryctl/internal/cli/root/perf/testdata/result/1009.golden
+++ b/mesheryctl/internal/cli/root/perf/testdata/result/1009.golden
@@ -1,2 +1,1 @@
-Response Status Code 500, possible Server Error.
-See https://docs.meshery.io/reference/mesheryctl/perf/result for usage details
+Response Status Code 400, possible Server Error
--- a/mesheryctl/internal/cli/root/perf/testdata/result/1011.golden
+++ b/mesheryctl/internal/cli/root/perf/testdata/result/1011.golden
@@ -1,1 +1,1 @@
-authentication token not found. please supply a valid user token with the --token (or -t) flag. or login with `mesheryctl system login`
+Authentication token not found. please supply a valid user token with the --token (or -t) flag. or login with `mesheryctl system login`


warunicorn19:
Hey @cpepper96 could we'd love it if you would demo this on this meshery dev call. 😅

cpepper96:
@warunicorn19 sure. it's tomorrow right?

warunicorn19:
Depending on timezone, yes tomorrow 😅. Wednesday

leecalcote:
Great presentation today, @cpepper96. 💪

piyushsingariya (commented):
@cpepper96 I have an idea to centralize this whole thing in one go. In utils package there's functions `NewRequest` which was recently added and is specifically designed for making requests for meshery server, however the function only returns a `*http.Request`, what we can do here is that the function now performs a request inside itself and return the response of the request instead of  `*http.Request` that way we won't have to write these Error functions in each package and these response checks could inside the function itself.

we won't need to add these `if` checks at all the places.

How does this sounds??

mesheryctl/internal/cli/root/perf/apply.go
```diff
@@ -342,6 +347,12 @@ func createPerformanceProfile(client *http.Client, mctlCfg *config.MesheryCtlCon
 	}
 
 	resp, err := client.Do(req)
+
+	// failsafe for having an expired token
+	if resp.StatusCode == 302 {
+		return "", "", ErrExpired()
+	}
+
```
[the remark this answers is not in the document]
cpepper96: honestly not sure how they would be different. I was just following along with the other return statements in that function. All the other return statements in that function return "", "", error [for example](https://github.com/meshery/meshery/blob/620ff1ef940af18cbfd846e5fe20c68bffb7b169/mesheryctl/internal/cli/root/perf/apply.go#L351).

cpepper96:
> @cpepper96 I have an idea to centralize this whole thing in one go. In utils package there's functions `NewRequest` which was recently added and is specifically designed for making requests for meshery server, however the function only returns a `*http.Request`, what we can do here is that the function now performs a request inside itself and return the response of the request instead of `*http.Request` that way we won't have to write these errors and `if` checks functions in each package/file and these checks could inside the function itself.
> 
> we won't need to add these `if` checks at all the places.
> 
> How does this sounds??

Piyush I am unfamiliar with this function but I can look into it today. Sounds like a good idea though!

mesheryctl/internal/cli/root/perf/apply.go
```diff
@@ -240,6 +240,11 @@ mesheryctl perf apply local-perf --url https://192.168.1.15/productpage --mesh i
 		log.Info("Initiating Performance test ...")
 
 		resp, err := client.Do(req)
+
+		// failsafe for having an expired token
```
navendu-pottekkat, line 244: We could add a comment here mentioning that expired tokens results in a redirect and it will keep redirecting since we are in a CLI client. So, we can check for the response code for redirect to check if the token has expired.

This comment could be helpful for future developers.

navendu-pottekkat:
@piyushsingariya handling it centrally makes sense as we will be using this throughout mesheryctl. Should we propagate these changes now across all commands and then later move to a central handling? Or should we handle it centrally now?

piyushsingariya:
> @piyushsingariya handling it centrally makes sense as we will be using this throughout mesheryctl. Should we propagate these changes now across all commands and then later move to a central handling? Or should we handle it centrally now?

handling it now would be much easier I feel, the function `NewRequest` would include all these `if` checks and will itself perform the request and return the response, only minor clean-ups would be required in each of the files

navendu-pottekkat:
@cpepper96 Would you consider this change?

@piyushsingariya can point you to the utility function.

piyushsingariya:
@navendu-pottekkat @cpepper96 
https://github.com/meshery/meshery/blob/ba8d97759de0be1e5fdd59c84ce25116dd84b3f0/mesheryctl/pkg/utils/auth.go#L30

warunicorn19:
@cpepper96 any updates to share here?

cpepper96:
@warunicorn19 @navendu-pottekkat @piyushsingariya sorry I missed that thread from a few days ago but yes I'm currently implementing those checks into the NewRequest function. At the moment I'm waiting for my new token to expire to try out the changes 😂

navendu-pottekkat:
@cpepper96 You could modify the token (the one you get from the Meshery provider) by editing the `auth.json` file. That should make the token invalid and you can test it without having to wait for it to expire.

piyushsingariya:
> @cpepper96 You could modify the token (the one you get from the Meshery provider) by editing the `auth.json` file. That should make the token invalid and you can test it without having to wait for it to expire.

he has to wait, modifying would make the token invalid not expired. The goal of the issue is to take care of expired tokens.

navendu-pottekkat:
> > @cpepper96 You could modify the token (the one you get from the Meshery provider) by editing the `auth.json` file. That should make the token invalid and you can test it without having to wait for it to expire.
> 
> he has to wait, modifying would make the token invalid not expired. The goal of the issue is to take care of expired tokens.

Wouldn't invalid tokens show the same error?

piyushsingariya:
> Wouldn't invalid tokens show the same error?

They'll show similar error, for testing the code he'll need to wait for it.

cpepper96:
So I'm running into an interesting error. When I try to get the response from the [client.Do function](https://github.com/meshery/meshery/blob/8b3b94d96cc12ba297ae72b0727032e08ba90837/mesheryctl/internal/cli/root/perf/profile.go#L137) using an invalid token there are times when I try to access the status code using resp.StatusCode that it throws a nil pointer exception like this
panic
The output I expect is something like this
failed
For reference this is my code in the auth.go file (I'm just printing out the status code to reference for myself)
auth

Is this a problem with just my local environment? Does the client.Do function always return a valid statuscode? I want to check the statuscode prior to this [check](https://github.com/meshery/meshery/blob/8b3b94d96cc12ba297ae72b0727032e08ba90837/mesheryctl/internal/cli/root/perf/profile.go#L138-L140) because otherwise an invalid token would just generate a failed request error which isn't that specific.

piyushsingariya:
@cpepper96 try logrus instead of `fmt` there must be an error from client.Do which I don't see being printed after getting panic because of nil reference, that error would help you better understanding the cause behind this.

cpepper96:
I was able to centralize all those errors into the NewRequest function in the auth.go file. I'm going to try to hop on the meshery dev call tomorrow!

warunicorn19:
Sounds great @cpepper96

mesheryctl/pkg/utils/auth.go
```diff
@@ -58,6 +60,28 @@ func NewRequest(method string, url string, body io.Reader) (*http.Request, error
...
+	if ContentTypeIsHTML(resp) {
+		return req, ErrUnauthenticated()
+	}
+
+	// failsafe for bad api call
+	if resp.StatusCode != 200 {
+		return req, ErrFailReqStatus(resp.StatusCode)
+	}
+
 	return req, nil
```
piyushsingariya, line 85:
```suggestion
	return resp, nil
```

Now that we executed request inside the function itself, let;s return the response back, instead of returning request.

cpepper96: if NewRequest returns a http response now, then how should we handle something like [this](https://github.com/meshery/meshery/blob/048517fa89c3028fd922af6abea5af2a0cb7a2d1/mesheryctl/internal/cli/root/perf/apply.go#L217-L222) where we need a request returned?

cpepper96: Nevermind I see, http.NewRequest function should work correct?

cpepper96: actually there are a few different instances where a request is expected from utils.NewRequest. Is it safe to say that anywhere that request is made http.NewRequest can be called instead?

piyushsingariya: yes, all those places will need a little cleaning, the function will directly returns the response fetched, and then that response can be used further and also closed with `defer` after getting the work done.

piyushsingariya: Nice @cpepper96 I completely forgot about such instances. I think we should leave it here then, now that the function makes sure token isn't expired, and then we make an actual request sounds also okay to me. We can go one step ahead and make this function more complicated but that just sounds like over-work to me. 

but now that we are making a "pre-request" to confirm our token condition, then in cases like `/profileID/run` it is going to trigger a performance test without those query parameters attached.

CC @navendu-pottekkat @hexxdump @leecalcote

piyushsingariya: Maybe we should make this "pre-request" to an API which doesn't change or trigger anything on backend but can confirm our token condition.

The API endpoint `/api/user/token` sounds good to me, It returns the token as file but also needs an attached token in the request. This endpoint looks good to me that can act as "pre-request" to confirm our token condition.

piyushsingariya: the attempt to centralize these expired and invalid token checks has lead us in the hole of "pre-request" and making an request after that again.

We need a little discussion on this again, I'm out of ideas rn, I'll think about what can this be handled better soon.

cpepper96:
@piyushsingariya @navendu-pottekkat @warunicorn19 any feedback?

mesheryctl/pkg/utils/auth.go
```diff
@@ -61,6 +62,37 @@ func NewRequest(method string, url string, body io.Reader) (*http.Request, error
...
+	resp, err := client.Do(req)
+
+	if err != nil && resp == nil {
+		return resp, ErrFailRequest(err)
+	}
+
+	// If statuscode = 302, then we either have an expired or invalid token
+	// We return the response and correct error message
+	if resp.StatusCode == 302 {
+		return resp, InvalidToken()
```
piyushsingariya, line 80:
```suggestion
		return nil, InvalidToken()
```

let's only return `nil, error` in error cases, to not make confusion for the folks who'll be new to golang and the codebase

mesheryctl/pkg/utils/error.go
```diff
@@ -2,11 +2,14 @@ package utils
...
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/layer5io/meshkit/errors"
 )
 
 var (
+	ErrFailRequestCode     = "1034"
+	ErrFailReqStatusCode   = "1035"
```
piyushsingariya, line 12:
```suggestion
	ErrFailRequestCode     = "not set"
	ErrFailReqStatusCode   = "not set"
```

workflows will take care of it and assign a code to it.

piyushsingariya:
also @cpepper96 unit tests are failing, try to hunt why's that happening.

cpepper96:
@piyushsingariya I went ahead and made your suggested changes. I believe the reason why the unit tests are failing is because the system is expecting a different error message from the apply command. [Here](https://github.com/meshery/meshery/blob/9bb5d308e5428e841d20a6ffe1252efff8f8f6cc/mesheryctl/internal/cli/root/perf/apply.go#L246-L251) we are expecting a FailTestRun error from these two error cases within apply however the [profile ](https://github.com/meshery/meshery/blob/9bb5d308e5428e841d20a6ffe1252efff8f8f6cc/mesheryctl/internal/cli/root/perf/profile.go#L141-L148) and [result](https://github.com/meshery/meshery/blob/9bb5d308e5428e841d20a6ffe1252efff8f8f6cc/mesheryctl/internal/cli/root/perf/result.go#L166-L173) expect the ErrUnauthenticated and ErrFailReqStatus errors. The last two error messages are what I have implemented in the MakeRequest function.

piyushsingariya:
Cool, I'll update the tests in this PR in a few hours.

piyushsingariya (approved):

PR #5021, opened 2022-02-01
[mesheryctl] Make `stop` wait for namespace to be deleted 

### LINKED ISSUE #4951: [mesheryctl] `system restart` colliding with helm delete operation

#### Current Behavior
Meshery CLI users that execute `mesheryctl system restart` while using in-cluster deployments find that `system restart` fails to start Meshery due to a helm uninstall operation ongoing (initiated by the `system stop` included in `system restart`).

#### Desired Behavior
Mesheryctl to wait for `stop` to complete, then proceed with `start`.

#### Screenshot
```
 mesheryctl system restart
Restarting Meshery...
Meshery deployments will be deleted from your cluster. Are you sure you want to continue [y/n]? y
Stopping Meshery...
Deleting Meshery Namespace...
Meshery is stopped.
Starting Meshery...
⡿ Deploying Meshery on Kubernetes Error: cannot start Meshery: create: failed to create: secrets "sh.helm.release.v1.meshery.v1" is forbidden: unable to create new content in namespace meshery because it is being terminated
Usage:
  mesheryctl system restart [flags]

Flags:
  -h, --help          help for restart
      --skip-update   (optional) skip checking for new Meshery's container images.

Global Flags:
      --config string    path to config file (default "/Users/lee/.meshery/config.yaml")
  -c, --context string   (optional) temporarily change the current context.
  -v, --verbose          verbose output
  -y, --yes              (optional) assume yes for user interactive prompts.
```

---

#### Contributor Resources

gr455:
**Description**
With stop operation, mesheryctl now polls kubernetes client waiting for meshery namespace to be deleted (times out at 300s with a message). This also fixes `restart` operation conflicting with namespace delete operation.

This PR fixes #4951

**Notes for Reviewers**
To test the changes
1. Run `make` inside `mesheryctl/` directory to generate the binary
2. Run `./mesheryctl system start` from within `mesheryctl/` directory and let the deployment complete
3. Now run `./mesheryctl system stop` and wait for the cli to exit
4. Verify that `kubectl get namespace` does not show the meshery namespace
5. Run `./mesheryctl system restart` also to make sure the restart happens without any errors.

### DIFF
--- a/mesheryctl/pkg/utils/healthcheck.go
+++ b/mesheryctl/pkg/utils/healthcheck.go
@@ -243,12 +243,12 @@ func AreAllPodsRunning() (bool, error) {
 
 // CheckMesheryNsDelete waits for Meshery namespace to be deleted, returns (done, error)
 func CheckMesheryNsDelete() (bool, error) {
-	kubeClient, err := meshkitkube.New([]byte(""))
+	client, err := meshkitkube.New([]byte(""))
 	if err != nil {
 		return false, err
 	}
 
-	if err := WaitForNamespaceDeleted(kubeClient, MesheryNamespace, 300); err != nil {
+	if err := WaitForNamespaceDeleted(client, MesheryNamespace, 300); err != nil {
 		return false, err
 	}
 


gr455:
It was suggested in the last development meeting that we cancel the `restart` operation if the `stop` operation times out without the deployments stopping.

gr455:
Hey @piyushsingariya @leecalcote, mind having a look at this? Thanks :D

mesheryctl/internal/cli/root/system/stop.go
```diff
@@ -168,8 +168,17 @@ func stop() error {
...
-
-	log.Info("Meshery is stopped.")
+	// Wait for meshery to stop
+	done, err := mesheryStopCheck()
+	if err != nil {
+		log.Info(err)
+	}
+	// If stop is not done until timeout, display info
+	if !done {
+		log.Info("Meshery has not stopped yet.\nPlease check the status of the pods by executing “mesheryctl system status” and Meshery-UI endpoint with “mesheryctl system dashboard”.")
```
hexxdump, line 178: whats the use of checking the Meshery-UI endpoint in this stop workflow?
I feel, the below log would be sufficient:

"Meshery has not stopped yet.\nPlease check the status of the pods by executing “mesheryctl system status”".
Why I am not in favor of asking users to run "mesheryctl system dashboard" is - this command might fail as some resources might have got deleted already.
// @piyushsingariya

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -251,3 +251,32 @@ func WaitForPodRunning(c *meshkitkube.Client, desiredPod, namespace string, time
 
 	return pollForPodRunning(c, namespace, desiredPodName, time.Duration(timeout)*time.Second)
 }
+
```
hexxdump, line 254: looks good

mesheryctl/internal/cli/root/system/stop.go
gr455: Ah yes. I copied this message from the start workflow.

gr455:
> It was suggested in the last development meeting that we cancel the `restart` operation if the `stop` operation times out without the deployments stopping.

@piyushsingariya @hexxdump. I would love to know your thoughts on this

piyushsingariya (commented):
I like this strategy but I'm not sure if this is a solid one.
waiting for the namespace to get deleted isn't a good option, rn in `restart` we are deleting namespace and we create it again, although that should not be the case, `restart` should only delete resources and restart the resources.

I would like you to push in a direction where we could poll till the resources are deleted.

gr455:
> I like this strategy but I'm not sure if this is a solid one. waiting for the namespace to get deleted isn't a good option, rn in `restart` we are deleting namespace and we create it again, although that should not be the case, `restart` should only delete resources and restart the resources.
> 
> I would like you to push in a direction where we could poll till the resources are deleted.

This could use the `AreMesheryComponentsRunning` from https://github.com/meshery/meshery/commit/8b8d110283743ccd8b025ae5f51c50c2cddfbc02 right?

piyushsingariya:
@gr455 I think that would work here better!

gr455:
> @gr455 I think that would work here better!

Unfortunately, this does not work because AreMesheryComponentsRunning only checks if the deployments are deleted. When deployments are deleted, the pods might still exist in `terminating` state. So this still fails.

piyushsingariya:
> > @gr455 I think that would work here better!
> 
> Unfortunately, this does not work because AreMesheryComponentsRunning only checks if the deployments are deleted. When deployments are deleted, the pods might still exist in `terminating` state. So this still fails.

Ohh I see, can you check the `--force` flag in the `stop` command, I think you can come up with a function that would be `WaitForMesheryResourceDeleted` which would be a better and best option for doing this.

gr455:
Sorry for bugging you @piyushsingariya, can you give me an idea of what all resources to poll for. Should it be the same as those that are deleted in the `ForceCleanupCluster` function? If so then I am confused why we are deleting replicaSets explicitly, aren't replicasets a part of deployments?

piyushsingariya:
> Sorry for bugging you @piyushsingariya, can you give me an idea of what all resources to poll for. Should it be the same as those that are deleted in the `ForceCleanupCluster` function? If so then I am confused why we are deleting replicaSets explicitly, aren't replicasets a part of deployments?

Yeah, deployment is a higher-level concept that manages replicaSets, I was in flow while writing that code, I forgot and added that piece of code, deleting deployments would delete replicaSets because of the delete policy. You can remove that particular part of code(do test it though).

In an ideal world, we should poll for all kinds of resources Meshery uses in clusters.

navendu-pottekkat:
@gr455 Quick question: Does this handle `system restart` failing to start Meshery as it is being stopped?

gr455:
> @gr455 Quick question: Does this handle `system restart` failing to start Meshery as it is being stopped?

Yes this should handle that. I'm actually changing this PR a bit to poll for components instead of the namespace, but either way it should handle that

gr455:
Seems like another linting problem which was not caught locally :(.
Anyway, this time we poll resources and if namespace is set to be deleted, then we poll namespace too (without this, the restart error persists)

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -290,3 +290,138 @@ func WaitForPodRunning(c *meshkitkube.Client, desiredPod, namespace string, time
...
+				services, err := c.KubeClient.CoreV1().Services(namespace).List(context.TODO(), metav1.ListOptions{})
+				if err != nil {
+					return false, err
+				}
+				return len(services.Items) == 0, nil
+			}
+		}
+
+		return true, nil
+	}
```
piyushsingariya, line 355: Let's not do this with switch case and do it for all resources in one-go.

also not pods/replicasets exactly we should be checking for higher level concepts such as deployments/services/roles.

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -290,3 +290,138 @@ func WaitForPodRunning(c *meshkitkube.Client, desiredPod, namespace string, time
...
+	err = wait.PollImmediate(time.Second, timeout, areMesheryComponentsDeleted(c, namespace, "statefulset"))
+	if err != nil {
+		return err
+	}
+	// Wait for services to delete
+	err = wait.PollImmediate(time.Second, timeout, areMesheryComponentsDeleted(c, namespace, "service"))
+	if err != nil {
+		return err
+	}
+
```
piyushsingariya, line 409: We shouldn't run multiple poll methods like this, it's bad practice.

piyushsingariya (changes requested):

mesheryctl/pkg/utils/healthcheck.go
gr455: The problem is, even when the deployment is deleted, the pods are still in terminating state and are not actually yet deleted. So if we are waiting for the resources to be deleted, then we might not want to return before everything is actually deleted

mesheryctl/pkg/utils/healthcheck.go
gr455: Oh, I did this so that it doesn't poll for everything everytime. For example if all pods are deleted and and some service is yet to be deleted, I would not want the pods to be polled again since we know they're deleted.

mesheryctl/pkg/utils/healthcheck.go
piyushsingariya: Nice point! Understood we can keep it this way then.

gr455: Piyush can you check the comment on the original issue by Lee, We can also not add the polling behaviour for resources at all if we don't need to (it doesn't break anything)

mesheryctl/pkg/utils/healthcheck.go
piyushsingariya: See, what we should do is if any of the resources are present, we poll again. That's how it should work.

And this polling shouldn't call the boolean function each second on each failure instead, as we know stopping a system is a time taking process from the user's perspective and that's why we can take a lease here and call the function after an interval, maybe 7 Seconds.

makes sense??

piyushsingariya: about the linter, it never works locally, at least for me. You can check your lint errors by clicking on 'Files Changed' in the top navigation.

gr455: Yep got it, thanks! (also nudging you to check the comment I made under the issue :sweat_smile:)

piyushsingariya: Yeah, I just checked it, that makes sense, You're right on that. I forgot the original issue and over-complexed it in my head and misdirected you.

I apologies! you can revoke these unnecessary changes. Let's get this wrapped up.

gr455: That's completely fine! Thanks for sticking around for this for so long. Let's get this done with :rocket:

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -290,3 +290,32 @@ func WaitForPodRunning(c *meshkitkube.Client, desiredPod, namespace string, time
...
+			}
+		}
+
+		return true, nil
+	}
+}
+
+// Poll up to timeout seconds every 5 seconds until the namespace no more exists.
+func pollForNamespaceDeleted(c *meshkitkube.Client, namespace string, timeout time.Duration) error {
+	return wait.PollImmediate(5*time.Second, timeout, isNamespaceDeleted(c, namespace))
```
piyushsingariya, line 315: We don't want to PollImmediate here, we should do it interval-based, we discussed right? Deleting namespace takes a good time, we can take a lease here and do it at intervals.

mesheryctl/internal/cli/root/system/check.go
```diff
@@ -581,6 +581,20 @@ func mesheryReadinessHealthCheck() (bool, error) {
...
+		return false, err
+	}
+
+	if err := utils.WaitForNamespaceDeleted(kubeClient, utils.MesheryNamespace, 300); err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
```
piyushsingariya, line 597: shift this to healthcheck.go in utils, this function doesn't align with other functions in check.go

mesheryctl/internal/cli/root/system/stop.go
```diff
@@ -177,11 +178,20 @@ func stop() error {
 			if err = deleteNs(utils.MesheryNamespace, client.KubeClient); err != nil {
 				return err
 			}
+			// Wait for the namespace to be deleted
+			done, err := mesheryNsDeleteCheck()
+			if err != nil {
+				log.Info(err)
+			} else if !done {
+				log.Info("Meshery is taking too long to stop.\nPlease check the status of the pods by executing “mesheryctl system status”.")
```
piyushsingariya, line 186: These two conditions can be merged! The error will normally be thrown by `wait` once the timeout is expired.

mesheryctl/internal/cli/root/system/stop.go
```diff
@@ -177,11 +178,20 @@ func stop() error {
...
+			}
+		} else {
+			log.Info("Meshery is stopped.")
 		}
 	}
 
-	log.Info("Meshery is stopped.")
-
 	// Reset Meshery config file to default settings
 	if utils.ResetFlag {
```
piyushsingariya, line 196: Let's shift this part of the code above because I feel this should be executed before the wait, the wait might return an error in some condition and then this won't be executed.

Let me know your thoughts!!

piyushsingariya (changes requested):

mesheryctl/pkg/utils/healthcheck.go
gr455: This does poll every 5 seconds, it only starts with a poll. I can replace it with `Poll` which starts with a wait which makes more sense because it's not very useful to poll in the start since it'll definitely take more time to stop :+1:

mesheryctl/internal/cli/root/system/check.go
gr455: Got it

mesheryctl/internal/cli/root/system/stop.go
gr455: We are not returning the error from `wait` any further, so it will still get executed (correct me if I'm wrong). If we move it above, we will have to move it before the deleting happens because this bit is common for docker and kubernetes setup.

mesheryctl/internal/cli/root/system/stop.go
gr455: Ah okay. Will do

mesheryctl/internal/cli/root/system/stop.go
piyushsingariya: ahh, it will be okay now that we'll be logging in both cases above.

gr455: So you reckon it's okay to move the config reset code above the delete code for both (k8s and docker)?

piyushsingariya: let's leave it where it is. it would always execute because of logging. I'm already ending up introducing more and more bugs these days :laughing:

mesheryctl/internal/cli/root/system/stop.go
```diff
@@ -177,11 +178,18 @@ func stop() error {
 			if err = deleteNs(utils.MesheryNamespace, client.KubeClient); err != nil {
 				return err
 			}
+			// Wait for the namespace to be deleted
+			_, err := utils.CheckMesheryNsDelete()
+			if err != nil {
```
piyushsingariya, line 183:
```suggestion
			if err != nil || !bool {
```

piyushsingariya (changes requested):
Looks good!

mesheryctl/pkg/utils/healthcheck.go
```diff
@@ -241,6 +241,20 @@ func AreAllPodsRunning() (bool, error) {
 	return true, nil
 }
 
+// CheckMesheryNsDelete waits for Meshery namespace to be deleted, returns (done, error)
+func CheckMesheryNsDelete() (bool, error) {
+	kubeClient, err := meshkitkube.New([]byte(""))
```
piyushsingariya, line 246:
```suggestion
	client, err := meshkitkube.New([]byte(""))
```
Minor edit for consistency across files. 

This is ready to merge.

gr455: Done!

piyushsingariya (approved):
LGTM! Amazing work @gr455 :partying_face: :balloon:

piyushsingariya:
Let's go @gr455 !
image

gr455:
> LGTM! Amazing work @gr455 partying_face balloon

Finally got this done! :tada: Thank you for sticking around

Issue #5008, opened 2022-01-28
[Apps] Application deploy not working despite `name` field being supplied.


leecalcote:
#### Current Behavior
User pastes a K8s manifest into an existing application entry and manually adds the `name` field.

#### Expected Behavior
Application fails to deploy. Empty pattern file is being sent for deployment. The pattern sent for deployment only contains a name field and is missing the rest of the k8s manifest.

#### Server Logs
Example:

```
map[string]interface {}{"name":"Book Info App"}
```

---
#### Contributor [Guides](https://docs.meshery.io/project/contributing) and Resources
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/build-and-release)
- 📚 [Instructions for contributing to documentation](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#documentation-contribution-flow)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and designs for Meshery UI in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
- 🙋🏾🙋🏼 Questions: [Layer5 Discussion Forum](https://discuss.layer5.io) and [Layer5 Community Slack](http://slack.layer5.io)

leecalcote:
Reminder...

Swathi-L:
Hi @Revolyssup,
Is this addressed?

PR #5039, opened 2022-02-02
Support applications #5008 

Revolyssup:
**Description**

This PR makes the following changes-
Applications can be pattern file or any other format. Current supported other format is "K8s". By default it assumes application to be pattern. But , for eg if you pass "format" as "K8s", it will treated as K8s manifest

**Notes for Reviewers**

### DIFF
--- a/handlers/meshery_application_handler.go
+++ b/handlers/meshery_application_handler.go
@@ -23,8 +23,9 @@ type MesheryApplicationRequestBody struct {
 	ApplicationData *models.MesheryApplication `json:"application_data,omitempty"`
 }
 
+//Supported formats other than pattern
 const (
-	K8S_FORMAT = "K8s"
+	k8sFormat = "K8s"
 )
 
 // swagger:route POST /api/application/deploy ApplicationsAPI idPostDeployApplicationFile
@@ -300,7 +301,7 @@ func (h *Handler) formatApplicationOutput(rw http.ResponseWriter, content []byte
 //If an empty or invalid format is passed in header, it will fallback to be treated as a pattern
 func convertApplicationToPattern(r *io.ReadCloser, format string) error {
 	switch format {
-	case K8S_FORMAT:
+	case k8sFormat:
 		data, err := ioutil.ReadAll(*r)
 		if err != nil {
 			return err


Revolyssup:
@sudo-NithishKarthik  In UI, we will have to add a drop down for users to select a format between "K8s" and "pattern".
If format is K8s, We will set the header before sending the request. format=K8s

handlers/meshery_application_handler.go
```diff
@@ -40,7 +50,14 @@ func (h *Handler) ApplicationFileHandler(
 	user *models.User,
 	provider models.Provider,
 ) {
-	// Application files are just pattern files
+	// Applications can be- pattern, k8s manifests,etc. By default if no "format" is present in headers, it will be treated as a pattern file.
+	format := r.Header.Get("format")
```
sudo-NithishKarthik, line 54: Nit: we might want to ask the client to send this information in the body.

sudo-NithishKarthik: just like what we are doing [here](https://github.com/meshery/meshery/blob/master/handlers/meshery_pattern_handler.go#L199)

bariqhibat (approved):
Looks good

handlers/error.go
```diff
@@ -411,3 +412,7 @@ func ErrParsePattern(err error) error {
 func ErrConvertPattern(err error) error {
 	return errors.New(ErrConvertPatternCode, errors.Alert, []string{"Error failed to convert PatternFile to Cytoscape object"}, []string{err.Error()}, []string{}, []string{})
 }
+
+func ErrApplicationConversion(err error) error {
+	return errors.New(ErrApplicationConversionCode, errors.Alert, []string{"Error converting application to a pattern file"}, []string{err.Error()}, []string{"Invalid syntax in the passed format"}, []string{"Make sure the syntax is proper. For eg- If the format is of type kubernetes, make sure it is a valid kubernetes yaml"})
```
leecalcote, line 417: It's great to see longer descriptions like these.

handlers/meshery_application_handler.go
leecalcote: Given today's discussion, I'll withhold approval until the approach is consistent.

Revolyssup: Given the second approach, we have everything already on backend. UI would need to make different API calls and this will fix that. So I am closing the pull request.

Revolyssup:
we have everything already on backend. UI would need to make different API calls and this will fix that. So I am closing the pull request.

PR #5055, opened 2022-02-10
Improve application endpoint

Revolyssup:

**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/models/default_local_provider.go
+++ b/models/default_local_provider.go
@@ -1076,7 +1076,6 @@ func githubRepoApplicationScan(
 		RegisterFileInterceptor(func(f walker.File) error {
 			ext := filepath.Ext(f.Name)
 			if ext == ".yml" || ext == ".yaml" {
-
 				af := MesheryApplication{
 					Name:            strings.TrimSuffix(f.Name, ext),
 					ApplicationFile: string(f.Content),


Revolyssup:
@piyushsingariya  , @Abhishek-kumar09 is changing the way applications are handled at the UI and this PR gets rid of some of the issues as well as introduces this new field "format". We can discuss the changes on a call. Although the new behaviour in the UI will be needed across mesheryctl as well
// @navendu-pottekkat

Revolyssup:
This will also require changes in migration files(table structure) for application table

handlers/meshery_application_handler.go
```diff
@@ -109,18 +109,6 @@ func (h *Handler) handleApplicationPOST(
...
-			obj := "save"
-			h.log.Error(ErrApplicationFailure(err, obj))
-			http.Error(rw, ErrApplicationFailure(err, obj).Error(), http.StatusBadRequest)
-			return
-		}
-
-		// Assign a name if no name is provided
-		if parsedBody.ApplicationData.Name == "" {
-			parsedBody.ApplicationData.Name = applicationName
-		}
```
Abhishek-kumar09, line 123: We have to make sure as the client that we are passing name field. @piyushsingariya 
like: 
```js
{
	"application_data": {
		"name": "Any name",
		"application_file": ".....",
		"format": "pattern" // default is pattern
	}
}
```

Abhishek-kumar09 (approved):
Thanks for the PR @Revolyssup

models/default_local_provider.go
```diff
@@ -1076,13 +1076,9 @@ func githubRepoApplicationScan(
 		RegisterFileInterceptor(func(f walker.File) error {
 			ext := filepath.Ext(f.Name)
 			if ext == ".yml" || ext == ".yaml" {
-				name, err := GetApplicationName(string(f.Content))
-				if err != nil {
-					return err
-				}
```
leecalcote, line 1079:
```suggestion
```

handlers/meshery_application_handler.go
Revolyssup: Yes that is right as we can't extract name from K8s manifest

PR #5057, opened 2022-02-11
[mesheryctl] Make system stop --force not return errors

gr455:
**Description**
Earlier `mesheryctl system stop --force` would return error if it didn't find some resource while deleting, which would panic the application and cause it to return without deleting all the resources.

`mesheryctl system stop --force` does not anymore return error when some resources are not found. It silently logs and goes on.

cc @piyushsingariya 

### DIFF
--- a/mesheryctl/pkg/utils/platform.go
+++ b/mesheryctl/pkg/utils/platform.go
@@ -750,22 +750,6 @@ func ForceCleanupCluster() error {
 		}
 	}
 
-	replicaSetInterface := client.KubeClient.AppsV1().ReplicaSets(MesheryNamespace)
-	replicaSetList, err := replicaSetInterface.List(context.TODO(), metav1.ListOptions{})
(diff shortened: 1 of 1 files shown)


mesheryctl/pkg/utils/platform.go
```diff
@@ -761,7 +761,7 @@ func ForceCleanupCluster() error {
 			if err := replicaSetInterface.Delete(context.TODO(), replicaSet.GetName(), metav1.DeleteOptions{
 				PropagationPolicy: &deletePolicy,
 			}); err != nil {
-				return err
```
piyushsingariya, line 764: we decided to remove replicaSet as we already delete deployments above. remember??

gr455: Ahh yeah, totally forgot about that

piyushsingariya (approved):

PR #5098, opened 2022-02-21
[UI] Fix UI crash due to invalid hook call

Abhishek-kumar09:

Fixes: https://layer5io.slack.com/archives/CFGG6U10E/p1645404841404479

**Description**
Fix problems of hook call inside function.
The hooks follow principles of being called only from the react functional components. 
This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/ui/components/ErrorHandling.js
+++ b/ui/components/ErrorHandling.js
@@ -6,33 +6,35 @@ import { useSnackbar } from "notistack";
  *  Show Snackbar when error occurs. Can be used in catch blocks
  *  of functions.
  *
- * @param {Object} err
- * @param {string} prefixMessage
- * @param {("normal"|"fatal")} severity
  */
-function handleError(err, prefixMessage, severity = "normal") {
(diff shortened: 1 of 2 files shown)


warunicorn19 (commented):
Getting this now, @Abhishek-kumar09 this is an issue with the schema right? 
Screenshot 2022-02-21 at 9 02 33 AM

Abhishek-kumar09:
Yeah, So read though this: 
image

We were right kinda wrong. 
We were expecting all the fields to have `type` property in the schema (necessary to take a lot of decisions), but reading through point 1, it is clear that if any of the two field is present in `x-kubernetes-int-or-string` or `x-kubernetes-preserve-unknown-fields` the schema will not have a `type` property, that was the exception we were missing from the start. 

Thus, the schema is **not** invalid.

warunicorn19:
Got you, Thank you @Abhishek-kumar09

warunicorn19 (approved):
LGTM! Merge at will

PR #5131, opened 2022-02-26
[CI] Relocate workflow config file

leecalcote:


navendu-pottekkat (changes requested):
`golangci-lint` expects a file `.golangci.yml` in the root of the repo as mentioned here. If it doesn't find one, it will search till it reaches the root [as mentioned here](https://golangci-lint.run/usage/configuration/).

This PR changes the filename and the path of the file which is not needed. What this means is that `golangci-lint` can't find the configuration file `.golangci.yml` and defaults to its preset configuration which runs tests we haven't previously (see [commented out code in .golangci.yml](https://github.com/meshery/meshery/blob/e2770a703afc5510eb66119d012576ceae8d0ebc/.golangci.yml#L77)). This is what causing the new lint errors in this PR.

I propose to keep the file back in the root of the repo and reinvestigate the tests that are skipped and check if they should be added to the linter and open a PR fixing those lint errors along with it.

leecalcote:
I'm trying to cleanup our root directory. Cleanup is the purpose of this investigation.

leecalcote:
This linger uses the .github/workflows/ directory in its example setup configuration. Does it not support this config file location?

navendu-pottekkat:
@leecalcote The file in `.github/workflows` runs the action. We already have the action configured to run in places like this https://github.com/meshery/meshery/blob/e2770a703afc5510eb66119d012576ceae8d0ebc/.github/workflows/mesheryctl-ci.yml#L21 (mesheryctl and Meshery server CI).

From [their documentation](https://golangci-lint.run/usage/configuration/), it will check for the configuration file in the current working directory. Which in this case would be the root of our repo.

It will also go up the tree till it reaches root to check for the configuration file. This can work in cases like running the lint check when you are in the mesheryctl folder.

You can also see that `golangci-lint` have their configuration file in the root of the repo: https://github.com/golangci/golangci-lint/blob/master/.golangci.yml

navendu-pottekkat:
It was good that we looked into this. I wasn't aware of this configuration. I will close the existing issue and open up a new one so that we can propose changes to the checks being run with the linter.

navendu-pottekkat:
Opened issue to track this: https://github.com/meshery/meshery/issues/5163

PR #5147, opened 2022-02-28
Fix redundant kubeopenjsonschema instances running at a time

Revolyssup:
**Description**

This PR fixes redundant kubeopenjsonschema instances running at a time
Also adds component creation per context

**Notes for Reviewers**


handlers/error.go
```diff
@@ -428,3 +429,7 @@ func ErrParsePattern(err error) error {
 func ErrConvertPattern(err error) error {
 	return errors.New(ErrConvertPatternCode, errors.Alert, []string{"Error failed to convert PatternFile to Cytoscape object"}, []string{err.Error()}, []string{}, []string{})
 }
+
+func ErrCreatingKubernetesComponents(err error) error {
+	return errors.New(ErrCreatingKubernetesComponentsCode, errors.Alert, []string{"failed to register/create kubernetes components"}, []string{err.Error()}, []string{"Invalid kubeconfig", "Filters passed incorrectly in config", "Could not fetch API resources from Kubernetes server"}, []string{"Make sure that the configuration filters passed are in accordance with output from /openapi/v2"})
```
leecalcote, line 434:
```suggestion
	return errors.New(ErrCreatingKubernetesComponentsCode, errors.Alert, []string{"failed to register/create Kubernetes components"}, []string{err.Error()}, []string{"Invalid kubeconfig", "Filters passed incorrectly in config", "Could not fetch API resources from Kubernetes server"}, []string{"Make sure that the configuration filters passed are in accordance with output from /openapi/v2"})
```

handlers/k8sconfig_handler.go
```diff
@@ -293,41 +301,69 @@ func (h *Handler) GetCurrentContext(token string, prov models.Provider) (*models
...
-					go func(l logger.Handler, config []byte, ctx string) {
-						err := registerK8sComponents(h.log, config, ctxID)
-						if err != nil {
-							logrus.Error(err)
-						}
-					}(h.log, cfg, ctxID)
-				}
-
-				return &ctx, nil
+					h.log.Info("Registration of k8s native components completed for contextID ", ctx)
```
leecalcote, line 321: Are we ensuring that these are Kubernetes native and not custom resources?

handlers/k8sconfig_handler.go
```diff
@@ -293,41 +301,69 @@ func (h *Handler) GetCurrentContext(token string, prov models.Provider) (*models
...
 		}
 	}
 
 	return &cc, nil
 }
-func registerK8sComponents(l logger.Handler, config []byte, ctx string) error {
-	l.Info("Starting to register k8s native components")
-	man, err := core.GetK8Components(config, ctx)
+
+//manages creation of k8s components per context. Such that for each k8s context we have one componentcreation process running at a time
```
leecalcote, line 334:
```suggestion
//manages creation of k8s components per context. Such that for each k8s context we have one component creation process running at a time
```

handlers/k8sconfig_handler.go
```diff
@@ -293,41 +301,69 @@ func (h *Handler) GetCurrentContext(token string, prov models.Provider) (*models
...
+				}
 
-			if ctx.IsCurrentContext {
-				_, _ = prov.SetCurrentContext(token, ctx.ID) // Ignore the error
-
-				if !viper.GetBool("SKIP_COMP_GEN") {
-					ctxID := ctx.ID
-					cfg, err := ctx.GenerateKubeConfig()
+				go func(config []byte, ctx string) {
+					compCreationSingleton.cancelPreviousRun(ctx) //if a registerK8sComponents is still running for the same context then we safely cancel that assuming that run to be stale now
```
leecalcote, line 312: I might not be reading this in the right "context" ( 🙂 ). Will you confirm that Meshery Server will not cancel an existing component generation process upon Meshery UI refresh, but instead, Meshery Server will finish it's current generation task? 

Will you confirm that Meshery Server will only generate Kubernetes components upon `create` of a K8s context?

handlers/k8sconfig_handler.go
Revolyssup: Yes

Revolyssup: We cannot do that. All api-resources are treated similarly by kubernetes whether those are custom or k8s native.
When kubernetes cannot distinguish, so cant we. And we should not as well, although more importantly we cannot.

handlers/k8sconfig_handler.go
Revolyssup: To second question- yes, components are generated on create of a k8s context.
On first question,  no component generation will not be cancelled due to UI refresh.
It will be cancelled when context is reloaded or deleted. Because for reloads or deletes of a k8s context, components generated from it will not be used anywhere anyways and in that case we are wasting computing resources.

Revolyssup: In case of reloads, the component generation will restart with the new reloaded k8s context.

handlers/k8sconfig_handler.go
leecalcote: We can and will.

handlers/error.go
```diff
@@ -428,3 +429,7 @@ func ErrParsePattern(err error) error {
 func ErrConvertPattern(err error) error {
 	return errors.New(ErrConvertPatternCode, errors.Alert, []string{"Error failed to convert PatternFile to Cytoscape object"}, []string{err.Error()}, []string{}, []string{})
 }
+
+func ErrCreatingKubernetesComponents(err error, ctxID string) error {
+	return errors.New(ErrCreatingKubernetesComponentsCode, errors.Alert, []string{"failed to register/create kubernetes components for contextID " + ctxID}, []string{err.Error()}, []string{"component generation was cancelled due to deletion or reload of k8s context", "Invalid kubeconfig", "Filters passed incorrectly in config", "Could not fetch API resources from Kubernetes server"}, []string{"If there is the log \"Starting to register ...\" for the same contextID after this error means that for some reason the context was reloaded which caused this run to abort. In that case, this error can be ignored.", "Make sure that the configuration filters passed are in accordance with output from /openapi/v2"})
```
leecalcote, line 434:
```suggestion
	return errors.New(ErrCreatingKubernetesComponentsCode, errors.Alert, []string{"failed to register/create kubernetes components for contextID " + ctxID}, []string{err.Error()}, []string{"component generation was canceled due to deletion or reload of K8s context", "Invalid kubeconfig", "Filters passed incorrectly in config", "Could not fetch API resources from Kubernetes server"}, []string{"If there is the log \"Starting to register ...\" for the same contextID after this error means that for some reason the context was reloaded which caused this run to abort. In that case, this error can be ignored.", "Make sure that the configuration filters passed are in accordance with output from /openapi/v2"})
```

leecalcote (approved):

Revolyssup:
@leecalcote  good to merge?

leecalcote:
@Revolyssup thanks, yes. 👍 

There is a codecov item to inspect, but yes....

handlers/k8sconfig_handler.go
```diff
@@ -268,12 +270,18 @@ func (h *Handler) GetCurrentContext(token string, prov models.Provider) (*models
 					return cc, nil
 				}
 
-				go func(l logger.Handler, config []byte, ctx string) {
-					err := registerK8sComponents(h.log, config, ctxID)
+				go func(config []byte, ctx string) {
+					compCreationSingleton.cancelPreviousRun(ctx)
+					ctxt, cancel := context.WithTimeout(context.Background(), 300*time.Second)
+					compCreationSingleton.registerCancelFunc(ctx, cancel)
+					h.log.Info("Starting to register k8s native components for contextID ", ctx)
```
leecalcote, line 277:
```suggestion
					h.log.Info("Starting to register Kubernetes core and custom components for contextID ", ctx)
```

leecalcote:
@Revolyssup will you confirm that the APIVERSION, NAMESPACED, and KIND attributes are being cataloged as metadata associated to each generated component (in support of the UI being able to differentiate core vs. custom)?

Revolyssup:
> @Revolyssup will you confirm that the APIVERSION, NAMESPACED, and KIND attributes are being cataloged as metadata associated to each generated component (in support of the UI being able to differentiate core vs. custom)?

Yes

leecalcote (approved):

PR #5261, opened 2022-03-24
[Fix] Use relative path for build files generated by react

humblenginr:
**Description**

The UI was not working initially because the `src` in the `script` tag was using a publicPath rather than a relativePath. Adding the `homepage` field and pointing it to the current directory makes the path relative and hence the UI will work as expected now. 


warunicorn19:
This is awesome.

leecalcote:
> CITY

😆 Ditto

PR #5283, opened 2022-03-27
[CI] Restructure makefile

leecalcote:

**Description**

This PR organizes the Makefile by section and breaks out global variables into a separate file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,34 +13,22 @@
 # limitations under the License.
 
 include install/Makefile.core.mk
-
-all: error run
-
-#-----------------------------------------------------------------------------
-# Meshery Server, CLI and API Errors
-#-----------------------------------------------------------------------------
-.PHONY: error
-error:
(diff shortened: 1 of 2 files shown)


Abhijay007 (commented):
Suggestion: it will be great if we include `default` target in the `makeFile` and add the following instructions related to different targets/commands so that in case the user just run `make` will get the instructions to follow further targets and get an idea about which commands to run next in order to run, build, or test, etc.

leecalcote:
@Abhijay007 good suggestions. Does `all` achieve our desired "default"?

leecalcote:
@Abhijay007, I did another pass, adding the ability for the makefile to self-document, printing help when `make` is run by itself.
Screen Shot 2022-03-27 at 9 30 39 PM

Abhijay007:
LGTM 👍

leecalcote:
Thanks for this review @Abhijay007

PR #5368, opened 2022-04-11
[UI] JSS: Dependency update

leecalcote:

**Description**

This PR fixes #

**Notes for Reviewers**


Abhishek-kumar09 (commented):
the package locks are not human maintained and they don't change someone else's npm builds, and they are prone to modification on npm install.

leecalcote:
> the package locks are not human maintained and they don't change someone else's npm builds, and they are prone to modification on npm install, which is a part of make run

@Abhishek-kumar09, will you rephrase? I'm not sure that I understand the point that you're making. Are you saying that only dependabot should be updating dependencies? Meshery UI is not using the latest JSS, and so this PR is to upgrade to the latest as part of the routine of keeping pace with dependencies. Humans can do this, too, and before Dependabot was acquired by GitHub a couple years ago, humans had to do this manually.

Abhishek-kumar09:
Yeah, the `package-lock` are not modified by us. They are modified automatically based on the package.json file.

Abhishek-kumar09:
This change is good, if we wan't to enforce the latest JSS, but it's not a guarantee.

Abhishek-kumar09 (approved):

PR #5374, opened 2022-04-12
[Docs] Update multiple-of-the-same-adapter.md

leecalcote:
**Description**

This PR updates the multiple adapters FAQ with a little more info.

@jbyers19 might like to expand on this with four or five reasons while users might want to deploy such a configuration.

### DIFF
--- a/docs/pages/guides/multiple-of-the-same-adapter.md
+++ b/docs/pages/guides/multiple-of-the-same-adapter.md
@@ -26,7 +26,9 @@ Alternatively, directly modify youn the `~/.meshery/meshery.yaml` configuration 
 
 #### Configuration: Running more than one instance of the same Meshery adapter
 
-The default configuration of a Meshery deployment includes one instance of each of the Meshery adapters (that have reached a stable version status). You may choose to run multiple instances of the same type of Meshery adapter; e.g. two instances of the Meshery Adapter for Istio. To do so, modify `~/.meshery/meshery.yaml` to include multiple copies of the given adapter.
+The default configuration of a Meshery deployment includes one instance of each of the Meshery adapters (that have reached a stable version status). You may choose to run multiple instances of the same type of Meshery adapter; e.g. two instances of the Meshery Adapter for NGINX Service Mesh. To do so, you can use either of Meshery's clients or to modify your Meshery deployment:
+ - Using `mesheryctl`, modify `~/.meshery/meshery.yaml` to include multiple copies of the given adapter.
+ - Using Meshery UI, navigate to the Settings page and enter the host and port of your additional adapter.
 
 #### Configuration: Choosing an adapter while installing Mesheryctl
 


leecalcote:
An item for @jbyers19's approval...

docs/pages/guides/multiple-of-the-same-adapter.md
```diff
@@ -26,7 +26,9 @@ Alternatively, directly modify youn the `~/.meshery/meshery.yaml` configuration
 
 #### Configuration: Running more than one instance of the same Meshery adapter
 
-The default configuration of a Meshery deployment includes one instance of each of the Meshery adapters (that have reached a stable version status). You may choose to run multiple instances of the same type of Meshery adapter; e.g. two instances of the Meshery Adapter for Istio. To do so, modify `~/.meshery/meshery.yaml` to include multiple copies of the given adapter.
+The default configuration of a Meshery deployment includes one instance of each of the Meshery adapters (that have reached a stable version status). You may choose to run multiple instances of the same type of Meshery adapter; e.g. two instances of the Meshery Adapter for NGINX Service Mesh. To do so, you can use either of Meshery's clients or to modify your Meshery deployment:
+ - Using `mesheryctl`, modify `~/.meshery/meshery.yaml` to include multiple copies of the given adapter.
+ - Using Meshery UI, navigate to the Settings page and enter the host and port of your additional adapter.
```
jbyers19, line 31: Wouldn't it be better just to update the adapter deployment manifest with the number of replicas desired? Since the adapters use a `svc`, you can spin up as many replicas (instances) of the adapter as you want without having to worry about the host and port. The only exception would be if you wanted to have adapters in different namespaces, which I would recommend against since it adds unnecessary complexity at this time.

leecalcote: Good thinking. It might be that we should provide the Kubernetes-based instruction you've given here.

The approach taken currently accounts for Docker-based deployments as well.

jbyers19: I'll be honest, I completely forgot about the Docker use-case because I always use k8s. However, would it be possible to have a similar configuration using [Docker services](https://docs.docker.com/engine/swarm/how-swarm-mode-works/services/)? I'm not very familiar with using Docker swarms or how Meshery is deployed in Docker, but the documentation looks promising.

leecalcote: I think your suggestions are spot on: that users can first look to their container orchestrator for multiple replicas of an adapter and that should they desire to have separate, named copies (for any of these reasons - https://discuss.layer5.io/t/is-there-ever-a-case-where-more-than-1-adapter-will-be-deployed/726/2) that they can then use this sort of configuration within Meshery. 

Please don't hesitate to open an update on this set of text. I think your comments are well-placed and helpful to users.

PR #5400, opened 2022-04-17
Do not change service name in patternfile for designs unless duplicated

Revolyssup:

**Description**
Currently all patternfile services names are suffixed with random alpha-numeric. This creates the following problem:
In a lot of cases, exact name matters to users. For instance, if a `Role` is created with a specific name, that exact name will be used in Role binding. If a k8s `service` is created with a specific name, that exact name might be used in some other resource. 
Randomizing the names will cause problem in such cases.

Solution:
If users want to pass their own unique names, then that behavior must be permissible to mitigate the above problem. If however, duplicate service names are found, in that case those duplicated names can be randomized considering user probably does not care much about the names.

**Notes for Reviewers**

### DIFF
--- a/models/pattern/core/pattern.go
+++ b/models/pattern/core/pattern.go
@@ -209,14 +209,21 @@ func NewPatternFileFromCytoscapeJSJSON(name string, byt []byte) (Pattern, error)
 	dependsOnMap := make(map[string][]string, 0) //used to figure out dependencies from traits.meshmap.parent
 	eleToSvc := make(map[string]string)          //used to map cyto element ID uniquely to the name of the service created.
 	countDuplicates := make(map[string]int)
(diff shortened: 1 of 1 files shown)


leecalcote:
The description you've given makes intuitive sense. Another way of stating this is: "If there's no conflicting names, don't bother mutating the names as their uniqueness is already account for."

Revolyssup:
> The description you've given makes intuitive sense. Another way of stating this is: "If there's no conflicting names, don't bother mutating the names as their uniqueness is already account for."

Yes, precisely.

leecalcote (approved):

leecalcote:
A follow up to this PR is one of avoiding use of `-` at the start of generated names as K8s considers this outside of acceptable naming convention.

PR #5413, opened 2022-04-19
Specify namespace as default when missing

Revolyssup:

**Description**

This PR fixes #

**Notes for Reviewers**


warunicorn19 (approved):

models/pattern/core/pattern.go
```diff
@@ -370,7 +370,9 @@ func createPatternServiceFromCoreK8s(manifest map[string]interface{}) (string, S
 	namespace, _ := metadata["namespace"].(string)
 	labels, _ := metadata["labels"].(map[string]interface{})
 	annotations, _ := metadata["annotations"].(map[string]interface{})
-
+	if namespace == "" {
```
leecalcote, line 373: Let's confirm that objects that are not namespaced are not affected by this logic; that non-namespaced objects are NULL, not empty string.

leecalcote (approved):

models/pattern/core/pattern.go
leecalcote: Do we have a null vs empty string concern?

PR #5492, opened 2022-05-08
Add `--provider` flag to `system login`

alphaX86:

**Description**

This PR adds provider flag to system login

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/pkg/utils/auth.go
+++ b/mesheryctl/pkg/utils/auth.go
@@ -396,7 +396,7 @@ func chooseDirectProvider(provs map[string]Provider, option string) (Provider, e
 	}
 
 	for i := range provNames {
-		if strings.ToLower(provNames[i]) == strings.ToLower(option) {
+		if strings.EqualFold(provNames[i], option) {
 			return provArray[i], nil
 		}
 	}


leecalcote:
We need a `--provider` flag instead, I believe.

leecalcote:
Meshery supports more than one provider (conceptually any number). For the same use case that someone might want to specify `--none` to script their login, others will also want to supply alternative provider names, e.g. `--provider Meshery`.

Yes, please change it (without necessarily waiting for me to ask twice, pretty please).

leecalcote (changes requested):
Requesting change for holistic consideration of provider flag, not None as a one-off.

alphaX86:
> Meshery supports more than one provider (conceptually any number). For the same use case that someone might want to specify `--none` to script their login, others will also want to supply alternative provider names, e.g. `--provider Meshery`.
> 
> Yes, please change it (without necessarily waiting for me to ask twice, pretty please).

Understood captain, it'll be executed soon (_Have to do this before another comment arises 🌚_)

mesheryctl/pkg/utils/auth.go
```diff
@@ -372,6 +413,26 @@ func selectProviderPrompt(provs map[string]Provider) Provider {
...
+	for _, prov := range provArray {
+		provNames = append(provNames, prov.ProviderName)
+	}
+
+	for i := 0; i < len(provNames); i++ {
+		if provNames[i] == option {
+			return provArray[i], nil
+		}
+	}
+	return provArray[1], fmt.Errorf("the specified provider '%s' is not available. Please try giving correct provider name", option)
```
piyushsingariya, line 433: why don't we directly iterate the map itself?? and match _lowercased_ keys with the _lowercased_ option

mesheryctl/pkg/utils/auth.go
```diff
@@ -283,6 +283,47 @@ func InitiateLogin(mctlCfg *config.MesheryCtlConfig) ([]byte, error) {
 	return data, nil
 }
 
+// Initiate Login on provider mentioned by user
+func InitiateLoginDirect(mctlCfg *config.MesheryCtlConfig, option string) ([]byte, error) {
```
piyushsingariya, line 287:
```suggestion
func InitiateDirectLogin(mctlCfg *config.MesheryCtlConfig, option string) ([]byte, error) {
```

leecalcote:
I'm trying to recall the use case... I believe there is the need to use `system login` from scripted environments, in which case, the interactive selector doesn't meet the need.

mesheryctl/pkg/utils/auth.go
alphaX86: Never thought at that time 😶

alphaX86:
The prompt is messed up when it comes to script execution and even when Unicode is passed, it's messing up totally. That's why the need of `--provider` is needed

mesheryctl/pkg/utils/auth.go
alphaX86: Function refactored

piyushsingariya: I don't think the function is optimised, we're increasing the memory use here, isn't iterating the map itself a good approach in terms of code?

alphaX86:
Almost forgot the map, I always thought its array, I'll change the way soon

mesheryctl/pkg/utils/auth.go
alphaX86: Will change soon

mesheryctl/pkg/utils/auth.go
```diff
@@ -393,8 +395,8 @@ func chooseProviderDirect(provs map[string]Provider, option string) (Provider, e
 		provNames = append(provNames, prov.ProviderName)
 	}
 
-	for i := 0; i < len(provNames); i++ {
-		if provNames[i] == option {
+	for i := range provNames {
+		if strings.ToLower(provNames[i]) == option {
```
piyushsingariya, line 399:
```suggestion
		if strings.ToLower(provNames[i]) == strings.ToLower(option) {
```

leecalcote (approved):

PR #5499, opened 2022-05-10
Allow connecting to multiple brokers 

Revolyssup:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/internal/graphql/model/helpers.go
+++ b/internal/graphql/model/helpers.go
@@ -311,7 +311,7 @@ func (k *K8sConnectionTracker) Get(id string) (url string) {
 	return
 }
 
-// Takes the meshkit Logger and logs a comma seperated list of currently tracked Broker Endpoints
+// Takes the meshkit Logger and logs a comma separated list of currently tracked Broker Endpoints
 func (k *K8sConnectionTracker) Log(l logger.Handler) {
 	var e = "Connected broker endpoints : "
 	k.mx.Lock()


leecalcote (approved):

internal/graphql/model/operator_helper.go
```diff
@@ -172,6 +176,7 @@ func SubscribeToBroker(provider models.Provider, mesheryKubeClient *mesherykube.
 		if err == nil && broker.Status.Endpoint.External != "" {
 			break
```
leecalcote, line 177: @Revolyssup the error handling here seems like it needs improvement.

internal/graphql/model/helpers.go
```diff
@@ -291,3 +291,31 @@ func SetOverrideValues(delete bool, adapterTracker models.AdaptersTrackerInterfa
...
+}
+
+func (k *K8sConnectionTracker) Get(id string) (url string) {
+	k.mx.Lock()
+	defer k.mx.Unlock()
+	url = k.ContextToBroker[id]
+	return
+}
+
+func (k *K8sConnectionTracker) Log(l logger.Handler) {
```
leecalcote, line 313: A couple of comments on these functions would be helpful.

internal/graphql/model/operator_helper.go
```diff
@@ -206,20 +211,36 @@ func SubscribeToBroker(provider models.Provider, mesheryKubeClient *mesherykube.
...
 		MaxReconnect:   5,
 	})
 	// Hack for minikube based clusters
-	if err != nil {
+	if err != nil && conn == nil {
 		return endpoint, err
 	}
+	defer func() {
+		if conn == nil {
+			return
```
leecalcote, line 230: Is connection state being tracked anywhere? If so, should this update that state?

leecalcote (changes requested):

internal/graphql/model/operator_helper.go
Revolyssup: This connection is local to the function call SubscribeToBroker so it doesn't need to be tracked anywhere else. 
Just few lines below, you will find a DeepCopy method being invoked which actually copies this local instance of connection into globally maintained broker connection if no errors are encountered.

internal/graphql/model/helpers.go
Revolyssup: Sure

internal/graphql/model/operator_helper.go
Revolyssup: I think it is fine. It spins in a loop with a timeout until we can get the broker without any errors AND it has been assigned an External IP. This logic seems fine to me. Can you point out what part of it needs improvement?

leecalcote (approved):

PR #5512, opened 2022-05-13
[Bug] Solve panic during performance test

theBeginner86:
**Description**
This PR replaces the improper err handling for performing load tests.
Earlier, we were logging an err through an Err function by sending nil err value. This was specific to an environment in which performance tests are performed without any K8s cluster on the host machine or meshery is not able to detect the k8s context.
So it was leading to the following panic:
Screenshot 2022-05-13 at 5 57 58 AM

This was even making smp-action false positive. 
Check this: 
1. https://github.com/layer5io/meshery-smp-action/runs/6415031249?check_suite_focus=true#step:6:1819
2. https://github.com/layer5io/meshery-smp-action/runs/6407123058?check_suite_focus=true#step:7:1308

Fix:
Created a new Meshkit err to be used in this case.

Demo:

https://user-images.githubusercontent.com/73700530/168194984-ee4476d1-8cfe-4bb5-8c63-ee96dee66afb.mov

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/handlers/load_test_handler.go
+++ b/handlers/load_test_handler.go
@@ -365,7 +365,8 @@ func (h *Handler) executeLoadTest(ctx context.Context, req *http.Request, profil
 	// Get the context
 	mk8scontext, ok := req.Context().Value(models.KubeContextKey).(*models.K8sContext)
 	if !ok || mk8scontext == nil {
-		h.log.Error(ErrLoadTest(err, "unable to perform: failed to identify kubernetes context"))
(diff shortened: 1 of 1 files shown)


leecalcote:
Some quality work here, @theBeginner86

leecalcote:
This is clearly an oversight. Thanks for documenting and fixing it, @theBeginner86

leecalcote (approved):

PR #5536, opened 2022-05-25
[Server] Add support to get MeshSync version

MUzairS15:
**Description**

This PR retrieves the version from MeshSync which is set as an env variable .
Related https://github.com/meshery/meshsync/pull/128.

**Notes for Reviewers**

### DIFF
--- a/internal/graphql/model/error.go
+++ b/internal/graphql/model/error.go
@@ -19,7 +19,7 @@ const (
 	ErrNoMeshSyncCode               = "1011"
 	ErrNoExternalEndpointCode       = "1012"
 	ErrApplyHelmChartCode           = "2155"
-	ErrGetVersionCode               = "2175"  
+	ErrGetVersionCode               = "2175"
 )
 
 var (
--- a/internal/graphql/model/operator_helper.go
+++ b/internal/graphql/model/operator_helper.go
@@ -271,8 +271,8 @@ func getVersion(clientset *mesherykube.Client) (string, error) {
 	meshsyncPod, _ := clientset.KubeClient.CoreV1().Pods("meshery").List(context.TODO(), metav1.ListOptions{
 		LabelSelector: "component=meshsync",
 	})
-	
-	if (!(len(meshsyncPod.Items) > 0)) {
+
+	if !(len(meshsyncPod.Items) > 0) {
 		return "", nil
 	}
 	podName := meshsyncPod.Items[0].Name
@@ -280,7 +280,7 @@ func getVersion(clientset *mesherykube.Client) (string, error) {
 	request := clientset.KubeClient.CoreV1().RESTClient().Post().Resource("pods").Name(podName)
 
 	request.VersionedParams(
-        &v1.PodExecOptions{
+		&v1.PodExecOptions{
 			Command: []string{"env"},
 			Stdin:   true,
 			Stdout:  true,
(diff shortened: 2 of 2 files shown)


MUzairS15:
Pod.yml
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
  - name: test
    image: test:v0.0.1
    env:
    - name: ENV2 <--------------
      value: "test2"
```
Dockerfile
```
FROM nginx:latest  
ENV ENV1="test1"  <---------------
CMD ["nginx", "-g", "daemon off;"]
```

`kubectl describe pod test`
```
Containers:
  test:
    Container ID:   docker://856a77bef30d0c614a4e942ff994854bfbbce8a7ea9e1d6e79a2e30ed012580e
    Image:          test:v0.0.1
    Image ID:       docker://sha256:55281ef707f82df3772c25f45ff304b2e3f54f7374d748f1c2ed2e92aecae7bb
    Port:           <none>
    Host Port:      <none>
    State:          Running
      Started:      Fri, 27 May 2022 12:47:04 +0530
    Ready:          True
    Restart Count:  0
    Environment:
      ENV2:  test2 <-----------
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-9g8vf (ro)

**Comment by MUzairS15** [comment]
@leecalcote

File: internal/graphql/model/error.go [hunk 1]
```diff
@@ -64,3 +65,7 @@ func ErrMesheryClient(err error) error {
 func ErrApplyHelmChart(err error) error {
 	return errors.New(ErrApplyHelmChartCode, errors.Alert, []string{"Error occurred while applying Helm Chart"}, []string{err.Error()}, []string{"Kubernetes cluster might not be connected", "Leftover resources from partial install"}, []string{"Try reinstalling", "Try reconnecting your kubernetes cluster", "Clean up artifacts from preinstalled helm release manually"})
 }
+
+func ErrGetVersion(err error) error {
+	return errors.New(ErrGetVersionCode, errors.Alert, []string{"Error getting MeshSync version.", err.Error()}, []string{"Unable to reach MeshSync Pod"}, []string{"Check MeshSync is healthy"}, []string{})
```
leecalcote, line 70: Extremely concise, but good to have each of these fields defined.

internal/graphql/model/operator_helper.go
```diff
@@ -117,11 +122,12 @@ func GetMeshSyncInfo(mesheryclient operatorClient.Interface, mesheryKubeClient *
...
 		return meshsyncStatus, ErrMesheryClient(err)
 	}
-
-	meshsyncDeployment, err := mesheryKubeClient.KubeClient.AppsV1().Deployments("meshery").Get(context.TODO(), "meshery-meshsync", metav1.GetOptions{})
-	meshsyncVersion := ""
-	if err == nil {
-		meshsyncVersion = imageVersionExtractUtil(meshsyncDeployment.Spec.Template, "meshsync")
+	meshsyncVersion, err := getVersion(mesheryKubeClient)
+	if meshsyncVersion == "" {
+		meshsyncVersion = "latest"
```
leecalcote, line 127: Can we also retrieve the associated version (git) tag? e.g. `stable-v0.5.19`?

MUzairS15: Yes.

MUzairS15:
@leecalcote I Just realised, that since we are using distroless image for MeshSync we don't have access to the shell. This will fail unless we use alpine or similar as base image in MeshSync.

internal/graphql/model/operator_helper.go
leecalcote: Good. We’ll want to send back the specific version tag as opposed to `latest`

MUzairS15: In any case, MeshSync is not running or not started, instead of showing nothing (blank), latest will be shown first-time and when again MeshSync starts successfully actual version tag will be visible.

PR #5548, opened 2022-05-28
Added multiple k8s version for adapter tests

### LINKED ISSUE #5547: [CI] Modify core E2E workflow to accommodate multiple k8s versions
#### Current Behavior

Currently we only allow users to pass one k8s version and run e2e tests on that. But we need to be able to check our compatibility with latest as well as older versions of kubernetes.
#### Desired Behavior

Allow users to pass comma separated values of versions in `k8s_versions` inside the test_adapters.yaml workflow. Use those to run tests for each of the passed version
#### Acceptance Tests

---
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/build-and-release)

4molybdenum2:
Signed-off-by: Tathagata Paul <tathagatapaul7@gmail.com>

**Description**

Closes #5547 

**Notes for Reviewers**
This change will essentially mean that we can remove the workflow input parameter: ```k8s_version``` from every adapter's e2e test file which references the ```test_adapters``` file

### DIFF
--- a/.github/workflows/test_adapters.yaml
+++ b/.github/workflows/test_adapters.yaml
@@ -29,6 +29,9 @@ on:
         required: false
         type: string
         default: 'bridge'
+      k8s_version:
+        required: false
+        type: string
       sm_version:
         required: true
         type: string


4molybdenum2:
cc @Revolyssup

.github/workflows/test_adapters.yaml
```diff
@@ -25,9 +25,6 @@ on:
       patternfile_name:
         required: true
         type: string
-      k8s_version:
-        required: true
-        type: string
```
gr455, line 30: This will remove backwards compatibility. We will need another sweeping change across the adapters. I think it's better to keep using string parsed arrays like the other fields. What do you think @4molybdenum2?

4molybdenum2: Hmm, thats true...
Currently we cant pass array types as workflow inputs... so another workaround maybe passing strings and parsing them according to some delimiter (eg: comma).

On the other hand if we pass the k8s version using matrix then the same k8s version will be used to test against all the adapters. As for the backwards compatitbility I don't think the tests will fail because the k8s_version taken from the inputs are all replaced by ```matrix.k8s_version```. So, it will not technically break the CI tests if I am correct..? But yes, we will have to remove all those k8s_version fields from the individual adapters.

gr455: It does fail, I'm afraid. Have a look at https://github.com/gr455/meshery-istio/actions/runs/2401529126
It expects the input field to be in the core workflow

4molybdenum2: Ah, I see...then let me see how to do it by passing a string of versions. I guess that's one way to tackle it

gr455: Sounds good!

4molybdenum2: @gr455 I just made the k8s_version a field and set it to ```not required```. Can you check if it works now?

.github/workflows/test_adapters.yaml
```diff
@@ -50,6 +50,10 @@ on:
 
 jobs:
   TestAdapter:
+    strategy:
+      fail-fast: false
+      matrix:
+        k8s_version: ['v1.20.11', 'v1.21.5', 'v1.22.2']
```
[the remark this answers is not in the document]
4molybdenum2: We can but then we will have to make changes to all the adapters. See here: https://github.com/meshery/meshery/pull/5548#discussion_r884170036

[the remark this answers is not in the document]
4molybdenum2: For this change too we need to remove all the k8s_version variable from all the adapters...but we can do that separately.

warunicorn19:
hey @4molybdenum2, could you please share this topic on the call.

gr455:
@4molybdenum2 If you want to test out the github workflow, you can just fork the meshery-istio repository and change the [uses](https://github.com/meshery/meshery-istio/blob/master/.github/workflows/e2etests.yml#L50) section of the workflow to `4molybdenum2/meshery/.github/workflows/test_adapters.yaml@allow-multiple-k8s-version`

4molybdenum2:
@Revolyssup @gr455 tests seems to be running fine for the k8s versions https://github.com/4molybdenum2/meshery-istio/runs/6703379423?check_suite_focus=true. I believe the update docs failure is due to missing credentials?

gr455:
> @Revolyssup @gr455 tests seems to be running fine for the k8s versions https://github.com/4molybdenum2/meshery-istio/runs/6703379423?check_suite_focus=true. I believe the update docs failure is due to missing credentials?

Yep. Looks good to me! @Revolyssup

gr455 (approved):

Revolyssup (approved):

PR #5743, opened 2022-07-02
[Server] Lint Fix

gr455:
**Description**

Lint fix

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -240,7 +240,10 @@ func main() {
 	k8sComponentsRegistrationHelper := models.NewComponentsRegistrationHelper(log)
 
 	// purge the meshsync objects in the database on startup
-	_ = models.RemoveStaleObjects(*dbHandler)
+	err = models.RemoveStaleObjects(*dbHandler)
+	if err != nil {
+		logrus.Error(err)
+	}
 	h := handlers.NewHandlerInstance(hc, meshsyncCh, log, brokerConn, k8sComponentsRegistrationHelper, mctrlHelper, dbHandler)
 
 	b := broadcast.NewBroadcaster(100)


leecalcote:
A couple more static analysis items…

cmd/main.go
```diff
@@ -240,7 +240,7 @@ func main() {
 	k8sComponentsRegistrationHelper := models.NewComponentsRegistrationHelper(log)
 
 	// purge the meshsync objects in the database on startup
-	models.RemoveStaleObjects(*dbHandler)
+	_ = models.RemoveStaleObjects(*dbHandler)
```
sudo-NithishKarthik, line 243: it's better to handle the error instead of ignoring it..

gr455: You got it. I just didn't want to step on your toes in case that ignore was intentional.

cmd/main.go
```diff
@@ -240,7 +240,10 @@ func main() {
 	k8sComponentsRegistrationHelper := models.NewComponentsRegistrationHelper(log)
 
 	// purge the meshsync objects in the database on startup
-	_ = models.RemoveStaleObjects(*dbHandler)
+	err = models.RemoveStaleObjects(*dbHandler)
+	if err != nil {
+		logrus.Fatal(err);
```
sudo-NithishKarthik, line 245: Generally, responding with a fatal error when something goes wrong while initialization is the right thing to do. But in this case, the server will run properly without any issues if it could not purge the database. So, a warning or an error would be a more suitable choice.

sudo-NithishKarthik:
oops, another lint error 😅

gr455:
> oops, another lint error sweat_smile

I gotta add the gofmt extension to my Sublime :laughing:

sudo-NithishKarthik (approved):
LGTM 🎉

PR #5806, opened 2022-07-13
Disable un/deploy operation when no contexts detected.

MUzairS15:
**Description**

This PR disables un/deploy button when no context is detected. 

**Notes for Reviewers**


ui/components/ConfirmationModal.js
```diff
@@ -155,13 +155,20 @@ function ConfirmationMsg(props) {
   const { classes, open, handleClose, submit,
     selectedK8sContexts, k8scontext, title, validationBody, setK8sContexts, enqueueSnackbar, closeSnackbar, componentCount, tab, isVerify } = props
 
-  const [contexts, setContexts] = useState(k8scontext);
   const [tabVal, setTabVal] = useState(tab);
+  const [disabled, setDisabled] = useState(true);
+
+  let isDisabled = typeof selectedK8sContexts.length === "undefined" || selectedK8sContexts.length === 0
```
leecalcote, line 161: How is this not a global variable? is this something that v0.7.0's approach to state management will address, @sudo-NithishKarthik?

ui/components/ConfirmationModal.js
```diff
@@ -373,10 +380,13 @@ function ConfirmationMsg(props) {
                   {/* colorchange  */}
                 </Button>
                 <Button onClick={handleSubmit}
-                  className={tabVal == 1 ? classes.button3 : classes.button0} autoFocus type="submit"
+                  className={isDisabled ? (tabVal === 1 ? classes.disabledBtnDel : classes.button0) : tabVal == 1 ? classes.button3 : classes.button0}
```
leecalcote, line 383: @MUzairS15 please don't use numeric index references, but instead use named keys.

leecalcote (changes requested):

Revolyssup:
@MUzairS15  Why are we disabling undeploy or deploy in case of no k8s context? We are completely extensible and do not enforce presence of k8s context for any operation. // @leecalcote  
If the operation requires k8s context(which it does currently in most cases), it will throw an appropriate error. But we should not enforce this otherwise we are tying ourselves to kubernetes in future.

leecalcote:
> @MUzairS15 Why are we disabling undeploy or deploy in case of no k8s context? We are completely extensible and do not enforce presence of k8s context for any operation. // @leecalcote
> If the operation requires k8s context(which it does currently in most cases), it will throw an appropriate error. But we should not enforce this otherwise we are tying ourselves to kubernetes in future.

This is a great point and forward-thinking. I like it. Of the lifecycle management and configuration management operations offered in the current version of Meshery, all require Kubernetes. We'll want to disallow users from shooting themselves in the foot with an underlay operation that will fail in the absence of a K8s connection (and quite candidly, might fail in the presence of one, too), while at the same time, knowing that the platform will expand, we don't want to paint ourselves into an architectural corner. Thanks for keeping a vigil eye here.

In summary, these flowery words mean: don't show "un/deploy" unless a target system is available to provision / configure against.

ui/components/ConfirmationModal.js
MUzairS15: `const [disabled, setDisabled] = useState(true)` In a sense it is a global variable, as wherever we perform an operation that requires context, this modal is present. And no other component requires it

MUzairS15:
@Revolyssup Take this as of all the environments we provide (or will provide) if none of them is available disable the operation.

ui/components/ConfirmationModal.js
MUzairS15: done

ui/components/ConfirmationModal.js
leecalcote: @MUzairS15 are you aware of how this will change when there is global state management in v0.7.0? What note(s) do we need to take now in preparation?

ui/components/ConfirmationModal.js
```diff
@@ -346,7 +357,7 @@ function ConfirmationMsg(props) {
                 </DialogContentText>
               </DialogContent>
           }
-          {tabVal === 2 && isVerify &&// Validate
+          {tabVal === ACTIONS.VERIFY &&// Validate
```
leecalcote, line 360: Thank you, @MUzairS15

leecalcote (approved):

MUzairS15:
@Chadha93 ok to merge?

Chadha93:
> @Chadha93 ok to merge?

Yes.

PR #5885, opened 2022-07-26
[Meshery-Extension] setup single source of truth for svgs

Abhishek-kumar09:

**Description**
- Adds up the component svgs to be used by meshery as well as meshery-extension under a static file host.
This PR fixes #

**Notes for Reviewers**


Abhishek-kumar09:
Depends on https://github.com/meshery/meshery/pull/5885

leecalcote (changes requested):
Things like the `cluster-role.svg` will need two versions: one without the shape (circle, which should be a rounded rectangle) and one with the shape.

Abhishek-kumar09:
> Things like the `cluster-role.svg` will need two versions: one without the shape (circle, which should be a rounded rectangle) and one with the shape.

Hmmm, Okay. This PR don't has the cluster role changes. I think @Harshit-Dandriyal is tracking that in his seperate branch. So the svgs can be added and chosen.

Abhishek-kumar09:
For now merging this PR, since it has some crucial changes wrt visualiser fixes.

PR #5888, opened 2022-07-27
fix(ci): publish linux and mac artifacts in tar.gz format

nishantwrp:
Signed-off-by: Nishant Mittal <mittalnishant14@outlook.com>

**Description**

This PR fixes part of #5881 

**Notes for Reviewers**
Publish linux & macos artifacts for mesheryctl in `tar.gz` format. Artifacts for windows will still be published in `zip` format.

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [X] Yes, I signed my commits.

Issue #5881, opened 2022-07-24
[mesheryctl] installing mesheryctl via the script shouldn't require unzip


nishantwrp:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

#### Current Behavior
When `mesheryctl` is installed via our bash script it requires `unzip` to be present so that the script can uncompress the released cli binary. 

https://github.com/meshery/meshery/blob/df9eb35b8d45d1b1899b29111c80eeaed423d57b/meshery#L120-L123

#### Desired Behavior
This shouldn't be the case because linux by default doesn't have support for `zip` but has for `tar.gz`. The plan is to
- Publish the release artifact in both `tar.gz` and `zip` formats.
- Update the bash script to use the `tar.gz` file if possible.

#### Environment

- Host OS: Mac Linux Windows
- Meshery Server Version: stable-v
- Meshery Client Version: stable-v
- Platform: Docker Kubernetes

---

#### Contributor Resources
- [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- 🙋🏾🙋🏼 Questions: [Layer5 Discussion Forum](https://discuss.layer5.io) and [Layer5 Community Slack](http://slack.layer5.io)

heysujal:
Hi @nishantwrp  i want to work on this issue , can you guide me how do I get started ?
I am very new to this project and tried to gain some familiarity but I am not able to understand terms like Mesh and so forth.

hexxdump:
@leecalcote, meshery bash script is used only by the unix based operating systems, so if I modify the script to switch to tar.gz, it would be enough, right ? I dont need to make any windows specific differentiation in the meshery script, right?

nishantwrp:
[...]

PR #5911, opened 2022-07-29
[mesheryctl]: change meshery bash script to consider the .tar.gz format

hexxdump:
**Description**
Change meshery bash script to consider the .tar.gz format bundle

To avoid the dependency on unzip in the *nix based systems,
goreleaser.yml will publish mesheryctl package bundled in .tar.gz format
rather than in .zip format. For Windows platform, mesheryctl is packaged
in zip format it self
This PR fixes part of #5881 

**Notes for Reviewers**
This commit should go in after PR https://github.com/meshery/meshery/pull/5888 goes in.

### DIFF
--- a/meshery
+++ b/meshery
@@ -117,8 +117,8 @@ if ! command_exists curl ; then
     exit;
 fi
 
-if ! command_exists unzip ; then
-    echo "Missing required utility: 'unzip'. Please install 'unzip' and try again."
+if ! command_exists tar; then
+    echo "Missing required utility: 'tar'. Please install 'tar' and try again."
     exit;
 fi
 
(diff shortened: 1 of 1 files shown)


nishantwrp (changes requested):
Hi @hexxdump, we should probably remove this block too. 

https://github.com/meshery/meshery/blob/25559fc547f6702ebc7ad040e30a06c880ec198d/meshery#L120-L123

hexxdump:
> Hi @hexxdump, we should probably remove this block too.
> 
> https://github.com/meshery/meshery/blob/25559fc547f6702ebc7ad040e30a06c880ec198d/meshery#L120-L123

yes, good catch, fixed it

nishantwrp (dismissed):
lgtm, thanks for the changes @hexxdump. 

Just one thought, should we have a preflight check for `tar` like we had for `unzip`? Most linux distros have `tar` by default but just in case. What do you think @hexxdump?

hexxdump:
> lgtm, thanks for the changes @hexxdump.
> 
> Just one thought, should we have a preflight check for `tar` like we had for `unzip`? Most linux distros have `tar` by default but just in case. What do you think @hexxdump?

makes sense @nishantwrp, added the preflight check for tar.

nishantwrp (approved):
Lgtm

alphaX86 (approved):
LGTM

leecalcote:
@AugustasV is this along the lines of what you were thinking? Does this satisfy your deployment scenario?

nishantwrp:
@leecalcote, i think this pr should be merged too. This updates the mesheryctl installation script to use `tar.gz` artifacts.

leecalcote (approved):
LGTM

leecalcote:
Thank you, @hexxdump! 😄

Issue #5905, opened 2022-07-28
[mesheryctl] Failure to recognize valid meshconfig


leecalcote:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

#### Current Behavior
```
$ mesheryctl system channel switch stable
Valid context is not available in meshconfig
Error: failed to initialize healthchecker: Valid context is not available in meshconfig
```
```
$ mesheryctl system context view
Error: current context not set
```
```
$ mesheryctl system start
Valid context is not available in meshconfig
Error: Valid context is not available in meshconfig
```
```
$ cat ~/.meshery/config.yaml
contexts:
    local:
        endpoint: http://localhost:9081
        token: Default
        platform: kubernetes
        components:
            - meshery-app-mesh
            - meshery-istio
            - meshery-linkerd
            - meshery-consul
            - meshery-nsm
            - meshery-kuma
            - meshery-osm
            - meshery-traefik-mesh
            - meshery-nginx-sm
            - meshery-cilium
        channel: edge
        version: latest
current-context: local
tokens:
    - location: auth.json
      name: Default
```

#### Environment

- Host OS: Mac 
- Meshery Server Version: stable-v0.6.0-rc.6d
- Meshery Client Version: stable-v0.6.0-rc.6d
- Platform: Kubernetes

---

#### Contributor Resources
- [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- 🙋🏾🙋🏼 Questions: [Layer5 Discussion Forum](https://discuss.layer5.io) and [Layer5 Community Slack](http://slack.layer5.io)

warunicorn19:
Might be something coming from this PR. https://github.com/meshery/meshery/pull/5877

alphaX86:
Something related to path referencing... I'll take this up and resolve quick today

PR #5930, opened 2022-08-01
Add tests for mesheryctl root config and flags

nishantwrp:
Signed-off-by: Nishant Mittal <mittalnishant14@outlook.com>

**Description**

This PR fixes n/a. These tests detect errors in config loading preventing bugs like: https://github.com/meshery/meshery/issues/5905

**Notes for Reviewers**


mesheryctl/internal/cli/root/root_test.go
```diff
@@ -0,0 +1,86 @@
...
+	testConfigPath := filepath.Join(fixturesDirPath, "testConfig.yaml")
+
+	tests := []RootCmdTestInput{
+		{
+			Name: "Gets version",
+			Args: []string{"version"},
+		},
+		{
+			Name:             "view the channel to test if meshconfig is correctly created & loaded",
+			Args:             []string{"system", "channel", "view"},
```
nishantwrp, line 45: Used this command i.e `mesheryctl system channel view` to test configs as it's output is simple and short. Any other better alternatives?

alphaX86 (commented):
Ok, this is an integration test and it's great. I though have a suggestion, continuing the convo from slack, it's noticed that both config, tar.gz and manifests are being created... it's ok since they are created via `system` commands. But we need to delete them after the tests get over, since we didn't want to leave a trace.

@nishantwrp could you add the above said requirement into the test? Other than that, this test looks pretty neat to me

nishantwrp:
@alphaX86 updated the code to remove all files it generates while testing

mesheryctl/internal/cli/root/root_test.go
leecalcote: Good one. This is as short and sweet as it gets, I believe.

leecalcote (approved):

PR #5963, opened 2022-08-04
[UI] Import Modal update and reference to Extensions 

Chadha93:
Signed-off-by: Gaurav Chadha <gauravchadha1676@gmail.com>
**Description**
Import Modal update and reference to Extensions keeping the UX as desired.

This PR fixes #
in reference to https://github.com/layer5labs/meshery-extensions/pull/1087

**Notes for Reviewers**

https://user-images.githubusercontent.com/65453826/182917825-0a49230b-e854-4ecd-8976-83cc9d18211b.mp4

### DIFF
--- a/docs/pages/reference/mesheryctl/mesheryctl-app-list.md
+++ b/docs/pages/reference/mesheryctl/mesheryctl-app-list.md
@@ -58,7 +58,7 @@ mesheryctl app list
 ## Screenshots
 
 Usage of mesheryctl app list
-![app-list-usage](../../../../docs/assets/img/mesheryctl/app-list.png)
+![app-list-usage](/assets/img/mesheryctl/app-list.png)
 
 ## See Also
 
--- a/docs/pages/reference/mesheryctl/mesheryctl-app-onboard.md
+++ b/docs/pages/reference/mesheryctl/mesheryctl-app-onboard.md
@@ -75,7 +75,7 @@ mesheryctl app onboard -f ./application.yml -s "Kubernetes Manifest"
 ## Screenshots
 
 Usage of mesheryctl app onboard
-![app-onboard-usage](../../../../docs/assets/img/mesheryctl/app-onboard.png)
+![app-onboard-usage](/assets/img/mesheryctl/app-onboard.png)
 
 ## See Also
 
--- a/docs/pages/reference/mesheryctl/mesheryctl-app-view.md
+++ b/docs/pages/reference/mesheryctl/mesheryctl-app-view.md
@@ -76,7 +76,7 @@ mesheryctl app view --all
 ## Screenshots
 
 Usage of mesheryctl app view
-![app-view-usage](../../../../docs/assets/img/mesheryctl/app-view.png)
+![app-view-usage](/assets/img/mesheryctl/app-view.png)
 
 ## See Also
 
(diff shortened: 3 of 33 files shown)


ui/components/MesheryApplications.js
```diff
@@ -783,7 +802,18 @@ function MesheryApplications({
         <div className={classes.topToolbar} >
           {!selectedApplication.show && (applications.length>0 || viewType==="table") && <div className={classes.createButton}>
             <div>
-              <UploadImport supportedTypes={types} isApplication = {true} aria-label="URL upload button" handleUrlUpload={urlUploadHandler} handleUpload={uploadHandler} configuration="Application"  />
+              <Button
+                aria-label="Add Pattern"
```
Chadha93, line 806:
```suggestion
                aria-label="Add Application"
```

ui/components/Filters.js
```diff
@@ -694,7 +713,18 @@ function MesheryFilters({ updateProgress, enqueueSnackbar, closeSnackbar, user,
         <div className={classes.topToolbar} >
           {!selectedFilter.show && (filters.length>0 || viewType==="table") && <div className={classes.createButton}>
             <div>
-              <UploadImport supportedTypes="null" aria-label="URL upload button" handleUrlUpload={urlUploadHandler} handleUpload={uploadHandler} configuration="Filter" />
+              <Button
+                aria-label="Add Pattern"
```
Chadha93, line 717:
```suggestion
                aria-label="Add Filter"
```

Harshit-Dandriyal:
woah, So, you Separated <UploadImport> and used buttons instead of that, Great but it's creating redundancy in code, I mean we should at least make sure that these buttons don't get repeated. I suggest making a component at least for buttons as they are even also similar in design.

Chadha93:
> woah, So, you Separated UploadImport and used buttons instead of that, Great but it's creating redundancy in code, I mean we should at least make sure that these buttons don't get repeated. I suggest making a component at least for buttons as they are even also similar in design.

Yes, I'm with you, this was a fix we could go, for now, in the configurations code (filter/apps/patterns) there are lots of more lines of code causing redundancy. It is a very long process to remove all of it. We need to make sure all of these should not occur in refactoring. cc @YashKamboj

leecalcote:
```
AssertionError: Timed out retrying after 4000ms: Expected to find element: `[data-cy="import-button"]`, but never found it.```

**Comment by Chadha93** [comment]
All the tests here are queued.... What could be the issue?? The runner is not picking up the jobs.

**Comment by leecalcote** [comment]
@Chadha93 looks like it finally ran:

```
AssertionError: Timed out retrying after 4000ms: Expected to find element: `[data-cy="import-button"]`, but never found it.
```

Abhishek-kumar09:
SO filters test needs to be updated to pass the integration test. Seems like the api request is not sent on file upload. Any changes in that area by you? Previously, the import actions directly runs after selecting browse, probably you have progressed it make it better? that may be reason of fail.

leecalcote:
Still failing - https://github.com/meshery/meshery/runs/7772269872?check_suite_focus=true

MarioArriaga92:
@Chadha93 @leecalcote taking a look at failing test right now

leecalcote:
Thank you. :)

Chadha93:
Yay! All tests passed. Thank you @MarioArriaga92 @Abhishek-kumar09 
Merging this.

MarioArriaga92 (commented):
LGTM

PR #6134, opened 2022-08-31
UI: Fix icons alignment in header

Nikhil-Ladha:

**Description**
The icons seemed a bit off, so added a touch of CSS.
Subtle but irritating

Before:
Screenshot from 2022-08-31 13-10-24

Now:
Screenshot from 2022-08-31 13-10-13

**Notes for Reviewers**

### DIFF
--- a/ui/components/Header.js
+++ b/ui/components/Header.js
@@ -291,7 +291,7 @@ function K8sContextMenu({
           ? 'menu-list-grow'
           : undefined}
         aria-haspopup="true"
-        style={{ marginRight : "0.5rem", top : "0.25rem" }}
+        style={{ marginRight : "0.5rem" }}
       >
         <div className={classes.cbadgeContainer}>
           <img className="k8s-image" src="/static/img/kubernetes.svg" width="24px" height="24px" style={{ zIndex : "2" }} />


leecalcote:
Oh, yes, hmm. The feeling you get here is because of the badge on the K8s logo. The thing is that the logos should all be the same size, while the badge is extra and as such might hang past other logos without badges.  

Do this, if you would: look at other systems (phones, macOS, others) that have badges and compare. Here's a screenshot from my system:
Screen Shot 2022-08-31 at 7 34 04 AM

Nikhil-Ladha:
Hmm....makes sense.
Though there was a slight issue with the overall icons set and for that the `align-items: center` is still needed.

ui/components/Header.js
```diff
@@ -289,7 +291,7 @@ function K8sContextMenu({
           ? 'menu-list-grow'
           : undefined}
         aria-haspopup="true"
-        style={{ marginRight : "0.5rem" }}
+        style={{ marginRight : "0.5rem", top : "0.25rem" }}
```
Nikhil-Ladha, line 294:
```suggestion
        style={{ marginRight : "0.5rem" }}
```

leecalcote (approved):
LGTM 👍

PR #6153, opened 2022-09-01
Add support for dynamic higher order component generation for imported designs via dynamic schemas

Revolyssup:
**Description**

This PR is first step in direction of dynamic schemas and dynamic components. Meshery server can now override schemas on for components based on `type`
This is a helpful feature where `type` is `$(#use ...)` and we are importing another pattern/design.
In future we can have even more sophisticated logic to control this dynamic schema creation based on other things.
**Notes for Reviewers**


leecalcote:
@Revolyssup Application Source Type as a complete feature is _still_ yet to be delivered. Let's see other work completed before taking on items that are not defined in the backlog, yet.

leecalcote:
I realize that this was put together hastily, but we do want to refactor the SVG to only include the logo (unless we are still using an SVG, not a CSS system for the Details pane).

Abhishek-kumar09:
@leecalcote Need for review.

spec file: https://docs.google.com/document/d/1B2N78EdRiZF-yVo1-HY3syppwBBDumgMuYg6seD-AJ4/edit#heading=h.ry753nvepfyk

Aisuko (commented):
We need some test cases here that would be good.

server/models/pattern/core/pattern.go
```diff
@@ -149,6 +152,14 @@ func (p *Pattern) GenerateApplicationConfiguration() (v1alpha1.Configuration, er
 func (p *Pattern) GetServiceType(name string) string {
 	return p.Services[name].Type
 }
+func (p *Pattern) GenerateDynamicSchema() {
+	for _, svc := range p.Services {
+		if strings.HasPrefix(svc.Type, "$(#use ") { //Currently pattern import is supported only via URL
```
Aisuko, line 157: The prefix string of `strings.HasPrefix(s, prefix string) bool` has a blank at the last. The next line of code does not have. So, here I suggest that we define a new variable as the parameters. It can avoid this situation.

PR #6165, opened 2022-09-02
chore: Updated titles in all pages for consistency

Nikhil-Ladha:

**Description**
Updated the titles on pages to maintain consistency throughout the UI.
Also, updated the drawer behavior for Meshery UI menu extensions.

### DIFF
--- a/ui/components/Navigator.js
+++ b/ui/components/Navigator.js
@@ -547,7 +547,6 @@ class Navigator extends React.Component {
         commitsha : "",
         release_channel : "NA",
       },
-      // drawerSetForMeshmap : false,
     };
   }
 
@@ -592,35 +591,7 @@ class Navigator extends React.Component {
       },
       (err) => console.error(err)
     );
-
-    // // Collapse the drawer, on meshmap page to provide more room for the canvas
-    // const path = window.location.pathname;
-    // if (path.includes("/extension/meshmap")) {
-    //   this.drawerCollapsedForMeshmap();
-    // }
   }
-
-  // // Collapse the drawer, on meshmap page to provide more room for the canvas
-  // componentDidUpdate() {
-  //   const path = window.location.pathname;
-  //   const { isDrawerCollapsed } = this.props;
-  //   const { drawerSetForMeshmap } = this.state;
-  //   if (!drawerSetForMeshmap && !isDrawerCollapsed && path.includes("/extension/meshmap")) {
-  //     this.drawerCollapsedForMeshmap();
-  //   } else if (drawerSetForMeshmap && !path.includes("/extension/meshmap")) {
-  //     // Update state if the page is changed, to re-collapse on next visit
-  //     this.setState({
-  //       drawerSetForMeshmap : false
-  //     })
-  //   }
-  // }
-
-  // drawerCollapsedForMeshmap() {
-  //   this.toggleMiniDrawer(false);
-  //   this.setState({
-  //     drawerSetForMeshmap : true
-  //   })
-  // }
 
   /**
    * @param {import("../utils/ExtensionPointSchemaValidator").NavigatorSchema[]} children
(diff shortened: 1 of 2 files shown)


Nikhil-Ladha:
@Abhishek-kumar09 I have added the code in this PR to handle the drawer behavior on meshmap page. Please do review it :)

ui/components/Navigator.js
```diff
@@ -591,6 +592,34 @@ class Navigator extends React.Component {
       },
       (err) => console.error(err)
     );
+
+    // Collapse the drawer, on meshmap page to provide more room for the canvas
+    const path = window.location.pathname;
+    if (path.includes("/extension/meshmap")) {
+      this.drawerCollapsedForMeshmap();
+    }
```
Abhishek-kumar09, line 600: 🤝

ui/components/Navigator.js
```diff
@@ -591,6 +592,34 @@ class Navigator extends React.Component {
...
+  }
+
+  // Collapse the drawer, on meshmap page to provide more room for the canvas
+  componentDidUpdate() {
+    const path = window.location.pathname;
+    const { isDrawerCollapsed } = this.props;
+    const { drawerSetForMeshmap } = this.state;
+    if (!drawerSetForMeshmap && !isDrawerCollapsed && path.includes("/extension/meshmap")) {
```
Abhishek-kumar09, line 608: `drawerSetForMeshMap` is this variable providing any value to us? I don't see its requirement or am I missing the logic here

Abhishek-kumar09 (commented):
🚀

Abhishek-kumar09 (approved):
btw, I had only 1 query rest of the PR looks great as @Nikhil-Ladha himself.

leecalcote:
> btw, I had only 1 query rest of the PR looks great as @Nikhil-Ladha himself.

Good, good.

leecalcote (approved):

ui/components/Navigator.js
Nikhil-Ladha: Yes, there is.
Consider these use cases:
1. The user directly opens the extension page, at that time the whole page would be rendered and  `compomemntDidMount` will be called. So, the drawer will get collapsed.
2. The user reloads the page, then similar situation as above.
3. The user starts at Dashboard i.e, `localhost:9081` and then navigated to extensions page. This time the drawer is already mounted so the `componentDidMount` will not run, but we want the drawer to get collapsed when the user goes to meshmap extension page. So, the `componentDidUpdate` is needed.
4. Now, the user navigates away to some other page and comes back, then we would still want the drawer to collapse.

All the above scenario could be covered without the extra state. But, now conisder the user wants to expand/collapse the drawer manually when on the meshmap extensions page. They won't be able to because on doing so a re-render would happen and the isDrwerCollapsed state would get updated and that would cause the whole collapsing on meshmap page to go on an infinite loop.
So, to prevent that we need another state to keep track that this is the first/returning-first instance of the user to meshmap page and collapse the drawer. After that whatever user does (expand/collapse) on the page is upto them.

A bit long explanation, but exactly printed what I went through while writing the code yesterday 😂

Abhishek-kumar09: hmmm, btw great explanation, that was the reason I was talking about putting it in useEffect hook, I meant the extension code that time. In this code: https://github.com/meshery/meshery/blob/master/ui/components/ExtensionSandbox.js

the drawer-expand state is saved as redux variable, so no prop drilling is required. and the use-case is pretty forward without any extra logic. 
1. load the extension-sandbox component: close the drawer.
2. this removes the overhead from navigator.js to manage any collapse event on any button-click other than that (expnand-collapse) button.

But anyways, this is also good.

Nikhil-Ladha: Ok, that sounds like a better one.
Though I tried that logic and it doesn't seem to work for some reason. Can't seem to figure out the issue, do you mind taking a look?

I have pushed the probable code change for it to the PR, let me know what am I missing here 🤔

Nikhil-Ladha: NVM, fixed the issue. I forgot to pass the toggle to props :p

Everything works fine now.

Nikhil-Ladha:
Can you please do a second round of review @Abhishek-kumar09 ?

Nikhil-Ladha:
@Abhishek-kumar09 I think we are safe to bypass the integration-tests and merge the PR? 🤔

Chadha93:
> @Abhishek-kumar09 I think we are safe to bypass the failing integration-tests and merge the PR? thinking

+1

leecalcote (approved):
LGTM

PR #6223, opened 2022-09-17
[bug] Fix Validation of PatternFile

theBeginner86:
**Description**
Prettification of the keys in user pattern file wasn't performed before validation so the system ceases to match with the stored prettified schema definition.
Hence, it was leading to error during provisioning of a Pattern.
error:
```
invalid workload definition: invalid settings: [/containers/0/ports/0: {"containerPort":908... "container Port" value is required]
```

Eg: While storing definition for a pattern file, a key: `containerPort` was stored as `container Port` (due to prettification) but user pattern still has `containerPort` (due to no prettification before validation) so the key was not matching with the definition schema. Hence, leading to error on deploying.

This also lead to failing workflows for Istio adapter ([check](https://docs.meshery.io/project/compatibility-matrix/meshery-istio-past-results)). [Check the logs](https://github.com/meshery/meshery-istio/actions/runs/2995210586/jobs/4953360577#step:14:33) of last failing test for [TestServiceMeshAndSampleApp](https://github.com/meshery/meshery-istio/actions/runs/2995210586/jobs/4953360577).

After this update, Meshery Isito adapter tests should also pass.

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/server/models/pattern/stages/validate.go
+++ b/server/models/pattern/stages/validate.go
@@ -9,7 +9,6 @@ import (
 	"github.com/layer5io/meshery/server/models/pattern/jsonschema"
 	"github.com/layer5io/meshery/server/models/pattern/patterns/k8s"
 	"github.com/layer5io/meshery/server/models/pattern/resource/selector"
-	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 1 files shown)


leecalcote (approved):
Thank you, @theBeginner86 👏

PR #6237, opened 2022-09-23
Implement websocket signalling for collaboration

nishantwrp:
Signed-off-by: Nishant Mittal <mittalnishant14@outlook.com>

**Description**
Future Todo: Currently there is no authentication in this endpoint. 

This PR fixes #

**Notes for Reviewers**


server/handlers/pattern_collab_handler.go
```diff
@@ -0,0 +1,255 @@
...
+	patterns                map[string]map[*Client]bool
+	broadcast               chan *SocketMessage
+	register                chan *Client
+	unregister              chan *Client
+	updateSubscribedPattern chan *PatternUpdateRequest
+}
+
+type SocketMessage struct {
+	Type   string      `json:"type"`
+	Topics []string    `json:"topics,omitempty"`
```
Revolyssup, line 46: what is the difference between `topics` and `topic` here?

server/handlers/pattern_collab_handler.go
```diff
@@ -0,0 +1,255 @@
...
+
+var mainHub *Hub = nil
+
+func serveWs(hub *Hub, w http.ResponseWriter, r *http.Request) {
+	conn, err := upgrader.Upgrade(w, r, nil)
+	if err != nil {
+		return
+	}
+
+	client := &Client{hub: hub, conn: conn, patternID: "", send: make(chan []byte), connected: true}
```
Revolyssup, line 242: why is patternid empty here?

server/handlers/pattern_collab_handler.go
```diff
@@ -0,0 +1,255 @@
...
+
+		c.processMessage(message)
+	}
+}
+
+func (c *Client) writeMessages() {
+	ticker := time.NewTicker(pingPeriod)
+	defer func() {
+		ticker.Stop()
+		c.hub.unregister <- c
```
Revolyssup, line 115: This same client was send to unregister channel when readMessages() returns. Also the connection is closed when the readMessages() returns. There will be a send() on closed channel. Do both these things at one place.

Revolyssup (changes requested):

Revolyssup:
@nishantwrp  All channels in hub are unbufferred. This might become a bottleneck so make these channels unbufferred unless some synchronization is required.

server/handlers/pattern_collab_handler.go
nishantwrp: It is later sent by client when the type of message is "subscribe".

server/handlers/pattern_collab_handler.go
nishantwrp: done!

nishantwrp:
@Revolyssup got it. Made all the channels buffered. I don't think the communications need to be synchronous. I am unsure of the idea buffer length, so have made it `10` for hub channels and `100` for the client send channel. Let me know, if I should update that.

leecalcote:
@nishantwrp when will this be ready for final review?

nishantwrp:
@leecalcote it is ready for final review. // @Revolyssup

leecalcote:
Syncing with `master`

leecalcote:
Please update the REST API Endpoint spreadsheet and the REST API docs in Meshery Docs. (use make command)

PR #6244, opened 2022-09-23
Fix: GraphQL subscription bottleneck connecting to the server

Abhishek-kumar09:

**Description**
Fix GrapqhQL subscription connection loss due to client-side bottleneck
Screenshot 2022-09-24 at 12 40 12 AM

## Before

https://user-images.githubusercontent.com/48255244/192040826-277a7583-100f-43f1-913e-490c8bc4b394.mov

## After

https://user-images.githubusercontent.com/48255244/192040984-b9fcce6b-a986-40ec-a381-71e1c8684929.mov

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/ui/lib/relayEnvironment.js
+++ b/ui/lib/relayEnvironment.js
@@ -21,13 +21,13 @@ export let subscriptionClient;
 if (typeof window !== "undefined"){
   subscriptionClient = new SubscriptionClient("ws://"+window.location.host+"/api/system/graphql/query", {
     reconnect : true,
+    minTimeout: 4000
   });
 
 }
 
 function setupSubscription(config, variables, cacheConfig, observer) {
   const query = config.text;
-
 
   const subscribeObservable = subscriptionClient.request({ query, variables }, (error, result) => {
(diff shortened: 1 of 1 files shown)


ui/lib/relayEnvironment.js
```diff
@@ -21,14 +21,14 @@ export let subscriptionClient;
 if (typeof window !== "undefined"){
   subscriptionClient = new SubscriptionClient("ws://"+window.location.host+"/api/system/graphql/query", {
     reconnect : true,
+    minTimeout: 4000
```
leecalcote, line 24: This doesn't address root cause of the issue, does it? 

Why is the client choking?

Abhishek-kumar09: Yes, this might be the root cause. The connection closes itself even before it acknowledge the server response, take it as a relay bug or anything else, this has been the reason towards all the *subscription initialisation* bugs, now the operator and meshsync subscriptions updates and rest of them are also caught on fire with speed.

PR #6258, opened 2022-09-27
[Server] Backend for catalog for local provider

gr455:
**Description**
Adds backend for catalog capability to local provider.

### DIFF
--- a/server/models/meshery_filter_persister.go
+++ b/server/models/meshery_filter_persister.go
@@ -85,7 +85,7 @@ func (mfp *MesheryFilterPersister) CloneMesheryFilter(filterID string) ([]byte, 
 	filterUUID, _ := uuid.FromString(filterID)
 	err := mfp.DB.First(&mesheryFilter, filterUUID).Error
 	if err != nil || *mesheryFilter.ID == uuid.Nil {
-		return nil, fmt.Errorf("Unable to get filter: %w", err)
+		return nil, fmt.Errorf("unable to get filter: %w", err)
 	}
 
 	id, err := uuid.NewV4()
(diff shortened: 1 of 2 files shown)


leecalcote:
@Revolyssup feedback on this one?

leecalcote:
@gr455 has is this explicitly making local provider content public?

gr455:
> @gr455 has is this explicitly making local provider content public or is that for another PR?

This does make the local provider content public. This means the seeded content is seeded as public

leecalcote:
@Revolyssup @theBeginner86 a review here will be helpful.

theBeginner86 (commented):
Looking good!

Though cloning a public pattern/filter doesn't seem to work. Will that be added separately?

https://user-images.githubusercontent.com/73700530/193433762-18212431-84b9-4bc8-9970-120cf97868e4.mov

gr455:
> Looking good!
> 
> Though cloning a public pattern/filter doesn't seem to work. Will that be added separately?
> 
>  Screen.Recording.2022-10-02.at.6.38.27.AM.mov

Coming right up!

theBeginner86 (approved):
Thanks @gr455 for quick fix!

LGTM!

server/models/meshery_filter_persister.go
```diff
@@ -54,6 +56,49 @@ func (mfp *MesheryFilterPersister) GetMesheryFilters(search, order string, page,
...
+func (mfp *MesheryFilterPersister) GetMesheryCatalogFilters(search, order string) ([]byte, error) {
+	order = sanitizeOrderInput(order, []string{"created_at", "updated_at", "name"})
+
+	if order == "" {
+		order = "updated_at desc"
+	}
+
+	filters := []*MesheryFilter{}
+
+	query := mfp.DB.Where("visibility = 'public'").Order(order)
```
Revolyssup, line 69: We have the concept of public for local provider as well? We can lose the visibility filter because on Meshery instance level. I dont think we need to care about public and private.

Revolyssup: We would be storing with visibility as `private` for all in local provider so I think we can lose the `public` from here.

Revolyssup (dismissed):

PR #6263, opened 2022-09-27
Make core workflow more general

Revolyssup:

**Description**

This PR makes core E2E workflow more general so that it doesn't need an adapter to run always

**Notes for Reviewers**

### DIFF
--- a/.github/workflows/test_adaptersv2.yaml
+++ b/.github/workflows/test_adaptersv2.yaml
@@ -152,22 +152,19 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: patternfile
-      # This pattern will try to install  service mesh
-      - name: Pattern apply (Install service mesh)
+      - name: Wait for Meshery pods to come up 
+        run: sleep 100  
+      # This pattern will try to install  service mesh(the first attempt can fail currently fot k8s componets so we will wait 100 seconds after the first apply and then apply again. Since pattern apply is idempotent, this wont be an issue.)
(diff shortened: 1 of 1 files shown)


.github/workflows/test_adaptersv2.yaml
```diff
@@ -0,0 +1,258 @@
+# This workflow will be called/referenced in each of the adapters and runs in the context of adapter
+name: Adapter Integration - Core v2
```
leecalcote, line 2: Oh, we're versioning now. Fancy. 😉

.github/workflows/test_adaptersv2.yaml
```diff
@@ -0,0 +1,258 @@
...
+    secrets:
+      token:
+        required: true
+
+jobs:
+  TestAdapter:
+    strategy:
+      fail-fast: false
+      matrix:
+        k8s_version: ['v1.20.11', 'v1.21.5', 'v1.22.2','v1.23.9']
```
leecalcote, line 57: No, 1.24 or 1.25?

Revolyssup: For now, due to problems in other adapters. We can keep this to 1.23.

leecalcote: @Revolyssup the issue here is that minikube and docker-desktop are both already running v1.25. We need to get to 1.25 now. What's preventing us from moving forward?

PR #6264, opened 2022-09-27
Add end to end tests for kubernetes components

Revolyssup:
**Description**

This PR adds test via emojivoto app

**Notes for Reviewers**


Revolyssup:
@leecalcote  I haven't added the UpdateDocs step until we figure out the UI on compatibility matrix for K8s yamls with no service mesh adapters.

leecalcote:
I wonder how necessary this testing is, @Revolyssup.

Revolyssup:
> I wonder how necessary this testing is, @Revolyssup.

@leecalcote  we can add a whole bunch of designs from public facing catalogues in these tests. The tests will then dynamically fetch that content and run it on multiple k8s to see which designs are compatible with what k8s version. We can use results from here to directly show under the status of each design in the catalogue (status in the sense,which k8s version these are compatible with).
For instance, any design having v1beta as apiVersion of CRD after k8s 1.22 will break and we can catch stuff like that. This is also in part with the conversation we had with @halcyondude, where we discussed how it is useful to have some mechanism which can notify you if your deployments/configurations break with a k8s version change.

Revolyssup:
> One question, some of these templates for patterns are also there in service-mesh-patterns. Shouldn't we have like one source of truth for all pattern files unless required? @Revolyssup

Yes you are right. This PR just used this new emoji.yaml as example, as explained in my above comment. The end goal is to source them from a single place, the same place from where we will also source for catalogue.

leecalcote:
@Revolyssup what would you like to do here?

PR #6323, opened 2022-10-04
[mesheryctl] improve the error messages shown to users for "mesheryctl pattern"

### LINKED ISSUE #6271: [mesheryctl] improve the error messages shown to users for "mesheryctl pattern"
**Description**: Improve the error messages in mesheryctl commands from the user's perspective. 
**Commands in scope of this issue**: `mesheryctl pattern`, `mesheryctl pattern list`, `mesheryctl pattern apply`, `mesheryctl pattern view`, `mesheryctl pattern delete`

---

We need to make sure of the following things
- There are proper error messages when users
  - Give more arguments to a command than required
  - Misspell a command
  - Do not give the required input via flags or args
  - Use the command without any arguments or flags.
- The error messages being displayed to users are meaningful from user's perspective. And if applicable, a potential fix for the problem should be shown too.  For example
  - "failed to unmarshal response body" could be modified to "couldn't process JSON response from server"
  - "manifest not found" could be modified to "manifest not found. Please provide the path to the app manifest."
  
 --- 
  #### Contributor Resources

For additional context, head over to this [pr](https://github.com/meshery/meshery/pull/5661).

kamalbuilds:
**Description**

This PR fixes #6271 

**Notes for Reviewers**
1> A notable case considered which is resolved in this PR is 
image

`mesheryctl pattern view` assumes a name if ID doesn't match so that case is covered in .

2> Authentification fix via CLI is done.
3>Unmarshall code is reframed from user's perspective.

### DIFF
--- a/mesheryctl/internal/cli/root/pattern/delete.go
+++ b/mesheryctl/internal/cli/root/pattern/delete.go
@@ -42,7 +42,7 @@ mesheryctl pattern delete [file | URL]
 		if !govalidator.IsURL(file) {
 			content, err := os.ReadFile(file)
 			if err != nil {
-				return errors.New(utils.PatternError(fmt.Sprintf("failed to read file %s Ensure the file name or URL is correct", file)))
+				return errors.New(utils.PatternError(fmt.Sprintf("failed to read file %s. Ensure the filename or URL is valid", file)))
 			}
 
 			patternFile = string(content)
@@ -97,7 +97,7 @@ mesheryctl pattern delete [file | URL]
 
 			err = json.Unmarshal(body, &response)
 			if err != nil {
-				return errors.Wrap(err, "couldn't process JSON response from server")
+				return errors.Wrap(err, "couldn't process JSON response from Meshery Server")
 			}
 
 			patternFile = response[0].PatternFile
--- a/mesheryctl/internal/cli/root/pattern/view.go
+++ b/mesheryctl/internal/cli/root/pattern/view.go
@@ -57,7 +57,7 @@ mesheryctl pattern view [pattern-name | ID]
 			if viewAllFlag {
 				url += "/api/pattern?page_size=10000"
 			} else {
-				return errors.New("[pattern-name|pattern-id] is not specified, use -a to view all patterns")
+				return errors.New("Pattern name or ID is not specified. Use `-a` to view all patterns")
 			}
 		} else if isID {
 			// if pattern is a valid uuid, then directly fetch the pattern
@@ -90,7 +90,7 @@ mesheryctl pattern view [pattern-name | ID]
 
 		var dat map[string]interface{}
 		if err = json.Unmarshal(body, &dat); err != nil {
-			return errors.Wrap(err, "couldn't process JSON response from server")
+			return errors.Wrap(err, "couldn't process JSON response from Meshery Server")
 		}
 
 		if isID {
@@ -106,7 +106,7 @@ mesheryctl pattern view [pattern-name | ID]
 			// use the first match from the result when searching by pattern name
 			arr := dat["patterns"].([]interface{})
 			if len(arr) == 0 {
-				utils.Log.Info(fmt.Sprintf("pattern with name: %s not found , please enter the correct [pattern name | ID] ", pattern))
+				utils.Log.Info(fmt.Sprintf("pattern with name: %s not found. Enter a valid pattern name or ID", pattern))
 				return nil
 			}
 			if body, err = json.MarshalIndent(arr[0], "", "  "); err != nil {
--- a/mesheryctl/internal/cli/root/system/error.go
+++ b/mesheryctl/internal/cli/root/system/error.go
@@ -63,7 +63,7 @@ func ErrCreateDir(err error, obj string) error {
 }
 
 func ErrUnmarshal(err error, obj string) error {
-	return errors.New(ErrUnmarshalCode, errors.Alert, []string{"Error processing JSON response from server", obj}, []string{err.Error()}, []string{}, []string{"Either the JSON response or the Response is distorted"})
+	return errors.New(ErrUnmarshalCode, errors.Alert, []string{"Error processing JSON response from Meshery Server", obj}, []string{err.Error()}, []string{}, []string{"Either the JSON response or the Response is distorted"})
 }
 
 func ErrUnsupportedPlatform(platform string, config string) error {
--- a/mesheryctl/pkg/utils/error.go
+++ b/mesheryctl/pkg/utils/error.go
@@ -128,7 +128,7 @@ func formatError(msg string, cmd cmdType) string {
 
 func ErrAttachAuthToken(err error) error {
 	return errors.New(ErrAttachAuthTokenCode, errors.Alert, []string{err.Error()},
-		[]string{"Authentication token not found. Please supply a valid user token using the --token (or -t) flag. Or login with `mesheryctl system login`"}, []string{}, []string{})
+		[]string{"Authentication token not found. Login with `mesheryctl system login` or supply a valid user token using the --token (or -t) flag."}, []string{}, []string{})
 }
 
 func ErrFailRequest(err error) error {


mesheryctl/internal/cli/root/system/error.go
```diff
@@ -63,7 +63,7 @@ func ErrCreateDir(err error, obj string) error {
 }
 
 func ErrUnmarshal(err error, obj string) error {
-	return errors.New(ErrUnmarshalCode, errors.Alert, []string{"Error unmarshalling file ", obj}, []string{err.Error()}, []string{}, []string{})
+	return errors.New(ErrUnmarshalCode, errors.Alert, []string{"Error processing JSON response from server to file", obj}, []string{err.Error()}, []string{}, []string{})
```
hershd23, line 66: `processing JSON response from server to file`

Seems a little weird if we aren't unmarshalling files. Is this only handling files? If not just omit `to file`

mesheryctl/pkg/utils/error.go
```diff
@@ -128,7 +128,7 @@ func formatError(msg string, cmd cmdType) string {
 
 func ErrAttachAuthToken(err error) error {
 	return errors.New(ErrAttachAuthTokenCode, errors.Alert, []string{err.Error()},
-		[]string{"Authentication token not found. please supply a valid user token with the --token (or -t) flag. or login with `mesheryctl system login`"}, []string{}, []string{})
+		[]string{"Authentication token not found. please supply a valid user token using `mesheryctl system config [aks|eks|gke|minikube] -t <path>`. or login with `mesheryctl system login`"}, []string{}, []string{})
```
hershd23, line 131: I think we have the token flag for quite a few methods. Have we changed that? This seems a generic error type, not specific to `mesheryctl system config`

Also not sure if we should delete the `--token (or -t)` suggestion

hershd23 (commented):
A couple of comments. Let me know if my understanding about some of these is incorrect

mesheryctl/pkg/utils/error.go
legendarykamal: the idea here was not to delete the --token flag but to provide the full command i.e. `mesheryctl system config  [aks|eks|gke|minikube] -t <path>` which inturn will help the user in providing the token via this command. 

so the --token suggestion is not deleted but is intact to the helping command.

hershd23: So your command is an example right? Like this auth error comes on other commands as well right. Is it right to just specify one example?

legendarykamal: An example or a case, the case here considered is 
user is trying to list the pattern using `mesheryctl pattern list` but he is not authenticated. 
So a token is being asked from him to supply via the `mesheryctl system config [aks|eks|gke|minikube] -t <path>` cmd  or login with `mesheryctl system login` .

hershd23: I don't think a specific case should be mentioned for a generic authentication error type. 

@revolyssup

mesheryctl/internal/cli/root/system/error.go
legendarykamal: done 👍

mesheryctl/pkg/utils/error.go
legendarykamal: Ok @alphaX86 @hershd23 thank you for the feedback, I will change this accordingly 🙂 no sweat.

legendarykamal:
@nishantwrp could you please review this PR and tell If any modification are needed or this is good to go? 

@hershd23 review has been addressed succerssfully.

hershd23:
For most part changes suggested here and in #6356 replace `unmarshall` with couldn't process JSON. 

@legendarykamal @XDRAGON2002 if you can decide on one string that would be better. @alphaX86 do you have any ideas/what the default error string for unmarshalling should be?

alphaX86:
> For most part changes suggested here and in #6356 replace `unmarshall` with couldn't process JSON.
> 
> @legendarykamal @XDRAGON2002 if you can decide on one string that would be better. @alphaX86 do you have any ideas/what the default error string for unmarshalling should be?
> 
> EDIT:- Actually they are making changes to the same files. Can we standardize the error messages?
> 
> Otherwise I don't have any issues with this. Once this is resolved, i'll approve

Well, for unmarshal, either we can mention as can't process json data or couldn't convert byte to json. It depends on our way of converting error to users

legendarykamal:
@alphaX86 @hershd23  both of us are using can't process json data , should we replace it with couldn't convert byte to json. 

PS:- First approach sounds better than the 2nd in my opinion .

mesheryctl/internal/cli/root/pattern/view.go
```diff
@@ -106,7 +106,7 @@ mesheryctl pattern view [pattern-name | ID]
 			// use the first match from the result when searching by pattern name
 			arr := dat["patterns"].([]interface{})
 			if len(arr) == 0 {
-				utils.Log.Info(fmt.Sprintf("pattern with name: %s not found", pattern))
+				utils.Log.Info(fmt.Sprintf("pattern with name: %s not found , please enter the correct [pattern name | ID] ", pattern))
```
leecalcote, line 109:
```suggestion
				utils.Log.Info(fmt.Sprintf("pattern with name: %s not found. Enter a valid pattern name or ID", pattern))
```

mesheryctl/internal/cli/root/system/error.go
```diff
@@ -63,7 +63,7 @@ func ErrCreateDir(err error, obj string) error {
 }
 
 func ErrUnmarshal(err error, obj string) error {
-	return errors.New(ErrUnmarshalCode, errors.Alert, []string{"Error unmarshalling file ", obj}, []string{err.Error()}, []string{}, []string{})
+	return errors.New(ErrUnmarshalCode, errors.Alert, []string{"Error processing JSON response from server", obj}, []string{err.Error()}, []string{}, []string{"Either the JSON response or the Response is distorted"})
```
leecalcote, line 66:
```suggestion
	return errors.New(ErrUnmarshalCode, errors.Alert, []string{"Error processing JSON response from Meshery Server", obj}, []string{err.Error()}, []string{}, []string{"Either the JSON response or the Response is distorted"})
```

mesheryctl/internal/cli/root/pattern/view.go
```diff
@@ -90,7 +90,7 @@ mesheryctl pattern view [pattern-name | ID]
 
 		var dat map[string]interface{}
 		if err = json.Unmarshal(body, &dat); err != nil {
-			return errors.Wrap(err, "failed to unmarshal response body")
+			return errors.Wrap(err, "couldn't process JSON response from server")
```
leecalcote, line 93:
```suggestion
			return errors.Wrap(err, "couldn't process JSON response from Meshery Server")
```

mesheryctl/internal/cli/root/pattern/view.go
```diff
@@ -57,7 +57,7 @@ mesheryctl pattern view [pattern-name | ID]
 			if viewAllFlag {
 				url += "/api/pattern?page_size=10000"
 			} else {
-				return errors.New("[pattern-name|pattern-id] not specified, use -a to view all patterns")
+				return errors.New("[pattern-name|pattern-id] is not specified, use -a to view all patterns")
```
leecalcote, line 60:
```suggestion
				return errors.New("Pattern name or ID is not specified. Use `-a` to view all patterns")
```

mesheryctl/pkg/utils/error.go
```diff
@@ -128,7 +128,7 @@ func formatError(msg string, cmd cmdType) string {
 
 func ErrAttachAuthToken(err error) error {
 	return errors.New(ErrAttachAuthTokenCode, errors.Alert, []string{err.Error()},
-		[]string{"Authentication token not found. please supply a valid user token with the --token (or -t) flag. or login with `mesheryctl system login`"}, []string{}, []string{})
+		[]string{"Authentication token not found. Please supply a valid user token using the --token (or -t) flag. Or login with `mesheryctl system login`"}, []string{}, []string{})
```
leecalcote, line 131:
```suggestion
		[]string{"Authentication token not found. Login with `mesheryctl system login` or supply a valid user token using the --token (or -t) flag."}, []string{}, []string{})
```

mesheryctl/internal/cli/root/pattern/delete.go
```diff
@@ -42,7 +42,7 @@ mesheryctl pattern delete [file | URL]
 		if !govalidator.IsURL(file) {
 			content, err := os.ReadFile(file)
 			if err != nil {
-				return errors.New(utils.PatternError(fmt.Sprintf("failed to read file %s", file)))
+				return errors.New(utils.PatternError(fmt.Sprintf("failed to read file %s Ensure the file name or URL is correct", file)))
```
leecalcote, line 45:
```suggestion
				return errors.New(utils.PatternError(fmt.Sprintf("failed to read file %s. Ensure the filename or URL is valid", file)))
```

leecalcote (approved):

PR #6387, opened 2022-10-14
[MeshModel] Add workflow and script to generate components | CU-3yvwk9h

humblenginr:
**Description**

This PR 
- Adds a script that generates two files: `component_models.yaml` and `components.yaml` which contain the system metadata and components generated from those systems respectively
- It uses utilities from meshkit to generate components

This script generates **~18500** components in around **15 minutes**. The concurrency parameters such as [this](https://github.com/meshery/meshery/pull/6387/files#diff-76d1b6fe5a5ca5822c004882b83a214e655a9e59311faf7411411d6fc2f04787R108 ) can be modified later after observing for some time to make it even more efficient. 

### DIFF
--- a/.github/workflows/generators.yml
+++ b/.github/workflows/generators.yml
@@ -1,4 +1,4 @@
-name: ArtifactHub Component Generation
+name: Static Component Generation
 on: 
   workflow_dispatch:
 jobs:
@@ -27,4 +27,4 @@ jobs:
       - name: Commit results
         uses: EndBug/add-and-commit@v9
         with: 
-          add: "*.json"
+          add: "*.yaml"
--- a/scripts/component_generation/go.mod
+++ b/scripts/component_generation/go.mod
@@ -1,4 +1,4 @@
-module github.com/sudo-NithishKarthik/meshmodel_ahub_gen
+module github.com/layer5io/meshmodel_ahub_gen
 
 go 1.19
 
--- a/scripts/component_generation/main.go
+++ b/scripts/component_generation/main.go
@@ -1,35 +1,32 @@
 package main
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"net/http"
+	"io"
 	"os"
 	"path/filepath"
 	"sync"
-	"time"
 
-	// "time"
+	"time"
 
 	"github.com/layer5io/meshkit/models/meshmodel/core/v1alpha1"
 	"github.com/layer5io/meshkit/utils/artifacthub"
+	"gopkg.in/yaml.v3"
 )
 
-const ArtifactHubApiEndpoint = "https://artifacthub.io/api/v1/"
-
 var (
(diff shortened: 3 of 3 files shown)


l5io:
Task linked: [CU-3yvwk9h Cronjob for pre-generating components in GitHub Action in meshery-extensions-packages](https://app.clickup.com/t/3yvwk9h)

scripts/component_generation/go.mod
```diff
@@ -0,0 +1,136 @@
+module github.com/sudo-NithishKarthik/meshmodel_ahub_gen
```
leecalcote, line 1: We'll need to relocate this package.

leecalcote (changes requested):

scripts/component_generation/main.go
```diff
@@ -0,0 +1,195 @@
+package main
```
leecalcote, line 1: `/scripts` - where is the logic for self-registering apps and filters located currently?

scripts/component_generation/main.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func getApiResponse(offset int, facet bool) (map[string]([]map[string]interface{}), error) {
+	reqUrl := fmt.Sprintf("%s?offset=%v&limit=%s&facets=%v&kind=%v&deprecated=false", AhSearchEndpoint, offset, Limit, facet, HELMKind)
+	resp, err := http.Get(reqUrl)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != 200 {
+		err = fmt.Errorf("status code %d for %s", resp.StatusCode, reqUrl)
```
leecalcote, line 185: MeshKit Error?

.github/workflows/generators.yml
```diff
@@ -0,0 +1,30 @@
+name: ArtifactHub Component Generation
```
leecalcote, line 1: Considering other repos in the future, should this be a generic title?

scripts/component_generation/go.mod
leecalcote, line 1: Will need to land this module elsewhere.

leecalcote (changes requested):
@Revolyssup please review and confirm that the same approach taken for static component generation in Meshery Adapters is being performed here. There should be no code duplication between static generation and dynamic generation of components (i.e. when a user imports a helm chart containing CRDs). 

Given this, I don't immediately consider that `/scripts` is an appropriate location for core component generation logic.

.github/workflows/generators.yml
sudo-NithishKarthik: Changed to `Static Component Generation`

scripts/component_generation/go.mod
sudo-NithishKarthik: moved to `layer5io/meshmodel_ahub_gen`

leecalcote (approved):

PR #6435, opened 2022-10-25
[server] Delete MeshSync data on deleting K8s Context

theBeginner86:
**Description**

1. This PR add support in the backend to delete related MeshSyn data on deleting a K8s context, considering its the last context pointing to a `cluster_id`.

2. Also, enhanced existing functionality of flushing MeshSyn data of a specific cluster. Earlier, it was just deleting objects related to that `cluster_id`, now it would delete `key_values`, `resource_spec`, `resource_status`,  and `resource_object_meta` as well.

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/ui/components/Header.js
+++ b/ui/components/Header.js
@@ -275,7 +275,7 @@ function K8sContextMenu({
     )
   }
 
-  const handleKubernetesDelete = (name, ctxId) => async () => {
+  const handleKubernetesDelete = (name, ctxId) => async() => {
     let responseOfDeleteK8sCtx = await deleteCtxtRef.current.show({
       title : `Delete ${name} context ?`,
       subtitle : `Are you sure you want to delete ${name} cluster from Meshery?`,
--- a/ui/components/MeshConfigComponent.js
+++ b/ui/components/MeshConfigComponent.js
@@ -110,6 +110,7 @@ function MesherySettingsNew({ classes, enqueueSnackbar, closeSnackbar, updatePro
   const [discover, setLastDiscover] = useState(['']);
   const [_operatorState, _setOperatorState] = useState(operatorState || []);
   const ref = useRef(null);
+  const deleteCtxtRef = useRef(null);
   const meshSyncResetRef = useRef(null);
   const _operatorStateRef = useRef(_operatorState);
   _operatorStateRef.current = _operatorState;
@@ -392,23 +393,32 @@ function MesherySettingsNew({ classes, enqueueSnackbar, closeSnackbar, updatePro
     }, variables);
   };
 
-  const handleConfigDelete = (id, index) => {
-    updateProgress({ showProgress : true });
-    dataFetch(
-      "/api/system/kubernetes/contexts/" + id,
-      {
-        credentials : "same-origin",
-        method : "DELETE"
-      },
-      () => {
-        updateProgress({ showProgress : false });
-        if (index != undefined) {
-          let newData = data.filter((dt, idx) => index != idx);
-          setData(newData);
-        }
-      },
-      handleError("failed to delete kubernetes context")
-    );
+  const handleConfigDelete = (id, name, index) => {
+    return async () => {
+      let responseOfDeleteK8sCtx = await deleteCtxtRef.current.show({
+        title : `Delete ${name} context ?`,
+        subtitle : `Are you sure you want to delete ${name} cluster from Meshery?`,
+        options : ["CONFIRM", "CANCEL"]
+      });
+      if (responseOfDeleteK8sCtx === "CONFIRM") {
+        updateProgress({ showProgress : true });
+        dataFetch(
+          "/api/system/kubernetes/contexts/" + id,
+          {
+            credentials : "same-origin",
+            method : "DELETE"
+          },
+          () => {
+            updateProgress({ showProgress : false });
+            if (index != undefined) {
+              let newData = data.filter((dt, idx) => index != idx);
+              setData(newData);
+            }
+          },
+          handleError("failed to delete kubernetes context")
+        );
+      }
+    }
   }
 
   const handleChange = () => {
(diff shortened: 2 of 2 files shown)


server/handlers/contexts_handler.go
```diff
@@ -82,6 +82,13 @@ func (h *Handler) DeleteContext(w http.ResponseWriter, req *http.Request, prefOb
 		return
 	}
 	h.config.K8scontextChannel.PublishContext()
+
+	err = models.FlushMeshSyncData(req.Context(), mux.Vars(req)["id"], provider)
```
Revolyssup, line 86: Can you make this asynchronous by putting this under a go routine and returning immediately?
Use eventStreamer instance to stream back the result/error of this function to the client.

Revolyssup: Basically do not return a 500 in case the meshsync data flush fails. If the context was deleted then it is 200. The rest of the information regarding data cleanup can be part of events.

server/internal/graphql/resolver/meshsync.go
```diff
@@ -218,7 +218,28 @@ func (r *Resolver) resyncCluster(ctx context.Context, provider models.Provider,
 			if provider.GetGenericPersister() == nil {
 				return "", ErrEmptyHandler
 			}
-			err := provider.GetGenericPersister().Where("cluster_id = ?", sid).Delete(&meshsyncmodel.Object{}).Error
+
+			err := provider.GetGenericPersister().Where("id IN (?)", provider.GetGenericPersister().Table("objects").Select("id").Where("cluster_id=?", sid)).Delete(&meshsyncmodel.KeyValue{}).Error
```
Revolyssup, line 222: Why to manually do this? Does the deletion in objects table alone not trigger the removal of rows from other tables with foreign key reference provided the gorm tags that we have?

server/models/k8s_context.go
```diff
@@ -329,3 +330,61 @@ func (kc *K8sContext) AssignServerID() error {
...
+	k8sctxs, ok := ctx.Value(AllKubeClusterKey).([]K8sContext)
+	if !ok || len(k8sctxs) == 0 {
+		return ErrEmptyCurrentK8sContext
+	}
+	var sid string
+	// Gets the serverID for the passed contextID
+	for _, k8ctx := range k8sctxs {
+		if k8ctx.ID == ctxID && k8ctx.KubernetesServerID != nil {
+			sid = k8ctx.KubernetesServerID.String()
+			break
```
Revolyssup, line 346: Instead of looping again, add that extra check and increment the reference counter here only

Revolyssup (changes requested):

server/handlers/contexts_handler.go
theBeginner86: makes sense, will make this change.

server/models/k8s_context.go
theBeginner86: gotcha, making change

server/internal/graphql/resolver/meshsync.go
theBeginner86: No, the deletion of `objects` table alone doesn't trigger deletion of rows in other tables with same `id`

server/models/k8s_context.go
```diff
@@ -329,3 +332,138 @@ func (kc *K8sContext) AssignServerID() error {
 
 	return nil
 }
+
+// FlushMeshSyncData will flush the meshsync data for the passed kubernetes contextID
+func FlushMeshSyncData(ctx context.Context, ctxID string, provider Provider, eb *events.EventStreamer) {
```
MUzairS15, line 337: We are dropping tables with meshsync data in the ` resyncCluster` resolver, only not streaming events (which we should stream). Why duplicate things, instead here https://github.com/meshery/meshery/pull/6435/files#diff-9640f3bd7dcd176aec6e4ee8412922560c18e16cd87c7f1e92ff6795dab9d64eR205
 we should call `FlushMeshSyncData`

server/internal/graphql/resolver/meshsync.go
MUzairS15: Changing `gorm` tag for `ONDELETE` to `cascade` should work. 

https://github.com/meshery/meshsync/blob/6f4eefabd09a9b8f0e8aa0919a00607a988141a5/pkg/model/model.go#L17

server/models/k8s_context.go
theBeginner86: yeah, good idea. I'll update that as well

server/internal/graphql/resolver/meshsync.go
```diff
@@ -218,7 +218,28 @@ func (r *Resolver) resyncCluster(ctx context.Context, provider models.Provider,
 			if provider.GetGenericPersister() == nil {
 				return "", ErrEmptyHandler
 			}
-			err := provider.GetGenericPersister().Where("cluster_id = ?", sid).Delete(&meshsyncmodel.Object{}).Error
+
```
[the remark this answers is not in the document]
leecalcote: Did you guys finalize on this, @theBeginner86?

leecalcote:
@theBeginner86 will you present this in today's Meshery Dev meeting? I think one more review will get this PR merged.

theBeginner86:
> @theBeginner86 will you present this in today's Meshery Dev meeting? I think one more review will get this PR merged.

Sure!

server/internal/graphql/resolver/meshsync.go
theBeginner86: @MUzairS15  I was trying it but there seems to be some problem as just adding that constraint didn't work. 

So currently I'm explicitly deleting the rows from other tables by matching their foreign key.
I would revisit this area while implementing the tombstone feature as that too requires a change in DB schema.

leecalcote (approved):
thank you for today's demo. LGTM

PR #6552, opened 2022-11-19
[CI] Decrease Golang requirements for Server builds

leecalcote:
### Description

This PR decreases the strictness of Golang version requirements for Meshery Server contributors (or any Golang aspect of this repo).

### Notes for Reviewers
1. Contributors are only required to have the appropriate MAJOR.MINOR version of Golang, not the .PATCH version, too; e.g. 1.19.x suffices as meeting the requirement.
1. Contributors are no longer required to have binaries available with matching Golang version number, but in a less restrictive way, need to simply have the required MAJOR.MINOR version number present in the output of `go version`.
1. Golang dependency checks are not implicit to the execution of any `make` command (.SILENT was not used). Instead, `dep-check` is included as a rule in every Go-related make target.
    1. UI contributors are not burdened with Golang requirements in this approach.

### Follow Up
1. Post-merge these same lightened restrictions will need to be propagated to all other Meshery-related repos.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -330,9 +330,14 @@ graphql-build: dep-check
 INSTALLED_GO_VERSION=$(shell go version)
 
 dep-check:
-	ifeq (,$(findstring $(GOVERSION), $(INSTALLED_GO_VERSION)))
+	
+ifeq (,$(findstring $(GOVERSION), $(INSTALLED_GO_VERSION)))
 # Only send a warning.
-#		@echo "Dependency missing: go$(GOVERSION). Ensure 'go$(GOVERSION).x' is installed and available in your 'PATH'"	
+# @echo "Dependency missing: go$(GOVERSION). Ensure 'go$(GOVERSION).x' is installed and available in your 'PATH'"	
+	@echo "GOVERSION: " $(GOVERSION)
+	@echo "INSTALLED_GO_VERSION: " $(INSTALLED_GO_VERSION) 
 # Force error and stop.
-	$(error Dependency missing: go$(GOVERSION). Ensure 'go$(GOVERSION).x' is installed and available in your 'PATH')
-	endif
+	$(error Found $(INSTALLED_GO_VERSION). \
+	 Required golang version is: 'go$(GOVERSION).x'. \
+	 Ensure go '$(GOVERSION).x' is installed and available in your 'PATH'.)
+endif


acald-creator:
Reviewing...

acald-creator:
I'm reviewing your PR, but it is stopping after dep-check checked for go version. From the Makefile, you're checking the GOVERSION which is 1.19 against the installed the version which could be anything 1.19.x  Here is the error Makefile:334: *** Dependency missing: go1.19. Ensure 'go1.19.x' is installed and available in your 'PATH'.  Stop.

Are we only warning that the Go 1.19 needs to be installed, and continue with the script, or are we checking for the Go installed version, and stop the script in case of failure?

Makefile
```diff
@@ -331,5 +318,21 @@ graphql-docs:
...
 
 ## Build Meshery GraphQl API specifications
-graphql-build:
+graphql-build: dep-check
 	cd server; cd internal/graphql; go run -mod=mod github.com/99designs/gqlgen generate
+
+#-----------------------------------------------------------------------------
+# Dependencies
+#-----------------------------------------------------------------------------
+.PHONY: dep-check
```
acald-creator, line 327: Here is a suggestion.

Previously, this `dep-check` will check to see if you have a go version that returns this output `go version go1.19.3 darwin/arm64`

By comparing this to the specified `GOVERSION`, it returned an error stating that the go version is not found and needs to make sure that it is installed and available in the `PATH`.

However after some troubleshooting, I realized that the comparison was actually `1.19` versus `go version go1.19.3 darwin/arm64`. Since it does not match, it will fail.

I have a suggestion, which we can expand on.

```bash
GO_MAJOR_VERSION = $(shell go version | cut -c 14- | cut -d' ' -f1 | cut -d'.' -f1)
GO_MINOR_VERSION = $(shell go version | cut -c 14- | cut -d' ' -f1 | cut -d'.' -f2)
GO_PATCH_VERSION = $(shell go version | cut -c 14- | cut -d' ' -f1 | cut -d'.' -f3)
MINIMUM_SUPPORTED_GO_MAJOR_VERSION = 1
MINIMUM_SUPPORTED_GO_MINOR_VERSION = 19
MINIMUM_SUPPORTED_GO_PATCH_VERSION = 1
GO_VERSION_VALIDATION_ERR_MSG = Go version is not supported, please update to at least $(MINIMUM_SUPPORTED_GO_MAJOR_VERSION).$(MINIMUM_SUPPORTED_GO_MINOR_VERSION).$(MINIMUM_SUPPORTED_GO_PATCH_VERSION)

dep-check:
	@if [ $(GO_MAJOR_VERSION) -gt $(MINIMUM_SUPPORTED_GO_MAJOR_VERSION) ]; then \
		exit 0 ;\
	elif [ $(GO_MAJOR_VERSION) -lt $(MINIMUM_SUPPORTED_GO_MAJOR_VERSION) ]; then \
		echo '$(GO_VERSION_VALIDATION_ERR_MSG)';\
		exit 1; \
	elif [ $(GO_MINOR_VERSION) -lt $(MINIMUM_SUPPORTED_GO_MINOR_VERSION) ] ; then \
		echo '$(GO_VERSION_VALIDATION_ERR_MSG)';\
		exit 1; \
	elif [ $(GO_PATCH_VERSION) -lt $(MINIMUM_SUPPORTED_GO_PATCH_VERSION) ] ; then \
		echo '$(GO_VERSION_VALIDATION_ERR_MSG)';\
		exit 1; \
	fi
```

By listing out the version supported, we can ensure that each time someone runs a go command from the `Makefile`, it will run if the version installed on the machine matches the minimum Go version. 

If it doesn't, it should fail and ask to install at least the minimum. The only caveat is that if the installed version is Go 1.18, the script will continue, but the good thing is that if it is at `1.17` and it will fail because it's looking for that specified `go1.17.x` version.

leecalcote: > However after some troubleshooting, I realized that the comparison was actually 1.19 versus go version go1.19.3 darwin/arm64. Since it does not match, it will fail.

This comparison is intentional. We're only trying to find a match for `1.19`. Was the `dep-check` for `1.19` (as written in this PR) failing on your 1.19.3 system?

acald-creator: And this is even when checking against the substring match for `$(GOVERSION)` set to `1.19` versus `INSTALLED_GO_VERSION` which will return something like `go version 1.19.x arch`

Even when I changed the `GOVERSION` back to `1.19.x`, it still failed.

leecalcote: I'm confused. In what situation does a check work for you?

leecalcote:
@alphaX86 @sudo-NithishKarthik @nishantwrp does `make server` as changed in this PR work for you?

sudo-NithishKarthik:
Yup 👍

nishantwrp:
No, @leecalcote. This is what I get.

```
meshery on  leecalcote/ci/makefile/lighten-golang-requirements [$] via 🐹 v1.19.2 
➜ go version
go version go1.19.2 darwin/arm64

meshery on  leecalcote/ci/makefile/lighten-golang-requirements [$] via 🐹 v1.19.2 
➜ make server
Makefile:333: *** Dependency missing: go1.19. Ensure 'go1.19.x' is installed and available in your 'PATH'.  Stop.
```

leecalcote:
Thank you for testing. This confirms, @acald-creator's findings.

leecalcote:
Go version logic check has been updated. This should behave as desired now.

Revolyssup (approved):

PR #6667, opened 2022-12-09
[chore] Removing deprecated packages

acald-creator:
Signed-off-by: Antonette Caldwell <pullmana8@gmail.com>

**Notes for Reviewers**

This PR removes some of the deprecated packages as Next.js is currently on version 10.

Some of the major features have been deprecated since version 10, and as we move towards upgrading Next.js slowly, we can start to refactor or remove some features and/or packages that are no longer needed because Next.js supports this internally at a later version.

By removing `@zeit/next-css`, Next.js now supports CSS in modules, which means moving the previous `*.css` files to global modules. 

`redux-devtools-extension` has been migrated to `@redux-devtools/extension` with no break in code.

Added a step in `Makefile` to check for `node 19` as when I installed `nvm`, the default latest version if at `v19`.


ui/components/LoadingComponents/Animations/AnimatedMeshPattern.js
```diff
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from "react";
-import "./animatedMeshPattern.css";
+import "./animatedMeshPattern.module.css";
```
leecalcote, line 2: Is in prep for upgrade of Next.js?

leecalcote (approved):

PR #6762, opened 2022-12-22
Add go.mod sync workflow to sync Meshery and Meshery-extensions edge

Revolyssup:

**Notes for Reviewers**

This PR adds a workflow which:
1. Gets triggered only when the go.mod on meshery:master changes
2. It results in go.mod sync utility to run on meshery-extensions repo and commit the synced go.mod

### DIFF
--- a/.github/workflows/go-mod-sync.yaml
+++ b/.github/workflows/go-mod-sync.yaml
@@ -8,7 +8,7 @@ on:
 jobs:
   sync:
     name: sync
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
     steps:
     - name: check out code for meshery
       uses: actions/checkout@master


leecalcote:
Oooh, nice: a PR label and description. No use of `[<component>]` in the title, though, so not immediately clear what component this primarily affects.

leecalcote:
Oh, I see that the `area/ci` label was a freebie, automatically added.

.github/workflows/go-mod-sync.yaml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  push:
+    branches:
+      - "master"
+    paths:
+      - "go.mod"
+jobs:
+  sync:
+    name: sync
+    runs-on: ubuntu-18.04
```
leecalcote, line 11: @StarTrooper08 @acald-creator we've move to 22, correct?

.github/workflows/go-mod-sync.yaml
```diff
@@ -0,0 +1,41 @@
+name: Meshery extensions go.mod sync
```
leecalcote, line 1: @Revolyssup, perhaps, this workflow should be just another step in the existing Edge and Stable release workflows?

leecalcote (changes requested):

.github/workflows/go-mod-sync.yaml
StarTrooper08: Yes, we had move GitHub actions workflow Ubuntu version specifically to 22.04.
Cause it gives some warnings if we are using 18.04 version or specify latest.

.github/workflows/go-mod-sync.yaml
Revolyssup: This only gets triggered when the go.mod changes. The rest are general.

leecalcote: OK. 👍

.github/workflows/go-mod-sync.yaml
leecalcote: Thanks for confirming, @StarTrooper08 :thu

leecalcote (approved):

.github/workflows/go-mod-sync.yaml
StarTrooper08: Happy to Help😄

PR #6831, opened 2023-01-03
[MeshModel] Define Status Types

humblenginr:
This PR adds Issue Types based on the classification done in this [doc](https://docs.google.com/document/d/1y-LHhOcQGA9fiUbN5CR42CXvjaPxKAJHftGrLgxJPTo/edit#heading=h.m5cjnddimh0s). These types are to be used by the Pattern Engine and the future Design Evaluation Engine while sending Design processing Issues to its clients.


leecalcote:
@sudo-NithishKarthik I've updated the PR name. How can we be rid of "Validation" given that it is only one type of issue? Unless there's something more clever and generic, I'll suggest that the package name be changed to "status".

sudo-NithishKarthik:
Makes sense @leecalcote. I think the natural answer would be to use `ErrorType`. I am just thinking about how this will affect our `errors` package since `ErrorType` is generic and it might beg the question as to why it is not included as a part of the `errors` package.

sudo-NithishKarthik:
I've updated the package to be more generic

leecalcote (approved):

PR #6868, opened 2023-01-11
Telemetry Providers

### LINKED ISSUE #6701: [UI] Settings-->Metrics: Support GraphQL subscription
#### Current Behavior
When MeshSync discovers a Grafana or Prometheus instance, they will be listed in the dropdown fields in the Settings page under Metrics if the user refreshes their browser to GET those entries.

#### Desired Behavior
The Settings-->Metrics-->Grafana page and the Prometheus pages are in need of using a subscription in order to expose discovered Grafana and Prometheus endpoints for selection without requiring that the user refresh their browser.

#### Implementation
Create GraphQL Subscription for these connections.

#### Acceptance Tests
The list of discovered Graf and Prom instances populate without page refresh.

---

#### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](http://slack.layer5.io)

MUzairS15:
Signed-off-by: MUzairS15 <muzair.shaikh810@gmail.com>

**Notes for Reviewers**

This PR fixes #6701 
1. Fixes broken state of cluster resources subscription (new resources were not being updated)
2. Added mesh sync events subscription which sends detect object by MeshSync, utilising this subscription UI gets informed about telemetry component. (Ideally cluster resources subscription can be replaced and instead this can be used, though out of scope of this PR)
3. Removed handlers and utility function for kubernetes scan

### DIFF
--- a/server/internal/graphql/resolver/error.go
+++ b/server/internal/graphql/resolver/error.go
@@ -34,7 +34,7 @@ const (
 	ErrMeshModelSummarySubscriptionCode     = "2248"
 	ErrGettingMeshModelSummaryCode          = "2249"
 	ErrGettingRegistryManagerCode           = "2250"
-	ErrGettingTelemetryComponentsCode		= "2251"
+	ErrGettingTelemetryComponentsCode       = "2251"
 )
 
 var (


MUzairS15:
Tests failed because an API request had been removed from the settings page.

theBeginner86 (commented):
@MUzairS15 
Could you please tell, why do we need to rely on the `MeshSyncEvents` subscription for getting telemetry status? Wouldn't it be helpful if we have a separate first-class subscription that handles this specific use case? We can customize and expand it later on based on future telemetry use cases. I'm saying this because currently, we seem to be determining major stuff in UI from the response from `MeshSyncEvents` subscription.
I see that you have implemented `Telemetry query` and a global redux var to track the status of telemetry which looks great.

Also, could you please point to the fix for `clusterResources` subscription? I'm unable to find it in the current set of changes.
Thanks

I tested this out and it works well!

MUzairS15:
> @MUzairS15 Could you please tell, why do we need to rely on the `MeshSyncEvents` subscription for getting telemetry status? Wouldn't it be helpful if we have a separate first-class subscription that handles this specific use case? We can customize and expand it later on based on future telemetry use cases. I'm saying this because currently, we seem to be determining major stuff in UI from the response from `MeshSyncEvents` subscription. I see that you have implemented `Telemetry query` and a global redux var to track the status of telemetry which looks great.
> 
> Also, could you please point to the fix for `clusterResources` subscription? I'm unable to find it in the current set of changes. Thanks
> 
> I tested this out and it works well!

The cluster resources subscriptions, depends on resolver of meshsyncevents subscription to send an update, but the meshsync events  subscription never initiated from UI, hence it was not updating the UI correctly, I initated the meshsync events form UI and now UI is update as soon as change is detected,

For using the same subscription, I think we still be signalling, there is a change in cluster, then we wil query the database and we will get the same object that is being sent via meshsync events.

So we can have a generalised subscriptio (meshsync) and we can act according to incoming payload. With this, an extra lookup to DB as well as extra connection to server is reduced.
In the UI we are just exxtracting the URL from the component only if it's an telemtry service detected.

theBeginner86:
> > @MUzairS15 Could you please tell, why do we need to rely on the `MeshSyncEvents` subscription for getting telemetry status? Wouldn't it be helpful if we have a separate first-class subscription that handles this specific use case? We can customize and expand it later on based on future telemetry use cases. I'm saying this because currently, we seem to be determining major stuff in UI from the response from `MeshSyncEvents` subscription. I see that you have implemented `Telemetry query` and a global redux var to track the status of telemetry which looks great.
> > Also, could you please point to the fix for `clusterResources` subscription? I'm unable to find it in the current set of changes. Thanks
> > I tested this out and it works well!
> 
> The cluster resources subscriptions, depends on resolver of meshsyncevents subscription to send an update, but the meshsync events subscription never initiated from UI, hence it was not updating the UI correctly, I initated the meshsync events form UI and now UI is update as soon as change is detected,
> 
> If we use a dedicated subscription, we will signal, there is a change in cluster, then we wil query the database and we will get the same object that is being sent via meshsync events. Instead if we can have a generalised subscription (meshsync) and we can act according to incoming payload. With this, an extra lookup to DB as well as extra connection to server is reduced.
> 
> In the UI we are just extracting the URL from the component only if it's an telemtry service detected.

Yeah, the argument of the reduced overhead and advantages of reuses makes sense. 
Thanks @MUzairS15, for the explanation 🙏

theBeginner86 (approved):
LGTM

> Just a merge conflict to fix

PR #6871, opened 2023-01-12
[Deps] Chore: upgrade to github.com/99designs/gqlgen v0.17.24

leecalcote:

**Notes for Reviewers**

This PR fixes #


acald-creator:
Last I look, this PR should be good to go.

One thing I will say, `gqlgen` only supports `gorilla/websocket`, looks like they have no plans to support any other libraries at this time, but they did add `sse` support, which is Server-Side Events.

I think this is something we can look into, as to still use `gorilla/websocket` as a fallback option until `gqlgen` decides whether they want to proceed with moving away from `gorilla/websocket` or add support for other libraries.

acald-creator:
// @leecalcote 
Can we revisit this?

Since we just did a release on Meshery and MeshMap, we can upgrade `gqlgen` before the next release so this will match in MeshMap `go.mod`. I don't think there wasn't much changes on generating the files needed with `gqlgen`.

leecalcote:
@acald-creator, I missed this comment until now. Yes, allow me to handoff the baton on this upgrade.

acald-creator:
Closing, will create a new PR in time for the next release, as this will needs to be replicated in `meshmap`

PR #6747, opened 2022-12-19
Part one upgrade

acald-creator:
**Notes for Reviewers**

This PR is related to the Meshery UI library and packages upgrade at https://github.com/meshery/meshery/issues/6680

**Goal**: Upgrade the following main packages
1. `Next.js` **12** minimum
2. `React@17` & `React-DOM@17`
3. `MUI` **v5** (`@mui/material`)(will open a separate issue to track this)
4. `Eslint` v7 minimum (due to package conflict from `Next.js` **10** to **12**)

**Blockers**: Packages that are conflicted
1. `react-codemirror2` - relies on `React@16`, inactive for 2 years
    1. Suggestions:
        1. upgrade to  `@uiw/react-codemirror` latest version (will require refactoring as YAML linting capabilities is not part of the library)
        2. pin `@uiw/react-codemirror` at `v3` (will not work due to Next.js 11 and above Global CSS support). See error below. And see error referenced at https://nextjs.org/docs/messages/css-npm. **Note that the previously hackity solution within `_app.js` is not working, plus additional suggestions.**
        3. use `codemirror v6` directly as Codemirror have revised their library from `v5` to `v6` and refactor multiple files that call `CodeMirror`
        4. force `react-codemirror2` as `--legacy-peer-deps` until final decision made from above, although this may require a hackity solution in order to make sure this continues to work because of the SSR and Client issues
        5. not force `react-codemirror2` and suggest a different library such as `Monaco Editor` (There is a React version that supports up to React 18)

```bash
./node_modules/@uiw/react-codemirror/esm/codemirror.css
Global CSS cannot be imported from within node_modules.
Read more: https://nextjs.org/docs/messages/css-npm
Location: node_modules/@uiw/react-codemirror/esm/index.js
```

For now, we are testing `monaco-editor` as a replacement, and so far there are several pages that requires changes.
* Applications (have not started)
* Filters (partially done)
* Designs (partially done) - need to fix fullscreen and loading new monaco-editor first

2. `react-select` can be upgraded to `v5` and just need an import changed to `import Creatable from 'react-select/creatable';`
3. `eslint` will need to be pinned at `v7` due to conflicts in between Next.js `10` to `12`
4. Remove `npm uninstall eslint-loader eslint-config-airbnb eslint-config-google eslint-plugin-react-hooks eslint-plugin-jsx-a11y eslint-plugin-import` as these have already been supported internally when `Next` has built-in support for `eslint`. See here https://nextjs.org/docs/basic-features/eslint
[...]

Issue #6680, opened 2022-12-12
[UI] Framework and Library Packages Upgrades and Removal


acald-creator:
#### Desired Behavior
Would like to upgrade Meshery UI libraries to latest packages. **PLEASE NOTE** This is a work in progress to track our progress to `Next.js v13` at the very least.

Currently, `meshery/ui` has a rather large list of NPM packages, but it would be helpful to initiate a chore to clean up packages that are no longer needed for the `ui` project.

Here are some suggestions of packages that are no longer needed or not being used directly in `meshery/ui`. Any feedback is welcome.
* `glob-parent`
* `@types/react-relay`
* `caniuse-lite`
* `color-string`
* `@babel/traverse`
* `cron-time-generator`
* `set-value`
* `minimist`
~~* `webpack` and `webpack-cli` are not being directly, in the upcoming Next.js version above 11, Next.js is using `webpack` internally, so there is no need to use these directly, unless there is plan to continue using `webpack-cli` directly.~~ 
* **CORRECTION**: `webpack` is needed for Cypress testing.
~~* `next-compose` - Next.js has better supports for plugins, and the package is over 5 years old.~~

#### Concerns to address in Next.js 13
- [ ] `Next.js` 13 requires `React 18`, which means a couple of the packages will not support `React 18` directly or indirectly. (WIP)
    - [ ] `react-codemirror2` is about 2 years old, and no longer maintained. Suggest to use `@uiw/react-codemirror` which supports `React 17` and above. There will be about 11 files or so that needs to be refactored to rename the inports.
    - [ ] ~~`"react-select": "^2.4.3"` will needs to be ugpraded to the latest version `v5.7.0`~~ **DONE - Solve by renaming imports.**
- [x] **DONE** - Refactor `eslint` configuration in the following files: `package.json`, and `.eslintrc.js`. `Next.js` currently supports `eslint` default via `next lint`. There is also support for `prettier` and `lint-staged`.
    - [x] `eslint` is currently at `v7`, suggest to move to `v8` to follow the typical `Next.js` 12/13 project layout.
    - [ ] `eslint-loader` has been deprecated and moved to `eslint-webpack-plugin`. Reassess if this package is still needed.
    - [x] Multiple `eslint-plugin-*` have not been used within the project. Reassess to see if these packages are still needed in terms of linting standard.
        1. `eslint-config-airbnb`
        2. `eslint-config-google` 
        3. `eslint-plugin-react-hooks`
[...]

PR #6880, opened 2023-01-13
Fast Track Next.js

acald-creator:
**Notes for Reviewers**

This PR is related to https://github.com/meshery/meshery/issues/6680 and https://github.com/meshery/meshery/pull/6747.

This is simply to fast track Next.js to `v12` and React to `v17`.

Unfortunately I ran into some styling issues at Next.js `13` and React `18` which will require a review of the current Material UI v4 to be migrated to Material UI v5 sooner than later.

A few things to note:

`eslint` is pinned at v7 for now because of an issue with Next.js 12 which is looking for eslint to be installed as a dev dependencies but the message didn't go away until I pinned it at this version.

I removed some of the `eslint` plugins and config because they either have been merged together or deprecated in favor of something else, but in this PR, most of them have been merged into one, where Next.js has better built-in support for eslint rules. Eslint will be a bit noisy, so we will have to review and decide which to turn off.

Because we are upgrading to React 17-18, any new React Components will now need to start using React Function Components instead of React Class Components, and this also means using React Hooks.

Lastly, Cypress had some dependencies issues, so I upgraded webpack to v5 and installed ajv in the meantime. In my local development, the test seemed to be passing, but we will leave it others to provide feedback and test this PR.

### DIFF
--- a/.github/workflows/build-ui-server-reusable-workflow.yml
+++ b/.github/workflows/build-ui-server-reusable-workflow.yml
@@ -1,10 +1,10 @@
 name: Re-usable Workflow - Build UI and Server
 on:
-  workflow_call: 
-    secrets: 
+  workflow_call:
+    secrets:
       NODE_VERSION:
         required: true
-      GO_VERSION: 
+      GO_VERSION:
         required: true
 
 jobs:
@@ -46,7 +46,7 @@ jobs:
           fetch-depth: 1
       - uses: actions/setup-node@v3.0.0
         with:
-          node-version: ${{ secrets.NODE_VERSION }}
+          node-version: 18
       - name: Cache node modules
         id: node-cache
         uses: actions/cache@v3
@@ -81,6 +81,7 @@ jobs:
           path: ui/out/
       - name: upload Meshery-provider-ui artifacts
         uses: actions/upload-artifact@v3
-        with: 
+        with:
           name: meshery-provider-ui build
           path: provider-ui/out/
+
--- a/docs/_releases/v0.6.43.md
+++ b/docs/_releases/v0.6.43.md
@@ -1,0 +1,28 @@
+---
+layout: release
+date: 2023-01-13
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -235,7 +235,7 @@ ui-build:
 
 ## Build only Meshery UI on your local machine.
 ui-meshery-build:
-	cd ui; npm run $(UI_BUILD_SCRIPT) && npm run export; cd ..
+	cd ui; npm run build && npm run export; cd ..
```
Abhishek-kumar09, line 238: Why this change?

ui/.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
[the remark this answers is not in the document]
acald-creator: `react-codemirror2` and `mui v4` are the only real blockers.

There are other smaller packages but those are already noted on my other PR.

install/docker/Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN go clean -modcache; cd server; cd cmd; GOPROXY=https://proxy.golang.org GOSU
 
 FROM node:lts as ui
 ADD ui ui
-RUN cd ui; npm install --only=production; npm run build && npm run export; mv out /
+RUN cd ui; npm install --omit-dev; npm run build && npm run export; mv out /
```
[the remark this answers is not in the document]
acald-creator: When I was running docker-build, I saw the message regarding this particular line was deprecated, node:lts is node 18

Makefile
acald-creator: For the purpose of testing Cypress

leecalcote (approved):

Makefile
Abhishek-kumar09: I dont think this is affecting cypress at all. Now a new problem arises, that will stop us from building meshery with any node version later than 16

PR #6911, opened 2023-01-18
`provider-ui` MUI framework upgrade

acald-creator:
**Notes for Reviewers**

This PR is for updating `provider-ui` to the latest
* Next.js 13
* React@18 and React-DOM@18
* MUI v5

One thing to note is that the Dockerfile has been changed to use `node:lts` which is `v18`, which is the same thing `ui` uses.

This also included some rearrangement of components, to reduce bloated and unused styles, and reduce the code being used in `_app.js`. This results in separate components called `Provider` and `Footer` along with custom styling for Provider.

The content via the drop-down menu has been changed to fit the width of the content, as before it was not.

The theme is one file with a few overrides, as there are only a few React components to create and use.

### DIFF
--- a/provider-ui/components/Footer.js
+++ b/provider-ui/components/Footer.js
@@ -12,7 +12,7 @@ const Item = styled(Paper)(({ theme }) => ({
   border : '0'
 }))
 
-export default function Footer () {
+export default function Footer() {
   const handleL5CommunityClick = () => {
     if (typeof window !== 'undefined') {
       const w = window.open('https://layer5.io', '_blank')
--- a/provider-ui/.eslintrc.js
+++ b/provider-ui/.eslintrc.js
@@ -9,10 +9,11 @@ module.exports = {
     es2021 : true
   },
   extends : [
+    'eslint:recommended',
+    'next',
     'plugin:react/recommended',
     'plugin:@next/next/recommended',
     "plugin:cypress/recommended",
-    'next/core-web-vitals',
   ],
   overrides : [
   ],
--- a/provider-ui/lib/createEmotionCache.js
+++ b/provider-ui/lib/createEmotionCache.js
@@ -13,5 +13,5 @@ export default function createEmotionCache () {
     insertionPoint = emotionInsertionPoint ?? undefined
   }
 
-  return createCache({ key: 'mui-style', insertionPoint })
+  return createCache({ key:'mui-style', insertionPoint })
 }


Abhishek-kumar09 (commented):
Did you compeleted the changes, solving the problems you were last discussing on the call around drop-downs?

acald-creator:
Here is a screenshot of how it looks at the moment with the drop down menu issue. I'll need to make some more changes and this PR should be ready for review today.
provider-mui

Abhishek-kumar09 (changes requested):
Looks Good to Me!
I just tested the UI locally, it works fine, just some style changes I would like to ask for:
old:
Screenshot 2023-01-26 at 10 29 46 PM

New:
Screenshot 2023-01-26 at 10 31 32 PM

---
Although I was not able to successfully `npm i`, the cypress installation was taking a lot of time, anyways I think that was my network issue.

provider-ui/.gitignore
```diff
@@ -1,6 +1,18 @@
-# Cypress Tests Files
-cypress/videos
-cypress/screenshots
```
leecalcote, line 3: This seems like something we would want to ignore...

provider-ui/.eslintrc
```diff
@@ -0,0 +1,62 @@
+module.exports = {
```
leecalcote, line 1: This change is necessary for MUI v5 compatibility?

provider-ui/components/Footer.js
```diff
@@ -0,0 +1,47 @@
+import React, { Fragment } from 'react'
```
leecalcote, line 1: I might be missing something obvious. Why was this footer rewritten?

provider-ui/ui_dev_server.js
```diff
@@ -1,39 +1,49 @@
...
+            } else {
+                if (req.url.startsWith('/provider')) {
+                    req.url = req.url.replace('/provider', '')
+                }
+                await handle(req, res, parsedUrl)
+            }
+        } catch (err) {
+            console.error('Error occurred handling', req.url, err)
+            res.statusCode = 500
+            res.end('internal server error')
```
leecalcote, line 43: Can we offer something more informative?

provider-ui/ui_dev_server.js
```diff
@@ -1,39 +1,49 @@
...
+    createServer(async (req, res) => {
+        try {
+            // Be sure to pass `true` as the second argument to `url.parse`.
+            // This tells it to parse the query portion of the URL.
+            const parsedUrl = parse(req.url, true)
+            const { pathname, query } = parsedUrl
+
+            if (pathname.startsWith('/api')) {
+                proxy.web(req, res, {
+                    target: 'http://localhost:9081',
```
leecalcote, line 32: Why are hostname and port being defined as constants, but not referenced here?

provider-ui/cypress.config.js
```diff
@@ -14,7 +14,7 @@ module.exports = defineConfig({
     setupNodeEvents(on, config) {
       return require('./cypress/plugins/index.js')(on, config)
     },
-    baseUrl: 'http://localhost:3001',
+    baseUrl: 'http://localhost:3001/provider',
```
leecalcote, line 17: Let's consider a separate improvement (separate issue) to dynamically attempt to use a higher port number in the case of port conflict on the contributor's system.

provider-ui/components/ProviderLayout.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import NoSsr from '@mui/material/NoSsr'
+
+export default function ProviderLayout({ children }) {
+  return (
+    <>
+      <NoSsr>
+        <div data-cy="root" style={{ padding: '170px 0px', textAlign: 'center' }}>
```
leecalcote, line 9: If we're rewriting, then let's not propagate bad practices of static styling, but instead move to flex.

provider-ui/ui_dev_server.js
acald-creator: Let me think of something

provider-ui/components/Footer.js
acald-creator: Footer wasn't rewritten, and it was moved to its own component. the `_app.js` was bloated with styles that weren't being used

provider-ui/.eslintrc
acald-creator: Not directly, sorry this was more to Next.js 13

provider-ui/.gitignore
acald-creator: Fixed

provider-ui/components/Footer.js
leecalcote: Ok, I see. Thanks for this. Makes much sense. 👍

leecalcote (dismissed):
LGTM

Abhishek-kumar09 (approved):
Looks Good to Me

Abhishek-kumar09:
Screenshot 2023-01-27 at 1 08 06 AM
eslint fixes pending

provider-ui/cypress.config.js
acald-creator: I reverted the change, it was not intended to be this way.

provider-ui/components/Provider.js
```diff
@@ -0,0 +1,260 @@
+import React, { Fragment, useEffect, useState } from "react";
```
Abhishek-kumar09, line 1: The tabwidth used in this file is 4 while in other it is 2. 
Can you please enforce 2, all around the project via eslintrc.

provider-ui/.eslintignore
```diff
@@ -5,5 +5,7 @@ static
 package.json
 package-lock.json
 nodemon.json
-next.config.js
-cypress/
\ No newline at end of file
+lib/
+cypress/
+pages/_document.js
```
Abhishek-kumar09, line 10: any specific reason for putting it in eslintignore?

Abhishek-kumar09: `pages/_document.js`

Abhishek-kumar09 (approved):

PR #6920, opened 2023-01-20
[Relationship] Add deployment pod relationship

Abhishek-kumar09:
Add deployment pod hierarchical relationship.
**Notes for Reviewers**

This PR fixes #

### DIFF
--- a/server/meshmodel/relationships/deployment_pod_relationship.json
+++ b/server/meshmodel/relationships/deployment_pod_relationship.json
@@ -1,0 +1,25 @@
+{
+  "apiVersion": "core.meshery.io/v1alpha1",
+  "kind": "Hierarchical",
+  "metadata": {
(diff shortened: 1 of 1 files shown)


leecalcote:
@Abhishek-kumar09 this is great considering that Deployments do contain one or more Pods. With that said, I'm sorry to say that there is a bit of discussion needed as the Deployment spec already has it's own internal Pod spec, which means that adding this relationship (and allowing these two components to combine) means that there will be accidental overlap of those specs. // @Revolyssup

leecalcote:
> @Abhishek-kumar09 this is great considering that Deployments do contain one or more Pods. With that said, I'm sorry to say that there is a bit of discussion needed as the Deployment spec already has it's own internal Pod spec, which means that adding this relationship (and allowing these two components to combine) means that there will be accidental overlap of those specs. // @Revolyssup

@sudo-NithishKarthik can you think of a way around this?

leecalcote:
Let's close this out for now.

PR #7046, opened 2023-02-09
Deny designs save with empty service name from Meshery Extensions

Revolyssup:

**Notes for Reviewers**

This PR fixes #

### DIFF
--- a/server/models/pattern/core/pattern.go
+++ b/server/models/pattern/core/pattern.go
@@ -200,12 +200,13 @@ func (p *Pattern) ToYAML() ([]byte, error) {
 }
 
 // NewPatternFileFromCytoscapeJSJSON takes in CytoscapeJS JSON
-// and creates a PatternFile from it
+// and creates a PatternFile from it.
+// This function always returns meshkit error
 func NewPatternFileFromCytoscapeJSJSON(name string, byt []byte) (Pattern, error) {
 	// Unmarshal data into cytoscape struct
 	var cy cytoscapejs.GraphElem
 	if err := json.Unmarshal(byt, &cy); err != nil {
-		return Pattern{}, err
+		return Pattern{}, ErrPatternFromCytoscape(err)
 	}
 	if name == "" {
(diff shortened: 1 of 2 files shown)


server/models/pattern/core/pattern.go
```diff
@@ -317,6 +317,9 @@ func processCytoElementsWithPattern(eles []cytoscapejs.Element, pf *Pattern, cal
 		if err := json.Unmarshal(svcByt, &svc); err != nil {
 			return fmt.Errorf("failed to create service from the metadata in the scratch")
 		}
+		if svc.Name == "" {
+			return fmt.Errorf("cannot save service with empty name")
```
leecalcote, line 321: What Meshery error code belongs to this error? Suggested remediation?

leecalcote (changes requested):

server/models/pattern/core/pattern.go
Revolyssup: This internal error is returned and wrapped into a meshklt error where the function is called. `ErrSavePattern`

leecalcote: Good deal. Although - https://docs.meshery.io/search/?q=ErrSavePattern. ☹️

leecalcote: I see "ErrSavePatternCode", but its description is not helpful in this situation. https://docs.meshery.io/reference/error-codes

leecalcote: @acald-creator We need to continue our audit of error codes.

leecalcote:
Pending creation of a new error code.

Revolyssup:
> Pending creation of a new error code.

Added a new error code for this specific function.

leecalcote:
With an eye toward MeshModel, will this error code fall to the wayside?

leecalcote (approved):
LGTM. Will you please open an issue or two regarding improving our error codes? One of the issues should be to combine the moniker and the code into the same column on the docs page.

PR #7093, opened 2023-02-15
[Fix] Cypress integration Test failure

Abhishek-kumar09:
**Notes for Reviewers**
1. The subscription-transport-ws had problem in its code, that was thorwing uncaught exceptions in UI, that was hindering cypress working, since the package is not actively maintained, added the patch that handles the situation.
2. The old fixture of MeshAdapters has expired and new reponse data is feeded 
3. Lastly, fix in API interception position, i.e., interception code should before the request is fired. The race condition was causing the bug there, ultimately failing the integration tests in some situations.

This PR fixes [Integration Test Failures](https://github.com/meshery/meshery/actions/runs/4162317179/jobs/7201453476)

### DIFF
--- a/provider-ui/cypress/e2e/provider-ui.cy.js
+++ b/provider-ui/cypress/e2e/provider-ui.cy.js
@@ -14,7 +14,7 @@ describe('Provider UI', () => {
     it('selects None as provider', () => {
         cy
             .get('[data-cy="select_provider"]').click()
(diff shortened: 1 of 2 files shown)


Abhishek-kumar09:
All tetsts in Meshery-UI passes ✅ I didn't know the provider also had test failures. Looking into it.

leecalcote (approved):
✅

PR #7103, opened 2023-02-16
use omitEmpty JSON and YAML tags for Namespace

Abhishek-kumar09:
Pull request created in favour of catering non-namespaced components. Currently, if any component is non-namespaced, the design-yaml is adding empty string in that place. With this PR the property is tagged as `omitEmpty`

**Notes for Reviewers**
The dry-run is still not functional for non-namespaced propeties, @Revolyssup 
https://layer5io.slack.com/archives/G01BD9056TD/p1676562518902659

This PR fixes #

### DIFF
--- a/ui/components/MesheryMeshInterface/PatternService/RJSF.js
+++ b/ui/components/MesheryMeshInterface/PatternService/RJSF.js
@@ -65,14 +65,16 @@ function RJSFForm(props) {
     // prop should be present in order for the cloned element to override this property
     transformErrors,
     override,
+    uiSchema={}
   } = props;
+
   const templates = {
     ArrayFieldTemplate,
     ObjectFieldTemplate,
     WrapIfAdditionalTemplate,
     FieldTemplate : CustomFieldTemplate, // applying field template universally to every field type.
   }
-  const globalTheme = useTheme()
+  const globalTheme = useTheme();
   useEffect(() => {
     const extensionTooltipPortal = document.getElementById("extension-tooltip-portal");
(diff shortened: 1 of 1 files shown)


leecalcote (dismissed):
@Abhishek-kumar09 LGTM, but more importantly, can we stop investing in OAM and shift all focus to MeshModel?

Abhishek-kumar09:
Searching/Querying
Duplicate components from MeshModel API
camelCase naming

leecalcote (commented):
Was `omitempty` added for MeshModel components, too?

server/models/pattern/core/pattern.go
```diff
@@ -43,7 +43,7 @@ type Service struct {
 	Name        string            `yaml:"name,omitempty" json:"name,omitempty"`
 	Type        string            `yaml:"type,omitempty" json:"type,omitempty"`
 	APIVersion  string            `yaml:"apiVersion,omitempty" json:"apiVersion,omitempty"`
-	Namespace   string            `yaml:"namespace" json:"namespace"`
+	Namespace   string            `yaml:"namespace,omitempty" json:"namespace,omitempty"`
```
Abhishek-kumar09, line 46: > Was omitempty added for MeshModel components, too?

@leecalcote namespace is a pattern-file contruct and independent of any interface. So it will work for both without discrimination.

leecalcote: Oh, very good. :thu

leecalcote (approved):

PR #7105, opened 2023-02-17
Upgrade gqlgen

acald-creator:
**Notes for Reviewers**

This PR upgrade `gqlgen`, and it's actually important now to upgrade as I ran into an error generating the resolvers and others.

This will need to match in meshmap once the go.mod sync completes.


leecalcote:
Bad build...

leecalcote:
...I meant to say that GitHub failed us with a bad build.

acald-creator:
I thought `golangci-lint` has a timeout arg set, but apparently not.

leecalcote:
> I thought `golangci-lint` has a timeout arg set, but apparently not.

Oh, is that what is going on?

acald-creator:
> > I thought `golangci-lint` has a timeout arg set, but apparently not.
> 
> Oh, is that what is going on?

Actually the timeout arg is set at 3m, so we will need to set it a bit higher then, 5m should work. Ill update this PR

leecalcote (approved):
Yay!

Issue #7018, opened 2023-02-06
[UI] Upgrade `eslint` to `v8`


acald-creator:
#### Current Behavior

Currently we are on `eslint@v7`, which occurred because of the npm packages upgrade.

However, we halted some progress of it, to focus on others open issues. I wanted to open up an issue to track this, as somewhere in this year, we will need to move to `eslint@v8` as the rules we're currently using is `next/core-web-vitals`, which is a bit strict, and since we're on Next.js `v12`, this version also correlate with `eslint@v8` as well.

To slowly migrate and upgrade, I will be reverting to use `next`, which is the basic of the Eslint rules provided by Next.js, once we're a bit more mature, we can switch to using a different set of rules, if we decide to do this.

#### Environment

- Browser: Chrome Safari Firefox
- Host OS: Mac Linux Windows
- Meshery Server Version: stable-v
- Meshery Client Version: stable-v
- Platform: Docker Kubernetes

---

#### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🎨 Wireframes and designs for Meshery UI in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](http://slack.layer5.io)

PR #7124, opened 2023-02-20
[chore] Upgrade eslint to v8

acald-creator:
**Notes for Reviewers**

This PR fixes [#7018](https://github.com/meshery/meshery/issues/7018).

At this time, we are upgrading `eslint` to `v8` as there were some potential miss with linting in general.

As of right now, there are three rules that have been changed for now.

`"react-hooks/rules-of-hooks" : "warn"` has been changed to warn, it was set off
`"react-hooks/exhaustive-deps" : "off"` has been added to be turned off for now
`"valid-typeof" : "warn"` is added and set to warn for now

After upgrading `eslint`, there are a number of files that were formatted based on the linting/formatting rules stated in `.eslintrc.js`.

As stated in the original issue, we can revisit this, when we are a bit mature.

### DIFF
--- a/ui/.eslintrc.js
+++ b/ui/.eslintrc.js
@@ -9,7 +9,7 @@ module.exports = {
     "eslint:recommended",
     "plugin:react/recommended",
     "plugin:cypress/recommended",
-    "next/core-web-vitals"
+    "next"
   ],
   "globals" : {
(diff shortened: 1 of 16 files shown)


leecalcote (approved):
LGTM

PR #7148, opened 2023-02-22
[WIP] [UI] Migrate to `graphql-ws`

acald-creator:
**Notes for Reviewers**

* Removed `subscriptions-transport-ws`, and migrate to `graphql-ws`.
* Removed `subscriptions-transport-ws` patch file. Added a timer to check for Broker status.

I added a timer here to check for the Broker status.

Using `mesheryctl system start` will create and deploy the Kubernetes resources. However, after accessing the Dashboard, there is a graphql query that searches for the status of the Broker. I added a timer so that this can be checked if the status is up.

This can be improved so that there wouldn't be too many error messages showing the console, that may confuse any newcomers or anyone else that is trying out Meshery.


ui/components/Header.js
```diff
@@ -527,11 +527,16 @@ class Header extends React.Component {
...
   componentDidMount() {
     this._isMounted = true;
-    const brokerStatusSub = subscribeBrokerStatusEvents(data => {
-      console.log({ brokerData : data })
-      this.setState({ brokerStatus : data?.subscribeBrokerConnection })
-    });
-    this.setState({ brokerStatusSubscription : brokerStatusSub })
+    const checkServiceStatus = () => {
+      const brokerStatusSub = subscribeBrokerStatusEvents(data => {
+        console.log({ brokerData : data });
```
Abhishek-kumar09, line 532: Isn't it creating a new subscription every 5 sec? In case it is, the previous subscription should be flushed down. Addionally, the health check is great way of maintaining the correct UI status wrt to connectivity to broker.

acald-creator: It shouldn't be, it should be checking to see if it is active, not creating a subscription.

We just need a healthcheck, because the only time you're seeing this is during Docker build. I can't test this locally, which I haven't figured out the easiest way to do this.

Abhishek-kumar09: https://github.com/meshery/meshery/blob/0c4bc75fda98985607735ce9b521653951bf22bd/ui/components/graphql/subscriptions/BrokerStatusSubscription.js

As it written, I think that it is creating a subscription to the grapqhQL query `BrokerStatusSubscription`. If we don't flush the previous subscription data then we will be in loop. 

Although if you're saying it isn't I will confirm this after checking out to your PR in a while. I may be wrong here.

acald-creator: I can test it again too, I can revert part of it to the original, and just keep the health check portion.

acald-creator: After some testing, I don't see the subscription being recreated.

The lines of code are simply moved into their own function so that when the health check for the subscription if it is live or not, it will ping for this every 5 minutes. Which should give enough time to load from the Docker build.

MUzairS15 (commented):
I don't see any of the subscriptions created (there is no initialisation of subscription taking place, not even once)
Screenshot 2023-02-24 at 10 45 22 AM

acald-creator:
> I don't see any of the subscriptions created (there is no initialisation of subscription taking place, not even once) Screenshot 2023-02-24 at 10 45 22 AM

Thanks for bringing this up. I need more people to test to ensure that they can see what I can see.

As mentioned in my original PR description, you could only see this through the Docker build, and if you're using `mesheryctl system start` which will deploy the Kubernetes services.

Without this, you will just see the errors about unable to connect to the necessary endpoint, from what I've seen.

Once I get a working build, I can show a quick video of what happens and you can see the subscriptions been made and the queries happening.

acald-creator:
Closing PR, will reopen a new once I have tested this a bit more.

PR #7162, opened 2023-02-23
[chore] Clean up Makefile

acald-creator:
* Removed `(requires go${GOVERSION})`.

Not needed, as we already have a dependency check for Go version.

* Added another step called `ui-provider-build` to `ui-server`.

The above step failed in running `make ui-server` because it is missing the step to build and export `provider-ui` when doing these troubleshooting steps from `meshery-extensions`.

**Notes for Reviewers**

This PR fixes #


leecalcote (approved):
It pains me that `ui-provider-build` only adds one more step to the go-to build flow with `make ui-build`, but you've made a good call here (in addition to the removal of the redundant go version check). LGTM.

PR #7299, opened 2023-03-07
fix meshmodel duplicates 

Revolyssup:
**Notes for Reviewers**

This PR fixes #

### DIFF
--- a/server/cmd/main.go
+++ b/server/cmd/main.go
@@ -250,9 +250,6 @@ func main() {
 				case comp := <-compChan:
 					if comp.Metadata != nil && comp.Metadata["published"] == true {
 						utils.WriteSVGsOnFileSystem(&comp)
-						if comp.Model.Name == "core" {
-							fmt.Println(comp)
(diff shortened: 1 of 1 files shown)


leecalcote:
What was the cause of this issue?

Revolyssup:
> What was the cause of this issue?

There were two issues. One of which was my oversight, where components of all modelVersion were present inside a specific ModelVersion due to an incorrect filter.

leecalcote:
Thanks. It'd be nice to see this logic moved out of main.go and into an import function.

leecalcote:
Perhaps, something that @Azanul can work on. Having said that, I having seen any progress on a connection state machine just yet, though.

Azanul:
> Perhaps, something that @Azanul can work on.

Sure

> Having said that, I having seen any progress on a connection state machine just yet, though.

I'm assuming that's a "haven't". The connection state machine 🔮 presents itself to mortals when they're ready for it.

PR #7309, opened 2023-03-08
fix potential race condition

MUzairS15:
**Notes for Reviewers**

This PR fixes a potential race condition which may occur. 

### DIFF
--- a/server/internal/graphql/resolver/configuration.go
+++ b/server/internal/graphql/resolver/configuration.go
@@ -28,7 +28,7 @@ func (r *Resolver) subscribeConfiguration(ctx context.Context, provider models.P
 		r.Log.Error(models.ErrGenerateUUID(err))
 		return nil, err
 	}
-	
+
 	r.Config.ConfigurationChannel.SubscribePatterns(chp, key)
 	r.Config.ConfigurationChannel.SubscribeApplications(cha, key)
 	r.Config.ConfigurationChannel.SubscribeFilters(chf, key)
--- a/server/models/configuration_helper.go
+++ b/server/models/configuration_helper.go
@@ -42,7 +42,7 @@ func (c *ConfigurationChannel) UnSubscribeApplications(key uuid.UUID) {
 
 func (c *ConfigurationChannel) PublishPatterns() {
 	for _, ch := range c.PatternsChannel {
-		ch <- struct{}{}		
+		ch <- struct{}{}
 	}
 }
 
@@ -77,4 +77,3 @@ func (c *ConfigurationChannel) UnSubscribeFilters(key uuid.UUID) {
 	close(c.FiltersChannel[key])
 	delete(c.FiltersChannel, key)
 }
-


server/internal/graphql/resolver/configuration.go
```diff
@@ -22,6 +22,7 @@ func (r *Resolver) subscribeConfiguration(ctx context.Context, provider models.P
 	chp <- struct{}{}
 	cha <- struct{}{}
 	chf <- struct{}{}
+	key := len(r.Config.ConfigurationChannel.ApplicationsChannel)
```
Revolyssup, line 25: @MUzairS15 Given the fact that multiple instances of this function will be called for each client, here this key will introduce another race condition. 
Imagine this: Two clients connect at approx the same time. In both instances of this function, len will turn out to be the same and hence two different subscription will get the same key.
Use UUID or something like that.

Revolyssup (changes requested):

Azanul:
@MUzairS15 I couldn't see a race. Will you please point it out for me.

MUzairS15:
> @MUzairS15 I couldn't see a race. Will you please point it out for me.

Basically, here in this function https://github.com/meshery/meshery/blob/1949742026f4828ddbd3c5bf90d5008b51550c78/server/helpers/utils/utils.go#L112
we are expecting value and, based on this, deciding whether the channel is closed or not. https://github.com/meshery/meshery/pull/7309/files#diff-c9ba355fc474472a3b057eed347eb447b73f8ae534b2277c78f1767ac04a8ad0L33 and here we are publishing to the same channel.

Secondly, we were closing channel on the receiver's side.

Azanul:
> > @MUzairS15 I couldn't see a race. Will you please point it out for me.
> 
> Basically, here in this function
> 
> https://github.com/meshery/meshery/blob/1949742026f4828ddbd3c5bf90d5008b51550c78/server/helpers/utils/utils.go#L112
> 
> 
> we are expecting value and, based on this, deciding whether the channel is closed or not. https://github.com/meshery/meshery/pull/7309/files#diff-c9ba355fc474472a3b057eed347eb447b73f8ae534b2277c78f1767ac04a8ad0L33 and here we are publishing to the same channel.

I don't think that's an issue since the select is non-blocking, an empty struct will be received making `IsClosed` function return `true` hence the line you mention never executes. I think that's why the `IsClosed` is used.


> Secondly, we were closing channel on the receiver's side.
Good point.

MUzairS15:
Suppose there are 2 instances of this subscription, in first instance we are signalling there is change in data and (say context switch happens ) in second instance we are checking for same channel is closed or not, since data was send to this channel by first instance the first case in select will be chosen returning true, indicating that channel is closed but in actual that channel is not closed


I might not be able to explain clearly but I can see the condition 😅, anyways I think this is better than the current implementation.

Azanul:
> Suppose there are 2 instances of this subscription, in first instance we are signalling there is change in data and (say context switch happens ) in second instance we are checking for same channel is closed or not, since data was send to this channel by first instance the first case in select will be chosen returning true, indicating that channel is closed but in actual that channel is not closed

Wouldn't both instances have there own channels? The access to channels cross-request is only available through `r.Config.ConfigurationChannel` methods, where every channel is being checked before being pushed to.
 
> I might not be able to explain clearly but I can see the condition sweat_smile, anyways I think this is better than the current implementation.
I agree to a certain degree.

Try this once if possible (detect-data-races)[https://yourbasic.org/golang/detect-data-races/]

PR #7314, opened 2023-03-09
create meshmodel helper

Azanul:
**Notes for Reviewers**

This PR fixes  https://github.com/meshery/meshery/pull/7299#issuecomment-1458568931


Azanul:
@leecalcote @Revolyssup

server/meshmodel/helper.go
```diff
@@ -0,0 +1,135 @@
...
+		case comp := <-ch.componentChan:
+			err = ch.regManager.RegisterEntity(meshmodel.Host{
+				Hostname: ArtifactHubComponentsHandler,
+			}, comp)
+		case rel := <-ch.relationshipChan:
+			err = ch.regManager.RegisterEntity(meshmodel.Host{
+				Hostname: ArtifactHubComponentsHandler,
+			}, rel)
+		case mhErr := <-ch.errorChan:
+			fmt.Println("err: ", mhErr.Error())
```
Aisuko, line 124: Thanks for your contribution, please make sure the same logic is in case `mhErr`.

Aisuko (changes requested):

server/meshmodel/helper.go
Azanul: I didn't get it

server/meshmodel/helper.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// seed the local meshmodel components
+func (ch ComponentHelper) SeedComponents() {
+	// Read component and relationship definitions from files and send them to channels
+	ch.generateComponents("/components")
+	ch.generateRelationships("/relationships")
+
+	// Register components and relationships with the registry manager
+	go ch.watchComponents()
```
Revolyssup, line 44: @Azanul Shouldn't this be called before generateComponentst? generateComponents is a synchronous call that will block on the componentChannel after pushing first component until the component is received. And unless you run watchcomponent, it wont be received.

server/meshmodel/helper.go
```diff
@@ -0,0 +1,135 @@
...
+			if err != nil {
+				return nil
+			}
+			ch.relationshipChan <- rel
+		}
+		return nil
+	})
+
+	ch.errorChan <- errors.Wrapf(err, "error while generating relationships")
+	ch.doneSignal <- true
```
Revolyssup, line 107: @Azanul Move this ch.doneSignal <- true to SeedComponent method at the end of both synchronous calls.

Revolyssup (changes requested):

server/meshmodel/helper.go
Azanul: You're right, my bad. I made them async first but then removed go and didn't notice this. How about executing both `generateComponents` and `generateRelationships` in their own go routines.

server/meshmodel/helper.go
```diff
@@ -0,0 +1,135 @@
...
+	ch.errorChan = make(chan error)
+}
+
+// seed the local meshmodel components
+func (ch ComponentHelper) SeedComponents() {
+	// Watch channels and register components and relationships with the registry manager
+	go ch.watchComponents()
+
+	// Read component and relationship definitions from files and send them to respective channels
+	ch.generateComponents("/components")
```
Revolyssup, line 43: use filepath.Join. Inside of the function you are calling for its Abs, which would just result in /components

server/meshmodel/helper.go
```diff
@@ -0,0 +1,135 @@
...
+		select {
+		case comp := <-ch.componentChan:
+			err = ch.regManager.RegisterEntity(meshmodel.Host{
+				Hostname: ArtifactHubComponentsHandler,
+			}, comp)
+		case rel := <-ch.relationshipChan:
+			err = ch.regManager.RegisterEntity(meshmodel.Host{
+				Hostname: ArtifactHubComponentsHandler,
+			}, rel)
+		case mhErr := <-ch.errorChan:
```
Revolyssup, line 123: @Azanul nil check?

Revolyssup (changes requested):

server/meshmodel/helper.go
Azanul: nil checks are already there before pushing err to the channel

Revolyssup: Nope. They are not. https://github.com/meshery/meshery/pull/7314/files#diff-ab4499e6d5cf39814e6b6dd3607e23566fe60b939e1bb2fbb068a832003046c4R77

Revolyssup: And even if they were. The consumer should always be responsible for nil checks to avoid panic

server/meshmodel/helper.go
```diff
@@ -0,0 +1,135 @@
...
+	handlerConfig    *models.HandlerConfig
+	regManager       *meshmodel.RegistryManager
+	componentChan    chan v1alpha1.ComponentDefinition
+	relationshipChan chan v1alpha1.RelationshipDefinition
+	doneSignal       chan bool
+	errorChan        chan error
+}
+
+func (ch *ComponentHelper) Init(hc *models.HandlerConfig, rm *meshmodel.RegistryManager) {
+	ch.handlerConfig = hc
```
Revolyssup, line 29: @Azanul No need to pass the handlerconfig here. Execute the SeedComponents inside an anonymous func go routine and once it returns, Call publish. No need to use doneChannel. The original code was already abusing go channels. I should have refactored before

Revolyssup (changes requested):

server/meshmodel/helper.go
```diff
@@ -0,0 +1,135 @@
...
+	err = filepath.Walk(path, func(path string, info fs.FileInfo, err error) error {
+		if info == nil {
+			return nil
+		}
+
+		if !info.IsDir() {
+			var comp v1alpha1.ComponentDefinition
+			byt, err := os.ReadFile(path)
+			if err != nil {
+				return err
```
Revolyssup, line 64: @Azanul Do not return anywhere while walking the tree so that error in one file doesn't stop execution. Just push to the error channel and continue.

Revolyssup (changes requested):

server/meshmodel/helper.go
Azanul: I think it's better encapsulated this way

Revolyssup: It is not the responsibility of this function to publish anything. Passing a handlerconfig is pointless. Main should handle that. It's unnecessary complexity. This function should do one job and that job properly.

Revolyssup: There is no further synchronization routine needed after registration ends. Hence we dont need channels. A simple rule of thumb is to ask "Do we need synchronisation here". What you are trying to synchronise is effectively same as returning at the end a boolean value. Therefore channels are being abused here. Normal returns take place on stack,  channels require heap. Unnecessary.

server/meshmodel/helper.go
```diff
@@ -0,0 +1,137 @@
...
+	// Watch channels and register components and relationships with the registry manager
+	go ch.watchComponents()
+
+	// Read component and relationship definitions from files and send them to respective channels
+	ch.generateComponents("/components")
+	ch.generateRelationships("/relationships")
+	ch.doneSignal <- true
+}
+
+func (ch ComponentHelper) generateComponents(pathToComponents string) {
```
Revolyssup, line 50: Please make these methods for Pointer to ComponentHelper. You are lucky that all variables are effectively pointers so a shallow copy didn't hurt. But please make it a pointer to the methods, especially when you are communicating values in the same instance of the struct.

Revolyssup (changes requested):

server/meshmodel/helper.go
Azanul: How do you suggest closing `watchComponents` ?

server/meshmodel/helper.go
Azanul: Wouldn't just removing the "/" do the job?

server/meshmodel/helper.go
Azanul: > The consumer should always be responsible for nil checks to avoid panic

I'd agree in case of fetching data from an external source. Why would we follow this in case of internal system?

Revolyssup: nil checking multiple places vs nil checking once. We have multiple producers and one consumer.

server/meshmodel/helper.go
Revolyssup: No. Abs() uses working directory in that case. The working directory is /server/cmd. 
Abs() will return /server/cmd/components. do filepath.Abs(filepath.Join("../meshmodel","components"))

Revolyssup (changes requested):

Azanul:
>nil checking multiple places vs nil checking once. We have multiple producers and one consumer.

It's not a "vs" case. We have to nil check err at most of the producing points. Adding the same to consumer will reduce 2 line of visible code and add an extra a nil check everywhere else.

Azanul:
>No. Abs() uses working directory in that case. The working directory is /server/cmd.
Abs() will return /server/cmd/components. do filepath.Abs(filepath.Join("../meshmodel","components"))

So sorry for this blunder. I'll fix it right away. Instead of using join, "../meshmodel/components" would also work exactly the same I think.

server/meshmodel/helper.go
Aisuko: Should we set the error to the variable the code can handle at the end?

Azanul: This is watching all the errors while generating components and relationships. If I understand you correctly, you're suggesting handling this error like returning it to calling functions, but that is not the purpose of it. It'll run asynchronously, watch errors and print them to console.

Aisuko: Ok, I see. You can ignore my suggestion.

Aisuko: Please add a code comment to mention this in your next commit, thanks.

server/meshmodel/helper.go
Revolyssup: use context (create one and pass on watchComponents. Wait on ctx.Done() and call done() defer done() in SeedComponents)

Revolyssup:
> > nil checking multiple places vs nil checking once. We have multiple producers and one consumer.
> 
> It's not a "vs" case. We have to nil check err at most of the producing points. Adding the same to consumer will reduce 2 line of visible code and add an extra a nil check everywhere else.

what is meant by "have to nil check at producers?". Producers do not know how those errors are going to be used. There job shouldn't be to do checks. This will be code duplication at multiple places for no reason. The consumer is responsible for doing something with that error and error nil check should be done as close to calling methods on that error as possible to avoid unnecessary panics. Why to enforce any nil check on producer side? You would effectively create a channel where "only non-nil errors should be passed" with no enforcement of this rule at runtime or compile time. It's just left to the programmer to figure out. Why to have such implicit rules? Make things explicit by not caring whether errors are nil or non-nil and passing them on the channel. At one point, at one place, the nil check will take place.

server/meshmodel/helper.go
Revolyssup: Whenever the question comes "how to make sure X exits when Y exits?". The first thing that should pop in the head is passing context and not using state variables. (Unless it is a complicated use case with multiple states interacting with each other.)

server/meshmodel/helper.go
```diff
@@ -0,0 +1,144 @@
...
+		case comp := <-ch.componentChan:
+			err = ch.regManager.RegisterEntity(meshmodel.Host{
+				Hostname: ArtifactHubComponentsHandler,
+			}, comp)
+		case rel := <-ch.relationshipChan:
+			err = ch.regManager.RegisterEntity(meshmodel.Host{
+				Hostname: ArtifactHubComponentsHandler,
+			}, rel)
+		case mhErr := <-ch.errorChan:
+			fmt.Println("err: ", mhErr.Error())
```
Revolyssup, line 133: @Azanul Do not use println. Create appropriate meshkit error. And pass the meshkit logger initialized in main.go to ComponentHelper struct.

server/meshmodel/helper.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/layer5io/meshery/server/helpers/utils"
+	"github.com/layer5io/meshery/server/models"
+	"github.com/layer5io/meshkit/models/meshmodel"
+	"github.com/layer5io/meshkit/models/meshmodel/core/v1alpha1"
+	"github.com/pkg/errors"
+)
+
+const ArtifactHubComponentsHandler = "kubernetes" //The components generated in output directory will be handled by kubernetes
+
+type ComponentHelper struct {
```
Revolyssup, line 19: Rename to EntityRegistrationHelper

Revolyssup (changes requested):

server/meshmodel/helper.go
Azanul: context it is then

Azanul:
> what is meant by "have to nil check at producers?". Producers do not know how those errors are going to be used. There job shouldn't be to do checks. 

To check whether an error has occurred or not, we have to check if err is nil or not that is what is meant by "have to nil check at producers". 

>This will be code duplication at multiple places for no reason.

 Adding the same to consumer will reduce 2 line of visible code and add an extra a nil check everywhere else. 

>The consumer is responsible for doing something with that error and error nil check should be done as close to calling methods on that error as possible to avoid unnecessary panics. Why to enforce any nil check on producer side? You would effectively create a channel where "only non-nil errors should be passed" with no enforcement of this rule at runtime or compile time. It's just left to the programmer to figure out. Why to have such implicit rules? Make things explicit by not caring whether errors are nil or non-nil and passing them on the channel. At one point, at one place, the nil check will take place.

Entire code is left to the programmer to figure out. I can put in the nil check at the consumer too if you want, it'll make sure that no nil but just notice that in most error case it'll check err twice.

Revolyssup:
> > what is meant by "have to nil check at producers?". Producers do not know how those errors are going to be used. There job shouldn't be to do checks.
> 
> To check whether an error has occurred or not, we have to check if err is nil or not that is what is meant by "have to nil check at producers".
> 
> > This will be code duplication at multiple places for no reason.
> 
> Adding the same to consumer will reduce 2 line of visible code and add an extra a nil check everywhere else.
> 
> > The consumer is responsible for doing something with that error and error nil check should be done as close to calling methods on that error as possible to avoid unnecessary panics. Why to enforce any nil check on producer side? You would effectively create a channel where "only non-nil errors should be passed" with no enforcement of this rule at runtime or compile time. It's just left to the programmer to figure out. Why to have such implicit rules? Make things explicit by not caring whether errors are nil or non-nil and passing them on the channel. At one point, at one place, the nil check will take place.
> 
> Entire code is left to the programmer to figure out. I can put in the nil check at the consumer too if you want, it'll make sure that no nil but just notice that in most error case it'll check err twice.

How are there two nil checks????????

Azanul:
> How are there two nil checks???????? Do not check on producer. Only check at consumer. One nil check. What am I missing?

1. https://github.com/meshery/meshery/pull/7314/files#diff-ab4499e6d5cf39814e6b6dd3607e23566fe60b939e1bb2fbb068a832003046c4R55

2. https://github.com/meshery/meshery/pull/7314/files#diff-ab4499e6d5cf39814e6b6dd3607e23566fe60b939e1bb2fbb068a832003046c4R138

Revolyssup:
> > How are there two nil checks???????? Do not check on producer. Only check at consumer. One nil check. What am I missing?
> 
> 1. https://github.com/meshery/meshery/pull/7314/files#diff-ab4499e6d5cf39814e6b6dd3607e23566fe60b939e1bb2fbb068a832003046c4R55
> 2. https://github.com/meshery/meshery/pull/7314/files#diff-ab4499e6d5cf39814e6b6dd3607e23566fe60b939e1bb2fbb068a832003046c4R138

Name the channel something like `nonNilErrOnly`. And leave the redundant err checking at consumer .  It's okay to have this much code duplication rather than having an accidental panic.

Azanul:
> Name the channel something like `nonNilErrOnly`. And leave the redundant err checking at consumer . It's okay to have this much code duplication rather than having an accidental panic.

Understood. Leaving the renaming thing as is since we've decided to keep the nil check at consumer.

server/meshmodel/helper.go
```diff
@@ -0,0 +1,150 @@
...
+
+// seed the local meshmodel components
+func (erh *EntityRegistrationHelper) SeedComponents() {
+	// Watch channels and register components and relationships with the registry manager
+	ctx, cancel := context.WithCancel(context.TODO())
+	go erh.watchComponents(ctx)
+
+	// Read component and relationship definitions from files and send them to respective channels
+	erh.generateComponents("../meshmodel/components")
+	erh.generateRelationships("../meshmodel/components")
```
Revolyssup, line 49: Please change it to relationships

server/meshmodel/helper.go
```diff
@@ -0,0 +1,150 @@
...
+	cancel()
```
Revolyssup, line 50: Defer it on top

Revolyssup (changes requested):

Azanul:
@Revolyssup  GTG

server/meshmodel/helper.go
```diff
@@ -0,0 +1,151 @@
...
+		handlerConfig:    hc,
+		regManager:       rm,
+		componentChan:    make(chan v1alpha1.ComponentDefinition, 1),
+		relationshipChan: make(chan v1alpha1.RelationshipDefinition, 1),
+		errorChan:        make(chan error),
+		log:              log,
+	}
+}
+
+// seed the local meshmodel components
```
leecalcote, line 41: Code comments. 😍

server/meshmodel/helper.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		erh.errorChan <- errors.Wrapf(err, "error while generating components")
+	}
+	return
+}
+
+func (erh *EntityRegistrationHelper) generateRelationships(pathToComponents string) {
```
leecalcote, line 91: Code comments? 😢

Azanul: got that frown upside down?

leecalcote: Oh, very good. We have code comments. 🙂

leecalcote (approved):

leecalcote:
Syncing...

PR #7350, opened 2023-03-14
Changed integrations filetype for meshery.io repo from json to js

asubedy:
**Notes for Reviewers**
I have updated the script to update integrations on the meshery.io website, now the file which is saved on the repo is `data.js` instead of `data.json` with few changes required inside the file.

Reason for doing this:
>Because in vanilla JS if we try to access json files we need to write
import JSON from './data.json' assert {type: json}
This is will give us an error in Safari, assert not supported properly in safari
This PR fixes #

[Link to slack discussion](https://layer5io.slack.com/archives/C015QJKUMPU/p1678747178467119?thread_ts=1678735714.080779&cid=C015QJKUMPU)

[Link to PR that showed that `assert` does not work on safari](https://github.com/meshery/meshery.io/pull/1022#issuecomment-1433871342)

### DIFF
--- a/scripts/component_updation/main.go
+++ b/scripts/component_updation/main.go
@@ -98,7 +98,7 @@ func main() {
 		file.Close()
 		os.Remove(file.Name())
 		output = cleanupDuplicatesAndPreferEmptyComponentField(output, "model")
-		mesheryDocsJSON := "["
+		mesheryDocsJSON := "const data = ["
 		for _, out := range output {
 			var t pkg.TemplateAttributes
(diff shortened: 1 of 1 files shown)


Revolyssup:
If this will still stay in one file then what is the use of json->js ? Wasn't the purpose to split them?

leecalcote (approved):

asubedy:
> If this will still stay in one file then what is the use of json->js ? Wasn't the purpose to split them?

I didn't get your question @Revolyssup

PR #7412, opened 2023-03-29
[chore] Send Meshery Version

theBeginner86:
**Notes for Reviewers**

This PR adds support for sending Meshery version to remote provider on initiating login process with it.

This change is not necessary, as Meshery version is already sent a step before this, that is when `capabilities` are requested. But this would make the process fail safe because request to capabilities is sent only once when Meshery boots up and if due to some reason remote_provider reboots in the middle, then that `meshery_version` is lost.

This would ensure that whenever a user tries to login, then we would be comparing the correct `current_version`
This PR fixes #


leecalcote (approved):
What is the expected behavior here when the Remote Provider is unavailable, @theBeginner86?

leecalcote:
I have seen tis issue occurring in the past, so it's nice to see a PR here.

theBeginner86:
> What is the expected behavior here when the Remote Provider is unavailable, @theBeginner86?

If remote provider is not available then, we won't be able to login anyways. So, not an issue if remote provider is unavailable.

leecalcote:
Makes sense. I meant to ask, is will this fail silently? In the event that the remote provider doesn’t support this… fail forward?

PR #7463, opened 2023-04-03
Modify prettification logic for end-user fields

Revolyssup:
**Notes for Reviewers**

This PR fixes prettification logic. Refer to the comment in the file modified for more context.

The logic and principle for prettification/deprettification.
1. Specific considerations are made when schema is passed to be prettified like handling kubernetes specific fields.
2. A general rule of thumb is to never prettify or deprettify the end-user fields, the ones which are entered by USER.
For non schema files, it would be all end string fields (user input) and for schema files it would be ENUMS as they are the only system defined fields that are used at end user input.


leecalcote (approved):
LGTM

PR #7466, opened 2023-04-03
[Relationship] Enable node connection via some intermediary node

Abhishek-kumar09:
**Notes for Reviewers**
1. Supports enabling connection between two entity via third entity, e.g Pod to PV via PVC
2. Added the schema support as well

This PR fixes #

### DIFF
--- a/server/meshmodel/schemas/relationship.json
+++ b/server/meshmodel/schemas/relationship.json
@@ -71,6 +71,11 @@
                     "$ref": "https://schema.meshery.io/core.json#versionString"
                   }
                 }
+              },
+              "allowMultiple": {
+                "type": "boolean",
+                "default": true,
+                "description": "Whether to allow multiple connections of same type between the same two components"
               }
             },
             "description": "Optional fields that are a part of the `from` selector. Absence of a field has an implied * meaning."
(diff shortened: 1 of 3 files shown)


server/meshmodel/schemas/relationship.json
```diff
@@ -79,8 +94,13 @@
             },
             "description": "Optional fields that are a part of the `to` selector. Absence of a field has an implied * meaning."
           }
+        },
+        "allowMultiple": {
```
Abhishek-kumar09, line 98: This helps us in knowing whether between the same two component multiple connection of same type is available,

MUzairS15: After reading the description it seems like there will 2 edges, pod to pvc and then pvc to pv.
Or this schema update is just facilitating the current behaviour we have

Abhishek-kumar09: Although we want to have some modification in the code written here. But to your question, it just tells about "via" node, and its on UI implementation to chose any of the two approaches you're mentioning.

leecalcote:
While we're not taking this specific approach, @MUzairS15 @Abhishek-kumar09 this is still an open concern.

PR #7513, opened 2023-04-08
[Server] Credential crud

Azanul:
**Notes for Reviewers**

This PR adds credential CRUD, same as meshery cloud.

### DIFF
--- a/server/models/remote_provider.go
+++ b/server/models/remote_provider.go
@@ -3387,7 +3387,7 @@ func (l *RemoteProvider) SaveUserCredential(req *http.Request, credential *Crede
 }
 
 // GetCredentials - to get saved credentials
-func (l *RemoteProvider) GetUserCredentials(req *http.Request, _, page, pageSize, search, order string) (*CredentialsPage, error) {
+func (l *RemoteProvider) GetUserCredentials(req *http.Request, _ string, page, pageSize int, search, order string) (*CredentialsPage, error) {
 	if !l.Capabilities.IsSupported(PersistCredentials) {
 		logrus.Error("operation not available")
 		return nil, ErrInvalidCapability("PersistCredentials", l.ProviderName)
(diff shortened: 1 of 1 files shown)


theBeginner86 (commented):
Good stuff 👍

Few things to consider:
1. Credentials is a general capability of Meshery that is specific to the provider of the Meshery user. So logic of these handlers would go inside the [provider](https://github.com/meshery/meshery/blob/master/server/models/providers.go) interface. That is, these handlers' logic (in this PR) would be just necessary for [local provider](https://github.com/meshery/meshery/blob/master/server/models/default_local_provider.go)
2. There is a need to ensure that `Credentials` table is created when Meshery boots up (for local provider).
3. We need to expose credential as a capability of remote provider. _I tagged you in that recent PR where I exposed this_
4. Credentials is resource of group Integrations. So, the endpoint structure would be like: `/api/integrations/credentials`

To know more about how to ensure that handlers specific to the provider chosen by Meshery user, you can have a look at [other](https://github.com/meshery/meshery/blob/master/server/models/providers.go#L212) remote provider capabilities like `performance`, `patterns`, `applications`, `connections` etc.

Would you please make necessary changes?

Azanul:
> Would you please make necessary changes?

Check it out

leecalcote:
cha-cha-chickety-check it out, yo.

Chadha93:
@Azanul  Let's discuss this on the Meshery Dev call. Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit#) if you would. :)

leecalcote:
@theBeginner86 and/or @MUzairS15 will need to review.

server/models/remote_provider.go
```diff
@@ -3400,3 +3400,43 @@ func (l *RemoteProvider) SetKubeClient(client *mesherykube.Client) {
 func (l *RemoteProvider) GetKubeClient() *mesherykube.Client {
 	return l.KubeClient
 }
+
+// SaveCredential - to save a creadential for an integration
+// TODO
```
theBeginner86, line 3405: @Azanul 

Why is this remote_provider support left out? Are you about to add this?

Azanul: @theBeginner86 

There you go

PR #7523, opened 2023-04-10
Add workflow to perform nightly cleanup on playground cluster

Revolyssup:
**Notes for Reviewers**

This PR adds a workflow which runs nightly and removes all namespaces except for the ones specified in the if statement.
NOTE: Users might be thrown of if they are on playground and a namespace or resources in a namespace suddenly go away as this script runs.

### DIFF
--- a/.github/workflows/reset-playground.yaml
+++ b/.github/workflows/reset-playground.yaml
@@ -1,4 +1,4 @@
-name: Reset running Meshery playground
+name: Meshery Playground: Reset
 on:
    workflow_dispatch:
    schedule:
@@ -6,6 +6,8 @@ on:
 
 jobs:
    reset:
+     name: Cluster Reset
+     if: github.repository == 'meshery/meshery'
      runs-on: ubuntu-latest 
      steps:
       - name: Install kubectl


leecalcote:
@alaeddinebenhassir your eyes here would be helpful. This might be the first version of the cleanup process.

.github/workflows/reset-playground.yaml
```diff
@@ -0,0 +1,24 @@
+name: Reset running Meshery playground
+on:
+   workflow_dispatch:
+   schedule:
+     - cron:  '0 0 * * *'  # Run the job every day at midnight
+
+jobs:
+   reset:
+     runs-on: ubuntu-latest
```
leecalcote, line 9:
```suggestion
     name: Cluster Reset
      if: github.repository == 'meshery/meshery'
     runs-on: ubuntu-latest 
```

Only allow to be run from this repo.

.github/workflows/reset-playground.yaml
```diff
@@ -0,0 +1,24 @@
+name: Reset running Meshery playground
```
leecalcote, line 1:
```suggestion
name: Meshery Playground: Reset
```

leecalcote (dismissed):
GTG given the suggested changes.

leecalcote (approved):

PR #7548, opened 2023-04-14
Updated EventStreamHandler for REST API docs

acald-creator:
**Notes for Reviewers**

This PR updates the swagger generation, adding EventStreamHandler entry with swagger

### DIFF
--- a/server/handlers/events_streamer.go
+++ b/server/handlers/events_streamer.go
@@ -25,7 +25,7 @@ var (
 // swagger:route GET /api/events EventsAPI idGetEventStreamer
 // Handle GET request for events.
 // Listens for events across all of Meshery's components like adapters and server, streaming them to the UI via Server Side Events
-// This API call never terminates and establishes a persistent keep-alive connection over which EventsResponse is sent back.
+// This API call never terminates and establishes a persistent keep-alive connection over which `EventsResponse`s are pushed.
 // responses:
 // 	200: EventsResponse
 


server/helpers/swagger.yaml
```diff
@@ -242,6 +242,16 @@ paths:
             summary: Handle POST request for Sharing content
             tags:
                 - ShareContent
+    /api/events/events-streamer:
+        get:
+            description: Fetches events from Remote Provider and streams it to the UI
+            operationId: idGetEventStreamer
+            responses:
+                "200":
+                    description: ""
```
leecalcote, line 251: @Revolyssup what do we have for a description here?

acald-creator: There aren't any descriptions among the other endpoints, but there is a general description anyway on the handler itself. If we need to add a description, this should be the same for the other endpoints too

leecalcote: @acald-creator you're so very much right: all of the endpoints need additional documentation. As an extensible platform, Meshery needs very well-documented APIs.  Yes, please.

server/handlers/events_streamer.go
```diff
@@ -22,6 +22,13 @@ var (
 	flusherMap map[string]http.Flusher
 )
 
+// swagger:route GET /api/events/events-streamer EventsAPI idGetEventStreamer
+// Handle GET request for events streamer
+//
+// Fetches events from Remote Provider and streams it to the UI
+// responses:
+// 	200:
+
```
Azanul, line 31: No parameters listed and no response object definition?

Azanul (commented):
.

server/handlers/events_streamer.go
acald-creator: Most of the others are like this.

This is why I mentioned that if this is going to be the case here, this type of change needs to be reflected across the rest of the endpoints.

server/handlers/events_streamer.go
```diff
@@ -22,6 +22,13 @@ var (
 	flusherMap map[string]http.Flusher
 )
 
+// swagger:route GET /api/events EventsAPI idGetEventStreamer
+// Handle GET request for events.
+// Listens for events across all of Meshery's components like adapters and server, streaming it to the UI
+// This API call never terminates and establishes a persistent keep-alive connection over which EventsResponse is sent back.
```
leecalcote, line 28:
```suggestion
// This API call never terminates and establishes a persistent keep-alive connection over which `EventsResponse`s are pushed.
```

l5io:
🚀 Preview for commit 86e758bee965b7362ebdf9341494df213933f234 at: https://643d5ab5b5cbf51d24482f7e--meshery-docs-preview.netlify.app

PR #7569, opened 2023-04-18
[Server] Kubernetes-based support for De/provisioning Meshery Adapters á la carte

Azanul:
**Notes for Reviewers**

This PR adds adapter deploy/undeploy support for K8s host.

### DIFF
--- a/server/helpers/adapters_tracker.go
+++ b/server/helpers/adapters_tracker.go
@@ -2,6 +2,7 @@ package helpers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -250,6 +251,11 @@ func (a *AdaptersTracker) setOverrideValues(adapterName string, exposedPort int,
 func (a *AdaptersTracker) applyHelmCharts(kubeClient *meshkitkube.Client, adapterName string, exposedPort int, enable bool) error {
 	// get value overrides to install the helm chart
 	overrideValues := a.setOverrideValues(adapterName, exposedPort, enable)
+	b, err := json.MarshalIndent(overrideValues, "", "  ")
+	if err != nil {
+		fmt.Println("error:", err)
+	}
+	fmt.Print(string(b))
 
 	// install the helm charts with specified override values
 	return kubeClient.ApplyHelmChart(meshkitkube.ApplyHelmChartConfig{
@@ -266,13 +272,11 @@ func (a *AdaptersTracker) applyHelmCharts(kubeClient *meshkitkube.Client, adapte
 	})
 }
 
-func setService(name string, port int) *corev1.Service {
-	return &corev1.Service{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
-			Namespace: core.MesheryNamespace,
-		},
-		Spec: corev1.ServiceSpec{
+func setService(name string, port int) map[string]interface{} {
+	return map[string]interface{}{
+		"name":      name,
+		"namespace": core.MesheryNamespace,
+		"spec": corev1.ServiceSpec{
 			Selector: map[string]string{
 				"app": name,
 			},


Chadha93:
@Azanul  Let's discuss this on the Meshery Dev call. Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit#) if you would. :)

leecalcote:
@meshery/adapter-maintainers please review.

server/helpers/adapters_tracker.go
```diff
@@ -119,11 +124,90 @@ func (a *AdaptersTracker) DeployAdapter(ctx context.Context, adapter models.Adap
...
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: map[string]string{
+							"app": adapter.Name,
+						},
+					},
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{
+								Name:  adapter.Name,
+								Image: "layer5/" + adapter.Name + ":stable-latest",
```
leecalcote, line 158: @Azanul is this same stable-latest assumption already being made in the current deployments?

server/helpers/adapters_tracker.go
```diff
@@ -119,11 +124,90 @@ func (a *AdaptersTracker) DeployAdapter(ctx context.Context, adapter models.Adap
...
+		if err != nil {
+			return ErrAdapterAdministration(err)
+		}
+
 	// switch to default case if the platform specified is not supported
 	default:
 		return ErrAdapterAdministration(fmt.Errorf("the platform %s is not supported currently. The supported platforms are:\ndocker\nkubernetes", platform))
 	}
 
+	adapter.Location = "localhost:" + adapter.Location
```
leecalcote, line 210: We can’t assume “localhost”.

server/helpers/adapters_tracker.go
```diff
@@ -119,11 +124,90 @@ func (a *AdaptersTracker) DeployAdapter(ctx context.Context, adapter models.Adap
...
+				Name:      adapter.Name,
+				Namespace: core.MesheryNamespace,
+			},
+			Spec: corev1.ServiceSpec{
+				Selector: map[string]string{
+					"app": adapter.Name,
+				},
+				Ports: []corev1.ServicePort{
+					{
+						Name:     "http",
```
leecalcote, line 187: Is the protocol not gRPC?

leecalcote (changes requested):

server/helpers/adapters_tracker.go
Azanul: @leecalcote No, currently the version is being inferred from the config file by meshery version

server/helpers/adapters_tracker.go
Azanul: Because it won't work in a multi-node cluster where meshery server and deployed adapter are different node?

server/helpers/adapters_tracker.go
Azanul: You're right, it is gRPC

server/helpers/adapters_tracker.go
leecalcote: @Azanul that is fantastic. 💯

server/helpers/adapters_tracker.go
```diff
@@ -119,11 +125,99 @@ func (a *AdaptersTracker) DeployAdapter(ctx context.Context, adapter models.Adap
 			return ErrAdapterAdministration(err)
 		}
 
+	case "kubernetes":
+		kubeClient, err := meshkitkube.New([]byte(""))
```
leecalcote, line 129: @Azanul we're using the helm client, yes? Ideally, we're using the exact same mechanism to deploy Meshery Adapters as we are to deploy Meshery Operator.

server/helpers/adapters_tracker.go
```diff
@@ -119,11 +125,99 @@ func (a *AdaptersTracker) DeployAdapter(ctx context.Context, adapter models.Adap
...
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: map[string]string{
+							"app": adapter.Name,
+						},
+					},
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{
+								Name:  adapter.Name,
+								Image: "layer5/" + adapter.Name + ":stable-latest",
```
leecalcote, line 166: Ideally, we don't presume the container registry org name. Let's externalize this configuration, if we can.

leecalcote: Doing so might require a follow up PR.

server/helpers/adapters_tracker.go
```diff
@@ -171,6 +265,28 @@ func (a *AdaptersTracker) UndeployAdapter(ctx context.Context, adapter models.Ad
...
+			adapter = a.adapters[adapter.Location]
+		}
+
+		// Delete the Deployment
+		err = kubeClient.KubeClient.AppsV1().Deployments(core.MesheryNamespace).Delete(context.Background(), adapter.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return ErrAdapterAdministration(err)
+		}
+
+		// Delete the associated Service objects
```
leecalcote, line 284: The collective sequence that we have here is a bit imperative. Ideally, we are simply reusing Helm client to hand Kubernetes a manifest (a chart) and Helm/Kubernetes does the rest.

server/helpers/adapters_tracker.go
Azanul: The version here is hardcoded because meshery server doesn't have that config file.

server/helpers/adapters_tracker.go
Azanul: kubernetes client not helm

server/helpers/adapters_tracker.go
Azanul: Externalise as in? take through user input?

server/helpers/adapters_tracker.go
Azanul: I've update the code to use Helm client. I don't think this way the adapter is completely decoupled from meshery.

Revolyssup (changes requested):

server/helpers/adapters_tracker.go
```diff
@@ -179,3 +213,91 @@ func (a *AdaptersTracker) UndeployAdapter(ctx context.Context, adapter models.Ad
 	a.RemoveAdapter(ctx, adapter)
 	return nil
 }
+
+func int32Ptr(n int32) *int32 {
```
Revolyssup, line 217: Where is it being used?

Revolyssup (changes requested):

server/helpers/adapters_tracker.go
leecalcote, line 217: Where is this used?

server/helpers/adapters_tracker.go
```diff
@@ -179,3 +213,91 @@ func (a *AdaptersTracker) UndeployAdapter(ctx context.Context, adapter models.Ad
...
+			Repository: utils.HelmChartURL,
+			Chart:      utils.HelmChartName,
+		},
+		Action:         meshkitkube.INSTALL,
+		DryRun:         false,
+		OverrideValues: overrideValues,
+	})
+}
+
+func setService(name string, port int) map[string]interface{} {
```
Revolyssup, line 285: Why can't we use the Service struct from k8s.io/api for this? Instead of manually creating map[string]interface{}

Revolyssup (changes requested):

server/helpers/adapters_tracker.go
```diff
@@ -171,6 +194,17 @@ func (a *AdaptersTracker) UndeployAdapter(ctx context.Context, adapter models.Ad
 			return ErrAdapterAdministration(err)
 		}
 
+	case "kubernetes":
+		kubeClient, err := meshkitkube.New([]byte(""))
+		if err != nil {
+			return ErrAdapterAdministration(err)
```
leecalcote, line 200: ErrAdapterAdministration() is missing error-specific probable cause and suggested remediation.

So, too, is ErrDockerHost()

leecalcote (changes requested):

server/helpers/adapters_tracker.go
```diff
@@ -75,6 +78,7 @@ func (a *AdaptersTracker) DeployAdapter(ctx context.Context, adapter models.Adap
 	}
 
 	// Deploy to current platform
+	targetIP := "127.0.0.1"
```
leecalcote, line 81: Have you confirmed that this change isn't going to cause a regression issue with Docker as the platform? What's the need for change from "localhost"?

server/helpers/adapters_tracker.go
```diff
@@ -179,3 +213,91 @@ func (a *AdaptersTracker) UndeployAdapter(ctx context.Context, adapter models.Ad
...
+	// Initialize the valueOverrides with default values and set config for selected adapter
+	valueOverrides := map[string]interface{}{}
+	for _, availableAdapter := range models.ListAvailableAdapters {
+		valueOverrides[availableAdapter.Name] = map[string]interface{}{
+			"enabled": false,
+		}
+		if enable && availableAdapter.Name == adapterName {
+			valueOverrides[availableAdapter.Name] = map[string]interface{}{
+				"enabled": true,
+				"service": map[string]interface{}{
```
Revolyssup, line 232: Why is that function used over there and not here?

Revolyssup (changes requested):

server/helpers/adapters_tracker.go
```diff
@@ -179,3 +213,91 @@ func (a *AdaptersTracker) UndeployAdapter(ctx context.Context, adapter models.Ad
...
+							"app": adapterName,
+						},
+						"ports": []map[string]interface{}{
+							{
+								"name":     "gRPC",
+								"protocol": "TCP",
+								"port":     exposedPort,
+							},
+						},
+						"type": "ClusterIP",
```
leecalcote, line 246: Instead of building this out by hand, so to speak, is there no existing struct / helm function to reference?

server/helpers/adapters_tracker.go
Azanul: Not used anymore, was using previously

server/helpers/adapters_tracker.go
leecalcote, line 81: Also, see line 73, "AddAdapter" should be "DeployAdapter"

server/helpers/adapters_tracker.go
```diff
@@ -119,11 +122,31 @@ func (a *AdaptersTracker) DeployAdapter(ctx context.Context, adapter models.Adap
...
+		}
+
+		targetIP = svc.Spec.ClusterIP
+
 	// switch to default case if the platform specified is not supported
 	default:
 		return ErrAdapterAdministration(fmt.Errorf("the platform %s is not supported currently. The supported platforms are:\ndocker\nkubernetes", platform))
 	}
 
+	adapter.Location = targetIP + ":" + adapter.Location
```
leecalcote, line 149: Is this how we have provisioned components of Meshery Operator? There's no need for TargetIP. Use the adapter name instead. Rely on DNS inside of Kubernetes and Docker-based deployments.

server/internal/graphql/resolver/adaptor.go
leecalcote, line 1: Please correct the misspelling of this filename.

server/helpers/adapters_tracker.go
Azanul: ErrDockerHost() does have
ErrAdapterAdministration() is generic, similar to ErrNewKubeClient()

server/internal/graphql/resolver/adaptor.go
Azanul: done

server/helpers/adapters_tracker.go
Azanul: fixed

Azanul:
@leecalcote Closing this since a different approach is to be taken

leecalcote:
Roger that, @Azanul. Thank you.

PR #7592, opened 2023-04-22
query the userDetails from remoteProvider ```/api/user```

harkiratsm:
**Notes for Reviewers**

This PR fixes #
Enhance the User Details by fetching the details from the remote provider. We now have user_roles along with user data.


server/handlers/user_handler.go
```diff
@@ -9,21 +9,24 @@ import (
...
-		h.log.Error(ErrEncoding(err, obj))
-		http.Error(w, ErrEncoding(err, obj).Error(), http.StatusInternalServerError)
+func (h *Handler) UserHandler(w http.ResponseWriter, r *http.Request, _ *models.Preference, user *models.User, provider models.Provider) {
+	resp, err := provider.GetUserDetails(r)
+	if err != nil {
+		h.log.Error(ErrGetResult(err))
+		http.Error(w, ErrGetResult(err).Error(), http.StatusNotFound)
+		return
+	}
+	fmt.Println(resp)
```
MUzairS15, line 25: Remove this please.

server/handlers/user_handler.go
```diff
@@ -9,21 +9,24 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 
 	"github.com/layer5io/meshery/server/models"
 )
 
 // UserHandler returns info about the logged in user
-func (h *Handler) UserHandler(w http.ResponseWriter, _ *http.Request, _ *models.Preference, user *models.User, _ models.Provider) {
```
MUzairS15, line 17: Why remove this?
In session middleware we are already fetching the user detail and passing, why can’t we use that??
https://github.com/meshery/meshery/blob/f7d0a28a250e7c9a078601e864aa68486b349ae2/server/handlers/middlewares.go#L232

MUzairS15 (changes requested):

server/handlers/user_handler.go
harkiratsm: Firstly, it is not removed, and the reason I don't use the Session middleware is that it serves us a different use case.

MUzairS15: By removing, I meant why you are removing the caching logic?
Also, you have the user details (struct); you are just re-fetching, and I don't understand why.
func (h *Handler) UserHandler(w http.ResponseWriter, r *http.Request, _ *models.Preference, ------>_ *models.User, provider models.Provider)

server/models/remote_provider.go
```diff
@@ -289,25 +289,13 @@ func (l *RemoteProvider) fetchUserDetails(tokenString string) (*User, error) {
 		return nil, ErrDataRead(err, "User Data")
 	}
 
-	up := &UserPref{
-		Preferences: &Preference{
-			AnonymousUsageStats:  true,
-			AnonymousPerfResults: true,
```
MUzairS15, line 295: Any reasons for this, does it prevent saving performance results, because we want to be greedy to have statistics recorded.

harkiratsm: No reason, I thought preferences could be controlled on the UI side.

MUzairS15: Yes, they are controlled in UI. This is also fine

harkiratsm: I added it back since, you wanted to enforce certain behaviour or preferences .

MUzairS15 (approved):

PR #7629, opened 2023-04-28
[Wasm] Initialise go wasm to be used in the Meshery-UI and Meshery-extensions

Abhishek-kumar09:
**Notes for Reviewers**
Need for wasm:
1. Some complex decisions like schema verification and its alterations is being done in JS that makes the code less performant, neither we can rely on server for its execution because of network trip, rather it is required to be executed via our wasm binaries that runs them in the browser itself in a type-safe manner(unlike JS)
2. In extension, the ctx-aware relationships and calculations should be done via some typesafe and faster language rather than JS. 

Setup:
1. Wasm go-files can be located anywhere, if it solves for Meshery, it can go under Meshery or if for extension, it can go under extension.
2. But, the final binary produced should come under "/public/static" folder so that next.js could pick it up and it is not bundled in extension.

Screenshot 2023-04-29 at 2 16 29 AM
playground route: /wasm-playground

This PR fixes #


Chadha93:
Woah, using WASM could be a game-changer in terms of long-term performance. This is nice. :rocket:

acald-creator:
@Abhishek-kumar09 How is this being built?

Is there a command being used to build, either with Tinygo or Go itself with the WASM turned on?

Abhishek-kumar09:
Thanks for pointing out, I've to add the command in the makefile and the npm scripts.
For handy, it can be: `GOOS=js GOARCH=wasm go build -o main.wasm`

acald-creator:
Okay, we're just doing a different directory that house the Go code for WASM purposes right?

Abhishek-kumar09:
Yes.

Yorisoft:
Closing PR - effort no longer being worked on.

PR #7656, opened 2023-05-04
 Migrate meshery-ui tracking to google tag manager

ayushthe1:
**Notes for Reviewers**

This PR attempts to migrate from `gtag.js` to `google tag manager` for tracking meshery-ui data. Also, this PR shouldn't be merged till it's in a draft state.

### DIFF
--- a/ui/pages/_document.js
+++ b/ui/pages/_document.js
@@ -17,7 +17,18 @@ class MesheryDocument extends Document {
 
           {/* Google Tag Manager */}
           <script dangerouslySetInnerHTML={{
-            __html: `(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start': new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0], j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src= 'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','GTM-WR7595H');`
+            __html: `(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start': new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0], j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src= 'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+            // Fetch user preferences
+            fetch("/api/userPreferences", { credentials: 'include' })
+              .then((res) => res.json())
+              .then((res) => {
+                if (res && res?.anonymousUsageStats === false) {
+                  // User opted out of tracking, disable GTM
+                  w[l] = []; // Clear the dataLayer array
+                  w['ga-disable-'+i] = true; // Disable Google Analytics tracking
+                }
+              });
+          })(window,document,'script','dataLayer','GTM-TFLZDSQ');`
           }} />
           {/* End Google Tag Manager */}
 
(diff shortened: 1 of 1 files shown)


ui/pages/_document.js
```diff
@@ -3,34 +3,6 @@ import PropTypes from "prop-types";
...
-  }
-
-  fetch("/api/user/prefs", { credentials: 'include' })
-    .then((res) => res.json())
-    .then((res) => {
-      if (res?.anonymousUsageStats) {
-        gtag("js", new Date());
-        gtag("config", "G-8Q51RLT8TZ", {
-          page_path: window.location.pathname,
-        });
```
ayushthe1, line 28: Do we need this logic of firing the tag only when `res?.anonymousUsageStats` is true ?

leecalcote: @Udit-takkar might have some insight.

Udit-takkar: we can directly call (without checking res?.anonymousUsageStats)  gtag  like here https://developers.google.com/tag-platform/gtagjs/install to track all users data anonymously. 

maybe this was added to allow users to opt out of anonymous tracking by checking user preferences and only then track the user behaviour

ayushthe1: Thanks @Udit-takkar for the info :+1: Is there any other change you would suugest ?

leecalcote: We track by default. Users can opt out under their User Preferences.

Is this PR complete or does it need changes @Udit-takkar?

Udit-takkar: @leecalcote @ayushthe1  ah sorry didn't saw this comment before.

Like you said that "you track by default but let user opt out in user preference". i can see this functionality in the removed code but i don't see this in the new added code that we are fetching user preferences and checking anonymousUsageStats and only then  tracking events to Google Analytics to record the page load and configuration information.

ui/pages/_document.js
```diff
@@ -43,34 +15,35 @@ class MesheryDocument extends Document {
           <meta name="referrer" content="no-referrer" />
           <link rel="icon" href="/static/favicon.png" />
 
+          {/* Google Tag Manager */}
+          <script dangerouslySetInnerHTML={{
+            __html: `(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start': new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0], j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src= 'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','GTM-WR7595H');`
+          }} />
+          {/* End Google Tag Manager */}
```
Udit-takkar, line 22: we can do this here if we want users to opt out based on user preferences

```
<script dangerouslySetInnerHTML={{
  __html: `(function(w,d,s,l,i){
    w[l]=w[l]||[];
    w[l].push({'gtm.start': new Date().getTime(),event:'gtm.js'});
    var f=d.getElementsByTagName(s)[0],
        j=d.createElement(s),
        dl=l!='dataLayer'?'&l='+l:'';
    j.async=true;
    j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;
    f.parentNode.insertBefore(j,f);

    // Fetch user preferences
    fetch("/api/userPreferences", { credentials: 'include' })
      .then((res) => res.json())
      .then((res) => {
        if (res && res.trackData === false) {
          // User opted out of tracking, disable GTM
          w[l] = []; // Clear the dataLayer array
          w['ga-disable-'+i] = true; // Disable Google Analytics tracking
        }
      });
  })(window,document,'script','dataLayer','GTM-WR7595H');`
}} />
```

Udit-takkar (approved):

Chadha93 (approved):

Issue #7616, opened 2023-04-26
[UI] Meshery UI stoped working on clicking on any filter card in Filters page


Yashsharma1911:
#### Current Behavior
On clicking on any filter card on the filters page whole Meshery UI stopped working

https://user-images.githubusercontent.com/71271069/234620967-ea3c59f9-d4ef-4eee-8391-169c1089f9a5.mov

#### Screenshots/Logs
![Screenshot 2023-04-26 at 8 49 28 PM](https://user-images.githubusercontent.com/71271069/234621408-509ef918-f77c-4d99-871d-c5aea5e79b79.png)

#### Environment

- Browser: Chrome Safari Firefox
- Host OS: Mac Linux Windows
- Meshery Server Version: stable-v
- Meshery Client Version: stable-v
- Platform: Docker Kubernetes

---

#### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🎨 Wireframes and designs for Meshery UI in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](http://slack.layer5.io)

leecalcote:
@Abhishek-kumar09

UsmanDev09:
@Yashsharma1911 This is working fine on my machine.

https://user-images.githubusercontent.com/31584129/235614476-974d347b-c353-4765-bf90-82de37c83016.mp4

Abhishek-kumar09:
Hmm, actually in playground as well, I could see it happening. @UsmanDev09 Would you try tha same thing on [playground](https://playground.meshery.io)

There is definately an issue related to memory leak or call-stack exceeding.

anshgoyalevil:
@leecalcote @Abhishek-kumar09 

Can I take up this issue if no one's working on it? 🛩️

leecalcote:
This issue needs to be addressed as promptly as possible, @meshery/maintainers

leecalcote:
@anshgoyalevil if you're able to knock it out quickly before a maintainer comes around, that'd be great. I'll leave it unassigned in the meantime, though, so that a maintainer can quickly grab it.

anshgoyalevil:
@leecalcote 

[...]

PR #7668, opened 2023-05-10
[Bug] Fix Filters page

theBeginner86:
**Notes for Reviewers**

Issue:
It was a performance issue of CodeMirror to render large file when content is not of YAML type but mode is set to `text/x-yaml`.

Solution:
Use mode of `text/plain` for rendering Filters using CodeMirror.

Long term we should be using more intuitive way of rendering Filters considering its encrypted and not in human readable format.

This PR fixes https://github.com/meshery/meshery/issues/7616

https://github.com/meshery/meshery/assets/73700530/03290892-66bf-4050-a4f4-27e6e33cc8a7


MUzairS15 (approved):
Thanks @theBeginner86 👌🏻

Right, these filters are binaries, and byte form is displayed when you flip the card, which doesn’t make sense. If want to show code, then we should display the actual wasm code

// @leecalcote

PR #7692, opened 2023-05-16
add endpoint to download the application/designs with particular :id

harkiratsm:
**Notes for Reviewers**

This PR fixes #
Applications can either be downloaded as Kubernetes Manifests or as design files. Here is the fix to support downloading application_file based on $id.

### DIFF
--- a/ui/components/MesheryPatterns.js
+++ b/ui/components/MesheryPatterns.js
@@ -33,7 +33,7 @@ import DoneAllIcon from '@material-ui/icons/DoneAll';
 import DoneIcon from '@material-ui/icons/Done';
 import PublicIcon from '@material-ui/icons/Public';
 import ConfirmationModal from "./ConfirmationModal";
-import PublishIcon from "@material-ui/icons/Publisnameh";
+import PublishIcon from "@material-ui/icons/Publish";
 import PromptComponent from "./PromptComponent";
 import ConfigurationSubscription from "./graphql/subscriptions/ConfigurationSubscription";
 import fetchCatalogPattern from "./graphql/queries/CatalogPatternQuery";


Abhishek-kumar09 (commented):
btw, source-content is something which is not available to UI, that's why there was an endpoint which lets you download the actual source file.

But, the application file could be get via `/api/application/id`, which gives the all info about application including application_file.

Now, the UI could use that `application_file` and download it as a text (from the UI only). 

Is there any other reason, this endpoint is created on server?

leecalcote:
@harkiratsm thanks for working on this. In addition to @Abhishek-kumar09's questions, will you reflect on how these changes fit into the existing APIs documented in the Meshery API Endpoints spreadsheet?

MUzairS15 (changes requested):
+1 To what @Abhishek-kumar09 said. 

All info regarding to application is already present with the UI (including the application file converted in design format).

Only thing that is not available is the original / source content i.e if you uploaded docker compose then that is not available readily, you need to make a request to get it (this implementation is also presnt in UI, when you click download btn this is what gets downloaded)

Only thing not present is ability to download an application converted to design format.

leecalcote:
> will you reflect on how these changes fit into the existing APIs documented in the Meshery API Endpoints spreadsheet? 

This will aid in my review, @harkiratsm

harkiratsm:
As soon as the changes are accepted, I will incorporate api endpoint into the spreadsheet.

harkiratsm:
https://docs.google.com/spreadsheets/d/1ABJCbfQRi0uN_YoP2kmHZ-lTI4S4QuvRk1_unomKNRE/edit?disco=AAAAvcY-AMc

harkiratsm:
It can also be used by mesheryctl to download the application file. It can also be used to download design files via catalog. I guess we're moving forward. 

Thank you @Abhishek-kumar09 for pointing out how this PR fits in other places.

leecalcote:
@harkiratsm it's unclear what the status of this is.

server/handlers/meshery_application_handler.go
```diff
@@ -646,6 +646,55 @@ func (h *Handler) GetMesheryApplicationTypesHandler(
...
+		obj := "download"
+		h.log.Error(ErrApplicationFailure(err, obj))
+		http.Error(rw, ErrApplicationFailure(err, obj).Error(), http.StatusNotFound)
+		return
+	}
+
+	var mimeType string
+	switch filepath.Ext(applicationID) {
+	case ".json":
+		mimeType = "text/plain"
```
Abhishek-kumar09, line 677: its not required. The file is always yaml in our case.

ui/components/MesheryApplications.js
```diff
@@ -395,19 +395,24 @@ function MesheryApplications({
...
-          method : "GET",
-        },
-        () => {
-          fileDownloader(id, name, source_type);
-          console.log("ApplicationFile API", `/api/application/download/${id}/${source_type}`);
-          updateProgress({ showProgress : false });
-        },
-        handleError(ACTION_TYPES.DOWNLOAD_APP)
-    );
+    try {
```
Abhishek-kumar09, line 398: There should be a change in designs as well in UI.

Abhishek-kumar09 (changes requested):

server/handlers/meshery_application_handler.go
harkiratsm: ok, updated

ui/components/MesheryApplications.js
harkiratsm: added for designs as well.

Abhishek-kumar09 (approved):

PR #7714, opened 2023-05-20
[Meshsync Subsription Leak] Fix GrapqhQL subscription memory leak

Abhishek-kumar09:
Problems in the current code:

1. One client receives data 3 times the data that it requires, the subscription client even on an ask to dispose doesn't stop sending the data to the client
Screenshot 2023-05-20 at 1 42 47 AM

<br />
In the screenshot below, the `instance number` tells the number of channel the same data is sent to. For any particular client, the same data is sent 3 times on three different channels(even though the subscription is disposed by client)
Screenshot 2023-05-20 at 1 49 41 AM

2. The client refresh doesn't dispose the channel data
Screenshot 2023-05-20 at 2 34 57 AM

In the screenshot above, on 3 refreshes, the 9 channels are created and data goes back and forth (I will have to check if client receives it or not, but definitely processed by the server, and a try to send it)

3. Meshsync being raw, send the data as-is to Meshery-Server, which is infact fine, but there is no processing unit on Meshery-server, that could clean the data and send only required item.
For instance, any changes in Node (specially when it is autoscaled) or kube-proxy server changes, which happens quite frequently sometimes more than 10-20 times in a second, may totally fill up the browser memory letting it to die eventually.

https://github.com/meshery/meshery/assets/48255244/555d3d64-8c6a-43b9-925b-ef4fa44a4911

(above recording is my older commit implementation, and just describes the problem of memory leak and how a processing unit is drastically reducing the data sent to client, The current code is much more better given it mutates data with mutex, and dispose the channel updates)

### Solution:
1 and 2.: Whenever a gql subscription is disposed or client is disconnected, the `ctx` is completed and gets `Done()` which should be the terminal point, and that's when no other data should be sent via graphQL.

3: There has to be batch processing middleware that should also act as rate-limiter for the number of responses sent by the server to client to not overflow the memory of browser.

Using multi-column index to de-duplicate and send the latest event for a particular resource given it is created for one purpose.

Have implemented the key-value map, whose key is a result of calculation of:
   1. meshsync-event-type(ADD, MODIFIED, DELETE, etc)
   2. resource-kind(POD, namespace, etc)
   3. context-id of that resource
   4. kubernetes object-metadata uid (id assigned by kubernetes for a particular resource)

These 4 columns create a key in hashmap, that under the activity of 5 seconds, gets any new value assigned by meshsync to stay up to-date. It drastically, de-deuplicates the same data sent to client and reduces the memory footprint to the factor of 10-20x. 

**Notes for Reviewers**
Getting some problems closing the channel completely, but making sure that the data received on disposed channel is not sent to the client.

This PR fixes #

### DIFF
--- a/server/internal/graphql/resolver/schema.resolvers.go
+++ b/server/internal/graphql/resolver/schema.resolvers.go
@@ -349,7 +349,6 @@ func processAndRateLimitTheResponseOnGqlChannel(publishChannel chan *model.MeshS
 
 	processMap := syncedProcessMap{processMap: make(map[string]*model.MeshSyncEvent)}
 
-	// processMap := make(map[string]*model.MeshSyncEvent)
 	isLast := false
 
 	return func(meshsyncEvent *model.MeshSyncEvent) {
@@ -371,10 +370,8 @@ func processAndRateLimitTheResponseOnGqlChannel(publishChannel chan *model.MeshS
 		processMap.mu.Unlock()
 
 		if !shouldWait {
-			// fmt.Println("starting to publish......")
 			shouldWait = true
 			isLast = false
-			// push the entry
 
 			go func() {
 				<-time.After(d)


server/internal/graphql/resolver/schema.resolvers.go
```diff
@@ -304,25 +306,110 @@ func (r *subscriptionResolver) SubscribeMeshSyncEvents(ctx context.Context, k8sc
...
 				// handle the events
 				res := &model.MeshSyncEvent{
 					ContextID: ctxID,
 					Type:      string(event.EventType),
 					Object:    event.Object,
 				}
-				resChan <- res
-				go r.Config.DashboardK8sResourcesChan.PublishDashboardK8sResources()
+				publishHandlerWithProcessing(res)
+				// go r.Config.DashboardK8sResourcesChan.PublishDashboardK8sResources()
```
Abhishek-kumar09, line 323: @theBeginner86 I didn't go deep in what this function is used for (since I have you 😉), but even without this publish endpoint, the dashboard data is being correctly synced and processed (as far as I could see). Would you help in checking and re-assuring, please educate me on this portion, I should have missed something.

theBeginner86: Thanks for pointing this out.

This `DashboardK8sResourcesChan.PublishDashboardK8sResources` is a function that publishes to the golang channels that are being used to track dashboard subscription responsible for dynamically updating `Workloads` widget in the dashboard.
If we remove this then those channels wouldn't receive updates when a new object is received from MeshSync and hence wouldn't be able to send updates in UI.

It has a sibling function [DashboardK8sResourcesChan.SubscribeDashboardK8sResources](https://github.com/meshery/meshery/blob/master/server/models/dashboard_k8s_resources.go#L20). This [subscribeClusterResources](https://github.com/meshery/meshery/blob/master/server/internal/graphql/resolver/kubernetes.go#L119) is invoked when UI sends a subscribe request to backend.

Even without this function, we would see data in Workload widget in dashboard because of [GetClusterResources](https://github.com/meshery/meshery/blob/master/server/internal/graphql/resolver/schema.resolvers.go#L151) query but that would just be a one time fetch and UI would update only on reload.

Abhishek-kumar09: Thanks @theBeginner86

MUzairS15 (commented):
> 1. meshsync-event-type(ADD, MODIFIED, DELETE, etc)
> 2. resource-kind(POD, namespace, etc)
> 3. context-id of that resource
> 4. kubernetes object-metadata uid (id assigned by kubernetes for a particular resource)

object-metadata uid doesn’t get updated (have you verified that this updates ?) so the update event will always be missed isn’t it?

Instead of using timeout, it is better in my opinion to use weighted semaphores on publishChannel.
By ensuring weight of semaphore acquired before publishing is greater than weight of being released we can achieve similar behaviour, and this is what semaphore is for.

Abhishek-kumar09:
> > 1. meshsync-event-type(ADD, MODIFIED, DELETE, etc)
> > 2. resource-kind(POD, namespace, etc)
> > 3. context-id of that resource
> > 4. kubernetes object-metadata uid (id assigned by kubernetes for a particular resource)
> 
> object-metadata uid doesn’t get updated (have you verified that this updates ?) so the update event will always be missed isn’t it?
> 
> For resources that are updated frequently MeshSync behaviour will change [meshery/meshery-operator#479](https://github.com/meshery/meshery-operator/issues/479) Timeout per component, so the timeout here will affect real time updates. Just want to ensure we don’t have same logic at two places.

> object-metadata uid doesn’t get updated

As far as I checked, it doesn't.

> so the update event will always be missed isn’t it?

No it doesn't, the map clears itself once the data is sent to client, so the next update event coming would always be registered and sent, even when it same as what just went in previous transfer. See the ratelimiter function to know where it optimizes.

> For resources that are updated frequently MeshSync behaviour will change https://github.com/meshery/meshery-operator/issues/479
Timeout per component, so the timeout here will affect real time updates. Just want to ensure we don’t have same logic at two places.

It is underlapping some criteria we're talking about here. It still talks about the giving user the power to set configurability and timeouts, even if it not, we might be adding some entries of resource for the timeout. While it is easy in this case to miss some resources, the function we are using in here is activated on all the resources and give updates to the "most recent activity" happening in that resource (alternatively, the listen timeout is not good imo, because while the timeout is not resolved, the new updates could be missed on that resource).

If we; at some point, implement that function correctly in meshsync, we can safely remove this portion, till then, it is required to unblock the client from filling memory.

leecalcote:
@sudo-NithishKarthik this is an intriguing update that you might want to review.

server/internal/graphql/resolver/schema.resolvers.go
```diff
@@ -304,25 +306,110 @@ func (r *subscriptionResolver) SubscribeMeshSyncEvents(ctx context.Context, k8sc
...
+	processMap := syncedProcessMap{processMap: make(map[string]*model.MeshSyncEvent)}
+
+	// processMap := make(map[string]*model.MeshSyncEvent)
+	isLast := false
+
+	return func(meshsyncEvent *model.MeshSyncEvent) {
+		// create a key to uniquely identify meshsync objects with its type, purpose, ctx and resource uniqueId
+		var key string
+		key += meshsyncEvent.Type
+		key = (meshsyncEvent.Object).(map[string]interface{})["kind"].(string)
```
Abhishek-kumar09, line 360:
```suggestion
		key += (meshsyncEvent.Object).(map[string]interface{})["kind"].(string)
```

leecalcote (approved):
LGTM

leecalcote:
@MUzairS15 good to merge?

PR #7757, opened 2023-05-31
[playground] Update docker based deployment

theBeginner86:
**Notes for Reviewers**

This PR documents recent changes done to the docker based deployment of [Cloudnative Playground](https://playground.meshery.io).

Changes:
1. Added missing watch label to all services so that watchtower can update containers of each service defined in docker-compose.
2. Reduced the [Polling interval](https://containrrr.dev/watchtower/arguments/#poll_interval) to 6hrs for watchtower service so that the frequency by which it pulls latest images increases. Reason being that by default watchtower pulls the latest image in 24 hrs which means if a release was made then playground deployment would be updated in 24hrs and not immediately.
3. Documented the complete nginx config. Earlier we were just documenting part of the config in `/etc/nginx/conf.d/default.conf` which was being imported into `/etc/nginx/nginx.conf` (default location). So also updated the container mount for nginx service.
4. Exposed port 80 for nginx service so that HTTP to HTTPS redirection works.

This PR fixes #


install/playground/docker/nginx.conf
```diff
@@ -25,7 +64,7 @@ server {
 	   proxy_pass http://meshery-meshery-1:9081;
```
Revolyssup, line 64: Is the uri correct here? It wont be http://meshery:9081 ?

MUzairS15: Yes, as the meshery container gets named meshery-meshery-1, (i.e. network-container name-serial no.)

MUzairS15 (commented):
Reducing the polling time to 6hrs from 24 hrs is helpful. But I will say, it will be nice if we update the release workflow to pull the latest image and deploy it again.

theBeginner86:
> Reducing the polling time to 6hrs from 24 hrs is helpful. But I will say, it will be nice if we update the release workflow to pull the latest image and deploy it again.

Sounds good. 
This caveat of watchtower was the reason we moved away from it in remote provider . I'll create an issue to follow up on this.

PR #7784, opened 2023-06-02
fix: check if the pattern upload, is valid

harkiratsm:
**Notes for Reviewers**

This PR fixes #

The designers sometimes get stuck when they try to upload docker-compose or a kubernetes manifest under design, since it is an invalid design construct.

Now we'll check for valid design files and only save them if they're valid, when users uploads.

### DIFF
--- a/server/models/pattern/core/pattern.go
+++ b/server/models/pattern/core/pattern.go
@@ -217,7 +217,7 @@ func IsValidPattern(stringifiedFile string) (err error) {
 	}
 
 	if pattern.Services == nil {
-		return errors.New("missing services field")
+		return errors.New("invalid design-file format: missing services field")
 	}
 
 	for serviceName, service := range pattern.Services {


server/models/pattern/core/pattern.go
```diff
@@ -207,6 +208,26 @@ func NewPatternFile(yml []byte) (af Pattern, err error) {
...
+// isValidPattern checks if the pattern file is valid or not
+func IsValidPattern(stringifiedFile string) (err error) {
+	pattern := Pattern{}
+
+	if err = yaml.Unmarshal([]byte(stringifiedFile), &pattern); err != nil {
+		return err
+	}
+
+	if pattern.Services == nil {
+		return errors.New("missing services field")
```
Abhishek-kumar09, line 220:
```suggestion
		return errors.New("invalid design-file format: missing services field")
```

server/models/pattern/core/pattern.go
```diff
@@ -207,6 +208,26 @@ func NewPatternFile(yml []byte) (af Pattern, err error) {
...
+	}
+
+	for serviceName, service := range pattern.Services {
+		if service.Traits == nil {
+			return errors.New("missing traits field for:" + serviceName)
+		}
+	}
```
Abhishek-kumar09, line 227: Not required for a pattern/design to always have traits, although we have traits for every pattern/design now, this is not an important check.

Abhishek-kumar09 (changes requested):

leecalcote:
@harkiratsm did you address this feedback? If so, re-request a review.

leecalcote:
and comment on how yo addressed the feedback.

server/models/pattern/core/pattern.go
```diff
@@ -207,6 +208,26 @@ func NewPatternFile(yml []byte) (af Pattern, err error) {
...
+
+	if pattern.Services == nil {
+		return errors.New("invalid design-file format: missing services field")
+	}
+
+	for serviceName, service := range pattern.Services {
+		if service.Traits == nil {
+			return errors.New("missing traits field for:" + serviceName)
+		}
+	}
```
leecalcote, line 227:
```suggestion
//	for serviceName, service := range pattern.Services {
//		if service.Traits == nil {
//			return errors.New("missing traits field for:" + serviceName)
//		}
//	}
```

PR #7789, opened 2023-06-05
[chore] Error codes cleanup

nebula-aac:
**Notes for Reviewers**

This PR removed some of the unused error codes.

I would create new ones, but it seems like they weren't being used anywhere in the repo.

There is one warning on one error code: `ErrInvalidRequestObjectCode` which does exist, but the `errorutil` doesn't check if the error message format is appending new strings, or if the fields are nil.

I will be adding a check in `meshkit` for it.

If there are any error codes you would like to use here, please let me know, and we can make revisions.


server/handlers/error.go
```diff
@@ -113,6 +110,7 @@ const (
 	ErrGetMeshModelsCode                = "2257"
 	ErrGetUserDetailsCode               = "2258"
 	ErrResolvingRelationship            = "2259"
+	ErrGetLatestVersionCode             = ""
```
Revolyssup, line 113: Why is this empty?

Revolyssup (changes requested):

server/handlers/error.go
nebula-aac: @Revolyssup This has been revised to use a random number, until I can clean up the duplicates.

leecalcote:
Refreshing with changes from HEAD. 

@Revolyssup your eyes on this are appreciated.

leecalcote:
@nebula-aac did we end up with new issues to track enhancements on the errorutils? I can't remember...

nebula-aac:
I will create a new issue tracking `errorutil`.

//@leecalcote

leecalcote:
Thank you much. 🙂

leecalcote (approved):
Lgtm

PR #7799, opened 2023-06-06
Fix failing cluster cleanup workflow

Revolyssup:
**Notes for Reviewers**

1. This PR assumes that a playground-cleanup script is present in the VM. A follow up issue should be created to add another workflow which can automatically update the script inside VM when scripts/playground-cleanup.sh changes in master.
Currently that is done manually.

2. A new secret is added at repo level. `METAL_SSH_KEY`

### DIFF
--- a/.github/workflows/cncf-playground-reset.yaml
+++ b/.github/workflows/cncf-playground-reset.yaml
@@ -6,22 +6,14 @@ on:
 
 jobs:
   reset:
-    name: Cluster Reset
-    if: github.repository == 'meshery/meshery'
-    runs-on: ubuntu-latest 
-    steps:
-    - uses: azure/setup-kubectl@v3
-    - name: Extract kubeconfig
-      run: mkdir ~/.kube; touch ~/.kube/config ;echo "${{ secrets.PLAYGROUND_CONFIG }}" | base64 --decode > ~/.kube/config
-    - name: Delete namespaces except the ones which are whitelisted in the if statement
-      run: |
-        for ns in $(kubectl get ns -o jsonpath="{.items[*].metadata.name}");
-        do
-        if [[ "$ns" != "kube-system" && "$ns" != "default" && "$ns" != "monitoring" && "$ns" != "kube-flannel" && "$ns" != "kube-node-lease" && "$ns" != "kube-public" && "$ns" != "meshery" && "$ns" != "metallb-system" && "$ns" != "projectcontour" ]];then
-          echo "Deleting namespace $ns"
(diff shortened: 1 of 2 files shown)


.github/workflows/cncf-playground-reset.yaml
```diff
@@ -6,22 +6,14 @@ on:
...
-        do
-        if [[ "$ns" != "kube-system" && "$ns" != "default" && "$ns" != "monitoring" && "$ns" != "kube-flannel" && "$ns" != "kube-node-lease" && "$ns" != "kube-public" && "$ns" != "meshery" && "$ns" != "metallb-system" && "$ns" != "projectcontour" ]];then
-          echo "Deleting namespace $ns"
-          kubectl delete ns "$ns"
-        fi
-        done
+    name: Cluster cleanup by running the cleanup script inside VM
+    uses: appleboy/ssh-action@master
+    with:
+        host: 147.75.47.9
```
theBeginner86, line 12: One quick question, shouldn't we be SSHing into the second VM root@c3-medium-x86-02-meshery on IP address `147.28.149.165`?

> As Meshery + Docker is to be in root@c3-medium-x86-01-meshery on IP address `147.75.47.9` and Kubernetes is to be in root@c3-medium-x86-02-meshery on IP address `147.28.149.165`.

Chadha93:
Hi @Revolyssup, if you are available for the dev call, let's discuss this PR. Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit#) if you would. :)

.github/workflows/cncf-playground-reset.yaml
Revolyssup: Although both will work but yes your suggestion makes better sense. I will change it.

Revolyssup: I did it on this node thinking it will be a test of another thing which is - "Is kubeconfig available to Meshery on node 1". Though wrangling cleanup with this pseudo test isn't good. Thanks for pointing out.

Revolyssup: Second reason to do it this way to have all control points out of the node where test cluster is running, so that we can take down that node and do cleanup without losing any scripts or configuration.

theBeginner86 (approved):

theBeginner86:
I'll create the issue to track another workflow to update script.

PR #7801, opened 2023-06-06
[Mesheryctl] Support tailing logs in Kubernetes platform

### LINKED ISSUE #3234: [mesheryctl] Support tailing logs in Kubernetes platform

#### Current Behavior
mesheryctl users are able to fetch logs from Kubernetes-based deployments using:

```
$ mesheryctl system logs
```
Users are not able to tail logs, however.

#### Desired Behavior

Support following / tailing logs of any Meshery component deployed on Kubernetes.

```
$ mesheryctl system logs meshery -f
```

#### Implementation
Ideally, a `-f`/`--follow` flag would provide support following / tailing logs of any Meshery component deployed on Kubernetes.

#### Acceptance Tests
1. Users can use -f or --follow as a command to stream Meshery component logs to their terminal.
1. Users can use CTRL+C or CMD+C to stop following logs.

---

#### Contributor Resources

Meyazhagan:
**Notes for Reviewers**

This PR fixes #3234 

### DIFF
--- a/mesheryctl/internal/cli/root/system/logs.go
+++ b/mesheryctl/internal/cli/root/system/logs.go
@@ -16,7 +16,6 @@ package system
 
 import (
 	"bufio"
-	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -48,16 +47,19 @@ func IsPodRequired(requiredPods []string, pod string) bool {
 	return false
 }
 
+func printLogs(logs string, podName string) {
+	for _, logMsg := range strings.Split(logs, "\n") {
+		logStr := fmt.Sprintf("%s\t|\t%s", podName, logMsg)
+
+		log.Print(logStr)
+	}
+}
+
 var (
 	follow bool
 )
 
 const BYTE_SIZE = 2000
-
-type logChanStruct struct {
-	PodName string
-	Message []byte
-}
 
 // logsCmd represents the logs command
 var logsCmd = &cobra.Command{
@@ -193,25 +195,7 @@ mesheryctl system logs meshery-istio
 			}
 
 			log.Info("Starting Meshery logging...")
-
-			logChan := make(chan logChanStruct)
 			wg := &sync.WaitGroup{}
-			wg.Add(1)
-			go func() {
-				defer wg.Done()
-				for {
-					value, ok := <-logChan
-					if ok {
-						for _, msg := range strings.Split(string(value.Message), "\n") {
-							logStr := fmt.Sprintf("%s\t|\t%s", value.PodName, msg)
-
-							log.Print(logStr)
-						}
-					} else {
-						break
-					}
-				}
-			}()
 
 			// List all the pods similar to kubectl get pods -n MesheryNamespace
 			for _, pod := range podList.Items {
@@ -243,9 +227,17 @@ mesheryctl system logs meshery-istio
 						return err
 					}
 					defer logs.Close()
-
-					if follow {
+					var logBuf []byte
+					if !follow {
+						logBuf, err = io.ReadAll(logs)
+						if err != nil {
+							return fmt.Errorf("error occurred while processing logs")
+						}
+						printLogs(string(logBuf), name)
+					} else {
+						wg.Add(1)
 						go func() {
+							defer wg.Done()
 							for {
 								buf := make([]byte, BYTE_SIZE)
 								numBytes, err := logs.Read(buf)
@@ -256,29 +248,18 @@ mesheryctl system logs meshery-istio
 									break
 								}
 								if err != nil {
+									log.Println("error occurred while processing logs", err)
 									break
 								}
-								logChan <- logChanStruct{PodName: name, Message: buf[0:numBytes]}
+								logBuf = buf[0:numBytes]
+								printLogs(string(logBuf), name)
 							}
 						}()
-					} else {
-						buf := new(bytes.Buffer)
-						_, err = io.Copy(buf, logs)
-						if err != nil {
-							return fmt.Errorf("error in copy information from logs to buf")
-						}
-
-						logChan <- logChanStruct{PodName: name, Message: buf.Bytes()}
 					}
 				}
 			}
-			if !follow {
-				close(logChan)
-			}
-
 			wg.Wait()
 		}
-
 		return nil
 	},
 }


mesheryctl/internal/cli/root/system/logs.go
```diff
@@ -212,27 +244,45 @@ mesheryctl system logs meshery-istio
...
+								buf := make([]byte, BYTE_SIZE)
+								numBytes, err := logs.Read(buf)
+								if numBytes == 0 {
+									continue
+								}
+								if err == io.EOF {
+									break
+								}
+								if err != nil {
+									break
```
Revolyssup, line 259: Here in the case of unknown error, you can throw the error to the client.

mesheryctl/internal/cli/root/system/logs.go
```diff
@@ -212,27 +244,45 @@ mesheryctl system logs meshery-istio
...
+								}
+								logChan <- logChanStruct{PodName: name, Message: buf[0:numBytes]}
+							}
+						}()
+					} else {
+						buf := new(bytes.Buffer)
+						_, err = io.Copy(buf, logs)
+						if err != nil {
+							return fmt.Errorf("error in copy information from logs to buf")
```
Revolyssup, line 268: This error message is leaky. You are talking about internal variables in user facing log messages. Replace them with appropriate messages that signal to the user that logs couldn't be fetched with possible reasons.

Revolyssup: You can also avoid creating another buffer variable by directly calling io.ReadAll on the logs in this case.

mesheryctl/internal/cli/root/system/logs.go
```diff
@@ -212,27 +244,45 @@ mesheryctl system logs meshery-istio
...
+								if err == io.EOF {
+									break
+								}
+								if err != nil {
+									break
+								}
+								logChan <- logChanStruct{PodName: name, Message: buf[0:numBytes]}
+							}
+						}()
+					} else {
```
Revolyssup, line 264: Instead of putting it as else condition. Pull it on top and return in the case of non-follow. Other wise continue as usual.

Revolyssup (changes requested):

mesheryctl/internal/cli/root/system/logs.go
Meyazhagan: @Revolyssup how to throw the error to the client from go routine? 

I saw we can use **errgroup** to group go routine and handle errors.

Revolyssup: Yeah you can use that as well. Thats usually used when a lot of synchronisation is needed.
Although here it feels like goroutine's are not really required. There isn't much synchronisation required and this is added complexitiy.

So currently you start a consumer blocking go routine and [ producer blocking go routine only when follow is present]. Then you block the parent function until producer returns and then you close the channel. Or in case of of non follow you immediately close the channel. 

A much simpler solution is:
Dont use any go routines here. Check if `!follow` then print out the outputs and return immediately. 
Else block the main function in a for loop and keep printing until EOF.  No extra synchronisation overhead needed here.

Revolyssup (changes requested):

mesheryctl/internal/cli/root/system/logs.go
```diff
@@ -182,6 +193,26 @@ mesheryctl system logs meshery-istio
...
+			logChan := make(chan logChanStruct)
+			wg := &sync.WaitGroup{}
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				for {
+					value, ok := <-logChan
+					if ok {
+						for _, msg := range strings.Split(string(value.Message), "\n") {
+							logStr := fmt.Sprintf("%s\t|\t%s", value.PodName, msg)
```
alphaX86, line 206: Are we sure that we're making this command only k8s exclusive? What about other platforms?

mesheryctl/internal/cli/root/system/logs.go
alphaX86, line 268:
```suggestion
							return fmt.Errorf("error in copy information from logs to buf")
```
This can be abstracted as @Revolyssup said

alphaX86 (changes requested):

Meyazhagan:
@Revolyssup @alphaX86 I made the updates. Take a look

mesheryctl/internal/cli/root/system/logs.go
Meyazhagan: @alphaX86 Yes. For docker that is already taken care

Revolyssup:
> @Revolyssup @alphaX86 I made the updates. Take a look

Its still using go routine.

Meyazhagan:
>>> A much simpler solution is:
Dont use any go routines here. Check if !follow then print out the outputs and return immediately.
Else block the parent function in a for loop and keep printing until EOF. No extra synchronisation overhead needed here.

>> Its still using go routine.

> @Meyazhagan, feedback for you to consider.

@Revolyssup  Can you suggest with sample code without use of go routines

Revolyssup:
> @Revolyssup @alphaX86 I made the updates. Take a look

You're using one go routine and in order to block that go routine you're using waitgroup. That's just sequential behaviour. I think its normal for developers to instinctively using concurrency primitives when they are so convenient, even if the execution is sequential.

if !follow, then log the output and return early.
Otherwise, go in a for loop and block the function until EOF.

Meyazhagan:
@Revolyssup I think you misunderstood the way go routine is used here

if this is the case

```go
go func() {
   for {
      // some code
   }
}
```
then > go in a for loop and block the function until EOF.

Meyazhagan:
```go
for pod := range pods {
    for container := range pod.containers {
        if !follow { // print log immediately
        } else {
               for {} // will be running until EOF
        }
    }
}
```
here the `for` loop is blocking `pod.containers` loop, because it waiting for loop to break (till EOF or error) at 1st container's logs.

so I consciously moved this `for` loop inside the goroutine (creating a subtask) for all containers. now the `for` loop of each container, will be running concurrently. I feel this is a simpler solution!

---

> You're using one go routine

No. By seeing code it seems to be one go routine! But it was in loop creating as many as pods and containers there.


> Otherwise, go in a for loop and block the function until EOF.

Why we need to use just `for` loop and wait until loop breaks (EOF) at 1st item in the loop. Is there a reason for this to do?


> developers to instinctively use concurrency primitives when they are so convenient

I am open to new suggestions and try new things! Why I didn't change concurrency (go routine) because if we removed it, code will won't work as expected. That's why I am asking you the reason to change or sample pseudo code of your suggestion.


> There isn't much synchronisation required and this is added complexitiy.

I don't think this will add complexity.


> No extra synchronisation overhead needed here.

Finally, If you want to do so
```go
// if follow 
for {
    for pod := range pods {
        for container := range pod.containers {
             // print log immediately
        }
    }
}
```
yeah we can do this way! But I don't know how efficient it will be.

@leecalcote @Revolyssup

Revolyssup (approved):

PR #7813, opened 2023-06-07
Update cncf-playground-reset.yaml

Revolyssup:
**Notes for Reviewers**

- Stop rebooting first VM and only reboot the VM running cluster.
- First perform cleanup before rebooting


leecalcote (approved):
Any longer... I question whether we need to reboot the Kubernetes cluster server at all. I'll propose that we cease doing this...

PR #7859, opened 2023-06-14
[UI] - Casl initialized in meshery UI

### LINKED ISSUE #7382: [UI] Permissions Framework
#### Current Behavior
Meshery UI lacks a permissions framework. The existing internal implementation is extremely simple and fragile and is “does user have a role with this name assigned?“. So, it must be completely set aside / rewritten.

#### Prologue
We’re about to go headlong into definition of permissions and their enforcement in Meshery with an aim for it to be quite dynamic in terms of each input component carrying a unique permission key id. Each key is then put into a group of keys in a keychain, keychains assigned to user roles, in turn, roles assigned to users. With users having the ability to create own custom roles, the framework will be dynamic based on the associated server-side permissions for the currently auth’ed user.

Its the library / framework that we’re interested in. We’re in need of an extensible framework in which permissions to for any given user to see or to interact with a UI component can be dynamically discerned based on that users’ assigned permissions and not hardcoded into the UI.

This code snippet offers the appropriate mindset to have when considering the needs at-hand - https://isamatov.com/react-permissions-and-roles/#maps-for-roles-and-permissions

Meshery UI's approach needs to be extensible. This framework will be an extension point for Remote Providers to supply their own permissions.

#### Permissions via Extension Point

Each “Key” here will be a combination of a UI input field / component ID + action, so for example: `<input type="text" id="settings-adapter-uri" />`

`id` is the unique identifier of the field. Separately, the key would then contain four different composites: `key-id` + `action`

Example: `settings-adapter-uri `+ `edit`

#### Research
Investigate available frameworks. We’re just looking for an existing permission framework to use as an internal library. 

---

#### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](http://slack.layer5.io)

smartdev09:
**Notes for Reviewers**

This PR fixes #7382

### DIFF
--- a/ui/components/Filters.js
+++ b/ui/components/Filters.js
@@ -1067,7 +1067,7 @@ function MesheryFilters({ updateProgress, enqueueSnackbar, closeSnackbar, user, 
     <>
       {ability.can(keys.VIEW_FILTER.subject, keys.VIEW_FILTER.action) ? (
         <NoSsr>
-            {selectedRowData && Object.keys(selectedRowData).length > 0 && (
+          {selectedRowData && Object.keys(selectedRowData).length > 0 && (
           <YAMLEditor filter={selectedRowData} onClose={resetSelectedRowData()} onSubmit={handleSubmit} classes={classes} />
         )}
         <div className={classes.topToolbar} >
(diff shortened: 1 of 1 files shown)


leecalcote:
@nebula-aac will you become familiar with our use of this framework? It will be pervasively used. Here is the [permissions framework design spec](https://docs.google.com/document/d/16b1tarq7iwIlC9sqQ6y_4utnuw6qgAJmZse3feE_BKo/edit?usp=sharing).

nebula-aac (changes requested):
I feel it would be better to rename the file to something that will help us know what does this do.

Since this is a React Context API, perhaps we can rename `Can.js` to indicate that this is a context api file?

We don't have a dedicated folder for contexts so we don't need to move this now, but in the near future, I would like to make a note to remember this so this will be moved into an appropriate folder for contexts.

PR #7930, opened 2023-06-24
feat: add a custom native select widget in rjsf

harkiratsm:
**Notes for Reviewers**

This PR fixes #

Adding a custom native select widget.

image


leecalcote:
@aabidsofi19 will you offer review here?

aabidsofi19:
@leecalcote  sure.

vishalvivekm:
@harkiratsm Let's discuss this on Meshery Dev call. Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit#) if you would. :)

Abhishek-kumar09:
Before starting this is what a [Select](https://mui.com/material-ui/react-select/) is called.

Now, select could be created by MUI, and it could be a native one, that uses the html/browser native APIs.

You just saw the MUI select, This is what [Native select](https://mui.com/material-ui/react-select/#native-select) looks like.

Problem:
The MUI select popup when opens up, an html element is created and appended at the `body` level of dom-tree. When we open a RJSF form, that also is created on the body level of dom-tree. Now if you notice the two elements when opened as a popup becomes sibling to each other, even if the `Select` element is a children of RJSF's tooltip. 

Remember the behaviour of RJSF: Whenever you interact with anything outside the RJSF tooltip, the tooltip closes itself to give you a better experience. e.g. clicking on another node closes the RJSF that was opened by the previous node. Anything outside the RJSF tooltip is considered a trigger point to close the RJSF form, thus when you select something from the dropdown list (while opening second level tooltip, because that is an extended version of tooltip, now you have to manage two layers of tooltips together), the RJSF form closes down because the select is considered to be a click outside the tooltip.

Using Native select, the html element are not created on portals (by portal I meant, creating a html node and appending it to the body), but they directly become the child of the their parent component, thus still following the usual parent-child relationship, and now clicking on the select is considered a click in its children (unlike sibling), that would keep the form stay open without any interference.

leecalcote:
Assigning to @harkiratsm for a response...

harkiratsm:
As I just verified, the second level tooltip works perfectly. 
Therefore, this PR fixes the odd behavior of select on the second level tooltip.


image

harkiratsm:
We're good to merge.

PR #7963, opened 2023-06-27
[Filters] Store filter as comp instance

MUzairS15:
**Notes for Reviewers**

This PR fixes #


vishalvivekm:
@MUzairS15 Let's discuss this on Meshery Dev call. Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit#) if you would. :)

server/models/meshery_filter.go
```diff
@@ -21,7 +21,7 @@ type MesheryFilter struct {
 	Location    sql.Map `json:"location"`
 	Visibility  string  `json:"visibility"`
 	CatalogData sql.Map `json:"catalogData"`
-	Config      string  `json:"config"` 
+	FilterResource      string  `json:"filter_resource"`
```
leecalcote, line 24: `Component`?

leecalcote (changes requested):

server/models/meshery_filter.go
MUzairS15: To be in accordance with how pattern_resources are stored, I chose `filter_resource`, as it is actually a instance of that component/ resource of that

theBeginner86: can we call it `component_resource`?

So that we can normalize this construct between patterns, filters etc.

server/models/pattern/utils/utils.go
```diff
@@ -100,3 +102,13 @@ func ToMapStringInterface(mp interface{}) map[string]interface{} {
 	}
 	return res
 }
+
+func GetRandomAlphabetsOfDigit(length int) (s string) {
+	charSet := "abcdedfghijklmnopqrstuvwxyz"
+	for i := 0; i < length; i++ {
```
Abhishek-kumar09, line 108: Isn't it there already? When the name of two services have conflict with each other the random string concatenated at the last.

MUzairS15: Yes, I have just moved this func here.

MUzairS15:
Merging this. // @Abhishek-kumar09

PR #8005, opened 2023-06-30
[MeshModel] Add Binding Relationship

0xt3j4s:
**Notes for Reviewers**

1. This PR adds a `Binding` relationship, which can be used to evaluate whether two resources can have a binding type of relationship.
2. As of now added this type of relationship for:
     - From [`Secrets`, `CronJob`] to [`ConfigMaps`, `Pods`]

**Logic:** If a component uses another component for its basic functionality or enhanced functionality or integrity, then we can count it as a binding relationship.

### DIFF
--- a/server/meshmodel/relationships/binding.json
+++ b/server/meshmodel/relationships/binding.json
@@ -26,15 +26,15 @@
           {
             "kind": "CronJob",
             "model": "kubernetes"
-          },
-          {
-            "kind": "PodTemplates",
-            "model": "kubernetes"
           }
         ],
         "to": [
           {
             "kind": "ConfigMaps",
+            "model": "kubernetes"
+          },
+          {
+            "kind": "Pods",
             "model": "kubernetes"
           }
         ]


server/meshmodel/relationships/binding.json
```diff
@@ -0,0 +1,43 @@
...
+    "apiVersion": "core.meshery.io/v1alpha1",
+    "kind": "Binding",
+    "metadata": {
+      "description": "A relationship between two resources where one resource uses the another resource for it's basic or enhanced functionality or Integrity.",
+      "name": "Binding Relationship",
+      "type": "Binding"
+    },
+    "model": {
+      "name": "kubernetes",
+      "version": "v1.25.2",
```
leecalcote, line 11: @MUzairS15 are we able to make this version optional? Or is it essentially inconsequential at this juncture?

leecalcote:
I wonder if we don't want to use `kind: Hierarchical` and `subType: Sibling` to represent this relationship.

MUzairS15 (commented):
Thankyou @0xt3j4s , for coming up with this.
I have few considerations though,
1. CronJob template which defines a Job can reference a configmap as well as a secret 
2. Pod can also refer a secret

ConfigMap, secrets, PV, PVCs are all related to storage so can we have a representation were all these and other resources pertaining to storage are represented correctly. A defined representation for storage resources and how they are referenced across eligible resources.

also consider this thing, a cronjob will spin up specified number of jobs (short lived container). If a job is running independently it can acquire it via matchLabels, so please do consider how the relation is established here via matchLabels/mathExoressions

Suppose a cron job spins 3 jibs can we represent them visually, can we use one of the four corners of a node (see the design in figma)

0xt3j4s:
@MUzairS15 

I got your point. We need to appropriately represent the storage-related resources and their connections with other resources.

Regarding `CronJob`, are you referring to the parent-child relationship between the jobs created by `CronJob` and `CronJob`?
And we need to establish relationships between `CronJob` and associated jobs through labels and selectors to identify the jobs it creates.

0xt3j4s:
Also, in this definition, there are four possible relationships; the explanation for each of them is given below:

- Secrets to ConfigMaps: Secrets are incorporated into ConfigMaps to include sensitive data.
- Secrets to Pods: Secrets are bound to Pods to provide secure access to sensitive data within the Pods.
- CronJob to ConfigMaps: CronJobs reference ConfigMaps to use the configuration data stored in the ConfigMap for its functionality.
- CronJob to Pods: CronJobs create and manage Pods to execute jobs according to the defined schedule, ensuring lifecycle management and adherence to the schedule.

Boombag0607 (commented):
> * **Secrets** to **ConfigMaps**: Secrets are incorporated into ConfigMaps to include sensitive data.
> * **Secrets** to **Pods**: Secrets are bound to Pods to provide secure access to sensitive data within the Pods.
> * **CronJob** to **ConfigMaps**: CronJobs reference ConfigMaps to use the configuration data stored in the ConfigMap for its functionality.
> * **CronJob** to **Pods**: CronJobs create and manage Pods to execute jobs according to the defined schedule, ensuring lifecycle management and adherence to the schedule.

@0xt3j4s Can we define a `subtype` based on these 4 relationships?

0xt3j4s:
@Boombag0607 Do You mean to define `Binding` as a sub-type?

Boombag0607:
We must define what a `subtype` will represent in a group-to-group relationship; it is difficult to say otherwise. Also, what will it signify in the corresponding visualization? 
For instance, here, out of the four relationships, three are `Binding` while one is `Sibling` according to the definitions in the [proposal](https://docs.google.com/document/d/1C0odRgu4qmo7x3CgTU1WLLVvznJ20vpHGED9xCE2bf0). Until we establish a concrete classification for all the attributes, making changes is problematic because it will be prone to discrepancies.

vishalvivekm:
@0xt3j4s Let's discuss this on Meshery Dev call. Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit#) if you would. :)

PR #8058, opened 2023-07-04
fixed extension infinite loading issue

### LINKED ISSUE #8043: [UI] Adapter Extension Cards improperly loading
### Description

 I’m seeing either one of two things when I load the /extensions page now:

1. a blank page
1. an indefinite, horizontal loading indicator

### Console Logs
```
[Error] Unhandled Promise Rejection: Error: The provided `as` value (/extensions) is incompatible with the `href` value (/extension/[...component]). Read more: https://nextjs.org/docs/messages/incompatible-href-as
	(anonymous function) (main-bc42d5a8245fe53d.js:1:7357)
```
Here’s a screenshot of #1
Screenshot 2023-07-03 at 2 13 49 PM

Here’s a screenshot of #2
Screenshot 2023-07-03 at 2 15 04 PM

### Environment
Browser: Safari
OS: macOS

vaibhav135:
**Notes for Reviewers**

This PR fixes #8043

The pr fixes the issue in which after loading the extension (when no extension adapter is deployed). The user was getting infinite loader.


ui/components/extensions/adapters/adapters.js
```diff
@@ -153,15 +169,6 @@ const Adapters = ({ updateProgress, classes, meshAdapters, meshAdaptersts }) =>
...
 
-const mapStateToProps = (state) => {
-  const meshAdapters = state.get("meshAdapters");
-  const meshAdaptersts = state.get("meshAdaptersts");
-
-  return {
-    meshAdapters,
-    meshAdaptersts
-  };
-};
```
theBeginner86, line 164: Can you please confirm, why we would have to use an extra fetch instead of relying on our Redux Store variables for consistency?

theBeginner86: If there is a lag between the values not being properly propagated when this UI components is loaded then we should probably check the cause for that bottleneck instead of switching over from the Redux store var to an extra GET request, IMO 🤔

WDYT?

vaibhav135: Yes there is lag. And then also issue of showing a loader. Because I am not sure how you would show the loader incase you are using a global state. I mean previously I did tried it but it resulted in more weird cases (e.g, a very short and abrupt loader). 
Since a loader should be related to the api request anyways.  Not so in this case though.

vishalvivekm:
@vaibhav135 Let's discuss this on Meshery Dev call. Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit#) if you would. :)

vaibhav135:
@Abhishek-kumar09 can you please have a look at this.

ui/components/extensions/adapters/adapters.js
leecalcote: A use case for an upgrade to redux-toolkit (now), @nebula-aac?

Abhishek-kumar09 (approved):

PR #8123, opened 2023-07-10
fix * available as version

MUzairS15:
**Notes for Reviewers**

This PR fixes #


server/meshmodel/relationships/hierarchical_inv_wasm_filters.json
```diff
@@ -6,7 +6,7 @@
     },
     "model": {
         "name": "kubernetes",
-        "version": "*",
+        "version": "v1.25.2",
```
MUzairS15, line 9: Relationships are defined under models, and each model has a version. Defining the version as `*` will result in the creation of a model with `name: Kubernetes` and `version: *`  (based on our design).
This is what was happening.

We do want the ability to mark a relationship such as it is defined for all models and this can be done in some way when we are defining selectors. https://github.com/meshery/meshery/blob/eb0bf628ad7e021ee6c318693f676efe34d42864/server/meshmodel/relationships/hierarchical_parent.json#L32

MUzairS15:
Merging by EOD.

server/meshmodel/relationships/hierarchical_inv_wasm_filters.json
leecalcote: Great comment. Agreed.

leecalcote (approved):

PR #8214, opened 2023-07-19
Fix visibility bug in catalog

sudhanshutech:
**Notes for Reviewers**

This PR fixes #
The patterns which are published will be reflect as published and rest as private and public in both cards and table which was earlier all visible as` public` and were having` publish` though it was a `published`
image
image
image


leecalcote (approved):
Looks great, @sudhanshutech

PR #8217, opened 2023-07-19
[UI] Memoized Charts And Graph Components

### LINKED ISSUE #8211: [dashboard] Unresponsive MeshModel Dashboard
Currently, when a user hovers over MeshModel charts then the page becomes unresponsive and leads to flickering of chart.

This behavior is unexpected and should be resolved for better UX for Meshery users.

aabidsofi19:
**Notes for Reviewers**
The Dashboard component is rerendering multiple times , so the charts as children were also getting rerendered a lot .
As chart component itself doesnt do any io so it is good candidate for memoizing .

Now the Chart Components won't be rerendered even if the Dashboard does .

Related To #8211  

### DIFF
--- a/ui/components/DashboardMeshModelGraph.js
+++ b/ui/components/DashboardMeshModelGraph.js
@@ -21,11 +21,11 @@ const useFetchTotal = (fetchr) => {
 
 function MeshModelContructs({ classes }) {
 
-  // API Calls 
+  // API Calls
   const totalModels = useFetchTotal(() => getMeshModels(1, 1))
   const totalComponents = useFetchTotal(() => getAllComponents(1, 1))
   const totalRelationships = useFetchTotal(() => getRelationshipsDetail(1, 1))
-  
+
   // Data Cleanup
   const data = useMemo(() => {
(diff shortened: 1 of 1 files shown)


leecalcote:
@vaibhav135 look good?

ui/components/DashboardComponent.js
```diff
@@ -1229,6 +1229,7 @@ class DashboardComponent extends React.Component {
   };
 
   render() {
+    console.info("Rerendering Dashboard")
```
vaibhav135, line 1232: @aabidsofi19 please remove this console if not needed.

vaibhav135 (changes requested):
Just a small comment. Everything else LGTM.

ui/components/DashboardComponent.js
aabidsofi19: I thought to keep it for an easy check to see if the dashboard rerenders are really increasing in production  .

vaibhav135 (approved):

leecalcote (approved):

PR #8234, opened 2023-07-20
[chore] Fix missing remote provider prefs

theBeginner86:
**Notes for Reviewers**

This PR ensure that change in remote provider prefs is persisted in local DB.

Issue:
Earlier a change in remoteProvider prefs was not persisted in local DB

Reason:
Change in remote prefs was not realized by Meshery server because there was a check for UpdatedAt value. So, if updatedAt value for changes in remoteProvder was after the changes in updatedAt value for changes in local Meshery DB only then the change was persisted.

Solution:
Do extra check to determine the diff in remoteProvider Prefs and persist in that case too. 

The need for this is because all the middleware in Meshery Server do an update for different keys in User Prefs leaving the remoteProviderPrefs. And due to middleware invoking those changes before the userPrefs UpdatedAt check so local changes were always having high UpdatedAt value and hence change in remoteProvider was rejected and not persisted

This PR fixes #

### DIFF
--- a/server/models/preference.go
+++ b/server/models/preference.go
@@ -68,7 +68,7 @@ type Preference struct {
 	AnonymousPerfResults      bool                   `json:"anonymousPerfResults"`
 	UpdatedAt                 time.Time              `json:"updated_at,omitempty"`
 	UsersExtensionPreferences map[string]interface{} `json:"usersExtensionPreferences,omitempty"`
-	RemoteProviderPreferences      map[string]interface{} `json:"remoteProviderPreferences,omitempty"`
+	RemoteProviderPreferences map[string]interface{} `json:"remoteProviderPreferences,omitempty"`
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


server/models/preference.go
```diff
@@ -68,7 +68,7 @@ type Preference struct {
 	AnonymousPerfResults      bool                   `json:"anonymousPerfResults"`
 	UpdatedAt                 time.Time              `json:"updated_at,omitempty"`
 	UsersExtensionPreferences map[string]interface{} `json:"usersExtensionPreferences,omitempty"`
-	RemoteProviderPrefs       map[string]interface{} `json:"mesheryCloudPrefs,omitempty"`
+	RemoteProviderPreferences      map[string]interface{} `json:"remoteProviderPreferences,omitempty"`
```
theBeginner86, line 71: @MUzairS15 I updated it to remoteProviderPreferences considering a different remoteProvider might not have Meshery Cloud as its name

MUzairS15: Is remote provider updated with thks change?

MUzairS15 (approved):

server/models/preference.go
theBeginner86: yes

PR #8285, opened 2023-07-26
[MeshModels] Add support for wildcard versions

Abhishek-kumar09:
**Notes for Reviewers**
Wildcard Version rules:
1. If the distinct version in the model is more than one, i.e. components from a model belongs to more than one version, then the new WildCard version is added which is given the first priority in sorting, and is the default view user gets
2. In case there is just 1 version, then the wildcard versions has no point.
3. Wildcard versions are the union of components from all the versions available in a model.

Screenshot 2023-07-27 at 12 09 43 AM

This PR fixes #

### DIFF
--- a/ui/utils/hooks/useMeshModelComponents.js
+++ b/ui/utils/hooks/useMeshModelComponents.js
@@ -1,6 +1,6 @@
 import _ from "lodash";
 
-export const WILDCARD_V = "*(All Versions)"
+export const WILDCARD_V = "All Versions"
 
 import {
   fetchCategories,


ui/utils/hooks/useMeshModelComponents.js
```diff
@@ -1,5 +1,7 @@
 import _ from "lodash";
 
+export const WILDCARD_V = "*(All Versions)"
```
leecalcote, line 3: Would it be ok if we displayed this string?

```suggestion
export const WILDCARD_V = "All Versions"
```

PR #8374, opened 2023-08-04
Show menu in subcommands under `mesheryctl system`

nebula-aac:
Removed args in certain subcommands under `system`.

Show help menu instead.

**Notes for Reviewers**
Open issue: https://github.com/meshery/meshery/issues/8400

This PR removes args in certain subcommands under 'system'. This makes the user experience a little bit jarring because when you go to navigate to certain subcommand, you expect the help menu to show up. On some of the subcommand, it didn't make sense to have args because there's already a help menu in place with other subcommands for that one.

A couple of the files that were added to the commit which I did not know until after the fact. These are just formatting issues.

### DIFF
--- a/mesheryctl/internal/cli/root/system/token.go
+++ b/mesheryctl/internal/cli/root/system/token.go
@@ -40,7 +40,9 @@ var tokenCmd = &cobra.Command{
 	Manipulate user tokens and their context assignments in your meshconfig`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if len(args) == 0 {
-			cmd.Help() // Display help information if no arguments are provided
+			if err := cmd.Help(); err != nil {
+				return errors.Wrap(err, "displaying help menu")
+			}
 			return nil
 		}
 		if ok := utils.IsValidSubcommand(availableSubcommands, args[0]); !ok {


leecalcote:
@nebula-aac are we holding onto this under commands outside of `system` have been updated or were those already reviewed?

nebula-aac:
@leecalcote any of the other commands didn't seem like it needs a help menu, so far i only saw `system` had some subcommand that would be better with a menu option. i can double check

leecalcote:
Menus… like that of `mesheryctl system login` as an example, right?

Clarification: we will still have MeshKit-style error codes with probable cause and suggested remediation, right?

nebula-aac:
@leecalcote 

Yes, whenever you run a command, some of the subcommand returns an error because we were requiring another command attached, but we wouldn't know unless we do `meshery system context`.

It will return this 

```bash
./mesheryctl system context
Error: please specify a flag or subcommand. Use 'mesheryctl system context --help' to display user guide.

See https://docs.meshery.io/reference/mesheryctl/system/context for usage details
```

When I was navigating through some of the commands, I felt it was an extra step to ask for the user to provider another flag, and then ask them use `--help` to get the commands needed for it, when you can just run it as normal and see the help menu.

And ask for `meshkit`, yes, the only reason the fields wouldn't be showing would be if you didn't add anything to. In `errors` packages from `meshkit`, we don't have a requirement to look for those fields to be filled, nor do we have a check for if a field has been nil or not.

I can make it so but it may cause some initial start for everyone to get used unless we documented in the official readme about how to setup a custom error message.

leecalcote:
Ok, so not a selectable menu, then, unlike the login command that has a menu.

suhail34 (commented):
Same problem with other subcommand too will you be doing it in this pr itself

nebula-aac:
I can check.

Philip-21:
> @leecalcote
> 
> Yes, whenever you run a command, some of the subcommand returns an error because we were requiring another command attached, but we wouldn't know unless we do `meshery system context`.
> 
> It will return this
> 
> ```shell
> ./mesheryctl system context
> Error: please specify a flag or subcommand. Use 'mesheryctl system context --help' to display user guide.
> 
> See https://docs.meshery.io/reference/mesheryctl/system/context for usage details
> ```
> 
> When I was navigating through some of the commands, I felt it was an extra step to ask for the user to provider another flag, and then ask them use `--help` to get the commands needed for it, when you can just run it as normal and see the help menu.
> 
> And ask for `meshkit`, yes, the only reason the fields wouldn't be showing would be if you didn't add anything to. In `errors` packages from `meshkit`, we don't have a requirement to look for those fields to be filled, nor do we have a check for if a field has been nil or not.
> 
> I can make it so but it may cause some initial start for everyone to get used unless we documented in the official readme about how to setup a custom error message.

@nebula-aac its best if the fields are filled because it gives the user an idea on how to proceed with a solution when an error occurs , our focus should just be making each error message for each field brief and concise

Philip-21 (commented):
@nebula-aac for the meshkit error package,  its best if the fields are filled because it gives the user an idea on how to proceed with a solution when an error occurs , our focus should just be making each error message in each field brief and concise.

nebula-aac:
@Philip-21 I agree but I'm not doing it on this PR though

That will be a separate PR once I test out the changes I can make in `meshkit`.

This PR is just to remove some of the required args in some of the subcommands.

Philip-21:
If you want to make changes in Meshkit please create an issue first, so we all can understand what you want to implement

leecalcote:
This has a dependency on https://github.com/meshery/meshkit/pull/340, is that right?

PR #8375, opened 2023-08-04
[chore] Removed OSM

nebula-aac:
**Notes for Reviewers**

This PR is related to https://github.com/meshery/meshery/issues/8305

OSM is being removed from
- [x] Meshery Server
- [x] Docker Extension
- [x] Docker Kubernetes deployment
- [x] `mesheryctl`
- [x] Meshery UI

Still needs to remove this from other places.

Also I'm removing CPX because we're not even using this, and it can be misleading on the UI that there is an adapter we may support.

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [ ] Yes, I signed my commits.

PR #8393, opened 2023-08-06
[lifecycle] Manage Un/Deployment of Meshery Adapters

theBeginner86:
**Notes for Reviewers**

- [x] Kubernetes
    - [x] Able to deploy/undeploy
   - [x] Remove OSM Adapter sub chart. (missed out in https://github.com/meshery/meshery/pull/8375)
- [x] Docker
  - [x] Able to deploy/undeploy
  - [x] Detect running Meshery container and install adapters' containers in same network

This PR fixes #

### DIFF
--- a/server/helpers/error.go
+++ b/server/helpers/error.go
@@ -158,13 +158,13 @@ func ErrDeployingAdapterInDocker(err error) error {
 }
 
 func ErrUnDeployingAdapterInDocker(err error) error {
-	return errors.New(ErrUnDeployingAdapterInDockerEnvCode, errors.Critical, []string{"Unable to undeploy adapter in k8s env"}, []string{err.Error()}, []string{"Possible issues with Docker configuration, container networking, or resource availability."}, []string{"Check Docker configuration settings, ensure containers have proper networking access, and verify available resources."})
+	return errors.New(ErrUnDeployingAdapterInDockerEnvCode, errors.Critical, []string{"Unable to undeploy Meshery Adapter in Kubernetes environment"}, []string{err.Error()}, []string{"Possible issues with Docker configuration, container networking, or resource availability."}, []string{"Check Docker configuration settings, ensure containers have proper networking access, and verify available resources."})
 }
 
 func ErrDeployingAdapterInUnknownPlatform(err error) error {
-	return errors.New(ErrDeployingAdapterCode, errors.Critical, []string{"Unable to deploy adapter given env"}, []string{err.Error()}, []string{"Your platform is not supported for deploying Meshery Adapters"}, []string{"Consider using a supported platform for deploying Meshery Adapters"})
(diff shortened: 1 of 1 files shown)


suhail34:
Doubt :
As you have removed adapters from docker-compose how will this thing work with mesheryctl as it uses the same docker-compose file and downloads it

theBeginner86:
> Doubt :
> As you have removed adapters from docker-compose how will this thing work with mesheryctl as it uses the same docker-compose file and downloads it

You can make use of Adapter Tracker Interface. It does have the info for adapters as it's seeded on Meshery boot as env vars.
FYI: https://github.com/meshery/meshery/blob/master/server/cmd/main.go#L125

theBeginner86:
This PR is ready.

System Flows:

1. For Docker env, performing un/deploying would create new containers (or stop and delete running containers) of each adapter. Would ensure that the new adapters are on the same Docker network as the one for Meshery container
2. For K8s env, performing un/deploying would internally use Helm with overrides for each adapter (enabled: true/false). Then on each un/deployment Helm upgrade is performed.

theBeginner86:
_The initial code written for this functionality is not clean would require some refactoring considering most of this functionality can be extracted into MeshKit. Like provisioning of containers or supporting multi-k8s deployment through our deployment modals._

install/docker/docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
 version: "3"
 services:
   meshery:
-    image: layer5/meshery:latest
+    image: layer5/meshery:stable-latest
```
theBeginner86, line 4: Looks like someone had changed this tag for meshery image. 🤔

As there is no image tag of latest for layer5/meshery image.  https://hub.docker.com/r/layer5/meshery/tags?page=1&name=latest

theBeginner86:
I have one more question,
Should we start publishing all Meshery Adapter's charts separately too?

So that users can directly do following for managing deployments of their individual adapters irrespective of Meshery deployment:
`helm install meshery/meshery-istio`

Instead of deploying them as follows:
```
helm pull meshery/meshery --untar
cd meshery/charts
helm install meshery-istio ./meshery-istio
```


The above would have no change in the current sub chart structure that we have its just that instead of referring a local charts dir we would be referring the remotely published charts.

The above would also be helpful when we would be deploying Meshery in K8s and trying to manage lifecycle of Adapters. As currently, since both Meshery and Adapters are part of the same release so during deployment/un-deployment of any adapter, the Meshery release is upgraded that leads to restart of Meshery pods. So, if Adapters charts are published separately then we can easily put them under separate release and that would not hinder Meshery pods.

leecalcote:
I'm not opposed to it, @theBeginner86.

@AugustasV @nebula-aac @Revolyssup thoughts?

leecalcote:
Good error messages.

leecalcote:
Error handling of local builds that try to perform deployments works.

PR #8471, opened 2023-08-15
Dry Run Support in Designs And Applications

aabidsofi19:
**Notes for Reviewers**

Add Dry Run Support in :
- Applications 
- Patterns

This PR fixes 

https://github.com/meshery/meshery/assets/65964225/ab8530c0-c7b8-440b-a8a0-08d50c75facd


ui/api/index.js
```diff
@@ -0,0 +1,15 @@
+import axios from "axios";
```
leecalcote, line 1: @aabidsofi19 make sure that you highlight this sort of pervasive change prominently in the PR description.

What is this for and what does it do?

ui/components/DryRunComponent.js
```diff
@@ -0,0 +1,326 @@
+import React from "react";
```
leecalcote, line 1: Please organize the components into respective folders. Create new folders as needed and move components around as you see fit.

aabidsofi19: The Patterns/Applications components were in root , so put it there also . i will create a new folder for this .

ui/api/index.js
aabidsofi19: sorry , forgot we were not using axios in meshery ui  , although it was in dependencies  . Btw this is the axios instance we use for data fetching (in  meshmap also in meshery-cloud ) but seems not in meshery ui . should i remove it .  @leecalcote

ui/components/DryRunComponent.js
leecalcote: Thank you much. Let's break those old bad habits.

PR #8483, opened 2023-08-16
[UI] Enhance Notification Center 

aabidsofi19:
**Notes for Reviewers**

Related to  #8485

This Pr Adds the following :

-  Custom Snackbar hook to enable sending toaster messages in notification center 
-  timestamp for events 
-  deeplink from toasts 
-  Refactored Applications Pages and Designs Page to use these new notifications for toasts 

https://github.com/meshery/meshery/assets/65964225/94d91322-71d3-4134-bf10-81865c0664de

### DIFF
--- a/ui/utils/hooks/useNotification.js
+++ b/ui/utils/hooks/useNotification.js
@@ -3,13 +3,11 @@ import { ToggleButtonGroup } from "@mui/material"
 import { useSnackbar } from "notistack"
 import { iconMedium } from "../../css/icons.styles"
 import CloseIcon from "@material-ui/icons/Close";
-import { NorthEast } from "@mui/icons-material";
 import { useDispatch } from "react-redux";
 import { pushEvent ,openEventInNotificationCenter, toggleNotificationCenter } from "../../lib/store";
 import moment from "moment";
 import { v4 } from "uuid";
-
-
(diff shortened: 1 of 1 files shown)


aabidsofi19:
TODO :
- start using these new notifications in all other places ( simple copy pasta work )

Yashsharma1911 (commented):
@aabidsofi19 I wonder if we can use hook similar that we have in [meshery cloud](https://github.com/layer5io/meshery-cloud/blob/master/ui/utility/notification-handlers.js) with customization so that we will be consistent on both side thoughts?

aabidsofi19:
Notes :

- Added Support for Class Based Components , we can use the custom withNotify HOC to pass notify as a prop in CBC  and then use it as normal 
- Making the whole snackbar causes memory leaks . it is a issue with the version of snackbar we are using , but these simple abstractions we can easily upgrade it  when we move to mui5 and react 18

leecalcote:
> Notes :
> 
> * Added Support for Class Based Components , we can use the custom withNotify HOC to pass notify as a prop in CBC  and then use it as normal
> * Making the whole snackbar clickable causes memory leaks . it is a issue with the version of snackbar we are using , but these simple abstractions we can easily upgrade it  when we move to mui5 and react 18

Great explanation.

PR #8494, opened 2023-08-16
fix crash due to empty response of filter_resource file

sudhanshutech:
**Notes for Reviewers**

This PR fixes #
> [!IMPORTANT]
> Currently the filters without filter_resource value when clicked on edit of config gets crashed. This should not happen , we can have empty filter_resource file so that on edit config the page should not crashed. This pr fixes that crash for temporary .

Example:
image
image


MUzairS15 (changes requested):
WIll you make a wrapper function around `JSON.parse` and use that wherever JSON.parse is used directly so that we don't need to add this same logic everywhere?

PR #8554, opened 2023-08-20
Fixed email-on-failure job in stable release

VaibhavMalik4187:
This job was added to notify support@layer5 via an email if any of the jobs defined in the build-and-relase-stable workflow failed.

Problem: The subject and the body of the email being sent were incorrect.

Reason: Previously the subject and the body were being generated using the variables assigned to the current job in execution i.e. email-on-failure.

Solution: Now the failed jobs are determined by iterating over all the jobs in the needs array and checking their result/status. If the status is `failure` then that particular job is put in the failed_jobs variable. And its details are appended in the failed_details variable. Now when the email will be sent, it will contain the right details.

### DIFF
--- a/.github/workflows/build-and-release-stable.yml
+++ b/.github/workflows/build-and-release-stable.yml
@@ -189,24 +189,41 @@ jobs:
     uses: meshery/meshery/.github/workflows/cncf-playground-deploy-meshery.yaml@master
     secrets: inherit
 
-  notify-via-email:
-    runs-on: ubuntu-latest
+  email-on-failure:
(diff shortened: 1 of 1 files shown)


VaibhavMalik4187:
@leecalcote @suhail34 Could you please review this PR when you have a moment?

leecalcote (approved):

leecalcote:
Let's see if this does the trick...

leecalcote:
Thanks, @VaibhavMalik4187 👏

Issue #8423, opened 2023-08-09
[mesheryctl] Enhance system dashboard to create a port forward using kubectl's


suhail34:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

#### Desired Behavior

Enhance system dashboard to create a port forward using kubectl's same approach with Meshery UI when it is deployed in kubernetes cluster

#### Mockups

---

#### Contributor Resources

- ⌨️ [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- 🙋🏾🙋🏼 Questions: [Layer5 Discussion Forum](https://discuss.layer5.io) and [Layer5 Community Slack](http://slack.layer5.io)

leecalcote:
Related https://github.com/meshery/meshery/issues/3035

Jaishree2310:
@suhail34 can you please assign this issue to me?

Jaishree2310:
@saurabh100ni ^

saurabh100ni:
Go On @Jaishree2310

AjaxAueleke:
Can I work on this if this isn't assigned to anyone @saurabh100ni?

ishwar170695:
**Status**: Already implemented

**Evidence**: The `mesheryctl system dashboard` command now supports a `--port-forward` (`-p`) flag and performs port forwarding using Kubernetes client APIs via `utils.NewPortForward()` in `mesheryctl/internal/cli/root/system/dashboard.go`.

```
portforward, err := utils.NewPortForward(
    cmd.Context(),
    client,
    utils.MesheryNamespace,
    "meshery",
    options.host,
    dashboardCmdFlags.Port,
    options.podPort,
    false,
)
```

The functionality requested in the issue is present in the current implementation, so the issue appears to have been completed and can likely be closed.

PR #8576, opened 2023-08-23
[mesheryctl] configured dashboard to update `config.yaml` to the allocated port

nebula-aac:
**Notes for Reviewers**

This PR partially fixes #https://github.com/meshery/meshery/issues/3035 and https://github.com/meshery/meshery/issues/8423

`configured dashboard to update config.yaml to the port allocated`

Will need another feature to utilize `kubectl` to proxy forward, and also if this is the case, probably remove the `port-forward` command in order to use `kubectl` to proxy forward.

### DIFF
--- a/mesheryctl/internal/cli/root/system/dashboard.go
+++ b/mesheryctl/internal/cli/root/system/dashboard.go
@@ -17,6 +17,7 @@ package system
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


suhail34:
Why do we need proxy forward if kubectl port-forwarding with `mesheryctl system dashboard --port-forward` is working fine i tried in my system

nebula-aac:
Isn't that the ask here in the open issue regarding using `kubectl` to port forward? https://github.com/meshery/meshery/issues/8423

nebula-aac:
Even though we already uses `mesheryctl system dashboard --port-forward`, anyone who uses the tool might not know this available.

It is probably assumed that when you use `mesheryctl system dashboard` that the port forwarding is already done.

I only mentioned it because we may be duplicating efforts that are not needed, or we do need to duplicate efforts in the appropriate place.

For this PR, I'm proposing to update the config.yaml when you start the deployment, then the port should be updated without manual intervention.

Last I checked, it worked for me but it might not work across all Kubernetes different deployments.

I tested `kind`, but it seemed to be finicky in `k3d`.

PR #8598, opened 2023-08-26
Prevent MeshMap from running on forked repo

VaibhavMalik4187:
Added an if condition that checks if the name of the target repository and the name of the source repository are same. If the name of the source and the base/target repos are different, the workflow will be skipped.

### DIFF
--- a/.github/workflows/meshmap.yml
+++ b/.github/workflows/meshmap.yml
@@ -3,7 +3,7 @@ on: # rebuild any PRs and main branch changes
   pull_request_target:
     types: [opened, synchronize, reopened]
 
-permissions: 
+permissions:
   actions: read
   contents: write
   security-events: write
@@ -13,6 +13,7 @@ permissions:
 
 jobs:
   MeshMapScreenshot:
+    if: ${{ github.event.pull_request.head.repo.full_name == github.repository && always() }}
     runs-on: ubuntu-latest
     steps:
       - name: Set PR number # To comment the final status on the Pull-request opened in any repository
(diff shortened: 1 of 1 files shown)


VaibhavMalik4187:
@theBeginner86 @leecalcote @suhail34 To stop MeshMap SnapShot action from running on forked repos, I've created this test PR. Your review would be of great help here.
Initially, I created an issue in the MeshMap SnapShot repo under layer5labs. However, if we disable running the action in forked repos, we won't be able to test it at all. Hence, I've come to this solution i.e. add an if statement at the top of the job which uses this action to prevent it from running in the forked repos.

leecalcote:
> @theBeginner86 @leecalcote @suhail34 To stop MeshMap SnapShot action from running on forked repos, I've created this test PR. Your review would be of great help here.
> Initially, I created an issue in the MeshMap SnapShot repo under layer5labs. However, if we disable running the action in forked repos, we won't be able to test it at all. Hence, I've come to this solution i.e. add an if statement at the top of the job which uses this action to prevent it from running in the forked repos.

Makes sense to me. Once this is proven as the only route to go, then the workflow template in layer5labs will need to be updated.

.github/workflows/meshmap.yml
```diff
@@ -13,6 +13,7 @@ permissions:
 
 jobs:
   MeshMapScreenshot:
+    if: ${{ github.event.pull_request.head.repo.full_name == github.repository && always() }}
```
leecalcote, line 16: Thanks. I'm also looking for this to be updated in the workflow template that is created when a user connects Meshery to their GitHub repo, @VaibhavMalik4187

leecalcote (approved):

Issue #8664, opened 2023-09-06
`email-on-failure` job sends email even when there is no failure


VaibhavMalik4187:
### Current Behavior

The build-and-release-stable workflow sends an email even if every job passed.
![image](https://github.com/meshery/meshery/assets/34866732/617d015e-8810-4938-bccd-150454999746)

### Expected Behavior
Email should not be sent if there is not failure

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#documentation-contribution-flow)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](http://slack.layer5.io)

PR #8665, opened 2023-09-06
Email will not be sent without any failure

VaibhavMalik4187:
Removed extra string: `EMAIL_BODY=` from the environment variable EMAIL_BODY.

Fixes: https://github.com/meshery/meshery/issues/8664

### DIFF
--- a/.github/workflows/build-and-release-stable.yml
+++ b/.github/workflows/build-and-release-stable.yml
@@ -191,7 +191,7 @@ jobs:
 
   email-on-failure:
     needs: [update-rest-api-docs, update-graphql-docs, build, ctlrelease, call-dde-release-workflow, call-helm-chart-releaser, email-meshery-release-notes-workflow, call-cncf-playground-rollout]
-    if: ${{ always() }}
+    if: ${{ failure() }}
     runs-on: ubuntu-latest
     steps:
       - name: Find failed jobs.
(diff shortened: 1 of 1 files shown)


Ghat0tkach:
@VaibhavMalik4187 Let's discuss this on Meshery Dev call. Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit#) if you would. :)

suhail34 (changes requested):
https://github.com/meshery/meshery/blob/56eb7b793a434fd2c6352ff8ef92e90c8a507996/.github/workflows/build-and-release-stable.yml#L194
I think the problem is here the condition should be something like this
`if: ${ always() && (needs.update-grapql-docs.result == 'failure' || needs.update-swagger-docs.result=='failure'and so on for other jobs) }`

VaibhavMalik4187:
> https://github.com/meshery/meshery/blob/56eb7b793a434fd2c6352ff8ef92e90c8a507996/.github/workflows/build-and-release-stable.yml#L194
> 
> I think the problem is here the condition should be something like this
> `if: ${ always() && (needs.update-grapql-docs.result == 'failure' || needs.update-swagger-docs.result=='failure'and so on for other jobs) }`

This is yet another way to check if we should start the job or not. However, this looks very tedious, especially for workflows having multiple jobs.
Do you mind if I ask what is wrong with:
```
      - name: Send Email Notification
        if: ${{ env.EMAIL_BODY != '' }}
```

suhail34:
Okay i got you now we can do that but if there are no jobs failed then why do we need to run email-on-failure we can stop it the start itself

VaibhavMalik4187:
> Okay i got you now we can do that but if there are no jobs failed then why do we need to run email-on-failure we can stop it the start itself

Your approach suggests checking for failed jobs by hard coding the values and then comparing the result of each job with "failure". This violates the DRY principle. It is also prone to error. Suppose, later, at some point, 5 more jobs are added in this workflow. The comparisons would keep on growing. That's why I'm using the REST API to determine the failed jobs without hardcoding any values. And just as you suggested, I'm also skipping the `Send email` step if none of the jobs have failed.

suhail34:
Yes you are right that was just a suggestion what i meant was to check if there was a failure in any of the previously runned job then only email-on-failure job should run we can also try
`if: failure()` so that it will run only on failure of previously runned job

VaibhavMalik4187:
> Yes you are right that was just a suggestion what i meant was to check if there was a failure in any of the previously runned job then only email-on-failure job should run we can also try `if: failure()` so that it will run only on failure of previously runned job

Great suggestion, I've changed the code accordingly.

suhail34 (approved):
LGTM 👌

PR #8691, opened 2023-09-08
Reduce unnecessary queries to remote provider.

MUzairS15:
**Notes for Reviewers**

This PR fixes #
1. Fetch connection credentials when required. (Specifically for K8s context UI does not have any use for credentials).
2. After importing a Kubeconfig from `Settings`, UI was making another request to add an entry for credentials, which is not required since when creating a connection, the credential is added implicitly. This other request was causing another credential to be created, which didn't impact any flow but was obsolete.
3. Stopped fetching credentials for K8s subscription as not required. (This has very very less impact, since the subscription only reaches to Remote Provider, when context was updated/created/deleted, but since UI do not require this, credentials are no longer fetched).

### DIFF
--- a/server/helpers/component_info.json
+++ b/server/helpers/component_info.json
@@ -1,5 +1,5 @@
 {
   "name": "meshery-server",
   "type": "component",
-  "next_error_code": 1534
+  "next_error_code": 1536
 }


MUzairS15:
Meshkit Error Utility action is not working on latest commit in this PR, ad I have updated next error code and the utility doesn’t report any error locally.

PR #8704, opened 2023-09-09
Show duplicates meshmodels on toggle

sudhanshutech:
**Notes for Reviewers**

This PR fixes #
To show duplicates meshmodels only when toggle

Whole earlier logic for duplicated was off, we already were showing duplicates in the table and also a sub-table in which the number of duplicates were not matching.
So I add frontend logic that filter and show only unique data when toggle is off and when on all data including duplicates are shown


sudhanshutech (commented):
Only one thing is weird that when unique values being shown, only single data is shown in the a page and when duplicates is switched then its working good

leecalcote (approved):

PR #8776, opened 2023-09-15
Fix crash on saving applications

sudhanshutech:
**Notes for Reviewers**

This PR fixes #
The crash and bug on saving applications


sudhanshutech:
@MUzairS15 i think we are not focusing on applications since will soon merge with designs. But one thing to notice that when name updates for any apps its updated for` appliaction_file` not directly the name column like for` pattterns`  thats why the columns seems empty

PR #8787, opened 2023-09-15
[UI] Prevent rjsf modal rerendering

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

Publish and import modal was losing focus when you open them in start, reason for that is Filter and Pattern files have too much logic which makes it rerender so many times and because of it Modal.js is also rerendering which cause this error

This PRs uses React memo and conditional rendering here to solve this issue and remove dead code

### DIFF
--- a/ui/components/Filters.js
+++ b/ui/components/Filters.js
@@ -25,7 +25,6 @@ import { toggleCatalogContent, updateProgress } from "../lib/store";
 import TableSortLabel from "@material-ui/core/TableSortLabel";
 import dataFetch from "../lib/data-fetch";
 import PromptComponent from "./PromptComponent";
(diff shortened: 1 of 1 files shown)


Yashsharma1911:
Okay it's fully fixed now, renders Modal.js only when it gets clicked and prevents losing focus issue

Yashsharma1911:
Actually the changes I made over here fix the issue properly, but as soon I merge latest changes this makes it worse so need to look what's wrong and will make changes tomorrow

Yashsharma1911:
Okay so it was cache 😬, I literally spend so much time, this PR fixes the issue I'm merging this PR so that this issue can be resolved

PR #8803, opened 2023-09-16
Update ConnectionCharts.js

gautammanak1:
In the shorter version of the code I provided, I made several changes to remove redundant elements and simplify certain parts of the code for brevity. Here's a breakdown of those changes

**Notes for Reviewers**
**Destructuring Data**: Instead of destructuring json.connections_status directly in the getConnectionStatusSummary().then callback, I've used object destructuring to directly access connections_status. This makes the code shorter and more readable.

**Before:**

getConnectionStatusSummary().then(json => {
  setChartData(json.connections_status
    .filter(data => isValidColumnName(data.status))
    .map(data => [data.status,data.count]))
})

**After:**

getConnectionStatusSummary().then(({ connections_status }) => {
  const filteredData = connections_status
    .filter(data => isValidColumnName(data.status))
    .map(data => [data.status, data.count]);
  setChartData(filteredData);
});

**Arrow Functions**: I've replaced some regular functions with arrow functions where the function body is just a return statement. This simplifies the function syntax.

**Before:**

format: function (value) {
  return value;
},
After:

format: (value) => value,

**Shortened JSX**: I've retained the structure and elements in the JSX, but I've removed some excessive spacing and redundant styles, making the code more concise.

Code Formatting: I've formatted the code to follow consistent indentation and spacing, making it more readable.

### DIFF
--- a/ui/components/Dashboard/ConnectionCharts.js
+++ b/ui/components/Dashboard/ConnectionCharts.js
@@ -1,83 +1,83 @@
-import React , { useEffect,useState } from "react";
-import {  Box , Typography } from "@material-ui/core";
+import React, { useEffect, useState } from "react";
+import { Box, Typography } from "@material-ui/core";
 import { donut } from "billboard.js";
 import BBChart from "../BBChart";
-import { dataToColors , isValidColumnName } from "../../utils/charts"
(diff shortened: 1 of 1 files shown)


GAUTAMMANAK1:
Done

Issue #8661, opened 2023-09-05
[UI] Setup Prettier


nebula-aac:
In Meshery UI, we do not use `prettier`, which means our code looks different throughout each files.

Even though we have Next.js lint for its linting capabilities, the linting job is not doing all what it is supposed to do with the current rules on certain formatting. However, for a true formatter, we're not using prettier, and it would be great to set one up here in Meshery UI and also setup pre-commit hook for this.

Here is what being used in `Meshery Cloud`, but it can be improved.

Please note that making this change will create a large PR. So it would be necessary to check through all the files that will be formatted, and probably won't need to be formatted.

The referenced file is at https://github.com/layer5io/meshery-cloud/blob/master/ui/.prettierrc.yml

```yaml
arrowParens: avoid
tabWidth: 2
trailingComma: none
printWidth: 90
```

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.layer5.io) and [Community Slack](http://slack.layer5.io)

dhiraj0911:
Hey @nebula-aac, I'd like to work on this issue, can you assign this to me?

Yana-Gupta:
@nebula-aac I would like to work on this issue. Would you assign me.

ShatilKhan:
Hi @nebula-aac , I would like to understand this issue a bit more. What kind of changes are we looking for?
I have some front-end experience , if possible do assign this issue to me .

aabidsofi19:
+1 @nebula-aac

arpitghura:
Hey, I would like to work on this issue. Can you please assign it to me? @nebula-aac

PR #8827, opened 2023-09-19
[UI] feat: Setup prettier and moved lint-staged to its own file

nebula-aac:
**Notes for Reviewers**

This PR fixes https://github.com/meshery/meshery/issues/8661

Here is the proposed `prettier` config, where some of the existing formatting rules in `eslint` can be migrated to allow Prettier format the files for you.

```
arrowParens: always
bracketSameLine: false
bracketSpacing: true
jsxSingleQuote: false
proseWrap: preserve
quoteProps: as-needed
semi: true
singleQuote: true
tabWidth: 2
trailingComma: all
printWidth: 100
useTabs: false
```

I'm testing other rules, and may have removed something that you feel that it should be in the current config.

Please note that in order for this rules to work, you need to use `next lint` instead of `eslint` to lint the files, because `eslint` and `next lint` works differently from each other.

The `lint-staged` has been moved to its own config file for clarity and also takes care linting and formatting with the scripts available.

Feedback is welcomed, and this PR is by no means the definite PR, this can be improved with the configs.


Abhishek-kumar09 (commented):
The lint changes looks good. Since, I am not aware of the motivation behind due to lack of availability, I would comment less. But the new config looks good.

ui/.eslintrc.js
```diff
@@ -1,99 +1,60 @@
...
-    "react/jsx-uses-vars" : [2],
-    "react/jsx-no-undef" : "error",
-    "no-console" : 0,
-    "no-unused-vars" : "error",
-    "react/jsx-key" : "warn",
-    "no-dupe-keys" : "error",
-    "react/jsx-filename-extension" : [
+  plugins: ['react', 'cypress', 'prettier'],
+  rules: {
+    '@next/next/no-img-element': 'off',
```
Yashsharma1911, line 35: @nebula-aac All looks good though I have a question should be doing same for meshery cloud as we are not using next Image there too, and getting bunch of lint error on that or we should not silence it?

Yashsharma1911 (approved):

ui/.eslintrc.js
nebula-aac: @Yashsharma1911 I have it turned off now because in using Image from Next, it requires the width and height to be used, and some of the images we have, we're doing without. It would be a good idea to turn it off in Meshery Cloud too.

nebula-aac: @Yashsharma1911 can you approve again? I updated the lint staged to fix the linting at commits

Yashsharma1911: Sure

Yashsharma1911 (approved):

leecalcote:
tabWidth: 2 - nice.

ui/README.md
```diff
@@ -3,36 +3,41 @@
...
-    > **NOTE:** Please run the steps in order to avoid issues, as Meshery server should be running and logged-in before accessing the development server
-    > on 3000 port.
 
-    > **NOTE:** Its strongly recommended to use  either [Node Version Manager](https://github.com/nvm-sh/nvm#node-version-manager---) in linux/mac os systems or [NVM for Windows](https://github.com/coreybutler/nvm-windows#nvm-for-windows) on Windows systems so single `nvm use` / `nvm install` simplifies installing and using correct node version locallly **(v18)**, see [NVM Intro](https://github.com/nvm-sh/nvm#intro) for details. Otherwise, you might experience issues during local `npm i` similar to [4674](https://github.com/meshery/meshery/issues/4674) due to how optional dependencies are resolved in npm v6.
-    
+   > **NOTE:** Please run the steps in order to avoid issues, as Meshery server should be running and logged-in before accessing the development server
```
leecalcote, line 20: Could we please have these notes added to one or more of the contributing docs on docs.meshery.io?

leecalcote:
Updates needed in:
1. UI Docs https://docs.meshery.io/project/contributing/contributing-ui
2. Windows Docs - https://docs.meshery.io/project/contributing/meshery-windows

What instructions do we have or preconfigured settings can we share for popular IDEs like VSCode?

nebula-aac:
> Updates needed in:
> 
> 1. UI Docs https://docs.meshery.io/project/contributing/contributing-ui
> 2. Windows Docs - https://docs.meshery.io/project/contributing/meshery-windows
> 
> What instructions do we have or preconfigured settings can we share for popular IDEs like VSCode?

I can create a separate PR for this.

At this time, VS Code will pick up whatever settings is in the prettier config, but depending on the existing setup, it may get confusion. So we may need additional configs to sure this works properly and lessen confusions. Probably by way of using VSCode settings.

PR #8833, opened 2023-09-20
[UI] Add Validation and Error Boundaries 

aabidsofi19:
**Notes for Reviewers**

This PR fixes the crash of ui on invalid/corrupt events and adds error boundaries for gracefull error handling 

# Notes :
- I have used react-error-boundary  ( we use it in other projects also )
- ajv for validation of response ( i have hardcoded the event schema for now but we can also use openapi codegen )

### DIFF
--- a/ui/pages/_app.js
+++ b/ui/pages/_app.js
@@ -100,15 +100,18 @@ const EventsSubsciptionProvider = () => {
       console.error("Invalid event received",result)
       return;
     }
-
-    rtkStore.dispatch(pushEvent(validatedEvent))
-    rtkStore.dispatch(mesheryApi.util.invalidateTags([PROVIDER_TAGS.EVENT]))
-    notify({
-      message : validatedEvent.description,
-      event_type : validatedEvent.severity,
-      id : validatedEvent.id,
-      showInNotificationCenter : true,
-    })
+    try {
+      rtkStore.dispatch(pushEvent(validatedEvent))
+      rtkStore.dispatch(mesheryApi.util.invalidateTags([PROVIDER_TAGS.EVENT]))
+      notify({
+        message : validatedEvent.description,
+        event_type : validatedEvent.severity,
+        id : validatedEvent.id,
+        showInNotificationCenter : true,
+      })
+    } catch (e) {
+      console.error("Error While Storing Event --Event-Subscription ",e)
+    }
   }), [])
 
   useEffect(() => {


aabidsofi19:
@theBeginner86

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "dependencies": {
-    "google-sheet-cli": "^1.6.0"
+    "google-sheet-cli": "^1.6.0",
+    "react-error-boundary": "^4.0.11"
```
nebula-aac, line 4: Any reason the `react-error-boundary` is added through root level?

This should be in `ui` to avoid confusion, and also this entire repo isn't a true monorepo, so we're not using this `package.json` at the root level.

aabidsofi19: oh ,thanks for points , its by mistake .

ui/components/General/ErrorBoundary.js
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <ReactErrorBoundary FallbackComponent={Fallback} {...props}>
+      {children}
+    </ReactErrorBoundary>
+  );
+}
+
+export const withErrorBoundary = (Component) => {
+  const WrappedWithErrorBoundary = (props) => (
+    <ErrorBoundary FallbackComponent>
```
MUzairS15, line 24:
```suggestion
    <ErrorBoundary Fallback>
```
Or I am missing something?

ui/components/NotificationCenter/constants.js
```diff
@@ -37,4 +39,86 @@ export const SEVERITY_STYLE = {
     color : NOTIFICATIONCOLORS.WARNING
   },
 
+}
+
+//TODO: This should be generated from OPENAPI schema
+const EVENT_SCHEMA = {
+  type : "object",
```
MUzairS15, line 46: The schema is already present but the contract b/w fronted and backend is missing.

also, this schema is inaccurate as it lacks category and action

ui/components/NotificationCenter/constants.js
```diff
@@ -37,4 +39,86 @@ export const SEVERITY_STYLE = {
...
+export const validateEvents = (events) => {
+  return events.map((event) => {
+    const [isValid, validatedEvent] = validateEvent(event)
+    return isValid ? validatedEvent : null
+  }).filter((event) => event)
+}
+
+
+
+const EVENT_METADATA_SCHEMA = {
```
MUzairS15, line 99: This isn’t always the case, in long operations i.e. an operation which invoked multiple other operation the metadata contains a summary which is not an error specifically.

ui/components/General/ErrorBoundary.js
```diff
@@ -0,0 +1,30 @@
+import { ErrorBoundary as ReactErrorBoundary } from "react-error-boundary";
+
+function Fallback({ error }) {
+  // Call resetErrorBoundary() to reset the error boundary and retry the render.
+
+  return (
+    <div role="alert">
+      <p>Something went wrong:</p>
```
leecalcote, line 8:
```suggestion
      <h3>Please pardon the mesh</h3>
```

leecalcote (approved):

MUzairS15 (commented):
We can add error boundary to the top level app , also the component lvl boundary could  be made more specific to include the comp which caused the error. As stacktrace might be difficult to trace through sometimes.

in case of multiple error boundary closest one gets executed

ui/components/NotificationCenter/constants.js
aabidsofi19: Thanks Uzair , i have not considered that . do you have an  example response for this kind of situation

theBeginner86 (commented):
> We can add error boundary to the top level app , also the component lvl boundary could be made more specific to include the comp name or filename which caused the error. As stacktrace might be difficult to trace through sometimes.
> 
> in case of multiple error boundary closest one will be executed

+1

I too think that adding error boundary at root level would be more beneficial. As by doing that we would resolve not only the notifications center issues but unwanted UI crashes in general. Would help us to be future-proof

MUzairS15 (approved):
Follow up on review suggestions separately.

aabidsofi19:
> > We can add error boundary to the top level app , also the component lvl boundary could be made more specific to include the comp name or filename which caused the error. As stacktrace might be difficult to trace through sometimes.
> > in case of multiple error boundary closest one will be executed
> 
> +1
> 
> I too think that adding error boundary at root level would be more beneficial. As by doing that we would resolve not only the notifications center issues but unwanted UI crashes in general. Would help us to be future-proof

error boundary at root level is good to catch all the errors but that also means we wont have a usable ui on error .  the more granular we go the better ux  and easy debugging . But I  Agree to your point that there should be a balance and  every component might not need it .

PR #8875, opened 2023-09-22
Fix Notification center bell icon toggle btn

aabidsofi19:
**Notes for Reviewers**

This PR fixes #

The bug was how we manage clickAwayListeners for globals event , so basically it was getting fired on clicking the open icon itself . 

i have fixed it here and also added helper functions and framework to streamline the process : 

## Here is a general / proposed  approach to follow : 

we  check whether the click event occurred within a specific component (represented by `drawerRef`) or on a particular UI element (represented by `toggleButtonRef`). If so, no further action is taken. 

The handler checks if the clicked element has a specific CSS class (identified by NOTIFICATION_CENTER_TOGGLE_CLASS). If the class is present, the handler exits early without taking any action. ( this is for if we have action buttons like the toggle in toast ,  you might have noticed that it also had the same issues ) . i have created a function `hasClass` to check if the toggleClass is present among the component or any of its parent .  

There's a special case handling for SVG icons, where it checks if the clicked element has a class name containing `MuiSvgIcon.` If this condition is met, the handler also exits early. ( this special check for notification center because we dont want it to close for dismissing notifs or minor stuff ) 

If none of the above conditions are met, the assumption is that the click event occurred outside of the designated UI elements, prompting the handler to perform a specific action, such as closing a UI component (represented by the handleClose function).

### DIFF
--- a/ui/utils/Elements.js
+++ b/ui/utils/Elements.js
@@ -1,0 +1,37 @@
+// Description: Contains functions related to DOM elements and react components.
+
+import React from "react"
+
+
+// recursively check if element or any of its parent has the class
+export const hasClass = (element, className) => {
+
+  try {
+    if (typeof element?.className == 'string' && element?.className?.includes(className)) {
+      return true
+    }
+    if (element?.parentElement) {
+      return hasClass(element.parentElement, className)
+    }
+  } catch (e) {
+    console.error(`Error in hasClass while checking for ${className} in `, element, e)
+  }
+  return false
+}
+
+// recursively got throught component and its children and add the class to each of them
+// This is required to prevent the clickaway listner from blocking the click event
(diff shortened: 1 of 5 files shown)


ui/utils/Elements.js
```diff
@@ -0,0 +1,37 @@
...
+      return hasClass(element.parentElement, className)
+    }
+  } catch (e) {
+    console.error(`Error in hasClass while checking for ${className} in `, element, e)
+  }
+  return false
+}
+
+// recursively got throught component and its children and add the class to each of them
+// This is required to prevent the clickaway listner from blocking the click event
```
aabidsofi19, line 23:
```suggestion

```

MUzairS15 (approved):

PR #8893, opened 2023-09-23
[extensions] Fix Collaborator Extension Point Re-rendering

theBeginner86:
**Notes for Reviewers**

This PR fixes the collaborator extension point re-rendering by the use of `shouldComponentUpdate` function of React. Reference: https://legacy.reactjs.org/docs/optimizing-performance.html#avoid-reconciliation

Now we have a redux state that manages whether the extension point is mounted. Once mounted we would not be remounting it.

This PR fixes #

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [ ] Yes, I signed my commits.

PR #8904, opened 2023-09-24
[extensions] Fix Collaborator Extension Point Re-rendering

theBeginner86:
Earlier, in this PR (https://github.com/meshery/meshery/pull/8893) I was skipping re-rendering of entire header component by the use of shouldComponentUpdate function of React. Reference: https://legacy.reactjs.org/docs/optimizing-performance.html#avoid-reconciliation. That fixed the issue of collaborator ext re-rendering but lead to other issues like title not updating, and dark mode not applying etc.

Now, if have just memoized the extension point sandbox component using https://react.dev/reference/react/memo so any of our extension point would not be re-rendered until the requested type changes. This ensures that the collaborator's extension point doesn't re-render.

I've tested thoroughly, it doesn't lead to any side affects now. Event checked the behavior of other extension points and all work as desired.

### DIFF
--- a/ui/components/Navigator.js
+++ b/ui/components/Navigator.js
@@ -806,15 +806,16 @@ class Navigator extends React.Component {
    * Removes id from openitems if present
    * Adds id in openitems if not present already
    */
-  toggleItemCollapse(id) {
+  toggleItemCollapse = (itemId) => {
+    const isItemOpen = this.state.openItems.includes(itemId);
     const activeItems = [...this.state.openItems];
-    if (this.state.openItems.includes(id)) {
-      this.setState({ openItems: activeItems.filter((item) => item !== id) });
+    if (isItemOpen) {
+      this.setState({ openItems: activeItems.filter((item) => item !== itemId) });
     } else {
-      activeItems.push(id);
-      this.setState({ openItems: activeItems });
+      activeItems.push(itemId);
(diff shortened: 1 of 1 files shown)


aabidsofi19 (approved):
:clap:  . Just for curiosity did you find why was getting rerrendered so much .

theBeginner86:
> 👏 . Just for curiosity did you find why was getting rerrendered so much .

Yes, it is due to the fact that our header component re-renders continuously. The reason for that is because it has `title`, `k8s context switcher`, `settings` and `user avatar` as child component. And also the recently introduced user collaborator extension point component.
Each of these component especially title and k8s context switcher relies on the redux store vars which are passed to the header component as props. Since, the value of title props changes on navigating to a different page (because the title of each page is different) so the header component re-renders it self. Since, the parent component re-renders so that leads to re-rendering of child component. And naturally user collaborator extension point being its child component was repeatedly re-rendered due to it.

Now this PR would memoized the `ExtensionSandbox` component (the one which is responsible for dynamically rendering our extension points, like user avatar in this case). This memoization is based on the fact that only if the `type` of the extension changes (which is passed to the ExtensionSandbox Component as props) then we would re-render that component. This ensure that even thought the header component would re-render based on navigating to different pages the ExtensionSandbox component (that is, collaborator extension point) would not re-render because the type would remain same.

Issue #8946, opened 2023-09-27
[UI] Dashboard: Invalid Hook Call


leecalcote:
### Current Behavior
This is affecting all users, all browsers, and is on the Dashboard:

<img width="884" alt="Screenshot 2023-09-27 at 4 10 57 PM" src="https://github.com/meshery/meshery/assets/7570704/711abf2a-bed0-45c1-9ce0-fbc9997b1472">

### Environment
**Meshery:** v0.6.148

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](http://discuss.meshery.io) and [Community Slack](https://slack.meshery.io)

PR #8948, opened 2023-09-27
fix: Remove useTheme from emotion react, and call theme as props

nebula-aac:
**Notes for Reviewers**

This PR fixes [#8946](https://github.com/meshery/meshery/issues/8946)

I remove `useTheme` from `Emotion` as this is a custom hook and can't be used directly in a Class Components.

I revised the `theme` to be used as a props instead of a custom hook from `Emotion`.


leecalcote (approved):
😅 a much needed fix!

PR #8963, opened 2023-09-29
Add debouncing to subscription

aabidsofi19:
**Notes for Reviewers**

This PR fixes #
The root cause of context polling issue seems to arise from the k8sSubscription itself ,  it is continuosly returning an updated version of the config ,  we prev fixed the comparison for checking if config is updated and it has reducted a few of the queries but in playground the config is really getting updated  frequently . (which  might have something to do with it being used by multiple player , not sure) . which then later invalidates the subscriptions to clusterresources other related ones inside the dashboard .   FY : the part of the config that gets changed is the context's updated_at field

I have `debounced` the callback for updating the config in subscription to 2sec for now which should slow down things for now . config is updated only inside the callaback . 

### DIFF
--- a/ui/lib/store.js
+++ b/ui/lib/store.js
@@ -353,7 +353,8 @@ export const openEventInNotificationCenter =
   };
 
 export const makeStore = (initialState, options) => {
-  return createStore(reducer, initialState, composeWithDevTools(applyMiddleware(thunkMiddleware)));
+  const composeEnhancers = composeWithDevTools({ trace: true, traceLimit: 50 });
+  return createStore(reducer, initialState, composeEnhancers(applyMiddleware(thunkMiddleware)));
 };
 
 export const resultsMerge = (arr1, arr2) => {


aabidsofi19:
image

aabidsofi19:
this is continuosly sending updated contexts : 
image

leecalcote:
@MUzairS15 did we figure out what do here?

MUzairS15:
I went through the k8s subscription and server notifies the UI only in case of
1. When contexts are loaded and persisted from the config. (LoadAndPersist function)
2. User manually add or delete the context (KubeconfigHandler and DeleeKubeconfig)

saurabh100ni:
Let's discuss this on the Meshery dev call tomorrow,
Please add this as your agenda item in Docs: https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit

leecalcote:
Close?

aabidsofi19:
yup

Issue #9003, opened 2023-10-03
[Server] Controller Mutex Crash


leecalcote:
### Current Behavior
Meshery Server crashes when the Vizualizer extension is opened. 

[server-log.txt](https://github.com/meshery/meshery/files/12787952/server-log.txt)

### Environment

- **Host OS:** Mac
- **Platform:** local
- **Meshery Server Version:** stable-v
- **Meshery Client Version:** stable-v

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#documentation-contribution-flow)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](http://discuss.meshery.io) and [Community Slack](https://slack.meshery.io)

Shiva953:
According to me, this is the problematic snippet that seems to be causing the error

https://github.com/meshery/meshery/blob/5d79e68f0ba5acc77f7c002d3fd6c7026897894f/server/models/meshery_controllers.go#L127C1-L152C2

Not sure about the exact cause though.

leecalcote:
@Shiva953 thank you for this. 👍

PR #9015, opened 2023-10-04
[Server] Fix race cond in k8s comp registration

MUzairS15:
**Notes for Reviewers**

This PR fixes the meshery Server crash reported here #9003.
Additionally, it also fixes the K8s comp registration.
The `ctxRegStatusMap` map is a shared resource. The read/write needs to be atomic. 
The map was accessed here https://github.com/meshery/meshery/pull/9015/files#diff-23d91387bec0232bd30f0206cf1ec7a8b114302fe631c782d2df879de85103a5R89 
w/o acquiring a mutex, it can be the case that in a separate goroutine the lock is acquired and the line https://github.com/meshery/meshery/pull/9015/files#diff-23d91387bec0232bd30f0206cf1ec7a8b114302fe631c782d2df879de85103a5R96 is executed.
Since the previous read is not synchronised it will lead to a race condition (concurrent read/write).

### DIFF
--- a/server/internal/graphql/resolver/operator.go
+++ b/server/internal/graphql/resolver/operator.go
@@ -87,6 +87,7 @@ func (r *Resolver) changeOperatorStatus(ctx context.Context, provider models.Pro
 			return
 		}
 		op, _ := ctx.Value(models.MesheryControllerHandlersKey).(map[string]map[models.MesheryController]controllers.IMesheryController)
+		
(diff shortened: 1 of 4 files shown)


leecalcote:
@Shiva953 here's the PR that was posted and included in the v0.6.154 release. The issue is still occurring, though, so we're still on the hunt for a fix.

PR #9024, opened 2023-10-05
Fix ui issues

sudhanshutech:
**Notes for Reviewers**

This PR fixes #
1. Custom search icon  to scale in every browser
2. odd behaviour of column dropdown
3. odd behaviour of search expansion

### DIFF
--- a/ui/components/MeshConfigComponent.js
+++ b/ui/components/MeshConfigComponent.js
@@ -482,7 +482,7 @@ function MesherySettingsNew({ classes, updateProgress, operatorState, k8sconfig 
       name: 'location',
       label: 'Location',
       options: {
-        filter: true,
+        filter: false,
         sort: true,
         searchable: true,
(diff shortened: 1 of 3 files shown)


ui/assets/icons/search.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import { makeStyles } from "@material-ui/core/styles";
```
leecalcote, line 2: No more makeStyles.

Please use styled components. (MUI v5 compatible)

leecalcote (changes requested):

ui/assets/icons/search.js
sudhanshutech: The thing all the Hoc and parent components are styles with v4 styles thats why when v5 compatible styles they either gets override or have conflicts and don't work

PR #9152, opened 2023-10-19
[UI] Disable credential in settings

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

This PR disable credentials in Meshery to make sure user doesn't get a bad user experience, as we will come up with design spec for credential we will enable this again


Yashsharma1911:
@theBeginner86 this PR does not do mapping of workspace management to role I think this can be done in UI itself so, if that's right let's merge this

leecalcote (approved):

PR #9170, opened 2023-10-23
[Server] K8s connection management 

MUzairS15:
**Notes for Reviewers**

This PR fixes #
1. Show receipt when the kubeconfig is uploaded.
2. Manage contexts which have the state as `connected`. (Dynamic k8s registration).
   I have changed the logic to only query for context which has state as `connected`, this logic should handle all cases.
   The subsequent PRs will be handling the logic for controlling state transitions completely, i.e. in this PR state transitions are not supported. Operations like deploying operators will right now be controlled from UI, i.e. Backend does not have restrictions if UI requests an invalid transition.

In all the current functionality (as present in the master) is not affected but an additive change with improvements.
3. For UI, I have added an API to retrieve all possible state transitions for a particular connection kind.

### DIFF
--- a/server/helpers/component_info.json
+++ b/server/helpers/component_info.json
@@ -1,5 +1,5 @@
 {
   "name": "meshery-server",
   "type": "component",
-  "next_error_code": 1540
+  "next_error_code": 1541
 }
--- a/server/models/error.go
+++ b/server/models/error.go
@@ -98,7 +98,7 @@ const (
 	ErrPersistEventCode                   = "1533"
 	ErrUnreachableKubeAPICode             = "1534"
 	ErrFlushMeshSyncDataCode              = "1535"
-	ErrUpdateConnectionStatusCode       = "replace_me"
+	ErrUpdateConnectionStatusCode         = "1540"
 )
 
 var (


MUzairS15:
// @leecalcote @theBeginner86

leecalcote:
I'm removing "[WIP]", @MUzairS15 ...

server/handlers/connections_handlers.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/layer5io/meshkit/models/events"
 )
 
+type connectionStatusPayload map[uuid.UUID]models.ConnectionStatus
```
leecalcote, line 16: @MUzairS15, does this map to the Status attribute in the Connections OpenAPI schema (are we being schema-driven)?

leecalcote:
@MUzairS15, as the Connections page goes live in Meshery UI (@senali-d, @Yashsharma1911), the Connections API will need to respond with the available set of states to which each individual connection can transition given that connection's current state.

leecalcote:
@MUzairS15 a few lint checks could use your attention.

MUzairS15:
@yash37158 or @senali-d or @Akshay Sharma can you handle the styling here for the config upload modal?
@ritiksaxena124 Wil you ensure someones takes the above item?

MUzairS15:
Screenshot 2023-11-01 at 8 20 40 PM

There are 3 lists, `Connected, Registered and Ignored contexts`, make them collapsible and make sure they follow the Meshery UI design.

yash37158:
sure @MUzairS15, I will fix that

MUzairS15:
Screenshot 2023-11-06 at 12 08 37 PM

// @leecalcote Merging .

leecalcote:
@MUzairS15 where is the “?” that we discussed?

PR #9313, opened 2023-11-06
Use Sistent components and packages

sudhanshutech:
**Notes for Reviewers**

This PR adds our Sistent components and packages.
Since sistent is all supporting v5 mui so some components are not able to migrate directly to sistent ones it breaks.
Since these components are also using theme colors which is also supporting v4 so the colors are also not applying.

Currently i have used sistent only those components that should not break. I am also trying to migrate some styles to v5 styles components parallel to test.

I will make a another pr to migrate the theming to v5


sudhanshutech:
@theBeginner86 should i check any other build to confirm that nothing breaks when merged

abhijeetgauravm:
Let's discuss this on Meshery Development Meeting tomorrow at 8:30 PM IST / 9 AM Central time. Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit#).

PR #9317, opened 2023-11-07
Enable status col in connection and add ping check

sudhanshutech:
**Notes for Reviewers**

This PR fixes #
Due to some ui crash the status col in connection was disabled its now being enabled by fixing the bug.
This pr also add ping check on kubernetes connection on clicking the chip

### DIFF
--- a/ui/components/connections/index.js
+++ b/ui/components/connections/index.js
@@ -127,8 +127,6 @@ function ConnectionManagementPage(props) {
           handleSubmit={handleCreateConnectionSubmit}
           title="Connect Helm Repository"
           submitBtnText="Connect"
-          // leftHeaderIcon={ }
-          // submitBtnIcon={<PublishIcon  className={classes.addIcon} data-cy="import-button"/>}
         />
       )}
     </>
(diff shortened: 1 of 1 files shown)


sudhanshutech:
@senali-d will you check the theme being not working on different connection status, i have fixed the bug

theBeginner86 (commented):
Fixed styling:

Screenshot 2023-11-07 at 5 26 43 PM

theBeginner86 (approved):
Thank you @sudhanshutech 

LGTM

ui/components/connections/index.js
```diff
@@ -150,7 +158,77 @@ function Connections({ classes, updateProgress, /*onOpenCreateConnectionModal,*/
   const { notify } = useNotification();
   const StyleClass = useStyles();
 
-  // const statuses = ['ignored', 'connected', 'REGISTERED', 'discovered', 'deleted'];
+  const statuses = ['ignored', 'connected', 'REGISTERED', 'discovered', 'deleted'];
```
MUzairS15, line 161: Use the statuses defined in constants.js

MUzairS15 (changes requested):

ui/components/connections/index.js
```diff
@@ -600,6 +684,19 @@ function Connections({ classes, updateProgress, /*onOpenCreateConnectionModal,*/
     }
   };
 
+  const handleKubernetesClick = (name, connectionID) => {
```
MUzairS15, line 687: Why there is need to redefine if you can use from ConnectionWizard or use `useKubernetes` hook (inside ui/hooks dir) (this will be ideal).

MUzairS15 (changes requested):

sudhanshutech:
oh okay, i have to look to the recent created things

ui/components/connections/index.js
```diff
@@ -150,7 +158,77 @@ function Connections({ classes, updateProgress, /*onOpenCreateConnectionModal,*/
...
+            <Chip
+              className={classNames(classes.statusChip, classes.discovered)}
+              value={value}
+              avatar={<ExploreIcon />}
+              label={value}
+            />
+          </MenuItem>
+        );
+    }
+  };
```
aabidsofi19, line 231: @sudhanshutech  if you would create a single status component that would map over STATUS enum . it will be more maintainable

aabidsofi19 (changes requested):

ui/components/connections/index.js
theBeginner86: switched to enum.js file 👍

ui/components/connections/index.js
theBeginner86: Switched to the new k8s hook 👍

ui/components/connections/index.js
theBeginner86: updated to map on states defined in enum 👍

theBeginner86:
> @sudhanshutech Please use this
> 
> https://github.com/meshery/meshery/blob/2c995673697ae24fed1de43b341f5929155d292e/ui/components/connections/ConnectionChip.js#L192
> 
> .
> adding logic in the same file would lead to trouble maintaining it. Also one comp should be responsible for one thing only.
> 
> The API used in handleStatusChange is incorrect. Use the one recently created. Refer docs if you find trouble understanding or any issue. We can chat.

switched to new API for status update 👍

theBeginner86 (approved):

PR #9383, opened 2023-11-17
[Connections] Connection State Management

MUzairS15:
**Notes for Reviewers**

This PR fixes #


leecalcote:
@MUzairS15 merge conflict.

Is this PR dependent upon the new k8s connection definition PR?

leecalcote:
Some extra logs that we might want to cut down on:

```
INFO[2023-11-17T14:18:06-06:00] entering event loop                           app=meshery
INFO[2023-11-17T14:18:06-06:00] inside getNextState: discoverytrue            app=meshery
INFO[2023-11-17T14:18:06-06:00] next state: discovered                        app=meshery
INFO[2023-11-17T14:18:06-06:00] transitioning to next state: discovered       app=meshery
INFO[2023-11-17T14:18:06-06:00] entry action executed, event emitted noop     app=meshery
```

leecalcote:
Each time that I refresh my browser, I get four of these:

Screenshot 2023-11-17 at 2 20 46 PM

MUzairS15:
removed logs.

MUzairS15:
It indicates you don't have any connected cluster. Server is trying to import local kubeconfig and trying to connect.
Previously, user's were not notified about this behaviour.

The concern with having repitive notifications, this is happening because when the browser is refreshed several initialisation requests are made from the UI, causing the same logic to get invoked.

MUzairS15:
Merging by EOD.

leecalcote:
> It indicates you don't have any connected cluster. Server is trying to import local kubeconfig and trying to connect. Previously, user's were not notified about this behaviour.
> 
> The concern with having repitive notifications, this is happening because when the browser is refreshed several initialisation requests are made from the UI, causing the same logic to get invoked.

This all makes sense (no news here). 

What I’m pointing out is that this needs to be taken care of. With this PR merged, is there a new issue tracking this bug regarding duplicative notifications?

PR #9436, opened 2023-11-23
Added feature: `mesheryctl exp model search [query-text]` command

zakisk:
**Notes for Reviewers**

This PR fixes #

mesh

### DIFF
--- a/mesheryctl/internal/cli/root/system/model.go
+++ b/mesheryctl/internal/cli/root/system/model.go
@@ -279,7 +279,7 @@ mesheryctl exp model search [query-text]
 	Args: func(_ *cobra.Command, args []string) error {
 		const errMsg = "Usage: mesheryctl exp model search [query-text]\nRun 'mesheryctl exp model search --help' to see detailed help message"
 		if len(args) == 0 {
-			return fmt.Errorf("query text must not be empty\n\n%v", errMsg)
+			return fmt.Errorf("Search term is missing\n\n%v", errMsg)
 		}
 		return nil
 	},


mesheryctl/internal/cli/root/system/model.go
```diff
@@ -246,24 +246,119 @@ mesheryctl system model view [model-name]
...
+		err = ctx.ValidateVersion()
+		if err != nil {
+			return err
+		}
+		return nil
+	},
+	Args: func(_ *cobra.Command, args []string) error {
+		const errMsg = "Usage: mesheryctl exp model search [query-text]\nRun 'mesheryctl exp model search --help' to see detailed help message"
+		if len(args) == 0 {
+			return fmt.Errorf("query text must not be empty\n\n%v", errMsg)
```
leecalcote, line 282:
```suggestion
			return fmt.Errorf("Search term is missing\n\n%v", errMsg)
```

mesheryctl/internal/cli/root/system/model.go
```diff
@@ -246,24 +246,119 @@ mesheryctl system model view [model-name]
...
+		}
+		return nil
+	},
+	RunE: func(cmd *cobra.Command, args []string) error {
+		mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
+		if err != nil {
+			log.Fatalln(err, "error processing config")
```
leecalcote, line 289: Are we able to use a MeshKit error here?

Either way, please enhance this error to be more palatable to users. What action should they take? This error should be standard across each command. Will you confirm? Let's make it consistent.

mesheryctl/internal/cli/root/system/model.go
```diff
@@ -246,24 +246,119 @@ mesheryctl system model view [model-name]
...
+
+		for _, model := range modelsResponse.Models {
+			if len(model.DisplayName) > 0 {
+				rows = append(rows, []string{model.Category.Name, model.Name, model.Version})
+			}
+		}
+
+		if len(rows) == 0 {
+			// if no model is found
+			utils.Log.Info("No model(s) found")
```
leecalcote, line 335: What about components? Are we not offering search support for components, too?

mesheryctl/internal/cli/root/system/model.go
```diff
@@ -246,24 +246,119 @@ mesheryctl system model view [model-name]
 	},
 }
 
-// ModelCmd represents the `mesheryctl system model` command
+// represents the `mesheryctl exp model search [query-text]` subcommand.
+var searchModelCmd = &cobra.Command{
+	Use:   "search",
+	Short: "search models",
```
leecalcote, line 252: The needs to support Components, too.

leecalcote (changes requested):
Thanks, @zakisk there are a few items to refine.

If the user is searching for model, let's have that column leftmost column. If they are searching for a component, then let's have that column be the leftmost column.

mesheryctl/internal/cli/root/system/model.go
zakisk: @leecalcote  I am calling this API [endpoint](https://github.com/meshery/meshery/blob/0dd426525ab7dda847dd02b5c0b1bbf4721c6127/server/router/server.go#L187) to search models and this endpoint utilizes [GetModels](https://github.com/meshery/meshkit/blob/0133432a3ce871fdc12b148745788d3813a5afe5/models/meshmodel/registry/registry.go#L290) meshkit function. this function doesn't search for components (what I learnt by reading codebase).

zakisk: if we have to increase capibility of search command to search for components as well then I will provide a new flag for components e.g. `--components [query-text] --components [query-text]`, sounds good?

mesheryctl/internal/cli/root/system/model.go
zakisk: @leecalcote `GetMesheryCtl` function [returns](https://github.com/meshery/meshery/blob/0dd426525ab7dda847dd02b5c0b1bbf4721c6127/mesheryctl/internal/cli/root/config/config.go#L70) meshkit error.

zakisk:
> Thanks, @zakisk there are a few items to refine.
> 
> If the user is searching for model, let's have that column leftmost column. If they are searching for a component, then let's have that column be the leftmost column.

this can be acheived after components search.

PR #9445, opened 2023-11-24
Add Support form to meshery

sudhanshutech:
**Notes for Reviewers**

This PR adds support form to meshery when we get any errors .
The files had to be refactored a little bit which i will be doing but prior to that I am getting` CORS` error when i am making post req to webhook of make.com

It will look like
image


sudhanshutech:
This the cors error:
image

leecalcote (changes requested):
This is to go into meshery extensions, not meshery/meshery

PR #9458, opened 2023-11-26
[Feature] Interactive pagination for `model exp model list` command

zakisk:
https://github.com/meshery/meshery/assets/49492007/fe79de59-9590-4ff4-b77a-e572a08237a8

**Notes for Reviewers**


mesheryctl/internal/cli/root/system/model.go
```diff
@@ -128,10 +129,40 @@ mesheryctl system model list --page 2
...
+		}
+
+		if cmd.Flags().Changed("page") {
 			utils.PrintToTable(header, rows)
+		} else {
+			paginator := utils.NewPaginator(header, rows)
+			paginator.Render()
+			keysEvents, err := keyboard.GetKeys(10)
+			if err != nil {
+				panic(err)
```
leecalcote, line 143: Please use MeshKit-style error handling, @zakisk

mesheryctl/pkg/utils/formatter.go
```diff
@@ -36,3 +41,62 @@ func SetupMeshkitLogger(debugLevel bool, output io.Writer) {
...
+	whiteBgPrinter       *color.Color
+	userMsg             string
+}
+
+func NewPaginator(header []string, rows [][]string) *Paginator {
+	var buf bytes.Buffer
+	renderedTableString := writeTableToBuffer(&buf, header, rows)
+	lastVisibleItemIndex := 26
+	data := strings.Split(renderedTableString, "\n")
+	userMsg := "Press ↓ to load more items, esc and Ctrl+C to quit"
```
leecalcote, line 57: Please change to spacebar as both the keystroke and the help description.
```suggestion
	userMsg := "Press spacebar to load more. Press ESC or Ctrl+C to stop."
```

mesheryctl/pkg/utils/helpers.go
```diff
@@ -537,6 +538,15 @@ func PrintToTableWithFooter(header []string, data [][]string, footer []string) {
 	table.Render() // Render the table
 }
 
+// ClearLine clear last line from output
+func ClearLine() {
+	if runtime.GOOS == "windows" {
+		fmt.Print("\r")
+	} else {
+		fmt.Print("\r\033[K")
```
leecalcote, line 546: @zakisk please provide reference that lists these ASCII characters as best practice for their respective OS'es. Include the reference as a comment.

leecalcote (changes requested):
Good stuff, @zakisk. I have a couple of points of feedback for you to consider.

mesheryctl/internal/cli/root/system/model.go
```diff
@@ -128,10 +129,40 @@ mesheryctl system model list --page 2
...
+			}
+			defer func() {
+				_ = keyboard.Close()
+			}()
+
+			for {
+				event := <-keysEvents
```
MUzairS15, line 150: @zakisk Do we have this paginated navigation for other commands eg: Perf profiles, designs and other resources?
I don't think so.

Please extract this paginated navigation logic and make it reusable across commands.

mesheryctl/internal/cli/root/system/model.go
MUzairS15: No need to panic @zakisk 😉

mesheryctl/internal/cli/root/system/model.go
zakisk: @MUzairS15 pagination login is abstracted into Pagination type.

zakisk:
@MUzairS15 yes, panic isn't ideal here, I am thinking that if there is any error while listening keyboard events, the whole table should be printed to output (without pagination) or help message shown to user for other pagination option using `--page` flag, what do you and @leecalcote suggest me?

leecalcote:
@zakisk, yes: I would consider invoking the same behavior as if someone had pressed CTRL+C, but perhaps, with an error message to let them know that something is awry.

MUzairS15:
@zakisk Merge conflicts

MUzairS15:
@zakisk Can you add a command for the components list as well in separate PR

mesheryctl/pkg/utils/formatter.go
leecalcote: @zakisk in the future, please accept these suggestions so that I don't have to go separately confirm that you incorporated the feedback.

mesheryctl/internal/cli/root/system/model.go
leecalcote: @zakisk still working on this?

leecalcote:
@alphaX86

leecalcote:
@RipulHandoo do you mind carrying this forward?

RipulHandoo:
Yes Sir.

PR #9538, opened 2023-12-01
[connection] - Fix issue loading connection icon and status from connection definition

senali-d:
**Notes for Reviewers**

This PR fixes #

- Connection kind name and component definition not same was reason for not load the image and also connection transition status. I'm handle those from ui

connection definition: `meshery-core`
connection - connection kind: `meshery`

Screenshot 2023-12-02 at 12 15 31 AM

### DIFF
--- a/ui/utils/Enum.js
+++ b/ui/utils/Enum.js
@@ -55,6 +55,6 @@ export const MesheryPatternsCatalog = 'meshery-patterns-catalog';
 export const MesheryFiltersCatalog = 'meshery-filters-catalog';
 
 export const CONNECTION_KINDS = {
-  MESHERY: 'meshery-core',
+  MESHERY: 'meshery',
   KUBERNETES: 'kubernetes',
 };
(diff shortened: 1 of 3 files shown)


senali-d:
// @MUzairS15

ui/utils/Enum.js
```diff
@@ -55,6 +55,6 @@ export const MesheryPatternsCatalog = 'meshery-patterns-catalog';
 export const MesheryFiltersCatalog = 'meshery-filters-catalog';
 
 export const CONNECTION_KINDS = {
-  MESHERY: 'meshery-core',
```
MUzairS15, line 58: @senali-d you are changing to meshery but above you are adding conditional logic to change to meshery-core. The real issue I think is because at https://github.com/meshery/meshery/pull/9538/files#diff-b14dd66531b84fdbb46ba6c386362eb972e04bf8d3abe3b6f26e95a6c16fa493L236
splitting logic. Makes sense.

thanks for handling  the issue though.

PR #9541, opened 2023-12-01
fix: add pagination while searching for models

### LINKED ISSUE #9494: [UI] Infinite API calls when trying to search in the Registry UI.
https://github.com/meshery/meshery/assets/7570704/66ebbba0-d6fe-4ca2-96f4-4a3c11a79bc2

[Slack Message](https://layer5io.slack.com/archives/C023QGMSH8D/p1701281565704769?thread_ts=1701281353.886169&cid=C023QGMSH8D)

VihasMakwana:
**Notes for Reviewers**

Add pagination while searching. 
Also, add custom logic to maintain consistency across searching different terms.

If new search is added
- Perform API call (14 records at a time), turn on `firstSearch` flag.

If the old term changes
- Reset pages, clear the state, and perform a search.

If the old term remains as it is and we scroll down
- Search for the next page and append entries to the previous state.

This PR fixes #9494


leecalcote (changes requested):
Thanks, @VihasMakwana. Is this proposed as a temp solution until backend support is offered?

VihasMakwana:
@leecalcote I think there has been some misunderstanding.

I'm *not* using frontend to filter results.
I'm leveraging on paging, just like all other dashboard components.
Using pagination reduces the latency and keeps all the filtering in the backend.
Perhaps, you missed my [comment](https://github.com/meshery/meshery/issues/9494#issuecomment-1836802077)?

VihasMakwana:
@leecalcote In this case, the backend is working as expected.
In the previous scenario, we used to request all the data, and that caused issues.
Now, I request data in chunks, just like other components do.
This solves:
- Latency problem
- infinite API illusion

MUzairS15 (commented):
@VihasMakwana why you are preserving the previous search results?
If the search term changes completely then results will be inconsistent.

also where are you making the check if search term is same use previous results ?

VihasMakwana:
> @VihasMakwana why you are preserving the previous search results? If the search term changes completely then results will be inconsistent.
> 
> also where are you making the check if search term is same use previous results ?

https://github.com/meshery/meshery/pull/9541/files#r1412855140

VihasMakwana:
> Why you are preserving the previous search results?

Well, I'm not. 
If you observe the check at https://github.com/meshery/meshery/pull/9541/files#diff-7af477cff69a5e9262f2993a52d081aae3f259f200e7747a33dbbdf6e85349bbR183-R187, you'll see that we only preserve state if the term is same (based on `firstSearch` check).
`firstSearch` is set to `true` if the term is changed, or if the view is changed. 
And we'll know based on `firstCheck` that the term and view are the same as before, so we can combine current results with the previous state.

I'll upload a recording on the issue shortly.

VihasMakwana:
closed in favour of https://github.com/meshery/meshery/pull/9573

PR #9560, opened 2023-12-04
chore: partial cleanup of info icons

leecalcote:


leecalcote:
@SudhanAruna the material UI icon still needs to be replaced.

PR #9573, opened 2023-12-04
[feat] add enhancements, fix latency while searching models

VihasMakwana:
This PR fixes https://github.com/meshery/meshery/issues/9494

Depends on https://github.com/meshery/meshkit/pull/428 to get merged.


server/handlers/component_handler.go
```diff
@@ -219,6 +219,9 @@ func (h *Handler) GetMeshmodelModels(rw http.ResponseWriter, r *http.Request) {
 		OrderOn:     queryParams.Get("order"),
 		Sort:        queryParams.Get("sort"),
 		Annotations: returnAnnotationComp,
+
+		Components:    queryParams.Get("components") == "true",
+		Relationships: queryParams.Get("relationships") == "true",
```
leecalcote, line 224: a couple of lint checks to look into, @VihasMakwana

VihasMakwana: yup. will push in a moment

VihasMakwana: @leecalcote these are due to undefined references in meshkit.
Once I get https://github.com/meshery/meshkit/pull/428 merged, all these lints will be green.
Let's keep this PR open until we merge the former.

leecalcote: Oh, gotcha. That makes sense.

MUzairS15:
The new Meshkit release is updated in the master.

MUzairS15 (commented):
@VihasMakwana All looks good.
Just one more thing, will you ensure the relationships and components are fetched only when the model is expanded?

VihasMakwana:
> @VihasMakwana All looks good. One more thing, will you ensure the relationships and components are fetched only when the model is expanded?

Hmm, that will be a rework. Can we merge this as it is right now, as current behavior only pulls components and relationships when a search is executed?
We can enhance this behavior later.

I'm asking because the UI expects components and relationships to be pre-loaded inside models.

MUzairS15:
> > @VihasMakwana All looks good. One more thing, will you ensure the relationships and components are fetched only when the model is expanded?
> 
> Hmm, that will be a rework. Can we merge this as it is right now, as current behavior only pulls components and relationships when a search is executed? We can enhance this behavior later.
> 
> I'm asking because the UI expects components and relationships to be pre-loaded inside models.

Since you already have a flag that controls whether to fetch relationships and components IMO it should be easy.

yes we can track in different PR.

MUzairS15 (approved):

MUzairS15:
Thankyou @VihasMakwana

PR #9608, opened 2023-12-09
[server][fix]: don't exit in case of context is persisted

VihasMakwana:
**Notes for Reviewers**
While playing around locally with Minikube, I discovered that my k8s resources weren't getting saved in DB. I found the offending code [here](https://github.com/meshery/meshery/blob/master/server/machines/kubernetes/discover.go#L49-L52). The k8s context is already saved before the state machine initialization [here](https://github.com/meshery/meshery/blob/master/server/handlers/middlewares.go#L235-L241).
This causes the former `SaveK8sContext` to fail with `kubernetes context already persisted`.
This error shouldn't be treated as something unexpected, as it is possible to encounter this for the first time on localhost:9081/

It however causes the following behaviour:
Screenshot_20231209_220213
My minikube nodes, namespaces, etc. show empty.

After the fix:
Screenshot_20231209_210416

This PR fixes #

### DIFF
--- a/install/docker-extension/ui/package-lock.json
+++ b/install/docker-extension/ui/package-lock.json
@@ -21,7 +21,7 @@
         "@testing-library/jest-dom": "^5.16.2",
         "@testing-library/react": "^12.1.4",
         "@testing-library/user-event": "^13.5.0",
-        "axios": "^0.26.1",
+        "axios": "^1.6.0",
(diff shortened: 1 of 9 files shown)


MUzairS15 (approved):
Thanks

leecalcote:
This is great 😊

PR #9609, opened 2023-12-09
[server][chore]: drop dead code

VihasMakwana:
**Notes for Reviewers**
There are some unused functions lurking around in the code base.
Their references were removed somewhere between v0.6.90 and v0.6.91 release https://github.com/meshery/meshery/compare/v0.6.90...v0.6.91

This PR fixes #


leecalcote (commented):
Thanks for cleaning up, @VihasMakwana.

I wonder if we are removing an API that we intend to support. Is this graphql api for resetting the database not exposed or not connected to a resolver?

is it only the REST API for resetting the database that is functional? 


If they both have the same behavior, even though the Meshery UI as a client may not be using this graph API, we may want to leave it available for other client’s use. 


I believe that mesheryctl is only using the REST API, so I don’t think it’s of concern.  

As an extensible platform, we tout the fact that we support both choice of REST and GraphQL for system to integrate with Meshery.

server/internal/graphql/model/helpers.go
leecalcote, line 1: Has the code Meshery Operator status only recently gone unused, do you know? If so, this would make sense in context of changes that @MUzairS15  has made, so your cleanup could be quite timely in that sense. 👍

VihasMakwana:
> Thanks for cleaning up, @VihasMakwana.
> 
> I wonder if we are removing an API that we intend to support. Is this graphql api for resetting the database not exposed or not connected to a resolver?
> 
> is it only the REST API for resetting the database that is functional?

Hard reset is done via graphql. The API I removed was to persist the data received by meshsync. And it is not connected to any resolver.
The persisting is done via https://github.com/meshery/meshery/blob/master/server/models/meshsync_events.go#L54-L77. 

> If they both have the same behavior, even though the Meshery UI as a client may not be using this graph API, we may want to leave it available for other client’s use.
> 
> I believe that mesheryctl is only using the REST API, so I don’t think it’s of concern.
> 
> As an extensible platform, we tout the fact that we support both choice of REST and GraphQL for system to integrate with Meshery.

I see. Makes sense.

server/internal/graphql/model/helpers.go
VihasMakwana: > Has the code Meshery Operator status only recently gone unused
I believe so. @MUzairS15 can you confirm?

abhijeetgauravm:
Hey @VihasMakwana  Let's discuss this on Meshery Development Meeting today at 8:30 PM IST / 9 AM Central time. Please add this as an agenda item in the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit#).

MUzairS15 (approved):

leecalcote:
Excellent 🙂

Issue #9655, opened 2023-12-18
[UI] Issue with routing of Connections page


leecalcote:
### Current Behavior
Visit http://localhost:9081/management/connections, click refresh. 

### Screenshots/Logs
<img width="325" alt="Screenshot 2023-12-18 at 1 45 52 PM" src="https://github.com/meshery/meshery/assets/7570704/162a562a-a862-4ef4-af73-d6ef0eec94ea">

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](http://discuss.meshery.io) and [Community Slack](https://slack.meshery.io)

PR #9673, opened 2023-12-19
Fix Lifecycle routing

aabidsofi19:
**Notes for Reviewers**

This PR fixes #
https://github.com/meshery/meshery/issues/9655


Makefile
```diff
@@ -278,6 +278,10 @@ ui-provider-build:
 ui-integration-tests: ui-setup
 	cd ui; npm run ci-test-integration; cd ..
 
+## rtk query code generation
+rtk-query-codegen:
```
leecalcote, line 282: Will you put this in a different PR? This needs to be added to the list of dependency checks and we need to make sure that we’re actually publishing this package.

aabidsofi19: got accidentaly pushed .

leecalcote (approved):

PR #9684, opened 2023-12-19
[bug] Fix Unnecessary re-rendering of Dashboard 

theBeginner86:
**Notes for Reviewers**

It was occurring due to unwanted props being passed to dashboard component. Since these redux stats where being updated as a result of side-effects of other components, hence the dashboard component was re-rendering due to change in state of these inbound props.

This PR removes the unnecessary props and redux state vars.

This PR fixes #


sudhanshutech:
Thanks @theBeginner86 , i was unaware of this section of props so couldn't get to the cause

PR #9702, opened 2023-12-22
Added logic of HTML entity encoding in RJSF Text fields widget

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

This PR adds the logic of encoding HTML entities inside RJSF field itself. In its default configuration, the RJSF fields do not encode HTML entities. However, adding `x-encode-to-uri` property in schema, will allow HTML entity encoding on a per-field basis. Setting the x-encode-in-uri property to true enables the automatic encoding of HTML entities for the specified field.

Data without property `x-encode-in-uri` 
```
Okay so this is test publish

Added space.  with  : speacial character

fdsa34324 432432614 " "

fsdf
```

Data with `x-encode-in-uri` as true
```
Okay%20so%20this%20is%20test%20publish%0A%0AAdded%20space.%20%20with%20%20%3A%20speacial%20character%0A%0Afdsa34324%20432432614%20%22%20%22%0A%0Afsdf
```

### DIFF
--- a/ui/components/MesheryMeshInterface/PatternService/RJSFCustomComponents/CustomSelectWidget.js
+++ b/ui/components/MesheryMeshInterface/PatternService/RJSFCustomComponents/CustomSelectWidget.js
@@ -49,7 +49,6 @@ export default function CustomSelectWidget({
   const xRjsfGridArea = schema?.['x-rjsf-grid-area']; // check if the field is used in different modal (e.g. publish)
 
   multiple = typeof multiple === 'undefined' ? false : !!multiple;
-  const isRequired = props?.required;
   const emptyValue = multiple ? [] : '';
   const isEmpty =
     typeof value === 'undefined' ||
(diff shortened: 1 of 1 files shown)


leecalcote:
The catalogs and the info and publish modals decode this as expected?

Yashsharma1911:
> The catalogs and the info and publish modals decode this as expected?

Yes they does, we need to make small changes in the schema which I already did here https://github.com/meshery/meshkit/pull/436

leecalcote:
> > The catalogs and the info and publish modals decode this as expected?
> 
> Yes they does, we need to make small changes in the schema which I already did here [meshery/meshkit#436](https://github.com/meshery/meshkit/pull/436)

In this response, you've confirmed that both Catalogs are compatible, too.

Yashsharma1911:
> > > The catalogs and the info and publish modals decode this as expected?
> > 
> > 
> > Yes they does, we need to make small changes in the schema which I already did here [meshery/meshkit#436](https://github.com/meshery/meshkit/pull/436)
> 
> In this response, you've confirmed that both Catalogs are compatible, too.

Write now not and this merge will not affect them, until we merge the schema PR, so I'll make changes their too, first pushing changes for meshery.io catalog

PR #9647, opened 2023-12-16
[Environments] Environments page

senali-d:
**Notes for Reviewers**

This PR fixes #

- [x] CRUD for environments
- [x] Allocate connection for environment

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #9705, opened 2023-12-23
[Environments] - Dark mode support and RTK migration

senali-d:
**Notes for Reviewers**

This PR fixes #

- Dark mode support for environments
- Migrate api to RTK
- Scroll loading for transfer list component
- Fix caching issues

Screenshot 2023-12-25 at 6 58 58 PM

Screenshot 2023-12-25 at 6 59 21 PM

### DIFF
--- a/ui/components/Environments/index.js
+++ b/ui/components/Environments/index.js
@@ -159,18 +159,18 @@ const Environments = ({ organization, classes }) => {
   // const loading = isEnvironmentsLoading || isEnvironmentConnectionsLoading || isConnectionsLoading;
 
   const environments = environmentsData?.environments ? environmentsData.environments : [];
-  const connectionsDataRtk = connections?.connections ? connections.connections : [];
-  const environmentConnectionsDataRtk = environmentConnections?.connections
-    ? environmentConnections.connections
-    : [];
+  // const connectionsDataRtk = connections?.connections ? connections.connections : [];
+  // const environmentConnectionsDataRtk = environmentConnections?.connections
+  //   ? environmentConnections.connections
+  //   : [];
 
-  useEffect(() => {
-    setConnectionsData((prevData) => [...prevData, ...connectionsDataRtk]);
-  }, [connections]);
+  // useEffect(() => {
+  //   setConnectionsData((prevData) => [...prevData, ...connectionsDataRtk]);
+  // }, [connections]);
 
-  useEffect(() => {
-    setEnvironmentConnectionsData((prevData) => [...prevData, ...environmentConnectionsDataRtk]);
-  }, [environmentConnections]);
+  // useEffect(() => {
+  //   setEnvironmentConnectionsData((prevData) => [...prevData, ...environmentConnectionsDataRtk]);
+  // }, [environmentConnections]);
 
   useEffect(() => {
     if (isEnvironmentsError) {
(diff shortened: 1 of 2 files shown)


leecalcote (commented):
The existing light mode coloring can be left as is, I think. The dark mode only needs the white card to be changed to a shade of gray, I believe. 

@Rexford74, we have pre-established colors in-use in Meshery UI for similar cards in dark mode. I assume that these lineup with the proposed colors in Sistent. Is that about right?

ritiksaxena124:
@senali-d we already have a mockup for the dark mode of environments card in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI?type=design&node-id=24380%3A25009&mode=design&t=VPJrl1doCmoZC5EO-1). I mentioned you on that you might have missed it previously.

Rexford74:
> The existing light mode coloring can be left as is, I think. The dark mode only needs the white card to be changed to a shade of gray, I believe.
> 
> @Rexford74, we have pre-established colors in-use in Meshery UI for similar cards in dark mode. I assume that these lineup with the proposed colors in Sistent. Is that about right?

Well, so far, just Layer5's colors have been defined and none of the colors for the different products: Meshery, Nighthawk, etc.
This is not an issue however as any colors for individual products of Layer5 can be added both now and in the future as the need arises. If it is possible, please I will like a confirmation of what the primary color for Meshery is; Charcoal or Teal Blue. This will help to provide an accurate tonal palette and incorporate the addition to the designed Sisten drafts.

ritiksaxena124:
> > The existing light mode coloring can be left as is, I think. The dark mode only needs the white card to be changed to a shade of gray, I believe.
> > @Rexford74, we have pre-established colors in-use in Meshery UI for similar cards in dark mode. I assume that these lineup with the proposed colors in Sistent. Is that about right?
> 
> Well, so far, just Layer5's colors have been defined and none of the colors for the different products: Meshery, Nighthawk, etc. This is not an issue however as any colors for individual products of Layer5 can be added both now and in the future as the need arises. If it is possible, please I will like a confirmation of what the primary color for Meshery is; Charcoal or Teal Blue. This will help to provide an accurate tonal palette and incorporate the addition to the designed Sisten drafts.

Teal is not a primary color @Rexford74

ui/themes/app.js
```diff
@@ -117,7 +117,7 @@ export var darkTheme = createTheme({
       disableButtonBg: '#666666',
       disableButton: '#9d9d9d',
       listItemHover: 'rgb(0, 187, 166, 0.5)',
-      white: 'rgb(255, 255, 255, 0.1)',
+      white: '#202020',
```
Yashsharma1911, line 120: // FYI @ritiksaxena124

Yashsharma1911: @senali-d have you tested this theme change in MeshMap too?

senali-d: This color define specifically for environments in my previous PR. I think this was not use anywhere other than this page.

senali-d: #9647

senali-d:
Suggestion for APIs: 

1. Allocated connections count return from the get environments API: `api/environments`
2. Bulk allocation of connections API

- When we go ahead with the scroll loading we need a separate api for get the assignable connections to a specific environment.

// @theBeginner86 @aabidsofi19

ui/themes/app.js
Yashsharma1911: Ohh make sense thanks for confirming

senali-d:
I have temporary disable scroll loading of transfer list items (connections) to move forward...

PR #9722, opened 2023-12-26
[Workspaces] - Workspaces page

senali-d:
**Notes for Reviewers**

This PR fixes #

- [x] CRUD for workspaces
- [x] Assign environments
- [x] Assign design 
- [x] Implement scroll loading for transfer list

### DIFF
--- a/ui/components/Lifecycle/Workspaces/workspace-card.js
+++ b/ui/components/Lifecycle/Workspaces/workspace-card.js
@@ -66,6 +66,7 @@ export const RedirectButton = ({ title, count, disabled = true, classes }) => {
  * @param {Function} props.onEdit - Function to edit the workspace.
  * @param {Function} props.onSelect - Function to select workspace for bulk actions.
  * @param {Array} props.selectedWorkspaces - Selected workspace list for delete.
+ * @param {Function} props.onAssignEnvironment - Function to open environment assignment modal open.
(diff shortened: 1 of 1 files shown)


ui/components/Workspaces/workspace-card.js
```diff
@@ -0,0 +1,306 @@
...
+                </DateLabel>
+              </Grid>
+            </Grid>
+          </CardWrapper>
+        }
+      />
+    </>
+  );
+};
+
```
Yashsharma1911, line 305: @senali-d can we split this code here for the back and front components separately within same file? If we could it would be much readable

ui/assets/icons/Workspace.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+
+const WorkspaceIcon = ({ width, height, fill, style = {}, secondaryFill }) => (
```
Yashsharma1911, line 3: Adding default prop value will be helpful
```suggestion
const WorkspaceIcon = ({ width = 24, height = 24, fill, style = {}, secondaryFill }) => (
```

senali-d: I'm going to use icons from sistent. So those icons are temporary added. I'll replace them.

senali-d:
Moving forward this PR. Remaining items will implement in a new pr.

PR #9783, opened 2024-01-01
[bug] Fix Relationships Registration Process 

theBeginner86:
**Notes for Reviewers**

Move from buffered to unbuffered chan for relationship and component.

Reason:
Earlier, we were using buffered chan of size 1. This means that during registration if the size of the buffer is reached its max limit of 1 then a particular entity will not be processed.
This could be fixed if we increase the size of the buffered chan to a larger value. But considering we would be adding more and more of relationship/components in future so we might again hit the buffer size. Make it dynamic based on the number of components/relationships to register might fix this but could be fragile.
Or we can use an unbuffered chan that makes the process synchronous so that the receiving/sending chan waits until each of them completes their tasks.

This PR fixes #

### DIFF
--- a/server/meshmodel/helper.go
+++ b/server/meshmodel/helper.go
@@ -149,7 +149,6 @@ func (erh *EntityRegistrationHelper) generateRelationships(pathToComponents stri
 				erh.errorChan <- errors.Wrapf(err, fmt.Sprintf("unmarshal json failed for %s", path))
 				return nil
 			}
-			fmt.Printf("rel: %+v\n", rel)
 			erh.relationshipChan <- rel
 		}
 		return nil


MUzairS15 (commented):
@theBeginner86 When the buffer gets full, further send to the channel will be stalled, no more entity will be processed unless one of the consumer takes the entity from the buffer.
But in no sense it will be lost. 

Making the process synchronus doesn't sound right unless we there's some other issue which mandates synchronisation bw some process.

MUzairS15:
What is the issue happening otherwise?

leecalcote:
> What is the issue happening otherwise?

@MUzairS15 open Slack

MUzairS15 (approved):

PR #9807, opened 2024-01-01
[OCI] Add basic support for Designs To OCI

theBeginner86:
**Notes for Reviewers**

Screenshot 2024-01-02 at 5 26 11 AM

This PR fixes #


server/models/meshery_pattern.go
```diff
@@ -43,6 +47,7 @@ const (
 	DockerCompose DesignType = "Docker Compose"
 	K8sManifest   DesignType = "Kubernetes Manifest"
 	Design        DesignType = "Design"
+	OCI           DesignType = "OCI"
```
leecalcote, line 50: @theBeginner86 this doesn't exactly compute given that "Design" and "OCI" are both assumed to be of the design format - just that one is directly yaml and the other needs to be unwrapped. Can you collapse these two and include logic that simply attempts to import either format (straight yaml or tarred yaml (in OCI))?

server/helpers/oci/oci.go
```diff
@@ -0,0 +1,30 @@
+package oci
+
```
leecalcote, line 2: Relocate this file under `server/models/designs`?

server/models/meshery_pattern.go
leecalcote: You can reuse whatever test oras does for oci and an OpenAPI validation against Meshery's Design schema as format type tests. //@MUzairS15

server/helpers/oci/oci.go
leecalcote: Given a conversation about broader OCI support for Models, too, no its best to centralize OCI encode/decode.

leecalcote:
1. mesheryctl pattern [ import | export ]
2. mesheryctl model [ import | export ]

leecalcote:
Successful prototype. This is great. Complete refactoring needed in order to be based on flux.

PR #9822, opened 2024-01-03
Prevent common modal from rerender

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

I observed a re-render issue in the MeshMap publish modal when someone joined my design during the publishing process. Despite I made efforts to optimize the modals in MeshMap, the re-rendering persisted because of [`ComponentDrawerTabContent`](https://github.com/layer5labs/meshery-extensions/blob/master/meshmap/src/components/designer/drawer/ComponentDrawerTabContent/index.js) which has prop drilling issues was getting rerendered to many times. We are passing function to those modals to submit data, causing React.memo to consider them as props that had changed, leading to continuous re-renders of the modal components.

However I observed we have other couple more places where we are using RJSF (e.g. cloud) and even on rerender we don't lose focus then I found passing the `RJSFWrapperComponent` prop as default for additional props to the RJSF form was making the component sensitive to re-renders, so now I made it conditional so that if it is not passed then don't even pass default RJSFWrapperComponent prop to RJSF form and make sure RJSF doesn't lose focus even if modal gets rerender.

Note: This change will not affect any behavior of the component

```diff
- RJSFWrapperComponent={RJSFWrapperComponent || RJSFWrapperComponentDefault(uiSchema)}
+ {...(RJSFWrapperComponent && { RJSFWrapperComponent })}
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/models.md
+++ b/.github/ISSUE_TEMPLATE/models.md
@@ -1,12 +1,12 @@
 ---
-name: 🖥 💡 MeshModel Update
-about: An update, addition, or removal of one or more of the models, components, relationships, or policies within MeshModel.
-title: '[MeshModel]: '
-labels: area/meshmodel
+name: 🖥 💡 Models Update
+about: An update, addition, or removal of one or more of the models, components, relationships, workflows, or policies within Meshery Models.
+title: '[Models]: '
+labels: area/models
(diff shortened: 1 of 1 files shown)


Yashsharma1911:
Moving forward here as this change small change

PR #9829, opened 2024-01-03
Registry log

### LINKED ISSUE #9555: A new log file for while boot time
### Desired Behavior
A new log line at boot-time is required regarding quantification of items that are successfully imported into the registry, and also, items not imported.

---

Jougan-0:
**Notes for Reviewers**

This PR fixes #9555 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ replace (
 	github.com/docker/cli => github.com/docker/cli v23.0.6+incompatible
 	github.com/docker/docker => github.com/docker/docker v23.0.6+incompatible
 	github.com/kubernetes/kompose => github.com/meshery/kompose v1.26.2-0.20230425025309-3bb778d54007
+	github.com/layer5io/meshkit v0.7.9 => github.com/Jougan-0/meshkit v0.7.9
 	k8s.io/client-go => k8s.io/client-go v0.28.3
 	k8s.io/kubectl => k8s.io/kubectl v0.28.3
 	oras.land/oras-go v1.2.4 => oras.land/oras-go v1.2.3
--- a/go.sum
+++ b/go.sum
@@ -652,6 +652,8 @@ github.com/DATA-DOG/go-sqlmock v1.4.1/go.mod h1:f/Ixk793poVmq4qj/V1dPUg2JEAKC73Q
 github.com/DATA-DOG/go-sqlmock v1.5.0 h1:Shsta01QNfFxHCfpW6YH2STWB0MudeXXEWMr20OEh60=
 github.com/DATA-DOG/go-sqlmock v1.5.0/go.mod h1:f/Ixk793poVmq4qj/V1dPUg2JEAKC73Q5eFN3EC/SaM=
 github.com/JohnCGriffin/overflow v0.0.0-20211019200055-46fa312c352c/go.mod h1:X0CRv0ky0k6m906ixxpzmDRLvX58TFUKS2eePweuyxk=
+github.com/Jougan-0/meshkit v0.7.9 h1:qaeym8hYUsFBikojXHpRiwVl6eXBgpQohxLv5AKhTEg=
+github.com/Jougan-0/meshkit v0.7.9/go.mod h1:QWzYCLhtjeLIkNk+gL2oAXgSksSYS9mkA+vCHzl8TtM=
 github.com/Knetic/govaluate v3.0.1-0.20171022003610-9aa49832a739+incompatible/go.mod h1:r7JcOSlj0wfOMncg0iLm8Leh48TZaKVeNIfJntJ2wa0=
 github.com/MakeNowJust/heredoc v1.0.0 h1:cXCdzVdstXyiTqTvfqk9SDHpKNjxuom+DOlyEeQ4pzQ=
 github.com/MakeNowJust/heredoc v1.0.0/go.mod h1:mG5amYoWBHf8vpLOuehzbGGw0EHxpZZ6lCpQ4fNJ8LE=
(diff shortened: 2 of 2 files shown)


Jougan-0:
https://github.com/meshery/meshkit/pull/427

Jougan-0:
Here's the video clip of the new features I went the thought that people would register some model and if they wanna see the logs of the new model they could refresh the webpage itself rather than restarting the server and will get the quantity of the entities that are registered successfully or not .
[Screencast from 2024-01-03 21-44-21.webm](https://github.com/meshery/meshery/assets/99207534/20a12c32-a35b-46b4-9bcb-4f7d2cd6ebf3)

Jougan-0:
It's send the event to the ui on the first load of the page as well as when the page is refreshed

server/handlers/component_handler.go
```diff
@@ -1296,7 +1378,7 @@ func (h *Handler) RegisterMeshmodelComponents(rw http.ResponseWriter, r *http.Re
 func (h *Handler) GetMeshmodelRegistrants(rw http.ResponseWriter, r *http.Request) {
 	rw.Header().Add("Content-Type", "application/json")
 	enc := json.NewEncoder(rw)
-
+	registry.Mutex.Lock()
```
Jougan-0, line 1381: @MUzairS15 should i use this to sync the registering and fetching of the data i Don't think it is necessary just a fail safe if someone opens the server if the entites are still being registered can you confirm this 
https://github.com/meshery/meshery/blob/dd67ed6e06ec6070cff8356ca5b3e9b641fa7d58/server/meshmodel/helper.go#L167

leecalcote:
@MUzairS15 is there some existing mechanism that we should be using or do we need this new paradigm of storing and displaying log files?

leecalcote:
Good of you to stick with today's extra meeting, @Jougan-0 👍

Jougan-0:
> Good of you to stick with today's extra meeting, @Jougan-0 👍

It was fascinating to realize that I haven't yet scratched the surface of the project and still have a lot to learn for me to be more comfortable and useful in future discussions.

leecalcote:
@Jougan-0 expected failures? 

```
Error: server/meshmodel/helper.go:167:12: undefined: meshmodel.Mutex
Error: server/meshmodel/helper.go:187:14: undefined: meshmodel.Mutex
Error: server/meshmodel/helper.go:246:29: undefined: meshmodel.FailedMsgCompute
Error: server/meshmodel/helper.go:248:24: undefined: meshmodel.ErrRegisteringEntity
Error: server/meshmodel/helper.go:253:48: undefined: meshmodel.RegisterAttempts
Error: server/meshmodel/helper.go:255:23: undefined: meshmodel.ErrMarshalingRegisteryAttempts
Error: server/meshmodel/helper.go:261:23: undefined: meshmodel.ErrWritingRegisteryAttempts
# github.com/layer5io/meshery/server/handlers
Error: server/handlers/component_handler.go:1342:31: undefined: registry.FailedMsgCompute
Error: server/handlers/component_handler.go:1381:11: undefined: registry.Mutex
Error: server/handlers/component_handler.go:1415:11: undefined: registry.Mutex
make: *** [Makefile:95: build-server] Error 1
Error: Process completed with exit code 2.
```

server/handlers/component_handler.go
MUzairS15: No not required

server/meshmodel/helper.go
```diff
@@ -162,6 +164,7 @@ func (erh *EntityRegistrationHelper) generateRelationships(pathToComponents stri
 // If an error occurs, it logs the error
 func (erh *EntityRegistrationHelper) watchComponents(ctx context.Context) {
 	var err error
+	meshmodel.Mutex.Lock()
```
MUzairS15, line 167: These mutex is for shared variables right used in registryLog function?
If so move this lock smd unlock within that function call.

Jougan-0: @MUzairS15 
but we are locking as the registering is started when it enters the watchComponent for the first time 
the main aim of this mutex is that the function GetmeshmodelRegistrant shouldn't be displaying the value if the registration is not complete  and unlock it when the registration is complete . That is why we are locking it at the start and then unlocking it when the registration is complete .
```https://github.com/meshery/meshery/blob/e2bbe45b50d5acdb9493bab4fbf90b0998b4db36/server/handlers/component_handler.go#L1378```

File: server/handlers/component_handler.go [hunk 3]
```diff
@@ -1275,6 +1278,85 @@ func (h *Handler) RegisterMeshmodelComponents(rw http.ResponseWriter, r *http.Re
...
+// Handle POST request for registering meshmodel components.
+//
+// Validate the given value with the given schema
+// responses:
+// 	200:MeshModelHostsWithEntitySummary
+
+// request body should be json
+// request body should be of Host format
+
+func (handler *Handler) NonRegisterEntity(responseWriter http.ResponseWriter, request *http.Request, _ *models.Preference, user *models.User, provider models.Provider) {
```
MUzairS15, line 1291: Still not according to API design.
@Jougan-0 The functionality that you are defining in the PR is crucial, while the output that we get from the current implementation is nice, the approach is not ideal.

How do you ensure that the registration might have been completed by the time UI makes a request to this endpoint?

My question still stands, if the server is doing the registration why you cannot consolidate all errors and send it all at once?
See the errorChan channel it listens to all the errors encountered, can't you send in more details on this channel `model, component and error` or the `event` object itself, store it in an array, and call the `cancel` function of the passed in context (when registration completes), when you will call the cancel func the ctx.Done case will be called, there you can iterate over that array and add in more details (i.e. summary, x model, y comps failed to register..... )

MUzairS15 (changes requested):

server/handlers/component_handler.go
Jougan-0: [Screencast from 2024-01-09 06-22-53.webm](https://github.com/meshery/meshery/assets/99207534/5573a2e4-69de-411a-a91b-4442ee9069e2)
@MUzairS15 the reason for not using the errorChan is the same I am displaying in the video when the schema is nil the code works fine and these components are not registered but if we return an error such as schema empty the registration breaks I don't know the reason for it but this doesn't make sense i searched but i don't know why we exit quietly when the schema is empty and if we return an error the code breaks so i opt for using the meshkit to store the error and store their in meshkit itself when the entites try to register .

MUzairS15: API design is still not accurate, invoking the event from UI is also inaccurate, find a way to send events from the server.

leecalcote:
@MUzairS15, fyi, @Jougan-0 made a hard commitment to complete this effort by Friday, Jan 19th. If not completed by then, we'll need to toss in the towel here and carry forward these concerns separately under this issue: https://github.com/meshery/meshery/issues/9943

Jougan-0:
- Tasks done:

- [ ] 1. Stored the file at .meshery/content folder
- [ ] 2. New viper variable for "SERVER_CONTENT_FOLDER"
- [ ] 3. The metadata has a download link and view link 
- [ ] 4. New function to create a download and view link just by passing a file path
- [ ] 5. Change the success type to informational event

- Tasks Left:

- [ ] 1. Proper api call
- [ ] 2. Send event directly from the server

The tasks are left because i can't figure out how can i send the event directly from the function that has logs the values to the terminal because the user.ID is not present when the make server command is ran.

Jougan-0:
> @Jougan-0 use all zeros for the user ID.

@leecalcote please correct me if I am wrong here but first the event is created with new event and then we do persistEvent and then we send the using the 
```erh.handlerConfig.EventBroadcaster.Publish(userID, successEvent)```

while setting the id to all 0s nothing is shown in the notification center 
```ID := "00000000-0000-0000-0000-000000000000"

	userID := uuid.FromStringOrNil(ID)````
I am using this id

MUzairS15:
Thanks @Jougan-0 for your effort, but we will not be proceeding with the approach proposed.
I hope you find some other issue to work upon, as for beginners this issue isn't easy to tackle.

PR #9904, opened 2024-01-10
[UI] Fix CASL issues

Yashsharma1911:
**Notes for Reviewers**

- [X] Subscribe navigator component to CASL ability update to reflect changes in UI
- [x] Prevent "you are unauthorized" component from reflecting

https://github.com/meshery/meshery/assets/71271069/11866ff3-5ace-401e-8afa-abb04c59b427

This PR fixes #


ui/utils/permission_constants.js
```diff
@@ -282,7 +282,7 @@ export const keys = {
     action: 'fdc038e3-1fdf-403a-af8a-53c0de8d7820',
   },
   CONNECT_ADAPTER: {
-    subject: 'Connect Adapter',
+    subject: 'Connect adapter',
```
leecalcote, line 285: Each of these should be converted to upper or lowercase before comparing

ui/pages/_app.js
```diff
@@ -363,39 +363,55 @@ class MesheryApp extends App {
...
-  loadAbility = async (orgID) => {
+  setOrganization = (org) => {
+    const { store } = this.props;
+    store.dispatch({
+      type: actionTypes.SET_ORGANIZATION,
+      organization: org,
+    });
+  };
+
+  loadAbility = async (orgID, reFetchKeys) => {
```
leecalcote, line 410: Is this storing org ID and the keys in a cookie?

ui/pages/_app.js
```diff
@@ -363,39 +363,55 @@ class MesheryApp extends App {
...
-      (result) => {
+      async (result) => {
         let organizationToSet;
 
-        if (currentOrgId !== null) {
-          const indx = result.organizations.findIndex((org) => org.id === currentOrgId);
-          organizationToSet = indx !== -1 ? result.organizations[indx] : result.organizations[0];
+        if (currentOrg) {
+          const indx = result.organizations.findIndex((org) => org.id === currentOrg.id);
+          if (indx === -1) {
```
leecalcote, line 386: Why are you only comparing to negative one? Aren’t there other possible values or scenarios in which the Org ID isn’t set but doesn’t equal negative one?

Yashsharma1911: No, if org doesn't match findIndex will only return -1 and that's when we refetch the keys otherwise we don't

ui/pages/_app.js
Yashsharma1911: No, it only updates the CASL ability based on keys in sessionStorage if not then it will fetch keys and then updates the ability.

senali-d (commented):
LGTM

Yashsharma1911:
> LGTM

Thankyou @senali-d

leecalcote:
@Yashsharma1911 did you make the authorized message 1) dark-mode compatible and 2) say something less off-putting?

PR #9918, opened 2024-01-10
[UI] Filter support for Designs and Filters

sudhanshutech:
**Notes for Reviewers**

This PR adds filter support for designs and filter . I have removed the eye toggle .

>[!NOTE]
>Since catalog content is enabled from preferences so the filter options are here will be shown accordingly. If catalog content is enabled then all filter option such as public, private and published will be shown and if disabled then can filter between public and private

- [x] Filter
- [x] Designs


leecalcote:
Great note. Makes much sense, @sudhanshutech

leecalcote (approved):

PR #9948, opened 2024-01-13
[Permission] Update permission spreadsheet

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

Update permission spreadsheet to contain new keys


Yashsharma1911:
@MUzairS15 @leecalcote I'm concerned that we should not include all keys here for security purposes, that's why I updated this spreadsheet to contain only keys that are necessary for the local provider however lemme know if we should keep all keys 🤔 

@leecalcote updating spreadsheet also fixes the issues of not being able to access UI in the local provider

leecalcote (approved):
Good call. Thank you for publishing only the appropriate keys and for confirming that this fixes the existing issues in the latest release.

PR #9961, opened 2024-01-14
[Env/Workspace] - Remove org selector from create environment and workspace modal

senali-d:
**Notes for Reviewers**

This PR fixes #

- [x] Remove org selector from create environment and workspace modal

### DIFF
--- a/ui/components/Lifecycle/Environments/index.js
+++ b/ui/components/Lifecycle/Environments/index.js
@@ -234,7 +234,7 @@ const Environments = ({ organization, classes }) => {
     );
   };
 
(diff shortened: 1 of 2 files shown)


leecalcote:
@senali-d is this using the org ID from the user's cookie?

leecalcote:
In other words, do we have a consistent mechanism for retrieval of the user's currently selected Org ID?

senali-d:
> In other words, do we have a consistent mechanism for retrieval of the user's currently selected Org ID?

Current organization we stored in the redux store. So here I get the selected org from redux store. // @leecalcote

PR #10091, opened 2024-01-24
[UI] Update publish catalog request payload

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [ ] Yes, I signed my commits.

PR #10095, opened 2024-01-25
Fix casing for displayName in publish payload

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

This (PR) is a correction for PR #10091. I initially used snake_case for model display name in line with the standard we have for every response. However, it turns out that we get camelCase format for model display name. I have updated it accordingly in this PR.


Yashsharma1911:
// @leecalcote

PR #10098, opened 2024-01-25
Prevent lossing technology in infoModal

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

After yesterday change in casing of compatibility few published designs still have compatibility in upper case (UI is expecting model.name now not model.displayName and model.name are always in lowercase), to make sure we are not losing them in info modal I'm converting them first in lower case which will make sure even if designs have compatibility in upper case we don't lose them
 
I tested with most of existing designs I didn't find any more issues, this implementation is already in MeshMap if you want see refer to MeshMap latest release


leecalcote (changes requested):
Do better.

Yashsharma1911:
We are good to go here

PR #10239, opened 2024-02-07
[Docs] Jekyll Build Cache

leecalcote:

Use of Jekyll Build Cache plugin

### DIFF
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -19,6 +19,7 @@ source "https://rubygems.org"
 # If you want to use GitHub Pages, remove the "gem "jekyll"" above and
 # uncomment the line below. To upgrade, run `bundle update github-pages`.
 gem "github-pages", group: :jekyll_plugins
+gem 'jekyll-include-cache'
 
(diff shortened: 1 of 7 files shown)


leecalcote:
See if your build times drop.

vishalvivekm:
> See if your build times drop.

we can also do this
```
          {% if page.category != "integrations" %}
          {% include_cached suggested-reading.html language="en" %} 
          {% endif %}
 ```
 
 In my opinion showing a list of 200 hundred other integrations as suggested reading on each individual integration page is costing us a lot of build time . @iArchitSharma

**Comment by leecalcote** [comment]
Sounds good. And having that list on each of the integration pages isn't helpful.

More than an opinion, have you tested this change to confirm the integration pages are the major offender? Irrespective of the performance improvement, I agree regarding the nil value of having all other integrations as suggested reading.

**Comment by leecalcote** [comment]
P.S. Here's a _very_ helpful tool in to use while answering this question:

```
bundle exec jekyll build --profile
```

Issue #10254, opened 2024-02-08
[UI] React Warnings in Meshery UI Build


leecalcote:
### Current Behavior
While building Meshery UI, the following warnings are shown.

### Desired Behavior
No warnings during `make ui-build` (no build warnings).

### Screenshots/Logs

```
./components/MesheryMeshInterface/PatternService/CustomTextTooltip.js
16:19  Warning: React Hook "useStyles" is called in function "renderTooltipContent" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use".  react-hooks/rules-of-hooks

./components/connections/index.js
987:5  Warning: React Hook "useUpdateConnectionMutator" is called in function "updateConnectionStatus" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use".  react-hooks/rules-of-hooks

./components/hooks/useTelemetryHook.js
18:22  Warning: React Hook "useNotification" is called in function "pingPrometheus" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use".  react-hooks/rules-of-hooks
19:20  Warning: React Hook "useDispatch" is called in function "pingPrometheus" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use".  react-hooks/rules-of-hooks
43:22  Warning: React Hook "useNotification" is called in function "pingGrafana" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use".  react-hooks/rules-of-hooks
44:20  Warning: React Hook "useDispatch" is called in function "pingGrafana" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use".  
```

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
[...]

PR #10266, opened 2024-02-10
Fix the React Warnings in Meshery UI Build

MeenuyD:
Fixed warnings related to improper usage of React Hooks in Meshery UI components.
Updated affected files to ensure that React Hooks are called within functional components or custom React Hook functions as per React guidelines.
#10254


Yashsharma1911 (changes requested):
@MeenuyD please see lint failure

sudhanshutech (approved):

PR #10207, opened 2024-02-04
Fix SQL injection in /api/system/meshsync/resources

gyohuangxin:
**Notes for Reviewers**

This PR fixes #

As mentioned in GitHub Security Lab (GHSL) Vulnerability Report, this PR is to fix a SQL injection in /api/system/meshsync/resources. 
Similar vulnerabilities has been fixed in https://github.com/meshery/meshery/pull/9372  using `SanitizeOrderInput` function, which you can uses as a reference.

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #9372, opened 2023-11-16
Filter "order by" parameter to fix sql inject vuln

phant0ms:
## and export SanitizeOrderInput function

**Notes for Reviewers**

This PR fixes #
fix a sql inject in route `api/system/database`,  in `GetSystemDatabase` function, filter order by parameter with a exiting function `models.SanitizeOrderInput`.

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #10280, opened 2024-02-13
Fix more SQL injections

gyohuangxin:
**Notes for Reviewers**

This PR fixes SQL injection in /api/v2/events and /api/system/meshync/resources/kinds.
Similar vulnerabilities has been fixed in https://github.com/meshery/meshery/pull/10207 and https://github.com/meshery/meshery/pull/9372 using SanitizeOrderInput function, which you can uses as a reference. 

### DIFF
--- a/mesheryctl/pkg/utils/helpers.go
+++ b/mesheryctl/pkg/utils/helpers.go
@@ -547,7 +547,10 @@ func ClearLine() {
 		clearCmd = exec.Command("cmd", "/c", "cls") // for Windows
 	}
 	clearCmd.Stdout = os.Stdout
-	_ = clearCmd.Run()
+	err := clearCmd.Run()
+	if err != nil {
+		log.Fatal(err)
+	}
 }
 
 // StringContainedInSlice returns the index in which a string is a substring in a list of strings


leecalcote:
This is so helpful, @gyohuangxin 💯

mesheryctl/pkg/utils/helpers.go
```diff
@@ -547,7 +547,10 @@ func ClearLine() {
 		clearCmd = exec.Command("cmd", "/c", "cls") // for Windows
 	}
 	clearCmd.Stdout = os.Stdout
-	clearCmd.Run()
+	err := clearCmd.Run()
+	if err != nil {
```
leecalcote, line 551: @gyohuangxin @MUzairS15 I wonder if we should be using a MeshKit Error here. https://docs.meshery.io/project/contributing/contributing-error

leecalcote: What's your perspective, @MUzairS15?

MUzairS15: Yes, the usage of meshkit error is appropriate

gyohuangxin: @leecalcote @MUzairS15 It's a good idea, can we create another issue to track this? So that we can fix the security issues as quickly as possible. What do you think?

leecalcote: Ok, sure.

leecalcote: Here we are - https://github.com/meshery/meshery/issues/10343

leecalcote (approved):

PR #10313, opened 2024-02-15
Fix head caching issue

iArchitSharma:
image

### DIFF
--- a/docs/_includes/head.html
+++ b/docs/_includes/head.html
@@ -18,24 +18,24 @@
   <link rel="icon" type="image/png" href="{{ site.baseurl }}/assets/favicons/android-144x144.png" sizes="144x144" />
   <link rel="icon" type="image/png" href="{{ site.baseurl }}/assets/favicons/android-192x192.png" sizes="192x192" />
 
-  <title>{% if page.title %}{{ page.title }}{% else %}{{ site.title }}{% endif %} | Meshery</title>
-  <meta property="og:title" content="{% if page.title %}{{ page.title }}{% else %}{{ site.title }}{% endif %}" />
+  <title>{% if include.title %}{{ include.title }}{% else %}{{ include.title }}{% endif %} | Meshery</title>
+  <meta property="og:title" content="{% if include.title %}{{ include.title }}{% else %}{{ site.title }}{% endif %}" />
   <meta property="og:description"
-    content="{% if page.abstract %}{{ page.abstract }}{% else %}{{ site.description }}{% endif %}" />
+    content="{% if include.description %}{{ include.description }}{% else %}{{ site.description }}{% endif %}" />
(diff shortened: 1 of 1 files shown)


l5io:
🚀 Preview for commit 68a7338e3ea7242c3b25c0bf067bc67faea5a7d8 at: https://65ce05268e1349066941a0bb--meshery-docs-preview.netlify.app

leecalcote:
Can this include be updated so that we can continue to benefit from the build cache plug-in?

Thanks. The build cache plug-in speaks of a particular way, in which include macros are to be written so that they are compatible with the plug-in. If we pass the name of the page as a parameter in the include call, we might be able to still benefit from the plug-ins performance improvement while avoiding the mislabeling.

l5io:
🚀 Preview for commit 1fdd749419faf0669060bf418358734aaa621dbe at: https://65ce224251a70a1be4dff7b6--meshery-docs-preview.netlify.app

iArchitSharma:
@leecalcote update, also i changed `page.description` to `page.abstract` i don't know for how long it has been there like this :grimacing:

leecalcote:
> @leecalcote update, also i changed `page.description` to `page.abstract` i don't know for how long it has been there like this 😬

Makes sense. I'll await merge based on use of include parameters to be able to continue to use the cache plugin.

iArchitSharma:
@leecalcote I've added the parameters already
image

vishalvivekm:
There are a couple of issues that these changes have caused: 

1. see view-source:https://65ce224251a70a1be4dff7b6--meshery-docs-preview.netlify.app/concepts/architecure/adapters 
a page using `default` layout, it has 404s in titles and has the default site description, despite it having its own title and description: 


The thing is when using `include_cached` we can not rely on page context to get the value of the page variables, 

What we can do is to pass those variable as arguments in `include_cached` what @iArchitSharma has done here
https://github.com/meshery/meshery/pull/10313/files#diff-ddb6373434059f8e97028b6be1bf40dea40384f654e85dabafcea168458db8d0R4-R7, and then get those values using `include.<argument-name>` in our [include](https://github.com/meshery/meshery/pull/10313/files#diff-932f0971e3260b1c7557053593d85a0a51a132209a76c08f9b152920efd4a49cR24) and not `page.<variable-name>`

3. AS we have pages, using both the layouts i.e. `default` as well as `enhanced` let's do it for them both, as if done only for default, pages using enhanced layout will either have 404 or the default titles/descriptions

l5io:
🚀 Preview for commit 096678a703ed44ae69c705a5f7cbf8c82bf81e36 at: https://65ce62446f1883296ad8456b--meshery-docs-preview.netlify.app

vishalvivekm (approved):

leecalcote (approved):

PR #10353, opened 2024-02-20
[Server] Fix events API.

MUzairS15:
**Notes for Reviewers**

This PR fixes #
1. With recent usage of `SanitizeOrderInput`, the API started failing with `invalid query param err`, this was happening as the input provided to the `SanitizeOrderInput ` were `column-name` instead it expects the input as `column-name order` (eg: created_at desc), causing the function to return empty string thereby failing the query.
2. The `page_size` query param, provided by the UI was not being used, as the API expects the param as `pagesize` instead of `page_size`. While it didn't cause any error, the requested page size of 15 was not respected and was getting overridded by a page size of 25 by the server.

### DIFF
--- a/server/models/events_persister.go
+++ b/server/models/events_persister.go
@@ -1,10 +1,11 @@
 package models
 
 import (
+	"fmt"
+
 	"github.com/gofrs/uuid"
 	"github.com/layer5io/meshkit/database"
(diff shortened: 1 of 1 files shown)


MUzairS15:
// @gyohuangxin

gyohuangxin:
@MUzairS15 Thank you for fixing it.

PR #10389, opened 2024-02-23
[Server] Eliminate dead code and Improvements for controllers/dataHandlers. 

MUzairS15:
**Notes for Reviewers**

This PR 
1. Removes dead GraphQL subscriptions from UI and server.
2. Replace the global `ctxTocontroller/dataHandlers` map into their connection management.
3. Remove the usage of `mutex`.
4. Currently in some of the graphQL queries `contextID` was still being used, adding extra processing and in some cases (extra requests to `getConnectionByContextID`). The PR replaces the usage of `contextID` with `connectionID` in the queries.
5. The PR reduces the race conditions in the server codebase.
6. Removes the usage of `taskqueue`. This queue was just initialized and never used.

### DIFF
--- a/ui/components/ConnectionWizard/Screens/MesheryOperatorScreen.js
+++ b/ui/components/ConnectionWizard/Screens/MesheryOperatorScreen.js
@@ -1,7 +1,6 @@
 /* eslint-disable no-unused-vars */
 import MesheryOperatorIcon from '../icons/MesheryOperatorIcon.js';
 import fetchMesheryOperatorStatus from '../../graphql/queries/OperatorStatusQuery';
-import subscribeOperatorStatusEvents from '../../graphql/subscriptions/OperatorStatusSubscription';
 import ServiceCard from '../ServiceCard';
 import { CircularProgress, Grid } from '@material-ui/core';
 import MesheryOperatorDataPanel from '../DataPanels/MesheryOperator';
@@ -39,7 +38,6 @@ const MesheryOperatorScreen = ({ setStepStatus }) => {
   }, [isConnected]);
 
   useEffect(() => {
-    subscribeOperatorStatusEvents(setOperatorState);
     setIsLoading(true);
     fetchMesheryOperatorStatus().subscribe({
       next: (res) => {


server/internal/graphql/resolver/operator.go
```diff
@@ -3,10 +3,13 @@ package resolver
 import (
 	"context"
 
+	"github.com/gofrs/uuid"
```
leecalcote, line 6: @MUzairS15 will you add a description at the top of this file as to what it's purpose is?

server/models/meshery_controllers.go
```diff
@@ -36,146 +36,157 @@ const (
...
+		dbHandler:              dbHandler,
 	}
 }
 
 // initializes Meshsync data handler for the contexts for whom it has not been
 // initialized yet. Apart from updating the map, it also runs the handler after
 // updating the map. The presence of a handler for a context in a map indicate that
 // the meshsync data for that context is properly being handled
-func (mch *MesheryControllersHelper) UpdateMeshsynDataHandlers(ctx context.Context, connectionID, userID, mesheryInstanceID uuid.UUID, provider Provider) *MesheryControllersHelper {
+func (mch *MesheryControllersHelper) AddMeshsynDataHandlers(ctx context.Context, k8scontext K8sContext, userID, mesheryInstanceID uuid.UUID, provider Provider) *MesheryControllersHelper {
```
leecalcote, line 86: How are we preparing for the support of Environments?

server/models/meshery_controllers.go
```diff
@@ -36,146 +36,157 @@ const (
...
+		if brokerEndpoint == "" {
+			if err != nil {
+				mch.log.Warn(err)
 			}
+			mch.log.Info(fmt.Sprintf("Meshery Broker unreachable for Kubernetes context (%v)", ctxID))
+			return mch
 		}
-	}(mch)
+		brokerHandler, err := nats.New(nats.Options{
+			// URLS: []string{"localhost:4222"},
```
leecalcote, line 105: Is this configured as default when Meshery version = "latest" (when user has locally built Meshery)?

server/models/meshery_controllers.go
```diff
@@ -223,46 +231,46 @@ func (mch *MesheryControllersHelper) DeployUndeployedOperators(ot *OperatorTrack
...
-	go func(mch *MesheryControllersHelper) {
+	// go func(mch *MesheryControllersHelper) {
 
-		mch.mu.Lock()
-		defer mch.mu.Unlock()
-		for ctxID, ctrlHandler := range mch.ctxControllerHandlersMap {
+	oprStatus := mch.ctxOperatorStatus
 
-			if oprStatus, ok := mch.ctxOperatorStatusMap[ctxID]; ok {
+	if oprStatus != controllers.Undeployed {
```
leecalcote, line 259: Or `controllers.Disabled`?

ui/components/MeshConfigComponent.js
```diff
@@ -315,13 +316,15 @@ function MesherySettingsNew({
         message: `Operator ${response.operatorStatus.toLowerCase()}`,
         event_type: EVENT_TYPES.SUCCESS,
       });
-      const tempSubscription = fetchMesheryOperatorStatus({ k8scontextID: contextId }).subscribe({
-        next: (res) => {
-          _setOperatorState(updateCtxInfo(contextId, res));
-          tempSubscription.unsubscribe();
+      const tempSubscription = fetchMesheryOperatorStatus({ connectionID: connectionID }).subscribe(
```
leecalcote, line 319: Why is a temp subscription needed? Will you document this in comments?

ui/components/graphql/subscriptions/OperatorStatusSubscription.js
leecalcote, line 1: We're keeping this around in case we need to bring it back?

ui/utils/multi-ctx.js
```diff
@@ -162,3 +162,20 @@ export function getClusterNameFromCtxId(ctxId, k8sConfig) {
   }
   return cluster.name;
 }
+
```
leecalcote, line 165: Meshery is also in need of something like this:

```
export function getConnectionIDsFromEnvironmentId(connections, environment) {
``

**Review by leecalcote — COMMENTED** [review]
First round of review complete.

File: server/handlers/middlewares.go [hunk 7]
```diff
@@ -332,10 +332,15 @@ func K8sFSMMiddleware(ctx context.Context, h *Handler, provider models.Provider,
 				go h.config.EventBroadcaster.Publish(userUUID, event)
 			}
 		}(inst)
-		mdh, ok := ctxToDataHandlerMap[k8sContext.ID]
-		if ok {
+		kubernesMachineCtx, err := utils.Cast[*kubernetes.MachineCtx](inst.Context)
+		if err != nil {
+			h.log.Error(err)
```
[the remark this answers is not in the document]
MUzairS15: Yes, it's an meshkit error

ui/pages/_app.js
```diff
@@ -327,17 +322,16 @@ class MesheryApp extends App {
   initSubscriptions = (contexts) => {
     const mesheryControllerSubscription = new GQLSubscription({
       type: MESHERY_CONTROLLER_SUBSCRIPTION,
-      contextIds: contexts,
+      connectionIDs: getConnectionIDsFromContextIds(contexts, this.props.k8sConfig),
```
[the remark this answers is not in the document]
MUzairS15: Yes, it is not a reference to kubeconfig file.

server/models/meshery_controllers.go
MUzairS15: Undeployed is one of the state that we have agreed upon, so I think it's better to use that.

ui/components/graphql/subscriptions/OperatorStatusSubscription.js
MUzairS15: All other traces of this was deleted, this should also go away.

MUzairS15: Removed

server/models/meshery_controllers.go
leecalcote: Enquiring minds want to know. :)

leecalcote (approved):
Looks good to me. Assuming this isn't a step away from the goal of upgrading to an Environment Switcher.

leecalcote:
The graphql docs build step is failing.

server/models/meshery_controllers.go
MUzairS15: Because environment resources are logical groups of resources that server primary concerns with, hence when the user is issued a session/environment is changed from the switcher the necessary environment related details will be send to the UI and with each request envID needs to be included.

All APIs in the server will be updated to include envID.  As a result you should expect the above handler (and similarly other) have a envID param passed in their context (`ctx`).
In some cases, connectionID is enough (as it itself is an primary key) to manage connection, but including envID adds a direction.

leecalcote: I'll request that we draw a hard line on whether Environment ID becomes a required parameter or not and suggest that we lean into the former.

leecalcote (approved):
What test scenarios have been run through, @MUzairS15?

MUzairS15:
1. Ensuring the connection states work as they should.
2.  If connection is administratively transitioned from connected state the connection state is respected and doesn’t greedily change back to connected state, even if no clusters are connected.
3. The AdHoc tests on the Controllers chip reports correct status.
4. Changes inside the cluster are streamed back to client.
5. Meshery extension features, Log/Term session functions correctly.
6. Real-time reporting of Controller status.

PR #10443, opened 2024-02-28
Add empty state for no result found in Registry

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

This PR adds no result found empty state if there is no result on search, this fixes issue showing previous data if search is not found and giving user a feedback that search is not working.


Yashsharma1911:
This add a simple empty state, considering that I'm moving forward

PR #10576, opened 2024-03-15
Feat: Update model generator

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

This PR updates model generator to source core Meshery components from `meshery/schemas` repo.

### DIFF
--- a/mesheryctl/internal/cli/root/registry/generate.go
+++ b/mesheryctl/internal/cli/root/registry/generate.go
@@ -287,7 +287,9 @@ func GenerateDefsForCoreRegistrant(model utils.ModelCSV) error {
 	}
 	path, err := url.Parse(model.SourceURL)
 	if err != nil {
-		return err
+		err = ErrGenerateModel(err, modelDef.Name)
+		utils.Log.Error(err)
+		return nil
 	}
 	gitRepo := github.GitRepo{
 		URL:         path,
@@ -295,7 +297,9 @@ func GenerateDefsForCoreRegistrant(model utils.ModelCSV) error {
 	}
 	owner, repo, branch, root, err := gitRepo.ExtractRepoDetailsFromSourceURL()
 	if err != nil {
-		return err
+		err = ErrGenerateModel(err, modelDef.Name)
+		utils.Log.Error(err)
+		return nil
 	}
 
 	//Initialize walker


mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -289,37 +292,48 @@ func GenerateDefsForCoreRegistrant(model utils.ModelCSV) error {
...
 	}
+	path, err := url.Parse(model.SourceURL)
+	if err != nil {
+		return err
+	}
+	gitRepo := github.GitRepo{
+		URL:         path,
+		PackageName: model.Model,
+	}
+	owner, repo, branch, root, err := gitRepo.ExtractRepoDetailsFromSourceURL()
```
Yashsharma1911, line 303: This depends on my PR in meshkit https://github.com/meshery/meshkit/pull/481

mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -263,8 +267,8 @@ func InvokeGenerationFromSheet(wg *sync.WaitGroup) error {
 	return nil
 }
 
-// For registrants eg: meshery, whose components needs to be directly created by referencing the sheet.
-// the sourceURL contains the "rangeString" : the value that describes the sheet in which the components are listed, and their starting row to ending row.
+// For registrants eg: meshery, whose components needs to be directly created by referencing meshery/schemas repo.
+// the sourceURL contains the path of models component definitions
 func GenerateDefsForCoreRegistrant(model utils.ModelCSV) error {
 	totalComps := 0
 	version := "1.0.0"
```
leecalcote, line 274: This hardcoding is no good. This needs to be source from tagged releases of `meshery/schemas`. 

Why is the generation process for `meshery-core` models being treated any differently from other models; why is there a function specifically written for generation of Meshery Core/Shape/Flow models? @MUzairS15

Yashsharma1911:
In my last PR of meshery/schemas I forget to push GCP and AWS components, here is PR https://github.com/meshery/schemas/pull/64

mesheryctl/internal/cli/root/registry/generate.go
MUzairS15: This hardcoding of version was a concern when we were referencing the Comps from the sheet (I somewhere mentioned it elsewhere I believe), because the version was not captured in the sheet.

But now because they are sourced from the schemas repo and live inside the versioned directory we can remove this.

I realised this yesterday and thought I will convey @Yashsharma1911 but later forgot.

leecalcote: @Yashsharma1911 make sense?

Yashsharma1911: Let's remove this hardcoding and update generator according to new changes in `meshery/schema` in iterative PR

Yashsharma1911:
Unit tests are not failing because of this PR

PR #10606, opened 2024-03-23
Upgrade to MeshKit v0.7.20

leecalcote:

**Notes for Reviewers**

This PR is a chore.


leecalcote:
Are we able to remove the dependency on the older version of kompose? Ideally, we’re able to update to later versions of k8s.io/api, apimachinery, client-go. I believe that the older version of kompose is holding us back in this regard, though. Please confirm that we still need to pin to older kompose package.

We’re pinned to our fork of kompose v1.0.1 in [MeshKit](https://github.com/meshery/meshkit/blob/v0.7.20/go.mod#L11):

```
github.com/kubernetes/kompose => github.com/meshery/kompose v1.0.1
```

to another version of our kompose fork in Meshery -

```
github.com/kubernetes/kompose => github.com/meshery/kompose v1.26.2-0.20230425025309-3bb778d54007
```

Does this seem right?


We might need to rollback the MeshKit v0.7.20 release, which seems to be pushing a newer version of k8s.io/api / client-go that have deprecated use of `v1alpha1` of the `FlowControl` resource used in mesheryctl’s `system stop` command.


Lee Calcote
  [2 minutes ago](https://layer5io.slack.com/archives/CFGG6U10E/p1711227373912019?thread_ts=1711226942.173569&cid=CFGG6U10E)
Meshery CLI (and by association Meshery Server) and MeshKit are now incompatible in their use of Kubernetes packages.

PR #10650, opened 2024-04-01
fix docs search field

vishalvivekm:
**Notes for Reviewers**

This PR brings back the search fields in meshery docs 
image

### DIFF
--- a/docs/_sass/tabbed-menu.scss
+++ b/docs/_sass/tabbed-menu.scss
@@ -22,9 +22,9 @@ section.tabbed {
 section:after{
   background-color: red;
 }
-input {
-  display: none;
-}
+// input {
+//   display: none;
+// }
 
 label {
   display: inline-block;


docs/_sass/tabbed-menu.scss
```diff
@@ -22,9 +22,9 @@ section.tabbed {
 section:after{
   background-color: red;
 }
-input {
-  display: none;
-}
+// input {
+//   display: none;
+// }
```
leecalcote, line 27: We’re in need of separate classes for the search in the sidebar and the radio buttons on the tabbed menu of the API reference page

vishalvivekm: okay, having a class for the <input> element on `api ref` page makes sense, we can just give `display:none` to that .

vishalvivekm: or just setting hidden attribute for those.
this should do it: https://github.com/meshery/meshery/pull/10664

PR #10624, opened 2024-03-26
upgrade alpine from 3.15.3 to 3.18.5 to solve vulnerabilities 

deepak4566:
<br />Keeping your Docker base image up-to-date means you’ll benefit from security fixes in the latest version of your chosen image.

#### Changes included in this PR 

- install/docker/Dockerfile
 recommended upgrading to `alpine:3.18.5`, as this image has only 0 known vulnerabilities. To do this, merge this pull request, then verify your application still works as expected.

---

PR #10673, opened 2024-04-04
Revert "upgrade alpine from 3.15.3 to 3.18.5 to solve vulnerabilities "

MUzairS15:
Reverts meshery/meshery#10624


MUzairS15:
Screenshot 2024-04-05 at 1 10 51 AM
// @deepak4566 Had to revert the upgrade. The meshery container fails to start.

PR #10683, opened 2024-04-08
[CI][CLI] Include Cloud as a Target in the Publish Workflow

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

- [X] Remove TODO and added code to publish models on remote provider
- [X] Updated publish workflow to target cloud also
- [X] Fix wrong path in publish workflow, which fixes issue of not able to publish models on meshery.io and layer5.io

### DIFF
--- a/ui/components/MeshModelRegistry/MeshModelComponent.js
+++ b/ui/components/MeshModelRegistry/MeshModelComponent.js
@@ -121,7 +121,7 @@ const MeshModelComponent_ = ({
             {
               params: {
                 page: searchText ? 1 : page.Models + 1,
-                pagesize: searchText || checked ? 'all' : rowsPerPage,
+                pagesize: 'all',
                 components: true,
                 relationships: true,
                 search: searchText || '',
--- a/ui/components/MeshModelRegistry/MeshModelDetails.js
+++ b/ui/components/MeshModelRegistry/MeshModelDetails.js
@@ -273,8 +273,8 @@ const RelationshipContents = withSuppressedErrorBoundary(({ relationship, view }
   return (
     <div>
       <div style={{ display: 'flex', flexDirection: 'column' }}>
-        <Title title={relationship.subType} />
-        <Description description={metadata.description} />
+        <Title title={relationship?.subType} />
+        <Description description={metadata?.description} />
       </div>
       <RenderContents
         metaDataLeft={metaDataLeft}


leecalcote:
#3 in the list is a big deal.

mesheryctl/internal/cli/root/registry/publish.go
```diff
@@ -146,6 +142,9 @@ mesheryctl registry publish website $CRED 1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwiz
 		case "remote-provider":
 			err = remoteProviderSystem()
 		case "website":
+			if outputFormat != "md" && outputFormat != "mdx" && outputFormat != "js" {
```
leecalcote, line 145: .svg is to be written out here as well?

leecalcote: Why is this condition needed? What's bug / issue is it fixing?

mesheryctl/internal/cli/root/registry/publish.go
```diff
@@ -177,8 +176,28 @@ func mesherySystem() error {
 	return nil
 }
 
-// TODO
 func remoteProviderSystem() error {
+	outputPath, _ := filepath.Abs(filepath.Join("../", modelsOutputPath))
```
leecalcote, line 180: Please add one or more comments.

mesheryctl/internal/cli/root/registry/publish.go
Yashsharma1911: This condition makes sure that user proveded right output format, otherwise we show error

mesheryctl/pkg/utils/registry.go
```diff
@@ -212,3 +212,49 @@ func GenerateMDStyleDocs(model ModelCSV, components []ComponentCSV, modelPath, i
 
 	return nil
 }
+
+func GenerateIcons(model ModelCSV, components []ComponentCSV, imgPath string) error {
```
leecalcote, line 216: Is this a copy from Meshery Server's code?

Does this come insert the necessary `<xml>` tag?

mesheryctl/pkg/utils/registry.go
```diff
@@ -212,3 +212,49 @@ func GenerateMDStyleDocs(model ModelCSV, components []ComponentCSV, modelPath, i
...
+		return err
+	}
+
+	err = utils.WriteToFile(filepath.Join(whiteIconsDir, modelName+"-white.svg"), model.SVGWhite)
+	if err != nil {
+		return err
+	}
+
+	// Generate components metadata and create SVG files
+	_iconsSubDir := filepath.Join(filepath.Join(strings.Split(imgPath, "/")[1:]...), modelName, "components")
```
leecalcote, line 252: What's `1:`?

mesheryctl/pkg/utils/registry.go
```diff
@@ -212,3 +212,49 @@ func GenerateMDStyleDocs(model ModelCSV, components []ComponentCSV, modelPath, i
...
+	_imgOutputPath := filepath.Join(imgsOutputPath, "components")
+	_, err = CreateComponentsMetadataAndCreateSVGsForMDStyle(components, _imgOutputPath, _iconsSubDir)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
leecalcote, line 259: Send feedback to user about the # of models published?

mesheryctl/internal/cli/root/registry/publish.go
```diff
@@ -177,8 +176,28 @@ func mesherySystem() error {
...
+			comps = []utils.ComponentCSV{}
+		}
+
+		err := utils.GenerateIcons(model, comps, imgsOutputPath)
+		if err != nil {
+			utils.Log.Debug(ErrGeneratingIcons(err, imgsOutputPath))
+			log.Fatalln(fmt.Printf("Error generating icons for model %s: %v\n", model.Model, err.Error()))
+		}
+
+		_, _, err = WriteModelDefToFileSystem(&model, "", modalDir)
```
leecalcote, line 195:
```suggestion
		_, _, err = WriteModelDefToFileSystem(&model, "", modelDir)
```

mesheryctl/internal/cli/root/registry/publish.go
```diff
@@ -177,8 +176,28 @@ func mesherySystem() error {
 	return nil
 }
 
-// TODO
 func remoteProviderSystem() error {
+	outputPath, _ := filepath.Abs(filepath.Join("../", modelsOutputPath))
+	modalDir := filepath.Join(outputPath)
```
leecalcote, line 181:
```suggestion
	modelDir := filepath.Join(outputPath)
```

.github/workflows/integrations-updater.yml
```diff
@@ -37,23 +37,32 @@ jobs:
...
         run: |
-          ./mesheryctl registry publish website ${{ secrets.INTEGRATION_SPREADSHEET_CRED }} 1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw layer5/src/collections/integrations layer5/src/collections/integrations -o mdx
+          ./mesheryctl registry publish website ${{ secrets.INTEGRATION_SPREADSHEET_CRED }} 1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw ../layer5/src/collections/integrations ../layer5/src/collections/integrations -o mdx
       - name: Pubish to Meshery docs
         working-directory: ./meshery/mesheryctl
         run: |
           ./mesheryctl registry publish website ${{ secrets.INTEGRATION_SPREADSHEET_CRED }} 1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw docs/pages/integrations docs/assets/img/integrations -o md
-          
+      - name: Pubish to Meshery docs
+        working-directory: ./meshery/mesheryctl
```
leecalcote, line 63: just checking... we don't need an extra ".", do we? -  `../meshery/mesheryctl`

mesheryctl/pkg/utils/registry.go
leecalcote: `<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg>`

Reference - https://docs.meshery.io/project/contributing/contributing-ui#static-files-icons-and-images

Yashsharma1911: Yes, it insert necessary xml tags, no it is not copy of server

Yashsharma1911: This code is present in mesheryctl however it was used in 3 places, so I exracted this code and made a generic function that can be reused instead of writing this code again and again

leecalcote: I propose that it be a copy of Meshery Server... in terms of the treatment of SVGs.

Good of you to consolidate to single place in mesheryctl.

mesheryctl/pkg/utils/registry.go
Yashsharma1911: 1: is a slicing operation that removes the first emtpy string in slice

For instance, if we have the path "/assets/img/integrations", splitting it by the "/" delimiter gives us ["", "assets", "img", "integrations"]. Now, applying 1: to this slice selects all elements from the second element onward, effectively removing the empty string "".

So, after the operation, we get "assets/img/integrations"

mesheryctl/internal/cli/root/registry/publish.go
Yashsharma1911: Okay

mesheryctl/internal/cli/root/registry/publish.go
Yashsharma1911: Thankyou for identifying

.github/workflows/integrations-updater.yml
Yashsharma1911: No it is correct, in current Github action context after checking out the other repos we will have file structure something like this, and by default we will be in root, so to run mesheryctl command we will be using `./meshery/mesheryctl` because if we use `../meshery/mesheryctl` it will show no file exist issue

- meshery
  | - [files in meshery directory]
- layer5
  | - [files in layer5 directory]
- meshery.io
  | - [files in meshery.io directory]

mesheryctl/internal/cli/root/registry/publish.go
Yashsharma1911: Thankyou

mesheryctl/pkg/utils/registry.go
Yashsharma1911: I think yes we should do that, I'll add that code thankyou for identifying

leecalcote (dismissed):

mesheryctl/pkg/utils/registry.go
Yashsharma1911: Okay I'll check, last time Pranav worked on this and added this code, okay I'll check if he used the same code that we have in server or different

mesheryctl/pkg/utils/registry.go
leecalcote: Excellent

mesheryctl/pkg/utils/registry.go
leecalcote: Thanks. If it doesn’t make sense to copy / be the same, then please ignore this comment.

PR #10693, opened 2024-04-08
move relationship generate command into registry

RipulHandoo:
**Notes for Reviewers**

This PR fixes #


mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -98,25 +112,55 @@ mesheryctl registry generate --registrant-def [path to connection definition] --
...
+			return nil
+		} else if relationshipFlag {
+			resp, err := srv.Spreadsheets.Values.Get(spreadsheeetID, "Relationships").Fields().Do()
+			if err != nil || resp.HTTPStatusCode != 200 {
+				utils.Log.Error(ErrUpdateRegistry(err, outputLocation))
+				return err
+			}
+
+			if len(resp.Values) == 0 {
+				utils.Log.Info("No data(relationships) found in the sheet")
```
hamza-mohd, line 151:
```suggestion
				utils.Log.Info("No relationships found in the sheet")
```

mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -389,7 +433,55 @@ func logModelGenerationSummary(modelToCompGenerateTracker *store.GenerticThreadS
 	utils.Log.Info(fmt.Sprintf("Generated %d models and %d components", totalAggregateModel, totalAggregateComponents))
 }
 
+func createJsonFile(resp *sheets.ValueRange) error {
+
+	var customResp []CustomValueRange
+
+	for _, row := range resp.Values[2:] {
+		if len(row) >= 15 && row[0] != "" {
```
hamza-mohd, line 441: @RipulHandoo, this approach is fragile and breaks the flexibility of offered in the other publishing commands that work based on column name.

@MUzairS15, is that right?

mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -389,7 +433,55 @@ func logModelGenerationSummary(modelToCompGenerateTracker *store.GenerticThreadS
...
+		}
+	}
+
+	jsonData, err := json.MarshalIndent(customResp, "", "    ")
+	if err != nil {
+		utils.Log.Error(err)
+		return nil
+	}
+
+	jsonFile, err := os.Create("../docs/_data/RelationshipsData.json")
```
hamza-mohd, line 468: Where is the helpful error that is presented to the user to let know that this command assumes they are running it from within the root of a cloned Meshery repo?

mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -389,7 +433,55 @@ func logModelGenerationSummary(modelToCompGenerateTracker *store.GenerticThreadS
...
+	if err != nil {
+		utils.Log.Error(err)
+		return nil
+	}
+	return nil
+}
+
 func init() {
+	generateCmd.Flags().BoolVar(&modelFlag, "model", false, "generate only model")
+	generateCmd.Flags().BoolVar(&relationshipFlag, "relationship", false, "generate only relationship")
```
leecalcote, line 484:
```suggestion
	generateCmd.Flags().BoolVar(&relationshipFlag, "relationship", false, "only generate relationships")
```

mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -389,7 +433,55 @@ func logModelGenerationSummary(modelToCompGenerateTracker *store.GenerticThreadS
...
+	_, err = jsonFile.Write(jsonData)
+	if err != nil {
+		utils.Log.Error(err)
+		return nil
+	}
+	return nil
+}
+
 func init() {
+	generateCmd.Flags().BoolVar(&modelFlag, "model", false, "generate only model")
```
leecalcote, line 483:
```suggestion
	generateCmd.Flags().BoolVar(&modelFlag, "model", false, "only generate models")
```

mesheryctl/internal/cli/root/registry/generate.go
iArchitSharma: yes json is good to go

mesheryctl/internal/cli/root/registry/generate.go
leecalcote: 👀

Yashsharma1911 (commented):
@iArchitSharma @RipulHandoo we don't need any golang command to publish relationship if they are already present in spreadhsheet, we can use existing workflows to fetch spreadsheet data in csv similar to Layer5 docs and use jkyll to show that data, is there any issue of using csv file in jekyll @iArchitSharma intead of JSON

Yashsharma1911:
Checkout - https://github.com/meshery/meshery/blob/master/.github/workflows/generate_keys.yml and change if required to make it generic in nature

mesheryctl/internal/cli/root/registry/generate.go
leecalcote: hmm. We'll need another set of eyeballs here.

PR #10701, opened 2024-04-09
Convert Class Component into Functional component

devhimanshuu:
ErrorBoundary functional components updated with Material Ui latest version

### DIFF
--- a/ui/components/ErrorBoundary.js
+++ b/ui/components/ErrorBoundary.js
@@ -10,10 +10,6 @@ const ErrorBoundary = ({ children }) => {
       setHasError(true);
       setError(error);
     };
-    window.addEventListener('error', handleErrors);
-    return () => {
-      window.removeEventListener('error', handleErrors);
-    };
   }, []);
 
   const resetErrorBoundary = () => {
@@ -36,5 +32,5 @@ const ErrorBoundary = ({ children }) => {
   return children;
 };
 
-export default ErrorBoundary;
+export default ErrorBoundary;       
 


vishalvivekm:
Thank you @devhimanshuu 
Let's discuss this on Meshery Dev call, add this as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit) , if you would, please.

abhishekmmgn (commented):
This is good. Thank you for the PR.

ui/components/ErrorBoundary.js
```diff
@@ -1,40 +1,43 @@
...
 
-  resetErrorBoundary = () => {
-    this.setState({ hasError: false, error: null });
+    const resetErrorBoundary = () => {
+      setHasError(false);
+      setError(null);
+    };
+
+    window.addEventListener('error', handleErrors);
+    return () => window.removeEventListener('error', handleErrors);
```
Yashsharma1911, line 20: We don't need to listen to event, @devhimanshuu is there specific reason to use this, how you make sure we don't use error boundaries for all error which can be ignored

Yashsharma1911 (changes requested):
Hey @devhimanshuu thanks for raising this PR, let's put your effort on right by by depricating this component, we already have a react boundary defined here, so can help in making sure that we delete this `ErrorBoundary` component and where ever we are using it, we replace with this [ErrorBoundaries](https://github.com/meshery/meshery/blob/387f71324f66e07b1bcb19ddb3bc9a6ca0cadda1/ui/components/General/ErrorBoundary.js#L33)

Yashsharma1911:
Lint checks are failing

ui/components/ErrorBoundary.js
devhimanshuu: I agree what you are saying but, i add these event listener for some global errors or unexpected error.  may be there is some optimized way of handling these errors.

leecalcote:
Thank you, @devhimanshuu

PR #10728, opened 2024-04-12
Meshery UI permissions addition/fix

dragon-slayer875:
**Notes for Reviewers**

This PR adds missing permissions for various actions in -
- Header
- Dashboard
    - ConnectionChart
    - DashboardMeshModelGraph
    - MesheryConfigurationChart
    - WorkloadChart  
- Filters
- Designs
    - Configurator 
    - CustomToolbarSelect
    - MesheryPatterns
    - MesheryPatternCards
- Environments
    - EnvironmentsCard
- Lifecycle
    - TransferList 
- Workspaces
    - WorkspaceCard
- User menu
- Settings
- MesheryMetrics

This also fixes -
- Settings displaying a blank screen when user doesn't have permissions.

### DIFF
--- a/ui/assets/icons/DryRunIcon.js
+++ b/ui/assets/icons/DryRunIcon.js
@@ -4,7 +4,7 @@ export const DryRunIcon = ({
   fill = "currentColor",
   height = "20",
   width = "20",
-  ...props 
+  ...props
 }) => {
   return (
     <svg width={height} height={width} viewBox="0 0 28 28" fill="none" xmlns="http://www.w3.org/2000/svg" {...props} >


ui/components/ConfirmationModal.js
```diff
@@ -383,10 +385,14 @@ function ConfirmationMsg(props) {
...
                 }
+                disabled={!CAN(keys.VALIDATE_DESIGN.action, keys.VALIDATE_DESIGN.resource)}
               />
             )}
             <Tab
-              disabled={disabled}
+              disabled={
+                !CAN(keys.UNDEPLOY_DESIGN.action, keys.UNDEPLOY_DESIGN.subject) ||
+                (CAN(keys.UNDEPLOY_DESIGN.action, keys.UNDEPLOY_DESIGN.subject) && !disabled)
+              }
```
Yashsharma1911, line 395: We can simplify this, any reason we should stick with the change you made?

```suggestion
              disabled={!CAN(keys.UNDEPLOY_DESIGN.action, keys.UNDEPLOY_DESIGN.subject) || !disabled}
```

Yashsharma1911 (dismissed):
Other LGTM

ui/components/ConfirmationModal.js
dragon-slayer875: @Yashsharma1911 I wrote it that way because we need to first check if the person has the permission to do that action or not, and then check if the disabled property (which is dependent on the length of the k8s context.) is true or false.
Your suggestion has the chance to enable the action even if the person doesn't have permission for that action.

dragon-slayer875: Though I just realised the '!' mark before diabled isn't needed and have updated the PR for the same.

Yashsharma1911 (approved):
LGTM, please keep your eye after merging this to make sure we don't introduce any error or issue

Yashsharma1911:
Ohh ho, I missed the pr do not merge label, should we revert this PR? is this contain any breaking change?

PR #10824, opened 2024-04-30
[Server] fix pagination logic

MUzairS15:
**Notes for Reviewers**

The current pagination logic returned the same results when the `page` param was passed as `0` and `1`, because of the broken logic of calculated `offset`.
 
The PR ensures that the `offset` is calculated correctly.

### DIFF
--- a/server/handlers/component_handler.go
+++ b/server/handlers/component_handler.go
@@ -55,12 +55,12 @@ func (h *Handler) GetMeshmodelModelsByCategories(rw http.ResponseWriter, r *http
 	}
 	pagestr := queryParams.Get("page")
 	page, _ := strconv.Atoi(pagestr)
-	if page <= 0 {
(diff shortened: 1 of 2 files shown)


MUzairS15:
// @leecalcote

server/handlers/component_handler.go
```diff
@@ -55,12 +55,12 @@ func (h *Handler) GetMeshmodelModelsByCategories(rw http.ResponseWriter, r *http
 	}
 	pagestr := queryParams.Get("page")
 	page, _ := strconv.Atoi(pagestr)
-	if page <= 0 {
```
MUzairS15, line 58: Replace the repetitive logic for extracting `page, pagesize, search, offset` from API handlers to use  `getPaginationParams ` in `utils.go`.

PR #10912, opened 2024-05-09
Remove boolean values from `selectablerows` prop

sudhanshutech:
**Notes for Reviewers**

This PR fixes the warning `depraction` notice which was coming in meshery because we were using boolean values for a prop `selectable row` which is responsible for selecting rows which does not takes any boolean values in last update of table package


sudhanshutech:
@leecalcote

PR #10961, opened 2024-05-15
[mesheryctl] Bug fixes in `mesheryctl system start`

MUzairS15:
**Notes for Reviewers**

This PR fixes #
1. `mesheryctl system start` command is run using the `MESHERY_SERVER_CALLBACK_URL` env var, the variable is not set inside the container. 
2. When using  `port-forward` flag, a random port is chosen on the host device, and traffic is then directed to the server within the Pod. This presents challenges in specifying a custom value for `MESHERY_SERVER_CALLBACK_URL`, as this value must be provided at server startup. The PR ensures that whatever value is assigned to the environment variable, the `port-forward` will correctly forward to the corresponding value of the environment variable `<host:port>`.

### DIFF
--- a/mesheryctl/internal/cli/root/system/dashboard.go
+++ b/mesheryctl/internal/cli/root/system/dashboard.go
@@ -72,8 +72,8 @@ mesheryctl system dashboard
 // Open Meshery UI in browser and use port-forwarding (if default port is taken already)
 mesheryctl system dashboard --port-forward
 
-// Open Meshery UI in browser and use port-forwarding, listen on port 30000 locally, forwarding traffic to meshery server in the pod
-mesheryctl system dashboard --port-forward -p 30000
+// Open Meshery UI in browser and use port-forwarding, listen on port 9081 locally, forwarding traffic to meshery server in the pod
+mesheryctl system dashboard --port-forward -p 9081
 
 // (optional) skip opening of MesheryUI in browser.
 mesheryctl system dashboard --skip-browser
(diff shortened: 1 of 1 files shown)


mesheryctl/internal/cli/root/system/dashboard.go
```diff
@@ -36,14 +36,15 @@ import (
...
 	runPortForward bool
+	localPort      int
 )
 
 // dashboardOptions holds values for command line flags that apply to the dashboard
 // command.
 type dashboardOptions struct {
-	host    string
-	port    int
-	podPort int
```
leecalcote, line 46: Or is this just the Kubernetes api address, not necessarily the specific host?

mesheryctl/internal/cli/root/system/dashboard.go
```diff
@@ -71,6 +72,9 @@ mesheryctl system dashboard
 // Open Meshery UI in browser and use port-forwarding (if default port is taken already)
 mesheryctl system dashboard --port-forward
```
leecalcote, line 74: Please clarify whether this port number assignment is by default or just an example. Why not use 9081 as the default (and as the example)?

mesheryctl/internal/cli/root/system/dashboard.go
```diff
@@ -168,7 +165,7 @@ mesheryctl system dashboard --skip-browser
 						}
 					}
 				}()
-				log.Info(fmt.Sprintf("Forwarding ports %v -> %v", options.podPort, port))
+				log.Info(fmt.Sprintf("Forwarding ports %v -> %v", options.podPort, localPort))
```
leecalcote, line 168: Since we’re only forwarding one port, don’t use the plural version of “ports”.

leecalcote (commented):
Will you confirm that the port-forward command gracefully handles `platform` != “kubernetes@ and informs the user that this is a platform-specific command?

MUzairS15:
When used with docker, it just skips the protforward logic and directly opens the browser at `endpoint` as specified in mesheryctl context.
Screenshot 2024-05-17 at 8 50 36 PM

// @Jougan-0 You can quickly update to add this warning message before continutinh to open the browser.

mesheryctl/pkg/utils/portforward.go
```diff
@@ -44,7 +44,7 @@ func NewPortForward(
 	ctx context.Context,
 	client *meshkitkube.Client,
 	namespace, deployName string,
-	host string, localPort, remotePort int,
+	host string, localPort, remotePort int, // I think "remotePort" should be idenitified dynamically based on the retrieved Pod configuration, instead of assuming 8080, or asking from user?
```
leecalcote, line 47: Yes, although, for the service to be listening on a port other than 8080, the user will have to be running custom Meshery image builds.

leecalcote (approved):

PR #10968, opened 2024-05-15
[server]fix missing services attribute for empty design

MUzairS15:
**Notes for Reviewers**
The `Services` struct had `omitempty` tag because of which if the design was empty, it was not being streamed inside the response.

### DIFF
--- a/ui/components/Header.js
+++ b/ui/components/Header.js
@@ -377,7 +377,7 @@ function K8sContextMenu({
               src={
                 connectionMetadataState
                   ? `/${connectionMetadataState[CONNECTION_KINDS.KUBERNETES]?.icon}`
-                  : ''
+                  : '/static/img/kubernetes.svg'
               }
               width="24px"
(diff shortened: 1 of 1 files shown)


MUzairS15:
// @aabidsofi19 @leecalcote

leecalcote (commented):
Removal of this also leaves the system vulnerable to a nil primary key...

MUzairS15:
The services key as you said should be ever present, and this PR ensures this happens.
Comparing  it with primary constraint not null, if design is empty, it is not that `services` will be null but an empty object.

skipping/removing the key for empty design makes it difficult to understand whether design is actually empty or have got corrupted.

leecalcote:
> The services key as you said should be ever present, and this PR ensures this happens. Comparing it with primary constraint not null, if design is empty, it is not that `services` will be null but an empty object.

Aha! Ok, that’s very good.

PR #10973, opened 2024-05-16
[Server] Fix `k8s manifest` to `pattern` conversion logic

MUzairS15:
**Notes for Reviewers**

When a manifest is uploaded containing cluster-scoped resources (eg, ClusterRole, PV..), the `namespace: default` property is assigned, leading to the invalid design and dry-run/deployment failures.
The PR fixes the issue by assigning the `namespace` property only to namespaced-scoped resources by referencing the registry.
In case the introspection fails, we skip assigning the namespace property, in this case on the deployment of the design, Kubernetes will automatically assign the `default` namespace to valid namespaced resources.


MUzairS15:
// @leecalcote

leecalcote:
**Good.**

> When a manifest is uploaded containing cluster-scoped resources (eg, ClusterRole, PV..), the namespace: default property is assigned, leading to the invalid design and dry-run/deployment failures.
The PR fixes the issue by assigning the namespace property only to namespaced-scoped resources by referencing the registry.
In case the introspection fails, we skip assigning the namespace property,

**Needs to improve to allow users to specify a namespace during the deploy modal moment,** @MUzairS15 

> in this case on the deployment of the design, Kubernetes will automatically assign the default namespace to valid namespaced resources.

PR #11002, opened 2024-05-20
Mesheryctl's unit tests fix

Jougan-0:
**Notes for Reviewers**

This PR fixes #

Update few golden files to work with pageSize=1 and for the command the model command needs an interactive terminal that causes the issue 
specifically ``` No such device or directory /dev/tty```

checking only for the first 25 models could iterator and check for rest but doesn't made sense .

The reset_docker.output.golden file was missing causing an error so fetched it from the old commit history . Reason for the deletion still unclear.

The server was not starting and due to the new implementation the server is working fine could be a case when the server doesn't start in 100 milisecond but couldn't find a case when it crashed for me.

This is fixing the linting in my repo hope it does in main too.Finger-Cross


mesheryctl/internal/cli/root/system/fixtures/view.model.api.response.golden
```diff
@@ -1 +1 @@
-{"page":1,"page_size":1,"total_count":1,"models":[{"name":"spire","version":"0.12.2","displayName":"SPIRE","category":{"name":"Security \u0026 Compliance","metadata":null},"metadata":{"svgColor":"ui/public/static/img/meshmodels/identity-manager/color/workloadidentity-color.svg","svgWhite":"ui/public/static/img/meshmodels/identity-manager/white/workloadidentity-white.svg"},"duplicates":0}]}
+{"page":0,"page_size":1,"total_count":1,"models":[{"id":"bda7b409-8025-5428-8274-0c1399d20069","schemaVersion":"core.meshery.io/v1beta1","version":"v1.0.0","name":"spire","displayName":"SPIRE","description":"","status":"enabled","hostID":"26e64aa8-2d86-57bd-92b3-2a73160d27b4","registrant":{"hostname":"artifacthub"},"category":{"name":"Security \u0026 Compliance","metadata":null},"subCategory":"","metadata":{"isAnnotation":false,"source_uri":"https://github.com/philips-labs/helm-charts/releases/download/spire-0.12.2/spire-0.12.2.tgz","svgColor":"ui/public/static/img/meshmodels/spire/color/spire-color.svg","svgWhite":"ui/public/static/img/meshmodels/spire/white/spire-white.svg"},"model":{"version":"0.12.2"},"components":null,"relationships":null,"duplicates":0}]}
```
leecalcote, line 1: Please remove this test altogether. It's silly to have such tests in the first place. We should have never allowed it.

Jougan-0: Done

PR #11022, opened 2024-05-23
[Server] Lazy conversion from design formats to meshery-design

MUzairS15:
**Notes for Reviewers**

Supports lazy conversion of the source content to design format.
Introspects and converts the pattern_file into design format if the following conditions are met:

pattern_file attribute is empty, and
The "type" (sourcetype/original content) is not Design. [is one of compose/helmchart/manifests]

### DIFF
--- a/server/handlers/meshery_pattern_handler.go
+++ b/server/handlers/meshery_pattern_handler.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"bytes"
+	"context"
 	"database/sql"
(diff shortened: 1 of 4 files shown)


leecalcote:
@MUzairS15 Does source type still need to be expressly defined?

MUzairS15:
While we are using the recently added introspector in meshkit to distinguish whether a chart/manifest during model generation, it is yet to be replaced in server/meshreyctl

PR #11048, opened 2024-05-25
Revert "[Server] Lazy conversion from design formats to meshery-design"

MUzairS15:
Reverts meshery/meshery#11022


MUzairS15:
Need minor change, build is failing

PR #11071, opened 2024-05-30
generate view link for the deployed design

MUzairS15:
**Notes for Reviewers**

This PR fixes #

### DIFF
--- a/server/handlers/design_engine_handler.go
+++ b/server/handlers/design_engine_handler.go
@@ -146,7 +146,14 @@ func (h *Handler) PatternFileHandler(
 		"summary": response,
 	}
 
-	event := eventBuilder.WithSeverity(events.Informational).WithDescription(description).WithMetadata(metadata).Build()
+	serverURL, _ := r.Context().Value(models.MesheryServerURL).(string)
+	
+	if action == "deploy" {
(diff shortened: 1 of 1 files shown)


server/handlers/design_engine_handler.go
```diff
@@ -146,7 +146,14 @@ func (h *Handler) PatternFileHandler(
...
 
-	event := eventBuilder.WithSeverity(events.Informational).WithDescription(description).WithMetadata(metadata).Build()
+	serverURL, _ := r.Context().Value(models.MesheryServerURL).(string)
+	
+	if action == "deploy" {
+		viewLink := fmt.Sprintf("%s/extension/meshmap?mode=visualize&design=%s", serverURL, patternID)
+		description = fmt.Sprintf("%s. View deployed design %s", description, viewLink)
+	}
+
+	event := eventBuilder.WithSeverity(events.Success).WithDescription(description).WithMetadata(metadata).Build()
```
leecalcote, line 156: Should this be an Info severity? @MUzairS15

MUzairS15: This event is emitted when the design gets deployed and the metadata of the event contains which comps were deployed and their type, so it should be of severity success (currently we have `info`) but `success` makes more sense.

leecalcote: Oh, ok. 👍

leecalcote (approved):

PR #11074, opened 2024-05-31
Add Key ID & Monikers

iArchitSharma:


l5io:
🚀 Preview for commit a86e6bb89bc490a0ddb5104a0707d323e5991ce4 at: https://6659c4db043c342e370496a5--meshery-docs-preview.netlify.app

docs/_includes/permissions.html
```diff
@@ -4,63 +4,69 @@
...
-  #tablesContainer {
-    max-width: 1200px; 
-    margin: auto;
+
+  .hidden-row {
+    display: none;
   }
-}
 
+  .odd-row {
```
dragon-slayer875, line 32: Instead of providing odd-row class maybe you can use the nth-child(odd/even) selector.
It should further help in reducing the code.

dragon-slayer875 (changes requested):

docs/_includes/permissions.html
iArchitSharma: previously that was the approach but odd event does not work in our case because every even row is a hidden row, so by default table will look fully grey

dragon-slayer875: Oh I see. makes sense.

dragon-slayer875 (approved):
LGTM. Thank you mate. :smile:

docs/_includes/permissions.html
dragon-slayer875: Maybe instead of even odd you can then try 
image

dragon-slayer875 (changes requested):
nth-child(3) might work perhaps?

docs/_includes/permissions.html
iArchitSharma: this would only select the third item in list, rest of the list will not be selected

l5io:
🚀 Preview for commit 028ffaf48c698ac9a4aed90ddfce947c135517c4 at: https://665c10b47c16716956c02c4e--meshery-docs-preview.netlify.app

vishalvivekm:
@iArchitSharma Let's discuss this on Websites's call on Monday. Adding this as an agenda item into the [meeting minutes](https://docs.google.com/document/d/1XczAHXVe2FIWPqiF57ospJ43zw5cZQ7ui8mn39v5EvA/edit).

dragon-slayer875 (approved):
LGTM. Thank you. :smile:

l5io:
🚀 Preview for commit 6b1f484db0af84d4f8e32bb3cc884483db7c99e6 at: https://665d3317003792efc90f34a0--meshery-docs-preview.netlify.app

PR #11144, opened 2024-06-10
fix ratelimit artifacthub w/signoff

Jougan-0:
**Notes for Reviewers**

This PR fixes the rate limit issue due to which status code 429 was shown in logs.

### DIFF
--- a/mesheryctl/internal/cli/root/registry/generate.go
+++ b/mesheryctl/internal/cli/root/registry/generate.go
@@ -184,6 +184,7 @@ func InvokeGenerationFromSheet(wg *sync.WaitGroup) error {
 		if err != nil {
 			break
 		}
+		utils.Log.Info("Current model: ", model.Model)
 		wg.Add(1)
 		go func(model utils.ModelCSV) {
 			defer func() {
@@ -231,7 +232,6 @@ func InvokeGenerationFromSheet(wg *sync.WaitGroup) error {
 				utils.Log.Error(ErrGenerateModel(err, model.Model))
 				return
 			}
-			utils.Log.Info("Current model: ", model.Model)
 			utils.Log.Info(" extracted ", len(comps), " components for ", model.ModelDisplayName, " (", model.Model, ")")
 			for _, comp := range comps {
 				comp.Version = defVersion


Jougan-0:
Local [Action](https://github.com/Jougan-0/meshery/actions/runs/9439977307) Run.
Known issue package with more than one word name is showing 400 error need to replace them with %20 will update afterwards.

mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -199,8 +206,10 @@ func InvokeGenerationFromSheet(wg *sync.WaitGroup) error {
 				return
 			}
 
-			if model.Registrant == "artifacthub" {
-				time.Sleep(10 * time.Second)
+			if model.Registrant == "Artifact Hub" {
```
MUzairS15, line 209: @Jougan-0 in the model.go file just format the registrant name using `ReplaceSpacesAndConvertToLowercase` (meshkit func) and then try if still ratelimitting is hit. (keeping the old logic for ratelimitting, as I think 5 mins is too long even if you are doing in bunch)

MUzairS15: Previously the registrant name was formatted as sheet was parsed through, but later I moved it so that the real name as listed in sheet gets utilised but missed update here. And before this I wasn't seeing 429s in the action run. 
I was of the belief that the sleep logic is being used but after seeing your PR I realised this if block was getting skipped because of the above mentioned change.

Jougan-0: Changing back to old logic and running the [action](https://github.com/Jougan-0/meshery/actions/runs/9447298507)  we are still getting the 429 error .
I don;t think so that 5 mins time is much because for every 100 artifacthub registrant we wait for 5 minutes so it takes 30 minutes at max now to run the whole script.
[Here](https://github.com/Jougan-0/meshery/actions/runs/9439977307) have a look.

mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -199,8 +206,10 @@ func InvokeGenerationFromSheet(wg *sync.WaitGroup) error {
 				return
 			}
 
-			if model.Registrant == "artifacthub" {
-				time.Sleep(10 * time.Second)
+			if mutils.ReplaceSpacesAndConvertToLowercase(model.Registrant) == "artifacthub" {
+				time.Sleep(1 * time.Second)
```
MUzairS15, line 210: Why do we need this, when we have https://github.com/meshery/meshery/pull/11144/files#diff-9af4ad85714a082c37d38006b390c594453af5b29a63e3c272a42805d8bffef8R211 ?

Jougan-0: Removed it it's redundant .

MUzairS15 (approved):

PR #11146, opened 2024-06-10
fix assign of connections in environment transferlist

sudhanshutech:
**Notes for Reviewers**

This PR fixes the regression issue which was causing in assigning connections in environment. This will solve the regression  issue but snot the complete issue because permissions are applied for assigning connections to environment and what i am seeing is the environment keychain is missing in any of the role.
> [!IMPORTANT]
> The environment management keychain need to be added in role like admin or org admin most probably @Yashsharma1911 (you may know well) because due to this the button is disabled since role does not contains this keychain assigned ,then this pr can be merged


ui/components/Lifecycle/Environments/index.js
```diff
@@ -385,15 +385,7 @@ const Environments = ({ organization, classes }) => {
...
       getAddedAndRemovedConnection(updatedAssignedData);
-    (addedConnectionsIds.length > 0 || removedConnectionsIds.length) > 0 &&
-    (CAN(
-      keys.ASSIGN_CONNECTIONS_TO_ENVIRONMENT.action,
-      keys.ASSIGN_CONNECTIONS_TO_ENVIRONMENT.subject,
-    ) ||
-      CAN(
-        keys.REMOVE_CONNECTIONS_FROM_ENVIRONMENT.action,
-        keys.REMOVE_CONNECTIONS_FROM_ENVIRONMENT.subject,
-      ))
```
Yashsharma1911, line 396: @sudhanshutech why are you removing keys

sudhanshutech: because the transferlist we use already holds the ability to add permissions(which checks to assign or not) and we are using it so we don't required here duplicative thing

sudhanshutech:
@Yashsharma1911 good to go?

ui/components/Lifecycle/Environments/index.js
```diff
@@ -598,13 +590,13 @@ const Environments = ({ organization, classes }) => {
...
                   originalRightCount={environmentConnections?.total_count}
                   leftPermission={CAN(
-                    keys.REMOVE_CONNECTIONS_FROM_ENVIRONMENT.action,
-                    keys.REMOVE_CONNECTIONS_FROM_ENVIRONMENT.subject,
-                  )}
-                  rightPermission={CAN(
                     keys.ASSIGN_CONNECTIONS_TO_ENVIRONMENT.action,
                     keys.ASSIGN_CONNECTIONS_TO_ENVIRONMENT.subject,
                   )}
+                  rightPermission={CAN(
```
dragon-slayer875, line 596: Why reverse keys?
The left is of remove and right of assign.

dragon-slayer875 (changes requested):

ui/components/Lifecycle/Environments/index.js
sudhanshutech: please check the other areas

leecalcote: @Yashsharma1911 help! 😜

dragon-slayer875: @sudhanshutech I just checked.
It seems that Akshay made a mistake while making his migration to transferList components at other places.
As I stated, and as you have defined in sistent, the left permission is used to move components to left side, which is the remove action.

PR #11156, opened 2024-06-12
workflow update to mail error log w/signoff

Jougan-0:
**Notes for Reviewers**

Update the workflow to send error logs whenever an error occurs while running mesheryctl registry generate. Additionally, create a new file to store only the errors.

Local run of the [workflow](https://github.com/meshery/meshery/pull/11156).


mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -417,6 +425,16 @@ func logModelGenerationSummary(modelToCompGenerateTracker *store.GenerticThreadS
 	utils.Log.Info(fmt.Sprintf("-----------------------------\n-----------------------------\nGenerated %d models and %d components", totalAggregateModel, totalAggregateComponents))
 }
 
+func logError(err error, modelName string) {
```
MUzairS15, line 428: No need.
Use log instance,  refer the existing approach which logs its output to registry log file.
Create a new log instance for errors logs when an error occurs use that log instance. As you initialize that instance set its log output to registry-generate-error.

Also make the directory structured, if you would.
`logs/registry/1. log (contains logs of al severity except error logs) 2. `error (contains only error logs)`

MUzairS15 (changes requested):

mesheryctl/internal/cli/root/registry/generate.go
Jougan-0: Will do so positively.

leecalcote (commented):
Doe this new email replace the existing email for this workflow or will it be a duplicate?

Jougan-0:
> Does this new email replace the existing email for this workflow or will it be a duplicate?

This will replace the component generator failure email in these scenarios:

- >     If the component generator job fails, the old failure email will be used.
- >     If the component generator is successful but the error log is not empty, a new email regarding the error logs will be sent, and the workflow will be marked as successful.

For the scenario if the generator works and updater fails.
- >   We will have two emails: one for the failure of the updater and one for the error logs.

leecalcote (approved):
LGTM

PR #11267, opened 2024-06-28
Extentsion to logs

Jougan-0:
**Notes for Reviewers**

This PR addresses the issue where a 404 "Page Not Found" error occurs when clicking on "View" or "Download" in the event metadata, specifically when not on the root page.

Additionally, this PR includes the addition of a file extension to the logs file.


mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -79,14 +79,14 @@ mesheryctl registry generate --registrant-def [path to connection definition] --
 			return ErrUpdateRegistry(err, modelLocation)
 		}
 		utils.Log.SetLevel(logrus.DebugLevel)
-		logFilePath := filepath.Join(logDirPath, "Logs")
+		logFilePath := filepath.Join(logDirPath, "Logs.txt")
```
leecalcote, line 82:
```suggestion
		logFilePath := filepath.Join(logDirPath, "model-generation.log")
```

Are these model-generation or registry-import logs? Either way, let's name accordingly. Please

leecalcote: It's it's both, then `registry.log` will work.

mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -79,14 +79,14 @@ mesheryctl registry generate --registrant-def [path to connection definition] --
...
 		logFile, err = os.Create(logFilePath)
 		if err != nil {
 			return err
 		}
 
 		utils.LogError.SetLevel(logrus.ErrorLevel)
-		logErrorFilePath := filepath.Join(logDirPath, "Errors")
+		logErrorFilePath := filepath.Join(logDirPath, "Errors.txt")
```
leecalcote, line 89:
```suggestion
		logErrorFilePath := filepath.Join(logDirPath, "registry-errors.log")
```

mesheryctl/internal/cli/root/registry/generate.go
Jougan-0: These are the logs of `Model component Generator`. In the action we only want to mail the `error` so it's better we have 2 separate logs one that shows the error and the other that stores the general logs.

leecalcote: Will you add a note here highlighting how thoughtful we have been and how users can take note that a missing `model-generation-errors.log` is a good thing, but that it and the `model-generation.log` files can be referenced? Explain how they are used, please.

PR #11295, opened 2024-07-02
Fix pattern and filter API for local provider

shubham251972:
**Notes for Reviewers**

This PR fixes #

`GetMesheryPatterns` and `GetMesheryFilters` function was not returning any data despite the database being properly seeded. This issue was due to incorrect initialization and conditional logic in the query construction.

Solutions

1. `query` variable is now always initialized with mpp.DB.Table("meshery_patterns"), ensuring it is never nil.
1. Visibility condition is properly checked and added to the query only when visibility is not empty.
1. Counting of results is done on the fully constructed query, and pagination is correctly applied to the final query.

### DIFF
--- a/server/models/meshery_filter_persister.go
+++ b/server/models/meshery_filter_persister.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/gofrs/uuid"
 	"github.com/layer5io/meshkit/database"
-	"gorm.io/gorm"
 )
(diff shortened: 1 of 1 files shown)


aabidsofi19 (dismissed):

Yashsharma1911 (approved):
Thankyou gents

PR #11309, opened 2024-07-03
Run regal linter against WASM policy and fix errors

ctcarrier:
**Notes for Reviewers**

This PR fixes lint errors returned by the Regal linter for the WASM policy. This is in preparation for integrating the Regal linter into the build process to enforce policies pass static analysis.

### DIFF
--- a/policies/wasm/policies/namespace_discovery_relationship_policy/namespace_discovery_relationship_policy.rego
+++ b/policies/wasm/policies/namespace_discovery_relationship_policy/namespace_discovery_relationship_policy.rego
@@ -1,62 +1,35 @@
-# https://play.openpolicyagent.org/p/wFNhyGsIej
 package hierarchical_policy
 
-available_namespaces[namespace_key] {
-  service := input.services[_];
-  namespace_key = service.namespace
+import rego.v1
+
+available_namespaces contains service.namespace if {
+	some service in input.services
 }
 
-available_namespaces[namespace_key] {
-  service := input.services[_];
-  service.type == "Namespace";
-  namespace_key = service.name
+available_namespaces contains service.name if {
+	some service in input.services
+	service.type == "Namespace"
 }
 
-parent_child_mapping = { namespace_map: ns |
-	some namespace_map
-	available_namespaces[namespace_map]
-    service := input.services[key]
-    service.namespace
-    service.namespace == namespace_map
-    ns := {ns_serv: comp_id |
-        some ns_serv
-         x := input.services[ns_serv]
-         x.namespace == namespace_map
-         comp_id := x.traits.meshmap.id
-    }
+parent_child_mapping[namespace_map] := ns if {
+	some namespace_map in available_namespaces
+	some key, service in input.services
+	service.namespace == namespace_map
+	ns := {ns_serv: comp_id |
+		some ns_serv, x in input.services
+		x.namespace == namespace_map
+		comp_id := x.traits.meshmap.id
+	}
 }
 
-namespaces_to_create[namespaces2] {
-	some namespace
-    available_namespaces[namespace]
-    print(namespace)
-    ns_creation_status = is_present(check_namespace_present_status(namespace))
-    print(ns_creation_status)
-    ns_creation_status != true
-    namespaces2 = namespace
+namespaces_to_create contains namespace if {
+	some namespace in available_namespaces
+	not namespace_present(namespace, input.services)
 }
 
 # incase of present: {"present": true}, in absent, returns empty set
-check_namespace_present_status(ns) = is_present {
-  is_present := {"present": is |
-    some svc
-    s := input.services[svc]
-    s.type == "Namespace"
-    s.name == ns
-    is = true
-  }
+namespace_present(ns, all_services) if {
+	some s in all_services
+	s.type == "Namespace"
+	s.name == ns
 }
-
-# Is present function wraps the result of check_namespace_present_status to true or false
-is_present(obj) {
-	obj.present == true
-}
-
-is_present(obj) = pre {
- not obj.present
- pre = false
-}
-
-
-
-


policies/wasm/policies/namespace_discovery_relationship_policy/namespace_discovery_relationship_policy.rego
```diff
@@ -1,62 +1,54 @@
 # https://play.openpolicyagent.org/p/wFNhyGsIej
 package hierarchical_policy
 
-available_namespaces[namespace_key] {
-  service := input.services[_];
-  namespace_key = service.namespace
+import rego.v1
+
+available_namespaces[namespace_key] if {
```
anderseknert, line 6: While this is valid Rego v1, it changes the semantics from the previous version. 

The Rego v1 equivalence of this:
```rego
available_namespaces[namespace_key] {
```
Would be this:
```rego
available_namespaces contains namespace_key if {
```
And what you've got here, i.e this:
```rego
available_namespaces[namespace_key] if {
```
would in Rego v1 mean "a map with a `namespace_key` and an implicit value of `true`. 

A good idea is to use `opa fmt --rego-v1` to have the file automatically formatted for Rego v1 compliance, and then start linting with Regal from there.

policies/wasm/policies/namespace_discovery_relationship_policy/namespace_discovery_relationship_policy.rego
```diff
@@ -1,62 +1,54 @@
...
-    ns := {ns_serv: comp_id |
-        some ns_serv
-         x := input.services[ns_serv]
-         x.namespace == namespace_map
-         comp_id := x.traits.meshmap.id
-    }
+parent_child_mapping[namespace_map] := ns if {
+	some namespace_map in available_namespaces
+	some key
+	service := input.services[key]
```
anderseknert, line 20: This is fine, but:
```rego
some key
service := input.services[key]
```
could just as well be rewritten as:
```rego
some key, service in input.services
```
and perhaps it'd be more consistent

anderseknert (commented):
Hi there! Very happy to see you working on Regal compliance here 😃 

It looks like a few things have been mistranslated along the way to Rego v1, and I've left a few comments for the things that you'll want to address. 

In addition to `regal lint`, you'll want to run `opa check` on the file to see that it compiles as well.

Happy to re-review later if you ping me!

leecalcote:
Thank you for this, @anderseknert! 👏

ctcarrier:
Thanks for the feedback @anderseknert that helped point me in the right direction.

policies/wasm/policies/namespace_discovery_relationship_policy/namespace_discovery_relationship_policy.rego
```diff
@@ -1,62 +1,39 @@
...
+namespaces_to_create contains namespace if {
+	some namespace in available_namespaces
+	ns_creation_status := check_namespace_present_status(namespace, input.services)
+	ns_creation_status != true
 }
 
-
-
-
+check_namespace_present_status(ns, all_services) := contains_ns if {
```
ctcarrier, line 34: My assumption based on the existing implementation is that "namespaces_to_create" should end up containing all the namespaces referenced within the input that does not have a matching namespace definition. I used this playground link to test with:

https://play.openpolicyagent.org/p/iS8qumDa1I

So in this example all the pods and services are in the "default" namespace and there is also a "nginx-1" Namespace defined in the list of services. So my assumption was that the "namespaces_to_create" should only contain "default" since there is no explicit "default" namespace defined in the list of services. The existing implementation returned both "default" and "nginx-1". I might be misunderstanding the intention.

policies/wasm/policies/namespace_discovery_relationship_policy/namespace_discovery_relationship_policy.rego
```diff
@@ -1,62 +1,39 @@
...
+	some svc, s in all_services
+	s.type == "Namespace"
+	s.name == ns
+	contains_ns := count(svc) > 0
+} else := false
```
ctcarrier, line 39: @anderseknert Is this the idiomatic way to implement "contains" type functionality on a list of objects? I tried a few different approaches and this is the first that worked as I expected.

policies/wasm/policies/namespace_discovery_relationship_policy/namespace_discovery_relationship_policy.rego
MUzairS15: Nope @anderseknert you are totally correct. It should only return diffs.

policies/wasm/policies/namespace_discovery_relationship_policy/namespace_discovery_relationship_policy.rego
```diff
@@ -1,62 +1,39 @@
...
 }
 
-
-
-
+check_namespace_present_status(ns, all_services) := contains_ns if {
+	some svc, s in all_services
+	s.type == "Namespace"
+	s.name == ns
+	contains_ns := count(svc) > 0
```
MUzairS15, line 38: @ctcarrier This might not be needed. As it isn't  the count of all services of type `Namespace` but the name of the service example in this case `nginx-1`

MUzairS15:
Given the fact, this particular policy and all policies under `wasm` directory are not used currently but created for prototyping use of WASM, we are not to worry about breaking changes much.

@anderseknert Thanks for your review and suggestions, it will really help us in updating policies which are currently in use. (stored under `/server/meshmodel/kubernetes` directory).

MUzairS15 (approved):
Except https://github.com/meshery/meshery/pull/11309/files#r1671063488
rest LGTM

leecalcote (approved):

Ashparshp:
@ctcarrier Could we discuss this issue during the Meshery dev meeting call? Please add it as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit?usp=drivesdk).

MUzairS15 (approved):

PR #11361, opened 2024-07-10
fix concurent write error w/signoff

Jougan-0:
**Notes for Reviewers**

This PR fixes #
Fix concurent write condition.
image

### DIFF
--- a/mesheryctl/internal/cli/root/model/import.go
+++ b/mesheryctl/internal/cli/root/model/import.go
@@ -175,13 +175,11 @@ func sendToAPI(data []byte, name string, dataType string) error {
 
 	var formFile io.Writer
 	if dataType == "file" {
-		formFile, err = writer.CreateFormFile("file", filepath.Base(name))
+		formFile, _ = writer.CreateFormFile("file", filepath.Base(name))
 	} else {
-		formFile, err = writer.CreateFormField("dir")
+		formFile, _ = writer.CreateFormField("dir")
 	}
-	if err != nil {
 
-	}
 	_, err = formFile.Write(data)
 	if err != nil {
 		err = meshkitutils.ErrWriteFile(err, name)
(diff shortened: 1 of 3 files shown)


Jougan-0:
//@dragon-slayer875 @Yashsharma1911

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -175,13 +175,11 @@ func sendToAPI(data []byte, name string, dataType string) error {
...
 	if dataType == "file" {
-		formFile, err = writer.CreateFormFile("file", filepath.Base(name))
+		formFile, _ = writer.CreateFormFile("file", filepath.Base(name))
 	} else {
-		formFile, err = writer.CreateFormField("dir")
+		formFile, _ = writer.CreateFormField("dir")
 	}
-	if err != nil {
 
-	}
```
Yashsharma1911, line 184: Why are we removing the error handling

server/helpers/common_registry_logs.go
```diff
@@ -70,12 +72,14 @@ func HandleError(h v1beta1.Host, en entity.Entity, err error, isModelError bool,
 		if err != nil {
 			handleModelOrRegistrantError(h, entity.Model.Name, err, isModelError, isRegistrantError)
 
+			LogHandler.RegisterAttempts[meshmodel.HostnameToPascalCase(h.Hostname)].mu.Lock()
```
Yashsharma1911, line 75: Looks good however the logic for v1beta1 and v1alpha two looks mostly similar, you can extract his logic to make sure we are not repeating, something for next PR

Yashsharma1911 (commented):
LGTM after this

mesheryctl/internal/cli/root/model/import.go
Jougan-0: In different PR added them.

server/helpers/common_registry_logs.go
Jougan-0: okay

mesheryctl/internal/cli/root/model/import.go
Jougan-0: When I write support for OCI should be done by today if I get the OCI from Nitish then I'll update this to support it too. Then only I would write the meshkit error for this.

Jougan-0: @Yashsharma1911

Issue #11352, opened 2024-07-09
[UI] Settings --> Registry: Model Import and Model Export


leecalcote:
Related to https://github.com/meshery/meshery/issues/9775

### Current Behavior
Meshery supports importing and exporting of individual models into an out of the registry in a few different formats, however, this functionality is only exposed in `mesheryctl model`, currently.

### Desired Behavior
The ability to import and export models using Meshery UI **- using the OCI format only.**

### Implementation
1. Add Import button in Settings --> Registry page to upload a model in OCI format.
1. Add Export button in Settings --> Registry page to export a model in OCI format.
---
Bonus:
3. Bulk export of more than one model at a time.

### Acceptance Tests
1. User can click "Import" to open a file browser and upload a model in OCI format to Meshery Server, which upon import, validates all of the models contained entities, registers valid entities, notifies the user with a toasted event  and one filed in the Notification Center that includes a summary of the number of entities successfully imported or details as to why the import failed and what to do about it (with MeshKit error reference)
2. User can select (checkmark) a single model name (that might contain one or more versions of the model). Click Export and have an OCI-formatted model file pushed to their client (web browser).

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](http://discuss.meshery.io) and [Community Slack](https://slack.meshery.io)

102vansh:
I would like to work on this issue. Could you please assign it to me

Ashparshp:
@102vansh Have you set up the repo?

102vansh:
I've completed setting up the repository

102vansh:
@Ashparshp I've completed setting up the repository

Ashparshp:
@102vansh Assigned to you! So, go ahead!

leecalcote:
@102vansh progress?

leecalcote:
@102vansh, I didn't hear back...

PR #11385, opened 2024-07-12
Refactor Entity Registration Logic 

humblenginr:
### Description

This PR refactors the Meshmodel entity registration logic to be schema driven, more manageable, and straightforward. 

The [`Register`](https://github.com/meshery/meshery/pull/11385/files#diff-a9565a7958209547380a383b747cc081d850db91f1e67633fe9b81e7eaba9fd7R39) function takes in a [`RegistrableEntity`](https://github.com/meshery/meshery/pull/11385/files#diff-b3ce7d7058a32bc5f15ae50917524e9b3832a48141f6c7796ba514b1de042fe3R4) (oci image, tar or a directory) as its input and starts registering the entities in it. It sets a requirement that the [`RegistrableEntity`](https://github.com/meshery/meshery/pull/11385/files#diff-b3ce7d7058a32bc5f15ae50917524e9b3832a48141f6c7796ba514b1de042fe3R4) should contain one and only one `ModelDefinition` and any number of `RelationshipDefinition`, `ComponentDefinition` or `PolicyDefinition`.

This PR makes `Models` the units of packaging for Meshery’s logical object representation. 

This PR modifies the logic to identify a Meshmodel entity to be [schema driven](https://github.com/meshery/meshery/pull/11385/files#diff-2c08829638b3e8aca48699c241f868611fede1f71e1523eb5ac924f307f31553R39) -  rather than depending on the filesystem structure. 

This PR modifies the logging procedure in registration process and error handling. The registration process in this PR can be seen as two steps:
1. Parse the `RegistrableEntity` and extract _valid_ Meshmodel definitions
2. Register the extracted Meshmodel definitions
3. Log failures and successes in the registration process

The first step will fail (and the registration process will break) if there is no `Model` found. If the 'Model' is found, It will register all the extracted valid Meshmodel definitions, ignoring the invalid definitions. These invalid definitions will be shown in the registry logs file after the registration process is over. 

When it comes to the second step, we are only attempting to register _valid_ Meshmodel definitions. So, the errors encountered in this step will not be of schema violation, but rather due to database and other Registry related issues. 

The [Logger](https://github.com/meshery/meshery/pull/11385/files#diff-3cc9e8b1413b5eacd706c24578a0a092246d5fd9035949e33a5175593d30c92dR22) now logs issues in two categories:
1. Invalid Definitions
2. Failures in registering valid Meshmodel definitions

### DIFF
--- a/docs/_data/mesheryctlcommands/cmds.yml
+++ b/docs/_data/mesheryctlcommands/cmds.yml
@@ -725,7 +725,7 @@ pattern:
           description: Specify the output directory to save the design
           usage: mesheryctl pattern export --output [directory-path]
           example: mesheryctl pattern export --output ./exports
-          
+
     offboard:
       name: offboard
       description: Offboard application
@@ -914,6 +914,14 @@ model:
       examples:
          - import model /path/to/[file.yaml|file.json]
          - import model /path/to/models
+    export:
+      name: export
+      description: Export a model by name to the given output type (default is oci).
+      usage: mesheryctl model export [ model-name ]
+      examples:
+         - model export [model-name] -l /home/meshery
+         - model export [model-name] --discard-components
+         - model export [model-name] -o yaml
 
 components:
   name: components
@@ -960,7 +968,7 @@ exp:
     name: components
     description: View list of components and detail of components
     usage: mesheryctl exp components [command] [flags]
-    flags: 
+    flags:
       count:
         name: --count
         description: (optional) Get the number of exp components in total
@@ -996,7 +1004,7 @@ exp:
     name: connections
     description: Manage connections
     usage: mesheryctl exp connections [command] [flags]
-    flags: 
+    flags:
       help:
         name: --help, -h
         description: help for exp connections
@@ -1024,12 +1032,12 @@ exp:
     name: environment
     description: View list of environments and detail of environments
     usage: mesheryctl exp environment [command] [flags]
-    flags: 
+    flags:
     subcommands:
       create:
         name: create
         description: create a new environments by providing the name and description of the environment
-        usage: mesheryctl exp environment create [flags] 
+        usage: mesheryctl exp environment create [flags]
         example: mesheryctl
         flags:
           description:
@@ -1058,7 +1066,7 @@ exp:
       view:
         name: view
         description: view a environments registered in Meshery Server
-        usage: mesheryctl exp environment view [flags] 
+        usage: mesheryctl exp environment view [flags]
         flags:
           output-format:
             name: -o, --output-format
@@ -1072,7 +1080,7 @@ exp:
     name: relationship
     description: View list of relationships and details of relationship
     usage: mesheryctl exp relationship [command] [flags]
-    flags: 
+    flags:
     subcommands:
       generate:
         name: generate
@@ -1093,7 +1101,7 @@ exp:
             usage: mesheryctl exp relationship list --page [page number]
             example: mesheryctl exp relationship list --page 2
       view:
-        name: view 
+        name: view
         description: view a relationship queried by the model name
         usage: mesheryctl exp relationship view [flags]
         flags:
@@ -1104,7 +1112,7 @@ exp:
     name: workspace
     description: View list of workspaces and detail of workspaces
     usage: mesheryctl exp workspace [command] [flags]
-    flags: 
+    flags:
     subcommands:
       create:
         name: create
--- a/docs/pages/reference/mesheryctl-commands.md
+++ b/docs/pages/reference/mesheryctl-commands.md
@@ -769,7 +769,7 @@ Installation, troubleshooting and debugging of Meshery and its adapters
   </tr>
   {% assign command12 = site.data.mesheryctlcommands.cmds.model %}
     <tr>
-      <td rowspan=9><a href="{{ site.baseurl }}/reference/mesheryctl/{{ command12.name }}">{{ command12.name }}</a></td>
+      <td rowspan=10><a href="{{ site.baseurl }}/reference/mesheryctl/{{ command12.name }}">{{ command12.name }}</a></td>
       <td></td>
       <td></td>
       <td>{{ command12.description }}</td>
--- a/docs/pages/reference/mesheryctl/mesheryctl-model-export.md
+++ b/docs/pages/reference/mesheryctl/mesheryctl-model-export.md
@@ -29,28 +29,21 @@ mesheryctl model export [flags]
 Export a model by name
 <pre class='codeblock-pre'>
 <div class='codeblock'>
-mesheryctl model export <modelname> -o json
+mesheryctl model export [model-name] -o [oci/json/yaml] (default is oci)
 
 </div>
 </pre> 
 
 <pre class='codeblock-pre'>
 <div class='codeblock'>
-mesheryctl model export <modelname> -o yaml
+mesheryctl model export [model-name] -l /home/meshery/
 
 </div>
 </pre> 
 
 <pre class='codeblock-pre'>
 <div class='codeblock'>
-mesheryctl model export <modelname> -l /home/meshery/
-
-</div>
-</pre> 
-
-<pre class='codeblock-pre'>
-<div class='codeblock'>
-mesheryctl model export <modelname> --include-components true --include-relationships true
+mesheryctl model export [model-name] --discard-components --discard-relationships
 
 </div>
 </pre> 
@@ -66,11 +59,11 @@ mesheryctl model export <modelname> --include-components true --include-relation
 
 <pre class='codeblock-pre'>
 <div class='codeblock'>
+  -c, --discard-components       (optional) whether to discard components in the exported model definition (default = false)
+  -r, --discard-relationships    (optional) whether to discard relationships in the exported model definition (default = false)
   -h, --help                     help for export
-  -c, --include-components       whether to include components in the model definition (default = false)
-  -r, --include-relationships    whether to include components in the model definition (default = false)
   -l, --output-location string   (optional) output location (default = current directory) (default "./")
-  -o, --output-type string       (optional) format to display in [json|yaml] (default = yaml) (default "yaml")
+  -o, --output-type string       (optional) format to display in [oci|json|yaml] (default = oci) (default "oci")
 
 </div>
 </pre>
--- a/mesheryctl/helpers/component_info.json
+++ b/mesheryctl/helpers/component_info.json
@@ -1,5 +1,5 @@
 {
   "name": "mesheryctl",
   "type": "client",
-  "next_error_code": 1127
+  "next_error_code": 1128
 }
--- a/mesheryctl/internal/cli/root/error.go
+++ b/mesheryctl/internal/cli/root/error.go
@@ -55,3 +55,4 @@ func ErrGettingRequestContext(err error) error {
 func ErrUnmarshallingAPIData(err error) error {
 	return errors.New(ErrUnmarshallingAPIDataCode, errors.Fatal, []string{"Error processing json API data"}, []string{"Error processing json API data", err.Error()}, []string{}, []string{})
 }
+
--- a/mesheryctl/internal/cli/root/model/error.go
+++ b/mesheryctl/internal/cli/root/model/error.go
@@ -1,0 +1,15 @@
+package model
+
+import (
+	"fmt"
+
+	"github.com/layer5io/meshkit/errors"
+)
+
+const (
+	ErrExportModelCode = "mesheryctl-1127"
+)
+
+func ErrExportModel(err error, name string) error {
+	return errors.New(ErrExportModelCode, errors.Fatal, []string{"Error exporting model"}, []string{fmt.Sprintf("Given model with name: %s could not be exported: %s", name, err)}, []string{"Model may not be present in the registry"}, []string{"Ensure that there are no typos in the model name"})
+}
--- a/mesheryctl/internal/cli/root/model/export.go
+++ b/mesheryctl/internal/cli/root/model/export.go
@@ -17,10 +17,9 @@ var exportModal = &cobra.Command{
 	Long:  "export the registered model to the specified output type",
 	Example: `
 // Export a model by name
-mesheryctl model export <modelname> -o json
-mesheryctl model export <modelname> -o yaml
-mesheryctl model export <modelname> -l /home/meshery/
-mesheryctl model export <modelname> --include-components true --include-relationships true
+mesheryctl model export [model-name] -o [oci/json/yaml] (default is oci)
+mesheryctl model export [model-name] -l /home/meshery/
+mesheryctl model export [model-name] --discard-components --discard-relationships
     `,
 	PreRunE: func(cmd *cobra.Command, args []string) error {
 		//Check prerequisite
@@ -56,7 +55,7 @@ mesheryctl model export <modelname> --include-components true --include-relation
 		}
 		baseUrl := mctlCfg.GetBaseMesheryURL()
 		modelName := args[0]
-		url := fmt.Sprintf("%s/api/meshmodels/models/%s?components=%t&relationships=%t&%s", baseUrl, modelName, includeCompsFlag, includeRelsFlag, utils.GetPageQueryParameter(cmd, pageNumberFlag))
+		url := fmt.Sprintf("%s/api/meshmodels/models/%s?components=%t&relationships=%t&%s", baseUrl, modelName, !discardComponentsFlag, !discardRelationshipsFlag, utils.GetPageQueryParameter(cmd, pageNumberFlag))
 		return exportModel(args[0], cmd, url, false)
 	},
 }
--- a/mesheryctl/internal/cli/root/model/model.go
+++ b/mesheryctl/internal/cli/root/model/model.go
@@ -46,10 +46,10 @@ var (
 	outLocationFlag string
 	// flag used to specify format of output of export {model-name} command
 	outTypeFlag string
-	// flag used to specify whether to include components in the model
-	includeCompsFlag bool
-	// flag used to specify whether to include relationships in the model
-	includeRelsFlag bool
+	// flag used to specify whether to discard components in the model
+	discardComponentsFlag bool
+	// flag used to specify whether to discard relationships in the model
+	discardRelationshipsFlag bool
 
 	// Maximum number of rows to be displayed in a page
 	maxRowsPerPage = 25
@@ -147,9 +147,9 @@ func init() {
 	viewModelCmd.Flags().StringVarP(&outFormatFlag, "output-format", "o", "yaml", "(optional) format to display in [json|yaml]")
 
 	exportModal.Flags().StringVarP(&outLocationFlag, "output-location", "l", "./", "(optional) output location (default = current directory)")
-	exportModal.Flags().StringVarP(&outTypeFlag, "output-type", "o", "yaml", "(optional) format to display in [json|yaml] (default = yaml)")
-	exportModal.Flags().BoolVarP(&includeCompsFlag, "include-components", "c", false, "whether to include components in the model definition (default = false)")
-	exportModal.Flags().BoolVarP(&includeRelsFlag, "include-relationships", "r", false, "whether to include components in the model definition (default = false)")
+	exportModal.Flags().StringVarP(&outTypeFlag, "output-type", "o", "oci", "(optional) format to display in [oci|json|yaml] (default = oci)")
+	exportModal.Flags().BoolVarP(&discardComponentsFlag, "discard-components", "c", false, "(optional) whether to discard components in the exported model definition (default = false)")
+	exportModal.Flags().BoolVarP(&discardRelationshipsFlag, "discard-relationships", "r", false, "(optional) whether to discard relationships in the exported model definition (default = false)")
 
 	ModelCmd.AddCommand(availableSubcommands...)
 	ModelCmd.Flags().BoolVarP(&countFlag, "count", "", false, "(optional) Get the number of models in total")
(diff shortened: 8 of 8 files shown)


leecalcote (changes requested):
Build check failing.

MUzairS15 (commented):
@humblenginr The logic inside the `RegisterEntity` for checking whether model exists if not register the model needs to be removed.

Instead before registering an entity (component and relationship) if model is not registered we should error out.
https://github.com/meshery/meshkit/blob/fd9565f3bbbfb6731c25726714029913f82b3743/models/meshmodel/core/v1beta1/component.go#L84

humblenginr:
> @humblenginr The logic inside the `RegisterEntity` for checking whether model exists if not register the model needs to be removed.
> 
> 
> 
> Instead before registering an entity (component and relationship) if model is not registered we should error out.
> 
> https://github.com/meshery/meshkit/blob/fd9565f3bbbfb6731c25726714029913f82b3743/models/meshmodel/core/v1beta1/component.go#L84

Yes, agreed. I'll update that in the meshkit pr. The registration process in this PR is such that it calls 'RegisterEntity' only after it's ensured the presence of 'Model'.

server/helpers/common_registry_logs.go
```diff
@@ -65,6 +65,8 @@ func HandleError(h v1beta1.Host, en entity.Entity, err error, isModelError bool,
 	switch entity := en.(type) {
 	case *v1beta1.ComponentDefinition:
 		entityName := "[ " + entity.Model.Name + " " + entity.Model.Model.Version + " ]" + "( " + entity.DisplayName + " )"
+		//TODO:  Move this outside this function - this function is supposed to handle errrors, not check
```
leecalcote, line 68: I wonder if during the time that this is under review whether we can tie-off this to-do item before merge.

server/meshmodel/helper.go
```diff
@@ -1,229 +1,65 @@
...
-	})
-	if err != nil {
-		erh.errorChan <- mutils.ErrReadDir(errors.Wrapf(err, "error while generating relationships"), pathToComponents)
+// seed the local meshmodel components
+func SeedComponents(log logger.Handler, hc *models.HandlerConfig, regm *meshmodel.RegistryManager) {
+	regHelper := reg.NewRegistrationHelper(log, hc, regm)
+	modelDirPaths, err := GetModelDirectoryPaths()
+	if(err != nil){
+		// handle error
+		fmt.Println(err)
```
leecalcote, line 58: Needs replaced.

leecalcote (changes requested):

server/models/meshmodel/registration/dir.go
```diff
@@ -0,0 +1,89 @@
...
+/*
+	The directory should contain one and only one `model`.
+	A directory containing multiple `model` will be invalid.
+*/
+func NewDir(path string) Dir {
+	return Dir{dirpath: path}
+}
+
+
+/* PkgUnit parses all the files inside the directory and finds out if they are any valid meshery definitions. Valid meshery definitions are added to the packagingUnit struct.
```
leecalcote, line 27: nice!

server/models/meshmodel/registration/dir.go
```diff
@@ -0,0 +1,89 @@
...
+	err = filepath.Walk(d.dirpath, func (path string, f os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if(f.IsDir()){
+				return nil
+		}
+		byt, _ := os.ReadFile(path)
+		var e entity.Entity
+		if(filepath.Ext(path) == ".yaml"){
```
leecalcote, line 45: Why? Is ".yml" not valid?

Why is the file extension a concern at all? Unless there's an identifiable justification, this approach is fragile and over-optimized.

leecalcote (changes requested):

server/helpers/common_registry_logs.go
humblenginr: That's fixed @leecalcote . We are not using that function anymore. I just forgot to remove that. I'll remove it.

server/models/meshmodel/registration/error.go
```diff
@@ -0,0 +1,31 @@
...
+func ErrGetEntity(err error) error {
+	return errors.New(ErrGetEntityCode, errors.Alert, []string{err.Error()}, []string{err.Error()}, []string{}, []string{})
+}
+
+func ErrDirPkgUnitParseFail(dirpath string, err error) error {
+	return errors.New(ErrDirPkgUnitParseFailCode, errors.Alert, []string{fmt.Sprintf("Directory %s is not a valid packaging unit for registration", dirpath)}, []string{fmt.Sprintf("Could not parse the given directory %s into a packaging unit:  %s", dirpath, err.Error())}, []string{}, []string{} )
+}
+
+func ErrInvalidMeshmodelDefinition(path string, err error) error {
+	return errors.New(ErrDirPkgUnitParseFailCode, errors.Alert, []string{fmt.Sprintf("The file at path %s is not a valid meshmodel definition", path)}, []string{fmt.Sprintf("Could not parse meshmodel entity:  %s", err.Error())}, []string{}, []string{} )
```
leecalcote, line 30: We're trying to move away from the term "MeshModel". Please don't perpetuate its use. Please eradicate any existing references as you come across them. 

Yes, /api/meshmodel needs to be renamed to /api/registry.

leecalcote (changes requested):

server/models/meshmodel/registration/error.go
leecalcote: This comment holds true and is relevant to the MeshModel references in the description of this PR, too.

leecalcote (changes requested):
Documentation update is missing:

- https://docs.meshery.io/concepts/logical/registry
- https://docs.meshery.io/concepts/logical/models#models-as-the-unit-of-packaging (maybe)

server/helpers/common_registry_logs.go
leecalcote: Oh, very good.

leecalcote:
Is support for Models as OCI images to be a follow-on PR? Delivery of https://github.com/meshery/meshery/issues/11352 needs OCI import --> registration support as a prerequisite.

server/models/meshmodel/registration/dir.go
```diff
@@ -0,0 +1,89 @@
...
+				rel, _ := e.(*v1alpha2.RelationshipDefinition)
+				pkg.relationships = append(pkg.relationships, *rel)
+		}
+		return nil
+	})
+	if err != nil {
+		return pkg, ErrDirPkgUnitParseFail(d.dirpath, fmt.Errorf("Could not completely walk the file tree: %e", err))
+	}
+	if (reflect.ValueOf(pkg.model).IsZero()){
+		err := fmt.Errorf("Cannot find `model` in the directory. Directory without a `model` definition is invalid.")
```
leecalcote, line 84: Not "directory", but "model package".

Here's a suggestion that needs a little tweak at ".....":

"Model definition not found in imported package. Model definitions often use the filename "model.json", but are not required to have this filename. One and exactly one entity containing `schema: model.core....... ......` must be present, otherwise the model package is considered malformed."

Are you including probable causes and suggested remediations, @humblenginr?

leecalcote (changes requested):

server/models/meshmodel/registration/dir.go
humblenginr: Yes, I thought of using the syntax to identify the file type. But I thought depending on the error type from the `json` and `yaml` packages would be fragile as they often have less detail. Look at this - https://stackoverflow.com/questions/68156758/robust-way-check-if-the-input-is-json-or-yaml , the first answer gives me the impression that it might be fragile, especially with YAML files.

I am not feeling good about this approach either, but I thought this would not be an issue, because at this stage, we already know that we are dealing with a directory, and all the standard parsers check for file extension as the first check before checking anything else.

server/models/meshmodel/registration/error.go
humblenginr: Got it, will change this.

server/models/meshmodel/registration/dir.go
humblenginr: A much better error message 😄 . I will update this. Yes, I will be adding probable causes and suggested remediations.

humblenginr:
> Is support for Models as OCI images to be a follow-on PR? Delivery of #11352 needs OCI import --> registration support as a prerequisite.

Yes @leecalcote . 10 lines of code, which have added in the Meshery UI Import Export PR.

server/models/meshmodel/registration/dir.go
humblenginr: I just am not entirely confident with the alternate approach.

leecalcote: Call me when you have stable Internet and I'll explain why figuring out whether it's json or yaml upfront, simply doesn't matter.

leecalcote: Just attempt to parse the file. Apply a JSON validator, if valid continue. If JSON fail, apply a YAML validator. If good, continue. If fail, exit.

The filename extension is indicative of which validator to run, and as an optimization, we can run the validator appropriate to the filename extension first, however, ultimately, we’re going to have to try to parse the file whether it has a filename extension or not.

Ashparshp:
@humblenginr Could we discuss this issue during the Meshery dev meeting call? Please add it as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit?usp=drivesdk).

leecalcote:
Re-reviewing... I don't see meaningful changes to review, yet.

leecalcote:
Stopping by to review again. No new commits to review.

humblenginr:
Will update today

humblenginr:
Updated

leecalcote:
@humblenginr respond to and resolve each comment made. Describe what was done to incorporate the feedback or why it was rejected.

leecalcote:
> @humblenginr respond to and resolve each comment made. Describe what was done to incorporate the feedback or why it was rejected.

I see new commits, but don't see diligence exercised, conducive to another review, @humblenginr.

server/models/meshmodel/registration/dir.go
leecalcote: @Yashs911 @Jougan-0 fyi

humblenginr: https://github.com/meshery/meshery/pull/11385/files#diff-2c08829638b3e8aca48699c241f868611fede1f71e1523eb5ac924f307f31553R13 - a generalised `unmarshal` function has been introduced, which unmarshals the entity irrespective of its file type.

server/models/meshmodel/registration/error.go
humblenginr: Error message has been changed

server/models/meshmodel/registration/dir.go
humblenginr: The registration logic has been updated to make use of this function without checking for filetypes.

server/meshmodel/helper.go
humblenginr: Has been replaced with `ErrSeedComponents`

humblenginr:
The comments got outdated, so I couldn't see them. Resolved them now.

humblenginr:
> Is support for Models as OCI images to be a follow-on PR? Delivery of #11352 needs OCI import --> registration support as a prerequisite.

Yes, it is a pre-requisite for #11352, and will be implemented in another PR

server/models/meshmodel/registration/error.go
humblenginr: Also, other instances of the word `MeshModel` has been changed to `Model`

server/helpers/common_registry_logs.go
humblenginr: Removed the comments

leecalcote:
>The [Logger](https://github.com/meshery/meshery/pull/11385/files#diff-3cc9e8b1413b5eacd706c24578a0a092246d5fd9035949e33a5175593d30c92dR22) now logs issues in two categories:

> 1. Invalid Definitions
> 1. Failures in registering valid Meshmodel definitions

While these aren't the only two failure cases, I understand that at a high-level failures (largely) fall between these two categories (another being definitions not available as an example). @Jougan-0, are you onboard with this Logger update?

humblenginr:
Yes, these are not the only failure cases. But I made that statement in context of the current registration procedure:

> The first step will fail (and the registration process will break) if there is no Model found. If the 'Model' is found, It will register all the extracted valid Meshmodel definitions, ignoring the invalid definitions. These invalid definitions will be shown in the registry logs file after the registration process is over.   <br> <br>When it comes to the second step, we are only attempting to register valid Meshmodel definitions. So, the errors encountered in this step will not be of schema violation, but rather due to database and other Registry related issues.

Jougan-0:
> > The [Logger](https://github.com/meshery/meshery/pull/11385/files#diff-3cc9e8b1413b5eacd706c24578a0a092246d5fd9035949e33a5175593d30c92dR22) now logs issues in two categories:
> 
> > 1. Invalid Definitions
> > 2. Failures in registering valid Meshmodel definitions
> 
> While these aren't the only two failure cases, I understand that at a high-level failures (largely) fall between these two categories (another being definitions not available as an example). @Jougan-0, are you onboard with this Logger update?

Yes, it makes sense as he mentioned that he mentioned 2 errors following the flow the logger will log different errors as well. Am I right @humblenginr ?

server/models/meshmodel/registration/logs.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return "", err
+	}
+	if failedMsg != "" {
+		filePath := viper.GetString("REGISTRY_LOG_FILE")
+		errorEventBuilder := events.NewEvent().FromUser(mesheryInstanceID).FromSystem(mesheryInstanceID).WithCategory("registration").WithAction("get_summary")
+		errorEventBuilder.WithSeverity(events.Error).WithDescription(failedMsg)
+		errorEvent := errorEventBuilder.Build()
+		errorEventBuilder.WithMetadata(map[string]interface{}{
+			"LongDescription":      fmt.Sprintf("The import process for a registrant %s encountered difficulties,due to which %s. Specific issues during the import process resulted in certain entities not being successfully registered in the table.", hostname, failedMsg),
```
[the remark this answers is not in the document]
humblenginr: Replaced this with a meshkit error

Jougan-0:
@humblenginr merge conflicts.

l5io:
🚀 Preview for commit f2f6c80e56ecbd480996034e86020b99b20ad48d at: https://66a8fbb859756a2ead6c89c0--meshery-docs-preview.netlify.app

server/meshmodel/error.go
```diff
@@ -0,0 +1,14 @@
+package meshmodel
```
leecalcote, line 1: @humblenginr can we be rid of this package name and use `registry` instead?

humblenginr: yes

server/models/meshmodel/registration/register.go
```diff
@@ -0,0 +1,157 @@
...
+
+
+/*
+	register will return an error if it is not able to register the `model`.
+	If there are errors when registering other entities, they are handled properly but does not stop the registration process.
+*/
+func (rh *RegistrationHelper)register(pkg packagingUnit) {
+	// 1. Register the model
+	model := pkg.model
+	if(model.Status != entity.Enabled){return}
```
leecalcote, line 58: Revisit, please.

server/models/meshmodel/registration/register.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// Dont register anything else if registrant is not there
+	if(model.Registrant.Hostname == ""){
```
leecalcote, line 61: Shift to a Connection.

server/models/meshmodel/registration/register.go
```diff
@@ -0,0 +1,157 @@
...
+		RegLog.InsertEntityRegFailure(model.Registrant.Hostname, "",entity.Model, model.Name, err)
+		return
+	}
+
+	hostname := model.Registrant.Hostname
+	modelName := model.Name
+	// 2. Register components
+	for _, comp := range pkg.components {
+		comp.Model = model
+		if(comp.Model.Status != entity.Enabled){
```
leecalcote, line 83: Good idea. Revisit, please. For now, allow registration of disabled models and components.

server/models/meshmodel/registration/register.go
```diff
@@ -0,0 +1,157 @@
...
+
+	sysID := uuid.FromStringOrNil(systemID)
+	hosts, _, err := erh.regManager.GetRegistrants(&v1beta1.HostFilter{})
+	if err != nil {
+		log.Error(err)
+	}
+
+	for _, host := range hosts {
+		eventBuilder := events.NewEvent().FromSystem(sysID).FromUser(sysID).WithCategory("entity").WithAction("get_summary")
+		successMessage := fmt.Sprintf("For registrant %s successfully imported", host.Hostname)
```
leecalcote, line 125:
```suggestion
		successMessage := fmt.Sprintf("For registrant %s imported", host.Hostname)
```

How many imported? -- -- see the predefined error message in MeshKit.

server/models/meshmodel/registration/utils.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	return nil
+}
+
+// TODO: refactor this and use CUE
+func getEntity(byt []byte) (et entity.Entity, _ error) {
+    var versionMeta v1beta1.VersionMeta
+	err  := unmarshal(byt, &versionMeta)
+	if err != nil || versionMeta.SchemaVersion == "" {
+		return nil, ErrGetEntity(fmt.Errorf("Does not contain versionmeta"))
```
leecalcote, line 29: versionmeta?

MUzairS15 (commented):
1. We should ensure all entities are registered, and model APIs by default only return entities having `status` as `enabled`.
2. It will be nice if we can discern whether the input type is directory/oci/targz/zip by using the meshkit functionality (OCI detection needs to be added)

Both YAML and JSON fall under the content type plain/text. We already have the ability to detect this plaintext type. The newly added unmarshall function can be used to decode them afterward.

We can also replace the existing `unmarshall` function in  meshkit with this new one.

server/models/registration/utils.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/layer5io/meshkit/models/meshmodel/entity"
+	"gopkg.in/yaml.v2"
+)
+
+func unmarshal(byt []byte, out interface{}) error {
+	err := json.Unmarshal(byt, out)
+	if(err != nil){
+		err = yaml.Unmarshal(byt, out)
+		if(err != nil){
+			return fmt.Errorf("Not a valid YAML or JSON")
```
MUzairS15, line 18: Please replace with meshkit error.

server/models/meshmodel/registration/utils.go
humblenginr: This `versionMeta` is just an internal struct that will not be propagated to the user. This is just a helper struct that holds `schemaVersion` and `version`.

humblenginr:
Made some changes:
1. Moved the `registration` package out of `meshmodel`. It is now present at `server/models/registration`
2. Refined the errors 
3. Register disabled models and components as well

server/models/meshmodel/registration/register.go
humblenginr: Created an issue.

server/models/meshmodel/registration/register.go
humblenginr: Removed it.

server/models/registration/dir.go
```diff
@@ -0,0 +1,80 @@
...
+			return nil
+		}
+		// set it to pkgunit
+		switch e.Type() {
+			case entity.Model:
+				if !reflect.ValueOf(pkg.model).IsZero() {
+					// currently models inside models are not handled
+					return nil
+				}
+				model, _ := e.(*v1beta1.Model)
```
MUzairS15, line 60: This means the entities that are corrupted are not logged, I think we should track such failures as well.

Better to use `utils.Cast`.

server/meshmodel/error.go
humblenginr: Moved this whole thing into `registration` package

server/models/registration/error.go
```diff
@@ -0,0 +1,82 @@
...
+	)
+}
+
+func ErrMissingRegistrant() error {
+	return errors.New(
+		ErrMissingRegistrantCode,
+		errors.Alert,
+		[]string{"Meshery models are always registered in context of a registrant"},
+		[]string{""},
+		[]string{"Given model might not contain information about the registrant"},
```
MUzairS15, line 35: We can include the modelName so that we are specific about which model lacks the registrant info.

server/models/registration/error.go
```diff
@@ -0,0 +1,82 @@
...
+		[]string{"Make sure that there is a valid model definition present in the directory. Meshery's registration pipeline currently does not support nested models, therefore the behaviour might be unexpected if it contains nested models.", "If there is an internal error, please try again after some time"},
+	)
+}
+
+func ErrImportFailure(hostname string, failedMsg string) error {
+	return errors.New(
+		ErrImportFailureCode,
+		errors.Alert,
+		[]string{},
+		[]string{fmt.Sprintf("The import process for a registrant %s encountered difficulties,due to which %s. Specific issues during the import process resulted in certain entities not being successfully registered in the table.", hostname, failedMsg)},
```
MUzairS15, line 78:
```suggestion
		[]string{fmt.Sprintf("The import process for a registrant %s encountered difficulties,due to which %s. Specific issues during the import process resulted in certain entities not being successfully registered inside the registry.", hostname, failedMsg)},
```

server/models/registration/error.go
```diff
@@ -0,0 +1,82 @@
...
+		[]string{},
```
MUzairS15, line 79: We might know/guess the cause.

humblenginr:
I will incorporate @MUzairS15 's feedback as a part of next refactor. It will be tracked here - https://github.com/meshery/meshery/issues/11512

server/models/registration/error.go
```diff
@@ -0,0 +1,82 @@
...
+		[]string{fmt.Sprintf("Visit docs with the error code %s", "https://docs.meshery.io/reference/error-codes"),
```
MUzairS15, line 80: It's not very helpful redirecting to the general error code reference, @Jougan-0 do you have the troubleshooting registry if so we can point it there.

MUzairS15 (commented):
The error at places must be replaced with the meshkit error.

server/models/registration/utils.go
humblenginr: this was just meant as a helper function.. the only usage of this function wraps it in a meshkit error.. but yes, I will use meshkit error once moved over to meshkit, as it will be residing in the `utils` package

server/models/registration/error.go
humblenginr: Updated

server/models/registration/error.go
humblenginr: Updated

server/models/registration/error.go
Jougan-0: No, we don't.

server/models/meshmodel/registration/register.go
humblenginr: Updated it

server/models/registration/error.go
humblenginr: Updated the suggested remediation

MUzairS15 (approved):

leecalcote (approved):

Jougan-0:
@humblenginr Conflicts. @MUzairS15 GTM???

humblenginr:
The plan is to move the registration over to Meshkit. Right @MUzairS15 ?

humblenginr:
https://github.com/meshery/meshkit/pull/550

Meshkit release is needed.

humblenginr:
https://github.com/meshery/meshkit/pull/550 - Have to make a release of Meshkit and then merge this

MUzairS15:
`Meshkit` release made.

MUzairS15 (approved):

server/models/seed_models.go
```diff
@@ -0,0 +1,65 @@
...
+	"os"
+	"path/filepath"
+	"github.com/layer5io/meshkit/logger"
+	meshmodel "github.com/layer5io/meshkit/models/meshmodel/registry"
+	"github.com/layer5io/meshkit/models/registration"
+)
+
+var ModelsPath = "../meshmodel"
+
+func getModelDirectoryPaths() ([]string, error) {
```
MUzairS15, line 14: @humblenginr I think this function can be now safely removed, as the implementation of `PkgUnit` for `Directory` handles it already.

Issue #11313, opened 2024-07-03
[mesheryctl] `relationships view` command unmarshall error.


Jougan-0:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

### Current Behavior

When we run `mesheryctl exp relationship view [model-name]` command  there is json unmarshal error.

### Screenshots/Logs

```
./mesheryctl exp relationship view kubernetes
Error: json: cannot unmarshal object into Go struct field MeshmodelRelationshipsAPIResponse.relationships of type entity.Entity
```

### Environment

- Host OS: Mac Linux Windows
- Meshery Server Version: stable-v
- Meshery Client Version: stable-v
- Platform: Docker Kubernetes

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)
- 📚 [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- 📺 [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- ⌨️ [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- ⌨️ [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](http://discuss.meshery.io) and [Community Slack](https://slack.meshery.io)

alloydm:
Hi @Jougan-0 
This my first time contributing to this project, can I work on it?

Jougan-0:
> Hi @Jougan-0 This my first time contributing to this project, can I work on this issue?

Surely. If you could confirm that you were able to run `mesheryctl` locally, I would assign it to you or else I would suggest you to run it locally and then as soon as you confirm. I'll assign it to you.

alloydm:
Hello @Jougan-0 
my earlier comment got deleted bymistakely
Yes i do have the `mesheryctl` locally now.
do I need to fill the member form as well and do other prereqs?

Jougan-0:
[...]

PR #11387, opened 2024-07-12
mesheryctl: fix relationships view command unmarshall error

alloydm:
**Notes for Reviewers**

This PR fixes https://github.com/meshery/meshery/issues/11313
The go's Json Unmarshal was not able to unmarshal output from relationshipApi, since the the []entity.entity struct object did not have the keys from relationshipApi output. Fixed it with adding new struct object from v1beta2.RelationshipDefinition


server/models/meshery_meshmodels_api_response.go
```diff
@@ -45,6 +46,14 @@ type MeshmodelRelationshipsAPIResponse struct {
 	Relationships []entity.Entity `json:"relationships"`
 }
 
+// API response model for meshmodel relationships view API
+type MeshmodelRelationshipsViewAPIResponse struct {
```
Jougan-0, line 50: No need to create a struct thier should be one already in meshery.
Regarding entity.Entity look up the code under directory 
`/server/handlers/component_generation.go` you would some code regarding component entity and would guide you how to fetch the response properly.

Jougan-0 (changes requested):

server/models/meshery_meshmodels_api_response.go
alloydm: > No need to create a struct thier should be one already in meshery. Regarding entity.Entity look up the code under directory `/server/handlers/component_generation.go` you would some code regarding component entity and would guide you how to fetch the response properly.

**Regarding entity.Entity look up the code under directory
/server/handlers/component_generation.go you would some code regarding component entity and would guide you how to fetch the response properly.**
But the issue here is that the entity.Entity type doesn't contain all the fields which the api returns
below are the fields which api returns and I found that []v1alpha2.RelationshipDefinition is there in meshery.
image

Ashparshp:
Let's discuss this on Websites' call. Add this as an agenda item into the meeting minutes, if you would )

Jougan-0 (approved):

PR #11419, opened 2024-07-18
[UI] Registry: Perform model fetch on toggle and InfiniteScroll

dottharun:
**Notes for Reviewers**

This PR fixes 
### settings/registry models tab:
- and only fetches when the version dropdown tree is toggled.
- lazy-loading of pages with `IntersectionObserver`.

### registrants tab:
- Observer added.

### Note:
- Aggregation of models is still done in UI for registrants - removing this requires fixing duplication of data from the API. (but can be done separately with removal of all duplications in the registry)

### DIFF
--- a/ui/components/MeshModelRegistry/MeshModelComponent.js
+++ b/ui/components/MeshModelRegistry/MeshModelComponent.js
@@ -325,6 +325,8 @@ const MeshModelComponent_ = ({
       setSearchText(null);
       setResourcesDetail([]);
     }
+    setModelsFilters({ page: 0 });
+    setRegistrantsFilters({ page: 0 });
     setPage({
       Models: 0,
       Components: 0,


Yashsharma1911:
Can you add video of change you made? on clicking on version we do show the loader right? incase API have some latency otherwise it will look like buggy

Yashsharma1911:
Registry supports deep linking are we making sure after this change it will be working?

ui/components/MeshModelRegistry/MesheryTreeView.js
```diff
@@ -176,70 +196,87 @@ const MesheryTreeViewItem = ({
             }
             check={true}
             onClick={() => {
+              getModelByNameTrigger({
+                name: versionedModelDef.name,
+                params: {
+                  version: versionedModelDef.model.version,
+                  components: true,
+                  relationships: true,
+                },
```
Yashsharma1911, line 205: By default lazy loading does not use caching, please add this prop to make sure it is using it's caching behaviour
```suggestion
                }, true
```

ui/components/MeshModelRegistry/MesheryTreeView.js
```diff
@@ -176,70 +196,87 @@ const MesheryTreeViewItem = ({
...
+                    view={MODELS}
+                    setShowDetailsData={setShowDetailsData}
+                    idForKindAsProp={`${registrantID ? `${registrantID}.1.` : ''}${modelDef.id}.${
+                      selectedVersionedModel.id
+                    }.2`}
+                  />
+                </StyledTreeItem>
+              </>
+            ) : (
+              <CircularProgress />
```
Yashsharma1911, line 278: Okay you already added here thanks, please add this prop to make sure it gets update with theme, also please test after update
```suggestion
              <CircularProgress color="inherit"/>
```

leecalcote:
Suggested code changes applied.

Yashsharma1911:
In video on first item we are not showing the name, I think it is because that model doesn't have displayName can you add a fallback if displayName is not present we by default show the "name" of model

Yashsharma1911:
Testing

Yashsharma1911:
https://github.com/user-attachments/assets/b76e2995-7923-4ab4-a3ed-ddc887eb8d74

@dottharun after scrolling few time in models tab and then switching to other tab and switch back to models, models gets load from page where I left before switching and causing issue of losing prev models info and there is no way I can fetch prev data

Yashsharma1911:
After switching just make sure we are clearing page state

dottharun:
@Yashsharma1911 Made the changes here.

Yashsharma1911:
Thankyou testing

leecalcote:
Is this the only bug left, @Yashsharma1911? You didn't find any others?

dottharun:
test

dottharun:
test

PR #11428, opened 2024-07-19
[Chore] UI: Sistent dependency update v0.14.70

leecalcote:


dragon-slayer875 (changes requested):
Can't update right now.
We need to make the delay of the tooltip governed by a prop and then set that prop at all the places where we want it have the delay.
Otherwise all our icon and minor component tooltips will also have that delay which will degrade the UX.

sudhanshutech:
Its already updated we can close this.

PR #11489, opened 2024-07-29
[CI] Pull Changes Before `auto-commit` Action 

humblenginr:
**Notes for Reviewers**

This PR fixes pulls the changes from the repository before running the [`auto-commit`](https://github.com/stefanzweifel/git-auto-commit-action) action. This is to make sure that there are no new changes introduced in the time between when the code was checked out and when the workflow is trying to commit the changes. 

This PR ensures that the code is being pulled before running the `auto-commit` action in all of Meshery's workflows. 


leecalcote (approved):
Given the high volume of activity in this repo, I wonder if we might not bump in to a race condition at some point. It’s something to keep in the back of our minds should something go awry in the near future.

PR #11542, opened 2024-08-02
Fix unique key constraint error coming on boot

shubham251972:
**Notes for Reviewers**

This PR fixes #

Error

```
ERRO[2024-08-02T11:52:49-05:00] UNIQUE constraint failed: keys.id             app=meshery code=meshery-server-1312 probable-cause="Record already exist.Database connection is not reachable" severity=2 short-description="Unable to create record" suggested-remediation="Delete the record or try updating the record instead of recreating.Rest the database connection"
```

This error was coming because on boot we try to seed keys again and as some keys are already exist in DB gorm try to return error that it is not able to seed keys, to fix this, we need to use gorm `SAVE` operation instead of `CREATE` which make sure if key already exist then replace it or if not then create new db entry

### DIFF
--- a/server/models/keys_persister.go
+++ b/server/models/keys_persister.go
@@ -45,7 +45,7 @@ func (kp *KeyPersister) GetUsersKeys(search, order, updatedAfter string) ([]byte
 
 // SaveUsersKey saves a key to the database
 func (kp *KeyPersister) SaveUsersKey(key *Key) (*Key, error) {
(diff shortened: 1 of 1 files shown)


shubham251972:
@Yashsharma1911 I fixed this issue

Yashsharma1911 (approved):
Thankyou LGTM

PR #11595, opened 2024-08-09
Update the design file in OCI package to design.yml

ctcarrier:
**Notes for Reviewers**

This PR fixes problems in the generation of the artifacthub-pkg file when exporting as an OCI package from MeshMap. The metadata generated was not using the expected date format. And the design yml file did not match the naming convention ArtifactHub expects. With these changes a user can extract the files from the OCI and drop them right into an ArtifactHub repo.

### DIFF
--- a/.github/workflows/build-ui-and-server.yml
+++ b/.github/workflows/build-ui-and-server.yml
@@ -178,8 +178,8 @@ jobs:
         env:
           MESHERY_SERVER_URL: "http://localhost:9081"
           REMOTE_PROVIDER_URL: "https://meshery.layer5.io"
-          REMOTE_PROVIDER_USER_EMAIL: ${{ secrets.LAYER5_CLOUD_TESTING_BOT_EMAIL }}
-          REMOTE_PROVIDER_USER_PASSWORD: ${{ secrets.LAYER5_CLOUD_TESTING_BOT_PASSWORD }}
+          REMOTE_PROVIDER_USER_EMAIL: ${{ secrets.REMOTE_PROVIDER_USER_EMAIL }}
+          REMOTE_PROVIDER_USER_PASSWORD: ${{ secrets.REMOTE_PROVIDER_USER_PASS }}
           PROVIDER_TOKEN: ${{ secrets.PROVIDER_TOKEN }}
         run: |
           make test-setup-ui


leecalcote:
@humblenginr what have you identified as the cause of the long-running build check? https://github.com/meshery/meshery/actions/runs/10322650732/job/28578362504?pr=11595

server/handlers/meshery_pattern_handler.go
```diff
@@ -1185,7 +1186,7 @@ func (h *Handler) DownloadMesheryPatternHandler(
 		}
 		defer os.RemoveAll(tmpDir)
 
-		tmpDesignFile := filepath.Join(tmpDir, pattern.Name+".yaml")
+		tmpDesignFile := filepath.Join(tmpDir, "design.yml")
```
leecalcote, line 1189: @ctcarrier, looking good. 👍

leecalcote (approved):

server/handlers/meshery_pattern_handler.go
```diff
@@ -2069,7 +2070,7 @@ func createArtifactHubPkg(pattern *models.MesheryPattern, user string) ([]byte,
 	if isCatalogItem {
 		version = pattern.CatalogData.PublishedVersion
 	}
-	artifactHubPkg := catalog.BuildArtifactHubPkg(pattern.Name, "", user, version, pattern.CreatedAt.String(), &pattern.CatalogData)
+	artifactHubPkg := catalog.BuildArtifactHubPkg(pattern.Name, "", user, version, pattern.CreatedAt.Format(time.RFC3339), &pattern.CatalogData)
```
humblenginr, line 2073: Thanks for this @ctcarrier . 

Ideally, we would want this to be handled within the `BuildArtifactHubPkg` in Meshkit. The `BuildArtifactHubPkg` function should take `time` type as a parameter rather than `string`. And inside the `BuildArtifactHubPkg` function, we would try to parse it to RFC3339 and return if it failed.

humblenginr (approved):
LGTM

server/handlers/meshery_pattern_handler.go
leecalcote: New issue created to track this enhancement - https://github.com/meshery/meshkit/issues/556

leecalcote:
The failing build check is unrelated and an item for @humblenginr to address.

PR #11625, opened 2024-08-12
[UI] Remove 'Service Mesh' from the Adapters Page

humblenginr:


humblenginr:
Screenshot 2024-08-13 at 12 22 27 AM

I have updated the `Manage Service Mesh` to `Manage Meshery Adapter`. 

I think the following have to be updated as well:
1. The url of the page from `/management/service-mesh` to `management/adapter`
2. The cards for each adapter 

But I am not really sure what the cards for the adapters should say. The cards here have direct correlation with the services exposed by the Adapter and should be dynamic. So I think we need more discussion here on what needs to be done. 

As for the url of the page, I am asking confirmation to change all the references in the codebase to use the updated url.

ui/components/MesheryAdapterPlayComponent.js
```diff
@@ -1313,7 +1313,7 @@ class MesheryAdapterPlayComponent extends React.Component {
               <Grid item xs={12}>
                 <div className={classes.paneSection}>
                   <Typography align="left" variant="h6" style={{ margin: '0 0 2.5rem 0' }}>
-                    Manage Service Mesh
+                    Manage Meshery Adapter
                   </Typography>
```
leecalcote, line 1317: Please delete lines 1315 through 1317. This section doesn't require a title.

leecalcote (changes requested):

ui/components/MesheryAdapterPlayComponent.js
humblenginr: Removed the title

humblenginr:
After removing the title altogether:

Screenshot 2024-08-13 at 12 57 46 AM

leecalcote (approved):

PR #11644, opened 2024-08-15
ci: comment on PR e2e test cases

jerensl:
**Notes for Reviewers**

Bassicly the current artifact being send by the `pull_request` workflow is not be accesible by the caller workflow `workflow_run`, it happened because the context in each workflow is different when it already on the main branch. `workflow_run` using context from main branch and `pull_request` using context from fork PR branch.

The fix is using the event id from the workflow is being called, let's see if it the right ID


.github/workflows/e2e-test-playwright.yml
```diff
@@ -9,12 +9,26 @@ on:
...
-      issues: write
-      pull-requests: write
     steps:
+      - name: Download meshery-ui artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: meshery-ui
+          path: /home/runner/work/meshery/meshery/ui/out
+          github-token: ${{ github.GH_ACCESS_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
```
jerensl, line 24: Let's see if `github.event.workflow_run.id` the right ID

.github/workflows/build-ui-and-server.yml
```diff
@@ -91,41 +91,43 @@ jobs:
         uses: actions/upload-artifact@v4
```
jerensl, line 91: Moving the artifact pull above, in case it fails, no need to wait for all the Kubernetes cluster and updater to be up and running

.github/workflows/e2e-test-playwright.yml
leecalcote: Good point, @jerensl.

PR #11646, opened 2024-08-15
Configure docker extension to use `node.js` path modules

sudhanshutech:
**Notes for Reviewers**

### This PR fixes:
There was error coming while building because the @layer5/sistent package or any project that is attempting to use the Node.js path module. Since Webpack 5 does not include polyfills for Node.js core modules by default, this causes the build to fail when trying to resolve path.
Now we can use sistent package easily.
<hr/>

### Why used react script rewired:
react-app-rewired allows you to customize React App configurations (like Webpack) without ejecting.

❌**Without react-app-rewired:**
 Customization requires ejecting, which exposes configuration files but makes maintenance harder.

✅**With react-app-rewired:** 
You can modify configurations while staying within the managed CRA setup, keeping upgrades easier and your project simpler to maintain.

> [!IMPORTANT]
> This is do not merge because i have to fix styles of catalog card properly from sistent and release then we can move forward.

### DIFF
--- a/install/docker-extension/ui/package-lock.json
+++ b/install/docker-extension/ui/package-lock.json
@@ -14,7 +14,7 @@
         "@fortawesome/fontawesome-svg-core": "^6.1.1",
         "@fortawesome/free-solid-svg-icons": "^6.1.1",
         "@fortawesome/react-fontawesome": "^0.1.18",
-        "@layer5/sistent": "^0.14.85",
+        "@layer5/sistent": "^0.14.87",
         "@mui/icons-material": "^5.5.1",
         "@mui/material": "^5.5.3",
         "@mui/styled-engine-sc": "^5.5.2",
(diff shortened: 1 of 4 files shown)


install/docker-extension/ui/config-overrides.js
```diff
@@ -0,0 +1,22 @@
+const webpack = require('webpack');
+module.exports = function override(config, env) {
```
humblenginr, line 2: @sudhanshutech why don't we use a webpack plugin for this - https://www.npmjs.com/package/node-polyfill-webpack-plugin

humblenginr:
Thanks @sudhanshutech . `react-app-rewired` makes sense ✅

install/docker-extension/ui/config-overrides.js
sudhanshutech: This plugin is not needed when we using the react app rewired so we can easily configure the webpack configurations.
And when i gone through there some discussion it not solving few things completely.

install/docker-extension/ui/package-lock.json
humblenginr, line 1: should we include package-lock in this pr?

sudhanshutech: yes it would be needed😅

leecalcote (approved):
LGTM

sudhanshutech:
@humblenginr please give a final look and let me know so we can merge and the filp card will be available in  next release. I have to little work in sistent for that . But here everything is working fine so we don't have to wait.

humblenginr (approved):
LGTM 🎉

PR #11713, opened 2024-08-20
remove prerun for subcommands w/signoff

Jougan-0:
**Notes for Reviewers**
Model Import and Export in other PR.
This PR fixes #

### DIFF
--- a/mesheryctl/internal/cli/root/environments/list.go
+++ b/mesheryctl/internal/cli/root/environments/list.go
@@ -112,7 +112,7 @@ https://docs.layer5.io/cloud/spaces/environments/
 				whiteBoardPrinter.Println("Total number of environments: ", len(rows))
 				whiteBoardPrinter.Println("Page: ", startIndex/maxRowsPerPage+1)
 
-				whiteBoardPrinter.Println("Press Enter or ↓ to continue, Esc or Ctrl+C (Ctrl+Cmd for OS user) to exit")
+				whiteBoardPrinter.Println("Press Enter or ↓ to continue. Press Esc or Ctrl+C to exit.")
 
 				utils.PrintToTable(header, rows[startIndex:endIndex])
 				keysEvents, err := keyboard.GetKeys(10)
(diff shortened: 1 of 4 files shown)


mesheryctl/internal/cli/root/components/list.go
```diff
@@ -40,39 +38,15 @@ mesheryctl components list --page 2
...
-	},
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if len(args) != 0 {
-			return errors.New(utils.SystemModelSubError("this command takes no arguments\n", "list"))
-		}
 		mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
 		if err != nil {
-			log.Fatalln(err, "error processing config")
+			utils.Log.Error(err)
+			return nil
```
leecalcote, line 45: Wouldn't we want to report this failure to the user and ask if they would like generate a new config? We have an input prompt on this question in another area already that could be used.

mesheryctl/internal/cli/root/components/search.go
```diff
@@ -67,7 +46,7 @@ mesheryctl components search [query-text]
 	RunE: func(cmd *cobra.Command, args []string) error {
 		mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
 		if err != nil {
-			log.Fatalln(err, "error processing config")
+			utils.Log.Error(err)
```
leecalcote, line 49: Seems like an opportunity for a centralized MeshKit Error Code

mesheryctl/internal/cli/root/components/view.go
```diff
@@ -42,27 +42,6 @@ var viewComponentCmd = &cobra.Command{
...
-
-		mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
-		if err != nil {
-			return err
-		}
-		err = utils.IsServerRunning(mctlCfg.GetBaseMesheryURL())
-		if err != nil {
-			return err
-		}
-		ctx, err := mctlCfg.GetCurrentContext()
```
leecalcote, line 56: We should probably continue to include this in PreRun-e

leecalcote (dismissed):
Thank you for walking through this, @Jougan-0.

mesheryctl/internal/cli/root/components/view.go
leecalcote: This function is run separately, later in the same command.

mesheryctl/internal/cli/root/components/search.go
leecalcote: Tested. This behavior works well given the separate handling of creating a meshconfig.

leecalcote (approved):

PR #11773, opened 2024-08-24
Fix: e2e test cases

jerensl:
**Notes for Reviewers**

There is an exit code when the test fails more than 1, just remove it and it will work. The fail test should come from the report comment

### DIFF
--- a/.github/workflows/build-ui-and-server.yml
+++ b/.github/workflows/build-ui-and-server.yml
@@ -178,7 +178,7 @@ jobs:
           ADAPTER_URLS: "localhost:10000 localhost:10001 localhost:10002 localhost:10003 localhost:10004 localhost:10009 localhost:10007"
         run: |
           make server &&
-          sleep 120
+          sleep 90
       - name: Install Playwright Browsers
         run: npx playwright install --with-deps
         working-directory: ui


ui/tests/e2e/custom-playwright-reporter.js
```diff
@@ -36,12 +36,10 @@ class MyReporter {
     const message = await this.buildMessage(result);
 
     try {
-      await writeFileSync('test-report.md', message);
+      writeFileSync('test-report.md', message);
     } catch (e) {
       console.log('Cannot write test reporter ', e);
     }
-
-    process.exit(this.failed > 0 ? 130 : 0); // Return non-zero status code if there are failed tests
```
jerensl, line 44: Removing the exit code, it should run now

.github/workflows/build-ui-and-server.yml
```diff
@@ -177,10 +177,11 @@ jobs:
           PORT: 9081
           ADAPTER_URLS: "localhost:10000 localhost:10001 localhost:10002 localhost:10003 localhost:10004 localhost:10009 localhost:10007"
         run: |
-          make server &
-          sleep 60
+          make server &&
+          sleep 90
```
leecalcote, line 181: While we should probably avoid using `sleep` and instead be event-driven, I understand that this can be a challenge to do.

marblom007 (commented):
LGTM

leecalcote (approved):

PR #11778, opened 2024-08-25
chore: add test keep commenting when fail

jerensl:
**Notes for Reviewers**

This PR fixes #

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	github.com/jinzhu/copier v0.4.0
 	github.com/layer5io/gowrk2 v0.6.1
 	github.com/layer5io/meshery-operator v0.7.0
-	github.com/layer5io/meshkit v0.7.67
+	github.com/layer5io/meshkit v0.7.68
 	github.com/layer5io/meshsync v0.6.24
 	github.com/layer5io/nighthawk-go v1.0.3
 	github.com/layer5io/service-mesh-performance v0.6.1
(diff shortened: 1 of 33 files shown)


Jougan-0:
Why is this happening could you please explain.
1. Why was the timeout on first place.
2. Why the next fix caused the error.
3. How would this fix it. 
Please if you would.

leecalcote (dismissed):

jerensl:
> Why is this happening could you please explain.
> 
> 1. Why was the timeout on first place.

That's purely my mistake, I thought the command `&` for running a task in the background  is the same as sequence `;`, which causes the e2e to fail silently in the past. I'm also add timeout regarding of the workflow to 45 minutes because GitHub timeout is 6 hours long by default

> 2. Why the next fix caused the error?

There is one error in the test that comes from a past error when I trying to fix a silently failed test, for more detail on this [workflow](https://github.com/meshery/meshery/actions/runs/10391293095/job/28774118878#step:14:1100)

There is always a tendency for playwrights to fail the workflow given one single error [more](https://github.com/microsoft/playwright/pulls) example from the playwright repository

> 3. How would this fix it.

It will fix the uploading artifact https://github.com/jerensl/meshery/actions/runs/10547623703/job/29220639056#step:15:1 regarding whether the test failed or not, and hopefully send the comment directly on the PR, I been testing it in the last week with the same approach I used for my [other project](https://github.com/jerensl/www.jerenslensun.com/actions/runs/10446686011/job/28924442683#step:7:7) so it can count as POC

Lastly, because I'm no longer using the testing command from makefile I rollback the changes to the used sequence again for testing in the makefile command

Hope this explain

Jougan-0 (approved):

PR #11804, opened 2024-08-27
Decrease registration time from 30 second to 10 second w/signoff

Jougan-0:
**Notes for Reviewers**
PR **DO NOT MERGE**
This PR fixes #
Merge this [Meshkit PR](https://github.com/meshery/meshkit/pull/581)

### DIFF
--- a/server/models/seed_models.go
+++ b/server/models/seed_models.go
@@ -7,6 +7,8 @@ import (
 
 	"github.com/layer5io/meshery/server/helpers/utils"
 	"github.com/layer5io/meshkit/logger"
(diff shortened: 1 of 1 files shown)


MUzairS15 (changes requested):
@Jougan-0 It's nice to see this improvement.
Let's first define a general framework for worker pool, to which tasks can be assigned.
The current implementation serves the purpose but is too specific/embedded in the registration process.

There are many places currently and will be in future as well, which requires concurrent processing, so it's not feasible to implement logic everywhere.

leecalcote:
I think that we're ok to remove the `do-not-merge` label now.

PR #11810, opened 2024-08-27
fix: comment not send when not find last comment

jerensl:
**Notes for Reviewers**

This PR fixes comment not send when last same comment not exist

### DIFF
--- a/.github/workflows/build-ui-and-server.yml
+++ b/.github/workflows/build-ui-and-server.yml
@@ -195,12 +195,12 @@ jobs:
         run: npx playwright test --project=chromium --project=firefox --reporter="./tests/e2e/custom-playwright-reporter.js"
         working-directory: ui
       - name: Save PR number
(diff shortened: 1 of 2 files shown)


.github/workflows/build-ui-and-server.yml
```diff
@@ -195,12 +195,12 @@ jobs:
         run: npx playwright test --project=chromium --project=firefox --reporter="./tests/e2e/custom-playwright-reporter.js"
         working-directory: ui
       - name: Save PR number
-        if: always()
+        if: ${{ !cancelled() }}
```
jerensl, line 198: This one is better than always() because we can cancel the workflow anytime

leecalcote: Oh, nice. I wasn't aware of this.

leecalcote (approved):
Good!

PR #11891, opened 2024-09-02
[meshreyctl] Use central function to determine latest version.

MUzairS15:
The logic to determine the latest version for `mesheryctl` and then display the "update available" message is spread across commands, the PR ensures that there is a central function to determine the latest version and print the message.

**Notes for Reviewers**

### DIFF
--- a/mesheryctl/pkg/utils/latest_version.go
+++ b/mesheryctl/pkg/utils/latest_version.go
@@ -1,0 +1,54 @@
+package utils
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+)
+
+func GetLatestVersionForMesheryctl() (string, error) {
+	req, err := http.NewRequest(http.MethodGet, "https://docs.meshery.io/project/releases/latest", nil)
+
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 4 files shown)


mesheryctl/pkg/utils/latest_version.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		Log.Warn(fmt.Errorf("unable to check for latest version of mesheryctl. %s", err))
+		return
+	}
+	if latestVersion == "" {
+		Log.Warn(fmt.Errorf("unable to check for latest version of mesheryctl. %s", fmt.Errorf("no version found")))
+		return
+	}
+	// If user is running an outdated release, let them know.
+	if latestVersion != build {
```
MUzairS15, line 47: Compare `semver version` instead of only including "not equal" check.

leecalcote (approved):

mesheryctl/pkg/utils/latest_version.go
MUzairS15: no need, we are good without it as well.

PR #11906, opened 2024-09-04
[fix] Remove redundant store wrapper and add error boundary

dragon-slayer875:
**Notes for Reviewers**

The workspaces and environments pages on being loading directly through URLs were displaying this error:
image

This was happening due to these pages using redundant redux store wrappers without our custom contexts, which were conflicting with the stores which we now use globally.
I have thus removed the store wrapper from these and further wrapped them in error boundary from sistent.

### DIFF
--- a/ui/components/Lifecycle/Environments/index.js
+++ b/ui/components/Lifecycle/Environments/index.js
@@ -1,6 +1,6 @@
 import { useEffect, useRef, useState } from 'react';
 import { Button, Grid, NoSsr, Typography, Box } from '@material-ui/core';
-import { Provider, connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


jerensl:
@dragon-slayer875 Actually this is when Module Federation can solve the problem but I'm not sure if we can proceed with re-architecture a bit. Module federation can work on top of Webpack 5, it has shared dependencies and act like a bridge between 2 web apps or more

https://module-federation.io/practice/overview.html

Yashsharma1911 (commented):
@dragon-slayer875 did we solve the bug that you were getting, is this good to merge?

PR #11909, opened 2024-09-04
Fix handling of connection API response from remote provider

MUzairS15:
**Notes for Reviewers**

The response schema for connection API was updated by the Remote Provider, but in Meshery the changes were not incorporated, i.e. migrating to different structs and hence causing issues in the connection management.

For example, K8s connection despite being reachable, was not transitioning to the next states.


Yashsharma1911 (dismissed):
Oops my bad thanks for fixing it ^^

PR #11923, opened 2024-09-06
update component w/signoff

Jougan-0:
**Notes for Reviewers**

This PR fixes #


leecalcote (commented):
Why wouldn’t we update the component it if it does not have a description?

Jougan-0:
> Why wouldn’t we update the component it if it does not have a description?

We update the component even though the description is not present. It is never returning it is just a check if the description is not present then don't create the key as `description` is not required in schema.

Jougan-0:
DO NOT MERGE I'll pass all the fixes in this PR itself and we would be done with this today.

leecalcote (approved):

leecalcote (approved):

leecalcote:
Complete with a new flag, too!

PR #11980, opened 2024-09-11
Anonymous user

Jougan-0:
**Notes for Reviewers**

This PR fixes #


server/models/remote_provider.go
```diff
@@ -310,7 +311,38 @@ func (l *RemoteProvider) InitiateLogin(w http.ResponseWriter, r *http.Request, _
...
 		}
+		if viper.GetBool("PLAYGROUND") {
+			ep, exists := l.Capabilities.GetEndpointForFeature(PersistAnonymousUser)
+			if !exists {
+				l.Log.Warn(ErrInvalidCapability("PersistAnonymousUser", l.ProviderName))
+				return
+			}
+			client := &http.Client{}
+			req, _ := http.NewRequest("GET", l.RemoteProviderURL+ep, nil)
+			req.Header.Set("X-API-Key", GlobalTokenForAnonymousResults)
```
leecalcote, line 322: The remote provider should be offering this and it should not be a burden on Meshery Server to request it.

@MUzairS15 @Jougan-0, we might need to go over notes from our in-depth architecture discussion.

server/models/remote_provider.go
```diff
@@ -310,7 +311,38 @@ func (l *RemoteProvider) InitiateLogin(w http.ResponseWriter, r *http.Request, _
...
+			}
 
+			var tokenResp AnonymousTokenResponse
+			err = encoding.Unmarshal(body, &tokenResp)
+			if err != nil {
+				l.Log.Error(err)
+				return
+			}
+			l.SetJWTCookie(w, tokenResp.AccessToken)
+			redirectURL := "/extension/meshmap"
```
leecalcote, line 342: @Jougan-0 this should not be hardcoded into Meshery Server. This is an extension point. The remote provider needs to provide this information.

leecalcote (changes requested):

server/models/user.go
```diff
@@ -46,3 +46,6 @@ type UserKeys struct {
 	Updated_at  string `json:"updated_at,omitempty"`
 	Deleted_at  string `json:"deleted_at,omitempty"`
 }
+type AnonymousTokenResponse struct {
```
leecalcote, line 49: Why would this be any different that the regular user token?

server/models/remote_provider.go
Jougan-0: I agree. In [token handler](https://github.com/meshery/meshery/blob/f079c3696de88a1689ed2b262db3ab80cb54fe16/server/models/remote_provider.go#L3442) we did the same so I used it rather I should have updated both.
I'll update both the places.

server/models/remote_provider.go
Jougan-0: I don't the question here? 
We need to make a rest call to an endpoint on remote provider to get the token and for authentication we need something so we are using the special token for authentication.
I might be wrong on the interpretation of the question.

leecalcote: Let's use a sequence diagram to be of the same understanding. 

I've been hopeful to avoid a number of requests back and forth between server and remote provider.

server/models/remote_provider.go
MUzairS15: This has been removed now.

MUzairS15:
Completed.

MUzairS15 (approved):
LGTM

PR #12035, opened 2024-09-14
configure meshery ui before building

codeSafari10:
**Notes for Reviewers**

This PR fixes #

add a custom ui config for meshery

### DIFF
--- a/ui/utils/disabledComponents.js
+++ b/ui/utils/disabledComponents.js
@@ -27,9 +27,9 @@ export class CapabilitiesRegistry {
     this.isPlaygroundEnv = capabilitiesRegistry?.restrictedAccess?.isMesheryUiRestricted || false;
   }
 
-   capabilities() {
-     return this.capabilitiesRegistry
-   }
+  capabilities() {
+    return this.capabilitiesRegistry;
+  }
 
   isNavigatorComponentEnabled(navigatorWalker) {
     if (!this.isPlaygroundEnv) {


.github/workflows/build-release.yml
```diff
@@ -23,6 +23,16 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+
```
leecalcote, line 26: Will this displace any "PLAYGROUND" mode code?

aabidsofi19: @leecalcote  yeah the code in ui that was used to lockdown the navigator and header component is not needed now ( i have removed the code for lockdown of header components )

ui/components/User.js
```diff
@@ -160,6 +161,21 @@ const User = (props) => {
   const { color, iconButtonClassName, avatarClassName, classes } = props;
 
   const open = Boolean(anchorEl);
+
+
+  if ( userData?.status == 'anonymous' ) {
+    const url =
+         `${capabilitiesRegistry?.provider_url}/user/login?anonymousUserID=${userData?.user_id}&redirect=${window.location.pathname}`
```
leecalcote, line 168: @MUzairS15 why is this parameter explicit and not generic like `userID`? This is asymmetric to the extent that there isn't an `ActiveUserID` and a `PendingUserID`, so what justifies breaking the pattern here of now explicitly exposing user account status intermingled with the user ID? If user account status is needed to be known, should it not be exposed as it own `status` parameter given that it _is_ its own property in the schema? Are we being schema-driven in our API endpoint behavior?

leecalcote (dismissed):
Some docs on this enhanced extension point is ideal.

leecalcote:
Once eslint issues are fixed, we can merge. 👍

ui/components/User.js
```diff
@@ -160,6 +161,21 @@ const User = (props) => {
...
+        
+    return (
```
leecalcote, line 170: It'd be great to have a small section in https://docs.meshery.io/extensibility/ui explain the purpose of this extension point.

ui/components/User.js
```diff
@@ -160,6 +161,21 @@ const User = (props) => {
   const { color, iconButtonClassName, avatarClassName, classes } = props;
 
   const open = Boolean(anchorEl);
+
+
+  if ( userData?.status == 'anonymous' ) {
+    const url =
+         `${capabilitiesRegistry?.provider_url}/user/login?anonymousUserID=${userData?.user_id}&redirect=${window.location.pathname}`
+
```
leecalcote, line 169: // EXTENSION POINT: Anonymous Users
// Insert custom logic here to handle Single User mode, Anonymous User mode, Multi User mode behavior.

PR #12058, opened 2024-09-18
[chore] Change meshery extension reference to generic capabilities property check

leecalcote:

### DIFF
--- a/ui/components/Popup.js
+++ b/ui/components/Popup.js
@@ -80,7 +80,7 @@ const isMesheryExtensionRegisteredUser = (capabilitiesRegistry) => {
 
   return (
     capabilitiesRegistry.extensions?.navigator?.length > 0 &&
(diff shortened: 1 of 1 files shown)


leecalcote:
@jerensl will you look into why "20. Project: chromium, Test: Toggle "Meshery Catalog Content", Status: ❌, Retry: 0 
File Location: Not Found" is failing?

leecalcote:
A rename to be coordinated with remote providers.

jerensl:
> @jerensl will you look into why "20. Project: chromium, Test: Toggle "Meshery Catalog Content", Status: ❌, Retry: 0 File Location: Not Found" is failing?

Seems like new bug, when I trying to disable the Meshery Catalog Content in user preference, the response is OK but the response body always true from the API

MUzairS15 (approved):

PR #12086, opened 2024-09-20
test(e2e): fix perfomance load test and add delete performance profile

jerensl:
**Notes for Reviewers**

- [x] Fix performance load test
- [x] Add test for deleting performance profile 


ui/tests/e2e/service_mesh_performance.spec.js
```diff
@@ -1,62 +1,80 @@
...
-    );
-    await expect(
-      performancePage.page.getByRole('button', { name: 'Run Test', exact: true }),
-    ).toBeVisible();
-    await performancePage.page.getByRole('button', { name: 'Run Test', exact: true }).click();
-    const notification = await performancePage.page
-      .locator('text=Initiating load test . . .')
-      .first();
-    await expect(notification).toBeVisible();
+      await expect(await page.getByText('Performance Profile Deleted!').first()).toBeHidden();
```
leecalcote, line 77: Should we be looking for the missing it test result by ID or test result by name instead of a message back? Asserting on the text message back seems fragile. Looking at it, I can tell you that I would like for it to be changed and I have the!m exclamation mark removed. If we were to go do that, Then the test will break.

l5io:
🚀 Preview for commit 6e65468af9520ae72a14ec52b08cc6e138c49f22 at: https://66eec9b0ef2baf53c523fa9f--meshery-docs-preview.netlify.app

PR #12176, opened 2024-09-30
update registration to support latest version w/signoff

### LINKED ISSUE #12155: [Server] Register latest model version only at boot
### Current Behavior
Meshery tracks and supports many versions of its components and their models. Quite commonly duplicate components are not needed at all versions of a given model.

### Desired Behavior
By default,  Meshery Server will ideally only load the latest version of each model.

In addition to performance, lack of relationship duplication, this also has the added benefit of showing users an accurate model and component count upon boot and log into Meshery UI, not inflated numbers with duplicate components.

---

Jougan-0:
**Notes for Reviewers**

This PR fixes #12155
Merge this [Meshkit PR](https://github.com/meshery/meshkit/pull/599) before merging this PR.


MUzairS15 (commented):
@Jougan-0 As discussed, both comps and rels will be registered from the latest version of the model.

To do that, the ability to generate relationships from Integrations sheet for all models needs to be incorporated before this gets merged correct?

Otherwise, there will be issues where all models don'thave relationships.

Jougan-0:
> @Jougan-0 As discussed, both comps and rels will be registered from the latest version of the model.
> 
> To do that, the ability to generate relationships from Integrations sheet for all models needs to be incorporated before this gets merged correct?
> 
> Otherwise, there will be issues where all models don'thave relationships.

Yes, that's what we're currently doing. However, I added a fail-safe to check if the latest version lacks relationships, and if an earlier version has them, we'll use that instead. Generally, we register the latest version, but since the logic for generating relationships isn't implemented yet, this approach will work without breaking anything. Moreover, this method should continue functioning smoothly in the future as well.

The code won't require updates, as it includes an additional check on the latest version.

PR #12177, opened 2024-09-30
[Server] Add design import event for local upload.

MUzairS15:
**Notes for Reviewers**

This PR fixes #

### DIFF
--- a/server/handlers/meshery_pattern_handler.go
+++ b/server/handlers/meshery_pattern_handler.go
@@ -128,9 +128,9 @@ func (h *Handler) handlePatternPOST(
 	}()
 
 	var err error
-	action := models.Create
+	action := models.Import
 	userID := uuid.FromStringOrNil(user.ID)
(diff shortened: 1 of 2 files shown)


server/handlers/meshery_pattern_handler.go
```diff
@@ -128,9 +128,9 @@ func (h *Handler) handlePatternPOST(
 	}()
 
 	var err error
-	action := models.Create
+	action := models.Import
```
lekaf974, line 131: We wanted to replace create by import ? The title is misleading

leecalcote: Good call. I believe that we're trying to make "create" synonymous with "import", preferring the latter term in the case of a user uploading a local file or importing remote file via URL.

leecalcote (approved):

server/handlers/meshery_pattern_handler.go
leecalcote: @MUzairS15 do we continue to have a create event? What event are we sending when a user starts a design from scratch?

PR #12209, opened 2024-10-02
feat(loading-screen-for-root-app): loading screen for the root app

jerensl:
**Notes for Reviewers**

This loading is rendered on servers that don't have direct access to this current MUI version configuration. Also because the server doesn't have access to the react state for example useState, useEffect, etc, so I need to remove some of the animation

MUI in SSR/SSG supported in NextJS 13 or above
https://mui.com/material-ui/integrations/nextjs/#pages-router

### DIFF
--- a/ui/pages/_app.js
+++ b/ui/pages/_app.js
@@ -681,12 +681,10 @@ class MesheryApp extends App {
     const canShowNav = !this.state.isFullScreenMode && uiConfig?.components?.navigator !== false;
 
     return (
-      <DynamicComponentProvider>
-        <RelayEnvironmentProvider environment={relayEnvironment}>
-          <MesheryThemeProvider>
-            {this.state.isLoading ? (
-              <LoadingScreen message="Meshery Loading..." />
-            ) : (
+      <LoadingScreen message="Meshery Loading..." isLoading={this.state.isLoading}>
+        <DynamicComponentProvider>
+          <RelayEnvironmentProvider environment={relayEnvironment}>
+            <MesheryThemeProvider>
               <NoSsr>
                 <ErrorBoundary>
                   <div className={classes.root}>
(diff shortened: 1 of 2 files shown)


leecalcote:
Thanks!

aabidsofi19:
@jerensl  while this is good , it would be better if this loading is also uplifted a bit , i.e in the root html while we are fetching the app.js / indexjs  we can also show the loading screen

ui/pages/_app.js
```diff
@@ -681,105 +684,109 @@ class MesheryApp extends App {
...
-                        <MesheryProgressBar />
-                        {!this.state.isFullScreenMode && (
-                          <Header
-                            onDrawerToggle={this.handleDrawerToggle}
-                            onDrawerCollapse={isDrawerCollapsed}
-                            contexts={this.state.k8sContexts}
-                            activeContexts={this.state.activeK8sContexts}
-                            setActiveContexts={this.setActiveContexts}
-                            searchContexts={this.searchContexts}
+            {this.state.isLoading ? (
```
aabidsofi19, line 687: please dont use a terninary operator here . it makes the whole block messy and hard to read . create a Loading component that take children and isLoading as props and renderes accordingly

aabidsofi19 (changes requested):

ui/pages/_app.js
```diff
@@ -681,105 +684,109 @@ class MesheryApp extends App {
...
+                    </div>
                   </div>
-                </div>
-                <PlaygroundMeshDeploy
-                  closeForm={() => this.setState({ isOpen: false })}
-                  isOpen={this.state.isOpen}
-                />
-              </ErrorBoundary>
-            </NoSsr>
+                  <PlaygroundMeshDeploy
```
leecalcote, line 783: We shouldn’t have a component named playground.

jerensl:
> @jerensl while this is good , it would be better if this loading is also uplifted a bit , i.e in the root html while we are fetching the app.js / indexjs we can also show the loading screen
> 
> said that it would need to be a plain html/css loading as we wont be able to access react there

I think it is fine to put in `_app.js` because the rendering happened in build time also we need to utilize loading state from `componentDidMount`, so far I have tested it in multiple pages, and I'm never seeing a blank white page

PR #12226, opened 2024-10-02
moved /regal to /policies/wasm.policies

### LINKED ISSUE #12139: Relocate .regal to `/policies` directory.
**Current Situation***

`meshery/meshery`'s root directory contains a `.regal` folder.

***Desired Situation**

The `.regal` folder is relocated to  `/policies` folder.

**Acceptance Criteria**

1. Regal linter continues to work.

---
[Slack Message](https://layer5io.slack.com/archives/CDM0ACDM5/p1727363051300079?thread_ts=1727363051.300079&cid=CDM0ACDM5)

moiz2405:
**Moved /.regal to /policies/wasm/policies/.regal**
Confirmed the build running successfully after changing the directory
This PR fixes #12139 
Screenshot 2024-10-02 233505

Closed the previous PR due to multiple merge commits 


ctcarrier:
@moiz2405 This configuration file is used by Regal when linting the Rego policies. The way the tool works is that it searches the current directory and any parent directories for a `.regal/config.yaml` file. If it doesn't find one it just uses the defaults. You can read about it in the docs here:

https://docs.styra.com/regal#configuration

In this case we would be running the regal lint from the root of the project and this PR moves the config into a child directory. In that case we would need to make sure to add a flag pointing to the new location for the config file because it won't automatically be detected. In the workflow file in `.github/workflows/regal-lint.yml` you would need to add a new flag to the command in the last line so it looks something like:

`run: regal lint --config-file ./policies/wasm/policies/.regal/config.yaml --format github ./server/meshmodel`

I wouldn't think this would be causing the end to end tests to fail though so maybe something else is going on there.

moiz2405:
> @moiz2405 This configuration file is used by Regal when linting the Rego policies. The way the tool works is that it searches the current directory and any parent directories for a `.regal/config.yaml` file. If it doesn't find one it just uses the defaults. You can read about it in the docs here:
> 
> https://docs.styra.com/regal#configuration
> 
> In this case we would be running the regal lint from the root of the project and this PR moves the config into a child directory. In that case we would need to make sure to add a flag pointing to the new location for the config file because it won't automatically be detected. In the workflow file in `.github/workflows/regal-lint.yml` you would need to add a new flag to the command in the last line so it looks something like:
> 
> `run: regal lint --config-file ./policies/wasm/policies/.regal/config.yaml --format github ./server/meshmodel`
> 
> I wouldn't think this would be causing the end to end tests to fail though so maybe something else is going on there.

I see, thanks for the input @ctcarrier. I was considering it to cause of test failures, will start looking into the tests

vishalvivekm:
@moiz2405 
Thank you for your contributions! Let's discuss this during the Meshery Development call today at 9:00 AM CT. Adding it as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit?pli=1#) || [meeting link](https://meet.layer5.io/meshery).

moiz2405:
@ctcarrier I've changed the necessary files, and checked the build to be working fine. Please do a final review

ctcarrier (approved):

jerensl (approved):

PR #12324, opened 2024-10-08
[CI] Enhance  make target for port as an argument

leecalcote:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,20 @@
 include install/Makefile.core.mk
 include install/Makefile.show-help.mk
 
+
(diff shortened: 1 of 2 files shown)


leecalcote:
@pontusringblom, sure: 

```
make server PORT=9082
```

This overrides the global variable in the Makefile.core.env file.

marblom007 (approved):
This example makes sense to me. What doesn't make sense is the fact that the if/else does the same echo statement... but, this logic still works. `make` is an interesting animal.

pontusringblom (approved):
Cool. Thank you.

PR #12361, opened 2024-10-13
Allow getting entities by ids

Yashsharma1911:
**Notes for Reviewers**

This PR fixes #

This PR update entities API to have `id` params, which allows to fetch entity like models, components and relationship by id

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/jinzhu/copier v0.4.0
 	github.com/layer5io/gowrk2 v0.6.1
 	github.com/layer5io/meshery-operator v0.7.0
-	github.com/layer5io/meshkit v0.7.81
+	github.com/layer5io/meshkit v0.7.82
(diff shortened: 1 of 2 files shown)


Yashsharma1911:
This PR depends on new release of meshkit

Yashsharma1911:
// @amitamrutiya

amitamrutiya:
@Yashsharma1911 for these two I still have to pass the modal name because the modal name is a variable, not an argument that we are passing.

`GetAllMeshmodelRelationships`
`GetMeshmodelComponentByModel`

Yashsharma1911:
> @Yashsharma1911 for these two I still have to pass the modal name because the modal name is a variable, not an argument that we are passing.
> 
> `GetAllMeshmodelRelationships` `GetMeshmodelComponentByModel`

Ohh I see I missed those once I guess, you can simply make changes to those functions similar to this PR

PR #12459, opened 2024-10-23
Fix duplicate match labels

alero-awani:
**Notes for Reviewers**

The use of `sorted_ids` in the policy ensures that the relationship between two components is consistently ordered by their IDs. By sorting the `from_decl.id` and `to_decl.id` using the sort function, the policy guarantees that the smaller ID is always assigned to `from`, and the larger to `to`. This consistency prevents the creation of duplicate relationships in reverse order (i.e., a relationship from A to B and another from B to A), ensuring that only one relationship is created between two components.

image


ctcarrier:
@Alero-Awani Change looks good to me but there are some errors on the Rego lint. I see you didn't change these files so not sure how they got out of compliance. You should be able to resolve by running the `opa fmt` command on these files like:
`opa fmt -w /path/to/file`

leecalcote:
@Jougan-0 @Alero-Awani, is this an item that can be fixed this weekend?

leecalcote:
@Jougan-0 @Alero-Awani is this change still needed?

PR #12540, opened 2024-11-01
remove section and texbox from the meshery-shapes

amitamrutiya:
**Notes for Reviewers**

- This PR fixes #

- For ui changes: Please add a screenshot showcasing the UI before/after changes. 


amitamrutiya:
@Jougan-0 If we delete some fields from the integration sheet why it's not deleted by workflow? Every time We have to delete it manually.

Jougan-0:
Once a model is generated it has n no. of components and if you want to delete a component it means you are upgrading the version so what you can do is delete the entry and bump the version of the model in the Models sheet.

amitamrutiya:
Thanks @Jougan-0 for your explanation.

PR #12578, opened 2024-11-05
fix publish and unpublish action trigger

sudhanshutech:
**Notes for Reviewers**

- This PR fixes #

- The publish and unpublish action trigger issues. Unpublish action was not triggering due to event propogation issue and was delayed to display because we were using a check called `canPublishPattern` which is not needed for unpublish action and this prop was taking time to render.
- Remove publish action from design table since we use info modal to publish.

### DIFF
--- a/ui/components/MesheryPatterns.js
+++ b/ui/components/MesheryPatterns.js
@@ -225,10 +225,12 @@ const styles = (theme) => ({
   },
 });
 
-function TooltipIcon({ children, onClick, title, placement }) {
+function TooltipIcon({ children, onClick, title, placement, disabled }) {
   return (
     <CustomTooltip title={title} placement={placement} interactive>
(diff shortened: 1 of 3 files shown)


leecalcote:
Does it hurt to have the publish action in design table? I assume that doing so means that we’re not able to validate all the necessary catalog details. Is that right?

leecalcote:
Unpublish does need a permission check. Is there one included?

sudhanshutech:
> Unpublish does need a permission check. Is there one included?

yes the keys are included.

sudhanshutech:
> Does it hurt to have the publish action in design table? I assume that doing so means that we’re not able to validate all the necessary catalog details. Is that right?

No it shouldn't hurt to have publish action but we brought the info modal for the purpose for this only so it can be used for both to publish and unpublish items and also yes it give a additional step to validate things

PR #12586, opened 2024-11-06
fix: playwright trace is missing on ci

jerensl:
**Notes for Reviewers**

- Fix trace
- Remove firefox test to avoid data race

### DIFF
--- a/docs/_releases/v0.7.139.md
+++ b/docs/_releases/v0.7.139.md
@@ -1,0 +1,25 @@
+---
+layout: release
+date: 2024-11-08
+tag: v0.7.139
+prerelease: false
+---
+
+## What's New
+## 🔤 General
+- Fix redirects @Jougan-0 (#12607)
+
+## ⌨️ Meshery CLI
+
+- [mesheryctl] Add some prereq checks for system config gke command @hexxdump (#12584)
+
+## 🖥 Meshery UI
+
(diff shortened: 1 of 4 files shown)


ui/playwright.config.js
```diff
@@ -38,7 +41,7 @@ module.exports = defineConfig({
       mode: 'on-first-retry',
     },
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: 'on-first-retry',
```
ianrwhitney, line 41: curious as the test suite grows if we will have to consider only keeping traces if the fail for ci perfermance. 

 This option keeps the trace on retry even if it succeeds/fails. Is that correct?

ianrwhitney (commented):
LGTM. one thought about future considerations. thanks

ui/playwright.config.js
ianrwhitney: curious thoughts on retain-on-failure option. As if we implemented 1 retry to reduce noise of flaky tests, we won't have to update this again. let me know your thoughts. Either way im good as along as there is awerness. thanks 🚀

ianrwhitney (approved):
LGTM. thanks 🚀 . 

Curious where the traces will be located and how PR owers will be able to view them? Any updates in docs required for contributers

leecalcote (approved):
Thank you, gents!

PR #12610, opened 2024-11-08
[UI] Server url has gone undefined for connection chips in Header

SAHU-01:
**Notes for Reviewers**

- This PR fixes the error caused by  [#12581](https://github.com/meshery/meshery/pull/12581)

image
image

### DIFF
--- a/ui/components/DesignLifeCycle/finalizeDeployment.js
+++ b/ui/components/DesignLifeCycle/finalizeDeployment.js
@@ -47,9 +47,7 @@ export const FinalizeDeployment = ({ design, openInVisualizer, setOpenInVisualiz
             {envNames.map((env) => (
               <StyledEnvironment key={env}>
                 <EnvironmentIcon fill={palette.text.neutral.default} />
-                <Typography variant="textB1Regular" key={env}>
-                  {env}
-                </Typography>
+                <Typography key={env}>{env}</Typography>
               </StyledEnvironment>
             ))}
           </Stack>
@@ -61,11 +59,15 @@ export const FinalizeDeployment = ({ design, openInVisualizer, setOpenInVisualiz
           justifyContent={'center'}
         >
           <Box display="flex" alignItems="baseline">
-            <Typography variant="textH1Bold" color={palette.text.neutral.default}>
+            <Typography
+              sx={{
+                fontSize: '3.2rem',
+              }}
+              color={palette.text.neutral.default}
+            >
               {configurableComponents.length}
             </Typography>
             <Typography
-              variant="textB2SemiBold"
               color={palette.text.disabled}
               style={{
                 textTransform: 'lowercase',


ui/components/DesignLifeCycle/SelectDeploymentTarget.js
```diff
@@ -56,9 +56,10 @@ const K8sContextConnection = ({ connection, environment }) => {
   );
   const dispatch = useDispatchRtk();
   const toggleK8sConnection = () => dispatch(toggleConnection(environment, connection));
+
   return (
     <K8sContextConnectionChip
-      ctx={connection}
+      ctx={{ ...connection.metadata, connection_id: connection.id }}
```
aabidsofi19, line 62: we should directly use the whole connection as the prop . this will otherwise make it confusing to know what is included in ctx

aabidsofi19: and also rename the prop to connection than ctx .

aabidsofi19 (changes requested):

ui/components/DesignLifeCycle/SelectDeploymentTarget.js
SAHU-01: The thing is, if I pass the whole connection it doesn't set the connection_id. And inside deployment it fails.
If we do this it specifically sets the connection_id
` ctx={{ ...connection, connection_id: connection.id }}`

aabidsofi19: the issue is connetion.metadata.id is getting intermixed with connection.id  ,

ui/components/Header.js
```diff
@@ -442,13 +441,13 @@ function K8sContextMenu({
...
                         <K8sContextConnectionChip
-                          key={ctx?.metadata?.id}
+                          key={ctx.id}
                           classes={classes}
                           ctx={ctx}
                           selectable
                           onDelete={handleKubernetesDelete}
-                          selected={activeContexts.includes(ctx?.id)}
-                          onSelectChange={() => setActiveContexts(ctx?.id)}
+                          selected={activeContexts.includes(ctx.id)}
```
aabidsofi19, line 449: this will become connection.metadata.id

ui/components/Header.js
```diff
@@ -243,13 +242,13 @@ const K8sContextConnectionChip_ = ({
...
             <UsesSistent>
-              <Checkbox checked={selected} onChange={() => onSelectChange(ctx?.metadata?.id)} />
+              <Checkbox checked={selected} onChange={() => onSelectChange(ctx.id)} />
             </UsesSistent>
           )}
           <_ConnectionChip
             title={ctx?.name}
-            onDelete={onDelete ? () => onDelete(ctx?.metadata?.name, ctx?.id) : null}
-            handlePing={() => ping(ctx?.metadata?.name, ctx?.metadata?.server, ctx?.id)}
+            onDelete={onDelete ? () => onDelete(ctx.name, ctx.connection_id) : null}
```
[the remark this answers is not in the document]
SAHU-01: I believe rather setting handle ping to connection.id would work here...

PR #12623, opened 2024-11-10
fix rel defintion w/signoff

Jougan-0:
**Notes for Reviewers**

- This PR fixes #

For all the relationships that are of network type between service and pod and if they are not working make sure the targetPort is specified and protocol is given.

I changed in this the rel from target Port to port as most of our design used it.


Jougan-0:
Rel were getting deleted because when the pod is inside a deployment and you match the ports since before they were only checked with 0 position of array so it was causing the rel to be deleted I have updated the rel to take this situation into account.

Jougan-0:
I would suggest we have 2 rel definition.
1. Service to deployment if the value given is port.
2. Service to deployment if the value given is target port.

I saw various example files and this is how the rels exist.

@Alero-Awani thoughts ?

PR #12680, opened 2024-11-18
Pass meshery store and organization store in remote component

sudhanshutech:
**Notes for Reviewers**

- This PR fixes #

### DIFF
--- a/ui/components/NavigatorExtension.js
+++ b/ui/components/NavigatorExtension.js
@@ -6,7 +6,7 @@ import {
   createRequires,
 } from '@paciolan/remote-component';
 import { bindActionCreators } from 'redux';
-import { updateLoadTestData, setK8sContexts, useLegacySelector, mesheryStore } from '../lib/store';
+import { updateLoadTestData, setK8sContexts, mesheryStore } from '../lib/store';
 import GrafanaCustomCharts from './telemetry/grafana/GrafanaCustomCharts';
 import MesheryPerformanceComponent from './MesheryPerformance';
 import dataFetch from '../lib/data-fetch';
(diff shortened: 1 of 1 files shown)


ui/components/NavigatorExtension.js
```diff
@@ -139,6 +146,14 @@ function NavigatorExtension({
             useFilterK8sContexts,
             useDynamicComponent,
           },
+          store: store,
+          currentOrg: setOrganization,
+          reduxHooks: {
```
aabidsofi19, line 151: dont pass the whole store , as it would add additional dependencies on meshery . only pass a object `mesheryStore` which would include setters , getters and selectors for the exposed properties  ( in this case oganization )

ui/components/NavigatorExtension.js
```diff
@@ -139,6 +146,14 @@ function NavigatorExtension({
             useFilterK8sContexts,
             useDynamicComponent,
           },
+          store: store,
+          currentOrg: setOrganization,
```
aabidsofi19, line 150: dont pass this as standlone property

ui/components/NavigatorExtension.js
aabidsofi19: eg mesheryStore  : { organization : { set : () => dispatch(....) , get : ()=> store.get(...) , useCurrentOrg : () => useSelector(useCurrentOrgSelector)) }

aabidsofi19 (changes requested):

sudhanshutech:
@aabidsofi19 seems fine now? Should i keep passing the reduxhooks object?

ui/lib/store.js
```diff
@@ -405,6 +405,15 @@ export const setKeys =
     return dispatch({ type: actionTypes.SET_KEYS, keys });
   };
 
+export const mesheryStore = {
+  organization: {
+    set: (organization) => (dispatch) =>
+      dispatch({ type: actionTypes.SET_ORGANIZATION, organization }),
```
aabidsofi19, line 411: Directly use store.dispatch this won't work

aabidsofi19 (changes requested):

ui/lib/store.js
amitamrutiya: I guess we are not using this meshery store in the latest update. Sudhanshu we can remove it now.

PR #12755, opened 2024-11-25
Improvement in errorBoundary

sudhanshutech:
**Notes for Reviewers**

- This PR fixes #
- Remove all other different error boundaries and use a single sistent error boundary
- Use this error boundary to inject in remote component and root level directly
- This error boundary also includes **Feedback** component which will display when error occurs to submit issues

{215249D2-5420-47F0-9C47-246A7002F28D}


leecalcote (changes requested):
Please confirm both local and remote provider compatibility.

leecalcote:
@sudhanshutech are we good on both local and remote provider experiences?

sudhanshutech:
> @sudhanshutech are we good on both local and remote provider experiences?

Remote provider is good but local will not work because the feedback and support action are using extensions api so that's why they will not work with local provider. I adding a check for now to do not show this action for local provider

PR #12887, opened 2024-12-13
Download event for downloading manifest or design type

pranjalg1331:
**Notes for Reviewers**
Adding download event with description when downloading K8 Manifest or Meshery Design.
- This PR fixes #


leecalcote (changes requested):
Build check failure

pranjalg1331:
@leecalcote, The test is failing cause we are not checking the error when creating event. But i don't think we need to since the error won't affect the main logic of handlers.

leecalcote:
Include an error handler error code specific to events.

PR #13026, opened 2024-12-26
[mesheryctl] improve `mesheryctl registry update`

### LINKED ISSUE #13016: [mesheryctl] improve `mesheryctl registry update`
so running this 
```console
$ mesheryctl registry update
Error: unexpected end of JSON input
```
this not a desired behavior, instead it should print out the help of this command as it seems that it's a false command that's missing args

TusharMohapatra07:
**Notes for Reviewers**

- This PR fixes #13016 


TusharMohapatra07:
Hey maintainers, i think this should fix the issue

lekaf974 (changes requested):
Thanks @TusharMohapatra07 but you need to fix the root cause of the issue and follow mesheryctl UX implementation. 

2 things needs to be done here:
1. Check why this error is triggered 
2. wrap it inside a Meshkit Error following UX put in place

I guess it comes from a missing argument

lekaf974:
your branch needs to be resynch with master

TusharMohapatra07:
@devhindo Please check if this was the intended fix. I used the `GetString` method to get the corresponding argument value of `spreadsheet-id` and `spreadsheet-cred`. In case of err, the function returns an empty string. I've checked for empty string condition in the if block and returned `cmd.Help()`

mesheryctl/internal/cli/root/registry/update.go
```diff
@@ -55,6 +55,12 @@ mesheryctl registry update --spreadsheet-id 1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdw
 mesheryctl registry update --spreadsheet-id 1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw --spreadsheet-cred $CRED --model "[model-name]"
 	`,
 	PreRunE: func(cmd *cobra.Command, args []string) error {
+		spreadsheetID, _ := cmd.Flags().GetString("spreadsheet-id")
+		spreadsheetCred, _ := cmd.Flags().GetString("spreadsheet-cred")
+
+		if spreadsheetID == "" || spreadsheetCred == "" {
+			return cmd.Help()
```
lekaf974, line 62: Should be a Meshkit Error

TusharMohapatra07: @lekaf974 you mean, wrap the error with the error from the "github.com/layer5io/meshkit/errors" package ?

lekaf974: yep this is the way we want to standardized error in mesheryctl

TusharMohapatra07:
Hey @lekaf974, I initially just tried wrapping the error around as such but that cause segmentation fault as we are trying to called the Error() method on the returned error inside `ErrUpdateRegistry`. After reading the source code of cmd.Help(), i came up with the changes made by above commit.

source:
https://github.com/spf13/cobra/blob/e94f6d0dd9a5e5738dca6bce03c4b1207ffbc0ec/command.go#L475

vishalvivekm:
@TusharMohapatra07 
Thank you for your contributions! Let's discuss this today, during Meshery Development call at 9:00 AM CT | 8:30 PM IST. Add it as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit?pli=1&tab=t.0#) || [meeting link.](https://meet.layer5.io/meshery), if you would, please.

mesheryctl/internal/cli/root/registry/update.go
```diff
@@ -55,6 +55,14 @@ mesheryctl registry update --spreadsheet-id 1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdw
 mesheryctl registry update --spreadsheet-id 1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw --spreadsheet-cred $CRED --model "[model-name]"
 	`,
 	PreRunE: func(cmd *cobra.Command, args []string) error {
+		spreadsheetID, _ := cmd.Flags().GetString("spreadsheet-id")
+		spreadsheetCred, _ := cmd.Flags().GetString("spreadsheet-cred")
+
+		if spreadsheetID == "" || spreadsheetCred == "" {
+			if err := cmd.Help(); err != nil {
+				return ErrUpdateRegistry(err, modelLocation)
```
lekaf974, line 63: The error returned should give clear explanation of the issue and how to remediate, ex: `Spreadsheet Id or Spreadsheet credentials are missing` so you could update the error or create a new one for handle this case

lekaf974 (changes requested):

mesheryctl/internal/cli/root/registry/update.go
TusharMohapatra07: @lekaf974 Please review the latest commit. Since `cmd.Help()` always returns nil as an error value, we should remove the nil check. Instead, we can display the help message directly and return our own custom error.

lekaf974: Could you share the current output please ?

lekaf974:
You need to fix DCO by signoff all your commits please

mesheryctl/internal/cli/root/registry/update.go
TusharMohapatra07: image
@lekaf974 this is the output

lekaf974: Ok this is not the current UX we have implemented in other mesheryctl command so you need to remove `cmd.Help()`

lekaf974: also you should follow following behavior https://github.com/meshery/meshery/blob/8d4be7a969eb28bf746db91b1b37ce80b4c161df/mesheryctl/internal/cli/root/model/model.go#L264

TusharMohapatra07: @lekaf974 Hey, but the issue said that it should print the help command. So, i called the cmd.Help()

TusharMohapatra07: Let's say we changed this to:
image
Then, it'll give us the same output:
image

SAHU-01:
Thank you for your contribution @TusharMohapatra07!

Could you look into the failing checks, and fix them?

Let's discuss this during the website call on Monday(6th Jan 2025) at 6:30 PM IST

Add it as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1XczAHXVe2FIWPqiF57ospJ43zw5cZQ7ui8mn39v5EvA/edit?usp=drivesdk), if you would.

TusharMohapatra07:
@SAHU-01 Yes, i'm waiting for approval on my current code so that i can proceed to fix the failing checks

mesheryctl/internal/cli/root/registry/update.go
TusharMohapatra07: The current state has same UX as this 
```
$ go run mesheryctl/cmd/mesheryctl/main.go model
```
OUTPUT:
```
Usage:
  mesheryctl model [flags]
  mesheryctl model [command]

Examples:

// To view total of available models
mesheryctl model --count

// To view list of models
mesheryctl model list

// To view a specific model
mesheryctl model view [model-name]

// To search for a specific model
mesheryctl model search [model-name]


Available Commands:
  export      export registered models
  import      Import models from mesheryctl command
  import      Import models from mesheryctl command
  list        list registered models
  search      search models
  view        view model

Flags:
      --count   (optional) Get the number of models in total
  -h, --help    help for model

Global Flags:
      --config string   path to config file (default "/home/spyd/.meshery/config.yaml")
  -v, --verbose         verbose output

Use "mesheryctl model [command] --help" for more information about a command.
Error: please provide a subcommand
exit status 1
```

lekaf974: add `return nil` after util.Log.Error(....)

lekaf974: @leecalcote do we have a clear documentation on how error should be implemented in mesheryctl somewhere ?

TusharMohapatra07: > add `return nil` after util.Log.Error(....)

@lekaf974  Okay, what about the help command logging ?

lekaf974: what is the output without help logging ?

TusharMohapatra07: @lekaf974 
```
missing required fields: spredsheet ID and credentials
unexpected end of JSON input
Error: unexpected end of JSON input
exit status 1
```
This is the output if we do 

```go
if spreadsheetID == "" || spreadsheetCred == "" {
    utils.Log.Error(fmt.Errorf("missing required fields: spredsheet ID and credentials"))
    return nil
}
```

TusharMohapatra07: @lekaf974 Yes, pushed a commit. Please check that but the logging gives this out in the end:
```
unexpected end of JSON input
Error: unexpected end of JSON input
exit status 1
```

This is because we're returning nil from the cmd.Help() (even if we explicitly return nil, it'll cause the same)

To tackle this, i suggest we put out some informative message (in the form of error).

Earlier, we weren't facing this issue because, i was calling cmd.Help() first then returning the error with message about missing id and credentials

TusharMohapatra07: Also, we cannot wrap these errors inside the meshkit errors because internally, meshkit error function is trying to access the error message and for nil instance it'll cause segmentation faults.

lekaf974:
@TusharMohapatra07 need sync with master

lekaf974 (approved):
LGTM

alphaX86 (approved):

PR #13141, opened 2025-01-03
Fix catalog content not getting seed during Meshery boot

Yashsharma1911:
**Notes for Reviewers**

#### Issue

Cloud server now returns pattern files in JSON format which we previously used to get in YAML format. and when Meshery try to read the name of pattern file during Meshery boot it is unable to do that because file is in JSON format and it expect a Yaml

#### Fix

This PR enhances the GetPatternName function to accept file in either JSON or YAML format and it will return the name of file instead of only accepting the YAML format.

- This PR fixes #


Yashsharma1911:
Issue of storing pattern files in JSON format with .yml extension still persists. One possible solution is to convert the JSON data into YAML format before storing it. The reason for this is that on meshery.io, when a file is downloaded, it is saved with a .yml extension, even though the content is in JSON format.

We can update the [script ](https://github.com/meshery/meshery.io/blob/master/assets/artifact-hub-pkg/package.go#L222) in the meshery/meshery.io repository to address this.

PR #13174, opened 2025-01-08
[UI] Fix truncated adapter selection dropdown in settings page

### LINKED ISSUE #13161: [UI] Truncated Adapter Selection Dropdown in Settings Page
### Current Behavior
The dropdown menu in the Adapters section is not fully visible, with the bottom portion of the menu being cut off or truncated. This affects the visibility and accessibility of the available adapter options.

### Desired Behavior
The dropdown menu should fully expand to display all available adapter options completely, ensuring users can view and select any option without visual obstruction.

### Screenshots/Logs
screenshot_meshery_settings_truncated_dropdown_dark
screenshot_meshery_settings_truncated_dropdown_light

### Environment

- Host OS: Mac Linux Windows
- Platform: Docker Kubernetes

---

augustmood:
**Notes for Reviewers**

- This PR fixes #13161 - [UI] Truncated Adapter Selection Dropdown in Settings Page

#### Before:
The dropdown menu in the Adapters section is cut off, preventing users from seeing all available options.
screenshot_meshery_settings_truncated_dropdown_dark
screenshot_meshery_settings_truncated_dropdown_light

#### After:
The dropdown menu now fully expands to show all options clearly, improving accessibility and usability.
screenshot_meshery_settings_fixed_dropdown_light
screenshot_meshery_settings_fixed_dropdown_dark


leecalcote:
@Venkata-Sai-Vishwanath-robo your review here would be helpful.

vishalvivekm:
@augustmood 

ui/components/ReactSelectWrapper.js
```diff
@@ -166,12 +166,18 @@ class ReactSelectWrapper extends React.Component {
         color: theme.palette.text.primary,
         '& input': { font: 'inherit' },
       }),
+      menuPortal: (base) => ({
+        ...base,
+        zIndex: 9999,
+      }),
```
Yashsharma1911, line 172: I just have one questions, if this is a reusable component then it will make this change for every other selector in UI right? this should not be a desired behaviour by default right?

augustmood: yes, this is a reusable component and it will make this change for every other selector... 
yea, I guess this might not be a desired behaviour by default.

augustmood: btw, I will try to find other way to solve this issue.

augustmood: By looking at the original source code again, I notice that the high z-index approach was already implemented in the paper styles (and this was intended to make the dropdown/select overlay other components right?)

```
  paper: {
    zIndex: 9999,
    width: '100%',
    position: 'absolute',
  },
```

And the dropdown is truncated is due to this z-index is not work as intended, to make it work, I feel the Portal configuration is necessary:

```
menuPortalTarget={document.body}
menuPosition="fixed"
```

```
      menuPortal: (base) => ({
        ...base,
        zIndex: 9999,
      }),
```

So rather than introducing a different/new solution, my implementation here just completes the original approach's intent, what do you guys think? @vishalvivekm @Yashsharma1911

leecalcote:
@Venkata-Sai-Vishwanath-robo your review here would be helpful.

ui/components/ReactSelectWrapper.js
Venkata-Sai-Vishwanath-robo: This looks good, but it seems redundant to set the zIndex twice. If I’m not mistaken, simply adding menuPosition="fixed" should achieve the same result.

While testing in Firefox at the default 100% zoom, I noticed that the last dropdown options were still hidden. However, when zooming out to 90% or less, the options became visible.

Video for reference : 
[Screencast from 2025-01-15 06-58-33.webm](https://github.com/user-attachments/assets/e9ffed70-b7d9-4612-9b17-25d8343bde53)

augustmood: Thanks for pointing that out! I’ll take a closer look. I don’t think it’s specific to Firefox—it seems more related to the window or screen size. I noticed that when the screen is too small, the MenuList isn’t fully visible. Let me see if we need to make the height of the dropdown MenuList adaptive or adjust its behavior. I’ll follow up once I’ve had a chance to look into it.

augustmood: Screenshot 2025-01-17 at 10 55 57 AM
I just fixed it, now it works across various window sizes.

PR #13212, opened 2025-01-12
[UI] Kubernetes Deployment Operation Debug Logging

leecalcote:

### DIFF
--- a/ui/components/Header.js
+++ b/ui/components/Header.js
@@ -39,7 +39,6 @@ import {
   Box,
   CustomTooltip,
   Typography,
-  useTheme,
   styled,
 } from '@layer5/sistent';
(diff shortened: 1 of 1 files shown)


ui/components/Header.js
```diff
@@ -321,10 +330,45 @@ function K8sContextMenu({
...
+          </Typography>
+          <Typography
+            variant="body2"
+            sx={{ textAlign: 'left', marginTop: '1rem', cursor: 'pointer' }}
+          >
+            <details>
+              <summary>
+                <strong>What does this mean?</strong>
+              </summary>
+              <div>
```
leecalcote, line 349: It would be nice to have a slightly different background color inside of the details. I tried adding one, but...

leecalcote: ...I could use some help, @aabidsofi19

aabidsofi19: i will update this .

PR #13222, opened 2025-01-13
feat: enhance registry publish cmd to send models to mesheryio

vishalvivekm:
**Notes for Reviewers**

- This PR fixes #

### DIFF
--- a/mesheryctl/pkg/utils/registry.go
+++ b/mesheryctl/pkg/utils/registry.go
@@ -47,6 +47,7 @@ func GetIndexForRegisterCol(cols []string, shouldRegister string) int {
 }
 
 func GenerateMDXStyleDocs(model ModelCSV, components []ComponentCSV, modelPath, imgPath string) error {
+	// ../layer5/src/collections/integrations ../layer5/src/collections/integrations
 	modelName := utils.FormatName(model.Model)
 	// create dir for model
 	modelDir, _ := filepath.Abs(filepath.Join("../", modelPath, modelName))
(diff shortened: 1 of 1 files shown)


mesheryctl/pkg/utils/model.go
```diff
@@ -412,10 +412,12 @@ func (m ModelCSV) CreateJSONItem(iconDir string) string {
 	return json
 }
 
-func (m ModelCSV) CreateMarkDownForMDStyle(componentsMetadata string) string {
+func (m ModelCSV) CreateMarkDownForMDStyle(componentsMetadata, relationshipMetadata string, componentsCount, relationshipsCount int, outputFor string) string {
```
lekaf974, line 415: Could we consider to pass a struct here instead of adding arguments to the method, that way if we want to add additional data we can add fields in the struct without modify the method signature.

PS: Not a blocker and it can be done in another PR as improvement

mesheryctl/pkg/utils/registry.go
```diff
@@ -142,13 +146,42 @@ func GenerateJSStyleDocs(model ModelCSV, docsJSON, imgPath string) (string, erro
...
+	_, err = io.WriteString(file, md)
+	if err != nil {
+		return "", err
+	}
+
 	return docsJSON, nil
 }
 
-func GenerateMDStyleDocs(model ModelCSV, components []ComponentCSV, modelPath, imgPath string) error {
+func GenerateMDStyleDocs(model ModelCSV, components []ComponentCSV, relationships []RelationshipCSV, modelPath, imgPath string) error {
```
lekaf974, line 180: Samr comment related to method signature change

lekaf974 (approved):
LGTM

PR #13254, opened 2025-01-14
fix(ui): improve input field behavior in performance settings

### LINKED ISSUE #13232: [UI] Input Field Editing for 'Concurrent Requests' and 'Queries Per Second' is Unintuitive
### Current Behavior

While trying to edit the "Concurrent requests" or "Queries per second" text fields in the User Preferences -> PERFORMANCE, they do not allow the user to delete the existing value using the backspace key. To enter a new value, the user must move the cursor to the left of the existing value and then enter and delete the right-most number or use the arrow keys to increase the count.

### Desired Behavior

These text fields should allow users to delete the existing value (e.g., 0) entirely using the backspace or delete keys. If the field is left empty, it should handle validation appropriately, such as reverting to a default value (e.g., 0) upon losing focus or when saving

### Screenshots/Logs

image

https://github.com/user-attachments/assets/67a60e84-15e1-45d4-a9f7-6ffe847ca9c6

### Environment

- Host OS: Mac Linux Windows
- Platform: Docker Kubernetes

---

augustmood:
### Input Type Change

- Changed input type from "number" to "text" while maintaining numeric validation
- Rationale: Native "number" inputs don't allow for empty states in a user-friendly way, making it impossible to clear values completely

### Validation Logic

- Added proper empty string handling while maintaining numeric constraints:
  ```javascript
  if (value === '') {
    setC(null);
  } else {
    const parsedValue = parseInt(value, 10);
    if (!isNaN(parsedValue)) {
      setC(parsedValue);
    }
  }
  ```
- Added `inputMode="numeric"` and `pattern="[0-9]*"` to maintain numeric keyboard on mobile while allowing for empty states
- Added blur handler to restore default value (0) when field is left empty

#### Before:

https://github.com/user-attachments/assets/9eb18250-3334-4887-8e66-2d9ff2aeb82a

#### After:

https://github.com/user-attachments/assets/9554e941-369c-4c5b-9c48-02d1c83151a1

**Notes for Reviewers**

- This PR fixes #13232

### DIFF
--- a/ui/components/MesherySettingsPerformanceComponent.js
+++ b/ui/components/MesherySettingsPerformanceComponent.js
@@ -160,13 +160,31 @@ const MesherySettingsPerformanceComponent = (props) => {
                 id="c"
                 name="c"
                 label="Concurrent requests"
-                type="number"
+                type="text"
                 fullWidth
-                value={c}
-                inputProps={{ min: '0', step: '1' }}
+                value={c === null ? '' : c}
+                inputProps={{ 
+                  inputMode: 'numeric',
+                  pattern: '[0-9]*'
+                }}
                 margin="normal"
                 variant="outlined"
-                onChange={handleChange('c')}
+                onChange={(event) => {
+                  const value = event.target.value;
+                  if (value === '') {
+                    setC(null);
+                  } else {
+                    const parsedValue = parseInt(value, 10);
+                    if (!isNaN(parsedValue)) {
+                      setC(parsedValue);
+                    }
+                  }
+                }}
+                onBlur={(event) => {
+                  if (c === null) {
+                    setC(0);
+                  }
+                }}
                 InputLabelProps={{ shrink: true }}
               />
             </Grid>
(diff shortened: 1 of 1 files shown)


vishalvivekm:
@augustmood 

leecalcote:
Thanks for digging into this. The issue here is that in both of these cases, these are required fields (meaning that having no numeric value supplied is not an option) and zero is meaningful to the underlying low generator which interprets zero request per second to be Max, meaning, as many as it can possibly generate.  A value of zero for the concurrent request means that the generator will be single threaded.

augustmood:
> Thanks for digging into this. The issue here is that in both of these cases, these are required fields (meaning that having no numeric value supplied is not an option) and zero is meaningful to the underlying low generator which interprets zero request per second to be Max, meaning, as many as it can possibly generate. A value of zero for the concurrent request means that the generator will be single threaded.

Thanks for the points about the required fields and zero values. We're still enforcing numeric input, just changing how we handle it. Instead of using `type="number"` which has some quirks in how it behaves, we're switching to `type="text"` with proper numeric validation implemented. When the field is emptied, it still defaults to 0, maintaining the same behavior as before. This gives us better input behavior while keeping all the same validation requirements.

PR #13263, opened 2025-01-15
 connections page migrate from muiv4 to sistent(v5)

sudhanshutech:
**Notes for Reviewers**

 This PR fixes #
- Connection and meshsync
- Metadata components
- Registeration model(register any connection from meshesync)
- All connection states

### DIFF
--- a/ui/components/DataFormatter/index.js
+++ b/ui/components/DataFormatter/index.js
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Tooltip, Typography, Grid, Box, IconButton, useTheme } from '@material-ui/core';
+import { CustomTooltip, Typography, Grid, Box, IconButton, useTheme } from '@layer5/sistent';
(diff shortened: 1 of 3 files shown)


ui/components/connections/styles.js
```diff
@@ -242,6 +244,168 @@ const styles = (theme) => ({
...
+  },
+  background: `${theme.palette.background.success.default}30 !important`,
+  '& .MuiSvgIcon-root': {
+    color: `${theme.palette.background.success.default} !important`,
+  },
+}));
+
+export const IgnoredChip = styled(Chip)(({ theme }) => ({
+  '& .MuiChip-label': {
+    color: theme.palette.mode === 'dark' ? '#9FAFB6' : '#51636B',
```
sudhanshutech, line 323: currently these particular state color are not available in sistent that's why kept as hard code names. Will later on add it to sistent so we can use it as tokens.

PR #13299, opened 2025-01-17
Revert "migrate navigator components to sistent"

sudhanshutech:
Reverts meshery/meshery#13271


amitamrutiya:
@sudhanshutech can you confirm the reason for this revert?

sudhanshutech:
oh yes, everything was fine but the click was not working means when we click on any section in navigator to go to any page its was not working.

amitamrutiya:
Okay, I'll look and revert this PR.

PR #13316, opened 2025-01-21
Hide api version from the meshsync resource data table

amitamrutiya:
**Notes for Reviewers**

- This PR fixes #

### DIFF
--- a/docs/_releases/v0.8.19.md
+++ b/docs/_releases/v0.8.19.md
@@ -1,0 +1,30 @@
+---
+layout: release
+date: 2025-01-21
+tag: v0.8.19
+prerelease: false
+---
+
+## What's New
+## 🔤 General
+- add: endpoints to set and get log levels while server is running @souvikinator (#13290)
+- Support for K8s manifest kind list while importing @souvikinator (#13233)
+
+## ⌨️ Meshery CLI
+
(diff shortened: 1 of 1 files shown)


ui/components/DashboardComponent/resources/configuration/config.js
```diff
@@ -60,24 +59,6 @@ export const ConfigurationTableConfig = (
             },
           },
         },
-        {
-          name: 'apiVersion',
```
aabidsofi19, line 64: Don't remove completely , only hide in default column selection

ui/components/DashboardComponent/resources/namespace/config.js
```diff
@@ -60,24 +59,6 @@ export const NamespaceTableConfig = (
           },
         },
       },
-      {
-        name: 'apiVersion',
```
aabidsofi19, line 64: Why are we redining tables for each category

aabidsofi19 (changes requested):

ui/components/DashboardComponent/resources/namespace/config.js
amitamrutiya: Yeah, that's true, For the Crds table i made it dynamic but for other than all for each category we are defining a new table.

PR #13341, opened 2025-01-23
Optimize and enhance matchlabel policy and general evaluation 

aabidsofi19:
**Notes for Reviewers**

- This PR fixes #

- No longer generate duplicate matchlabel relationships for same field and value
- No longer generate matchlabel relationships for each pair but group all component with same field value into single relationship
- Modularize code into separate package ( core_utils, feasibility_evaluation, etc ) to prevent accidental function shadowing and  conflicts 
-  Add Support for evaluating separate rules and policies without having conflicts
-  Cap total number of generated/identified matchlabel relationships
-  Cleanup deleted matchlabel relationships

### DIFF
--- a/server/meshmodel/meshery-core/0.7.2/v1.0.0/policies/alias_policy.rego
+++ b/server/meshmodel/meshery-core/0.7.2/v1.0.0/policies/alias_policy.rego
@@ -28,9 +28,9 @@ import data.feasibility_evaluation_utils
 #    - Selector configuration that specifies:
 #      • 'from': The child (alias) component
 #      • 'to': The parent component being referenced
-#    - Single-item mutatorRef/mutatedRef array where index[0] points to 
+#    - Single-item mutatorRef/mutatedRef array where index[0] points to
 #      the child component field acting as the alias
-#    
+#
 #    Note: New relationships start with "pending" status
 #
 # 3. Actions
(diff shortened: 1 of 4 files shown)


aabidsofi19:
single tagset relationship can span more than 2 components here ( 3 in this example) : 
screenshot-2025-01-23-14-55-56

No duplicate relationships ( 4 matching fields , for relationships ) : 
screenshot-2025-01-23-14-58-32

Jougan-0:
Looks good to me @aabidsofi19 I see 6 bubbles in total but I think it should be 3 or atleast 1 matchlabel btw 2 components.
Thoughts?

server/meshmodel/meshery-core/0.7.2/v1.0.0/policies/alias_policy.rego
```diff
@@ -35,66 +49,29 @@ package relationship_evaluation_policy
 #        - Remove the alias component from the design file.
 #        - Remove the aliased configuration from the parent component.
 
-import rego.v1
-
 is_alias_relationship(relationship) if {
 	lower(relationship.kind) == "hierarchical"
 	lower(relationship.type) == "parent"
 	lower(relationship.subType) == "alias"
 }
```
leecalcote, line 56: It is unlikely, that Meshery has a use case for supporting relationship.type == "child" aliases in the future.

leecalcote:
@aabidsofi19 is `evaluation.rego` still the entry point for all evaluations?

aabidsofi19:
> @aabidsofi19 is `evaluation.rego` still the entry point for all evaluations? I have added Rego METADATA specifying it as an `entrypoint`.

yeah it is the entrypoint , right now it cotains both the old eval flow ( for inventor, binding , network policies) and the new flow for the alias and matchlabels policies

PR #13470, opened 2025-02-04
Fixed the alignment issue of checkboxes in the Environments card

PR4NJ41:
**Notes for Reviewers**

- This PR fixes the alignment issue of checkboxes in the Environments card.

# Screenshots

## Before
Before

## After
After


yash37158:
Hi @PR4NJ41, could you please check on Slack first to confirm the relevancy of the issue? If a fix is needed, you can then proceed with opening an issue and a PR. That would be greatly appreciated!

miacycle (approved):
Using `rem` - lgtm

miacycle:
Thank you, @yash37158. 👍 I'm having trouble discerning what the changes since it's difficult to pick out in the before and after screenshots. Where is this change, @PR4NJ41?

PR4NJ41:
@yash37158 Sure, I will keep that in check while doing PRs. 

@miacycle If you look closely the checkboxes in the before image are little away from center and aren't aligned well with the labels. While in the after image it is aligned properly. 

The back of the flip cards are correct in [this](https://playground.meshery.io/management/workspaces). But are misaligned in [this](https://playground.meshery.io/management/environments).

I thought it was a bug so made the PR.

A more closer look:
Before:
Before

After:
After

ui/components/Lifecycle/Environments/styles.js
```diff
@@ -136,6 +136,7 @@ export const Record = styled('div')(() => ({
 export const BulkSelectCheckbox = styled(Checkbox)({
   padding: 0,
   marginRight: '0.5rem',
+  height: '1.75rem',
```
sudhanshutech, line 139: don't use height, use align center or text align styles to align them, otherwise this will create issues again.

sudhanshutech (changes requested):

ui/components/Lifecycle/Environments/styles.js
PR4NJ41: I had already tried this earlier, but it didn't work out. You can check the design of the other card, which uses a similar box—it is implemented the same way I did.

sudhanshutech (approved):

Issue #13622, opened 2025-02-14
[mesheryctl] Improve `registry generate` error message when user is not in cloned repo


leecalcote:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

### Current Behavior

When the `mesheryctl registry generate` command is executed outside of the root of a cloned `meshery/meshery` repo, the following error is message shown:

```
Error generating model: loxilb
 mkdir /Users/server: permission denied
```

### Desired Behavior
1. Users executing this command often do have permission to create a folder. The logic for this behavior needs improved.
2. A more informative error message with instructions on what to do should be offered.

---

### Contributor Resources

- ⌨️ [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)

leecalcote:
@riyaa14 fyi

kushal9897:
Hi @leecalcote, I'd like to work on this issue. Could you please assign it to me? Thanks!

riyaa14:
@kushal9897 sure, you can give it a shot.

riyaa14:
@leecalcote this command works fine for cases when we aren't at root of a forked repo. It only occurs when we are outside the repo.

Shall we remove the 'root' check? 
A problem with root check is, `model-generator.yaml` workflow would fail since it doesn't execute from root, it executes here -  `meshery/mesheryctl`

// @kushal9897

leecalcote:
@riyaa14, you're right. Please lead the way.

leecalcote:
@riyaa14 given our verbal review, where did we land on the final desired behavior?

PR #13634, opened 2025-02-17
 add repo root check in generate command

kushal9897:
**Notes for Reviewers**

- This PR fixes #
#13622
- [ *] Yes, I signed my commits.

### DIFF
--- a/provider-ui/package-lock.json
+++ b/provider-ui/package-lock.json
@@ -19,7 +19,7 @@
         "http-proxy": "^1.18.1",
         "isomorphic-unfetch": "^4.0.2",
         "mui-datatables": "^4.3.0",
-        "next": "^13.5.4",
+        "next": "^15.2.4",
         "prop-types": "^15.8.1",
         "react": "^18.3.1",
         "react-dom": "^18.3.1"
@@ -311,6 +311,16 @@
       },
       "engines": {
         "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@emnapi/runtime": {
+      "version": "1.4.0",
+      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.4.0.tgz",
+      "integrity": "sha512-64WYIf4UYcdLnbKn/umDlNjQDSS8AgZrI/R9+x5ilkUVFxXcA1Ebl+gQLc/6mERA4407Xof0R7wEyEuj091CVw==",
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "tslib": "^2.4.0"
       }
     },
     "node_modules/@emotion/babel-plugin": {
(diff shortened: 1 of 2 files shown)


mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -63,6 +63,16 @@ mesheryctl registry generate --directory <DIRECTORY_PATH>
...
 		// Prerequisite check is needed - https://github.com/meshery/meshery/issues/10369
 		// TODO: Include a prerequisite check to confirm that this command IS being the executED from within a fork of the Meshery repo, and is being executed at the root of that fork.
+		// Prerequisite check: ensure that the command is executed from the root of a cloned meshery/meshery repo.
+		wd, err := os.Getwd()
+		if err != nil {
+			return errors.New("unable to determine current working directory")
+		}
+		// Check for the presence of the "server" directory as a marker for the repo root.
+		if _, err := os.Stat(filepath.Join(wd, "server")); os.IsNotExist(err) {
+			return errors.New("Error: it appears you are not running this command from the root of a cloned meshery/meshery repository (missing 'server' directory).\nPlease change directory to the repository root and try again.\nFor more information, see https://github.com/meshery/meshery/blob/master/README.md")
```
leecalcote, line 73: Thanks, @kushal9897. Please instead point to the contributing guide for models.

leecalcote (changes requested):

vishalvivekm:
@kushal9897 

riyaa14:
@kushal9897 I have added a comment, kindly look into that. Thanks :)

kushal9897:
Hey @riyaa14, I’m not seeing any comment on my end. Could you please check and let me know? Thanks!

riyaa14 (changes requested):

riyaa14:
This is weird 😅, anyways, can you see it now?

kushal9897:
Yeah, I can see it now! Thanks for the heads-up

kushal9897:
Hello @riyaa14  When I changed the error package it gave me the error

kushal9897:
Hey @riyaa14 , could you please review my latest PR? Let me know if any changes are required. Happy to address them!

mesheryctl/internal/cli/root/registry/generate.go
```diff
@@ -50,7 +50,7 @@ var generateCmd = &cobra.Command{
 	Use:   "generate",
 	Short: "Generate Models",
 	Long:  "Prerequisite: Excecute this command from the root of a meshery/meshery repo fork.\n\nGiven a Google Sheet with a list of model names and source locations, generate models and components any Registrant (e.g. GitHub, Artifact Hub) repositories.\n\nGenerated Model files are written to local filesystem under `/server/models/<model-name>`.",
-	Example: `
+	Example: `fu
```
riyaa14, line 53:
```suggestion
	Example: `
```

this seems like a typo?

kushal9897:
Hey @riyaa14,

I reviewed the changes and confirmed that removing the root check doesn’t affect the functionality the error now works fine. Based on your feedback, I’ve updated the PR  Could you please review the latest commit and let me know if there’s anything else you’d like adjusted?

riyaa14:
@kushal9897 please cross check your changes, they don't solve the issue. According to your changes, if we can't fetch the current working directory, then it would say -> `Error: it appears you are not running this command from the root of a cloned meshery/meshery ...`, which doesn't make sense.

Also, I checked that we don't need to be inside of a cloned meshery repo, this command is throwing an error as shown in issue because users by default don't have permission to create directory in `Users/` directory.

**Behaviour explanation** 
It is preferable to execute this command from `meshery/mesheryctl` directory, but the command would work fine even if we are outside the repo.
Why `meshery/mesheryctl` directory is preferable?
- If the user is generating a model a new model, this command goes to `../server/meshmodel` from the current working directory and checks if the model is already present so that it can skip its generation, else model is generated.
- At `meshery/mesheryctl`, we reach the correct location which is `meshery/server/meshmodel`

**Approach**
Considering all stated above, the approach should be -
- check if user is at preferred directory in a **cloned repo** i.e at `meshery/mesheryctl` repo, if not we can send a WARNING (not an error) 
- permission issues at `Users/` directory: for such cases, users should know where the directory is being created, a log like this would help: "creating <directory_name> at <path> ...."

Do let me know if you have any feedback or queries.

kushal9897:
Sorry for the error, I appreciate you pointing it out. I'll work on fixing it right away. If I have any questions along the way, I'll be sure to reach out.

riyaa14:
@kushal9897 Here's the functionality that we would be ideal

1. If user isn't at `meshery/mesheryctl` directory, we mention the directory at which their model is getting generated, i.e, 
> "It is recommended to use this command from the mesheryctl directory of cloned meshery repo. You can also specify a particular directory using --output flag. Currently using default directory - `dir-path` ..."

Please use `RegistryError` for this message.

2. "server directory" check works but isn't ideal, we should implement something like this one -> https://github.com/meshery/meshery/pull/10490

Could you please take these points into account? Thanks!

riyaa14:
@kushal9897 wanted to follow up on this work. Any problems you're facing?

kushal9897:
@riyaa14, i will make the new pr soon

PR #13655, opened 2025-02-18
[Header.js] Convert class component to functional component

### LINKED ISSUE #8721: [UI] Convert `Header.js` to Functional Component
Original Issue: https://github.com/meshery/meshery/issues/7187

Tracking another issue for visibility: https://github.com/meshery/meshery/issues/8668

- [ ] Move `Header.js` to its own directory
- [ ] Break up `Header.js` and `K8sContextMenu` to its own exports
- [ ] Remove unused file `Headerstyle.js`

Future items for tracking (not necessary to complete this ticket)
- [ ] Identity the redux states
- [ ] Identity the static styles
- [ ] Identity the custom components

---

falgunmpatel:
**Notes for Reviewers**

- This PR fixes #8721

### DIFF
--- a/ui/components/Header.js
+++ b/ui/components/Header.js
@@ -473,16 +473,16 @@ const Header = ({
                 <UserInfoContainer>
                   <UserSpan style={{ position: 'relative' }}>
                     <K8sContextMenu
-                      contexts={this.props.contexts}
-                      activeContexts={this.props.activeContexts}
-                      setActiveContexts={this.props.setActiveContexts}
-                      searchContexts={this.props.searchContexts}
+                      contexts={contexts}
+                      activeContexts={activeContexts}
+                      setActiveContexts={setActiveContexts}
+                      searchContexts={searchContexts}
                       runningStatus={{
-                        operatorStatus: this.props.operatorState,
-                        meshSyncStatus: this.props.meshSyncState,
+                        operatorStatus: operatorState,
+                        meshSyncStatus: meshSyncState,
                       }}
-                      updateK8SConfig={this.props.updateK8SConfig}
-                      updateProgress={this.props.updateProgress}
+                      updateK8SConfig={updateK8SConfig}
+                      updateProgress={updateProgress}
                     />
                   </UserSpan>
 
@@ -499,13 +499,10 @@ const Header = ({
                   </div>
 
                   <UserSpan>
-                    <User color="inherit" updateExtensionType={this.props.updateExtensionType} />
+                    <User color="inherit" updateExtensionType={updateExtensionType} />
                   </UserSpan>
                   <UserSpan>
-                    <HeaderMenu
-                      color="inherit"
-                      updateExtensionType={this.props.updateExtensionType}
-                    />
+                    <HeaderMenu color="inherit" updateExtensionType={updateExtensionType} />
                   </UserSpan>
                 </UserInfoContainer>
               </Box>


leecalcote:
Thanks, @falgunmpatel. 👍 Please don't forget to sign your commits. For this particular PR, you can follow these instructions - https://github.com/meshery/meshery/pull/13655/checks?check_run_id=37421679691

ui/components/HeaderMenu.js
```diff
@@ -68,29 +68,37 @@ const HeaderMenu = (props) => {
...
+  }, [isTokenError]);
+
+  useEffect(() => {
+    if (!capabilitiesLoadedRef.current && capabilitiesRegistry) {
+      capabilitiesLoadedRef.current = true;
+      setAccount(
+        ExtensionPointSchemaValidator('account')(capabilitiesRegistry?.extensions?.account),
+      );
+    }
+  }, [capabilitiesRegistry]);
```
amitamrutiya, line 101: Why this changes needed @falgunmpatel

ui/rtk-query/ability.js
```diff
@@ -32,10 +33,19 @@ export const useGetUserAbilities = (org, skip) => {
...
   }
+
+  useLayoutEffect(() => {
+    if (res?.keys) {
+      requestAnimationFrame(() => {
+        setKeys({ keys: res.keys });
+      });
+    }
+  }, [res?.keys, setKeys]);
+
```
amitamrutiya, line 48: Why this changes as well

amitamrutiya (changes requested):

ui/components/HeaderMenu.js
amitamrutiya: like adding the useEffects

falgunmpatel: Screenshot 2025-02-19 at 11 28 21 AM

Checkout this error. 

- The component updates state during rendering, which causes a re-render. 
- Wrapping the state updates inside useEffect ensures that they only happen when needed, not on every render.

Also, this resolves this error. Similary, in other components as well.

ui/rtk-query/ability.js
falgunmpatel: - To resolve the error similarly in the SpaceSwither.js, I tried using useEffect as well.
- Issue with useEffect: When I used useEffect, the state update (setKeys) happened too late—after the browser had already painted the UI. This delay caused the HoneyComb component on the Dashboard to load infinitely because it was rendering with outdated state.
- This delay made the HoneyComb component on the Dashboard load infinitely and not show the "connect clusters" button.
- Why useLayoutEffect: useLayoutEffect runs synchronously after DOM mutations and before the browser paints. This means the state update happens immediately, ensuring that the UI is in the correct state before it is rendered.
- By switching to useLayoutEffect, the state update happens immediately after the DOM changes and before the UI is painted.
- As a result, the HoneyComb component stops loading infinitely when no clusters are attached and correctly shows the "create cluster" button.

This resolves all the errors related to this specific component. 

Screenshot 2025-02-19 at 11 56 30 AM


Screenshot 2025-02-19 at 11 54 11 AM

falgunmpatel: - To resolve the error similarly in the SpaceSwither.js, I tried using useEffect as well.
- Issue with useEffect: When I used useEffect, the state update (setKeys) happened too late—after the browser had already painted the UI. This delay caused the HoneyComb component on the Dashboard to load infinitely because it was rendering with outdated state.
- This delay made the HoneyComb component on the Dashboard load infinitely and not show the "connect clusters" button.
- Why useLayoutEffect: useLayoutEffect runs synchronously after DOM mutations and before the browser paints. This means the state update happens immediately, ensuring that the UI is in the correct state before it is rendered.
- By switching to useLayoutEffect, the state update happens immediately after the DOM changes and before the UI is painted.
- As a result, the HoneyComb component stops loading infinitely when no clusters are attached and correctly shows the "create cluster" button.

This resolves all the errors related to this specific component. 

Screenshot 2025-02-19 at 11 56 30 AM


Screenshot 2025-02-19 at 11 54 11 AM

ui/components/User.js
```diff
@@ -41,21 +42,27 @@ const User = (props) => {
...
+        details: getUserError?.data,
+      });
+    }
+  }, [userData, isGetUserSuccess, isGetUserError]);
 
-  if (!capabilitiesLoadedRef.current && capabilitiesRegistry) {
-    capabilitiesLoadedRef.current = true;
-    setAccount(ExtensionPointSchemaValidator('account')(capabilitiesRegistry?.extensions?.account));
-  }
+  useEffect(() => {
```
falgunmpatel, line 58: Seems like useEffect is not needed here!

ui/components/HeaderMenu.js
```diff
@@ -67,29 +67,37 @@ const HeaderMenu = (props) => {
...
+      props.updateUser({ user: userData });
+      setUserLoaded(true);
+    } else if (isGetUserError) {
+      notify({
+        message: 'Error fetching user',
+        event_type: EVENT_TYPES.ERROR,
+        details: getUserError?.data,
+      });
+    }
+  }, [userData, isGetUserSuccess, isGetUserError]);
```
falgunmpatel, line 81: **@aabidsofi19**  

I'm encountering an issue with the `updateUser` action creator on **lines 72 and 73**. The error seems to be directly related to its execution.  

- The same issue occurs in the `User.js` and `HeaderMenu.js` components.  
- Both components have a similar implementation where `updateUser` is being called when `isGetUserSuccess` is `true`.  
- So far, the only workaround I can think of is wrapping it inside `useEffect`, but I'm unsure if that's the best approach.  

Do you have any suggestions on how to properly handle the `updateUser` action here? Any insights would be really helpful!

falgunmpatel: However, we can safely remove unnecessary useEffects from other places.

falgunmpatel: - Updating the state triggers a new render, and React was already in the middle of a render when it encountered setUserLoaded(true). 
- The function re-executes, but React does not batch updates immediately. Therefore userLoaded will still be false in the next re-render. Causing infinite loop.
- This causes React to re-render the component immediately, which might lead to an infinite loop.

falgunmpatel: - Redux action triggers a re-render in a React component only if the action changes the state in the Redux store and the component is subscribed to that part of the state using useSelector or connect.
- Here, updateUser changes the state in the redux. Hence, creates a re-render.

ui/components/Header.js
```diff
@@ -381,16 +380,26 @@ function K8sContextMenu({
...
+  updateK8SConfig,
+  updateProgress,
+  updateCapabilities,
+  updateExtensionType,
+}) => {
+  const [collaboratorExt, setCollaboratorExt] = useState(null);
+
+  useEffect(() => {
     dataFetch(
       '/api/provider/capabilities',
```
amitamrutiya, line 404: use the rtk query so no need for extra useEffect or extra state variables

amitamrutiya (changes requested):

amitamrutiya (approved):

Issue #13656, opened 2025-02-18
[UI] Dashboard: Customized layout not honored on render


leecalcote:
### Current Behavior
I see that when I Edit my layout, my requested positioning and sizing is honored, however, when I am viewing the page, my latest layout preferences are not honored.

### Screenshots/Logs

https://github.com/user-attachments/assets/7afe8f0d-e417-4f54-b376-dbe8dbec6744

### Environment

- Browser: Safari 
- Host OS: Mac 
- Meshery Server Version: stable-v0.8.32

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)

leecalcote:
@PR4NJ41 FYI

PR4NJ41:
@leecalcote 
This issue arises due to inconsistent size differences between Edit mode and Non-Edit mode. The side panel, from which widgets are selected, is causing this effect.

I found that when changing the between the two modes, the Breakpoints are changing and hence causing this effect.

![Image](https://github.com/user-attachments/assets/e16681a7-550a-4a2a-abe2-74d2e45b99cb)

leecalcote:
Good work, @PR4NJ41! Sounds like a PR is imminent.

amitamrutiya:
@PR4NJ41 are you working on the fix for a smaller screen?

PR4NJ41:
@amitamrutiya Yes, I am

PR #13674, opened 2025-02-20
fix: responsive issue for the dashboard

amitamrutiya:
The issue is that the widget panel that opens creates the difference between the open and closed layout. So I changed the position of the widget panel and now it is working fine.


PR4NJ41:
@amitamrutiya 
This might also fix #13656
If not then i will fix it.

amitamrutiya:
Yeah it fixed so you can work on other issue

leecalcote:
@M-DEV-1 @Karan-Palan

PR #13718, opened 2025-02-22
fix: removed unused fetch util

ianrwhitney:
**Notes for Reviewers**

- This part of a multi-step process to fix the env setup in our playwright tests. This includes modifying the ENV.* usages which this util is using. Therefore, since this util is unused right now we will delete it.  

Context: 
In current playwright setup we are using storageState. This handles authentication by storing session cookies and local stroage values across tests. This means when you launch a new browser context w/ this storageState all stored cookies including auth tokens are present. 

We would need this funciton if we are trying to make api calls outside of the playwright tests. If we are doing this then we should move this util somewhere else 

Screenshot 2025-02-22 at 7 55 31 AM


Yashsharma1911 (approved):

Yashsharma1911:
Thankyou @ianrwhitney I agree, I think we already have such fetch function already which we can reuse if required, we can safely remove this

PR #13721, opened 2025-02-22
fix linting issue in provider ui

falgunmpatel:
**Notes for Reviewers**

- This PR fixes #

### DIFF
--- a/ui/components/DataFormatter/index.js
+++ b/ui/components/DataFormatter/index.js
@@ -369,7 +369,13 @@ export const FormatStructuredData = ({ propertyFormatters = {}, data, uiSchema, 
           propertyFormatters: propertyFormatters,
         }}
       >
-        <Grid container>
+        <Grid
+          container
+          style={{
+            wordBreak: 'break-word',
+            overflowWrap: 'break-word',
+          }}
+        >
           <DynamicFormatter data={data} uiSchema={uiSchema} isLevel={isLevel} />
         </Grid>
       </FormatterContext.Provider>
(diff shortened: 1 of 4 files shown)


falgunmpatel:
@leecalcote I’ve fixed the linting issues in the provider UI here.

Since my previous PR was for documentation, I created a separate PR for this fix. Would it be better to include such fixes in the original PR or keep them separate?

Thanks!

leecalcote:
Separate is ideal. 👍

leecalcote (changes requested):
Thank you for this, @falgunmpatel. Instead of fixing the linting errors in this way, will you please update the listing rule to be that of `key: value` with no whitespace after the `key`?

The current lint style is undesired.

falgunmpatel:
> Thank you for this, @falgunmpatel. Instead of fixing the linting errors in this way, will you please update the listing rule to be that of `key: value` with no whitespace after the `key`?
> 
> The current lint style is undesired.

Sure!

falgunmpatel:
Hey @Lee, I noticed that we aren’t using Prettier for the Provider UI. Is there a specific reason for that? If we were, we could align the ESLint configuration across both. If there are no plans to integrate Prettier, I can adjust some ESLint rules for now. Let me know your thoughts!"

falgunmpatel:
@leecalcote I’ve updated the ESLint configurations. I referenced the ESLint setup from the "UI" package for guidance, and it should be good now.

Let me know if you’d like any further updates! 🚀

leecalcote:
No, there’s no reason that we’re not using prettier here.

falgunmpatel:
@amitamrutiya
I need some help with fixing these failing E2E tests. Could you provide some guidance on how to resolve them?

falgunmpatel:
@amitamrutiya

I’ve fixed all ESLint issues and added consistency checks similar to meshery-ui. 

Let me know if any further refinements are needed. It's ok, if these are not worth merging yet.

leecalcote (approved):

leecalcote:
Good!

Issue #12956, opened 2024-12-23
Unable to remove/delete Prometheus connection on one attempt


sangramrath:
Probably related issue #12957 

### Current Behavior
Unable to remove / delete Prometheus connection from `settings?settingsCategory=Metrics&tab=Prometheus` in one attempt.
The first attempt throws an error **Unable to update connection status** and a second attempt is required to remove it. (screenshot of error attached)

### Expected Behavior
User should be able to remove the Prometheus connection in one attempt.

### Screenshots/Logs
![Screenshot 2024-12-23 at 10 14 13 PM](https://github.com/user-attachments/assets/3d73a15a-63b3-4996-9076-1f4ad5374c75)

### Environment

- **Host OS:** Linux
- **Platform:** Kubernetes
- **Meshery Server Version:** stable-v
- **Meshery Client Version:** stable-v

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#documentation-contribution-flow)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](http://discuss.meshery.io) and [Community Slack](https://slack.meshery.io)

codeknight03:
Working on this one.

PR #13814, opened 2025-03-01
Fixes the credential save for local provider.

codeknight03:
- The identified issue with the local provider was the use of map type which was not recognized by Gorm.
- The fix was to move to a pre-established type.

**Impacted Places**
- Credentials API 
- Grafana Connection 
- Prom Connection 
- State Machines

**Tests Done**
- Test adding credentials directly via API. It worked
- Test with remote provider. The new struct did not disrupt remote provider's flow.
- Test with local provider by adding prometheus connection and grafana connection. 

**Notes for Reviewers**

- This PR paritaly fixes #https://github.com/meshery/meshery/issues/12956

### DIFF
--- a/server/models/default_local_provider.go
+++ b/server/models/default_local_provider.go
@@ -1324,11 +1324,27 @@ func (l *DefaultLocalProvider) Cleanup() error {
 }
 
 func (l *DefaultLocalProvider) SaveUserCredential(token string, credential *Credential) (*Credential, error) {
-	result := l.GetGenericPersister().Table("credentials").Create(&credential)
-	if result.Error != nil {
-		return nil, fmt.Errorf("error saving user credentials: %v", result.Error)
+
+	if credential.ID == uuid.Nil {
+		id, err := uuid.NewV4()
+		if err != nil {
+			return nil, ErrGenerateUUID(err)
+		}
+		credential.ID = id
 	}
-	return nil, nil
+	err := l.GetGenericPersister().Transaction(func(tx *gorm.DB) error {
+		existingCred := Credential{}
+
+		if err := tx.First(&existingCred, "id = ?", credential.ID).Error; err == nil {
+			return err
+		}
+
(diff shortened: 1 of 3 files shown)


server/models/default_local_provider.go
```diff
@@ -1324,11 +1324,27 @@ func (l *DefaultLocalProvider) Cleanup() error {
...
+		if err := tx.First(&existingCred, "id = ?", credential.ID).Error; err == nil {
+			return err
+		}
+
+		return tx.Save(&credential).Error
+	})
+	if err != nil {
+		return nil, ErrPersistCredential(err)
+	}
+	return credential, nil
```
Yashsharma1911, line 1347: Thankyou this looks good to me, I'm just adding a point that, we always try to write business logic of handler in persister file of resource. Like for connection local provider handlers, business logic will be in connections_persister.go file it is just to make sure we don't make default_local_provider.go file too big. I think we don't have existing persister file for credential however we can iterate on that or create an issue for someone else to pick up. For now thanks for adding great description in PR, it conveys all tests for this PR have done. This PR LGTM

Yashsharma1911 (approved):

server/models/default_local_provider.go
codeknight03: @Yashsharma1911 , I have created and assigned an issue to myself with acceptance tests and implementation details. 

https://github.com/meshery/meshery/issues/13847

Do you think we can move forward with this for the time being, that would allow me move closer to my **v0.9** goals ?

vishalvivekm:
@codeknight03 

codeknight03:
As per the discussion, I should be looking to use the Schema for the Credential and Connection type and I'd be more than happy to continue in that direction. 
But merging this at least allows for us to move forward on the Prometheus and Grafana integrations with the local provider as currently `Meshery` provider is not supporting `Grafana`.

PR #13835, opened 2025-03-03
fix: dry run errors not showing up

souvikinator:
**Notes for Reviewers**

- This PR fixes #

- Refactored the code to use KubeStatus error parser from meshkit that is used in `ApplyManifest`

**Before:**
image

**After:**
image

### DIFF
--- a/server/handlers/design_engine_handler.go
+++ b/server/handlers/design_engine_handler.go
@@ -463,13 +463,33 @@ func parseDryRunFailure(settings map[string]interface{}, name string) *core.DryR
 	})
 
 	dResp.Causes = make([]core.DryRunFailureCause, 0)
-	for _, ld := range longDescription {
-		msg := ld
-		field := "unknown"
-		typ := "Failure"
-		failureCase := core.DryRunFailureCause{Message: msg, FieldPath: field, Type: typ}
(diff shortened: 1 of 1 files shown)


leecalcote (commented):
There's some good work happening in here. What consideration(s) do we need to make for "undeploy dry run" version "deploy dry run"?

Here's an item - https://github.com/meshery/meshery/actions/runs/13624239241/job/38078640793?pr=13835

server/handlers/design_engine_handler.go
```diff
@@ -449,33 +450,27 @@ func parseDryRunFailure(settings map[string]interface{}, name string) *core.DryR
...
+	}
+
+	_, longDescription, _, _ := utils.ParseKubeStatusErr(&kubeerror.StatusError{
+		ErrStatus: kubeStatus,
+	})
+
+	dResp.Causes = make([]core.DryRunFailureCause, 0)
+	for _, ld := range longDescription {
+		msg := ld
+		field := "unknown"
```
aabidsofi19, line 470: field is hardcoded here

aabidsofi19 (changes requested):

server/handlers/design_engine_handler.go
Jougan-0: Fixed

PR #14021, opened 2025-03-13
docs: refine documentation structure with unified headings and logica…

zihanKuang:
This PR is based on [#13982](https://github.com/meshery/meshery/pull/13982) by @miacycle 
### **Changes Introduced:**  
- **Improved Documentation Structure:**
  - **Refactored headings** for better clarity and organization.
  - **Enhanced logical grouping** of sections for better flow.
  - **Added a table of contents with jump links** to improve navigation.
  
- **Updated Target Audience Section:**
  - Reorganized personas into **Users, Contributors & System Integrators, and Administrators**.
  - Added **clear descriptions** of each audience type and their documentation needs.

- **Refined High-Level Outline:**
  - Ensured all documentation sets align with user personas.
  - Made structure **more intuitive** by arranging sections in a progressive learning path.

- **Navigation & Table of Contents (TOC) Fixes:**
  - Added **Docs Structure** entry under `toc.yml` to improve discoverability.
  - Ensured seamless **internal linking** across documentation pages.

PR #13982, opened 2025-03-11
[Docs] Audience, high-Level outline & information architecture for Meshery Documentation

miacycle:
**Notes for Reviewers**

- This PR pulls the audience, high-Level outline & information architecture for Meshery Documentation out of the docs/readme.md and into Meshery Docs itself.

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #14022, opened 2025-03-13
Restore Documentation Structure Refinement PR (Recovering #14021 & #13982)

zihanKuang:
This PR is created to restore the changes that were previously submitted in #14021 and #13982, which were unintentionally closed due to the deletion of the meshery:miacycle-patch-1 branch.


miacycle (approved):
Whew!

PR #14027, opened 2025-03-14
feat: refactored bats setup

ianrwhitney:
**Notes for Reviewers**

This PR enhances the cli setup bats test. 

It appears the first test will pass regarding if the $MESHERYCTL variable is set or not as empty return is considered a success. 

The next test in this bats file was missusing $status. It appears that the bats framework only saves $status and $output if the command is used with run. Additionally $status was misspelled so it wasnt getting picked up anyways.

This two are minor issues but helped gain better understanding of the bats framework. The enhancements may help others onboarding to this effort. 


mesheryctl/tests/e2e/000-prerequisites/00-cli.bats
```diff
@@ -3,12 +3,19 @@
 # Basic tests to validate cli has been build and behave properly at root
 
 @test "cli is available" {
+    if [[ -z "$MESHERYCTL_BIN" ]]; then
```
lekaf974, line 6: This is the purpose of the next step so instead we should stop if [ "$status" -eq 0 ] is failing

lekaf974: If the issue was locally it is addressed in #14033

ianrwhitney: Hey @lekaf974, the motivation for the change is that the issue w/ the way the test was written is that it _passes_ if the variable, $MESHERYCTL_BIN, is SET **AND** _passes_ if the variable is NOT SET.

lekaf974: Gotcha, we could simply remove that test in such case, since other test will  fail ?

The point here is that we are doing end-to-end testing against `mesheryctl` so in the suite test we should keep cli behavior under test. 

If  the variable is not set properly, it  is a setup issue and not the cli that is not working as expected,
 
Make sense ?

lekaf974: we can also do 
```bash
@test "cli is available" {
    run $MESHERYCTL_BIN --help
    [ "$status" -eq 0 ]
}
```
that way we have the `--help` flag and the test will fail

ianrwhitney: Yea, I will make this change. Thanks for input and expertise.

vishalvivekm:
@ianrwhitney 
Thank you for your contributions! Let's discuss this today, during Meshery Development call at 9:00 AM CT | 7:30 PM IST. Add it as an agenda item to the :memo: [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit?pli=1&tab=t.0#), if you would.
https://meet.meshery.io/dev

ianrwhitney:
@lekaf974 done!

lekaf974:
lint issue looks it needs another sync

riyaa14 (approved):
LGTM ✅

lekaf974 (approved):

PR #14061, opened 2025-03-19
[mesheryctl] add csv import in model import command

riyaa14:
**Notes for Reviewers**

- This PR fixes #

**Changes made**
- To make model import consistent with model import option in UI, add a CSV import in this mesheryctl model import command
- Enhance error logging for CSV import (currently, no error is shown if CSV import fails)

Enhanced error message:
Screenshot 2025-03-19 at 7 20 04 PM

Demo:

https://github.com/user-attachments/assets/a285e181-97d7-4d8c-b113-c4e5cc306476

**Implementation**
Check if path provided by user has CSV files, in yes, we continue with CSV model generation and import logic

- [*] Yes, I signed my commits.

### DIFF
--- a/mesheryctl/internal/cli/root/model/import.go
+++ b/mesheryctl/internal/cli/root/model/import.go
@@ -16,6 +16,7 @@ import (
 	"github.com/layer5io/meshery/server/models"
 	"github.com/layer5io/meshkit/encoding"
 	"github.com/layer5io/meshkit/errors"
+	meshkitRegistryUtils "github.com/layer5io/meshkit/registry"
 	meshkitutils "github.com/layer5io/meshkit/utils"
 	schemav1beta1 "github.com/meshery/schemas/models/v1beta1"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


Tanishq1604 (commented):
other than tests.
it looks good to me.

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -69,6 +73,55 @@ mesehryctl model import -f [path-to-models]
...
+						utils.Log.Infof("\n  %s:\n  %s", utils.BoldString("PROBABLE CAUSE"), strings.Join(meshkitErr.ProbableCause, ". "))
+					}
+					if len(meshkitErr.SuggestedRemediation) != 0 {
+						utils.Log.Infof("\n  %s:\n  %s", utils.BoldString("SUGGESTED REMEDIATION"), strings.Join(meshkitErr.SuggestedRemediation, ". "))
+					}
+				} else {
+					utils.Log.Error(err)
+				}
+
+				return nil
```
lekaf974, line 97: we should return the error because cli will exit with 0 instead of 1

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -69,6 +73,55 @@ mesehryctl model import -f [path-to-models]
...
+				}
+				relationshipData, err := os.ReadFile(relationshipcsvpath)
+				if err != nil {
+					return utils.ErrFileRead(err)
+				}
+				err = registerModel(modelData, componentData, relationshipData, "model.csv", "csv", "", !register)
+				if err != nil {
+					utils.Log.Error(err)
+					return nil
+				}
```
lekaf974, line 115: same here, return the error

lekaf974 (changes requested):
Could we return the errors to return proper exit code 1 from cli ?

PR #14109, opened 2025-03-23
refactor: mesheryctl add reusable function list function

lekaf974:
**Notes for Reviewers**

- This PR fixes #

# Content

## mesheryctl environment command

- Add tests
- Enhance UX
- Refactoring using common methods Fetch and List

## mesheryctl exp relationship

- Refactoring using common methods Fetch and List

### DIFF
--- a/mesheryctl/internal/cli/root/environments/environment_test.go
+++ b/mesheryctl/internal/cli/root/environments/environment_test.go
@@ -1,0 +1,10 @@
+package environments
+
+import "flag"
+
+var update = flag.Bool("update", false, "update golden files")
+
+var testConstants = map[string]string{
+	"orgID":           "3f8319e0-33a9-4736-b248-12nm3kiuh3yu",
+	"environmentName": "test-environment",
+}
--- a/mesheryctl/internal/cli/root/relationships/relationship_test.go
+++ b/mesheryctl/internal/cli/root/relationships/relationship_test.go
@@ -1,0 +1,5 @@
+package relationships
+
+import "flag"
+
+var update = flag.Bool("update", false, "update golden files")


lekaf974:
@riyaa14 @alphaX86 @gentcod @leecalcote 

Can I have feedback on this PR related to have a common method to list data in mesheryctl ?

mesheryctl/internal/cli/pkg/api/meshery.go
```diff
@@ -0,0 +1,43 @@
...
+func Fetch[T any](url string) (*T, error) {
+	return makeReques[T](url, http.MethodGet)
+}
+
+// Send a Http request to meshery server from mesheryctl cli
+func makeReques[T any](url string, httpMethod string) (*T, error) {
+	req, err := utils.NewRequest(httpMethod, url, nil)
+	if err != nil {
+		return nil, err
+	}
```
gentcod, line 21: What if there is a request body

lekaf974: Good question, during this first iteration I just add support GET http call that's why I did not add body but  `utils.NewRequest` already support it, I'll add a new param in makeRequest

lekaf974: done

mesheryctl/internal/cli/root/environments/create_test.go
```diff
@@ -0,0 +1,114 @@
...
+			Args:             []string{"create"},
+			URL:              testContext.BaseURL + "/api/environments",
+			Method:           "POST",
+			Fixture:          "",
+			ExpectedResponse: "create.environment.without.name.golden",
+			ExpectError:      true,
+		},
+		{
+			Name:             "Create environment successfully",
+			Args:             []string{"create", "--name", "test-environment", "--description", "integration test", "--orgID", "3f8319e0-33a9-4736-b248-12nm3kiuh3yu"},
```
gentcod, line 53: I think some of these values would be best slated as constant values, doesn't really make a difference but it's makes it more readable

mesheryctl/internal/cli/root/environments/delete_test.go
```diff
@@ -0,0 +1,114 @@
...
+			URL:              testContext.BaseURL + "/api/environments",
+			Method:           "DELETE",
+			Fixture:          "",
+			ExpectedResponse: "delete.environment.without.name.golden",
+			ExpectError:      true,
+		},
+		{
+			Name:             "Delete environment successfully",
+			Args:             []string{"delete", "d56fb25b-f92c-4cd6-821b-2cfd6bb87259"},
+			URL:              testContext.BaseURL + "/api/environments/d56fb25b-f92c-4cd6-821b-2cfd6bb87259",
```
gentcod, line 54: same here.. the orgID as a const value

mesheryctl/internal/cli/root/environments/environment.go
```diff
@@ -15,44 +15,44 @@
...
-	"github.com/fatih/color"
 	"github.com/manifoldco/promptui"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
+var update = flag.Bool("update", false, "update golden files")
+
 var (
```
gentcod, line 33: why not const instead of var. var makes the variable mutable.

lekaf974: we could investigate but I suggest implement it the same way it is done every where else but const should be a better option. I'll try it

gentcod: oh, yeah constant works selected types too.. so it's fine

lekaf974: nope it does not compile using constant type since it is a pointer behind the scene

```
# github.com/layer5io/meshery/mesheryctl/internal/cli/root/environments [github.com/layer5io/meshery/mesheryctl/internal/cli/root/environments.test]
./environment.go:31:16: flag.Bool("update", false, "update golden files") (value of type *bool) is not constant
FAIL	github.com/layer5io/meshery/mesheryctl/internal/cli/root/environments [build failed]
```

gentcod: yeah.. I figured, it's fine

lekaf974:
@alphaX86 @riyaa14 Could you have a look here please

gentcod (approved):
so far the implementation and changes  LGTM..

miacycle (dismissed):
New code comments! 😍 

I wonder if there are any changes to existing error codes or any new error codes as part of this PR.

leecalcote (approved):
I went ahead and ran `cd mesheryctl; make error`. I don't believe any new error codes are defined. Should there be?

lekaf974:
No errors were added only content and how it was displayed to follow the new standard added in other commands

mesheryctl/internal/cli/root/environments/environment.go
```diff
@@ -15,44 +15,44 @@
...
-	outFormatFlag string
-	saveFlag      bool
-
-	maxRowsPerPage       = 25
-	whiteBoardPrinter    = color.New(color.FgHiBlack, color.BgWhite, color.Bold)
 	availableSubcommands = []*cobra.Command{listEnvironmentCmd, createEnvironmentCmd, deleteEnvironmentCmd, viewEnvironmentCmd}
+	testConstants        = map[string]string{
+		"orgID":           "3f8319e0-33a9-4736-b248-12nm3kiuh3yu",
+		"environmentName": "test-environment",
+	}
```
riyaa14, line 38: I understand that `testContants` is being used in testing for all subcommands, but still shouldn't we include this in some other file? something like testdata or fixtures?

This might be confusing if we include any test data in a command file. What do you think?

lekaf974: finally create `environment_test.go` file to handle that

lekaf974:
All comment addressed, one approval required to merge

leecalcote (approved):

miacycle (commented):
✅

PR #14251, opened 2025-04-03
fix(ci): update golangci-lint version and go to 1.24

lekaf974:
**Notes for Reviewers**

- This PR fixes golanci-lint fail

### DIFF
--- a/.github/workflows/go-testing-ci.yml
+++ b/.github/workflows/go-testing-ci.yml
@@ -23,7 +23,7 @@ jobs:
   golangci:
     strategy:
       matrix:
-        go: [1.24]
+        go: [1.23]
         os: [ubuntu-24.04]
     name: golangci-lint
     if: github.repository == 'meshery/meshery'


leecalcote (commented):
Thanks for tackling this, @lekaf974  🙌 

_Suggestion:_ Rather than linting at a go version (1.24) that the project isn't currently using, I recommend enhancing the configuration of the setup of golangci-lint to specify use of golang 1.23. What do you think?

In order to specify this, it might be that we have to ensure that golang is setup in a step prior to golangci-lint, like so:

```
jobs:
  lint:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v4

      - name: Set up Go 1.23
        uses: actions/setup-go@v5
        with:
          go-version: '1.23'

      - name: Run golangci-lint
        uses: golangci/golangci-lint-action@v6
```

leecalcote (approved):
All good here! ✅

PR #14288, opened 2025-04-04
chore(mesheryctl-workspace): update workspace command using reusable fetch and list function

lekaf974:
**Notes for Reviewers**

- This PR refactor workspace command
  - update workspace command using reusable fetch and list function
  - enhance common Fetch managing base meshery api url locally and update all commands accordingly
  - Add a new common function to test not list command + implement in workspace and updatte environment

### DIFF
--- a/mesheryctl/internal/cli/root/components/component.go
+++ b/mesheryctl/internal/cli/root/components/component.go
@@ -73,14 +73,7 @@ mesheryctl component view [component-name]
 	RunE: func(cmd *cobra.Command, args []string) error {
 		count, _ := cmd.Flags().GetBool("count")
 		if count {
-			mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
-			if err != nil {
-				log.Fatalln(err, "error processing config")
-			}
-
-			baseUrl := mctlCfg.GetBaseMesheryURL()
-			url := fmt.Sprintf("%s/%s", baseUrl, componentApiPath)
-			return listComponents(cmd, url)
+			return listComponents(cmd, componentApiPath)
 		}
 
 		if ok := utils.IsValidSubcommand(availableSubcommands, args[0]); !ok {
--- a/mesheryctl/internal/cli/root/components/component_test.go
+++ b/mesheryctl/internal/cli/root/components/component_test.go
@@ -24,20 +24,20 @@ func TestComponent(t *testing.T) {
 	// test scenarios for fetching data
 	tests := []utils.MesheryListCommamdTest{
 		{
-			Name:             "display components count",
-			Args:             []string{"--count"},
-			URL:              fmt.Sprintf("/%s", componentApiPath),
-			Fixture:          "components.api.response.golden",
-			ExpectedResponse: "components.list.count.output.golden",
-			ExpectError:      false,
-		},
-		{
 			Name:             "launch component with invalid subcommand name",
 			Args:             []string{"invalidCommand"},
 			URL:              "",
 			Fixture:          "components.api.response.golden",
 			ExpectedResponse: "components.invalid.subcommand.output.golden",
 			ExpectError:      true,
+		},
+		{
+			Name:             "display components count",
+			Args:             []string{"--count"},
+			URL:              fmt.Sprintf("/%s", componentApiPath),
+			Fixture:          "components.api.response.golden",
+			ExpectedResponse: "components.list.count.output.golden",
+			ExpectError:      false,
 		},
 	}
 
--- a/mesheryctl/internal/cli/root/components/list.go
+++ b/mesheryctl/internal/cli/root/components/list.go
@@ -17,10 +17,8 @@ package components
 import (
 	"fmt"
 
-	"github.com/layer5io/meshery/mesheryctl/internal/cli/root/config"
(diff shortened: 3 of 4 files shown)


lekaf974:
@n2h9 could you have a look here, thanks

mesheryctl/internal/cli/root/organizations/organization.go
```diff
@@ -43,7 +41,7 @@ mesheryctl organizations list
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if count {
-			orgs, err := getAllOrganizations()
+			orgs, err := api.Fetch[models.OrganizationsPage](fmt.Sprintf("%s?all=true", organizationsApiPath))
```
leecalcote, line 44: Are you finding this behavior is consistent across the REST API - such that you have to specify `?all=true`?

mesheryctl/internal/cli/root/relationships/relationship.go
```diff
@@ -72,14 +71,7 @@ mesheryctl exp relationship view [model-name]
...
 		if countFlag {
-			mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
-			if err != nil {
-				log.Fatalln(err, "error processing config")
-			}
-
-			baseUrl := mctlCfg.GetBaseMesheryURL()
-			url := fmt.Sprintf("%s/api/meshmodels/relationships?page=1", baseUrl)
-			models, err := api.Fetch[MeshmodelRelationshipsAPIResponse](url)
+			models, err := api.Fetch[MeshmodelRelationshipsAPIResponse]("api/meshmodels/relationships?page=1")
```
leecalcote, line 74: Why do we need to specify `?page=1`? The default behavior of every REST API in Meshery should be to return a paginated list. If we can remove this parameter by confirming the behavior of the API, that be ideal.

mesheryctl/internal/cli/root/relationships/view.go
```diff
@@ -49,17 +47,9 @@ mesheryctl exp relationship view [model-name]
...
-			return utils.ErrLoadConfig(err)
-		}
-
-		baseUrl := mctlCfg.GetBaseMesheryURL()
 		model := args[0]
 
-		url := fmt.Sprintf("%s/api/meshmodels/models/%s/relationships?pagesize=all", baseUrl, model)
-
-		relationshipsResponse, err := api.Fetch[MeshmodelRelationshipsAPIResponse](url)
+		relationshipsResponse, err := api.Fetch[MeshmodelRelationshipsAPIResponse](fmt.Sprintf("api/meshmodels/models/%s/relationships?pagesize=all", model))
```
leecalcote, line 52: I wonder if we are able to avoid `?pagesize=all` as queries like this are a performance hit on Meshery Server (when there is a large result set).

mesheryctl/internal/cli/root/workspaces/fixtures/list.workspace.api.response.golden
```diff
@@ -0,0 +1,67 @@
...
+      "created_at": "2024-04-08T11:01:52.364684Z",
+      "updated_at": "2024-04-08T11:01:52.364695Z",
+      "deleted_at": {
+        "Time": "0001-01-01T00:00:00Z",
+        "Valid": false
+      }
+    },
+    {
+      "id": "d47de0e3-dfc2-4778-a440-100221adf88c",
+      "name": "workspace-4",
```
leecalcote, line 56: Maybe this is intentionally a string, but its worth noting that the value of the `owner` property should be returning a UUID.

mesheryctl/internal/cli/root/relationships/relationship.go
lekaf974: here it is for count only that's why targeting page 1

mesheryctl/internal/cli/root/relationships/search.go
```diff
@@ -103,14 +93,14 @@ func init() {
 	searchCmd.Flags().StringVarP(&searchType, "type", "t", "", "search particular type of relationships")
 }
 
-func buildSearchUrl(baseUrl string) string {
+func buildSearchUrl() string {
 	var searchUrl strings.Builder
 
 	if searchModelName == "" {
-		searchUrl.WriteString(fmt.Sprintf("%s/api/meshmodels/relationships?", baseUrl))
+		searchUrl.WriteString("api/meshmodels/relationships?")
```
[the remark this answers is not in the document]
lekaf974: it will always have query params, that's why ? is added

mesheryctl/internal/cli/root/relationships/view.go
lekaf974: I think it is probably better not using it, since relationship is still in exp we can create a ticket for that

mesheryctl/internal/cli/root/organizations/organization.go
lekaf974: added this issue #14464 to make the command to follow the same behavior from other commands

mesheryctl/internal/cli/root/relationships/relationship.go
lekaf974: removed

lekaf974:
@riyaa14 @gourav-k-shaw can you have a look here

PR #14289, opened 2025-04-04
Fix: Login Logout Error due to Parallel Testing

vr-varad:
**Notes for Reviewers**

- This PR fixes #

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #14301, opened 2025-04-04
fix: fix logout test flakiness

ianrwhitney:
**Notes for Reviewers**

Currently there is flakiness introduced by adding additional workers in the playwright config. This flakiness is d/t a logout test modifying the shared authentication state across the projects. This causes test that run after the logout occured to fail. 

There are multiple approaches to fix this one being an additional project(s) for unauthenticated flows. While this will work i dont believe the added complexity of creating a separate stored auth state per worker is warrented d/t to the increased complexity to simply test the logout functionality. Addtionally we dont currently have any critical flows outside of an authenticated state. 

Therefore, i think mocking the logout call is the simpliest approach to our current situation which prevents the modification of the shared auth state. 

image

### DIFF
--- a/ui/tests/e2e/logout.spec.js
+++ b/ui/tests/e2e/logout.spec.js
@@ -9,16 +9,16 @@ test.describe('Logout Page Tests', () => {
       route.fulfill({
         status: 302,
         headers: {
-          'Location': '/provider'
-        }
+          Location: '/provider',
+        },
       });
     });
-  
+
     const waitForLogoutRequest = page.waitForRequest('/user/logout');
-  
+
     await page.getByTestId('header-menu').click();
     await page.getByRole('menuitem', { name: 'Logout' }).click();
-  
+
     const request = await waitForLogoutRequest;
     expect(request.url()).toContain('/user/logout');
 


vr-varad:
I tried as well using setup and teardown @ianrwhitney  https://github.com/meshery/meshery/pull/14289

leecalcote (commented):
I believe that there is an underlying bug beneath the tests that are causing this flakiness (not the tests themselves).

aabidsofi19:
from the discussion earlier  it seems like logout test was clearing the auth state before all tests were done so the later tests were failing due no auth token .

aabidsofi19:
this test will now test the ui for the flow , so we  should push the actual logout to teardown .

ianrwhitney:
@leecalcote 

Yea, @aabidsofi19 said it correctly. Since we added workers it exposed this flakiness more clearly, which is when this test runs it modifies the shared auth state across all of the tests. I believe the same realization would have come with time without the worker increase as the logic still holds. 

Earlier this morning, I was trying to make the case that the solution for this currently should be the simplest(mocking logout api) vs trying to support multiple playwright project definitions to sequence this ONE test case to run after all auth'ed-required tests run. 

If we see that the use cases to test against un-auth'ed flows coming such as top of a funnel or prior to some purchase flows then I would advocate for putting additional support in for multiple playwright projects to run and sequence accordingly no-auth/auth.

PR #14318, opened 2025-04-07
chore(model-export): fix panic server side and error not handled properly in cli

lekaf974:
**Notes for Reviewers**

- This PR fixes an issue with model export if we have on of `--discard-components`  or `--discard-relationships` flags

## Current behavior

```bash
./mesheryctl model export apisix --discard-components --discard-relationships
Get "http://localhost:9081/api/meshmodels/export?name=apisix&output_format=yaml&file_type=oci&components=false&relationships=false&pagesize=all": EOF
```

---
2 issues found
- a panic was raise in server when trying to access a nil value
- cli was not handling the error returned properly

### CLI
If an error is returned from the server
```bash
./mesheryctl model export apisix --discard-components --discard-relationships
Error: Given model with name: apisix could not be exported: an error occured on server during export process
```

### Server
On server we prevent manipulation of nil value and now export is working properly from cli
```bash
./mesheryctl model export apisix --discard-components --discard-relationships 
Exported model to ./apisix.tar


ls ./apisix.tar 
./apisix.tar
```

### DIFF
--- a/mesheryctl/internal/cli/root/model/export.go
+++ b/mesheryctl/internal/cli/root/model/export.go
@@ -90,7 +90,7 @@ func export(modelName string, url string, output *outputDetail) error {
 
 	resp, err := utils.MakeRequest(req)
 	if err != nil {
-		return ErrExportModel(errors.New("an error occured on server during export process"), modelName)
+		return ErrExportModel(errors.New("an error occurred on server during export process"), modelName)
 	}
 
 	// ensure proper cleaning of resources


gourav-k-shaw (approved):

mesheryctl/internal/cli/root/model/export.go
```diff
@@ -90,8 +90,7 @@ func export(modelName string, url string, output *outputDetail) error {
 
 	resp, err := utils.MakeRequest(req)
 	if err != nil {
-		utils.Log.Error(err)
-		return nil
+		return ErrExportModel(errors.New("an error occured on server during export process"), modelName)
```
riyaa14, line 93:
```suggestion
		return ErrExportModel(errors.New("an error occurred on server during export process - " + err.Error()), modelName)
```

Adding the exact error message as well from the server can be useful for debugging. 
(not very useful in this case when server just throws the endpoint, but can be helpful for other error logs from server)

riyaa14 (commented):
a minor suggestion, otherwise LGTM ✅

mesheryctl/internal/cli/root/model/export.go
lekaf974: the error return by server is not useful that's why I removed it see below

```bash
Get "http://localhost:9081/api/meshmodels/export?name=apisix&output_format=yaml&file_type=oci&components=false&relationships=false&pagesize=all": EOF
```

riyaa14 (approved):

mrmerlin320 (approved):

PR #14342, opened 2025-04-09
fix(mesheryctl-model): fix model export with a non existing version raising a panic

lekaf974:
**Notes for Reviewers**

- This PR fixes panic in meshery server

After previous fix made #14318, the error displayed mask the real issue server side while trying to access index 0 in a nil array.

## Before changes
```bash
./mesheryctl model export accure --version v1.1.1 
Error: Given model with name: accure could not be exported: an error occurred on server during export process
```

## After changes
```
bash./mesheryctl model export accure --version v1.1.1 
Error: Response Status Code 404. Server emitted an error: model with name accure version v1.1.1 has not been found
```

### DIFF
--- a/mesheryctl/internal/cli/root/model/export.go
+++ b/mesheryctl/internal/cli/root/model/export.go
@@ -90,7 +90,7 @@ func export(modelName string, url string, output *outputDetail) error {
 
 	resp, err := utils.MakeRequest(req)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


marblom007 (approved):
It’s confidence inspiring to know that we’re proactively identifying and fixing caveats like these. 😌

leecalcote (approved):
Ditto. What @marblom007 said.

leecalcote:
I don’t believe that the lint failure is related to this PR.

PR #14353, opened 2025-04-10
chore(mesheryctl-e2e): add bats-file library and use bats features for tests enhancement

lekaf974:
**Notes for Reviewers**

- This PR enhance bats tests implementation

Goal was to improve test output on fail using tools provided by bats

## Before
```bash
 ✗ mesheryctl model list --count is succeeded and return total numbers of models
   (in test file 002-model/01-model-list.bats, line 16)
     `[ "$status" -eq 0 ]' failed
```

## After
```bash
 ✗ mesheryctl model list --count is succeeded and return total numbers of models
   (from function `__assert_stream' in file helpers/bats-libs/bats-assert/src/assert_output.bash, line 230,
    from function `assert_output' in file helpers/bats-libs/bats-assert/src/assert_output.bash, line 125,
    in test file 002-model/01-model-list.bats, line 20)
     `assert_output --regexp "^Total number of models: [0-9]+$"' failed
   
   -- regular expression does not match output --
   regexp : ^Total number of models: [0-9]+$
   output : Total number of  model : 270
   --
```

### DIFF
--- a/mesheryctl/tests/e2e/000-prerequisites/00-cli.bats
+++ b/mesheryctl/tests/e2e/000-prerequisites/00-cli.bats
@@ -1,5 +1,10 @@
 #!/usr/bin/env bats
 
+setup() {
+    load "$E2E_HELPERS_PATH/tests_helpers"
+	_tests_helper
+}
+
 # Basic tests to validate cli has been build and behave properly at root
 
 @test "cli is available" {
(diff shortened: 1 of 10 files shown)


riyaa14 (approved):
This enhancement looks great ! 🙌

miacycle (approved):
Looking good! 

From @leecalcote 
> Ooh, this is a good topic. I wonder if we might eventually arrive at pushing result output to Meshery Docs akin to how adapter test results and I think now Meshery UI tests results are being pushed.

I also wonder if we've begun to use the same test _readiness_ nomenclature:

- "stable"
- "flaky"

And the same test status nomenclature (passing/failing - something in the middle?)

leecalcote (approved):
🥇

PR #14372, opened 2025-04-11
Fix: Verify Kanvas Detail Test in chromium Local Provider

vr-varad:
**Notes for Reviewers**

- This PR fixes #

chromium-meshery-provider
image

chromium-local-provider
image


ui/tests/e2e/extensions.spec.js
```diff
@@ -43,11 +43,18 @@ test.describe('Extensions Section Tests', () => {
...
   });
 
-  test('Verify Kanvas Details', async ({ page }) => {
+  test('Verify Kanvas Details', async ({ page, context }) => {
     await expect(page.getByTestId('kanvas-signup-heading')).toBeVisible();
     const kanvasDetailsButton = page.getByTestId('kanvas-signup-btn');
     await expect(kanvasDetailsButton).toBeVisible();
-    await expect(kanvasDetailsButton).toBeDisabled();
+    if (await kanvasDetailsButton.isEnabled()) {
+      const [docsPage] = await Promise.all([
```
sudhanshutech, line 51: ensure the error handling is proper here, promise.all is generally not safe. If either of the event fails the test case will fail.

ianrwhitney: I think we are fine here in playwright regarding this pattern as we have two events that are tightly coupled and if one fails (i/e the click doesn't open the page) thats a valid reason for the test to fail anyways.

ianrwhitney (commented):
Overall, looks good for a short term fix to a failing test. One nit on the extra await in the promise.all block.

Additionally, I added some comments in team slack to suggest we all align on an approach to handle the auth context differences in a scalable manner as we may face this more as we improve test coverage and it would be good to keep this auth-context conditional logic as clean as possible.

 So we can see what team thinks on that.

ianrwhitney (approved):
LGTM.

PR #14388, opened 2025-04-13
Test: E2E playwright test for model create and model import wizard

### LINKED ISSUE #14240: [Test] E2E playwright test for model create and model import wizard
### Current Behavior
Currently, we don't have any model test generate flow from the registry page.

### Desired Behavior
E2E playwright test is needed.

### Mockups

---

vr-varad:
**Notes for Reviewers**

- This PR fixes #14240

### DIFF
--- a/docs/_includes/test-report.md
+++ b/docs/_includes/test-report.md
@@ -1,17 +1,17 @@
 ### END-TO-END TESTS
 
-- Testing started at: April 25th 2025, 8:08:10 am
+- Testing started at: April 25th 2025, 2:09:46 pm
 
 **📦 Test Result Summary**
 
-- ✅ 60 passed
-- ❌ 0 failed
-- ⚠️ 2 flaked
+- ✅ 56 passed
+- ❌ 1 failed
+- ⚠️ 5 flaked
 - ⏩ 0 skipped
 
-⌛ _Duration: 5 minutes and 10 seconds_
+⌛ _Duration: 6 minutes and 32 seconds_
 
-**Overall Result**: 👍 All tests passed.
+**Overall Result**: 👎 Some tests failed.
 
 
 
@@ -26,11 +26,15 @@
 | 3 | chromium-meshery-provider | Transition to ignored state and then back to connected state | unstable | ⚠️ |
 | 4 | chromium-meshery-provider | Transition to not found state and then back to connected state | unstable | ⚠️ |
 | 5 | chromium-meshery-provider | Delete Kubernetes cluster connections | unstable | ⚠️ |
-| 6 | chromium-local-provider | Add a cluster connection by uploading kubeconfig file | unstable | ⚠️ |
-| 7 | chromium-local-provider | Transition to disconnected state and then back to connected state | unstable | ⚠️ |
-| 8 | chromium-local-provider | Transition to ignored state and then back to connected state | unstable | ⚠️ |
-| 9 | chromium-local-provider | Transition to not found state and then back to connected state | unstable | ⚠️ |
-| 10 | chromium-local-provider | Delete Kubernetes cluster connections | unstable | ⚠️ |
+| 6 | chromium-meshery-provider | Configure Existing Istio adapter through Mesh Adapter URL from Management page | unstable | ⚠️ |
+| 7 | chromium-meshery-provider | Ping Istio Adapter | unstable | ⚠️ |
+| 8 | chromium-meshery-provider | Connect to Meshery Istio Adapter and configure it |  | ❌ |
+| 9 | chromium-local-provider | Add a cluster connection by uploading kubeconfig file | unstable | ⚠️ |
+| 10 | chromium-local-provider | Transition to disconnected state and then back to connected state | unstable | ⚠️ |
+| 11 | chromium-local-provider | Transition to ignored state and then back to connected state | unstable | ⚠️ |
+| 12 | chromium-local-provider | Transition to not found state and then back to connected state | unstable | ⚠️ |
+| 13 | chromium-local-provider | Delete Kubernetes cluster connections | unstable | ⚠️ |
+| 14 | chromium-local-provider | Configure Existing Istio adapter through Mesh Adapter URL from Management page | unstable | ⚠️ |
 
 </div>
 </details>
--- a/ui/components/ConfirmationModal.js
+++ b/ui/components/ConfirmationModal.js
@@ -2,23 +2,20 @@ import {
   Box,
   Button,
   Checkbox,
-  Chip,
-  Dialog,
   DialogActions,
   DialogContent,
   DialogContentText,
-  DialogTitle,
-  IconButton,
   TextField,
-  Tooltip,
   Typography,
   styled,
   Tab,
   Tabs,
-  CloseIcon,
   DoneAllIcon,
   DoneIcon,
   RemoveDoneIcon,
+  Modal,
+  ModalBody,
+  useTheme,
 } from '@layer5/sistent';
 import { Search } from '@mui/icons-material';
 import { connect } from 'react-redux';
@@ -41,34 +38,9 @@ import CAN from '@/utils/can';
 import { keys } from '@/utils/permission_constants';
 import { K8sContextConnectionChip } from './Header';
 import { useFilterK8sContexts } from './hooks/useKubernetesHook';
-
-const ContextChip = styled(Chip)(({ theme }) => ({
-  height: '50px',
-  fontSize: '15px',
-  position: 'relative',
-  top: theme.spacing(0.5),
-  [theme.breakpoints.down('md')]: {
-    fontSize: '12px',
-  },
-}));
-
-const ContextIcon = styled('img')(({ theme }) => ({
-  display: 'inline',
-  verticalAlign: 'text-top',
-  width: theme.spacing(2.5),
-  marginLeft: theme.spacing(0.5),
-}));
-
-const DialogTitleStyled = styled(DialogTitle)(({ theme }) => ({
-  textAlign: 'center',
-  padding: theme.spacing(1),
-  color: '#fff',
-  backgroundColor: theme.palette.background.tabs,
-  fontSize: '1rem',
-}));
+import { TooltipWrappedConnectionChip } from './connections/ConnectionChip';
 
 const DialogSubtitle = styled(DialogContentText)({
-  minWidth: 400,
   overflowWrap: 'anywhere',
   textAlign: 'center',
   padding: '5px',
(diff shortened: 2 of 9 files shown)


vr-varad:
### ✅ Feature Coverage in Tests

- [x] **Create Model**

https://github.com/user-attachments/assets/9eeecd57-a499-491c-8c71-1010f0ca2a2e


- [x] **Search Model**

https://github.com/user-attachments/assets/51596d40-75b4-4f80-8d8c-434f246879ca


- [x] **Import Model**

https://github.com/user-attachments/assets/ada54aef-7ec2-48af-bf00-4165cab7542a



---

### 💡 Issue / Advice

- Currently using a **random model GitHub link** to test "Create Model" — this may lead to flaky behavior or test inconsistency.
- Used a random **`.tar` file** to test the import model functionality.

---

### 🚧 Works to be Done

- [ ] Add **`data-testid`** attributes for more reliable and consistent selectors in tests.
- [ ] Fix issue where **notifications are hidden behind the "Create Model" modal** — this affects visibility and confirmation of successful model creation.

vr-varad:
@leecalcote @aabidsofi19 @FaheemOnHub @amitamrutiya @Tanishq1604 @ianrwhitney 

Would love your review, suggestions, and any improvements you’d recommend

ui/tests/e2e/models.spec.js
```diff
@@ -0,0 +1,62 @@
...
+  await page.getByRole('button', { name: 'Next' }).click();
+
+  await page.getByRole('radio', { name: 'GitHub' }).check();
+  await page.getByRole('textbox', { name: 'Model URL' }).fill(MODEL_URL);
+
+  await page.getByRole('button', { name: 'Next' }).click();
+  await page.getByRole('checkbox', { name: /components in this model/i }).check();
+  await page.getByRole('button', { name: 'Next' }).click();
+
+  await page.getByRole('button', { name: 'Generate' }).click();
```
amitamrutiya, line 33: I think this should be get from id not name

amitamrutiya (changes requested):

amitamrutiya:
We have two other ways to improve model. If you test those then it will be good but so far this is good wanted work.

ui/tests/e2e/models.spec.js
vr-varad: Yes @amitamrutiya , using id instead of name is definitely a better approach for test reliability. However, at this stage, adding ids to all relevant components wouldn't be efficient. I’ve already discussed this with @ianrwhitney , and we’re considering introducing data-testid attributes across all Sistent components to improve testability. For now, the current approach works, and once the ids or consistent test IDs are added, I’ll update the tests accordingly.

ui/tests/e2e/models.spec.js
```diff
@@ -0,0 +1,62 @@
...
+
+test('Create a Model', async ({ page }) => {
+  await page.getByTestId('create-model-button').click();
+
+  await page.getByRole('textbox', { name: 'Model Name' }).fill(MODEL_NAME);
+  await page.getByRole('textbox', { name: 'Model Display Name' }).fill(MODEL_DISPLAY_NAME);
+
+  await page.getByRole('button', { name: 'Next' }).click();
+  await page.getByRole('button', { name: 'Next' }).click();
+  await page.getByRole('button', { name: 'Next' }).click();
```
ianrwhitney, line 24: can we assert on some intermediate steps here to ensure UI is behaving as anticpated.

ui/components/NotificationCenter/formatters/model_registration.js
```diff
@@ -193,7 +193,7 @@ export const ModelImportedSection = ({ modelDetails }) => {
         return (
           <Box key={index} mb={2}>
             <div style={{ display: 'flex', alignItems: 'center', justifyContent: 'space-between' }}>
-              <Typography gutterBottom>
+              <Typography gutterBottom data-testid="model-info">
```
ianrwhitney, line 196: CUrious if we can name this after the component as this may be to generic here. Also raise it one level vs just having the label included.

ui/components/NotificationCenter/formatters/model_registration.js
```diff
@@ -250,7 +250,7 @@ export const ModelImportedSection = ({ modelDetails }) => {
 };
 
 export const ModelImportMessages = ({ message }) => (
-  <Typography>
+  <Typography data-testid="model-summary">
```
ianrwhitney, line 253: Here as well this may be to generic

ui/utils/custom-search.js
```diff
@@ -123,7 +123,12 @@ const SearchBar = ({ onSearch, placeholder, expanded, setExpanded, value = '' })
       ) : (
         <CustomTooltip title="Search">
           <IconButton onClick={handleSearchIconClick}>
-            <SearchIcon fill={theme.palette.icon.secondary} height={'1.5rem'} width={'1.5rem'} />
+            <SearchIcon
+              data-testid="search-icon"
```
ianrwhitney, line 127: Also would we be able to lift this to a higher level on the component? If its not we may need to add a modifier to this as this toggles between expanded/collpased states

ui/tests/e2e/models.spec.js
```diff
@@ -0,0 +1,62 @@
...
+test('Search a Model and Export it', async ({ page }) => {
+  await page.getByTestId('search-icon').click();
+  await page.getByRole('textbox', { name: 'Search' }).click();
+  await page.getByRole('textbox', { name: 'Search' }).fill(MODEL_DISPLAY_NAME);
+  await page.getByText(MODEL_DISPLAY_NAME).click();
+
+  const downloadPromise = page.waitForEvent('download');
+  await page.getByRole('button', { name: 'Export Model' }).click();
+  const download = await downloadPromise;
+  expect(download).toBeDefined();
```
ianrwhitney, line 48: is there any success we can expect on after a download that is in the UI?

ianrwhitney (commented):
Thanks for including the videos 🚀

ui/tests/e2e/models.spec.js
vr-varad: Nope, @ianrwhitney it just downloads

leecalcote:
I'm excited about these tests! 🚀

vr-varad:
Hi @ianrwhitney,  
I've made the changes based on your feedback. Please let me know if anything else needs to be updated.  

Also, I had a quick question regarding the `data-testid`—would you recommend adding it directly inside the components or within the `sistent`?

vr-varad:
image

ui/tests/e2e/models.spec.js
```diff
@@ -0,0 +1,123 @@
...
+  MODEL_CSV_IMPORT: {
+    Models: path.resolve('tests/e2e/assets/models.csv'),
+    Components: path.resolve('tests/e2e/assets/components.csv'),
+    Relationships: path.resolve('tests/e2e/assets/relationships.csv'),
+  },
+};
+
+test.describe('Model Workflow Tests', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto(ENV.MESHERY_SERVER_URL);
```
ianrwhitney, line 24: can we use the baseUrl that is configured in the playwright.configs.

ui/tests/e2e/models.spec.js
```diff
@@ -0,0 +1,62 @@
...
+  await page.getByRole('heading', { name: 'File Import' }).click();
+
+  await page.setInputFiles('input[type="file"]', TAR_FILE_PATH);
+
+  await page.getByRole('button', { name: 'Next' }).click();
+
+  await expect(page.getByTestId('model-summary')).toBeVisible();
+  await expect(page.getByTestId('model-info')).toBeVisible();
+  await page.getByRole('button', { name: 'Finish' }).click();
+});
```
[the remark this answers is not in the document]
ianrwhitney: If we import and create models over and over without cleanup action taken, there will soon be thousands as this test runs many times a day.

ui/tests/e2e/models.spec.js
```diff
@@ -0,0 +1,123 @@
...
+    'https://raw.githubusercontent.com/vr-varad/meshery/refs/heads/feat/create_import_model_testing/ui/tests/e2e/assets/test.tar',
+  MODEL_FILE_IMPORT: path.resolve('tests/e2e/assets/test.tar'),
+  MODEL_CSV_IMPORT: {
+    Models: path.resolve('tests/e2e/assets/models.csv'),
+    Components: path.resolve('tests/e2e/assets/components.csv'),
+    Relationships: path.resolve('tests/e2e/assets/relationships.csv'),
+  },
+};
+
+test.describe('Model Workflow Tests', () => {
```
ianrwhitney, line 22: It seems like if these tests are sharing the same model state and depend on a previous step running. Do we need to set serial mode or look at make them indepedent would be best if there is the oppertunity to do so.

ui/components/NotificationCenter/formatters/model_registration.js
```diff
@@ -250,7 +253,7 @@ export const ModelImportedSection = ({ modelDetails }) => {
 };
 
 export const ModelImportMessages = ({ message }) => (
-  <Typography>
+  <Typography data-testid="model-import-messages">
```
ianrwhitney, line 256: Wondering if error messages in logs would be more descriptive and targeted if we used the component name as the first element in these ids. So this would become: `data-testId=ModelimportMessages`

Same suggestion for others.

ianrwhitney (commented):
Overall, looks great. some suggestions for considerations. thank you

ianrwhitney:
Lets make sure we track the coverage here too:  https://docs.google.com/spreadsheets/d/13Ir4gfaKoAX9r8qYjAFFl_U9ntke4X5ndREY1T7bnVs/edit?gid=838298230#gid=838298230

ui/tests/e2e/models.spec.js
vr-varad: make them serial would be good and first 2 tests are dependent and rest 3 are independent but making them serial won't be troublesome @ianrwhitney

ui/components/NotificationCenter/formatters/model_registration.js
vr-varad: yeah that would be easy to locate the error is there is any!!
thanks for the suggestion

vr-varad: but most of the components are formcontrol or button so using a unique testid would be best, but ya if we have a unique component we could use that in case of summary and message!!!
thoughts @ianrwhitney

ianrwhitney: Yeah, that’s why I think we should keep some of the sub-elements optional, as noted with the brackets in the proposed Slack format:

[] = optional
<> = required

Suggested Format:
<ComponentName>-[SubComponentName]-[Modifier]-[UniqueId]

The goal of this nudge is to stay consistent with a clear naming convention across tests, which should make things easier to debug and maintain. As long as we’re meeting those goals, I think we’re in a good spot.

Happy to get additional input from the team too — @aabidsofi19  @leecalcote

leecalcote: You have my resounding support in any effort to ensure that we are organized, code and tests are legible, purposeful, attest-able, maintainable, and so forth.

[UniqueID] is the test case ID, I presume. Aligning the first column in the Meshery Test Plan spreadsheet's list of test cases might further our traceability. The extent to which a test case ID might also be the same as the (permission) key ID, could be a good thing. See the Permission Keys Reference, https://docs.meshery.io/reference/permissions, which for some (really silly) reason is missing key UUIDs on each row. @Yashsharma1911, is this intentional?

ianrwhitney: So uniqueID in my mind would only be present if we we had to distinguish between several same components or components similar in same viewport but completely optional if additional context doesnt warrant it.

vr-varad:
I understood that we should follow a format for setting data-testid.
So I used this pattern:

<ComponentName>-<SubComponentName>-<Modifier>-<UniqueId>

Let me know what you think @ianrwhitney

PR #14467, opened 2025-04-20
feat: fix settings spec faliures

ianrwhitney:
**Notes for Reviewers**

- This PR aims to fix the failing settings spec. It appears we may have been using an old meshery-isito adapter so i updated this to the newer version of this. 


leecalcote (approved):
Let’s see…

good find!

PR #14508, opened 2025-04-24
remove routing for workspace modal

amitamrutiya:
**Notes for Reviewers**

- This PR fixes #

### DIFF
--- a/ui/components/Lifecycle/Workspaces/WorkspaceDataTable.js
+++ b/ui/components/Lifecycle/Workspaces/WorkspaceDataTable.js
@@ -26,11 +26,10 @@ import {
   Slide,
 } from '@layer5/sistent';
 import { useLegacySelector } from 'lib/store';
-import { useContext, useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { iconSmall } from 'css/icons.styles';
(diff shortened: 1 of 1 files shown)


ui/components/Lifecycle/Workspaces/index.js
```diff
@@ -119,19 +118,16 @@ const Workspaces = () => {
...
-        id: router.query.id,
-        name: router.query.name,
-      });
-    }
-  }, [router.query.id, router.query.name]);
+  const workspaceSwitcherContext = useContext(WorkspaceSwitcherContext);
+  if (workspaceSwitcherContext.selectedWorkspace.id) {
+    selectedWorkspace = workspaceSwitcherContext.selectedWorkspace;
+    setSelectedWorkspace = workspaceSwitcherContext.setSelectedWorkspace;
+  }
```
aabidsofi19, line 129: why are we making state a variable , and not using the setting . this will cause bugs and wrong behavior

amitamrutiya: I made this because we don't have context setup if we are not come from modal and directly access the workspace page. Still this is wrong behavior. I'll improve logic for that.

PR #14635, opened 2025-05-06
fix workspace modal crash in annonymous session

aabidsofi19:
**Notes for Reviewers**

- This PR fixes #


aabidsofi19:
@amitamrutiya  we need to do the following refactoring to prevent such bugs

#1  we need to get rid of legacy redux for organizations and workspaces 
#2  there should be reusable selectors for selectCurrentOrg , selectCurrentWorspace

PR #14672, opened 2025-05-10
fix: added connection name to remote provider

ianrwhitney:
**Notes for Reviewers**

- This PR fixes #
This PR fixes the search functionality to the remote provider. Previously we were not able to search clusters in the connection page via name. This was d/t the name being left off the connection payload.

Before: 
Screenshot 2025-05-07 at 8 07 54 PM

After: 
Screenshot 2025-05-09 at 8 54 54 PM

### DIFF
--- a/server/models/remote_provider.go
+++ b/server/models/remote_provider.go
@@ -732,6 +732,7 @@ func (l *RemoteProvider) SaveK8sContext(token string, k8sContext K8sContext) (co
 		Kind:    "kubernetes",
 		Type:    "platform",
 		SubType: "orchestrator",
+		Name: k8sContext.Name,
 		// Eventually the status would depend on other factors like, whether user administratively processed it or not
 		// Is clsuter reachable and other reasons.
 		Status:           connections.DISCOVERED,


server/models/remote_provider.go
```diff
@@ -732,6 +732,7 @@ func (l *RemoteProvider) SaveK8sContext(token string, k8sContext K8sContext) (co
 		Kind:    "kubernetes",
 		Type:    "platform",
 		SubType: "orchestrator",
+		Name: k8sContext.Name,
```
leecalcote, line 735: Do we know if this section of code strictly and only deals with Kubernetes connections or do we need to also consider this change in context of different connection types?

leecalcote:
it’s most encouraging to see this fixed coming through.

server/models/remote_provider.go
aabidsofi19: i guess from the values above Kind: "kubernetes"  . this piece is specifically for kuberntes

aabidsofi19 (approved):

Issue #14175, opened 2025-03-29
[Docs] Investigate and Plan Integration of Vulnerability Exploitability eXchange (VEX) into Meshery


gentcod:
### Current State
Meshery, as a cloud-native management plane, requires a robust approach to security vulnerability reporting and mitigation. Currently, Meshery lacks direct integration with the [Vulnerability Exploitability eXchange (VEX)](https://cyclonedx.org/capabilities/vex/), which could enhance its ability to provide actionable security insights.

While Meshery does utilize vulnerability scanning tools, incorporating VEX would enable Meshery to reduce false positives, prioritize vulnerabilities more effectively, and improve security posture.

### Desired State

We aim to:

- Investigate how VEX can be integrated into Meshery’s security reporting workflow.

- Explore available tooling and standards around VEX (e.g., OpenVEX, CSAF, and SBOM integrations).

- Define a concrete implementation plan for integrating VEX with Meshery.

- Develop an initial proof-of-concept (PoC) to showcase how Meshery can consume and generate VEX documents.

#### Research and Documentation:

- Understand VEX and its role in vulnerability management.

- Investigate existing open-source implementations and tools supporting VEX.

- Identify any CNCF projects that already leverage VEX.

#### Assessment of Current Workflow:

- Analyze how Meshery currently handles vulnerability management.

- Identify gaps and potential touchpoints for VEX integration.

- Proposal for Implementation:

- Draft a plan for incorporating VEX into Meshery’s security workflow.

- Outline necessary architectural changes, dependencies, and required tooling.

- Define potential APIs or interfaces for VEX ingestion and generation.

#### Proof-of-Concept (Optional but Encouraged):

- Implement a small PoC that integrates VEX with Meshery’s existing security tooling.

- Demonstrate how Meshery can parse and utilize VEX data.

### Expected Outcome

- A well-documented research report on the feasibility and benefits of VEX integration.

- A concrete implementation plan that can be used as the foundation for future development.

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)
- 📚 [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
[...]

PR #14701, opened 2025-05-14
[chore]: Test Vulnerability Scanning

gentcod:
**Notes for Reviewers**

- This PR is related to #14175
- It tests the feasibility of incorporating vulnerability scanning and tracking into Meshery's CI 

### DIFF
--- a/.github/workflows/vex.yaml
+++ b/.github/workflows/vex.yaml
@@ -1,0 +1,54 @@
+name: Vulnerability Scan workflow
+
+on:
+  workflow_dispatch: {}
+
+jobs:
+  vex_scan:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      issues: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install Grype
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin
+
+      - name: Generate CycloneDX SBOM
+        uses: CycloneDX/gh-gomod-generate-sbom@v2
+        with:
+          version: v1
+          args: mod -licenses -json -output bom.json
+
+      - name: Scan vulnerabilities with Grype
+        run: grype sbom:./bom.json -o json > grype-results.json
+
+      - name: Parse Grype Results
+        id: parse_results
+        run: |
+          vulns=$(jq '.matches | length' grype-results.json)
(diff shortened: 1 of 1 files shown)


gentcod:
@leecalcote @lekaf974 @sangramrath

.github/workflows/vex.yaml
```diff
@@ -0,0 +1,54 @@
...
+        id: parse_results
+        run: |
+          vulns=$(jq '.matches | length' grype-results.json)
+          echo "Found $vulns vulnerabilities" > issue_body.txt
+          echo '```json' >> issue_body.txt
+          jq '.matches[0:5]' grype-results.json >> issue_body.txt
+          echo '```' >> issue_body.txt
+          echo "vuln_count=$vulns" >> $GITHUB_OUTPUT
+
+      - name: Meterian Scanner
```
lekaf974, line 40: any output from this step ?

lekaf974 (approved):
LGTM

.github/workflows/vex.yaml
gentcod: So the one I used in my personal repo has a report link output within the logs but the workflow is different and an API key is provided.

This current is a different type that was specified to work with oss projects, so I can say exactly what the output would look like

sangramrath (commented):
I understand we are creating an issue out of them but should we also write these reports somewhere?

gentcod:
Hmmm.. that wouldn't be a bad idea. I guess we can when we make a final decision on the strategy for VEXing

.github/workflows/vex.yaml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+            oss: true
+
+      - name: Create GitHub Issue
+        if: steps.parse_results.outputs.vuln_count != '0'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh issue create \
+            --title "Vulnerability Scan Report - $(date +'%Y-%m-%d %H:%M:%S %Z')" \
```
leecalcote, line 52: Thanks for this PR, @gentcod.

So long as this workflow is seldomly dispatched, having a new issue opened won't be too much of an annoyance. It's worth noting that this approach to reporting is a slippery slope, though.

leecalcote (approved):

.github/workflows/vex.yaml
gentcod: I believe while we iterate and keep on making research and observations; we'd figure out more possible effective ways to tackle it.

PR #14774, opened 2025-05-21
Models: Cleanup and archival of old models

miacycle:
Signed-off-by: miacycle <184569369+miacycle@users.noreply.github.com>


leecalcote (approved):
Thanks, @miacycle!

PR #14862, opened 2025-05-31
fix: registry filter undefined. Added null checks && filter checks.

FaheemOnHub:
this issue was not easily reproducing on localhost, but got it 2-3 times. 
Added some null checks && fixes to prevent undefined.

The error happens because when switching tabs quickly:
The Models tab data is still in the state
When clicking Registrants tab, this Models data gets passed to MesheryTreeViewRegistrants
The component tries to access registrant.models on a model object which doesn't have that property, causing the error

after fix , was not getting this issue in localhost.

### DIFF
--- a/ui/components/Settings/Registry/MesheryTreeViewRegistrants.js
+++ b/ui/components/Settings/Registry/MesheryTreeViewRegistrants.js
@@ -34,7 +34,12 @@ const MesheryTreeViewRegistrants = ({
       selected={selected}
     >
       {data
-        ?.filter((item) => item?.id && (item?.summary || item?.models))
+        ?.filter((item) => item?.summary || item?.models)
+        ?.map((registrant) => (
+          <StyledTreeItem
+            key={registrant.id}
+            nodeId={registrant.id}
+            data-id={registrant.id}
             top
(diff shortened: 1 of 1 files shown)


FaheemOnHub:
@amitamrutiya kindly take a look!

ui/components/Settings/Registry/MesheryTreeViewRegistrants.js
```diff
@@ -33,47 +33,48 @@ const MesheryTreeViewRegistrants = ({
...
-                    setShowDetailsData={setShowDetailsData}
-                    showDetailsData={showDetailsData}
-                  />
-                ))}
-            </StyledTreeItem>
-          </div>
-        </StyledTreeItem>
-      ))}
+      {data
+        ?.filter((item) => item?.id && (item?.summary || item?.models))
```
amitamrutiya, line 37:
```suggestion
        ?.filter((item) => item?.summary || item?.models)
        ?.map((registrant) => (
          <StyledTreeItem
            key={registrant.id}
            nodeId={registrant.id}
            data-id={registrant.id}
```

amitamrutiya (approved):

PR #14984, opened 2025-06-06
fix: stabilize Kanvas snapshot workflow by removing duplicate checkou…

Ayushmore1214:
…t and adding concurrency  - Removed redundant checkout step causing possible conflicts - Added concurrency group and cancel-in-progress to prevent overlapping runs - Simplified filePath input logic for better default handling - Added timeout to avoid hanging workflow runs  This improves reliability and reduces flaky errors in the Kanvas snapshot GitHub Action. Signed-off-by: Ayush More <ayushmore42595@gmail.com>

Pull Request description
Summary:
This PR fixes the Meshery Kanvas snapshot workflow (kanvas.yml) by addressing several issues that were causing flaky or incorrect failures.

Details of the fix:

Removed a duplicated actions/checkout@v4 step that was unnecessarily checking out the repository twice, which could cause conflicts during workflow runs.

Added a concurrency section to ensure only one instance of the workflow runs for a given pull request number, and cancel any in-progress runs when a new one triggers. This helps reduce overlapping runs causing failures.

Simplified the conditional expression for the filePath input to correctly default to 'install/deployment_yamls/k8s' when no input is given.

Added a timeout-minutes parameter (15 minutes) to avoid the workflow hanging indefinitely in rare cases.

Testing:

Tested this workflow on my fork by opening multiple pull requests and pushing changes to simulate real usage.

Verified that the Kanvas snapshot step completes successfully without the previous flaky errors.

Confirmed that concurrency cancels older runs as expected.

**Notes for Reviewers**

- This PR fixes #

### DIFF
--- a/.github/workflows/kanvas.yml
+++ b/.github/workflows/kanvas.yml
@@ -1,11 +1,12 @@
 name: Meshery Snapshot Service
-on: # rebuild any PRs and main branch changes
+
+on:
   pull_request_target:
     types: [opened, synchronize, reopened]
-  
+
   workflow_call:
     inputs:
-      fileName: 
+      fileName:
         description: Relative file path from the root directory
         required: true
         type: string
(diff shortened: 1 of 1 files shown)


Ayushmore1214:
Note: The CI failure is due to unrelated lint errors in the UI codebase and not caused by this workflow fix. This PR only modifies the GitHub Actions YAML file.

PR #15008, opened 2025-06-08
improvement: graphql resolver resync cluster handler

n2h9:
**Notes for Reviewers**

This PR:
- fixes the logic of resync cluster graphql resolver. 
- add functionality to trigger meshsync resync;

Updates graphql resolver:
- access machine for particular connection;
- call resync resources on machine;

Updates to state machine:
- on the machine level add helper to resync resources (it is in helper package to avoid cyclic links, as we need to access kubernetes machine implementation);
- in kubernetes machine access  MesheryControllersHelper through the context relationship and  call Resync meshsync ;

Updates to MesheryControllersHelper and MeshsyncDataHandler
- add function to send request via broker to trigger meshsync resync; 

GetMachineCtx
this function requries event builder, because it is used in asynchronous actions of state machine; 
we need to convert interface to machine context, but we have synchronous action (api call). So probably do not need event builder here, but I am not sure :slightly_smiling_face:  
That is why I introduced a separate function GetMachineCtxPlain which operates without event builder. 


server/machines/kubernetes/helpers.go
```diff
@@ -32,13 +32,19 @@ func GenerateClientSetAction(k8sContext *models.K8sContext, eventBuilder *events
...
+	if err != nil {
 		eb.WithSeverity(events.Error).WithMetadata(map[string]interface{}{
 			"error": err,
 		})
-		return nil, err
+	}
+	return machineCtx, err
+}
+
+func GetMachineCtxPlain(machinectx interface{}) (*MachineCtx, error) {
```
aabidsofi19: the eventbuilder in the GetMachineCtx is a pointer , so might just add some nil checks there and pass nil where we dont have access to the eventbuilder . ( will prevent duplication of logic a bit )

n2h9: Makes sense, thank you Aabid :hugs: ! 

Updated :white_check_mark:

aabidsofi19 (approved):

PR #15014, opened 2025-06-09
Fix: Avatar Redirect Link

vr-varad:
**Notes for Reviewers**

- This PR fixes #

The PR redirects the avatar clicking to the user account. (cloud.layer5.io/user/<user_id>)


ui/components/User.js
```diff
@@ -28,7 +29,7 @@ const User = (props) => {
   } = useGetLoggedInUserQuery();
 
   const getProfileUrl = () => {
-    return (account || [])?.find((item) => item.title === 'Cloud Account')?.href;
+    return `${MESHERY_CLOUD_PROD}/user/${userData?.id}`
```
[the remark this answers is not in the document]
vr-varad: u are right

ui/components/User.js
```diff
@@ -3,19 +3,18 @@ import { Avatar, Button } from '@sistent/sistent';
 import NoSsr from '@mui/material/NoSsr';
 import Link from 'next/link';
 import { useGetLoggedInUserQuery } from '@/rtk-query/user';
-import ExtensionPointSchemaValidator from '../utils/ExtensionPointSchemaValidator';
 import { useNotification } from '@/utils/hooks/useNotification';
 import { EVENT_TYPES } from 'lib/event-types';
 import { IconButtonAvatar } from './Header.styles';
 import { useDispatch, useSelector } from 'react-redux';
 import { updateUser } from '@/store/slices/mesheryUi';
+import { MESHERY_CLOUD_PROD } from '@/constants/endpoints';
```
leecalcote: This needs to be renamed to something like REMOTE_PROVIDER_URL.

leecalcote: This is true of all other references to Meshery Cloud, too.

leecalcote:
Thanks for updating pervasively. It looks like the build tests are passing, too.

amitamrutiya (changes requested):
@vr-varad Currently, we are hardcoding the remote provider URL from a constants file in multiple places, which is not ideal. I understand this was done previously, but with the current setup, if you run Meshery Cloud and Meshery UI locally at the same time, it still redirects to the production Cloud URL.

I recommend fetching the provider URL from the user’s capabilities instead. I think we are storing capability information in the capabilitiesRegistry Redux store, so you can retrieve it from there and use it consistently throughout the app.

sudhanshutech:
@vr-varad any update over here?

Makefile
```diff
@@ -56,7 +56,7 @@ docker-local-cloud:
 docker-cloud:
 	(docker rm -f meshery) || true
 	docker run --name meshery -d \
-	-e PROVIDER_BASE_URLS=$(MESHERY_CLOUD_PROD) \
+	-e PROVIDER_BASE_URLS=$(REMOTE_PROVIDER_URL) \
```
leecalcote, line 59: @vr-varad is the inclusion of this in this PR intentional? I think this is a good change, but it's off base from the PR's title.

saurabhraghuvanshii:
Closing due to inactivity.

PR #15022, opened 2025-06-09
Fix: sorting by count now works as expected

padmaksh-mishra:
**Notes for Reviewers**

This PR fixes **wrong sorting behavior** when sorted by count in the dashboard/settings/reset and some beautification of comments by using multi-line comments.

### Approach
- Figured out that Counting & sorting by count isn’t supported by GORM unless you expose that as a real column or view.
- Thus switched from GORM-based paging + in-Go counting to a single raw-SQL “UNION ALL” query. 
- We still PLUCK the list of table names, but then build one derived table that returns (name, count) for each in a single SQL statement.
- SanitizeOrderInput now allows count

### Result
image

image


Ayushmore1214:
Hey @padmaksh-mishra 👋,

Great job on identifying the GORM limitation and creatively working around it with the raw UNION ALL query! The switch to raw SQL makes sense here and looks cleanly implemented. I also appreciate the multi-line comment formatting—it really improves readability!

A few small suggestions echoing Gemini's review:
- Cleaning up the fmt.Printf debug statement would keep things tidy.
- Adjusting the SanitizeOrderInput whitelist to just "name" and "count" would prevent possible query mismatches.
- A quick check for an empty tableNames slice before building the SQL query would make this more robust.
Really solid work overall—this will definitely improve UX in the database view. Thanks for the contribution!

padmaksh-mishra:
> Hey @padmaksh-mishra 👋,
> 
> Great job on identifying the GORM limitation and creatively working around it with the raw UNION ALL query! The switch to raw SQL makes sense here and looks cleanly implemented. I also appreciate the multi-line comment formatting—it really improves readability!
> 
> A few small suggestions echoing Gemini's review:
> 
>     * Cleaning up the fmt.Printf debug statement would keep things tidy.
> 
>     * Adjusting the SanitizeOrderInput whitelist to just "name" and "count" would prevent possible query mismatches.
> 
>     * A quick check for an empty tableNames slice before building the SQL query would make this more robust.
>       Really solid work overall—this will definitely improve UX in the database view. Thanks for the contribution!

Yes I have done those changes and pushed the commits.

vishalvivekm:
@padmaksh-mishra 

Thank you for your contributions! Let's discuss this today, during Meshery Development call at 9:00 AM CT | 7:30 PM IST. Add it as an agenda item to the 📝 [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit?pli=1&tab=t.0#), if you would. https://meet.meshery.io/dev

padmaksh-mishra:
> @padmaksh-mishra
> 
> Thank you for your contributions! Let's discuss this today, during Meshery Development call at 9:00 AM CT | 7:30 PM IST. Add it as an agenda item to the 📝 [meeting minutes](https://docs.google.com/document/d/1Dg4Uz9rGTuGHmYZwIt47z1j5x3LmnpxD1M1SgypAkQ0/edit?pli=1&tab=t.0#), if you would. https://meet.meshery.io/dev

Will hop in the next meet currently OOO 
Thank you for responding

ui/components/DatabaseSummary.js
```diff
@@ -105,8 +105,9 @@ const DatabaseSummary = (props) => {
       if (searchText) setPage(0);
       setSearchText(searchText != null ? searchText : '');
     }),
-    onColumnSortChange: (_, direction) => {
-      setSortOrder(`name ${direction}`);
+    onColumnSortChange: (changedColumn, direction) => {
```
sudhanshutech, line 108: why changecolumn prop?

padmaksh-mishra: It is because we want to do a raw sql query to return the respective column in sorted manner. Since we are offloading the sorting to the db we need to specify that in the query.

Please look into a line form database_handlers.go

```
raw := fmt.Sprintf(`
		SELECT name, count
		  FROM (%s)
		 ORDER BY %s <---- THIS COULD BE ORDER BY COUNT ASC etc etc
		
	`,
```

Now we will use the retrieved table

server/handlers/database_handlers.go
```diff
@@ -7,68 +7,117 @@ import (
...
-//
-// ```?page={page-number}``` Default page number is 1
-//
-// ```?pagesize={pagesize}``` Default pagesize is 10. To return all results: ```pagesize=all```
-//
-// ```?search={tablename}``` If search is non empty then a greedy search is performed
-// responses:
-//
-//	200: systemDatabaseResponseWrapper
+/*
```
n2h9: Hey hey :innocent: 

Probably not very important,  but I think that every where else in api handlers this decsription comment is arranged with 
// 
comments

I don know which is better, but if everywhere else it is like //, do we really need to update this one to /**/ here? :thinking:

Ayushmore1214:
I have used the changedColumn parameter here , bcoz it allows the backend to dynamically sort by the column selected in the UI also it makes sure that the raw sql query applies the right ordering

server/handlers/database_handlers.go
padmaksh-mishra: I understand. Reverted the changes.
Did it because it looked cleaner, but yes keeping it consistent is more apt.
Thank you for reviewing

padmaksh-mishra:
> I have used the changedColumn parameter here , bcoz it allows the backend to dynamically sort by the column selected in the UI also it makes sure that the raw sql query applies the right ordering

Yes, you are correct. Since I am querying directly SQL and there is an order by clause we need the column name

matrixkavi:
@padmaksh-mishra - Are you still working on the issue? If yes, ensure that the branch is updated and you post a video of sort by count working locally.

padmaksh-mishra:
> @padmaksh-mishra - Are you still working on the issue? If yes, ensure that the branch is updated and you post a video of sort by count working locally.

No longer working on the issue. Closed the PR

Thank you

PR #15029, opened 2025-06-10
Resolved code break for doc(mesheryctl registry generate page)

Paladian78:
**Notes for Reviewers**

- This PR fixes code break in the page
https://docs.meshery.io/reference/mesheryctl/registry/generate#examples

image

### DIFF
--- a/docs/pages/reference/mesheryctl/mesheryctl-registry-generate.md
+++ b/docs/pages/reference/mesheryctl/mesheryctl-registry-generate.md
@@ -33,32 +33,32 @@ mesheryctl registry generate [flags]
 Generate Meshery Models from a Google Spreadsheet (i.e. "Meshery Integrations" spreadsheet).
 <pre class='codeblock-pre'>
 <div class='codeblock'>
-mesheryctl registry generate --spreadsheet-id "1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw" --spreadsheet-cred $CRED
-
+<code class="clipboardjs">mesheryctl registry generate --spreadsheet-id "1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw" --spreadsheet-cred $CRED
+</code>
 </div>
 </pre> 
 
 Directly generate models from one of the supported registrants by using Registrant Connection Definition and (optional) Registrant Credential Definition
 <pre class='codeblock-pre'>
 <div class='codeblock'>
-mesheryctl registry generate --registrant-def [path to connection definition] --registrant-cred [path to credential definition]
-
+<code class="clipboardjs">mesheryctl registry generate --registrant-def [path to connection definition] --registrant-cred [path to credential definition]
+</code>
 </div>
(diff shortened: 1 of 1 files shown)


l5io:
🚀 Preview for commit d81171d01d9a599111ca5fd40a448e77b251f266 at: https://6848213ca0a22309640a621c--meshery-docs-preview.netlify.app

l5io:
🚀 Preview for commit a5c1d3fca2296c3b8fc8e0c73e288daa9252e247 at: https://68482884d2833b1d76d728bf--meshery-docs-preview.netlify.app

l5io:
🚀 Preview for commit 56c58c9c6d71b53328f179424b923b5678e232d9 at: https://68482f14da0f0a2b48e4c0f9--meshery-docs-preview.netlify.app

leecalcote (approved):
I wonder if these will return after the CLI docs are auto-generated again.  @lekaf974 or @alphaX86, do you know?

Paladian78:
@leecalcote  It is still not resolved. In the preview section I checked and the behavior is still the same. Last time when i had this issue @zihanKuang suggested this and it worked but in this case it's not resolving. @zihanKuang Can you help me with this?

zihanKuang:
Hi @Paladian78,

I haven't had a chance to test this myself, but here's a thought, building on what @leecalcote suspected.

The issue is likely that these docs are auto-generated from the Go source code, so the fix probably needs to happen in the source file itself, not the Markdown file.

My guess is you might need to find the `Example:` string in the command's `.go` file and wrap the placeholder in backticks. For example:
https://github.com/meshery/meshery/blob/master/mesheryctl/internal/cli/root/model/generate.go

From: `... mesheryctl registry generate --directory <DIRECTORY_PATH> ...`
To: ```... mesheryctl registry generate --directory \`<DIRECTORY_PATH>\` ...```

After that, you'd have to regenerate the docs with the project's build script.

It's just a theory, but might be worth a try! If this approach works, you could then apply the same fix to the other files with this issue, or feel free to use a better method if you discover one.👍

alphaX86:
> I wonder if these will return after the CLI docs are auto-generated again.  @lekaf974 or @alphaX86, do you know?

The docs will be generated via GH Actions only when PR is merged. Alternatively, to verify the changes, the docs can be built manually by running the make command

lekaf974:
@Paladian78 

this will not fix the issue since it will be overridden by automation, the real issue is that `<` and `>` are not escaped properly with jekyll so an easy fix is to update https://github.com/meshery/meshery/blob/4e3b1965c3cb83069b4a15b28d28d133db8a563b/mesheryctl/internal/cli/root/registry/generate.go#L60 replacing `<DIRECTORY PATH>` by `[directory-path]` which will make the doc following the same convention used currently

pontusringblom:
@lekaf974, will you please implement your suggested change? @Paladian78 has abandoned his effort here.

lekaf974:
> @lekaf974, will you please implement your suggested change? @Paladian78 has abandoned his effort here.

#15453

PR #15101, opened 2025-06-16
Revert "fix: omit non func design-id"

leecalcote:
Reverts meshery/meshery#15100

@FaheemOnHub, the request is for the event to display a hyperlink to the design. 

Did you attempt to open the design and the design would not open? If the design opens, but produces another warning message, that's fine.

### DIFF
--- a/ui/components/NotificationCenter/metadata.js
+++ b/ui/components/NotificationCenter/metadata.js
@@ -110,7 +110,7 @@ export const FormattedMetadata = ({ event }) => {
   }
 
   const metadata = {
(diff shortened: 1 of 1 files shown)


FaheemOnHub:
> Reverts #15100
> 
> @FaheemOnHub, the request is for the event to display a hyperlink to the design.
> 
> Did you attempt to open the design and the design would not open? If the design opens, but produces another warning message, that's fine.

Yes i had attempted & the design does not opens , throws the same warning again 🔁 .

https://github.com/user-attachments/assets/36ee8765-e65a-4f40-bab7-432139f5c98d

leecalcote:
@FaheemOnHub, ok, thank you for confirming. 👍

Issue #13546, opened 2025-02-10
[mesheryctl] model init and build cmds


vishalvivekm:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

### Mockups

---

### Contributor Resources

- ⌨️ [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)

rabelmervin:
Hello @vishalvivekm can I contribute this issue ?

lekaf974:
> Hello [@vishalvivekm](https://github.com/vishalvivekm) can I contribute this issue ?

Do you have already used mesheryctl and mesheryctl model command ?

rabelmervin:
No @lekaf974 I'm new but, can learn and do it !

leecalcote:
@rabelmervin, a progress report?

leecalcote:
@rabelmervin, good! Allow me to set clear expectations: If you don't demo in this week's dev meeting, I will assume that you have in fact not made any progress, will not be making any progress, and will free assignment for another volunteer. Deal?

rabelmervin:
@leecalcote  I gave my best effort to solve this issue, but I'm currently stuck on the model scaffolding. This has made me realize I need to work on improving my problem-solving skills.

I'm finding it more challenging than I expected, so I’d like to unassign myself to let someone else take it up. Thank you for the opportunity!

leecalcote:
I appreciate this, @rabelmervin. There are plenty of other issues where this comes from. While this particular issue didn’t immediately work out, please don’t be discouraged. Let’s get you assigned to another one.

gourav-k-shaw:
i would like to work on this.
may you please assign
thanks

[...]

PR #15165, opened 2025-06-23
update model build command to receive an output location and no-version flag

vishalvivekm:
**Notes for Reviewers**
- This PR adds two flags to the model build command:
  - `--output-location`: to specify the output directory for the generated OCI artifact
  - `--no-version`: boolean flag to exclude version name from the generated artifact's name

**Reason for no-version flag**

Currently `./mesheryctl exp model build aws-vpc-cni/1.19.5` would generate an OCI artifact named `aws-vpc-cni-1-19-5.tar`. 

If we were to write this tar file to sites like https://github.com/meshery/meshery.io under `assets/modelsFiles`, we can't reference the file on its individual page for download because the model version information `<model-name>/<model-version>/v1.0.0` is not available to these sites.

```html
{% assign modelName = page.path | split: '/' | last | replace: '.md', '' %} 
<a href="/assets/modelsFiles/{{ modelName }}.tar" download> 
  <button class="import" style="width: 10rem;">Download Model</button> 
</a> 
```

We simply have the modelName, not the version number, so the flag `--no-version` (or short `-n`) ensures the version is stripped from the generated OCI image, allowing the artifact to be easily referenced.

### DIFF
--- a/mesheryctl/internal/cli/root/model/build.go
+++ b/mesheryctl/internal/cli/root/model/build.go
@@ -23,6 +23,18 @@ Documentation for exp model and subcommands can be found at https://docs.meshery
 	Example: `
 // Create an OCI-compliant package from the model files
 mesheryctl exp model build [model-name]/[model-version]
+
+// Create package in specific output location
+mesheryctl exp model build [model-name]/[model-version] -o /path/to/output
+
+// Create package without version in filename (for Jekyll sites)
+mesheryctl exp model build [model-name]/[model-version] --no-version
+
+// Use short flag
+mesheryctl exp model build [model-name]/[model-version] -n
+
+// Combine both flags
+mesheryctl exp model build [model-name]/[model-version] -o /path/to/output -n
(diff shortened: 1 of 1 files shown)


n2h9:
Hey Vivek, :wave: ! 

Thank you for your enhancements to model build command :innocent: :bow: !

Looks good for me, I think we are good to go.  :hugs: 

`--output-location` has a good naming, very clear how is it supposed to use.
`--no-version` is less  obvious, I didn't catch from the very begging its purpose. 

I think it is fine to have it as you  have suggested.

Also some thoughts on possible work around image naming. 

1) if we generally do not need version in file name, maybe we could just remove it in general version :woman_shrugging: , I think there were no such requirement in initial issue https://github.com/meshery/meshery/issues/13546 

2) maybe we can have a general name by default like [model-name].tar and give an option to provide a custom name.

n2h9 (approved):

PR #15840, opened 2025-09-18
[UI] Fix some errors flooding the console on page load.

kuv2707:
**Notes for Reviewers**

- This PR fixes some of the errors that appear on the console mainly during app load. These errors are relatively minor.
- Many of the remaining errors require more involved work, and it might not be worth it as almost all of these will surface when/if we migrate to TypeScript and can be fixed then.

[Slack discussion](https://mesheryio.slack.com/archives/C023QGMSH8D/p1758091777096229)

### DIFF
--- a/ui/components/Dashboard/style.js
+++ b/ui/components/Dashboard/style.js
@@ -43,7 +43,9 @@ export const ConnectClusterText = styled(Typography)({
   marginBottom: '0.5rem',
 });
 
-export const HoneycombRoot = styled(Card)(({ theme, isEditMode }) => ({
+export const HoneycombRoot = styled(Card, {
+  shouldForwardProp: (prop) => prop !== 'isEditMode',
+})(({ theme, isEditMode }) => ({
   backgroundColor:
     theme.palette.mode === 'dark' ? theme.palette.background.card : theme.palette.common.white,
   padding: theme.spacing(2),
--- a/ui/components/General/style.js
+++ b/ui/components/General/style.js
@@ -246,7 +246,9 @@ export const SideBarListItem = styled(ListItemButton, {
   fontSize: '14px',
 }));
 
-export const SideBarText = styled(ListItemText)(({ drawerCollapsed }) => ({
+export const SideBarText = styled(ListItemText, {
+  shouldForwardProp: (prop) => prop !== 'drawerCollapsed',
+})(({ drawerCollapsed }) => ({
   opacity: drawerCollapsed ? 0 : 1,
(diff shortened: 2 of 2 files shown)


ui/components/LoadingComponents/DynamicFullscreenLoader.js
```diff
@@ -4,27 +4,29 @@ import LoadingScreen from './LoadingComponentServer';
...
+                top: 0,
+                left: 0,
+                zIndex: 9999999,
+                height: '100vh',
+                width: '100vw',
+              }}
+            />,
+            document.body,
+          )}
+        </>
```
ARYANSHAH1567, line 29: @kuv2707, Why is there a extra react fragment used here, because we are returning only a single element here.

kuv2707: I've mentioned the reason in the corresponding commit message:

>The component NoSsr expects a react node and not a
children prop. So we wrap its children in an empty
jsx fragment. This is indeed a minor nuance.

Without the fragment wrap, it showed an error in the console.

ARYANSHAH1567: On the official page for MUI, https://mui.com/material-ui/api/no-ssr/
It states that the component accepts a children prop. Can you please verify once is the change necessary?

Darshan174:
@kuv2707 Thank You for your contribution!!
If this work is complete, this is a good item to add to the weekly [Meshery Development meeting agenda](https://docs.google.com/document/d/15IEwtTxPOkQXN2r_1LzGXuAv2mIUlmbdtmxpyxXF-KI/edit). You can add this item in the doc, attend, and present it. Meeting details at https://meet.layer5.io/meshery.

ui/components/LoadingComponents/DynamicFullscreenLoader.js
kuv2707: image

This is the error that appeared prior to adding the fragment.

I guess the wording is deceptive. Looks like it expects `children` to be of type `ReactNode` but the return value of `createPortal` isn't exactly of that type.

kuv2707: https://github.com/mui/material-ui/blob/master/packages/mui-material/src/NoSsr/NoSsr.tsx

The source file indicates that the package `prop-types` is used for runtime prop checking.

kuv2707: @ARYANSHAH1567 Anything else here?

PR #15857, opened 2025-09-19
[mesheryctl] Remove redundant error guard clause.

kuv2707:
We're already checking for `err` to be `nil` a little earlier. This was discovered through the static analyzer.

image

### DIFF
--- a/mesheryctl/internal/cli/root/config/config.go
+++ b/mesheryctl/internal/cli/root/config/config.go
@@ -250,7 +250,7 @@ func (ctx *Context) ValidateVersion() error {
 	client := &http.Client{}
 	resp, err := client.Do(req)
 	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


n2h9 (approved):
good catch, thank you @kuv2707 :bowing_woman:

mesheryctl/internal/cli/root/config/config.go
```diff
@@ -250,7 +250,7 @@ func (ctx *Context) ValidateVersion() error {
 	client := &http.Client{}
 	resp, err := client.Do(req)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to make GET request to %s", url)
```
Darshan174, line 253: 👍

Darshan174 (approved):

PR #16049, opened 2025-10-03
Create .aiexclude

yi-nuo426:
These folders don't require AI-assisted code review.


simihablo (approved):
I agree. The comments only get in the way.

Issue #16186, opened 2025-10-11
Change Kubernetes connection event action from "registering" to "register"


leecalcote:
### Environment

- **Host OS:** Mac Linux Windows
- **Platform:** Docker or Kubernetes
- **Meshery Server Version:** stable-v
- **Meshery Client Version:** stable-v

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/qJy1c95qirjgWQODApilR9)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)

<img width="665" height="536" alt="Image" src="https://github.com/user-attachments/assets/266b546d-098f-4fc4-9647-b02a492724a8" />

JunaidAli-dev:
Hello @leecalcote I would like to work on this issue, could you please assign this to me!

vtushar06:
Hey @leecalcote,
I’d like to start working on this issue if it’s still available, as I noticed it’s not currently assigned to anyone. I’ve been going through the Meshery documentation, and since Kubernetes is a topic I’ve recently been studying, I believe this would be a great opportunity to contribute and apply my learnings. I’ll make sure to work on it promptly and aim to resolve it as soon as possible.

Thanks!

shri33:
Hi all & @leecalcote,
I noticed others showed interest in this issue, but I had already worked on it and completed the fix, so I’ve submitted this PR for review. The changes focus solely on ensuring consistency by updating the Kubernetes event action from "registering" to "register".

Looking forward to feedback from the maintainers.

PR #16190, opened 2025-10-12
Fix: Change Kubernetes event action from 'registering' to 'register'

shri33:
## Description
This PR fixes issue #16186 by changing the Kubernetes connection event action from "registering" to "register" for consistency.

## Changes Made
- Updated `RegistrationStatus.String()` method in `server/models/k8s_components_registration.go`
- Changed `Registering` case to return "register" instead of "registering" on line 46
- Ensures consistency with other connection event actions throughout the codebase

## Issue
Resolves: #16186

## Impact
- UI filters will show consistent `action: register` 
- Event logs will use uniform "register" action terminology  
- API responses will have consistent event action naming

## Testing
- [x] Code change applied and verified
- [x] Single line modification confirmed
- [ ] UI testing needed to verify "register" action appears instead of "registering"

### DIFF
--- a/server/models/k8s_components_registration.go
+++ b/server/models/k8s_components_registration.go
@@ -45,7 +45,7 @@ func (rs RegistrationStatus) String() string {
 	case NotRegistered:
 		return "not_registered"
 	case Registering:
-		return "registering"
+		return "register"
 	default:
 		return ""
 	}


leecalcote:
@shri33 thank you. Can you confirm that this is the string used as the event action and not the string used to track the connection state?

shri33:
@leecalcote Thank you for reviewing! 

Yes, I can confirm this change affects the **event action string** that gets displayed in the UI and logs, not the internal connection state tracking.

**Context:**
- The `RegistrationStatus.String()` method returns the string representation used for **displaying/logging** the registration status
- This string appears in UI filters and event logs as the "action" field
- The internal enum value `Registering` remains unchanged - only its string representation changes from "registering" to "register"

**Evidence from the issue #16186:**
The issue specifically mentions inconsistency in UI display where users see `action: registering` instead of the expected `action: register` format used by other connection events.

This change only affects the **display/logging layer**, not the underlying state management logic.

shri33:
@leecalcote please review this

PR #16266, opened 2025-10-22
Feat: refactor and remove duplicate code make query fast

saurabhraghuvanshii:
**Notes for Reviewers**
#### Optimize Models Directory: Performance & Code Quality Improvements

- HTTP Client Pooling: Implemented connection reuse across remote provider operations
- Database Query Optimization: Fixed N+1 query problems with batch operations 
- Batch Operations: Added CreateInBatches() for bulk inserts (100 records per batch)
- Query Builder Pattern: Fluent API for consistent, optimized database queries

#### Code Deduplication & Architecture

- Generic Marshal Utilities: Replaced 20+ duplicate marshal functions with 2 reusable utilities
- Base Persister Pattern: Common database operations abstracted into BasePersister class
- Consistent Error Handling: Proper error propagation instead of silent failures
- Constructor Functions: Standardized persister initialization patterns

#### Backward compatible
 - All existing APIs maintained, just optimized implementation.

 This PR fixes #


server/models/base_persister.go
```diff
@@ -0,0 +1,146 @@
...
+import (
+	"strings"
+
+	"github.com/gofrs/uuid"
+	"github.com/meshery/meshkit/database"
+	"gorm.io/gorm"
+)
+
+// BasePersister provides common database operations for all persisters
+type BasePersister struct {
```
lekaf974, line 12:
```suggestion
type BaseDBPersister struct {
```

Since it is only db related I think we should add it in the name but not sure, what do you think about it ?

server/models/http_client_pool.go
```diff
@@ -0,0 +1,78 @@
...
+// Global HTTP client pool instance
+var (
+	globalHTTPClientPool *HTTPClientPool
+	poolOnce             sync.Once
+)
+
+// GetHTTPClientPool returns the global HTTP client pool instance
+func GetHTTPClientPool() *HTTPClientPool {
+	poolOnce.Do(func() {
+		globalHTTPClientPool = NewHTTPClientPool(10)
```
lekaf974, line 75: Should we make pool size configurable instead of hardcoded 10 value ?

server/models/http_client_pool.go
lekaf974, line 1: not sure this file should be in `models` folder what do you think to move it a new one example `internal/http` ?

server/models/marshal_utils.go
lekaf974, line 1: If I am right we have already a function is doing that in `mehsery/meshkit` could you reuse it and enhance existing one if needed ?

Also I'd not create an util file in `models` folder

server/models/base_persister.go
```diff
@@ -0,0 +1,146 @@
...
+// Paginate applies pagination to the query
+func (qb *QueryBuilder) Paginate(page, pageSize uint64) *QueryBuilder {
+	if pageSize > 0 {
+		qb.query = Paginate(uint(page), uint(pageSize))(qb.query)
+	}
+	return qb
+}
+
+// Count returns the total count of records
+func (qb *QueryBuilder) Count(tableName string) int64 {
```
lekaf974, line 99: Not sure query builder should be responsible to return count of a table, I think it should be another struct that should call the builder and then implement Count,, does it make sense to you ?

server/models/base_persister.go
```diff
@@ -0,0 +1,146 @@
...
+	var count int64
+	qb.query.Table(tableName).Count(&count)
+	return count
+}
+
+// Find executes the query and populates the result
+func (qb *QueryBuilder) Find(dest interface{}) error {
```
lekaf974, line 106: Same comment than I did for `Count`, should be another struct which is responsible  of it

server/models/sql-utils.go
```diff
@@ -29,28 +29,38 @@ func SanitizeOrderInput(order string, validColumns []string) string {
 
 	inputCol := parsedOrderStr[0]
 	typ := strings.ToLower(parsedOrderStr[1])
+
+	// Create a map for O(1) lookup instead of O(n) loop
```
lekaf974, line 33: Looks to me that the comment is a little misleading, we are still doing a O(n) loop to feed the map, but this approach is more cleaner to understand

lekaf974 (changes requested):
Good Job, I added some comments could you take a loook.

PR #16661, opened 2025-12-15
Update Eslint-check workflow to run in check only mode 

### LINKED ISSUE #16657: [CI]Update the eslint-check workflow to run ESLint in check-only mode
### Current Behavior
The `eslint-check `workflow fails for fork-based PRs because it attempts to auto-commit and push lint fixes using `git-auto-commit-action`. For security reasons, GitHub provides a read-only `GITHUB_TOKEN` to workflows triggered from forks, which prevents pushing commits back to the PR branch. As a result, the workflow errors out (e.g., `fatal: invalid reference`) even when ESLint passes.

### Desired Behavior
CI workflows should not modify or commit code during PR validation, linting should run in check-only mode instead.

### Implementation
Update the `eslint-check` workflow to run ESLint in check-only mode and remove the auto-commit step. Replace `npm i eslint` with deterministic installs (`npm ci`) and delete the `git-auto-commit-action` step so the workflow no longer attempts to push changes. This makes the workflow fork-safe, avoids write operations during PR validation, and ensures CI fails only when real lint errors exist.

---
- 🧪 [Meshery Test Plan Spreadsheet](https://docs.google.com/spreadsheets/d/13Ir4gfaKoAX9r8qYjAFFl_U9ntke4X5ndREY1T7bnVs/edit#gid=0)

CodexRaunak:
**Notes for Reviewers**
- CI workflows should not modify or commit code during PR validation, linting should run in check-only mode instead, therefore removed the `GitHub auto commit action` and add separate make targets for checking lint.

- This PR fixes #16657 


CodexRaunak:
I wonder why the `KanvasScreenshot` job Is failing

.github/workflows/eslint-gh.yml
```diff
@@ -10,19 +10,9 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v6
-    - name: Install eslint in Meshery UI and Provider UI
-      run: |
-        cd ui && npm i eslint && cd ..
-        cd provider-ui && npm i eslint && cd ..
+    - name: Install UI dependencies
+      run: make ui-setup-ci
```
leecalcote: This is missing in the makefile.

leecalcote (changes requested):

leecalcote (commented):
This PR is missing a logical check to verify when to --fix and when not to fix.

.github/workflows/eslint-gh.yml
CodexRaunak: [I am referring to this target](https://github.com/meshery/meshery/blob/master/Makefile#L342-L344), as in ci,`npm ci` is the preferred command as it make sure to gets the exact version mentioned in `package.lock.json`. So rather than installing the `eslint`, I preferred this one.
Is that fine?

CodexRaunak:
@lee can u review the PR, we have added a precommit hook that runs the make targets that lints for UI and provider UI.
For the setup:

We just have to do `npm install` in the `ui` or `provider-ui` directory 
 we can run `make ui-setup` as well.
 
 That will run the `prepare` script in the `package.json`, which configures `husky` by setting
` git config core.hooksPath ui/.husky`, It tell git to use instead of `.git/hooks`, look for hooks in `ui/.husky/`.

The commit hook also don't take much time, which is a desired behaviour when committing.
This ensures best developer practices to push linted good code, rather than firstly pushing bad code and then `CI`, validating that it's bad code.

We still have the es linter workflow as for if we by pass commit hooks, CI can catch any issues.

We Would also be adding this to today's development meet (17th Dec) for the discussion.

Thanks for helping @Suhani95.

leecalcote:
I don't see Suhani's approval or feedback here.

Suhani95:
I’ve added my approval and shared my feedback on the PR.

From my side, the intent of this change is to ensure CI is used only for validation and not for modifying or committing code during PR checks. ESLint now runs in check-only mode in CI, while lint fixing is handled locally through pre-commit hooks for both UI and Provider UI. This keeps fork-based PRs safe and avoids CI failures caused by write restrictions.

CI is still retained as a safety net in case commit hooks are bypassed

CodexRaunak:
> I don't see Suhani's approval or feedback here.

Here is the recording of the yesterday's meet discussion as well with timestamp.
https://youtu.be/z3X1CyzloSY?t=1789

Thank You

test.txt
saurabhraghuvanshii, line 1: @CodexRaunak  remove this

saurabhraghuvanshii (changes requested):

ui/.husky/pre-commit
saurabhraghuvanshii, line 1: @CodexRaunak Shouldn’t Husky always be set up at the root level of the project?

saurabhraghuvanshii (changes requested):

ui/.husky/pre-commit
CodexRaunak: Perhaps for meshery it should not currently, as we are running the pre commit hook only for the `ui` and `provider ui` projects.
Also if we configure it in the root, we need to add another `package.json` in the root, and then have a dev dependency  for `husky` for its installation.

Then the developer would be forced to hit `npm install` from root to install and run `husky`, which only lints the `ui` and `provider ui` projects currently.
If in future we also add a go linter or something maybe we can have it in root directory and then modify all the make targets to make sure `husky` is installed initially (added as an extra step).

Also previously in the meshery husky was configured in the [ui project](https://github.com/meshery/meshery/blob/master/ui/.husky/pre-commit).

saurabhraghuvanshii: Thanks @CodexRaunak lgtm!!

CodexRaunak:
Hey @leecalcote , we have accounted the feedback from @hexxdump and @ianrwhitney 
Here’s our thought.

`Approach 1(pre commit hook +CI Validation):` We run `pre commit hook` on staging files rather than linting the whole project.
On top of that we would be having the same CI workflow, that we have currently that acts as a gatekeeper or another layer of safety net, if someone bypasses the hook, the CI workflow will catch it and fail on PR.

Pros:
•⁠  ⁠The master branch always remains clean by catching issues early, reducing CI failure.

`Approach 2(Pre commit hook +Scheduled Auto-Fix):` Same we would having a precommit hook like in `approach 1 `running on staged files.
In addition to that, rather than having a linting workflow on PR, we run it via a cron job on a particular schedule of day to fix linting and commit back the changes to master branch.

Cons: It may pollute the commit history of the project, when the actions would be having a `“Fix linting commit”`, though most of then will be handled by precommit hook, but still.

•⁠  ⁠It may also have a merge conflict when we are modifying the same file locally that linter fixed on the master branch.

We think `Approach 1` is better because it ensures  quality code from the start, keeps the master branch clean, and avoids conflicts and unnecessary commits.
We are open to any suggestions or thoughts on any of the approach.

How we did:
•⁠  ⁠To lint all the staged files only we are using a tool `lint-staged`, which was also already configured in the repo. We modified its config.
•⁠  ⁠Inverted the make target behaviour, now the default behaviour is `auto fix` and if u don’t want, just pass 
`NOT_FIX=true` to run in check only mode.

saurabhraghuvanshii (dismissed):

lekaf974 (dismissed):
LGTM

lekaf974:
@CodexRaunak any reason this has not been merged yet ?

CodexRaunak:
> @CodexRaunak any reason this has not been merged yet ?

@leecalcote Is there anything else that needs to be addressed in this PR

leecalcote:
@guan404ming thoughts?

guan404ming (commented):
Thanks for the change, I think the path parsing could be better without regex in all files

like in `.lintstagedrc.js` we could try

```js
module.exports = {
  '**/*.{js,ts,jsx,tsx}': [
    'eslint --fix --max-warnings=0',
    'prettier --write'
  ]
}
```

PR #16725, opened 2025-12-19
fix(patterns): prevent data race on errs slice in Process  Guard conc…

aravindhaj17:
### What this PR does

This PR fixes a data race in the `Process` function located at  
`server/models/pattern/patterns/patterns.go`.

The `errs` slice was being appended to from multiple goroutines without synchronization, which caused race conditions when processing multiple Kubernetes clients or when running tests with the Go race detector.

### What was changed

- Introduced a `sync.Mutex` (`errsMx`) to guard concurrent writes to the shared `errs` slice
- Ensured all `append(errs, err)` operations are protected by the mutex
- Preserved existing behavior while making the function concurrency-safe

### Why this is needed

Running the following command previously reported data races:

```bash
go test ./... -race
```

### DIFF
--- a/server/models/pattern/patterns/patterns.go
+++ b/server/models/pattern/patterns/patterns.go
@@ -31,7 +31,19 @@ type DeploymentMessagePerContext struct {
 	Location   string
 }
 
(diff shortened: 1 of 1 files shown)


jamieplu (changes requested):
@aravindhaj17 open an issue, documenting the problem that you're addressing.

aravindhaj17:
@jamieplu Thanks! I’ve opened an issue documenting the problem and linked it to this PR. 
Issue: #16754

PR #16733, opened 2025-12-19
[Server] Surface Adapter Plugin details in Notification Center

### LINKED ISSUE #10509: [Server] Surface Plugin details in Notification Center
### Current Behavior
As a platform with me to extension points, Meshery offers the the ability to enable any different number of plugins. Currently, as a platform Meshery doesn't expose any information into Meshery UI about plugins that have been loaded or have been attempted to load outside of Adapter plugins.

### Desired Behavior
Users are aware of the version, size, and details of the plugins that are loaded.

### Implementation
See https://docs.meshery.io/project/contributing/contributing-server-events

### Acceptance Tests
1. On failure to load plugin, sent this to the Notification Center.
1. On success of loading a plugin, send plugin details (manifest and version) to Notification Center.

---

shri771:
**Notes for Reviewers**
Previously, adapter operations showed simple toast notifications. Now, they are sent to the Notification Center with enhanced details such as version, manifest, and error remediation guidance.

-  Additionally when attempting to configure another adapter at an already-used URL, a specific notification is sent to alert the user of the conflict

https://github.com/user-attachments/assets/71f8a9f9-7733-4c9b-bdff-8bf6bab4f24f

- While fixing this I found another bug at  [server/models/default_local_provider.go](https://github.com/org/repo/blob/master/file.go#L123) GORM event persistence
```
// Before (BUG)
err := e.DB.Save(event).Error
// After (FIXED)
err := e.DB.Save(&event).Error

```
- Without the pointer, GORM cannot update the original event.ID field after database insertion, causing event tracking failures.

###  Tests Ran
- [x] On failure to load plugin, sent this to the Notification Center.
- [x] On success of loading a plugin, send plugin details (manifest and version) to Notification Center.

Fixes #10509 


server/handlers/mesh_ops_handlers.go
```diff
@@ -146,22 +212,100 @@ func (h *Handler) MeshAdapterConfigHandler(w http.ResponseWriter, req *http.Requ
...
+				WithDescription("Error adding Adapter").
+				WithAction(models.Register).
+				WithMetadata(metadata).
+				Build()
+
+			_err := provider.PersistEvent(*event, nil)
+			if _err != nil {
+				h.log.Debug(fmt.Sprintf("Failed to persist event: %v", _err))
+			}
+			go h.config.EventBroadcaster.Publish(userID, event)
```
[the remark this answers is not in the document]
shri771: The duplication is intentional here. In the error case (lines 240-247), we need to return immediately after publishing the event without calling RecordPreferences (line 310), since we don't want to persist an incomplete adapter configuration. Consolidating the event handling would require adding `err == nil` checks before RecordPreferences, which would make the control flow less clear.

shri771:
@lekaf974

server/handlers/mesh_ops_handlers.go
```diff
@@ -146,22 +212,100 @@ func (h *Handler) MeshAdapterConfigHandler(w http.ResponseWriter, req *http.Requ
...
-			http.Error(w, ErrRetrieveData(err).Error(), http.StatusInternalServerError)
+			err = ErrRetrieveData(err)
+
+			h.log.Error(err)
+
+			errDescription := errors.GetSDescription(err)
+			errRemedy := errors.GetRemedy(err)
+			errProbableCause := errors.GetCause(err)
+
+			metadata := map[string]interface{}{
```
aabidsofi19, line 185: @shri771 i think errors have toJson method / helper  for creating marshalling into this .

server/handlers/mesh_ops_handlers.go
```diff
@@ -114,6 +117,58 @@ func (h *Handler) AdapterPingHandler(w http.ResponseWriter, req *http.Request, p
 	_, _ = w.Write([]byte("{}"))
 }
 
+// For Configuration category, it further sub-categorizes by prefix (Add-ons, Policies, etc.)
+func categorizeAdapterOps(ops []*meshes.SupportedOperation) map[string]interface{} {
```
aabidsofi19: @shri771  whats the need for this .  we shouldnt be hardcoding the op values again here

server/handlers/mesh_ops_handlers.go
shri771: Hi @aabidsofi19,Do you mean to use a simply `"errors": err` in metadata, also this is used in other handlers.

server/handlers/mesh_ops_handlers.go
shri771: This function is needed to group the adapter ops because without it manifest could look like this 
image
Just the key and value.

shri771: I have renamed the function to `groupOpsByCategory` and removed the hardcoding that was there in function.

suzy-g38:
@saurabhraghuvanshii we can close this pr

PR #16736, opened 2025-12-21
chore: Add missing unit tests on meshery api client

lekaf974:

**Notes for Reviewers**

- This PR fixes #


mesheryctl/internal/cli/pkg/api/meshery_test.go
shri771, line 1: As you suggested, comparing error codes is a better approach since they're more stable and explicit than error messages.

Issue #16735, opened 2025-12-21
[Server] Upgrade from meshkit v0.8.53 to v0.8.54 causes all mesheryctl golden file tests to fail


shri771:
### Current Behavior

The upgrade from meshkit v0.8.53 to v0.8.54 causes all mesheryctl golden file tests to fail. The `Error()` method in meshkit v0.8.54 now appends metadata to error strings:

```
message | Short Description: X | Probable Cause: Y | Suggested Remediation: Z
```

This causes mismatches between the golden files (which expect plain messages) and the actual error output.

**Example failure:**
```
exp: "Please provide a model name. Usage: mesheryctl model search [query-text]"
got: "Please provide a model name. Usage: mesheryctl model search [query-text] | Short Description: Invalid Argument | Probable Cause: Invalid Argument | Suggested Remediation: Please check the arguments passed"
```

### Expected Behavior

Tests should pass. Either:
1. Golden files should be updated to match the new meshkit error format

### Screenshots/Logs

```bash

=== RUN   TestModelInit/model_init_no_model_name
Error: must provide only one argument: model name | Short Description: Error model init | Probable Cause: Error during run of model init command | Suggested Remediation: Ensure passing all params according to the command description
init_test.go:274:

        exp: "must provide only one argument: model name"

        got: "must provide only one argument: model name | Short Description: Error model init | Probable Cause: Error during run of model init command | Suggested Remediation: Ensure passing all params according to the command description"
```

### Environment

- **Host OS:** Linux
- **Platform:** N/A (unit tests)
- **Meshery Server Version:** N/A
- **Meshery Client Version:** N/A
- **meshkit version:** v0.8.56

### To Reproduce

1. Ensure `go.mod` has `github.com/meshery/meshkit v0.8.56`
2. Run `go test ./mesheryctl/internal/cli/root/model/... -v`

### Affected Tests

- `mesheryctl/internal/cli/root/model/build_test.go`
- `mesheryctl/internal/cli/root/model/init_test.go`
- `mesheryctl/internal/cli/root/model/search_test.go`
- Other tests comparing `err.Error()` against golden files

### Proposed Fix

Update all error golden files to include the new meshkit metadata format, or modify test utilities to normalize error strings before comparison.

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
[...]

PR #16776, opened 2025-12-28
[mesheryctl] Update model cmd unit test to get rid of error golden files

shri771:
**Notes for Reviewers**
- Replace golden file-based error comparisons with structured meshkit error assertions
- Added `ExpectedError` field to test structs for direct error comparison using `GetCode`() and `GetLDescription()`
- Removed obsolete error golden files

- This PR Partially addresses [#16735](https://github.com/meshery/meshery/issues/16735)

### Passing unit test
```
ok      github.com/meshery/meshery/mesheryctl/internal/cli/root/model
```


shri771:
@lekaf974

mesheryctl/internal/cli/root/model/build_test.go
```diff
@@ -168,19 +172,23 @@ func TestModelBuild(t *testing.T) {
 			Name:             "model build no params",
 			Args:             []string{"build"},
 			ExpectError:      true,
-			ExpectedResponse: "model.build.error.usage.golden",
+			ExpectedResponse: "",
+			IsOutputGolden:   false,
+			ExpectedError:    ErrModelBuildFromStrings("Usage:\nmesheryctl model build [model-name]\nor\nmesheryctl model build [model-name]/[model-version]\n\nRun 'mesheryctl model build --help' to see detailed help message"),
```
lekaf974: Why not create a constant with the error message in build command and then reuse it in the test, so if we update the error message it will be reflected in test as well.

same for other test cases

shri771: That makes sense. I will implement that.

mesheryctl/internal/cli/root/model/error.go
```diff
@@ -15,6 +15,20 @@ const (
...
+	ErrBuildMultiVersionNotSupported = "\nCommand does not support multiple versions build under one image"
+	
+	ErrInitOneArg = "must provide only one argument: model name"
+	ErrInitUnsupportedFormat = "[ %s ] are the only format supported"
+	ErrInitInvalidVersion = "version must follow a semver format, f.e. v1.2.3"
+	ErrInitFolderExists = "folder %s exists, please specify different model name or version"
+	ErrInitInvalidModelName = "invalid model name: name must match pattern ^[a-z0-9-]+$"
+
+	ErrSearchUsage = "Usage: mesheryctl model search [query-text]\nRun 'mesheryctl model search --help' to see detailed help message"
+	ErrSearchModelName = "Please provide a model name. " + ErrSearchUsage
```
lekaf974: all these var should start in lower case since it is specific to model package

mesheryctl/internal/cli/root/model/error.go
```diff
@@ -15,6 +15,20 @@ const (
...
+	// Error Constants
+	ErrBuildUsage = "Usage:\nmesheryctl model build [model-name]\nor\nmesheryctl model build [model-name]/[model-version]\n\nRun 'mesheryctl model build --help' to see detailed help message"
+	ErrBuildFolderNotFound = "\nfolder %s does not exist"
+	ErrBuildMultiVersionNotSupported = "\nCommand does not support multiple versions build under one image"
+	
+	ErrInitOneArg = "must provide only one argument: model name"
+	ErrInitUnsupportedFormat = "[ %s ] are the only format supported"
+	ErrInitInvalidVersion = "version must follow a semver format, f.e. v1.2.3"
+	ErrInitFolderExists = "folder %s exists, please specify different model name or version"
+	ErrInitInvalidModelName = "invalid model name: name must match pattern ^[a-z0-9-]+$"
```
lekaf974: I think the error message should be more explicit for non developers 
```suggestion
	ErrInitInvalidModelName = "invalid model name: name must contains alphanumeric characters or hyphens"
```

shri771: The `errInitInvalidModelName` constant holds the exact error message returned during validation. Only the test uses this same constant.

PR #16864, opened 2026-01-07
Remove auto commit for fork based PRs in Rego Lint and Test Workflow

CodexRaunak:
**Notes for Reviewers**
- [Failing Step](https://github.com/meshery/meshery/actions/runs/20725791210/job/59500887375)
- Remove auto commit for fork based PRs in `Rego Lint and Test Workflow`  as workflow don't have write permissions
- Remove branch checkout in auto commit 


CodexRaunak:
> Why not give the workflow these permissions instead?

Even if we provide permissions, it will still fail as Github only provides a `read only token` to a workflow in PR targeting the upstream from the fork in case of `pull request` event.
If we use the `pull request target`, then we can get the `write` permissions.
But that's not recommended as CI workflows should only check and fail not correct that failure. 

Ref: https://github.com/actions/labeler/issues/121#issuecomment-1036490254

leecalcote:
@guan404ming thoughts?

.github/workflows/rego-lint-and-test.yml
```diff
@@ -57,7 +59,6 @@ jobs:
           commit_author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
           commit_message: "chore: auto-format Rego files"
           commit_options: "--no-verify"
-          branch: ${{ github.head_ref || github.ref_name }}
```
guan404ming, line 60: I would like to ask that why do we need to remove this?

CodexRaunak: Maybe not in scope of the issue, perhaps we want to rely on the default branch detection of the auto commit action

PR #16883, opened 2026-01-09
fix(CI): update workflow to publish multiple artifacts for each tested k8s versions

### LINKED ISSUE #16882: [Extensions] [Workflows] E2E tests run on multiple K8s versions but only latest version results are published to docs
### Current Behavior

The reusable workflow test_adaptersv2.yaml runs E2E tests against multiple Kubernetes versions using a matrix strategy:
```
strategy:
  matrix:
    k8s_version: ['v1.31.4', 'v1.32.0']

```
but, both test jobs upload artifacts with the same filename (data.json), causing the second upload to overwrite the first
```yaml
- name: UploadResults
  uses: actions/upload-artifact@v6
  with:
    name: ${{ inputs.output_filename }}  # Same name for both versions
    path: ./${{ inputs.output_filename }}
    overwrite: true  # Overwrites previous upload
```

so as a result Only the latest K8s version's test results (v1.32.0) are preserved, while v1.31.4 results are lost.

When adapter repos e.g., meshery-istio download the artifact to publish compatibility docs, they only receive the last version's results

```
UpdateDocsForServicemeshInstall:
  steps:
    - name: DownloadJSON
      uses: actions/download-artifact@v4
      with:
        name: patternfile  # Only contains v1.32.0 results
```
This causes compatibilty documentation to reflect only one version of k8s rather than all versions.

### Desired Behavior

Each K8s version test should upload a uniquely named artifact:

for example,
```
data.json-v1.31.4
data.json-v1.32.0

```

Adapter repos should download all artifacts and publish results for all tested K8s versions in a single compatibility doc

### Screenshots/Logs

Image

Image

### Environment

- Host OS: Mac Linux Windows
- Meshery Extension Version: stable-v
- Platform: Docker Kubernetes

---

- 📚 [Meshery Build and Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 🎨 UI Design For Meshery Extensions [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 📚 [Meshery Extensibility Providers](https://docs.meshery.io/extensibility/providers)
- ⌨️ [Meshery API Docs](https://docs.meshery.io/extensibility/api)

kaizakin:
**Notes for Reviewers**

- This PR fixes #16882

I have updated the testadaptersv2.yaml workflow to include dynamic Kubernetes version extraction from the versions defined in the matrix and upload artifacts with version names appended to it.

Currently the workflow uploads both versioned artifacts and the old data.json named artifacts for backward compatibilty once all the adapters consuming this file has been changed to accept these changes I'll raise a PR to remove this legacy artifact upload.

@leecalcote @aabidsofi19

### DIFF
--- a/.github/workflows/test_adaptersv2.yaml
+++ b/.github/workflows/test_adaptersv2.yaml
@@ -54,7 +54,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        k8s_version: ['v1.31.4', 'v1.32.0']
+        k8s_version: ["v1.31.4", "v1.32.0"]
     runs-on: ubuntu-24.04
     outputs:
       exitstatus: ${{ steps.runresources.outputs.exitstatus }}
@@ -259,7 +259,15 @@ jobs:
           MESHERY_VERSION=$(curl -L -s https://github.com/meshery/meshery/releases/latest | grep -oP 'releases/tag/\K[^"]+' | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+' | head -1)    
           echo $(jq '.metadata."meshery-server-version" |="'$MESHERY_VERSION'"' data.json ) > data.json
           echo "exitstatus=$exitstatus" >> $GITHUB_ENV
-      - name: UploadResults
+      - name: UploadResults (Versioned)
+        uses: actions/upload-artifact@v6
+        with:
+          name: ${{ inputs.output_filename }}-${{ matrix.k8s_version }}
+          path: ./${{ inputs.output_filename }}
+          overwrite: true
+        continue-on-error: true
+
+      - name: UploadResults (Legacy)
         uses: actions/upload-artifact@v6
         with:
           name: ${{ inputs.output_filename }}
@@ -271,4 +279,3 @@ jobs:
           if [ "$exitstatus" -eq 1 ];then
               exit 1
           fi
-


.github/workflows/test_adaptersv2.yaml
```diff
@@ -54,10 +54,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        k8s_version: ['v1.31.4', 'v1.32.0']
+        k8s_version: ["v1.31.4", "v1.32.0"]
     runs-on: ubuntu-24.04
     outputs:
       exitstatus: ${{ steps.runresources.outputs.exitstatus }}
+      k8s_versions: ${{ steps.set-versions.outputs.versions }}
```
lekaf974: I am not sure to understand how we declare this output ?

kaizakin:
```
versions=$(yq -o=json '.jobs.TestAdapter.strategy.matrix.k8s_version' .github/workflows/test_adaptersv2.yaml | jq -c .)
          echo "versions=$versions" >> $GITHUB_OUTPUT
```

the kubernetes versions for this comes from the versions mentioned in the job matrix

image
 
 
 is this what you were asking please let me know @lekaf974

lekaf974: My question is more where this output is reused ?

kaizakin: yep, you're right we don't need this output.

my initial idea was to create this output versions so that it can be used in the adapter repos to generate compatibility docs 
I thought adapter repos might want to dynamically discover which K8s versions were tested, 

and my initial implementation i was doing sth like this
```
echo "versions=[\"v1.31.4\",\"v1.32.0\"]" >> $GITHUB_OUTPUT
```
then later i was using this output to name the data.json-v{versions}


but apparently the data.json file already has the metadata in it.

thanks for the nudge, I should have looked into it more thoroughly


I'll remove this output right away, 
adapter repo workflows can get the k8s version from the metadata and update compatibility docs for each respective version.

kaizakin:
Hey @lekaf974 i've fixed it now, could you please take a look.

lekaf974 (approved):
LGTM

.github/workflows/test_adaptersv2.yaml
```diff
@@ -259,7 +259,15 @@ jobs:
...
-      - name: UploadResults
+      - name: UploadResults (Versioned)
+        uses: actions/upload-artifact@v6
+        with:
+          name: ${{ inputs.output_filename }}-${{ matrix.k8s_version }}
+          path: ./${{ inputs.output_filename }}
+          overwrite: true
+        continue-on-error: true
+
+      - name: UploadResults (Legacy)
```
ianrwhitney, line 270: any reason to keep the legacy one around?

kaizakin: meshery-extensions are not updated to handle this new upload yet, 
I'll raise PR's for all of those repos, once they're merged ill raise a PR remove this legacy upload. @ianrwhitney

lekaf974: I had the same question and I had this in mind, so looks good to keep both until all repo are migrated. Perhaps create an issue to make the cleanup once everything is updated

kaizakin: For sure. @lekaf974 😀

I'll raise an issue for this cleanup once everything is updated.
and i'll make sure the cleanup is performed as well.

PR #17006, opened 2026-01-17
fix: prevent panic on closed channel in event stream handler

Sanchit2662:
## Summary

This PR fixes a **server-crashing race condition** in Meshery’s event streaming pipeline that could trigger a `panic: send on closed channel` when a client disconnects while events are actively being streamed.

The issue occurs in the `/api/events` SSE handler, where multiple goroutines continue writing to a response channel after it has been closed during connection teardown. This results in an unrecovered panic and brings down the entire Meshery server.

The fix ensures **proper goroutine synchronization, safe channel lifecycle management, and graceful shutdown** of all event streaming workers.

---

### 1. Synchronize event stream writers

A `sync.WaitGroup` is introduced in `EventStreamHandler` to ensure **all writer goroutines exit before the response channel is closed**.

```go
var wg sync.WaitGroup

wg.Add(1)
go func() {
    defer wg.Done()
    listenForCoreEvents(ctx, buffer, respChan, log, p)
}()

wg.Wait()
close(respChan)
```

---

### 2. Make sends context-aware

All event writers now guard channel sends with a `select` to respect context cancellation and avoid blocking or panicking after disconnect.

```go
select {
case respChan <- data:
case <-ctx.Done():
    return
}
```

---

### 3. Remove unsafe channel state checks in event broadcasting

The previous `IsClosed` helper was removed and replaced with a non-blocking send pattern, eliminating TOCTOU races and preventing accidental data consumption.

```go
select {
case listener <- event:
default:
}
```

---

## Result

Meshery can now safely handle client disconnects during active event streaming without panicking, leaking goroutines, or dropping events unexpectedly.

---

### DIFF
--- a/mesheryctl/internal/cli/root/model/init.go
+++ b/mesheryctl/internal/cli/root/model/init.go
@@ -1,7 +1,6 @@
 package model
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -16,6 +15,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"golang.org/x/mod/semver"
+	"gopkg.in/yaml.v2"
 )
 
 var initModelCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


Sanchit2662:
Hi @aabidsofi19 , @Siddhi-sahu,
This PR fixes a server-crashing race condition in the event stream handler by synchronizing writer goroutines and making sends context-aware. 
Would appreciate a review when you get a chance.

nexusai97-gif:
I own an API with multiple models and high performance at a cheap price. Message me on Telegram @Nexusai97.

matrixkavi:
Looks like the effort was abandoned midway as the branch is outdated and lots of unit tests are failing. Hence closing this PR.

PR #17008, opened 2026-01-17
[mesheryctl] fix: e2e test meshery server not reachable

lekaf974:

**Notes for Reviewers**

- This PR fixes meheryctl e2e server not reachable issue


.github/workflows/mesheryctl-e2e.yaml
```diff
@@ -62,6 +62,12 @@ jobs:
       - name: Get Date
         id: date
         run: echo "date=date +'%Y-%m-%d-%H.%M.%S'" >> $GITHUB_OUTPUT
+      - name: Build mesheryctl and start Meshery
+        run: |
+          make build && \
+          ./mesheryctl system start --provider ${{ matrix.platform }} --skip-browser --skip-update
```
PragalvaXFREZ: `--provider` expects a provider name (e.g., `Meshery`), not a platform. We should use `-p` instead. This is causing test # 4  to fail - the provider gets set to `kubernetes` instead of `Meshery`.

PragalvaXFREZ (changes requested):
One change needed, I left an inline comment with the suggested fix.

.github/workflows/mesheryctl-e2e.yaml
lekaf974: Good catch, this is a typo

lekaf974:
> No more port forwarding necessary?

Nope minikube tunnel is enough, we can see the current errors are legit and no more related to server not reachable

PR #17073, opened 2026-01-21
[mesheryctl] Refactor endpoint discovery with shared helper function

PragalvaXFREZ:
## Description

This PR enhances `mesheryctl system status` to dynamically discover the actual Meshery service endpoint from Kubernetes, rather than relying solely on the static configuration in `~/.meshery/config.yaml`. This addresses scenarios where users deploy Meshery with **LoadBalancer** or **NodePort** services.

Additionally, this PR refactors duplicate endpoint discovery code into a shared helper function for better maintainability.

### 1. New Shared Helper Function
Created `utils.GetMesheryEndpoint()` in platform.go to centralize endpoint discovery logic, eliminating duplicate code between status.go and dashboard.go.

### 2. Enhanced status.go
- Displays the real-time accessible endpoint (LoadBalancer IP/NodePort)
- Shows warning if discovered endpoint differs from configured endpoint
- Provides actionable guidance to run `mesheryctl system dashboard`

### 3. Refactored dashboard.go
- Uses shared helper function instead of inline `ServiceOptions` setup
- Fixed trailing period in URL log message

### Before
The command displayed the stale/local config value regardless of the actual service status:
Image

### After
Image


mesheryctl/internal/cli/root/system/status.go
```diff
@@ -208,7 +208,29 @@ mesheryctl system status --verbose
 			// Print the data to a table for readability
 			utils.PrintToTable(columnNames, data, nil)
 
-			log.Info("\nMeshery endpoint is " + currCtx.GetEndpoint())
+			// Display endpoint information
+			configuredEndpoint := currCtx.GetEndpoint()
+
+			// Try to get the actual service endpoint from Kubernetes
+			var opts meshkitkube.ServiceOptions
+			opts.Name = "meshery"
```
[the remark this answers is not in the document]
PragalvaXFREZ: Thanks for the suggestion! I noticed that the hardcoded `"meshery"` string is consistent with the existing pattern in `dashboard.go` (lines 141 and 191), which this PR's implementation is based on.

I intentionally kept the same pattern to maintain consistency across the codebase. If we'd like to refactor this to use a constant, it might be better addressed as a separate cleanup PR that updates both `status.go` and `dashboard.go`
Happy to make the change here if you prefer, but wanted to flag the existing pattern first!

mesheryctl/internal/cli/root/system/status.go
lekaf974, line 1: Was this implementation duplication of existing code ?

PragalvaXFREZ: Yes, you're right, the initial implementation duplicated the existing pattern from `dashboard.go`.  I've now refactored the code to address this by creating a shared `utils.GetMesheryEndpoint() ` helper function in `platform.go`.

Please have a look at the updated changes!

mesheryctl/pkg/utils/platform.go
```diff
@@ -966,3 +966,14 @@ func ForceCleanupCluster() error {
 
 	return nil
 }
+
+// GetMesheryEndpoint discovers the Meshery service endpoint from Kubernetes.
```
lekaf974, line 970: this command is also used if platform is docker, did you test it ?

lekaf974:
Did a test locally with current implementation and it looks to wok as expected

```bash
mevrin@aiolos:~/Repos/github/lekaf974/meshery/meshery/mesheryctl$ ./mesheryctl system status
NAME              READY  STATUS   RESTARTS  AGE         
meshery           1/1    Running  2         179h34m41s  
meshery-broker    2/2    Running  4         179h34m39s  
meshery-meshsync  1/1    Running  2         179h34m17s  
meshery-operator  2/2    Running  4         179h34m40s  

Meshery endpoint is http://192.168.49.2:31759

```

mesheryctl/pkg/utils/platform.go
PragalvaXFREZ: This PR doesn't change the logic for when the platform is Docker. The `GetMesheryEndpoint`
function I added is specifically for Kubernetes service discovery, it queries the K8s API to find the Meshery service's external IP/port.

However, due to the existing fallthrough in `status.go` (line 145), when the platform is Docker, the code still falls through to the Kubernetes case and executes GetMesheryEndpoint if a kubeconfig is available.

PragalvaXFREZ:
> This command is really critical, did you do tests on docker/kubernetes and run existing tests to ensure nothing is broken ?

Yes, I tested it on k8s and all the 4 existing tests passed.

lekaf974:
@PragalvaXFREZ just see this PR, it was lost in all other I reviewd since then, ok to merge ?

PragalvaXFREZ:
> @PragalvaXFREZ just see this PR, it was lost in all other I reviewd since then, ok to merge ?

@lekaf974 , I have one edge case to test that's on my mind. If necessary I will update it, else it's good to merge.

lekaf974:
Perfect I think you have permission to do so, let you do it at your convenience

PragalvaXFREZ:
@lekaf974 , Latest commit guards the Kubernetes endpoint discovery behind a currPlatform == "kubernetes" check. Previously, the discovery ran inside the kubernetes switch case which docker falls through into, causing a misleading warning on Docker when --verbose is used.

mesheryctl/internal/cli/root/system/status.go
```diff
@@ -206,7 +206,24 @@ mesheryctl system status --verbose
...
 			utils.PrintToTable(columnNames, data, nil)
 
-			log.Info("\nMeshery endpoint is " + currCtx.GetEndpoint())
+			configuredEndpoint := currCtx.GetEndpoint()
+			displayEndpoint := configuredEndpoint
+
+			if currPlatform == "kubernetes" {
+				endpoint, err := utils.GetMesheryEndpoint(cmd.Context(), client)
+				if err != nil && verboseStatus {
+					log.Warnf("Could not discover Meshery service endpoint: %v", err)
```
lekaf974:
```suggestion
					utils.Log.Warnf("Could not discover Meshery service endpoint: %v", err)
```

mesheryctl/internal/cli/root/system/status.go
```diff
@@ -206,7 +206,24 @@ mesheryctl system status --verbose
...
+				}
+				if err == nil && endpoint.External != nil && endpoint.External.Address != "" {
+					displayEndpoint = fmt.Sprintf("%s://%s:%d", utils.EndpointProtocol, endpoint.External.Address, endpoint.External.Port)
+				}
+			}
+
+			log.Info(fmt.Sprintf("\nMeshery endpoint is %s", displayEndpoint))
```
lekaf974:
```suggestion
			utils.Log.Info(fmt.Sprintf("\nMeshery endpoint is %s", displayEndpoint))
```

mesheryctl/internal/cli/root/system/status.go
```diff
@@ -206,7 +206,24 @@ mesheryctl system status --verbose
...
+			if displayEndpoint != configuredEndpoint {
+				log.Info(fmt.Sprintf("Note: Your configured endpoint (%s) differs from the discovered endpoint.", configuredEndpoint))
```
lekaf974:
```suggestion
				utils.Log.Info(fmt.Sprintf("Note: Your configured endpoint (%s) differs from the discovered endpoint.", configuredEndpoint))
```

mesheryctl/internal/cli/root/system/status.go
```diff
@@ -206,7 +206,24 @@ mesheryctl system status --verbose
...
+				log.Info("Run 'mesheryctl system dashboard' to update your configuration.")
```
lekaf974:
```suggestion
				utils.Log.Info("Run 'mesheryctl system dashboard' to update your configuration.")
```

PragalvaXFREZ:
> Small commment LTGM otherwise

I have addressed it in the latest commit.

lekaf974 (dismissed):
LGTM

PragalvaXFREZ:
@lekaf974, sorry for the extra nudge  😮‍💨. I added a guardrail for minikube users, since this PR could otherwise surface a misleading endpoint for them.

`if err == nil && endpoint.External != nil && endpoint.External.Address != "" && endpoint.External.Address != "localhost" {`

Added this `&& endpoint.External.Address != "localhost"`

There is an alternative approach: reuse the endpoint resolution logic from `mesheryctl system dashboard` without persisting the resolved value to config, so `system status` stays read-only. I did not take that route because it would make `system status` noticeably slower.

Thoughts?

PR #17146, opened 2026-01-25
Fix unsafe os.Stat handling during mesheryctl config initialization

SuryanshGarg04:
### What this PR does

- Guards access to stat.Size() behind err == nil
- Separates decision/action logic for default config creation
- Prevents potential nil pointer dereference during CLI bootstrap

### Why this change is needed

os.Stat may return a non-nil error with invalid FileInfo. Previous logic accessed stat.Size() without guaranteeing safety, which could crash mesheryctl under certain filesystem error conditions.

### How it was tested

- Missing config (rm ~/.meshery/config.yaml)
- Empty config (touch ~/.meshery/config.yaml)
- Permission denied (chmod 000 ~/.meshery/config.yaml)

Additionally verified that stat.Size() is only accessed when err == nil by simulating filesystem permission errors.

Confirmed no panic and graceful failure.

### DIFF
--- a/mesheryctl/internal/cli/root/root_initconfig_test.go
+++ b/mesheryctl/internal/cli/root/root_initconfig_test.go
@@ -79,6 +79,18 @@ func TestInitConfigUseCases(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			viper.Reset()
 
+			// Save global state
+			origMesheryFolder := utils.MesheryFolder
+			origDefaultConfigPath := utils.DefaultConfigPath
+			origCfgFile := cfgFile
+
+			// Restore after each test
+			t.Cleanup(func() {
+				utils.MesheryFolder = origMesheryFolder
+				utils.DefaultConfigPath = origDefaultConfigPath
+				cfgFile = origCfgFile
+			})
+
 			tt.setup(t)
 
 			// ensure default config path branch is exercised


mesheryctl/internal/cli/root/root.go
```diff
@@ -159,22 +159,31 @@ func initConfig() {
...
 			if _, err := os.Stat(utils.MesheryFolder); err != nil {
 				if os.IsNotExist(err) {
 					err = os.MkdirAll(utils.MesheryFolder, 0775)
 					if err != nil {
 						log.Fatal(err)
 					}
+				} else {
+					log.Fatal(err)
 				}
 			}
```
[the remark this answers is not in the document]
SuryanshGarg04: Thanks, updated to use direct os.MkdirAll as suggested.

mesheryctl/internal/cli/root/root.go
```diff
@@ -159,51 +159,45 @@ func initConfig() {
...
-			if _, err := os.Stat(utils.MesheryFolder); err != nil {
-				if os.IsNotExist(err) {
-					err = os.MkdirAll(utils.MesheryFolder, 0775)
-					if err != nil {
-						log.Fatal(err)
-					}
-				}
+		// Only create + mutate config when needed
+		if createDefaultConfig {
+			if err := os.MkdirAll(utils.MesheryFolder, 0775); err != nil {
```
lekaf974:
```suggestion
			if err := os.MkdirAll(utils.MesheryFolder, 0o775); err != nil {
```

mesheryctl/internal/cli/root/root.go
lekaf974, line 1: If you take time here, perhaps you could consider to make decision tree more readable instead of multiple if/else which make code less easy to follow

for example a map[string]bool could make thing prettier

```golang
defaultConfig := map[string]bool{
  "meshconfigEmpty": stat.Size() == 0,
  "meshconfigMissing": os.IsNotExist(err),
  etc.
}
```

Then based on the values diplay proper message and/or take proper action does it make sense ?

SuryanshGarg04: Thanks for the suggestion, I agree the decision flow could be clearer. Would it be okay if I refactor this using a switch for explicit control flow instead of a string-based map? Happy to proceed based on your preference.

lekaf974: you can try with a switch to see

SuryanshGarg04: I’ve pushed the switch based refactor for the initConfig decision flow

SuryanshGarg04:
current root_test.go mainly covers happy-path CLI integration. I’m adding table-driven unit tests for initConfig to explicitly cover missing config, empty file, stat/permission error, and normal load paths. Will push shortly

lekaf974:
> current root_test.go mainly covers happy-path CLI integration. I’m adding table-driven unit tests for initConfig to explicitly cover missing config, empty file, stat/permission error, and normal load paths. Will push shortly

Yep, I think that if you take time to make this part more robust, adding test is also part of the improvement process. Thanks taking care of this.

SuryanshGarg04:
Added table driven unit tests for initConfig covering missing config, empty file, permission/stat error, and normal load paths. Also fixed a regression where --config was being overridden. All tests are passing

mesheryctl/internal/cli/root/root.go
```diff
@@ -159,56 +159,53 @@ func initConfig() {
...
+		case os.IsNotExist(err):
 			log.Printf("Missing Meshery config file.")
+			createDefaultConfig = true
+
+		case err == nil && stat.Size() == 0:
+			log.Println("Empty meshconfig. Please populate it before running a command")
+			createDefaultConfig = true
+
+		case err != nil:
+			log.Printf("Unable to stat config file: %v", err)
```
lekaf974: `Unable to stat config file` I think we can enhance this error message I am not sure non developer understand **Unable to stat** we need to keep in my our end users can not be technical people

mesheryctl/internal/cli/root/root_initconfig_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}{
+		{
+			name: "missing config",
+			setup: func(t *testing.T) {
+				utils.MesheryFolder = filepath.Join(tmp, "meshery-missing")
+				utils.DefaultConfigPath = filepath.Join(utils.MesheryFolder, "config.yaml")
+			},
+		},
+		{
+			name: "empty config",
```
lekaf974:
```suggestion
			name: "given an empty config file when calling initConfig then xxx",
```
We started to introduce Given when then in our test to make explicit the expected behavior based on inputs.

mesheryctl/internal/cli/root/root.go
SuryanshGarg04: Thanks for the suggestion , updated the message to be more user friendly and avoid exposing low-level terminology. Please let me know if this wording works.

alexquincy:

- `docs/_includes/test-report.md`
- `mesheryctl/internal/cli/root/root.go`
- `mesheryctl/internal/cli/root/root_initconfig_test.go`
- `mesheryctl/tests/e2e/001-system/03-system-checks.bats`
- `mesheryctl/tests/e2e/003-design/02-design-delete.bats`
- `mesheryctl/tests/e2e/004-perf/00-perf-apply.bats`

**What to do:**

SuryanshGarg04:
Sorry about that , branch picked up unrelated upstream commits during rebase. Cleaned now.

mesheryctl/internal/cli/root/root_initconfig_test.go
SuryanshGarg04: done

mesheryctl/internal/cli/root/root_initconfig_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+func TestInitConfigUseCases(t *testing.T) {
+	tmp := t.TempDir()
+
+	tests := []struct {
+		name  string
+		setup func(t *testing.T)
+	}{
+		{
+			name: "missing config",
```
lekaf974: please use the Given, When, Then for all tests name

mesheryctl/internal/cli/root/root_initconfig_test.go
```diff
@@ -0,0 +1,91 @@
...
+			setup: func(t *testing.T) {
+				utils.MesheryFolder = filepath.Join(tmp, "meshery-missing")
+				utils.DefaultConfigPath = filepath.Join(utils.MesheryFolder, "config.yaml")
+			},
+		},
+		{
+			name: "given an empty config file when calling initConfig then xxx",
```
lekaf974: what means `xxx`, it was a template could you adjust `then <what is the expected behavior>`.

SuryanshGarg04: xxx was a template placeholder. I’m away for the next 2 days, but will adjust it to reflect the expected behavior once I’m back and push an update. Appreciate the review!

SuryanshGarg04: xxx was a template placeholder. I’m away for the next 2 days, but will adjust it to reflect the expected behavior once I’m back and push an update. Appreciate the review!

mesheryctl/internal/cli/root/root_initconfig_test.go
SuryanshGarg04: Got it, will update all test names to follow Given–When–Then. Thanks!

lekaf974 (approved):

mesheryctl/internal/cli/root/root_initconfig_test.go
```diff
@@ -0,0 +1,91 @@
...
+					t.Fatal(err)
+				}
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			viper.Reset()
+
```
PragalvaXFREZ, line 81:
```suggestion
origMesheryFolder := utils.MesheryFolder
origDefaultConfigPath := utils.DefaultConfigPath
origCfgFile := cfgFile

t.Cleanup(func() {
	utils.MesheryFolder = origMesheryFolder
	utils.DefaultConfigPath = origDefaultConfigPath
	cfgFile = origCfgFile
})
```
Would it be better to clean up the global config state before/after each test case so one test doesn’t leak into the next? Specifically utils.MesheryFolder, utils.DefaultConfigPath, and cfgFile.
Thoughts @lekaf974 @SuryanshGarg04 ?

SuryanshGarg04: Added t.Cleanup to restore global config state between subtests (and to keep things safe for future parallelization). Thanks for the suggestion!

SuryanshGarg04:
Hi, thanks for the approval! 🙏
Just checking in ,please let me know if anything else is needed from my side before the merge

PR #17166, opened 2026-01-27
Fix nil pointer dereference in state machine context extraction

Yashika0724:
### Summary

This PR adds defensive checks when extracting User and SystemID from context.Context in the state machine and related connection action handlers to prevent potential nil pointer dereferences.

Previously, several handlers assumed these values were always present and dereferenced them without validating the type assertion result, which could lead to server panics in certain async or edge cases.

---

### Fix
	•	Added validation for User and SystemID context values before use
	•	Introduced structured errors for missing or invalid context
	•	Applied the same checks consistently across all affected state machine actions

This preserves existing behavior for valid requests while preventing crashes when context is incomplete.

---

### Impact

Prevents Meshery server panics during connection lifecycle operations (connect, disconnect, register, discover) and allows such failures to be handled gracefully with proper error reporting instead of crashing the process.

### DIFF
--- a/server/machines/actions.go
+++ b/server/machines/actions.go
@@ -31,9 +31,16 @@ func (da *DefaultConnectAction) ExecuteOnEntry(ctx context.Context, machineCtx i
 }
 
 func (da *DefaultConnectAction) Execute(ctx context.Context, machineCtx interface{}, data interface{}) (EventType, *events.Event, error) {
-
-	user, _ := ctx.Value(models.UserCtxKey).(*models.User)
-	sysID, _ := ctx.Value(models.SystemIDKey).(*uuid.UUID)
+	user, ok := ctx.Value(models.UserCtxKey).(*models.User)
+	if !ok || user == nil {
+		err := ErrMissingUserContext()
+		return NoOp, nil, err
+	}
+	sysID, ok := ctx.Value(models.SystemIDKey).(*uuid.UUID)
+	if !ok || sysID == nil {
+		err := ErrMissingSystemIDContext()
+		return NoOp, nil, err
+	}
 	userUUID := user.ID
 
 	token, _ := ctx.Value(models.TokenCtxKey).(string)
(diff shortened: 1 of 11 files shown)


Yashika0724:
Hi @PragalvaXFREZ , this PR adds defensive checks to avoid potential nil pointer panics in state machine actions when context values are missing.
Happy to make any changes if needed. 
Thanks for taking a look!

server/machines/kubernetes/connect.go
```diff
@@ -20,10 +20,22 @@ func (ca *ConnectAction) ExecuteOnEntry(ctx context.Context, machineCtx interfac
...
+	}
+	sysID, ok := ctx.Value(models.SystemIDKey).(*uuid.UUID)
+	if !ok || sysID == nil {
+		err := machines.ErrMissingSystemIDContext()
+		return machines.NoOp, nil, err
+	}
 	userUUID := user.ID
-	provider := ctx.Value(models.ProviderCtxKey).(models.Provider)
+	provider, ok := ctx.Value(models.ProviderCtxKey).(models.Provider)
+	if !ok || provider == nil {
```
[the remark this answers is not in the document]
PragalvaXFREZ: @Yashika0724 , I would suggest the same thing.

server/machines/error.go
```diff
@@ -12,6 +12,8 @@ const (
 	ErrInititalizeK8sMachineCode  = "meshery-server-1216"
 	ErrAssetMachineCtxCode        = "meshery-server-1217"
 	ErrInvalidTypeCode            = "meshery-server-1218"
+	ErrMissingUserContextCode     = "meshery-server-1219"
+	ErrMissingSystemIDContextCode = "meshery-server-1220"
```
PragalvaXFREZ, line 16: these error codes are already in use and shouldn't be duplicated here. 

Also:  `ErrMissingProviderContextCode` is missing

server/machines/actions.go
PragalvaXFREZ, line 56:
```suggestion
	_, ok = payload.CredentialSecret["id"]
```

ritzorama:
@Yashika0724, first open an issue instead of drive-by bombing with AI contributions.

PR #17188, opened 2026-01-28
Fix potential data race in Prometheus scan handlers

### LINKED ISSUE #17153: Potential data race in Prometheus scan handlers due to concurrent slice writes
Summary

Multiple Prometheus-related scan handlers launch goroutines that append to a shared slice without synchronization, which can result in a data race under concurrent execution.

Details

In server/handlers/prometheus_handlers.go, several scan handlers iterate over Kubernetes contexts and launch goroutines to perform work concurrently. These goroutines append error messages to a shared []string slice (errs) without any synchronization mechanism.

In Go, appending to a slice from multiple goroutines is not thread-safe and can lead to data races, lost writes, or undefined behavior.

Affected locations
ScanPromGrafanaHandler
ScanPrometheusHandler
ScanGrafanaHandler

Example pattern:

errs := []string{}

for _, mk8scontext := range customK8scontexts {
    go func(mk8scontext models.K8sContext) {
        if err != nil {
            errs = append(errs, err.Error())
        }
    }(mk8scontext)
}

Impact

Under concurrent execution, this can result in:
Data races detected by the Go race detector
Lost or corrupted error messages
Unpredictable handler behavior

dipak0000812:
This PR fixes a potential data race in Prometheus-related scan handlers where multiple goroutines were appending to a shared slice without synchronization.

The change ensures error collection is concurrency-safe while preserving the existing execution flow.

Closes #17153.


dipak0000812:
Hi, I’m picking this PR back up and will address the concurrency issues raised in the review. I’ll update the implementation shortly.

dipak0000812:
Updated this PR based on the earlier review feedback.

### Changes made

* Removed concurrent writes to `http.ResponseWriter`; responses are now encoded only after all goroutines complete
* Replaced the mutex-based error collection with buffered channels for results/errors
* Ensured `wg.Wait()` completes before channels are closed and drained
* Added handling for empty context lists with a successful empty JSON response
* Re-enabled the Prometheus/Grafana scan helpers
* Added unit tests covering empty-context and error paths
* Rebased onto latest `master` and refreshed commit sign-offs

Ready for another review when convenient. Thanks again for the detailed feedback.

Issue #16667, opened 2025-12-16
[mesheryctl] golangci-lint config update


lekaf974:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

### Current Behavior

After upgrade to golangci-lint v2 we had to temporary disabled the following linters:
- errcheck
- staticcheck

We need to re-enable them and fix all related issues

this is one run that listed all issues https://github.com/meshery/meshery/actions/runs/20253290033/job/58149853089

### Desired Behavior

both linters are removed from disable linters and check pass

### Mockups

---

### Contributor Resources

- ⌨️ [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- ⌨️ [Meshkit errors guide](https://docs.meshery.io/project/contributing/contributing-error)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

Sumitsh28:
@lekaf974 Can I work on this issue?

lekaf974:
@Sumitsh28  Do you already have everything setup to work on this issue ?

Sumitsh28:
@lekaf974 Yes, I have everything set up and I’m ready to work on this issue.

lekaf974:
Please don't generate a big Pull request and create small batches that make sense

Deepak-negi11:
@lekaf974 @Sumitsh28  what's the status of  the issue can i help ?

lekaf974:
@Sumitsh28 you did not provide any update, what the current status ?

KhaledSayed04:
Hi @lekaf974 , 
I think @Sumitsh28 isn't interested in solving this issue anymore.
[...]

PR #17247, opened 2026-01-31
style: fix staticcheck capitalization and switch usage (Batch 1)

KhaledSayed04:
**Notes for Reviewers**

-  This PR fixes style violations reported by staticcheck to prepare for enabling strict linting. This is Batch 1 of the cleanup series closing #16667.

**Changes:**
* **`QF1003`:** Refactored `if/else` chains to `switch` statements or applied "Guard Clauses" for cleaner control flow.
* **`ST1005`:** Fixed error string formatting (removed capitalization and trailing punctuation).
* **`QF1004`:** Replaced `strings.Replace(..., -1)` with `strings.ReplaceAll` for better readability.

> Some errorcheck for fmt.Fprintf are fixed in this PR.


Bhumikagarggg:
@KhaledSayed04 Thank you for your contribution! Let's discuss this during the website call tomorrow at 6:30 PM IST | 7 AM CST Add it as an agenda item to the [meeting minutes](https://docs.google.com/document/d/1XczAHXVe2FIWPqiF57ospJ43zw5cZQ7ui8mn39v5EvA/edit?usp=drivesdk), if you would 🙂

KhaledSayed04:
@Bhumikagarggg 
Thanks for the invite!
Since this PR focuses entirely on Backend Go (CLI/Server) staticcheck fixes and doesn't touch the website or docs, I don't think the websites call is the right venue.

server/models/meshery_controllers.go
```diff
@@ -475,7 +475,7 @@ func NewOperatorDeploymentConfig(adapterTracker AdaptersTrackerInterface) contro
 		// if unable to fetch latest release tag, meshkit helm functions handle
 		// this automatically fetch the latest one
 		if err != nil {
-			// mch.log.Error(fmt.Errorf("Couldn't check release tag: %s. Will use latest version", err))
+			// mch.log.Error(fmt.Errorf("couldn't check release tag: %s. Will use latest version", err))
```
lekaf974: this log is commented, just remove it

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -385,7 +387,8 @@ Description: Configures mesheryctl to actively use one one context vs. the anoth
 			const errMsg = `Try running the following to create the context:
 mesheryctl system context create `
 
-			return fmt.Errorf("requested context does not exist \n\n%v%s", errMsg, args[0])
+			cmd.Printf("%s%s\n", errMsg, args[0])
+			return fmt.Errorf("requested context does not exist")
```
lekaf974: not a valid output, it is not following how we are raising error in other commands. should return a meshkit error

PragalvaXFREZ:
I just checked the failing CI run and confirmed the failure is not introduced by this PR.The root cause appears unrelated, and I’ve opened a separate fix to unblock CI. I’ll continue reviewing this PR in parallel.

mesheryctl/internal/cli/root/filter/error.go
```diff
@@ -34,7 +34,7 @@ const (
 )
 
 const (
-	errFilterNameOrIDNotProvided = "Filter Name or ID not provided"
+	errFilterNameOrIDNotProvided = "filter Name or ID not provided"
```
PragalvaXFREZ:
```suggestion
	errFilterNameOrIDNotProvided = "filter name or ID not provided"
```
Nice catch on starting with a lowercase 'f'! To fully match ST1005, we should also lowercase 'name'.

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -363,13 +363,15 @@ var switchContextCmd = &cobra.Command{
...
-	Args: func(_ *cobra.Command, args []string) error {
+	Args: func(cmd *cobra.Command, args []string) error {
 		const errMsg = `Usage: mesheryctl system context switch [context name]
 Example: mesheryctl system context switch k8s-sample
 Description: Configures mesheryctl to actively use one one context vs. the another context`
 
 		if len(args) != 1 {
-			return fmt.Errorf("please provide exactly one context name\n\n%v", errMsg)
+			cmd.PrintErr(errMsg)
+
```
PragalvaXFREZ: Any reason we’re printing the usage/error details separately and returning only a generic error now? This changes the returned error string. Please confirm it doesn’t break any tests that assert on error output.

KhaledSayed04: Good catch. I was attempting to resolve the **ST1005** static analysis warning that error strings should not contain newlines.

However, I agree that changing the return structure risks breaking tests that assert on the exact error output.
I will revert this specific change to preserve the original behavior and ensure backward compatibility.

PragalvaXFREZ:
@lekaf974 I've opened PR https://github.com/meshery/meshery/pull/17374 to unblock the merge of this PR.

server/handlers/grafana_handlers.go
```diff
@@ -58,6 +58,11 @@ func (h *Handler) GrafanaConfigHandler(w http.ResponseWriter, req *http.Request,
 		return
 	}
 
+	if req.Method == http.MethodDelete {
```
lekaf974: I'd rather have this change in another PR instead of so many changes

KhaledSayed04: Done :saluting_face:

lekaf974:
Could we also activate this rule in ci linter config ?

KhaledSayed04:
Done!
@lekaf974

lekaf974 (approved):

PR #17306, opened 2026-02-06
replace logrus - Part 1 - internal/cli/root

### LINKED ISSUE #4705: [mesheryctl] Replace logrus with meshkit logger

#### Current Behavior

mesheryctl is using https://github.com/sirupsen/logrus everywhere

#### Desired Behavior

`Logrus` needs to be replaced with `meshkit` logger

---

#### Contributor Resources

Yorisoft:
**Notes for Reviewers**

- This PR fixes #4705

### DIFF
--- a/mesheryctl/internal/cli/root/root.go
+++ b/mesheryctl/internal/cli/root/root.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"log"
 
 	"github.com/meshery/meshery/mesheryctl/internal/cli/root/adapter"
 	"github.com/meshery/meshery/mesheryctl/internal/cli/root/components"
@@ -32,7 +33,7 @@ import (
 	"github.com/meshery/meshery/mesheryctl/internal/cli/root/registry"
 	"github.com/meshery/meshery/mesheryctl/internal/cli/root/system"
 	"github.com/meshery/meshery/mesheryctl/pkg/utils"
-	log "github.com/sirupsen/logrus"
+	logrus "github.com/sirupsen/logrus"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -90,7 +91,7 @@ func Execute() error {
 func init() {
 	err := utils.SetFileLocation()
 	if err != nil {
-		utils.Log.Fatal(err)
+		log.Fatal(err)
 	}
 
 	cobra.OnInitialize(setupLogger)
@@ -158,7 +159,6 @@ func initConfig() {
 		// Otherwise, use the default `config.yaml` config file
 	} else {
 		stat, err := os.Stat(utils.DefaultConfigPath)
-
 		createDefaultConfig := false
 
 		switch {
@@ -210,10 +210,10 @@ func initConfig() {
 
 // setVerbose sets the log level to debug if the -v flag is set
 func setVerbose() {
-	utils.Log.SetLevel(log.InfoLevel)
+	utils.Log.SetLevel(logrus.InfoLevel)
 
 	if verbose {
-		utils.Log.SetLevel(log.DebugLevel)
+		utils.Log.SetLevel(logrus.DebugLevel)
 	}
 }
 


mesheryctl/internal/cli/root/root_test.go
```diff
@@ -76,8 +75,8 @@ func TestRootCmdIntegration(t *testing.T) {
 			viper.Reset()
 
 			b := bytes.NewBufferString("")
-			logrus.SetOutput(b)
-			utils.SetupLogrusFormatter()
+			utils.Log = utils.SetupMeshkitLogger("mesheryctl", verbose, b)
+			utils.LogError = utils.SetupMeshkitLogger("mesheryctl-error", verbose, b)
```
Yorisoft: The intention here is to simulate the two loggers we have. Open to feedback on whether or not to include both.

lekaf974: I don't understand why we have 2 log handlers ?

leecalcote:
Thank you for breaking this into bite-sized parts, @Yorisoft.

mesheryctl/internal/cli/root/root_test.go
lekaf974: I mean usually loggers is initiate once no matter what type of log (info, debug, error, etc..)

Yorisoft: Okay, thank for clarifying - that makes sense to me 
I'll leave it to one logger, utils.Log.

lekaf974: @Yorisoft any update ?

Yorisoft: Updated last night. This should be resolved. I'll let you make the decision.

Yorisoft:
@lekaf974 @PragalvaXFREZ @leecalcote 
Question to anyone in Meshery about best practices:  
Do I resolve threads/feedback as the PR contributor or do I let the reviewers resolve the thread?

PragalvaXFREZ:
> @lekaf974 @PragalvaXFREZ @leecalcote
> Question to anyone in Meshery about best practices:
> Do I resolve threads/feedback as the PR contributor or do I let the reviewers resolve the thread?

Typically the contributor resolves the thread after implementing the feedback and verifying it.
If it still needs reviewer confirmation or a decision, it should stay open.

lekaf974:
@Yorisoft  we trust contributors so you can resolve them once fixed.

mesheryctl/internal/cli/root/root.go
```diff
@@ -91,11 +90,11 @@ func Execute() error {
 func init() {
 	err := utils.SetFileLocation()
 	if err != nil {
-		log.Fatal(err)
+		utils.LogError.Fatal(err)
```
lekaf974: why using utils.LogError ? this mean we have 2 logrus handlers ?

lekaf974:
```suggestion
		utils.Log.Fatal(err)
```

mesheryctl/internal/cli/root/root_test.go
```diff
@@ -76,8 +75,7 @@ func TestRootCmdIntegration(t *testing.T) {
 			viper.Reset()
 
 			b := bytes.NewBufferString("")
-			logrus.SetOutput(b)
-			utils.SetupLogrusFormatter()
+			utils.Log = utils.SetupMeshkitLogger("mesheryctl", verbose, b)
```
lekaf974: can be replaced by 

```golang
b := utils.SetupMeshkitLoggerTesting(t, false)
```

mesheryctl/internal/cli/root/root.go
Yorisoft: I assumed the earlier feedback was for test specifically.  
  
If it applies to the entire project my question would be: what do we do about the `LogError` in `utils/helper.go`?  
It defines both of these loggers. They were pointed out when the task was brought up.

lekaf974: I have seen utils.LogError  but if you look it is only use in one place in the project, I had took a look as well and will do some changes since it does not make sense to have 2 log handlers IMO.

From what I have seen it was introduced to have log in os.Stderr but logrus support this by default, it needs a small change to have this initialezed properly

Yorisoft: Okay thanks for clarifying. I assume that change will be different scope than this PR - I'll resolve this thread and I think it's ready.

Yorisoft:
@lekaf974 @PragalvaXFREZ @aabidsofi19 
Does this look good to merge?

mesheryctl/internal/cli/root/root.go
```diff
@@ -91,11 +90,11 @@ func Execute() error {
 func init() {
 	err := utils.SetFileLocation()
 	if err != nil {
-		log.Fatal(err)
+		utils.Log.Fatal(err)
```
PragalvaXFREZ: I'm a bit concerned about the initialization order in `init()`. According to my understanding, `utils.Log` gets set up later via `cobra.OnInitialize`, won't it still be `nil` when `SetFileLocation` runs? If that returns an error, calling `utils.Log.Fatal` here looks like it would cause a panic before the logger is ready.

lekaf974: I also had the same question

Yorisoft: What do yall think about using Go's standard log package for fatal errors during `init()` ?

lekaf974: fmt.print + exit(1) would have done the trick as well

PR #17336, opened 2026-02-07
[mesheryctl] fixed failure test and improved test

DharunMR:
**Notes for Reviewers**

-Fixed the failing test in delete sub-command
-Fixed skipping tests
-When running list, it required manual intervention (pressing Enter) this was avoided by introducing the --page flag
-Added missing tests

### DIFF
--- a/mesheryctl/tests/e2e/007-connection/04-connection-delete.bats
+++ b/mesheryctl/tests/e2e/007-connection/04-connection-delete.bats
@@ -47,6 +47,4 @@ teardown_file() {
     run $MESHERYCTL_BIN connection delete "$CONNECTION_ID"
     assert_success
     assert_output --partial "deleted"
-
-    run $MESHERYCTL_BIN connection create --type minikube
 }


mesheryctl/tests/e2e/007-connection/02-connection-list.bats
```diff
@@ -3,10 +3,42 @@ setup() {
...
+    mkdir -p "$TESTDATA_DIR"
+}
+
+@test "given all requirements met when running mesheryctl connection list --status status then the conections of specified status is displayed" {
+    run $MESHERYCTL_BIN connection list --status connected --page 1
+
+    assert_success
+    assert_output --partial "Total number of connection"
+
+    CONNECTION_ID=$(
```
lekaf974: don't think it is a good idea to take connection id from a list, since we are creating a connection id and the connection view allow looking by name perhaps we can do this:

in create test create a connection with name: e2e-test-<datetime>
in view test view by name with name: e2e-test-<datetime> capture the connection id with yq then test view by connection id

DharunMR: how do i create connection with name???

DharunMR: @lekaf974 I verified this locally and everything works as expected, but the CI keeps failing. Could you please help me understand what might be causing the failure in CI?

mesheryctl/tests/e2e/007-connection/01-connection-create.bats
```diff
@@ -43,6 +44,13 @@ setup() {
     assert_output --partial "Token set in context minikube"
 
     #Extract connection ID if present and store it in temp dir
-    
-
+    CONNECTION_ID=$(
```
lekaf974, line 47: why re-run connection create ? it should already be in current ouput

DharunMR: fixed it

mesheryctl/tests/e2e/007-connection/04-connection-delete.bats
```diff
@@ -31,4 +47,6 @@ teardown_file() {
     run $MESHERYCTL_BIN connection delete "$CONNECTION_ID"
     assert_success
     assert_output --partial "deleted"
+
+    run $MESHERYCTL_BIN connection create --type minikube
```
lekaf974: not sure to understand about this one ?

mesheryctl/tests/e2e/007-connection/02-connection-list.bats
```diff
@@ -5,8 +5,29 @@ setup() {
...
+
+@test "given all requirements met when running mesheryctl connection list --kind kind-name then the connections of specified kind is displayed" {
+    run $MESHERYCTL_BIN connection list --kind kubernetes --page 1
+
+    assert_success
+    assert_output --partial "Total number of connection"
+}
+
+@test "given all requirements met when running mesheryctl connection list --pagesize size then the conections of specified size is displayed" {
+    run $MESHERYCTL_BIN connection list --pagesize 10000
```
lekaf974: if we have more than 10000 connections it will fail use --page-size 1 and --page 1 for now

mesheryctl/tests/e2e/007-connection/04-connection-delete.bats
DharunMR: this test will delete existing minikube connection in my machine , by running create --type minikube it will again reconnect with same id.

lekaf974: I don't really understand your point. The goal of using the same id from the test is to be sure to remove this connection id since it was created for the test purpose, also at the end it will run in our ci against our server.

DharunMR: i thought it will return error in ci if i delete connection ,i will remove it now👍🏻

mesheryctl/tests/e2e/007-connection/02-connection-list.bats
lekaf974: Can you share me what ci test is failing ? Not sure to understand since e2e are only triggered on PR merged.

DharunMR: earlier there was a kanvas screenshot check failing it was saying something like cypress error ,then now it passed

lekaf974: kanvas error is not related to your changes

PR #17322, opened 2026-02-06
Fix None provider redirecting to blank login screen

PragalvaXFREZ:
## Summary                                                                                                                                                                                  
- When `Provider=None` is selected, the server redirected to `/user/login`, but the local provider's `InitiateLogin` didn't redirect anywhere, resulting in a blank page                    
- `ProviderUIHandler` also didn't handle the `None` case when `PROVIDER=None` env var is set, redirecting to `/user/login?` instead of `/`                                                  
                  
## Changes                                                                                                                                                                                  
- **`server/handlers/provider_handler.go`**: Added check in `ProviderUIHandler` to redirect to `/` when `h.Provider == "None"`                                                              
- **`server/models/default_local_provider.go`**: Updated `InitiateLogin` to redirect to `/` instead of being a no-op

**Notes for Reviewers**

- This PR fixes #17321 

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

## Testing:
- Verified `Provider=None` redirects to `/` instead of blank login screen
- Verified provider selection UI still works when no `PROVIDER` env var is set (200 OK serving provider-ui)
- Tested on fresh install / incognito to reproduce original bug

### Provider selection UI (no provider set)
[Screencast from 2026-02-06 23-06-59.webm](https://github.com/user-attachments/assets/7429d5b1-6cb6-49d3-a7ed-6658842c8624)

PR #17342, opened 2026-02-08
[Server] Fix None provider blank login by routing through /user/login

### LINKED ISSUE #17321: Provider=None redirects to blank login screen instead of dashboard
### Current Behavior
When selecting the "None" provider (or when PROVIDER=None is set), the server redirects to /user/login, which renders a blank page. The local provider's InitiateLogin doesn't perform any redirect, resulting in an empty response. On a subsequent page refresh, the dashboard loads because the session cookie is already set.

### Expected Behavior
Selecting the "None" provider should redirect users directly to the dashboard (/) without passing through the blank login screen.

#### Current Behavior:

https://github.com/user-attachments/assets/c3ffebe6-5702-44b1-b7df-e77f2f4f676d

#### Expected Behavior:

https://github.com/user-attachments/assets/014161f4-ffd1-4c2c-b76d-bca7f81d889b

### Environment
  - Host OS: Linux
  - Platform: Kubernetes
  - Meshery Server Version: v0.8.202

---

leecalcote:
## Description

Fixes #17321 – When `Provider=None` is selected (or enforced via `PROVIDER` env var), users are redirected to a blank login page because `DefaultLocalProvider.InitiateLogin` was a no-op (the method body was commented out). This PR fixes the root cause by making `InitiateLogin` complete the login flow with a redirect.

### Root Cause

`DefaultLocalProvider.InitiateLogin` (in `server/models/default_local_provider.go`) had its body commented out (`// l.issueSession(w, r, fromMiddleWare)`). When `LoginHandler` called `provider.InitiateLogin(w, r, false)`, no HTTP response was written, resulting in a blank page.

### Fix

**`server/models/default_local_provider.go`**
- `InitiateLogin` now redirects to `/` (or to the deep-link target from the `ref` query param) when called from the `/user/login` route (`fromMiddleWare=false`)
- When called from `AuthMiddleware` (`fromMiddleWare=true`), it remains a no-op to avoid sending 302 responses to REST API clients
- Uses `core.DecodeRefURL()` for centralized base64 ref decoding
- Adds `isSafeRedirect()` validation to prevent open redirect attacks via the `ref` parameter (rejects absolute URLs, protocol-relative URLs, and non-empty scheme/host)

**`server/handlers/provider_handler.go`**
- Removed the `provider == "None"` special-case bypass from `ProviderHandler` — the None provider now flows through `/user/login` like all other providers
- Fixed `ProviderUIHandler` to avoid appending a trailing `?` when `RawQuery` is empty

**`server/handlers/common_handlers.go`**
- Fixed swagger annotation from `/api/user/login` to `/user/login` to match the actual route

### Login Flow (After Fix)

1. User selects "None" provider → `ProviderHandler` sets cookie → redirects to `/user/login?provider=None`
2. `/user/login` → `ProviderMiddleware` resolves provider → `LoginHandler` → `DefaultLocalProvider.InitiateLogin` → **302 to `/`** (dashboard)

When `PROVIDER=None` env var is set:
1. User visits `/provider` → `ProviderUIHandler` sets cookie → redirects to `/user/login`
2. Same flow as above → dashboard

### Tests

Added `server/handlers/provider_handler_test.go` with 7 tests:
- `TestProviderHandler_NoneRedirectsThroughLogin` — verifies cookie + redirect to `/user/login`
- `TestProviderUIHandler_NoneEnvVarRedirectsThroughLogin` — verifies env var flow (no trailing `?`)
- `TestLoginHandler_NoneProviderRedirectsHome` — verifies redirect to `/`
- `TestLoginHandler_NoneProviderDeepLink` — verifies `ref` param deep-link support
- `TestLoginHandler_NoneProviderRejectsOpenRedirect` — table-driven test for absolute URL, protocol-relative, and javascript scheme
- `TestLoginHandler_NoneProviderFromMiddleware` — verifies no redirect when `fromMiddleWare=true`

### Approach Comparison with #17322

Both PRs fix `InitiateLogin` to redirect. This PR differs by:
- **Removing** the `ProviderHandler` bypass instead of adding more bypasses
- Preserving the standard login flow (`/user/login` route) for all providers
- Adding open redirect protection for the `ref` parameter
- Using centralized `core.DecodeRefURL()` instead of inline base64 decoding
- Adding comprehensive test coverage

### DIFF
--- a/docs/pages/extensibility/providers.md
+++ b/docs/pages/extensibility/providers.md
@@ -60,6 +60,16 @@ Name: **"Layer5"** (default)
 - Save environment setup.
 - Retrieve performance test results.
 - Events are stored locally and can be published to remote provider. [Read more about server events](https://docs.meshery.io/project/contributing/contributing-server-events)
+
+#### Example remote provider
+
+Name: **"Acme"** (default)
+
+- Enforces user authentication.
+- Long-term term persistence.
+- Save environment setup.
+- Retrieve performance test results.
+- Events are stored locally and can be published to remote provider. [Read more about server events](https://docs.meshery.io/project/contributing/contributing-server-events)
 - Free to use.
 
 ### Local Provider


server/models/default_local_provider.go
```diff
@@ -140,9 +140,17 @@ func (l *DefaultLocalProvider) GetProviderCapabilities(w http.ResponseWriter, _
...
+// there instead of / so that deep-links survive provider selection.
+func (l *DefaultLocalProvider) InitiateLogin(w http.ResponseWriter, r *http.Request, _ bool) {
+	redirectURL := "/"
+	if ref := r.URL.Query().Get("ref"); ref != "" {
+		if decoded, err := base64.RawURLEncoding.DecodeString(ref); err == nil && len(decoded) > 0 {
+			redirectURL = string(decoded)
+		}
+	}
+	http.Redirect(w, r, redirectURL, http.StatusFound)
 }
```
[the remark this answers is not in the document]
PragalvaXFREZ: This is a valid concern. I've checked it locally the ref param is vulnerable to open redirects. 

Image

server/handlers/provider_handler_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+func TestLoginHandler_NoneProviderDeepLink(t *testing.T) {
+	local := &models.DefaultLocalProvider{}
+	local.Initialize()
+
+	providers := map[string]models.Provider{local.Name(): local}
+	h := newTestHandler(t, providers, "")
+
+	// ref is base64-raw-url-encoded "/extension/meshmap"
+	req := httptest.NewRequest(http.MethodGet, "/user/login?provider=None&ref=L2V4dGVuc2lvbi9tZXNobWFw", nil)
```
PragalvaXFREZ, line 125: Could we add a case here to verify that an external URL passed as ref doesn't result in an off-site redirect?

server/handlers/provider_handler.go
PragalvaXFREZ: Is there a reason we're not including h.Provider in the redirect query string here? I noticed an infinite redirect loop between /provider and /user/login? when testing without cookies (e.g., curl without a cookie jar). It works fine in browsers since cookies are sent automatically, but appending the provider (e.g., /user/login?provider=None) would make the flow less dependent on cookie state.

leecalcote:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/meshery/meshery/pull/17342#pullrequestreview-3768890668)

server/models/default_local_provider.go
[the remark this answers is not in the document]
leecalcote: Thanks for this, @PragalvaXFREZ. This feedback is being addressed in #17350

server/handlers/provider_handler.go
leecalcote: Thanks for bringing this up... 🤔

server/handlers/provider_handler_test.go
leecalcote: Yes, I'll let you add.

leecalcote: Implemented.

server/handlers/provider_handler_test.go
```diff
@@ -0,0 +1,137 @@
...
+	t.Helper()
+	log, _ := logger.New("test", logger.Options{Format: logger.SyslogLogFormat})
+	return &Handler{
+		config: &models.HandlerConfig{
+			Providers:          providers,
+			ProviderCookieName: "meshery-provider",
+		},
+		Provider: providerOverride,
+		log:      log,
+	}
```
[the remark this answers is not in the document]
leecalcote: Incorporated.

server/handlers/provider_handler_test.go
```diff
@@ -0,0 +1,137 @@
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusFound {
+		t.Fatalf("expected 302 Found, got %d", resp.StatusCode)
+	}
+
+	loc := resp.Header.Get("Location")
+	if loc != "/user/login?" {
+		t.Errorf("expected redirect to /user/login?, got %s", loc)
+	}
```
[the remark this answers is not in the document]
leecalcote: Incorporated.

docs/pages/extensibility/providers.md
hortison, line 73:
```suggestion

#### Example remote provider

Name: **"Acme"** (default)

- Enforces user authentication.
- Long-term term persistence.
- Save environment setup.
- Retrieve performance test results.
- Events are stored locally and can be published to remote provider. [Read more about server events](https://docs.meshery.io/project/contributing/contributing-server-events)
- Free to use.
```

PR #17345, opened 2026-02-08
[Server] Policies: Fix identified relationships not applied in rego evaluation

guan404ming:
**Notes for Reviewers**

- This PR fixes design_file_with_identified_rels was re-applying validation_actions instead of new_identified_rels_actions, so identified relationships were never applied.
 - Fix arguments to apply_all_actions_to_design to use design_file_with_validated_rels and new_identified_rels_actions

### DIFF
--- a/server/meshmodel/meshery-core/0.7.2/v1.0.0/policies/evaluation.rego
+++ b/server/meshmodel/meshery-core/0.7.2/v1.0.0/policies/evaluation.rego
@@ -233,8 +233,8 @@ evaluate := eval_results if {
 	})
 
 	design_file_with_identified_rels := actions.apply_all_actions_to_design(
-		design_file_to_evaluate,
-		validation_actions,
+		design_file_with_validated_rels,
+		new_identified_rels_actions,
 	)
 
 	print("New identified rels", count(new_identified_rels_actions))
--- a/server/meshmodel/meshery-core/0.7.2/v1.0.0/policies/evaluation.rego
+++ b/server/meshmodel/meshery-core/0.7.2/v1.0.0/policies/evaluation.rego
@@ -241,23 +241,18 @@ evaluate := eval_results if {
 
 	#3. Actions
 
-	design_file_to_apply_actions := actions.apply_all_actions_to_design(
-		design_file_with_identified_rels,
-		new_identified_rels_actions,
-	)
-
-	print("All relationships", count(design_file_to_apply_actions.relationships))
+	print("All relationships", count(design_file_with_identified_rels.relationships))
 
 	actions_to_apply := union({actions |
 		some identifier in relationship_policy_identifiers
-		actions := eval.generate_actions_to_apply_on_design(design_file_to_apply_actions, identifier)
+		actions := eval.generate_actions_to_apply_on_design(design_file_with_identified_rels, identifier)
 		print("action for", identifier, count(actions))
 	})
 
 	print("All Actions", count(actions_to_apply))
 
 	actions_response := trace_from_actions(actions_to_apply)
-	design_to_return := actions.apply_all_actions_to_design(design_file_to_apply_actions, actions_to_apply)
+	design_to_return := actions.apply_all_actions_to_design(design_file_with_identified_rels, actions_to_apply)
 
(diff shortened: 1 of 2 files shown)


hortison:
@leecalcote @aabidsofi19 @guan404ming @reaper8055, this is a big deal... we need to test this thoroughly. 

@guan404ming do you have before and after example(s), by chance?

guan404ming:
Before (master):                                                                                                                                                                  
step1_result = validate(original_design)        # removes stale rels                                                                                                              
step2_result = validate(original_design)        # oops, same thing again (copy-paste)                                                                                             
step3_result = add_new_rels(step2_result)        # new rels added here instead
                                                                                                                                                                                  
After (fix):                                                                
step1_result = validate(original_design)        # removes stale rels
step2_result = add_new_rels(step1_result)        # new rels added at the right step
step3_result = add_new_rels(step2_result)        # no-op, already there

Same final result, but step 2 now does what it's supposed to instead of redundantly repeating step 1. On master, the new rels never get added at step 2, but step 3 adds them anyway. On the branch, step 2 adds them correctly, and step 3 is a no-op (set union, already there). So the bug is real (step 2 does the wrong thing), but the final output is the same because step 3 happens to do the same work. It's kind of effectively dead code on master I think.

guan404ming:
Hi I found that I could edit my pr title. But I think to add [Server] as prefix would be better align to the convention here. Could someone help with me, thanks!

lekaf974:
Perhaps we can start to implement some unit test here so we will be more confident on changes ?

guan404ming:
> Perhaps we can start to implement some unit test here so we will be more confident on changes ?

Thanks for your suggestion! I think there are already a number of unit tests covering this area. Since this change doesn’t alter or add any new behavior like I mentioned above, the acceptance criteria should simply be passing the existing tests based on my understanding. What do you think @lekaf974?

leecalcote:
> Before (master): step1_result = validate(original_design) # removes stale rels step2_result = validate(original_design) # oops, same thing again (copy-paste) step3_result = add_new_rels(step2_result) # new rels added here instead
> 
> After (fix): step1_result = validate(original_design) # removes stale rels step2_result = add_new_rels(step1_result) # new rels added at the right step step3_result = add_new_rels(step2_result) # no-op, already there
> 
> Same final result, but step 2 now does what it's supposed to instead of redundantly repeating step 1. On master, the new rels never get added at step 2, but step 3 adds them anyway. On the branch, step 2 adds them correctly, and step 3 is a no-op (set union, already there). So the bug is real (step 2 does the wrong thing), but the final output is the same because step 3 happens to do the same work. It's kind of effectively dead code on master I think.

@guan404ming, "before and after" in context of a design being evaluated.

guan404ming:
I see your point. We should validate the state transition at each step, not just the final output. I'll add a unit test to specifically verify that step 2 is now correctly applying the relationships.

server/meshmodel/meshery-core/0.7.2/v1.0.0/policies/evaluation.rego
```diff
@@ -233,8 +233,8 @@ evaluate := eval_results if {
 	})
 
 	design_file_with_identified_rels := actions.apply_all_actions_to_design(
-		design_file_to_evaluate,
-		validation_actions,
+		design_file_with_validated_rels,
```
aabidsofi19, line 236: @guan404ming  actually this merge seems  completely redundant as the identified relationships are already merged on line 244 .  we probably dont need the intermediated design_file_with_identified_rels here .

aabidsofi19 (changes requested):

server/meshmodel/meshery-core/0.7.2/v1.0.0/policies/evaluation.rego
guan404ming: Good catch, you're right. The new_identified_rels_actions were being applied twice in sequence. Removed the redundant intermediate step and now use design_file_with_identified_rels directly for the remaining pipeline.

guan404ming:
Hi @aabidsofi19 could you help take another look? Thanks!

PR #17376, opened 2026-02-09
[Server] Fix rego evaluation depth-exceeded check being unreachable

guan404ming:
**Notes for Reviewers**

- This PR fixes the depth-exceeded error in EvaluateDesign was dead code, causing non-converging evaluations to silently return incomplete results as success.

### DIFF
--- a/server/handlers/policy_relationship_handler.go
+++ b/server/handlers/policy_relationship_handler.go
@@ -200,13 +200,16 @@ func (h *Handler) EvaluateDesign(
 		lastEvaluationResponse.Design = evaluationResponse.Design
 		lastEvaluationResponse.Actions = append(lastEvaluationResponse.Actions, evaluationResponse.Actions...)
 
-		if evalIterations == i+1 || doesntNeedReeval(evaluationResponse) {
+		if doesntNeedReeval(evaluationResponse) {
 			h.log.Info("Evaluation completed in iteration ", i+1)
 			break
 		}
 		if i == (MAX_RE_EVALUATION_DEPTH - 1) {
 			h.log.Info("Evaluation depth exceeded")
-			return lastEvaluationResponse, fmt.Errorf("Evaluation depth exceeded")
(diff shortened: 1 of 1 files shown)


RanitMukherjee (approved):
LGTM! 🚀

server/handlers/policy_relationship_handler.go
```diff
@@ -200,13 +200,13 @@ func (h *Handler) EvaluateDesign(
 		lastEvaluationResponse.Design = evaluationResponse.Design
 		lastEvaluationResponse.Actions = append(lastEvaluationResponse.Actions, evaluationResponse.Actions...)
 
-		if evalIterations == i+1 || doesntNeedReeval(evaluationResponse) {
```
aabidsofi19, line 210: @guan404ming  i dont think this code is unreachable . evalIterations parameter is different from MAX_RE_EVALUATION_DEPTH  , the former controls the number of iteration to evaulation which is intented in certain situations and not an error  while the later is a fail safe to prevent infinite recursion and returns an error if that happens

aabidsofi19 (changes requested):

aabidsofi19:
@guan404ming  i dont think these changes are needed the evaliterations parameter and MAX_RE_EVALUATION_DEPTH control different situations

server/handlers/policy_relationship_handler.go
guan404ming: You're right that evalIterations and MAX_RE_EVALUATION_DEPTH serve different purposes. But the depth check is unreachable in the current code because the only caller passes evalIterations = MAX_RE_EVALUATION_DEPTH (line 385), so evalIterations == i+1 always triggers first on the same iteration, and the depth error is never returned. I'll fix the ordering to keep them as separate checks: convergence (break), MAX_RE_EVALUATION_DEPTH (error), then evalIterations (break).

guan404ming:
Make sense to me, thanks.

PR #17411, opened 2026-02-11
style: fix staticcheck issues (QF1003, ST1019, QF1006, SA1019, QF1001)

KhaledSayed04:
**Notes for Reviewers**

- This PR fixes style violations reported by staticcheck to prepare for enabling strict linting. This is Batch 2 of the cleanup series closing #16667

**Key Changes:**
- **Deprecation Fix (SA1019):** Updated `runner.Run` to `runner.RunTests` in OPA tests. *Note: Passing `nil` for the transaction argument preserves the original behavior of the deprecated function.*
- **Control Flow Refactor:**
  - **Switch Statements (QF1003):** Converted peer `if/else` chains to `switch` statements for better readability.
  - **Guard Clauses:** Replaced `if/else` block ending with `return` with **Guard Clauses** (early return).
- **Cleanup:** Resolved duplicate imports (`ST1019`), simplified boolean logic (`QF1001`), and lifted conditions into loops (`QF1006`).

**Fixed Checks**
* [x] **QF1003**
* [x] **SA1019**
* [x] **ST1019**
* [x] **QF1006**
* [x] **QF1001**


leecalcote:
@guan404ming 👀

leecalcote:
@YASHMAHAKAL

YASHMAHAKAL:
@KhaledSayed04 @leecalcote, i checked it, looks good to me : )

.github/.golangci.yml
```diff
@@ -8,6 +8,11 @@ linters:
       checks: 
         - ST1005
         - QF1004
+        - QF1003
+        - QF1006
+        - ST1019
+        - SA1019
+        - QF1001
```
YASHMAHAKAL, line 15: working correctly

guan404ming (approved):
Looks nice. Thanks for the contribution!

server/handlers/grafana_handlers.go
```diff
@@ -52,13 +52,15 @@ func (h *Handler) GrafanaConfigHandler(w http.ResponseWriter, req *http.Request,
 	eventBuilder := events.NewEvent().ActedUpon(userUUID).WithCategory("connection").WithAction("update").FromSystem(*sysID).FromUser(userUUID).WithDescription("Failed to interact with the connection.")
 
 	token, _ := req.Context().Value(models.TokenCtxKey).(string)
-	if req.Method == http.MethodGet {
+	switch req.Method {
+	case http.MethodGet:
 		req = mux.SetURLVars(req, map[string]string{"connectionKind": "grafana"})
 		h.GetConnectionsByKind(w, req, prefObj, user, p)
```
lekaf974, line 58:
```suggestion
		h.GetConnectionsByKind(w, req, prefObj, user, p)
		return
```

server/handlers/grafana_handlers.go
```diff
@@ -52,13 +52,15 @@ func (h *Handler) GrafanaConfigHandler(w http.ResponseWriter, req *http.Request,
...
-		return
-	}
 
-	if req.Method == http.MethodPost {
+	case http.MethodDelete:
+		http.Error(w, "API is deprecated, please use connections API", http.StatusGone)
```
lekaf974, line 62:
```suggestion
		http.Error(w, "API is deprecated, please use connections API", http.StatusGone)
		return
```

server/handlers/grafana_handlers.go
```diff
@@ -126,11 +128,7 @@ func (h *Handler) GrafanaConfigHandler(w http.ResponseWriter, req *http.Request,
 		h.log.Debug(fmt.Sprintf("connection to grafana @ %s succeeded", grafanaURL))
 
 		_ = json.NewEncoder(w).Encode(connection)
```
lekaf974, line 132:
```suggestion
		_ = json.NewEncoder(w).Encode(connection)
		return
```

server/models/validators.go
```diff
@@ -24,7 +24,7 @@ func SMPPerformanceTestConfigValidator(perfTest *SMP.PerformanceTestConfig) erro
 		if testClient.Protocol.String() == "" {
 			return ErrProtocol
 		}
-		if !(testClient.LoadGenerator == Wrk2LG.Name() || testClient.LoadGenerator == FortioLG.Name()) {
+		if testClient.LoadGenerator != Wrk2LG.Name() && testClient.LoadGenerator != FortioLG.Name() {
```
lekaf974: why modify this logic ?

if it is really what we want add a comment to explain the wanted behavior.

before it was on case of other and now is both case so we need to explicitly explain the correct behavior or use variables to make the condition check more explicit without comments

lekaf974 (changes requested):
please check commects

server/handlers/grafana_handlers.go
KhaledSayed04: I think this `return` doesn't affect as this switch is the end of the function anyway. (The same for the next 2 `return` suggestions in the same switch)

server/models/validators.go
KhaledSayed04: The newline is the same logic but simplified. (They're identical in behaviour)
The old one gave `QF1001` issue in staticcheck.

KhaledSayed04: > or use variables to make the condition check more explicit without comments

Just to be sure, is this the current required change?

KhaledSayed04: @lekaf974

lekaf974: ok, I was just a little confused but yep logic keep the same, just that test boolean based on false looks not natural.

use a variable could make the test more explicit for example something like the following set clear expectations to someone else who read the code

```golang

isLoadGeneratorWrk2LGOrForlioLG := slices.Contains([]string{Wrk2LG.Name(), FortioLG.Name()}, testClient.LoadGenerator)

if !isLoadGeneratorWrk2LGOrForlioLG { 
```

PR #17413, opened 2026-02-12
[Model] AWS S3: four new relationship definitions

leecalcote:


leecalcote:
@Debashich, what do you think?

Debashich:
Hi @leecalcote, apologies for the slight delay in getting back to you, and thank you for the opportunity to review these definitions!

I’ve verified the logic locally and everything is working as intended. Apart from a small suggestion around keeping the patch scope limited to ARN injection (to avoid any unintended replace behavior), the definitions look good to me. LGTM overall.

To help further build out these AWS integrations, I noticed we could also add the reciprocal access relationships for the target components. Since SQS is already being handled, I’d be happy to help out with the SNS side.
Would it be alright if I open a new issue for this? I’d love to contribute by helping to complete that functional path.

PR #17415, opened 2026-02-12
[Server] Use file name instead of "Autogenerated" for imported designs

### LINKED ISSUE #12062: Use file name instead of "Autogenerated".
### Current Behavior

Designs can be imported in multiple ways from MesheryUI, for some of the scenarios the file name or name provided as input are not respected and server assigns name as "Autogenerated".
### Expected Behavior

Designs can be imported in various ways from MesheryUI. However, the filename or name is not honoured in some cases, and the server assigns the name "Autogenerated."

### Environment

---

guan404ming:
**Notes for Reviewers**

- This PR fixes #12062

## Why                                                                                                                  
  
Imported designs ignore the provided filename and get named "Autogenerated" because the server unconditionally       
overwrites the design name with an empty payload field.                    

## How

- Remove the "Autogenerated" fallback and strip the file extension from the filename for use as the design name
- Preserve the design name from conversion when no explicit name is provided in the import payload
- Fall back to the filename (sans extension) only when both the payload name and conversion name are empty


YASHMAHAKAL:
This is a good change @guan404ming !

guan404ming:
Thanks!

PragalvaXFREZ:
Thank you @guan404ming. I had a small comment: extension stripping in the import naming path currently removes only the last suffix, so a file like `my-chart.tar.gz` becomes `my-chart.tar` instead of `my-chart`. Since `.tar.gz/.tgz` are supported import types, can we strip compound extensions explicitly (ideally via a shared helper) so design names stay clean and consistent? LGTM otherwise.

guan404ming:
@PragalvaXFREZ Thanks for the good catch! Added a .tar.gz strip before the single-extension trim in both design_import.go and pattern.go. It's the only compound extension in the supported import types, so a targeted trim keeps it simple.

PragalvaXFREZ (dismissed):
LGTM.

server/models/pattern/core/pattern.go
```diff
@@ -327,8 +329,9 @@ func processCytoElementsWithPattern(eles []cytoscapejs.Element, callback func(sv
 }
 
 func NewPatternFileFromK8sManifest(data string, fileName string, ignoreErrors bool, reg *registry.RegistryManager) (pattern.PatternFile, error) {
-	if fileName == "" {
-		fileName = "Autogenerated"
+	if fileName != "" {
```
lekaf974: what if `fileName == ""`

guan404ming: It will have some issue. I just extracted the name-stripping logic into an exported DesignNameFromFileName helper in pattern.go that explicitly handles the empty-string case (early return). Both NewPatternFileFromK8sManifest and the handler now use this shared function, eliminating the duplication

server/handlers/design_import.go
```diff
@@ -276,7 +276,13 @@ func (h *Handler) DesignFileImportHandler(
...
 	}
 
-	design.Name = importDesignPayload.Name
+	if importDesignPayload.Name != "" {
+		design.Name = importDesignPayload.Name
+	} else if design.Name == "" {
+		if name := pCore.DesignNameFromFileName(fileToImport.FileName); name != "" {
+			design.Name = name
+		}
+	}
```
lekaf974, line 350:
```suggestion
	design.Name = importDesignPayload.Name
	if design.Name == "" {
		if name := pCore.DesignNameFromFileName(fileToImport.FileName); name != "" {
			design.Name = name
		}
```

server/models/pattern/core/pattern_test.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func TestDesignNameFromFileName(t *testing.T) {
+	tests := []struct {
+		name         string
+		fileName     string
+		expectedName string
+	}{
+		{
+			name:         "regular yaml file",
```
lekaf974:
```suggestion
			name:         "given regular yaml file when DesignNameFromFileName then return name",
```
please try using given, when then to make test name useful.
This way when a test is failing it is easy to know the input the method call and expected behavior

server/models/pattern/core/pattern_test.go
lekaf974, line 1: can you add edge cases like unsupported extensions (eg. `.xls`, `.zip`, etc.)

lekaf974:
@guan404ming could you address or return feedback on review

server/models/pattern/core/pattern_test.go
guan404ming: Just added. Thanks!

server/handlers/design_import.go
guan404ming: Updated

server/models/pattern/core/pattern_test.go
guan404ming: Updated, thanks!

guan404ming:
> @guan404ming could you address or return feedback on review

Sorry for the late I've handled all comments. Please take another look, thanks!

lekaf974 (dismissed):

lekaf974 (approved):
LGTM

PragalvaXFREZ (approved):

Issue #16684, opened 2025-12-17
[mesheryctl] `environment list --orgID` offers invalid response


leecalcote:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

### Current Behavior
```
$ mesheryctl environment list --orgID 4667e08e-6cc2-42c9-bf9a-fee1e4af91e7
Error: uuid: incorrect UUID length 10 in string "Gingergeek"
```

Gingergeek is the name of the organization, so it's evident that the command is retrieving some information about the organization, but it is failing somewhere, responding oddly.

### Environment

- Host OS: Mac
- Meshery Server Version: stable-v0.8.181
- Meshery Client Version: stable-vv0.8.181
- Platform: Kubernetes

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 📚 [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- 📺 [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- ⌨️ [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- ⌨️ [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- ⌨️ [Meshkit errors Guide](https://docs.meshery.io/project/contributing/contributing-error)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

1012Charan:
Hey @leecalcote , I looked into this and may have found the issue.

The length 10 matches the string "Gingergeek" (the org name). It looks like the server is returning the Organization Name in the `organization_id` field, but the CLI struct expects a UUID type there.

Since this is a type mismatch (Server sending Name string vs CLI expecting UUID), maybe we fix the server side to return the actual UUID

ItsYash1421:
can i work on this?

[...]

PR #17420, opened 2026-02-12
[mesheryctl] Handle non-UUID organization_id in environment list

ashwingopalsamy:
**Notes for Reviewers**

- This PR fixes [#16684](https://github.com/meshery/meshery/issues/16684) by making `mesheryctl environment list --orgID` resilient when the server returns a non-UUID `organization_id` value (for example, organization name strings like `"Gingergeek"`).

#### Why this change is needed

`environment list` previously decoded API responses into a strict UUID-backed type. If `organization_id` was returned as a string name, decode failed with:
`uuid: incorrect UUID length 10 in string "Gingergeek"`

This prevented listing environments even though the request itself was valid.

**Before:**

```bash
# Before (on parent commit, reproducible test failure)
$ go test ./mesheryctl/internal/cli/root/environments -run TestListEnvironment/List_environments_when_server_returns_non-UUID_organization_id -count=1
Error: uuid: incorrect UUID length 10 in string "Gingergeek"
--- FAIL: TestListEnvironment/List_environments_when_server_returns_non-UUID_organization_id
FAIL
```

**After:**

```bash
# After (current branch)
$ go test ./mesheryctl/internal/cli/root/environments -run TestListEnvironment/List_environments_when_server_returns_non-UUID_organization_id -count=1
ok  	github.com/meshery/meshery/mesheryctl/internal/cli/root/environments	0.xxxs
```

### DIFF
--- a/mesheryctl/internal/cli/root/environments/list.go
+++ b/mesheryctl/internal/cli/root/environments/list.go
@@ -15,8 +15,10 @@
 package environments
 
 import (
+	"encoding/json"
 	"fmt"
 
+	"github.com/gofrs/uuid"
 	"github.com/meshery/meshery/mesheryctl/internal/cli/pkg/display"
 	"github.com/meshery/meshery/mesheryctl/pkg/utils"
 	"github.com/meshery/schemas/models/v1beta1/environment"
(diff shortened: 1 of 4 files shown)


mesheryctl/internal/cli/root/environments/list.go
lekaf974, line 1: This exactly what is must not be done, we are using schema driven development and must not create struct individually. 

From what I understand is more like a data corruption since ID should be a UUID

lekaf974, line 1: instead of skip we need to understand why ID is not UUID and fix the root cause

lekaf974: any update here ?

lekaf974:
@ashwingopalsamy I identified the root cause and will create a fix from the source itself, closing it

Thanks for your effort

PR #17422, opened 2026-02-12
style: resolve remaining staticcheck violations and enable linter

KhaledSayed04:
- This PR addresses the final batch of staticcheck violations and fully enables `staticcheck` in the CI pipeline.
- This is Batch 3 of the cleanup series closing #16667.

**Key Changes:**
- **Code Cleanup (QF1008):** Removed redundant selector names for embedded struct fields (e.g., `dep.ObjectMeta.Name` -> `dep.Name`).
- **Performance (QF1010):** Replaced inefficient `fmt.Sprint` calls with direct string conversions/casts.
- **CI Configuration:** Updated `.golangci.yml` to fully enable `staticcheck` for all future commits.

**Fixed Checks**
* [x] **QF1008** (Redundant Selectors)
* [x] **QF1010** (Formatting)

### DIFF
--- a/docs/meetings/2026/week-of-02-26-2026/Raunak.md
+++ b/docs/meetings/2026/week-of-02-26-2026/Raunak.md
@@ -1,0 +1,1 @@
+Yada yada 
--- a/docs/meetings/2026/week-of-02-26-2026/Suhani_Garg.md
+++ b/docs/meetings/2026/week-of-02-26-2026/Suhani_Garg.md
@@ -1,0 +1,3 @@
+Hi everyone! 
+
+MySelf Suhani Garg, and I’m a B.Tech Computer Science student specializing in Software Product Engineering. I’m currently exploring cloud-native technologies and open-source contributions, and I’m really excited to be part of the Meshery community.
--- a/docs/meetings/2026/week-of-02-26-2026/Vibhanshu_Singh.md
+++ b/docs/meetings/2026/week-of-02-26-2026/Vibhanshu_Singh.md
@@ -1,0 +1,11 @@
+Hello,
+
+My name is Vibhanshu Singh.
+I'm a 3rd-year B.Tech student and have been contributing to many open-source projects for a while now.
+I can code proficiently in Go, Java, Python, and TypeScript.
+I am here to learn and grow.
+Looking forward to helping the community and making good connections.
+
+You can find my work on [GitHub](https://github.com/singhvibhanshu)
+
+Thanks!
--- a/docs/meetings/2026/week-of-02-26-2026/gangireddypavan-attendance.md
+++ b/docs/meetings/2026/week-of-02-26-2026/gangireddypavan-attendance.md
@@ -1,0 +1,7 @@
+Gangireddy Pavan - Attendance
+
+Hi, I am Gangireddy Pavan. I am new to open source and recently started contributing to Meshery.
+I am currently exploring:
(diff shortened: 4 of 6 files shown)


KhaledSayed04:
Hi @lekaf974 ,
With this PR merged, we re-enable staticcheck linter

server/models/remote_provider.go
lekaf974, line 1: did you run this locallly to ensure eveything is still working as expected ?

KhaledSayed04: @lekaf974 
Yes.
- The old implementation `fmt.Sprint(bdr)` actually formats the `[]byte` as an array of base-10 integers (e.g., `[123 34 101...]`), which makes the error logs hard to read. Casting it with `string(bdr)` correctly converts the payload into a readable UTF-8 string (e.g., `{"error": "..."}`). 
```go 
	// Simulating a failed JSON response from a server
	bdr := []byte(`{"message": "unauthorized access"}`)

	// The old code
	oldWay := fmt.Sprint(bdr)
	fmt.Println("OLD WAY LOG OUTPUT:")
	fmt.Println(oldWay)
	
	fmt.Println("-------------------")

	// The new code
	newWay := string(bdr)
	fmt.Println("NEW WAY LOG OUTPUT:")
	fmt.Println(newWay)
```
- I also ran `go test ./...` and `go build ./...` in the `models` package to ensure the type change didn't break any downstream assertions or compilation.

lekaf974:
Was busy

KhaledSayed04:
@lekaf974
I fixed the `S1034` and `S1002` issues (along with the unused variable) that appeared in the `validator.go` file.
The linter works fine now

mesheryctl/internal/cli/pkg/flags/validator.go
```diff
@@ -51,8 +50,8 @@ func validateSemver(fl validator.FieldLevel) bool {
 // This is necessary because the default validator does not have a built-in validation behaving as expected for boolean fields,
 // especially when using flags in cobra
 func validateBoolean(fl validator.FieldLevel) bool {
-	if val, ok := fl.Field().Interface().(bool); ok {
-		return val == true || val == false
+	if _, ok := fl.Field().Interface().(bool); ok {
+		return true
```
lekaf974, line 54: why doing this change, there is a comment that explain this is done explicitly. this will break command flag validation

KhaledSayed04: @lekaf974 
I understand the concern about breaking the flag validation.
To clarify why I made this specific change: the type assertion
```golang
if _, ok := fl.Field().Interface().(bool); ok
```
already does the complete validation. If the underlying field is not a boolean (e.g., a string or nil), the type assertion fails (ok is false) and it correctly returns false.

If the assertion succeeds, it guarantees the field is a Go bool. Since a Go bool can physically only be true or false, evaluating `val == true || val == false` inside that block is a tautology that will always return `true`. In fact, that specific tautology is what caused the CI pipeline to fail with the staticcheck `S1002` error.

Using `_, ok` and returning `true` preserves the exact same validation behavior (returning `true` if it's a `bool`, `false` if it's not) while satisfying the linter. 
Let me know if you'd still prefer me to revert it and use a `//nolint` directive instead!

lekaf974: Make sense, LGTM

PR #17429, opened 2026-02-12
[UI] chore: update to Next 15 and react 18

saurabhraghuvanshii:
**Notes for Reviewers**
Why CustomRadioWidget:
1. The Import Design modal uses a radio field for “File Upload” vs “URL Import”. RJSF was using the theme’s default RadioWidget, which in this app caused:
“Element type is invalid… Check the render method of RadioWidget.”

So the default radio widget wasn’t a valid React component here (wrong export or different UI library). CustomRadioWidget replaces it with a widget built from MUI components (Radio, RadioGroup, FormControlLabel, FormLabel), so the options render without that error.

2. Safe labels
- It uses safeDisplayValue(option.label) so when the schema sends objects as labels (which caused React # 130 elsewhere), they’re turned into strings instead of breaking the UI.

3. Wired in via RJSF
In RJSF.tsx, the form’s widgets include RadioWidget: CustomRadioWidget, so any RJSF form that uses a radio (including Import Design) uses this component instead of the theme default.

In short: We need it so the Import Design “File Upload / URL Import” radio works and doesn’t throw. Without it, RJSF falls back to the theme’s RadioWidget and the “Element type is invalid” error comes back when we open Import Design.

https://github.com/user-attachments/assets/90cc1991-b783-4099-965a-e12813ccb7e8

- This PR fixes #

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1694,9 +1694,9 @@
       }
     },
     "node_modules/@microlink/react-json-view": {
-      "version": "1.31.1",
-      "resolved": "https://registry.npmjs.org/@microlink/react-json-view/-/react-json-view-1.31.1.tgz",
-      "integrity": "sha512-UsWwhSgNn06RDVrhnNejFypmu3xtfpxAzQ2BqXWY7NWs5FKZnjaNDFmB5kNpC1NDjzAH2v2oDq4JFEV3eLyDkw==",
+      "version": "1.31.3",
+      "resolved": "https://registry.npmjs.org/@microlink/react-json-view/-/react-json-view-1.31.3.tgz",
+      "integrity": "sha512-S9zkcgtciN6YphL+XAHnoRnv/e08r/TFYmcANrqTHMHNGJRHdPVq6oRzlCeQ4MIzOBs3CmPdC7gktmmmaOlzZA==",
       "license": "MIT",
       "dependencies": {
         "react-base16-styling": "~0.10.0",
@@ -2404,9 +2404,9 @@
       }
     },
     "node_modules/@rjsf/core": {
-      "version": "6.3.1",
-      "resolved": "https://registry.npmjs.org/@rjsf/core/-/core-6.3.1.tgz",
-      "integrity": "sha512-LTjFz5Fk3FlbgFPJ+OJi1JdWJyiap9dSpx8W6u7JHNB7K5VbwzJe8gIU45XWLHzWFGDHKPm89VrUzjOs07TPtg==",
+      "version": "6.4.1",
+      "resolved": "https://registry.npmjs.org/@rjsf/core/-/core-6.4.1.tgz",
+      "integrity": "sha512-+QaiSgQnOuO6ghIsohH2u/QcylkN+Da2968a75g/i4oARYJRYVxXDm2u3JR5aXndpMb4t4jTFrYyG8cNIv6oEg==",
       "license": "Apache-2.0",
       "dependencies": {
         "lodash": "^4.17.23",
@@ -2418,14 +2418,14 @@
         "node": ">=20"
       },
       "peerDependencies": {
-        "@rjsf/utils": "^6.3.x",
+        "@rjsf/utils": "^6.4.x",
         "react": ">=18"
       }
     },
     "node_modules/@rjsf/mui": {
-      "version": "6.3.1",
-      "resolved": "https://registry.npmjs.org/@rjsf/mui/-/mui-6.3.1.tgz",
-      "integrity": "sha512-htZgruxZioYpsqtBUOkWmm7MeBzsHaYDhHTEcaw9O08R/mGpyvrZw9f93ryhEH3hBQs5SSsHZeNsLwr00PGEEg==",
+      "version": "6.4.1",
+      "resolved": "https://registry.npmjs.org/@rjsf/mui/-/mui-6.4.1.tgz",
+      "integrity": "sha512-1oz5INDZ7/rEwGaCn52ZMt+EqROeHNAI5kL6us2AMl7yKc3D6K1/X3H5quHdnr0dnh9Y8ivpOXroGy1ka4IBsA==",
       "license": "Apache-2.0",
       "engines": {
         "node": ">=20"
@@ -2435,15 +2435,15 @@
         "@emotion/styled": "^11.6.0",
         "@mui/icons-material": "^7.0.0",
         "@mui/material": "^7.0.0",
-        "@rjsf/core": "^6.3.x",
-        "@rjsf/utils": "^6.3.x",
+        "@rjsf/core": "^6.4.x",
+        "@rjsf/utils": "^6.4.x",
         "react": ">=18"
       }
     },
     "node_modules/@rjsf/utils": {
-      "version": "6.3.1",
-      "resolved": "https://registry.npmjs.org/@rjsf/utils/-/utils-6.3.1.tgz",
-      "integrity": "sha512-ve2KHl1ITYG8QIonnuK83/T1k/5NuxP4D1egVqP9Hz2ub28kgl0rNMwmRSxXs3WIbCcMW9g3ox+daVrbSNc4Mw==",
+      "version": "6.4.1",
+      "resolved": "https://registry.npmjs.org/@rjsf/utils/-/utils-6.4.1.tgz",
+      "integrity": "sha512-5NL3jwt3rIS5/WRTrKt++y40FS/ScKGVwYJ3jIrHSQHSwBdLnd4cHf2zcnA97L1Klj8I6tvS/ugh+blf/Diwuw==",
       "license": "Apache-2.0",
       "dependencies": {
         "@x0k/json-schema-merge": "^1.0.2",
@@ -2467,9 +2467,9 @@
       "license": "MIT"
     },
     "node_modules/@rjsf/validator-ajv8": {
-      "version": "6.3.1",
-      "resolved": "https://registry.npmjs.org/@rjsf/validator-ajv8/-/validator-ajv8-6.3.1.tgz",
-      "integrity": "sha512-2RHDxBc0gBplPniau5UZj7aznpTelSBm1b3DNybok8L0NuIfmndbp9kNXgFuRvlyfsQSyYmZSBjbzeYqr0Hpcw==",
+      "version": "6.4.1",
+      "resolved": "https://registry.npmjs.org/@rjsf/validator-ajv8/-/validator-ajv8-6.4.1.tgz",
+      "integrity": "sha512-Gx28sRIV7E4CYs2c7BxOGLX44p5IlJE+IaD7GbVk1S+6TxDATqFBSYYZukLB+/vNk3urpndQMreQLKW3W7POHQ==",
       "license": "Apache-2.0",
       "dependencies": {
         "ajv": "^8.17.1",
@@ -2481,7 +2481,7 @@
         "node": ">=20"
       },
       "peerDependencies": {
-        "@rjsf/utils": "^6.3.x"
+        "@rjsf/utils": "^6.4.x"
       }
     },
     "node_modules/@rjsf/validator-ajv8/node_modules/ajv-formats": {


YASHMAHAKAL:
@saurabhraghuvanshii, looks good !

Rajesh-Nagarajan-11:
Can you address Gemini’s high-priority comments on the Next.js update?

ui/lib/data-fetch.ts
```diff
@@ -1,5 +1,3 @@
-import fetch from 'isomorphic-unfetch';
```
pontusringblom, line 1: Is this no longer used?

ui/pages/extension/[...component].tsx
```diff
@@ -20,6 +20,22 @@ import {
   updateTitle,
 } from '@/store/slices/mesheryUi';
 import { useDispatch, useSelector } from 'react-redux';
+import { GetStaticPaths, GetStaticProps } from 'next';
```
Comment (pontusringblom) [inline ui/pages/extension/[...component].tsx:23]: Why is there a change here?

ui/pages/_document.tsx
```diff
@@ -1,12 +1,58 @@
...
+} from 'next/document';
+import createEmotionServer from '@emotion/server/create-instance';
+import createCache from '@emotion/cache';
 import { PureHtmlLoadingScreen } from '@/components/LoadingComponents/LoadingComponentServer';
 
-const registry = createStyleRegistry();
-const flush = registry.flush();
-class MesheryDocument extends Document {
+// Create emotion cache for SSR
+const createEmotionCache = () => {
```
leecalcote, line 15: Is this change strictly necessary for Next 15 or React 18 compatibility?

ui/pages/_document.tsx
```diff
@@ -111,64 +160,4 @@ class MesheryDocument extends Document {
   }
 }
 
-MesheryDocument.getInitialProps = (ctx) => {
-  // resolution order
```
leecalcote, line 115: @saurabhraghuvanshii why has our documentation been removed?

ui/rtk-query/meshModel.ts
```diff
@@ -170,9 +170,12 @@ export const useGetCategoriesSummary = () => {
     return categoryMap;
   };
 
-  useEffect(async () => {
-    const categoryMap = await fetchModelsForCategories();
-    setCategoryMap(categoryMap);
+  useEffect(() => {
```
leecalcote, line 173: @gemini-code-assist, why is this change necessary? Is it an improvement? Why is it worse? Why is it better?

ui/next.config.js
```diff
@@ -1,14 +1,37 @@
 /** @type {import('next').NextConfig} */
 
-const removeImports = require('next-remove-imports')();
-const nextConfig = removeImports({
-  reactStrictMode: true,
+const nextConfig = {
+  reactStrictMode: false,
```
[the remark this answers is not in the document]
leecalcote: Why is this disabled? What challenges are insurmountable when it is enabled?

ui/package.json
```diff
@@ -8,10 +8,10 @@
     "lint": "npx --yes eslint@8.57.1 --ext .js,.jsx .",
     "lint:fix": "npx --yes eslint@8.57.1 --fix --ext .js,.jsx .",
     "format": "npx prettier --write \"**/*.{ts,tsx,js,jsx,md,json}\"",
-    "build": "npm run clean && cross-env next build",
+    "build": "npm run clean && cross-env NODE_OPTIONS=--max-old-space-size=8192 next build",
```
leecalcote: Remove this option. This is not necessary for upgrade. This is a separate concern.

ui/next.config.js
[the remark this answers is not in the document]
saurabhraghuvanshii: I disabled reactStrictMode temporarily for the migration. With Next 15 and React 18, Strict Mode was triggering additional dev-time issues that required broader refactoring. I’ll re-enable it after things stabilize.

ui/lib/data-fetch.ts
saurabhraghuvanshii: yeah,  this is no longer needed as fetch is built‑in.

ui/pages/extension/[...component].tsx
Comment (saurabhraghuvanshii) [reply ui/pages/extension/[...component].tsx:23]: Because with Next.js 15 + output: 'export', any dynamic route in the Pages Router must define getStaticPaths (and usually getStaticProps) so that Next knows which paths to prerender as static HTML.

ui/pages/_document.tsx
saurabhraghuvanshii: No, this is required for Emotion, MUI, and Sistent SSR to function correctly and consistently.

I’ll revert it after the next MUI-Datatables release so I can properly reassess what’s necessary and what can be removed.

ui/pages/_document.tsx
saurabhraghuvanshii: Here is new form of https://github.com/meshery/meshery/pull/17429/changes#diff-f08b73499e013f3a336345f3b08dfc0b21f66bcc2c2ccc3f50012cd42033d469R19   

 rewritten for Emotion instead of JSS and updated for Next 15/React 18.

ui/next.config.js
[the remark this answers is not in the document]
leecalcote: @saurabhraghuvanshii, no. This doesn't answer my questions, which called for specific examples of items that we're not able to handle at the moment. What are they?

ui/rtk-query/meshModel.ts
leecalcote: @gemini-code-assist, why use `useEffect` at all and not another approach?

ui/package.json
```diff
@@ -8,7 +8,7 @@
     "lint": "npx --yes eslint@8.57.1 --ext .js,.jsx .",
     "lint:fix": "npx --yes eslint@8.57.1 --fix --ext .js,.jsx .",
     "format": "npx prettier --write \"**/*.{ts,tsx,js,jsx,md,json}\"",
-    "build": "npm run clean && cross-env NODE_OPTIONS=--max-old-space-size=8192 next build",
+    "build": "npm run clean && cross-env next build",
```
leecalcote, line 11: There should be no changes to this file, correct?

saurabhraghuvanshii: Already reverted.

ui/next.config.js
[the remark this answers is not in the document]
saurabhraghuvanshii: It is turned off as a tactical choice to reduce noise and scope during the Next 15, React 18  SWC, not because of a specific bug we can’t fix.

Example:
`useEffect(async () => { ... })` returned a Promise; React expects a cleanup function or `void`.
Strict Mode made it worse: In dev, React 18 double-invokes effects, so this anti‑pattern caused extra runs and surfaced as `TypeError: n is not a function more often.` 

And I want to keep changes minimum for this migration.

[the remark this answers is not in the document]
saurabhraghuvanshii: By the way, I just checked — the build works fine with true as well, so I’ve reverted it to true. Thanks for pointing that out.

rishiraj38:
Overall, this PR is excellent and the upgrade is highly compatible. The Webpack aliasing for React/MUI is a great solution for hook invocation bugs, and the Emotion SSR configuration in `_document.tsx` is perfectly executed to prevent FOUC. The `useEffect` fix for React 18 is also spot-on.

Two minor points to consider before merging:
1. **Windows support in Scripts**: In `package.json`, substituting `"clean": "rm -rf .next out"` with `"clean": "rimraf .next out"` (and adding `rimraf` to dev dependencies) ensures cross-platform compatibility for Windows contributors.
2. **TypeScript Build Errors**: I noticed `ignoreBuildErrors: true` was added to `next.config.js`. I tried running the build with `ignoreBuildErrors: false` and it successfully passed without any TypeScript errors, so it looks highly compatible. We should probably remove this flag so we don't accidentally silence future errors!

Looks good to me!

ui/components/Dashboard/index.tsx
```diff
@@ -118,6 +114,11 @@ const Dashboard = () => {
     width: '40',
   };
 
+  const WIDGETS = getWidgets({ iconsProps, isEditMode });
+  const availableHandles = ['s', 'w', 'e', 'n', 'sw', 'nw', 'se', 'ne'];
+
+  const cols = { lg: 12, md: 10, sm: 6, xs: 4, xxs: 2 };
+
```
leecalcote, line 121: Why is this change included in this PR?

ui/components/General/style.tsx
```diff
@@ -159,6 +159,10 @@ export const NavigatorListItem = styled(ListItemButton, {
     ? theme.palette.background.brand.default
     : theme.palette.background.constant.disabled,
   fill: isActive ? theme.palette.background.brand.default : '#fff',
+  '& a': {
+    color: 'inherit',
+    textDecoration: 'none',
+  },
```
leecalcote, line 165: Why is this change included in this PR?

ui/components/General/style.tsx
```diff
@@ -229,6 +233,10 @@ export const SideBarListItem = styled(ListItemButton, {
...
```
leecalcote, line 239: Why is this change included in this PR?

ui/components/Dashboard/index.tsx
saurabhraghuvanshii: before that change  `iconsProps`  is used on line 109 https://github.com/meshery/meshery/blob/master/ui/components/Dashboard/index.tsx#L109 
 before being defined on line 114 https://github.com/meshery/meshery/blob/master/ui/components/Dashboard/index.tsx#L114  

So that was given `TDZ` (Temporal Dead Zone). that's why I have to make change here.

ui/components/General/style.tsx
saurabhraghuvanshii: SideBarListItem sets the color on the parent ListItemButton, but the Link component renders an <a> element which gets the browser default color (-webkit-link). Because of that, the link was not inheriting the sidebar color.

Adding

'& a': { color: 'inherit', textDecoration: 'none' }

forces the <a> to inherit the parent color so the active/inactive theme styles work correctly after the Next.js 15 / React 18 upgrade.

ui/components/General/style.tsx
saurabhraghuvanshii: Same here.

marblom007:
@saurabhraghuvanshii all of the e2e tests are failing. This PR is not ready for merge, right?

ui/next.config.js
```diff
@@ -1,14 +1,37 @@
...
+  // Static export (replaces removed `next export` in Next.js 15)
+  output: 'export',
+
   eslint: {
     ignoreDuringBuilds: true,
   },
+
+  typescript: {
+    ignoreBuildErrors: true,
+  },
```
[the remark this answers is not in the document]
marblom007: @saurabhraghuvanshii, why is this critical comment not being addressed?

ui/package.json
```diff
@@ -4,14 +4,14 @@
...
+    "dev": "npm run lint:fix && cross-env node ui_dev_server.js",
     "lint": "npx --yes eslint@8.57.1 --ext .js,.jsx,.ts,.tsx .",
     "lint:fix": "npx --yes eslint@8.57.1 --fix --ext .js,.jsx,.ts,.tsx .",
     "format": "npx prettier --write \"**/*.{ts,tsx,js,jsx,md,json}\"",
     "build": "npm run clean && cross-env next build",
     "start": "NODE_ENV=production node ui_dev_server.js",
-    "export": "next export",
-    "export-playground": "PLAYGROUND=true next export -o playground",
+    "export": "echo 'Static export is handled by next build with output: export in next.config.js'",
+    "export-playground": "cross-env PLAYGROUND=true next build",
```
leecalcote: Undo this change. Move to another PR.

ui/package.json
```diff
@@ -4,14 +4,14 @@
   "description": "collaborative management plane",
   "main": "index.js",
   "scripts": {
-    "dev": "npm run lint:fix && node ui_dev_server.js",
+    "dev": "npm run lint:fix && cross-env node ui_dev_server.js",
```
leecalcote: Undo this change. Move to another PR.

ui/package.json
```diff
@@ -22,7 +22,7 @@
     "relay": "relay-compiler",
     "prepare": "cd .. && husky ui/.husky",
     "postinstall": "patch-package",
-    "clean": "rm -rf .next,out"
+    "clean": "rm -rf .next out"
```
leecalcote: Undo this change. Move to another PR.

ui/package.json
```diff
@@ -34,21 +34,23 @@
...
-    "@rjsf/core": "^5.24.8",
-    "@rjsf/mui": "^5.24.8",
-    "@rjsf/utils": "^5.24.8",
-    "@rjsf/validator-ajv8": "^5.24.8",
+    "@rjsf/core": "^6.3.1",
+    "@rjsf/mui": "^6.3.1",
+    "@rjsf/utils": "^6.3.1",
+    "@rjsf/validator-ajv8": "^6.3.1",
     "@sistent/mui-datatables": "^5.1.4",
     "@sistent/sistent": "^0.16.13",
```
leecalcote: Why does this PR not include new new versions of sistent and mui-datatables?

ui/next.config.js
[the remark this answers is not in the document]
saurabhraghuvanshii: I initially left it for a follow-up PR, but I’ve now addressed it in this PR.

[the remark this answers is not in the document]
leecalcote: Thanks. Please signal when this item is complete.

ui/pages/_document.tsx
leecalcote: @saurabhraghuvanshii given the @sistent/mui-datatables v6.0.1 release, I wonder what your findings are here.

ui/next.config.js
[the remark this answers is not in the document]
saurabhraghuvanshii: @gemini-code-assist Your suggestion already addressed.

PR #17453, opened 2026-02-14
feat(mesheryctl): update no data found output

lekaf974:

**Notes for Reviewers**

- This PR fixes #

### DIFF
--- a/mesheryctl/internal/cli/root/components/view.go
+++ b/mesheryctl/internal/cli/root/components/view.go
@@ -78,7 +78,7 @@ mesheryctl component view [component-name] -o [json|yaml] --save
 		}
 
 		if componentResponse.Count == 0 {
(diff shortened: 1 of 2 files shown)


mesheryctl/internal/cli/root/components/view.go
```diff
@@ -78,7 +78,7 @@ mesheryctl component view [component-name] -o [json|yaml] --save
 		}
 
 		if componentResponse.Count == 0 {
-			utils.Log.Info("No component(s) found for the given name: ", componentDefinition)
+			fmt.Println("No component(s) found with the name:", componentDefinition)
```
PragalvaXFREZ, line 81: Any reason we switched to `fmt.Println` here instead of `cmd.Println`?

lekaf974: just reuse the same tool that we use in other commands

PR #17455, opened 2026-02-15
[mesheryctl] Changed context to case sensitive

### LINKED ISSUE #17461: [mesheryctl] Inconsistent case while context creation

### Current Behavior
- The CLI accepts non-lowercase and mixed-case context names during creation (e.g., Name). However, it stores the name in lowercase (e.g., Name -> name). This creates inconsistency between user input and stored value. It may lead to unexpected overrides if a lowercase context already exists.
- When creating a context with an existing name the validation correctly restricts duplicate names only if the name is entirely in lowercase. However if the name contains any uppercase letters the system does not enforce this restriction and instead allows the creation of the context and overriding already existing context.

Image

### Desired Behavior

Since `context create` ultimately stores context names in lowercase, the CLI should validate input and reject non-lowercase names during creation.
For example, kubectl rejects non-lowercase pod names (e.g., `kubectl run Nginx-Pod` throws an RFC 1123 validation error), so following a similar approach here would make the behavior more predictable and consistent

### Approach
By implementing new meshkit error code stating
```bash
The context name "Name" is invalid: 
A name must consist of lower case alphanumeric characters or '-' and must start and end with an alphanumeric character (e.g. 'name', 'my-name', 'name1')
``` 

### Environment

- Host OS: Linux 
- Platform: Kubernetes

---

DharunMR:
**Notes for Reviewers**

- This PR fixes #17461

When creating a context using mixed case (e.g., "Name"), the CLI stores it as lowercase ("name"). If a lowercase context already exists, the new command ("Name") overrides it.

Before:
Screenshot_2026-02-15_11 11 54

After:
Screenshot_2026-02-16_15 52 29

### DIFF
--- a/mesheryctl/internal/cli/root/system/testdata/reset/TestResetContexts.yaml
+++ b/mesheryctl/internal/cli/root/system/testdata/reset/TestResetContexts.yaml
@@ -22,7 +22,6 @@ contexts:
         platform: docker
         channel: stable
         version: latest
-        provider: Layer5
 current-context: local2
 tokens:
     - name: default
--- a/docs-new/content/en/project/contributing/contributing-docs.md
+++ b/docs-new/content/en/project/contributing/contributing-docs.md
@@ -11,195 +11,30 @@ Before contributing, please review the [Documentation Contribution Flow](https:/
 Use the following steps to set up your development environment depending on your Operating System of choice, fork the repository, install dependencies and run the site locally.  You can then make changes, test locally, sign-off and commit, and push the changes made for review.
 
 {{% alert color="info" title="Meshery Documentation Design Specification" %}}
-See the [Meshery Documentation Design Specification](https://docs.google.com/document/d/17guuaxb0xsfutBCzyj2CT6OZiFnMu9w4PzoILXhRXSo/edit#) which serves to provide an overview of the tooling and approach used to create Meshery's documentation and it information architecture.
+See the [Meshery Documentation Design Specification](https://docs.google.com/document/d/17guuaxb0xsfutBCzyj2CT6OZiFnMu9w4PzoILXhRXSo/edit#) which serves to provide an overview of the tooling and approach used to create Meshery's documentation and its information architecture.
 {{% /alert %}}
 
 ## Documentation Framework
 
 Meshery documentation is built using the following components:
 
-- Framework - [Jekyll](https://jekyllrb.com)
-- Theme - [https://github.com/vsoch/docsy-jekyll](https://github.com/vsoch/docsy-jekyll)
-- Repo - [https://github.com/meshery/meshery/tree/master/docs](https://github.com/meshery/meshery/tree/master/docs)
+- Framework - [Hugo](https://gohugo.io) (Extended)
+- Theme - [Docsy](https://www.docsy.dev) (imported as a Hugo module)
+- Repo - [https://github.com/meshery/meshery/tree/master/docs-new](https://github.com/meshery/meshery/tree/master/docs-new)
 - Site - [https://docs.meshery.io](https://docs.meshery.io)
-- AWS API GW - an instance is configured to redirect from docs.meshery.io to cloud.layer5.io, because of the repo location of where the docs currently reside.
 
 ## Set up your development environment
 
-{{% alert color="info" title="Jekyll" %}}
-The Meshery Docs site is built using Jekyll - a simple static site generator. Jekyll can be installed on different platforms like Windows, Linux, and MacOS by the following steps
+{{% alert color="info" title="Hugo" %}}
+The Meshery Docs site is built using Hugo - a fast static site generator. Hugo is installed automatically as an npm dependency, so you only need Node.js and Go installed on your system.
 {{% /alert %}}
 
-### For Windows
+### Prerequisites
 
-{{% alert color="light" title="Note" %}}
-Windows users can run Jekyll by following the [Windows Installation Guide](https://jekyllrb.com/docs/installation/windows/) and also installing Ruby Version Manager [RVM](https://rvm.io). RVM is a command-line tool which allows you to work with multiple Ruby environments on your local machine. Alternatively, if you're running Windows 10 version 1903 Build 18362 or higher, you can upgrade to Windows Subsystem for Linux [WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10) and run Jekyll in Linux instead.
-{{% /alert %}}
+The following tools are required to build and run the Meshery documentation site locally:
 
-- Fire up your WSL VM and install the ruby version manager (RVM):
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">sudo apt update
-  sudo apt install curl g++ gnupg gcc autoconf automake bison build-essential libc6-dev \
-   	libffi-dev libgdbm-dev libncurses5-dev libsqlite3-dev libtool \
-   	libyaml-dev make pkg-config sqlite3 zlib1g-dev libgmp-dev \
-   	libreadline-dev libssl-dev
-  sudo gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
-  curl -sSL https://get.rvm.io | sudo bash -s stable
-  sudo usermod -a -G rvm `whoami`
-  </code>
-  </div></pre>
-
-  
-  If `gpg --keyserver` gives an error, you can use:
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">sudo gpg --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
-  </code>
-  </div></pre>
-  
-  OR
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">sudo gpg2 --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB %}
-  </code>
-  </div></pre>
-
-  Restart your WSL VM before moving forward to install Ruby.
-
-- To install Ruby, run:
-
-    <pre class="codeblock-pre"><div class="codeblock">
-    <code class="clipboardjs">bash
-    rvm install ruby
-    rvm --default use ruby 2.7.5
-    gem update
-    gem install jekyll bundler
-    </code>
-    </div></pre>
-  
-- Update the Gemfile located in meshery/docs directory
-
-    <pre class="codeblock-pre"><div class="codeblock">
-    <code class="clipboardjs">ruby '3.0.2'
-    </code>
-    </div></pre>
-
-  <strong>Note:</strong> In place of `3.0.2` add your installed version
-  
-- Also add this to the next line in the _Gemfile_.
-    <pre class="codeblock-pre"><div class="codeblock">
-    <code class="clipboardjs">gem 'wdm','>=0.1.0' if Gem.win_platform?
-    </code>
-    </div></pre>
-  
-{{% alert color="warning" title="Don't Commit Gemfile" %}}
-This is just a workaround for your local machine. So, do not commit or push the modified Gemfile or Gemfile.lock during Pull Requests.
-{{% /alert %}}
-  
-- Next, navigate to the following folder 
-  ```C:\Ruby24-x64\lib\ruby\gems\2.4.0\gems\eventmachine-1.2.5-x64-mingw32\lib```
-  
-- Add ```require 'em/pure_ruby' ``` in the first line of the <strong>eventmachine.rb</strong> file
-
-### For Linux
-
-- Prerequisites
-  
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">sudo apt-get update
-  sudo apt-get install autoconf bison build-essential libssl-dev libyaml-dev libreadline6-dev zlib1g-dev libncurses5-dev libffi-dev libgdbm3 libgdbm-dev
-  </code>
-  </div></pre>
-{{% alert color="info" title="Note" %}}
-The latest versions of Debian-based OS'es do not support <code>libgdbm3</code>, so if you encounter an issue with it's installation, try replacing <code>libgdbm3</code> with <code>libgdbm5</code> or <code>libgdbm6</code>.
-{{% /alert %}}
-
-#### Installing rbenv
-
-- Cloning the rbenv repository
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">git clone https://github.com/rbenv/rbenv.git ~/.rbenv
-  </code>
-  </div></pre>
-
-{{% alert color="info" title="Note" %}}
-Change bashrc with your shell specific rc file, for eg: if you are using zsh then the filename is zshrc.
-{{% /alert %}}
-
-- Setting the path
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">echo 'export PATH="$HOME/.rbenv/bin:$PATH"' >> ~/.bashrc
-  </code>
-  </div></pre>
-
-- rbenv init
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">echo 'eval "$(rbenv init -)"' >> ~/.bashrc
-  </code>
-  </div></pre>
-
-- Reload your bashrc
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">source ~/.bashrc
-  </code>
-  </div></pre>
-
-- Check installation
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">type rbenv
-  </code>
-  </div></pre>
-
-{{% alert color="info" title="Note" %}}
-After installing rbenv, you need to install the ruby-build plugin to use the `rbenv install` command. This step is necessary for both package manager and manual installations of rbenv.
-{{% /alert %}}
-
-- Install ruby-build as an rbenv plugin
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">git clone https://github.com/rbenv/ruby-build.git "$(rbenv root)"/plugins/ruby-build
-  </code>
-  </div></pre>
-  
-#### Install Ruby
-
-- rbenv install version
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">rbenv install 3.2.6
-  </code>
-  </div></pre>
-
-- To list all the versions that can be installed
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">rbenv install --list
-  </code>
-  </div></pre>
-
-- Set which Ruby version you want to use
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">rbenv global version
-  </code>
-  </div></pre>
-
-- Check Ruby installation
-
-  <pre class="codeblock-pre"><div class="codeblock">
-  <code class="clipboardjs">ruby -v
-  </code>
-  </div></pre>
-
-### For MacOS
-
-- Use docs here [Jekyll installation](https://jekyllrb.com/docs/installation/macos/)
+- [Node.js](https://nodejs.org/) (LTS recommended)
+- [Go](https://go.dev/dl/) (required for Hugo modules)
 
 {{% alert color="light" title="Note" %}}
 In case of any installation issues, use the [discussion forum](https://meshery.io/community#discussion-forums).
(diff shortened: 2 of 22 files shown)


DharunMR:
@PragalvaXFREZ Thanks for the feedback.
I felt that rejecting non-lowercase names might be too strict, but i think rejecting would simplify the behavior and avoids confusion.
For example, kubectl rejects non-lowercase pod names (e.g., `kubectl run nGinx-pod` throws an RFC 1123 validation error), so following a similar approach here would make the behavior more predictable and consistent???

lekaf974:
Like mentionned from a UX perspective it is a good idea. 
But on a community aspect is this change being discussed publicly in slack or ina dev meeting ? 
PS: this question is because this break current behavior which was like that since years so we need to ensure that we don't break anything.

Also what will happen if I have already a context in capitalize or upper case ?

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -103,12 +104,14 @@ Usage: mesheryctl system context create [context-name]`
 			tempCntxt.Components = components
 		}
 
-		err := config.AddContextToConfig(args[0], tempCntxt, viper.ConfigFileUsed(), set, false)
+		componentName := strings.ToLower(args[0])
```
lekaf974: is it a component or a context ?

DharunMR:
@lekaf974 
Since the server already stores context names in lowercase, I assume there wouldn’t be existing contexts stored in uppercase.

However, if there are edge cases where uppercase names were preserved earlier (before normalization), we should verify that to avoid breaking existing setups.

lekaf974:
> @lekaf974 Since the server already stores context names in lowercase, I assume there wouldn’t be existing contexts stored in uppercase.

Perhaps I was not aware of this but I was sure that mesheryctl context was stored locally and not on server.

> However, if there are edge cases where uppercase names were preserved earlier (before normalization), we should verify that to avoid breaking existing setups.

Even if it is stored locally or server side this PR should handle edge cases.

DharunMR:
@lekaf974 
Previously i was converting names to lowercase ,but now i have introduced meshkit error code.
So the newly created context will reject uppercase 
While `view`, `switch`, `delete` only validates existence, old users with upparcase context name can use those without getting any error

The current implementation only rejects uppercase names during creation, so new users won’t be confused.

lekaf974:
> @lekaf974 Previously i was converting names to lowercase ,but now i have introduced meshkit error code. So the newly created context will reject uppercase While `view`, `switch`, `delete` only validates existence, old users with upparcase context name can use those without getting any error
> 
> The current implementation only rejects uppercase names during creation, so new users won’t be confused.

Perhaps I didn't fully understand the intention of the PR but if the changes only affect new created contexts, we can just store it in lower case as long as the new created context name is clearly indicated and highlighted in the output.

DharunMR:
I think i need to be more clear,

- If i create context with name ("example") it stores as "example" everythig is fine
- If i create context with name ("Example", mixed uppercase or complete uppercase) also it stores it as "example" ,however, when the user later runs `mesheryctl system context view Example` it throws error saying `not found Example` because "Example" is automatically converted and stored as "example", same error follows for delete and switch.  
- The automatic convertion of "Example" to "example" could be suprising to users ,that's why i later changed the code 

Sorry for making it too complicated : ) 
Thanks for your feedback👍🏻

lekaf974:
> If i create context with name ("Example", mixed uppercase or complete uppercase) also it stores it as "example" ,however, when the user later runs mesheryctl system context view Example it throws error saying not found Example because "Example" is automatically converted and stored as "example", same error follows for delete and switch.

What if a context with `Example` was already created before we merge this PR ?

DharunMR:
@lekaf974 already existing context wont gets affected anyways

mesheryctl/internal/cli/root/system/error.go
```diff
@@ -336,5 +337,14 @@ func ErrUploadFileParams(err error) error {
 		[]string{err.Error()},
 		[]string{"Unable to upload parameters from config file with provided context"},
 		[]string{"Ensure you have a strong network connection and the right configuration set in your Meshconfig file." + FormatErrorReference()})
+}
 
+func ErrInvalidLowerCase(err error) error {
```
matrixkavi: Could we add some supporting positive and negative test cases for this please?

lekaf974:
> Left a review comment. It would be beneficial to add a test to support the validation change

this is a good point adding unit test so we will be confident on the changes

DharunMR:
@lekaf974  i added unit test 👍🏻

lekaf974:
Also all these question is really a matter of UX

matrixkavi:
@DharunMR - could you please squash all those commits into a single commit so that the commit history looks fine please?

lekaf974:
> @DharunMR - could you please squash all those commits into a single commit so that the commit history looks fine please?

this is not required, merge squash is also doing the trick

mesheryctl/internal/cli/root/system/context_test.go
```diff
@@ -68,9 +68,7 @@ func TestViewContextCmd(t *testing.T) {
 			}
 
 			if tt.ExpectError {
-				if err == nil {
-					t.Fatalf("expected error, got nil")
-				}
+				t.Fatalf("expected error, got nil")
```
lekaf974, line 74: not sure to understand why you remove this check ?

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -160,15 +174,15 @@ mesheryctl system context delete [context name]`
 					return errors.New("new context wrongly set")
 				}
 
-				if newContext == args[0] {
+				if newContext == contextName {
 					return errors.New("choose a new context other than the context being deleted")
```
lekaf974: meshkit error

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -72,13 +73,21 @@ mesheryctl system context create [context-name]
...
+	Args: func(_ *cobra.Command, args []string) error {
 		if len(args) == 0 {
 			const errMsg = `Please provide a context name.
 Usage: mesheryctl system context create [context-name]`
 			return errors.New(utils.SystemContextSubError(fmt.Sprintf("%s\n", errMsg), "create"))
 		}
 
+		if args[0] != strings.ToLower(args[0]) {
+			return ErrInvalidLowerCase(fmt.Errorf("context name %s is invalid", args[0]))
+		}
```
lekaf974: why not simply create name in lower case ?

DharunMR: I will put this in Today's dev meeting agenda, so that i can show the demo of whats happening.

edited: (due to time constraints I couldn’t present it)

mesheryctl/internal/cli/root/system/context_test.go
DharunMR: Previously we are checking `if err != nil` , so if that case doesnt pass it is obviously `err == nil` .That `if err == nil` case doesnt do much, also my code editor showed warning so i removed it

mesheryctl/internal/cli/root/system/context.go
DharunMR: It always creates name in lowercase no matter what user input. The code mentioned here will check name case and return error if it had uppercase. I'm slightly confused, could you please give me any suggestion of what should i do?

lekaf974: Why just not save to lowercase instead of returning an error ?  I don't see any value here

DharunMR: Yes i can do that, i just thought of making it strict with case

lekaf974: just apply it without error, LGTM

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -74,16 +74,16 @@ mesheryctl system context create context-name --components meshery-nsm --platfor
...
 	Annotations: linkDocContextCreate,
 	Args: func(_ *cobra.Command, args []string) error {
-		if len(args) == 0 {
+		if len(args) != 1 {
+			return utils.ErrInvalidArgument(fmt.Errorf("%s", errArgMsg))
+		}
+
+		if len(args) > 1 {
 			const errMsg = `Please provide a context name.
 Usage: mesheryctl system context create [context-name]`
```
lekaf974: error message does not explicitly explain too many args were provided

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -112,14 +112,17 @@ Usage: mesheryctl system context create [context-name]`
...
 
 		err := config.AddContextToConfig(contextName, tempCntxt, viper.ConfigFileUsed(), set, false)
 		if err != nil {
 			return err
 		}
 
 		log.Printf("Added `%s` context", contextName)
+		if args[0] != strings.ToLower(args[0]) {
+			log.Printf("Context `%s` converted to lowercase `%s`", contextName, strings.ToLower(args[0]))
+		}
```
lekaf974: this is not needed, updating previous message would have been enough

DharunMR:
@lekaf974  Made changes as suggestions 👍

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -74,14 +74,12 @@ mesheryctl system context create context-name --components meshery-nsm --platfor
...
-			return utils.ErrInvalidArgument(fmt.Errorf("%s", errArgMsg))
+		if len(args) == 0 {
+			return utils.ErrInvalidArgument(fmt.Errorf("%s\n%s", errNoArgMsg, errContextUsageMsg))
 		}
 
 		if len(args) > 1 {
-			const errMsg = `Please provide a context name.
-Usage: mesheryctl system context create [context-name]`
-			return errors.New(utils.SystemContextSubError(fmt.Sprintf("%s\n", errMsg), "create"))
+			return utils.ErrInvalidArgument(fmt.Errorf("%s\n%s", errMultiArgMsg, errContextUsageMsg))
```
lekaf974: this could be handle with one check `if lan(args) != 1`

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -119,9 +117,10 @@ Usage: mesheryctl system context create [context-name]`
 			return err
 		}
 
-		log.Printf("Added `%s` context", contextName)
 		if args[0] != strings.ToLower(args[0]) {
-			log.Printf("Context `%s` converted to lowercase `%s`", contextName, strings.ToLower(args[0]))
+			log.Printf("Added `%s` context (converted to lowercase)", contextName)
+		} else {
+			log.Printf("Added `%s` context", contextName)
```
lekaf974: we have a work in progress to get ride of logrus and use utils.Log from meshkit logger

DharunMR:
@lekaf974 @PragalvaXFREZ 
Made changes in arg validation and logging msg, Good to merge?

mesheryctl/internal/cli/root/system/testdata/check/check.output.golden
PragalvaXFREZ, line 1: is this intentional change?

mesheryctl/internal/cli/root/system/error.go
```diff
@@ -57,9 +57,11 @@ const (
...
 
 var (
-	cmdType     string
-	contextdocs string = "See https://docs.meshery.io/reference/mesheryctl/system/context for usage details."
-	contextDir  string = "see that you have a correct context in your  meshconfig at `$HOME/.meshery/config.yaml`."
+	errArgMsg          = "only one argument must be provided"
+	errContextUsageMsg = `Usage: mesheryctl system context create [context-name]`
+	cmdType            string
+	contextdocs        string = "See https://docs.meshery.io/reference/mesheryctl/system/context for usage details."
+	contextDir         string = "see that you have a correct context in your  meshconfig at `$HOME/.meshery/config.yaml`."
```
PragalvaXFREZ:
```suggestion
	contextDir         string = "See that you have a correct context in your meshconfig at `$HOME/.meshery/config.yaml`."
```

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -189,9 +201,9 @@ mesheryctl system context delete [context name]`
 			fmt.Printf("The current context is now %q\n", result)
 			viper.Set("current-context", result)
 		}
-		delete(configuration.Contexts, args[0])
+		delete(configuration.Contexts, contextName)
 		viper.Set("contexts", configuration.Contexts)
-		log.Printf("deleted context %s", args[0])
+		log.Printf("deleted context %s", contextName)
```
lekaf974:
```suggestion
		utils.Log.Infof("deleted context %s", contextName)
```

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -380,15 +392,18 @@ Description: Configures mesheryctl to actively use one one context vs. the anoth
...
+
+		contextName := strings.ToLower(args[0])
+
+		_, exists := configuration.Contexts[contextName]
 		if !exists {
 			const errMsg = `Try running the following to create the context:
 mesheryctl system context create `
 
-			return fmt.Errorf("requested context does not exist \n\n%v%s", errMsg, args[0])
+			return fmt.Errorf("requested context does not exist \n\n%v%s", errMsg, contextName)
```
lekaf974: We can take advantage to move to meshkit error

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -407,9 +422,9 @@ mesheryctl system context create `
 			utils.Log.Info("Meshery is running... switching context without stopping Meshery deployments.")
 		}
 
-		configuration.CurrentContext = args[0]
+		configuration.CurrentContext = contextName
 		viper.Set("current-context", configuration.CurrentContext)
-		log.Printf("switched to context '%s'", args[0])
+		log.Printf("switched to context '%s'", contextName)
```
lekaf974:
```suggestion
		utils.Log.Infof("switched to context '%s'", contextName)
```

mesheryctl/internal/cli/root/system/testdata/check/check.output.golden
DharunMR: thanks for mentioning, it is not intentional. I think some came when i did -update i'll revert those

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -103,12 +106,18 @@ Usage: mesheryctl system context create [context-name]`
...
 		if err != nil {
 			return err
 		}
 
-		log.Printf("Added `%s` context", args[0])
+		if args[0] != strings.ToLower(args[0]) {
+			utils.Log.Infof("Added `%s` context (converted to lowercase)", contextName)
+		} else {
+			utils.Log.Infof("Added `%s` context", contextName)
+		}
```
lekaf974:
```suggestion
    utils.Log.Infof("Added `%s` context ", contextName)
```

mesheryctl/internal/cli/root/system/testdata/check/check.output.golden
lekaf974, line 1: I was in this test yesterday and have the same output before which look weird for a unit test

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -380,15 +392,18 @@ Description: Configures mesheryctl to actively use one one context vs. the anoth
 			utils.Log.Error(ErrUnmarshallConfig(err))
 			return nil
```
PragalvaXFREZ, line 383: The positional argument is lowercased but the --context flag bypasses this completely. A user running `mesheryctl system context view --context MyCtx` might get an error even though `myctx` exists. 

Could you ensure that this is not the case?

mesheryctl/internal/cli/root/system/testdata/reset/TestResetContexts.yaml
PragalvaXFREZ, line 1: i think this should be reverted as well

mesheryctl/internal/cli/root/system/context.go
PragalvaXFREZ:
```suggestion
				utils.Log.Infof("Delete aborted")
```

PragalvaXFREZ, line 131: could you update this in a similar way as context create. Currently `mesheryctl system context delete ctx1 ctx2` silently ignores `ctx2`

DharunMR: Sure ok!

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -157,18 +169,18 @@ mesheryctl system context delete [context name]`
 			if newContext != "" {
 				_, exists := configuration.Contexts[newContext]
 				if !exists {
-					return errors.New("new context wrongly set")
+					return ErrSetCurrentContext(fmt.Errorf("new context wrongly set"))
```
PragalvaXFREZ, line 166: should use meshkit error `ErrSetCurrentContext`

mesheryctl/internal/cli/root/system/context.go
PragalvaXFREZ: +1

mesheryctl/internal/cli/root/system/context.go
DharunMR: Do we even need --context flag? It will work fine with `mesheryctl system context view MyCtx`

DharunMR:
@lekaf974 @PragalvaXFREZ 
- Changed log to utils.log added
- meshkit error for required err 
- One question is do we need --context flag for view? Because view works fine even without flag

mesheryctl/internal/cli/root/system/context.go
```diff
@@ -103,12 +105,15 @@ Usage: mesheryctl system context create [context-name]`
 			tempCntxt.Components = components
 		}
 
-		err := config.AddContextToConfig(args[0], tempCntxt, viper.ConfigFileUsed(), set, false)
+		contextName := strings.ToLower(args[0])
+
+		err := config.AddContextToConfig(contextName, tempCntxt, viper.ConfigFileUsed(), set, false)
```
lekaf974, line 110: should we check if a context name with same value already exists before saving ?

DharunMR: The duplicate check is already handled inside AddContextToConfig, which returns an error if a context with the same name already exists

mesheryctl/internal/cli/root/system/context_test.go
```diff
@@ -218,20 +219,54 @@ func TestAddContextCmd(t *testing.T) {
 	utils.SetupCustomContextEnv(t, currDir+"/testdata/context/ExpectedAdd.yaml")
 	tests := []utils.CmdTestInput{
 		{
-			Name:             "add given context",
-			Args:             []string{"context", "create", "local3"},
-			ExpectedResponse: "createContext.golden",
+			Name:                 "given a valid context name provided when running mesheryctl system context create [valid-name] then a context gets created",
```
lekaf974:
```suggestion
			Name:                 "given context name provided when system context create [valid-name] then context is created",
```

mesheryctl/internal/cli/root/system/context_test.go
```diff
@@ -218,20 +219,54 @@ func TestAddContextCmd(t *testing.T) {
...
+			Args:                 []string{"context", "create", "local3"},
+			ExpectedResponse:     "createContext.golden",
+			ExpectedResponseYaml: "addExpected.golden",
+		},
+		{
+			Name:                 "given a valid context name which contains uppercase letters provided when running mesheryctl system context create [valid-name] then a context gets created",
```
lekaf974:
```suggestion
			Name:                 "given context name which contains uppercase provided when system context create then context is created in lowercase",
```

test should also validate this behavior since it is the purpose of the PR

mesheryctl/internal/cli/root/system/context_test.go
```diff
@@ -218,20 +219,54 @@ func TestAddContextCmd(t *testing.T) {
...
+			Args:                 []string{"context", "create", "Local4"},
+			ExpectedResponse:     "createContext.uppercase.golden",
+			ExpectedResponseYaml: "addExpected.uppercase.golden",
+		},
+		{
+			Name:           "given no valid context name provided when running mesheryctl system context create [valid-name] then an error message displayed",
```
lekaf974:
```suggestion
			Name:           "given no context provided when system context create  then thorw error",
```

mesheryctl/internal/cli/root/system/context_test.go
```diff
@@ -218,20 +219,54 @@ func TestAddContextCmd(t *testing.T) {
...
+			Args:           []string{"context", "create"},
+			ExpectError:    true,
+			ExpectedError:  utils.ErrInvalidArgument(fmt.Errorf("%s\n%s", errArgMsg, contextCreateUsageMsg)),
+			IsOutputGolden: false,
+		},
+		{
+			Name:           "given multiple context name provided when running mesheryctl system context create then an error message displayed",
```
lekaf974:
```suggestion
			Name:           "given multiple context name provided when system context create then throw error",
```

mesheryctl/internal/cli/root/system/context_test.go
```diff
@@ -218,20 +219,54 @@ func TestAddContextCmd(t *testing.T) {
...
+			IsOutputGolden: false,
 		},
 	}
+
 	for _, tt := range tests {
 		t.Run(tt.Name, func(t *testing.T) {
-			utils.SetupMeshkitLoggerTesting(t, false)
-			b := utils.SetupLogrusGrabTesting(t, false)
+			b := utils.SetupMeshkitLoggerTesting(t, false)
+			utils.Log.SetLevel(logrus.InfoLevel)
```
lekaf974, line 326:
```suggestion
			b := utils.SetupMeshkitLoggerTesting(t, true)
```

mesheryctl/internal/cli/root/system/context_test.go
```diff
@@ -255,7 +290,7 @@ func TestAddContextCmd(t *testing.T) {
 				t.Error(err)
 			}
 			actualResponse = string(content)
-			golden = utils.NewGoldenFile(t, "addExpected.golden", testdataDir)
+			golden = utils.NewGoldenFile(t, tt.ExpectedResponseYaml, testdataDir)
```
lekaf974, line 367:
```suggestion
			golden = utils.NewGoldenFile(t, tt.ExpectedResponse, testdataDir)
```
why Yaml suffix ?

DharunMR: it needs 2 golden to validate ,one for cli output and another for verify context got created

Previously it was hardcoded , now i used existing var

mesheryctl/internal/cli/root/system/context_test.go
DharunMR: without this `utils.Log.SetLevel(logrus.InfoLevel)` it also prints debug msg in test so i added it

lekaf974: ok, let's keep it like this for now, I have a WIP that should be helpful to avoid this tweaks

mesheryctl/internal/cli/root/system/context_test.go
```diff
@@ -297,6 +299,27 @@ func TestAddContextCmd(t *testing.T) {
...
+			}
+
+			contexts := data["contexts"].(map[string]interface{})
+
+			if (len(tt.Args) > 2) && (tt.Args[2] != strings.ToLower(tt.Args[2])) {
+				_, existsUpper := contexts[tt.Args[2]]
+				if existsUpper {
+					t.Fatalf("uppercase context should not exist")
+				}
+			}
```
lekaf974, line 393: I don't think we need to validate it for all tests instead a dedicated test should be created

PR #17459, opened 2026-02-16
feat(mesheryctl): implement flags validation for model export command

### LINKED ISSUE #17458: [mesheryctl] improve flags validation implementation

### Current Behavior

flags validation are currently mainly based on hardcoded checks example

```golang
	if !slices.Contains(getValidOutputFormat(), strings.ToLower(modelViewOutputFormat)) {
		const errMsg = "[ yaml, json ] are the only format supported\n\nUsage: mesheryctl model view --output-format [yaml|json]\nRun 'mesheryctl model view --help' to see detailed help message"
		return ErrModelUnsupportedOutputFormat(errMsg)
	}
	return nil
},
```
### Desired Behavior

Try an existing framework to make validation straightforward

### Mockups

---

### Contributor Resources

- ⌨️ [Meshkit errors guide](https://docs.meshery.io/project/contributing/contributing-error)

lekaf974:

**Notes for Reviewers**

- This PR fixes #17458

new behavior

```golang
mevrin@aiolos:~/Repos/github/lekaf974/meshery/meshery/mesheryctl$ ./mesheryctl model export aerospike-kubernetes-operator --version test
Error: Invalid value 'test' for flag --version: version must be in format vX.X.X | Short Description: Invalid flag provided | Probable Cause: The value for the flag provided is invalid. | Suggested Remediation: Provide a valid flag value and try again.

mevrin@aiolos:~/Repos/github/lekaf974/meshery/meshery/mesheryctl$ ./mesheryctl model export aerospike-kubernetes-operator -t csv
Error: Invalid value 'csv' for flag --output-format: valid values are json yaml | Short Description: Invalid flag provided | Probable Cause: The value for the flag provided is invalid. | Suggested Remediation: Provide a valid flag value and try again.

```

### DIFF
--- a/mesheryctl/internal/cli/root/model/export.go
+++ b/mesheryctl/internal/cli/root/model/export.go
@@ -38,7 +38,7 @@ var exportModelCmd = &cobra.Command{
 	Use:   "export",
 	Short: "Export registered models",
 	Long: `Export the registered model to the specified output type
-Documentation for models export can be found at https://docs.meshery.io/reference/mesheryctl/model/export`,
+Find more information at: https://docs.meshery.io/reference/mesheryctl/model/export`,
 	Example: `
 // Export a model by name 
 mesheryctl model export [model-name] -o [oci|tar]  (default is oci)
--- a/docs/meetings/2026/week-of-02-19-2026/Arnav Sharma
+++ b/docs/meetings/2026/week-of-02-19-2026/Arnav Sharma
@@ -1,0 +1,1 @@
+Hey All , I am arnav sharma happy to contributr to Meshery
--- a/docs/meetings/2026/week-of-02-19-2026/Azzam.md
+++ b/docs/meetings/2026/week-of-02-19-2026/Azzam.md
@@ -1,0 +1,10 @@
+Hello! I am Ahmed Azzam, a third-year Computer and Communications Engineering student studying on a fully-funded USAID scholarship.
+My main focus lies in DevOps, Cloud Infrastructure, and declarative automation (Infrastructure as Code).
+I am also a technical writer for *In Plain English*, where I break down complex distributed systems and cloud architecture concepts.
+
+[Medium Articles Link](https://medium.com/@ahmedazzam.business)
+I wrote about:
+1- How OS Theory Powers High-Performance DevOps Systems (Part 1 & 2)
+2- Automating Azure Application Gateway for Containers (AGFC) with AKS, Terraform, Helm, CI/CD
+
+Some of my skills: Docker, Kubernetes, Helm, Terraform, CI/CD pipelines (GitHub Actions), Azure (AKS, Key Vault), AWS
--- a/docs/meetings/2026/week-of-02-19-2026/Farhan_saleem.md
+++ b/docs/meetings/2026/week-of-02-19-2026/Farhan_saleem.md
@@ -1,0 +1,7 @@
+# Farhan Saleem
+
+**GitHub:** [@farhann-saleem](https://github.com/farhann-saleem)
+**Project Interest:** AI Adapter for LFX Mentorship
+**Current Contributions:**
+- **Approved PR:** Fixed error handling across 11 files in server ([PR #17274](https://github.com/meshery/meshery/pull/17274))
+  
--- a/docs/meetings/2026/week-of-02-19-2026/GiGi.md
+++ b/docs/meetings/2026/week-of-02-19-2026/GiGi.md
@@ -1,0 +1,7 @@
+Hi team!
+
+I'm GiGi Koneti from Bengaluru.
+
+I feel Meshery is a very active and supportive community, and I'm excited to make meaningful open-source contributions.
+
+My background is in Machine Learning (ML), and I look forward to applying my skills here.
--- a/docs/meetings/2026/week-of-02-19-2026/Kannan.md
+++ b/docs/meetings/2026/week-of-02-19-2026/Kannan.md
@@ -1,0 +1,5 @@
+Hello,
+
+My name is Kannan Ravindran. I'm a Software Engineer with 3 years of experience, focused on the JavaScript environment. I joined Meshery to start contributing to open source.
+
+You can find me on my [GitHub](https://github.com/kannan-ravi) and [LinkedIn](https://www.linkedin.com/in/kannanravindrankr/) profiles.
--- a/docs/meetings/2026/week-of-02-19-2026/Kavitha_Karunakaran.md
+++ b/docs/meetings/2026/week-of-02-19-2026/Kavitha_Karunakaran.md
@@ -1,0 +1,8 @@
+Hello,
+
+My name is Kavitha Karunakaran. 
+I’m a software engineer with over 15 years of industry experience.
+I joined Meshery to start contributing to open source and have already made a few PRs.
+I’m excited to help with the upcoming CMC Weekly Study Group.
+
+You can find my bio on my [GitHub](https://github.com/matrixkavi) and [StackOverflow](https://stackoverflow.com/users/3981539/kavitha-karunakaran) profiles
--- a/docs/meetings/2026/week-of-02-19-2026/Mahmoud-Shaheen.md
+++ b/docs/meetings/2026/week-of-02-19-2026/Mahmoud-Shaheen.md
@@ -1,0 +1,4 @@
+Hello! 
+I am mahmoud a 3rd year CS student at GIU with main proficiencies in cloud,devops and data science.
+Some of my skills: Python, Matplotlib, azure, aws, gcp, k8, linux, docker, javascript, react.
+Some of my credentials: AWS MLA certified, Azure AZ104, AZ305, AZ400, AZ204, AI102, DP100 and GCP Professional cloud architect
--- a/docs/meetings/2026/week-of-02-19-2026/Nihar.md
+++ b/docs/meetings/2026/week-of-02-19-2026/Nihar.md
@@ -1,0 +1,1 @@
+Present. Final-year CSE student, focused on DevOps and learning CI/CD + Kubernetes.
--- a/docs/meetings/2026/week-of-02-19-2026/Raunak.md
+++ b/docs/meetings/2026/week-of-02-19-2026/Raunak.md
@@ -1,0 +1,1 @@
+Skills, Where you live
--- a/docs/meetings/2026/week-of-02-19-2026/Rodan_Demirali.md
+++ b/docs/meetings/2026/week-of-02-19-2026/Rodan_Demirali.md
@@ -1,0 +1,7 @@
+Hello everyone, I’m Rodan from Istanbul/Turkiye. 
+I'm passionate about open source and interested in contributing to meaningful projects. And that's how I come across with Meshery and very excited to contribute!
+I’m strong in DevOps, CI/CD automation, and the MERN stack.
+
+My current area of ​​interest is developing web applications and maintaining, testing, and debugging existing ones. I also participate in various algorithm competitions. Therefore, algorithms, data structures, and calculus are of great interest to me, and I dedicate a considerable amount of time to them. For this, I regularly solve Leetcode and HackerRank algorithm problems. 
+
+My passion is also developing Javascript/C# web applications using .NET Core MVC. To mention one of my projects that interests me, I have Backendexpress.com, a website I've been working on for 5 years, which contains over 100 applications. I'm still working on it, adding applications that are useful, solve real-life problems, or that people can enjoy and find entertaining.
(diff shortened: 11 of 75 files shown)


lekaf974:
@PragalvaXFREZ and @shri771 this is a WIP and wanted to have your feedback here. I am trying to improve overall flag validation process and found this framework that allow to remove the need of hardcoded validation and use struct validation instead.

This example to see how it could be done, what I like is:

-  we inject validator with a Singleton in one place at initialization in RootCmd and can just reuse it from children command by calling the context.
- Validation is declarative with struct 
- We can centralize and customize error message by flag type

Any toughts ?

PragalvaXFREZ:
> @PragalvaXFREZ and @shri771 this is a WIP and wanted to have your feedback here. I am trying to improve overall flag validation process and found this framework that allow to remove the need of hardcoded validation and use struct validation instead.
> 
> This example to see how it could be done, what I like is:
> 
> * we inject validator with a Singleton in one place at initialization in RootCmd and can just reuse it from children command by calling the context.
> * Validation is declarative with struct
> * We can centralize and customize error message by flag type
> 
> Any toughts ?

Hi @lekaf974, I reviewed the pr and really like the direction. This implementation is much cleaner than manual command-level checks.

One question: right now, if a user passes multiple invalid flags, the CLI reports them one by one across repeated runs. Should we show all validation errors at once so users can fix everything in a single run?


Happy to help with follow-up migrations once the pattern is finalized.

lekaf974:
> One question: right now, if a user passes multiple invalid flags, the CLI reports them one by one across repeated runs. Should we show all validation errors at once so users can fix everything in a single run?
> 

Oh yep initially the framework support multiple errors in one run but it was not readable due to how meshkit error was outputed so I just did a quick hack just to make it more readable but it is not the final version.
I have already an idea to have only one feedback for all errors

lekaf974:
@PragalvaXFREZ think it is good for final review.

mesheryctl/internal/cli/root/model/export.go
```diff
@@ -51,81 +72,59 @@ mesheryctl model export [model-name] --version [version (ex: v0.7.3)]
...
-func export(modelName string, url string, output *outputDetail) error {
-	// Find the entity with the model name
-	req, err := utils.NewRequest(http.MethodGet, url, nil)
-	if err != nil {
-		utils.Log.Error(err)
-		return nil
-	}
+		if exportedModelData == nil {
+			return utils.ErrNotFound(fmt.Errorf("model with name %s and version %s not found", modelName, exportModelFlagsProvided.Version))
+		}
```
shri771: This could be removed, as Model not found case is handled in `MakeRequest`.

mesheryctl/pkg/utils/auth.go
```diff
@@ -84,8 +84,17 @@ func MakeRequest(req *http.Request) (*http.Response, error) {
...
 	}
 
+	// failsafe for data not found on the server
+	if resp.StatusCode == http.StatusNotFound {
+		bodyBytes, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, ErrReadResponseBody(err)
+		}
+		return nil, ErrNotFound(errors.New(string(bodyBytes)))
+	}
```
shri771, line 95: Since this changes the error response for Status Not Found, the golden files for `design` and `workspace` will need to be updated to reflect the new output.

mesheryctl/internal/cli/root/model/export.go
```diff
@@ -43,6 +55,15 @@ mesheryctl model export [model-name] --discard-components --discard-relationship
 // Export a model version by name in YAML type
 mesheryctl model export [model-name] --version [version (ex: v0.7.3)]
 `,
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		validate := cmd.Context().Value("validators").(*validator.Validate)
+		err := validate.Struct(exportModelFlagsProvided)
+		if err != nil {
+			vErr := err.(validator.ValidationErrors)
```
shri771: While `validate.Struct` usually returns `ValidationErrors`, it can technically return `InvalidValidationError` if the struct tags are invalid. To prevent a  panic in the future, we should probably use the type assertion check.

lekaf974:
Will fix linter issue before merge

shri771 (changes requested):
LGTM. Using struct tags makes the validation much cleaner. I’ve left a small suggestion.

mesheryctl/internal/cli/root/model/export.go
PragalvaXFREZ: +1

mesheryctl/internal/cli/root/model/export.go
```diff
@@ -51,81 +72,59 @@ mesheryctl model export [model-name] --version [version (ex: v0.7.3)]
...
-	exportModelCmd.Flags().IntP("page", "p", 1, "(optional) List next set of models with --page (default = 1)")
+	exportModelCmd.Flags().StringVarP(&exportModelFlagsProvided.OutputFormat, "output-format", "t", "yaml", "(optional) format to display in [json|yaml] (default = yaml)")
+	exportModelCmd.Flags().StringVarP(&exportModelFlagsProvided.OutputLocation, "output-location", "l", "./", "(optional) output location (default = current directory)")
+	exportModelCmd.Flags().StringVarP(&exportModelFlagsProvided.OutputType, "output-type", "o", "oci", "(optional) format to display in [oci|tar] (default = oci)")
+	exportModelCmd.Flags().BoolVarP(&exportModelFlagsProvided.DiscardComponents, "discard-components", "c", false, "(optional) whether to discard components in the exported model definition (default = false)")
+	exportModelCmd.Flags().BoolVarP(&exportModelFlagsProvided.DiscardRelationships, "discard-relationships", "r", false, "(optional) whether to discard relationships in the exported model definition (default = false)")
+	exportModelCmd.Flags().StringVarP(&exportModelFlagsProvided.Version, "version", "", "", "(optional) model version to export (default = \"\", format: vX.X.X)")
```
PragalvaXFREZ, line 129: looks like page flag is missing here

mesheryctl/internal/cli/root/model/export.go
```diff
@@ -43,6 +55,15 @@ mesheryctl model export [model-name] --discard-components --discard-relationship
 // Export a model version by name in YAML type
 mesheryctl model export [model-name] --version [version (ex: v0.7.3)]
 `,
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		validate := cmd.Context().Value("validators").(*validator.Validate)
```
PragalvaXFREZ: This is the unguarded type assertion that would panic if the context value is missing.

mesheryctl/internal/cli/root/root.go
```diff
@@ -222,3 +228,11 @@ func setupLogger() {
 	utils.Log = utils.SetupMeshkitLogger("mesheryctl", verbose, os.Stdout)
 	utils.LogError = utils.SetupMeshkitLogger("mesheryctl-error", verbose, os.Stderr)
 }
+
+// Initialize a validator and add it to the command context
+// This allows us to use the same validator instance across all subcommands and avoid initializing multiple instances of the validator
+func initValidators(cmd *cobra.Command) {
+	validate := validator.NewValidator()
+	ctx := context.WithValue(context.Background(), "validators", validate)
```
PragalvaXFREZ: Using a plain string as a context key can collide with other packages. Would it be better to add a small helper in validator with an unexported key type? That also removes the unsafe type assertion here, since the helper can return a clean error if the value is missing instead of panicking.

mesheryctl/pkg/utils/auth.go
```diff
@@ -84,8 +84,17 @@ func MakeRequest(req *http.Request) (*http.Response, error) {
 		return nil, ErrUnauthenticated()
 	}
 
+	// failsafe for data not found on the server
+	if resp.StatusCode == http.StatusNotFound {
+		bodyBytes, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, ErrReadResponseBody(err)
+		}
+		return nil, ErrNotFound(errors.New(string(bodyBytes)))
```
PragalvaXFREZ, line 94: resp.Body is not closed on these early returns. Since we return nil for the response, the caller cannot close it either. We should consider closing it on all error paths, for example defer resp.Body.Close() right after the resp != nil check.

mesheryctl/internal/cli/pkg/flags/validator.go
```diff
@@ -0,0 +1,78 @@
...
+
+func validateSemver(fl validator.FieldLevel) bool {
+	return vSemverRegex.MatchString(fl.Field().String())
+}
+
+func validateBoolean(fl validator.FieldLevel) bool {
+	if val, ok := fl.Field().Interface().(bool); ok {
+		return val == true || val == false
+	}
+	return false
```
PragalvaXFREZ, line 55: What's the intended behavior here? val == true || val == false is always true for any Go bool, and Cobra's   BoolVarP already rejects invalid boolean input before validation runs. Is this meant to validate something else?

mesheryctl/pkg/utils/error.go
```diff
@@ -772,8 +772,8 @@ func ErrCreateFile(filepath string, err error) error {
 	return errors.New(ErrCreateFileCode, errors.Alert,
 		[]string{"Error creating file"},
 		[]string{fmt.Sprintf("Failed to create the file at path: %s", filepath), err.Error()},
-		[]string{"Insufficient disk page, filepath could be invalid."},
-		[]string{"Verify that the file path is valid, and ensure there is sufficient disk space available."})
+		[]string{"Insufficient disk page", " Filepath could be invalid", " Lack of permissions to create file at the specified path"},
```
PragalvaXFREZ: Should "Insufficient disk page" be "Insufficient disk space" ? I was wondering if this is a typo.

mesheryctl/internal/cli/pkg/flags/validator.go
lekaf974: This is mainly due to the validator framework default behavior which is take only `true` as valid boolean value

mesheryctl/internal/cli/root/model/export.go
lekaf974: I was not sure it was useful at first but I rethink about it

mesheryctl/pkg/utils/auth.go
lekaf974: should be resolved

mesheryctl/internal/cli/pkg/flags/validator.go
lekaf974: added a comment to make clear why this has been done

mesheryctl/internal/cli/root/model/export.go
lekaf974: good catch, will rework this part

mesheryctl/internal/cli/root/root.go
```diff
@@ -222,3 +228,11 @@ func setupLogger() {
 	utils.Log = utils.SetupMeshkitLogger("mesheryctl", verbose, os.Stdout)
 	utils.LogError = utils.SetupMeshkitLogger("mesheryctl-error", verbose, os.Stderr)
 }
+
+// Initialize a validator and add it to the command context
+// This allows us to use the same validator instance across all subcommands and avoid initializing multiple instances of the validator
+func initValidators(cmd *cobra.Command) {
+	validate := mesheryctlflags.NewValidator()
+	ctx := context.WithValue(context.Background(), "flags-validator", validate)
```
PragalvaXFREZ: I was checking out the [context.WithValue docs.](https://pkg.go.dev/context#WithValue) It suggests not using built-in types like string as context keys to avoid collisions between packages. The chances of "flags-validator" being overwritten are low, but I'd suggest wrapping it in an unexported named type (e.g. type contextKey string) so the key is unique by type, not just by value.

mesheryctl/internal/cli/root/model/export.go
lekaf974: I try another approach to wrap all in a centralized place

mesheryctl/internal/cli/root/model/export.go
lekaf974: restored the flag for now

mesheryctl/internal/cli/root/root.go
lekaf974: gotcha, updated implementation

mesheryctl/internal/cli/root/design/delete.go
```diff
@@ -52,6 +53,9 @@ mesheryctl design delete [file | URL]
 		if len(args) > 0 {
 			pattern, isID, err = utils.ValidId(mctlCfg.GetBaseMesheryURL(), args[0], "pattern")
 			if err != nil {
+				if errors.GetCode(err) == utils.ErrNotFoundCode {
+					return ErrDesignNotFound(args[0])
+				}
```
shri771, line 58: ❤️

PR #17527, opened 2026-02-20
[mesheryctl] Standardize select prompts and remove infinite prompt loops across commands

### LINKED ISSUE #17522: [mesheryctl] Fix relationship view crash and hang on multi-result selection
**Summary**

1) Panic on missing `EvaluationQuery`: When the CLI prompts a user to select a relationship from a list, it tries to display the `EvaluationQuery`. However, this field is optional, particularly in non-Kubernetes model sets like cloud registries. Currently, the `selectRelationshipPrompt` function in view.go dereferences EvaluationQuery (*string) without a nil check, risking a panic.This doesn't crash with current data (all values are empty strings, not null), but a guard-rail should be added for safety.

2)  The "Trap" Loop: If a user tries to cancel the selection prompt (e.g., hitting Ctrl+C or Esc), the logic currently defaults to continue. This creates an infinite loop where the user is "trapped" in the prompt and cannot exit the command gracefully.

**To reproduce**
`mesheryctl exp relationship view kubernetes` 
instead of selecting the relationship trying exiting via ctrl + c or esc.

PragalvaXFREZ:
**Summary**
 - Guard optional `EvaluationQuery` pointer access in relationship selection to avoid nil dereference.
  - Replace repeated `promptui.Select` retry loops with shared `utils.RunSelectPrompt`.
  - Return and propagate prompt errors from selection helpers instead of looping indefinitely.
  - Apply consistent prompt handling across:
    - `component view`
    - `connection create`
    - `connection view`
    - `design export`
    - `environment view`
    - `model view`
    - `relationship view`
    - provider selection in auth flow

**Notes for Reviewers**

- This PR fixes #17522

**Updated Behavior**
image

### DIFF
--- a/mesheryctl/internal/cli/root/relationships/view.go
+++ b/mesheryctl/internal/cli/root/relationships/view.go
@@ -19,7 +19,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/manifoldco/promptui"
 	"github.com/meshery/meshery/mesheryctl/internal/cli/pkg/api"
 	"github.com/meshery/meshery/mesheryctl/internal/cli/pkg/display"
 	"github.com/meshery/meshery/mesheryctl/pkg/utils"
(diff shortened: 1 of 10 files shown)


mesheryctl/internal/cli/root/relationships/view.go
```diff
@@ -130,14 +136,12 @@ func selectRelationshipPrompt(relationship []relationship.RelationshipDefinition
...
-	for {
-		i, _, err := prompt.Run()
-		if err != nil {
-			continue
-		}
-
-		return &relationship[i]
+	i, _, err := prompt.Run()
+	if err != nil {
+		return nil, fmt.Errorf("relationship selection cancelled: %w", err)
```
lekaf974: Meshkit error

l5io:
🚀 Preview for commit 023c40321a472c832249c4d15388ab49c35ed182 at: https://69987fc0eafad2e5b636990a--meshery-docs-preview.netlify.app

lekaf974 (dismissed):

PragalvaXFREZ:
@lekaf974  I updated this PR by introducing a shared prompt helper and applied the fix across 7 other commands for consistent cancellation/error handling.

PR #17551, opened 2026-02-20
[mesheryctl] Remove misleading [query-text] from relationship search

### LINKED ISSUE #17550: [mesheryctl] exp relationship search documents [query-text] but server doesn't support it
**Description**
`mesheryctl exp relationship search` documents `[query-text]` in its help text, examples, and error messages, but the server-side `RelationshipFilter` in meshkit only supports LIKE matching on the `Kind` field. This makes `[query-text]` a misleading duplicate of `--kind`.

**Expected Behavior**
Remove `[query-text]` from the CLI interface (help, examples, error messages) so the documented interface matches reality. Flags (`--kind`, `--type`, `--subtype`, `--model`) are the correct search interface.

  **To Reproduce**
  1. `mesheryctl exp relationship search --help` : shows `[query-text]` in usage
  2. `mesheryctl exp relationship search network --kind edge` : `network` is silently ignored
  3. `mesheryctl exp relationship search network` : errors because no flag is set, not because query-text isn't supported

PragalvaXFREZ:
**Notes for Reviewers**

- This PR fixes #17550 


mesheryctl/internal/cli/root/relationships/search.go
```diff
@@ -33,17 +33,17 @@ var (
...
+	Long:  "Search registered relationship(s) used by different models",
 	Example: `
-// Search for relationship using a query
-mesheryctl exp relationship search [--kind <kind>] [--type <type>] [--subtype <subtype>] [--model <model>] [query-text]`,
+// Search for a specific relationship
+mesheryctl exp relationship search [--kind <kind>] [--type <type>] [--subtype <subtype>] [--model <model>]`,
 	Args: func(cmd *cobra.Command, args []string) error {
 		const usage = "mesheryctl exp relationship search [--kind <kind>] [--type <type>] [--subtype <subtype>] [--model <model>]"
-		errMsg := fmt.Errorf("[--kind, --subtype or --type or --model] and [query-text] are required\n\nUsage: %s\nRun 'mesheryctl exp relationship search --help'", usage)
+		errMsg := fmt.Errorf("at least one of [--kind, --subtype, --type, --model] is required\n\nUsage: %s\nRun 'mesheryctl exp relationship search --help'", usage)
```
lekaf974: I think we should also refactor the way we are validating flags
- moving flags validation in PreRunE
- use the flag validator added in model export this will enforce UX consistency across commands

PragalvaXFREZ: Implemented. Moved flag validation to `PreRunE` using the `FlagValidator` pattern from `model export`. The validator is currently a no-op since all tags are `omitempty`, but it's wired up to maintain consistency across commands and to enable future field-level constraints.

PR #17608, opened 2026-02-24
[mesheryctl] Added view by ID in model

### LINKED ISSUE #17607: [mesheryctl] Added view by ID in model command

### Current Behavior

mesheryctl model view currently supports viewing a model only by its name.
Although the model ID is displayed in the command output header the CLI does not allow users to query a model using its ID. Attempting to pass a model ID results in no matches being found.

### Desired Behavior

mesheryctl model view should support viewing a model using either `model name` or `model ID`
This enhancement improves usability and aligns with the fact that model IDs are already exposed in the output. Supporting lookup by ID provides a more consistent and flexible CLI experience.

---

### Contributor Resources

- ⌨️ [Meshkit errors guide](https://docs.meshery.io/project/contributing/contributing-error)

DharunMR:
**Notes for Reviewers**

- This PR fixes #17607

Added view by id and test case for it.

### DIFF
--- a/docs/meetings/2026/week-of-03-05-2026/Aditya.md
+++ b/docs/meetings/2026/week-of-03-05-2026/Aditya.md
@@ -1,0 +1,8 @@
+# Aditya Upasani
+
+Hello everyone! 👋
+
+I'm Aditya Upasani, a cloud and DevOps engineer with a strong interest in Kubernetes, cloud-native infrastructure, and platform engineering.
+
+Attending the Meshery Newcomers Meeting  
+Week of 03-05-2026.
--- a/docs/meetings/2026/week-of-03-05-2026/Harshit_Kushwaha.md
+++ b/docs/meetings/2026/week-of-03-05-2026/Harshit_Kushwaha.md
@@ -1,0 +1,6 @@
+Hi, I'm Harshit Kushwaha, a 3rd-year Computer Science Engineering student.
+
+I'm continuously working to expand my skill tree and spend most of my time learning and exploring new technologies.
+
+* **GitHub:** [harshitt13](https://github.com/harshitt13)
+* **Website:** [harshitt13.in](https://harshitt13.in)
--- a/docs/meetings/2026/week-of-03-05-2026/Hrishikesh.md
+++ b/docs/meetings/2026/week-of-03-05-2026/Hrishikesh.md
@@ -1,0 +1,3 @@
+hi, im hrishikesh kulkarni 2nd year student from pune, studying in mit-adt, actively contributing in linux and making linux simple for use of everyone,
+im greatly intrested in meshery project as i am focusing on LFX internship,
+i hope we get along well and i can learn with amazing people
--- a/docs/meetings/2026/week-of-03-05-2026/Kavitha_Karunakaran.md
+++ b/docs/meetings/2026/week-of-03-05-2026/Kavitha_Karunakaran.md
@@ -1,0 +1,8 @@
+Hello,
+
+My name is Kavitha Karunakaran. 
+I’m a software engineer with over 15 years of industry experience.
+I joined Meshery to start contributing to open source and have already made a few PRs.
+I’m excited to help with the CMC Weekly Study Group.
+
+You can find my bio on [GitHub](https://github.com/matrixkavi) and [StackOverflow](https://stackoverflow.com/users/3981539/kavitha-karunakaran)
--- a/docs/meetings/2026/week-of-03-05-2026/Raunak.md
+++ b/docs/meetings/2026/week-of-03-05-2026/Raunak.md
@@ -1,0 +1,1 @@
+Introduction
--- a/docs/meetings/2026/week-of-03-05-2026/Riya Mahajan
+++ b/docs/meetings/2026/week-of-03-05-2026/Riya Mahajan
@@ -1,0 +1,1 @@
+Hi everyone, Riya here. I’m a final-year AI/ML undergraduate interested in LLM systems, AI infrastructure, and open-source development, excited to learn from the community and contribute to Meshery.
--- a/docs/meetings/2026/week-of-03-05-2026/Yash-Mahakal.md
+++ b/docs/meetings/2026/week-of-03-05-2026/Yash-Mahakal.md
@@ -1,0 +1,1 @@
+Hey there my name is yash 
--- a/docs/meetings/2026/week-of-03-05-2026/farhan-saleem.md
+++ b/docs/meetings/2026/week-of-03-05-2026/farhan-saleem.md
@@ -1,0 +1,4 @@
+4th semester student 
+Ai engineer
+open source cloud contributor 
+active contributor in Meshery
--- a/docs/meetings/2026/week-of-03-05-2026/joy_gundu.md
+++ b/docs/meetings/2026/week-of-03-05-2026/joy_gundu.md
@@ -1,0 +1,1 @@
+I attended today's meeting
--- a/docs/meetings/2026/week-of-03-05-2026/lucky.md
+++ b/docs/meetings/2026/week-of-03-05-2026/lucky.md
@@ -1,0 +1,35 @@
+I'm lucky singh (Mann) i'm a first year  Computer Science student with a deep passion for **cloud-native systems**, **distributed architecture**, and **backend engineering**. Over the past half years, I have been systematically studying Go, Kubernetes internals, and the design patterns that power production-grade infrastructure platforms.
+
+My approach to open source is deliberate and methodical:
+
+- I read the entire relevant codebase before opening a single issue
+- I reproduce bugs locally and document them with minimal reproducible examples
+- I submit small, focused PRs that are easy to review, building trust incrementally
+- I respond to reviewer feedback — both from human maintainers and automated tools like Gemini AI review — quickly and professionally
+- I never claim ownership of a task until I have demonstrated familiarity with the code it touches
+
+
+I discovered Meshery while exploring the CNCF landscape and was immediately struck by the quality of its engineering and the warmth of its community. The codebase presented real challenges — not tutorial-level problems — and that drew me in.
+
+what i do till now i make some pr 
+## Pull Request #17663 — Fix Variable Shadowing in `VerifyAndConvertToDesign`
+
+**Link:** https://github.com/meshery/meshery/pull/17663
+
+**Branch:** `fix/variable-shadowing-VerifyAndConvertToDesign`
+
+## Issue Filed — Nil Pointer Dereference in `handleFilterPOST`
+
+**File:** `server/handlers/meshery_filter_handler.go`
+
+**Function:** `handleFilterPOST` (~lines 106–121)
+
+**Bug:** `parsedBody` is declared as a pointer and initialized to `nil`. When `json.Decode` fails, the error handler immediately accesses `parsedBody.FilterData.Name` — dereferencing a nil pointer — causing a **guaranteed server panic / goroutine crash** on any malformed filter POST request.
+
+## Pull Request #17751 — Fix Missing `return` After Error in Catalog Publish/Unpublish Handlers
+
+**Link:** https://github.com/meshery/meshery/pull/17751
+
+**Closes:** Issue #17746
+
+**Problem Solved:** In 4 HTTP handler functions across `meshery_pattern_handler.go` and `meshery_filter_handler.go`, after a `json.Unmarshal` failure, the code called `http.Error(...)` but was **missing a `return` statement** — causing execution to continue into code that accesses `respBody.ContentName` on a nil pointer, resulting in a **guaranteed server panic**.
--- a/docs/meetings/2026/week-of-03-05-2026/taiwo-sholadoye.md
+++ b/docs/meetings/2026/week-of-03-05-2026/taiwo-sholadoye.md
@@ -1,0 +1,2 @@
+Name: Taiwo Sholadoye
+GitHub: https://github.com/hymmns
--- a/docs/meetings/2026/week-of-03-05-2026/thanuja.md
+++ b/docs/meetings/2026/week-of-03-05-2026/thanuja.md
@@ -1,0 +1,1 @@
+Hi, my name is Thanuja Kularathna. I am a third-year student currently pursuing a degree in Information and Communication Technology at Rajarata University of Sri Lanka. 


mesheryctl/internal/cli/root/model/view.go
```diff
@@ -105,6 +115,28 @@ func selectModelPrompt(models []model.ModelDefinition) model.ModelDefinition {
 	}
 }
 
+func getModelViewUrlPath(modelname string) (error, string) {
+	queryParams := url.Values{}
+	var modelsUrlPath string
+
+	// For validate an argument is uuid or name
+	isID, err := regexp.MatchString("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$", modelname)
```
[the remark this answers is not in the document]
PragalvaXFREZ: I think we should use utils.IsUUID here

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -52,7 +59,7 @@ mesheryctl model view [model-name]
 
 		switch modelsResponse.Count {
 		case 0:
-			fmt.Println("No model(s) found for the given name ", modelDefinition)
+			fmt.Println("No model(s) found for the given name ", args[0])
```
PragalvaXFREZ:
```suggestion
			fmt.Println("No model(s) found for the given name or ID ", args[0])
```

PragalvaXFREZ: Also can you ensure that the error.go is updated as well.

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -105,6 +112,28 @@ func selectModelPrompt(models []model.ModelDefinition) model.ModelDefinition {
...
+	isID, err := regexp.MatchString("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$", modelname)
+	if err != nil {
+		return err, ""
+	}
+
+	if !isID {
+		modelsUrlPath = fmt.Sprintf("%s/%s", modelsApiPath, url.PathEscape(modelname))
+	} else {
+		modelsUrlPath = modelsApiPath
+		queryParams.Add("id", url.QueryEscape(modelname))
```
PragalvaXFREZ:
```suggestion
		queryParams.Add("id", modelname)
```
according to my understanding, url.Values handles escaping during Encode()

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -105,6 +111,25 @@ func selectModelPrompt(models []model.ModelDefinition) model.ModelDefinition {
...
 }
 
+func getModelViewUrlPath(modelname string) (error, string) {
+	queryParams := url.Values{}
+	var modelsUrlPath string
+
+	// For validate an argument is uuid or name
+	isID := utils.IsUUID(modelname)
+
+	if !isID {
```
lekaf974:
```suggestion
	if !utils.IsUUID(modelname) {
```

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -105,6 +111,25 @@ func selectModelPrompt(models []model.ModelDefinition) model.ModelDefinition {
...
+		modelsUrlPath = fmt.Sprintf("%s/%s", modelsApiPath, url.PathEscape(modelname))
+	} else {
+		modelsUrlPath = modelsApiPath
+		queryParams.Add("id", modelname)
+	}
+
+	queryParams.Add("pagesize", "all")
```
lekaf974:
```suggestion
       queryParams.Add("pagesize", "all")
		modelsUrlPath = fmt.Sprintf("%s/%s", modelsApiPath, url.PathEscape(modelname))
	} else {
		modelsUrlPath = modelsApiPath
		queryParams.Add("id", modelname)
	}

```
pagesize all does not make sense in case of id

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -105,6 +111,21 @@ func selectModelPrompt(models []model.ModelDefinition) model.ModelDefinition {
 	}
 }
 
+func getModelViewUrlPath(modelname string) (error, string) {
```
lekaf974:
```suggestion
func getModelViewUrlPath(modelNameOrId string) (error, string) {
```

lekaf974: the method return only a string and error is always nil so we can remove it

ps: using go idiomatic when using a method with several values error is always the last one not first

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -52,7 +58,7 @@ mesheryctl model view [model-name]
 
 		switch modelsResponse.Count {
 		case 0:
-			fmt.Println("No model(s) found for the given name ", modelDefinition)
+			fmt.Println("No model(s) found for the given name or ID :", args[0])
```
lekaf974:
```suggestion
			fmt.Sprintf("No model(s) found for the given name or ID : %s\n", args[0])
```

DharunMR: i think Printf, because we dont return any string

lekaf974: Yep goal it to just use string format

lekaf974:
conflicts needs to be resolved

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -19,11 +20,14 @@ var modelViewOutputFormat string
 var viewModelCmd = &cobra.Command{
 	Use:   "view",
 	Short: "View model",
-	Long: `View a model queried by its name
+	Long: `View a model queried by its name or id
```
PragalvaXFREZ:
```suggestion
	Long: `View a model queried by its name or ID
```

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -105,6 +109,21 @@ func selectModelPrompt(models []model.ModelDefinition) model.ModelDefinition {
 	}
 }
 
+func getModelViewUrlPath(modelNameorId string) string {
+	queryParams := url.Values{}
+	var modelsUrlPath string
+
+	if !utils.IsUUID(modelNameorId) {
```
PragalvaXFREZ:
```suggestion
	if !utils.IsUUID(modelNameOrId) {
```

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -105,6 +109,21 @@ func selectModelPrompt(models []model.ModelDefinition) model.ModelDefinition {
...
+		modelsUrlPath = fmt.Sprintf("%s/%s", modelsApiPath, url.PathEscape(modelNameorId))
```
PragalvaXFREZ:
```suggestion
		modelsUrlPath = fmt.Sprintf("%s/%s", modelsApiPath, url.PathEscape(modelNameOrId))
```

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -105,6 +109,21 @@ func selectModelPrompt(models []model.ModelDefinition) model.ModelDefinition {
...
+		queryParams.Add("pagesize", "all")
+	} else {
+		modelsUrlPath = modelsApiPath
+		queryParams.Add("id", modelNameorId)
```
PragalvaXFREZ:
```suggestion
		queryParams.Add("id", modelNameOrId)
```

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -52,7 +56,7 @@ mesheryctl model view [model-name]
 
 		switch modelsResponse.Count {
 		case 0:
-			fmt.Println("No model(s) found for the given name ", modelDefinition)
+			fmt.Printf("No model(s) found for the given name or ID : %s\n", args[0])
```
PragalvaXFREZ:
```suggestion
			fmt.Printf("No model(s) found for the given name or ID: %s\n", args[0])
```

mesheryctl/internal/cli/root/model/testdata/view.model.empty.output.golden
```diff
@@ -1 +1 @@
-No model(s) found for the given name  invalid
+No model(s) found for the given name or ID : invalid
```
PragalvaXFREZ:
```suggestion
No model(s) found for the given name or ID: invalid
```

PragalvaXFREZ (commented):
nits, LGTM otherwise.

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -52,7 +56,7 @@ mesheryctl model view [model-name]
 
 		switch modelsResponse.Count {
 		case 0:
-			fmt.Println("No model(s) found for the given name ", modelDefinition)
+			fmt.Printf("No model(s) found for the given name or ID: %s\n", args[0])
```
lekaf974: just seeing this we could use utils.Log.Infof instead ?

lekaf974: let you update this, good to merge then

mesheryctl/pkg/utils/testing.go
lekaf974, line 1: this change broke other tests

DharunMR: utils.Log is not getting captured in golden files so i changed testing.go ,when i checked in my machine every test passes including other commands but failing in ci?

shri771: @DharunMR, in which test case is there a need to capture log + stout?

DharunMR: in TestViewModel testcase 2 "given an invalid model-name provided when running mesheryctl model view invalid-name then an error message is displayed" after changing print to utils.log it aint capturing in golden file (view.model.empty.output.golden) .So i made changes in testing.go and results fail in ci, but works fine in my machine

DharunMR: @shri771 The issue was that the output being written directly to os.Stdout inside the display package. Since our tests capture cmd.Out(), the output wasn’t being captured and the golden comparison failed.
I fixed it by using .WithOutput(cmd.OutOrStdout()) before calling Display(). So now in tests writes output to buffer and in CLI writes output to stdout as expected. Is this approach right?

shri771: Yes, that makes sense. It’s a good approach.

mesheryctl/internal/cli/root/model/view.go
shri771: We should use `display.PromptAsyncPagination `here to reduce duplicated logic of `selectModel`  while adding pagination support. For reference,  see the implementation of  [model delete](https://github.com/shri771/meshery/blob/d3b6d20ab4329b36c7b4e86a7e87191a77f548f6/mesheryctl/internal/cli/root/model/delete.go)

shri771: @DharunMR, now `PromptAsyncPagination` takes one more field `ErrNotFoundMsg`, and also no need to check the err code is not found. Please sync with master.

shri771: We should now remove the fetch logic, because `PrompAsyncPagination` handles feching of data and also diffrent cases.

shri771 (changes requested):

mesheryctl/internal/cli/root/model/view.go
DharunMR: @shri771 I have a doubt, does different models share same name? I listed through the models and i seen no models have same name. If models are unique PrompAsyncPagination this looks unwanted?

shri771: Yes, model names are unique. However, when a user provides a partial keyword like aws, multiple models match, so the selection prompt lets them pick the right one.

DharunMR:
@shri771 seen this [PR](https://github.com/meshery/meshery/pull/17737) ,now we are logging the not found msg or returning error?

lekaf974:
> @shri771 seen this [PR](https://github.com/meshery/meshery/pull/17737) ,now we are logging the not found msg or returning error?

not yet merged

lekaf974:
@DharunMR it is your reponsability to ensure follow up on our opened PRs by (re-)request review or post in slack. We are doing many stuff in parallel and it can be forgotten.

DharunMR:
@lekaf974
I’ll make sure to follow up on my open PRs by re-requesting reviews or posting in Slack. And thanks as well for all your reviews over time.

lekaf974:
@PragalvaXFREZ @shri771 PR merged since you request changes, could you retake a look.

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -81,19 +80,19 @@ func getValidOutputFormat() []string {
...
+func getModelViewUrlPath(modelNameOrId string) string {
+	queryParams := url.Values{}
+	var modelsUrlPath string
 
-	i, err := utils.RunSelectPrompt("Select a model", modelNames)
-	if err != nil {
-		return model.ModelDefinition{}, err
+	if !utils.IsUUID(modelNameOrId) {
+		modelsUrlPath = fmt.Sprintf("%s/%s", modelsApiPath, url.PathEscape(modelNameOrId))
+		queryParams.Add("pagesize", "all")
```
PragalvaXFREZ: I have a slight confusion here, why do we force pagesize=all here for name-based model view?

model delete uses paged prompt flow (UrlPath: modelsApiPath + SearchTerm), but this path is all-at-once. For names with many matches, prompt UX can show Load More..... and then end with no more items available.

Can we align this with model delete pagination behavior?

shri771: +1

mesheryctl/internal/cli/root/model/view.go
```diff
@@ -39,36 +42,32 @@ mesheryctl model view [model-name]
 		return nil
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
-		modelDefinition := args[0]
-
-		modelsResponse, err := api.Fetch[models.MeshmodelsAPIResponse](fmt.Sprintf("%s/%s?pagesize=all", modelsApiPath, modelDefinition))
-
+		urlPath := getModelViewUrlPath(args[0])
```
shri771: Please store the arg in a varibale, for readability or reusability.

mesheryctl/internal/cli/root/model/view.go
DharunMR: I removed pagesize=all since it seemed unnecessary and could bypass the pagination handled by PromptAsyncPagination.
We cannot fully align this with delete.go because model view supports querying both by model name and model ID, which require different query parameters. For name-based queries I switched to using the search parameter so pagination can work consistently, while UUID inputs use the id filter.

PragalvaXFREZ (approved):
LGTM

PR #17642, opened 2026-02-25
Chore: RTk clients to meshery/schemas

saurabhraghuvanshii:
**Notes for Reviewers**
The UI RTK setup now uses the generated Meshery API from meshery/schemas.

- This PR fixes #


go.mod
lekaf974, line 1: why go.mod and go.sum changes push in a PR related to UI ?

saurabhraghuvanshii: @lekaf974, when I run make server, it automatically modifies go.mod and go.sum.
I already mentioned this in the PR description, and I encountered the same issue on another branch as well.

lekaf974: from what I am seeing it is only reordering, gonna make a PR for it

saurabhraghuvanshii: Reverted, Thanks @lekaf974

lekaf974: #17650

ui/rtk-query/environments.ts
YASHMAHAKAL, line 1: @saurabhraghuvanshii Looks like splitting getEnvironments and createEnvironment onto a separate rtk instance means their cache tags can;t talk to the rest of the environment endpoints, so mutations won't trigger the expected refetches. Same thing might apply to teams and user keys as well. What do you think, or am I missing something?

saurabhraghuvanshii: You're right. Thanks for pointing out!!

YASHMAHAKAL (approved):
@saurabhraghuvanshii Looks clean to me now : )

PR #17654, opened 2026-02-26
Fix: missing return after http.Error in catalog filter handlers to prevent nil pointer dereference panic

### LINKED ISSUE #17648: [Bug] Missing `return` After Error Response in PublishCatalogFilterHandler and UnPublishCatalogFilterHandler Causes Nil Pointer Dereference Panic
### Current Behavior

In `server/handlers/meshery_filter_handler.go`, both `PublishCatalogFilterHandler` and `UnPublishCatalogFilterHandler` are missing a `return` statement after calling `http.Error` during JSON response unmarshalling failure. This causes execution to fall through into subsequent code that dereferences `respBody`, which is `nil` when unmarshalling has failed — resulting in a nil pointer dereference panic that crashes the server.

**Affected Functions:**

| Function | Lines |
|---|---|
| `PublishCatalogFilterHandler` | 443–455 |
| `UnPublishCatalogFilterHandler` | 520–532 |

### Expected Behavior

When `json.Unmarshal(resp, &respBody)` fails inside either handler, the function should immediately return after calling `http.Error` to prevent accessing `respBody.ContentName` or `respBody.Status` when `respBody` is nil.

### Screenshots/Logs

**PublishCatalogFilterHandler — Buggy Code (Lines 443–455):**

The error block at line 452 calls `http.Error` but has no `return`, so execution continues to line 455 which accesses `respBody.ContentName` causing a nil pointer dereference panic.

**UnPublishCatalogFilterHandler — same pattern at Lines 520–532.**

Expected panic:
```
panic: runtime error: invalid memory address or nil pointer dereference
```

### Environment

- **Meshery Server Version:** latest master
- **Meshery Client Version:** N/A (server-side bug)

### To Reproduce

1. Run the Meshery server locally
2. Trigger a malformed provider response for PublishCatalogFilter or UnPublishCatalogFilter
3. Send POST to /api/filter/catalog/publish or DELETE to /api/filter/catalog/unpublish
4. Server panics with nil pointer dereference

### Suggested Fix

Add `return` after `http.Error` in both functions (line 452 and line 529).

---

- [Meshery Build and Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)

luckyxhq:
## Summary

Fixes nil pointer dereference panic by adding missing `return` statements after `http.Error` calls in `PublishCatalogFilterHandler` and `UnPublishCatalogFilterHandler` in `server/handlers/meshery_filter_handler.go`.

Closes #17648

## Root Cause

Both `PublishCatalogFilterHandler` and `UnPublishCatalogFilterHandler` call `json.Unmarshal` to parse the provider response into `respBody`. When this fails, the error block calls `http.Error` to respond with a 500 status — but was **missing a `return` statement**. Without `return`, execution falls through to the next line which accesses `respBody.ContentName` and `respBody.Status`. Since `respBody` remains `nil` after a failed unmarshal, this results in a **nil pointer dereference panic** that crashes the server.

## Solution

Added `return` immediately after `http.Error` in both error blocks:

- `PublishCatalogFilterHandler` — after line 452
- `UnPublishCatalogFilterHandler` — after line 529

This is consistent with every other error path in the same file (all other `if err != nil` blocks correctly use `return` after sending an error response).

## Files Changed

| File | Change |
|------|--------|
| `server/handlers/meshery_filter_handler.go` | +2 lines (`return` added in 2 error blocks) |

## Testing Performed

- Verified the diff is syntactically correct Go (adding a bare `return` inside an `if err != nil` block).
- Manually traced both execution paths to confirm `respBody` is no longer accessed when `nil`.
- No existing handler unit tests exist for these functions, so no regressions are introduced.
- The fix follows the existing coding convention in the file — every other error handler block already uses `return`.

## Assumptions / Limitations

- This is the minimal correct fix; no new functionality is added.
- No new tests are added since the handler test infrastructure in this package is currently fully commented out. Adding proper unit tests for these handlers would be a worthwhile follow-up.

### DIFF
--- a/docs/meetings/2026/week-of-02-26-2026/Raunak.md
+++ b/docs/meetings/2026/week-of-02-26-2026/Raunak.md
@@ -1,0 +1,1 @@
+Yada yada 
--- a/docs/meetings/2026/week-of-02-26-2026/Suhani_Garg.md
+++ b/docs/meetings/2026/week-of-02-26-2026/Suhani_Garg.md
@@ -1,0 +1,3 @@
+Hi everyone! 
+
+MySelf Suhani Garg, and I’m a B.Tech Computer Science student specializing in Software Product Engineering. I’m currently exploring cloud-native technologies and open-source contributions, and I’m really excited to be part of the Meshery community.


luckyxhq:
Thanks for the review! I've amended the commit to include the required DCO `Signed-off-by` sign-off and force-pushed the branch. The DCO check should now pass. Please let me know if there's anything else to address. 🙏

luckyxhq:
Thanks for the review! I've addressed the feedback, ensured DCO compliance, and all checks are now passing. I've also updated the branch with the latest changes from the base branch.

Please let me know if anything else is needed — happy to make further improvements. @matrixkavi

matrixkavi:
Unfortunately, I do not have merging rights for this repo.

PR #17662, opened 2026-02-27
fix(pattern): resolve variable shadowing bug causing silent design import failure

### LINKED ISSUE #17661: [Bug] Shadowed Variable Causes Silent Design Import Failure in `VerifyAndConvertToDesign`
### Summary

A **variable shadowing bug** in the `VerifyAndConvertToDesign` function (`server/handlers/meshery_pattern_handler.go`) causes fetched design source content to be silently discarded. When a design of non-Design type (e.g., Helm chart, K8s manifest, Docker Compose) has no in-memory source content and the code fetches it remotely, the result is assigned to a **new local variable** via `:=` instead of updating the outer variable. This means the conversion always proceeds with an **empty byte slice**, producing a silent failure or an empty/corrupt design file.

---

### Current Behavior

When a design pattern requires source content to be fetched from the remote provider (because `mesheryPattern.SourceContent` is empty), the fetched content is **silently dropped** due to variable shadowing. The `FileToImport` struct receives an empty `Data` field instead of the actual file bytes, causing:

- Silent conversion failure (no error is returned)
- The exported/downloaded design file is empty or malformed
- No error is surfaced to the user — the download appears to succeed but delivers a broken design

---

### Expected Behavior

The fetched `sourceContent` should be used for the design conversion. The `FileToImport.Data` field must receive the bytes returned from `provider.GetDesignSourceContent(...)`.

---

### Steps to Reproduce

1. Save a design of type **Helm Chart**, **K8s Manifest**, or **Docker Compose** (i.e., `Type != "MesheryDesign"`) using a remote provider
2. Ensure `SourceContent` is **not** cached in memory (e.g., after a server restart)
3. Call the **download endpoint** for that pattern: `GET /api/pattern/download/{id}`
4. Observe that the downloaded file is empty or produces an import error when re-imported

---

### Root Cause

**File:** `server/handlers/meshery_pattern_handler.go`
**Function:** `VerifyAndConvertToDesign`
**Lines:** ~273–285

```go
// BUGGY CODE:
sourceContent := mesheryPattern.SourceContent  // outer variable (empty []byte)

if len(mesheryPattern.SourceContent) == 0 {
    h.log.Info("Pattern file doesn't contain SourceContent, fetching from remote provider")
    sourceContent, err := provider.GetDesignSourceContent(token, mesheryPattern.ID.String())
    //        ↑ := creates a NEW local variable that shadows the outer `sourceContent`
    if err != nil {
        return ErrDesignSourceContent(err, "get ")
    }
    mesheryPattern.SourceContent = sourceContent
}

fileToImport := FileToImport{
    Data:     sourceContent,  // ← uses OUTER (still empty) variable, not the fetched one
    FileName: mesheryPattern.Name,
}
```

The `:=` on the line `sourceContent, err := provider.GetDesignSourceContent(...)` declares a **brand new local** `sourceContent` that is scoped to the `if` block. The outer `sourceContent` (used by `FileToImport`) is never updated and remains an empty `[]byte{}`.

---

### Suggested Fix

Change `:=` to `=` on the assignment, and declare `err` separately:

```go
sourceContent := mesheryPattern.SourceContent

if len(mesheryPattern.SourceContent) == 0 {
    h.log.Info("Pattern file doesn't contain SourceContent, fetching from remote provider")
    var err error
    sourceContent, err = provider.GetDesignSourceContent(token, mesheryPattern.ID.String())
    //             ↑ = (assignment) instead of := (short variable declaration)
    if err != nil {
        return ErrDesignSourceContent(err, "get ")
    }
    mesheryPattern.SourceContent = sourceContent
}

fileToImport := FileToImport{
    Data:     sourceContent,  // now correctly uses the fetched content
    FileName: mesheryPattern.Name,
}
```

---

### Impact

| Attribute | Value |
|---|---|
| **Severity** | Medium |
| **Area** | Design Import / Export |
| **Affected Providers** | Remote Provider |
| **Affected Design Types** | Helm Chart, K8s Manifest, Docker Compose |
| **User Impact** | Silent data loss — downloaded designs are empty or corrupt |

---

### Environment

- **Component:** Meshery Server (`server/handlers/meshery_pattern_handler.go`)
- **Function:** `VerifyAndConvertToDesign`
- **Language:** Go
- **Type of bug:** Variable shadowing (`:=` vs `=`)

---

### Additional Context

This bug is detectable with Go's static analysis tools:

```bash
go vet ./server/handlers/...
# or
staticcheck ./server/handlers/...
```

Adding a unit test that:
1. Creates a `MesheryPattern` with `SourceContent = nil` and `Type = HelmChart`
2. Mocks `provider.GetDesignSourceContent` to return known bytes
3. Calls `VerifyAndConvertToDesign`
4. Asserts `FileToImport.Data` equals the mocked bytes

...would catch this regression permanently.

---

🚨 DO NOT EDIT BELOW THIS LINE 🚨

---
*This issue was identified through static analysis of the Meshery server codebase.*
*File: `server/handlers/meshery_pattern_handler.go` | Function: `VerifyAndConvertToDesign` | Lines: ~273–285*

MayankRaj435:
### Description
This PR fixes issue #17661 where importing non-Design patterns (such as Helm charts, Kubernetes manifests, or Docker Compose files) would silently fail. 

The bug was caused by variable shadowing in [server/handlers/meshery_pattern_handler.go](cci:7://file:///d:/meshery/meshery/server/handlers/meshery_pattern_handler.go:0:0-0:0) ([VerifyAndConvertToDesign](cci:1://file:///d:/meshery/meshery/server/handlers/meshery_pattern_handler.go:258:0-312:1)). When fetching source content remotely, the `:=` operator was mistakenly used inside the [if](cci:1://file:///d:/meshery/meshery/server/models/pattern/core/pattern.go:26:0-32:1) block, preventing the outer `sourceContent` variable from being populated. Consequently, an empty byte slice was passed to the conversion function, resulting in corrupted or empty patterned files without any error being surfaced to the user.

**Changes made:**
- Changed `:=` to `=` when assigning `sourceContent` to properly update the outer scope variable.
- Added a new unit test [TestVerifyAndConvertToDesign_VariableShadowing](cci:1://file:///d:/meshery/meshery/server/handlers/meshery_pattern_handler_shadowing_test.go:35:0-73:1) in [server/handlers/meshery_pattern_handler_shadowing_test.go](cci:7://file:///d:/meshery/meshery/server/handlers/meshery_pattern_handler_shadowing_test.go:0:0-0:0) that mocks the provider to ensure source content is correctly populated and regressions are prevented.

### Notes for Reviewers
- This PR fixes #17661
- A regression test using a mock [Provider](cci:2://file:///d:/meshery/meshery/server/models/providers.go:353:0-514:1) was included to validate the fix.

### Signed commits

### DIFF
--- a/server/handlers/meshery_pattern_handler.go
+++ b/server/handlers/meshery_pattern_handler.go
@@ -272,7 +272,8 @@ func (h *Handler) VerifyAndConvertToDesign(
 		sourceContent := mesheryPattern.SourceContent
 		if len(mesheryPattern.SourceContent) == 0 {
 			h.log.Info("Pattern file doesn't contain SourceContent, fetching from remote provider")
-			sourceContent, err := provider.GetDesignSourceContent(token, mesheryPattern.ID.String())
+			var err error
+			sourceContent, err = provider.GetDesignSourceContent(token, mesheryPattern.ID.String())
 			if err != nil {
 				return ErrDesignSourceContent(err, "get ")
 			}
--- a/server/handlers/meshery_pattern_handler_shadowing_test.go
+++ b/server/handlers/meshery_pattern_handler_shadowing_test.go
@@ -1,0 +1,74 @@
+package handlers
+
+import (
+	"context"
+	"testing"
+	"bytes"
+
+	"github.com/gofrs/uuid"
+	"github.com/meshery/meshery/server/models"
+	coreV1 "github.com/meshery/schemas/models/v1alpha1/core"
+	"github.com/meshery/meshkit/logger"
+	"database/sql"
+)
+
+// shadowingMockProvider implementation for testing VerifyAndConvertToDesign
+type shadowingMockProvider struct {
+	models.Provider // Embed the interface so we don't have to implement all methods
+	
+	fetchedSourceContent []byte
+	fetchError           error
+	
+	savedPattern         *models.MesheryPattern
+	saveError            error
+	saveResp             []byte
+}
+
+func (m *shadowingMockProvider) GetDesignSourceContent(token, patternID string) ([]byte, error) {
+	return m.fetchedSourceContent, m.fetchError
+}
+
+func (m *shadowingMockProvider) SaveMesheryPattern(token string, pattern *models.MesheryPattern) ([]byte, error) {
+	m.savedPattern = pattern
+	return m.saveResp, m.saveError
+}
+
+func TestVerifyAndConvertToDesign_VariableShadowing(t *testing.T) {
+	log, _ := logger.New("test", logger.Options{})
+	
+	h := &Handler{
+		log: log,
+		// registryManager can be nil if ConvertFileToDesign doesn't crash, 
+		// but ConvertFileToDesign requires a valid filename with extension.
+	}
+	
+	patternID, _ := uuid.NewV4()
+	pattern := &models.MesheryPattern{
+		ID:   &patternID,
+		Name: "test-helm-chart.tgz", // needs extension for conversion to work without error
+		Type: sql.NullString{String: string(coreV1.HelmChart), Valid: true},
+		SourceContent: []byte{}, // empty initially
+		PatternFile: "", // empty initially
+	}
+	
+	expectedBytes := []byte("mocked helm chart content")
+	
+	provider := &shadowingMockProvider{
+		fetchedSourceContent: expectedBytes,
+		saveResp: []byte(`[]`), // mock successful json response from save
+	}
+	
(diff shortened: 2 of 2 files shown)


matrixkavi:
Please start raising PRs only after issue is assigned. Also, considering the review queue, it would be great if you could help with reviewing others codes too. We need to maintain some etiquettes in this regard as we are having only a handful of maintainers looking at the code and it becomes extremely difficult to bring down the review queue if multiple PRs are randomly generated at will to fix the same issue. Since this PR was raised before assigning the issue, we are closing it.

The DCO is also failing

PR #17697, opened 2026-02-28
fix: k8s icon rendering issue

saurabhraghuvanshii:
**Notes for Reviewers**
Issue : 
- MeshKit has a MaxDepth limit when dereferencing these schemas to prevent infinite recursion and keep response sizes manageable.
- Unresolved $ref references remain in the schema - When the depth limit is hit, some $ref pointers are left in the schema pointing to definitions like #/components/schemas/io.k8s.api.core.v1.Affinity. 

The fix in component_handler.go processes the schema before sending it to the frontend.

- This PR fixes #


YASHMAHAKAL:
@saurabhraghuvanshii @rishiraj38 
Screenshot from 2026-03-03 12-46-45

PR #17706, opened 2026-02-28
Fix broken Discourse badge URL in README

farhann-saleem:
The Discourse users badge in README was pointing to `meshery.io/community` as the Discourse server URL, which is not a valid Discourse instance and returns a "not found" badge. 
  Updated it to point to `discuss.layer5.io` which is the actual Discourse forum.    

before :
Screenshot from 2026-02-28 22-32-22

after(my markdown previewer in VS is in light mode :) :
Screenshot from 2026-02-28 22-42-22

cc: @saurabhraghuvanshii


saurabhraghuvanshii:
@farhann-saleem It’s good, we don’t need to add direct Discourse links in Meshery repos. If you find a broken link in Layer5 repos, you can raise a PR there. Thanks!!

PR #17719, opened 2026-03-01
fix: prevent nil pointer panic in handleFilterPOST when request body …

### LINKED ISSUE #17718: [Bug] Nil Pointer Dereference / Server Panic in handleFilterPOST when request body is malformed
### Summary

A **nil pointer dereference** exists in `server/handlers/meshery_filter_handler.go` inside the `handleFilterPOST` function. When the HTTP request body fails to decode (e.g., malformed JSON is sent), the error-handling block immediately accesses `parsedBody.FilterData.Name` — but `parsedBody` is still `nil` at this point because the decode failed. This causes a **server-side panic** that crashes the goroutine handling the request instead of returning a proper `400 Bad Request` to the client.

---

### Current Behavior

When a `POST /api/filter` request is made with a malformed or empty JSON body:
1. `json.NewDecoder(r.Body).Decode(&parsedBody)` returns an error.
2. The error handler tries to build an event description using `parsedBody.FilterData.Name`.
3. Since decoding failed, `parsedBody` is `nil` — accessing `.FilterData.Name` on a nil pointer causes a **runtime panic**.
4. The server goroutine crashes; the client receives no meaningful HTTP error response.

---

### Expected Behavior

When an invalid or malformed request body is sent to `POST /api/filter`:
- The server should **not panic**.
- The server should return `HTTP 400 Bad Request` with a useful error message.
- The error event should be persisted and broadcast **without** dereferencing the nil `parsedBody`.

---

### Steps to Reproduce

1. Start the Meshery server locally.
2. Send a POST request to `/api/filter` with invalid JSON:

```bash
curl -X POST http://localhost:9081/api/filter \
  -H "Content-Type: application/json" \
  -d "THIS IS NOT VALID JSON"
```

3. Check the server logs — you will see a panic, and the client will get no proper error response.

---

### Root Cause Analysis

**File:** `server/handlers/meshery_filter_handler.go`
**Function:** `handleFilterPOST`
**Lines:** 103–121

`parsedBody` is declared as a pointer (`*models.MesheryFilterRequestBody`), which defaults to `nil`. When `json.Decode` fails, `parsedBody` is never assigned a value and remains `nil`. The error handler then accesses `parsedBody.FilterData.Name` — dereferencing a nil pointer — causing a fatal runtime panic.

---

### Buggy Code

```go
// server/handlers/meshery_filter_handler.go — lines 103–121
var parsedBody *models.MesheryFilterRequestBody   // parsedBody is nil here

if err := json.NewDecoder(r.Body).Decode(&parsedBody); err != nil {
    invalidReqBody := ErrRequestBody(err)
    h.log.Error(invalidReqBody)

    event := eventBuilder.WithSeverity(events.Error).WithMetadata(map[string]interface{}{
        "error": invalidReqBody,
    //  BUG: parsedBody is still nil — this line causes a runtime panic!
    }).WithDescription(fmt.Sprintf("Filter %s is corrupted.", parsedBody.FilterData.Name)).Build()

    _ = provider.PersistEvent(*event, nil)
    go h.config.EventBroadcaster.Publish(userID, event)
    http.Error(rw, ErrSaveFilter(err).Error(), http.StatusBadRequest)
    addMeshkitErr(&res, ErrGetFilter(err))
    go h.EventsBuffer.Publish(&res)
    return
}
```

---

### Proposed Fix

Replace the unsafe nil dereference with a safe static description string:

```go
// FIXED CODE
var parsedBody *models.MesheryFilterRequestBody

if err := json.NewDecoder(r.Body).Decode(&parsedBody); err != nil {
    invalidReqBody := ErrRequestBody(err)
    h.log.Error(invalidReqBody)

    //  Safe: parsedBody may be nil on decode failure — do NOT access its fields
    event := eventBuilder.WithSeverity(events.Error).WithMetadata(map[string]interface{}{
        "error": invalidReqBody,
    }).WithDescription("Filter request body is corrupted or malformed.").Build()

    _ = provider.PersistEvent(*event, nil)
    go h.config.EventBroadcaster.Publish(userID, event)
    http.Error(rw, ErrSaveFilter(err).Error(), http.StatusBadRequest)
    addMeshkitErr(&res, ErrGetFilter(err))
    go h.EventsBuffer.Publish(&res)
    return
}
```

---

### Diff

```diff
     event := eventBuilder.WithSeverity(events.Error).WithMetadata(map[string]interface{}{
         "error": invalidReqBody,
-    }).WithDescription(fmt.Sprintf("Filter %s is corrupted.", parsedBody.FilterData.Name)).Build()
+    }).WithDescription("Filter request body is corrupted or malformed.").Build()
```

---

### Affected File

| File | Function | Lines |
|------|----------|-------|
| `server/handlers/meshery_filter_handler.go` | `handleFilterPOST` | 112 |

---

### Environment

- **Meshery Server Version:** master branch (all versions with `handleFilterPOST`)
- **Go Version:** 1.21+
- **Platform:** Any (Docker / Kubernetes / Local)
- **Host OS:** Mac, Linux, Windows

---

### Severity

| Field | Value |
|-------|-------|
| **Type** | Runtime Panic — Nil Pointer Dereference |
| **Impact** | Server goroutine crashes on every malformed filter POST request |
| **Exploitability** | Any client sending bad JSON to `POST /api/filter` triggers the panic |
| **Priority** | High |

---

### Additional Notes

- This is a **one-line fix** — just replace the unsafe `parsedBody.FilterData.Name` access with a static string.
- The fix preserves all existing error-handling behavior: correct HTTP 400 status, event logging, and buffer publishing.
- Adding a unit test that sends invalid JSON to `POST /api/filter` would prevent regressions.

---

### Contributor Resources
- 📚 [Contributing to Documentation](https://docs.meshery.io/project/contributing/contributing-docs)
- 🙋 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)

luckyxhq:
## Description

Fixes a **nil pointer dereference panic** in `handleFilterPOST` (`meshery_filter_handler.go`).

Closes #17718

## What was wrong

When a malformed JSON body is sent to `POST /api/filter`, `parsedBody` stays `nil` after a failed decode. The code then accessed `parsedBody.FilterData.Name`, which caused a **server panic**.

## What changed

- Added `parsedBody == nil` check alongside the decode error check
- Set a clear fallback error (`"request body cannot be null"`) when body is nil
- Used the correct error function `ErrSaveFilter` instead of `ErrGetFilter`

## Diff

```diff
- if err := json.NewDecoder(r.Body).Decode(&parsedBody); err != nil {
+ if err := json.NewDecoder(r.Body).Decode(&parsedBody); err != nil || parsedBody == nil {
+   if err == nil {
+     err = fmt.Errorf("request body cannot be null")
+   }
```


luckyxhq:
All checks are passing ,Could you please review when you have time? Happy to make any further changes if needed.
@saurabhraghuvanshii , Thanks.

matrixkavi:
@luckyxhq 
While you are at this, please do a favor to the reviewers by ensuring that you are not copy pasting the issue description also along with the LLM output you are using. This is just an advice out of good will. For a one line change, having a 2 page description doesn’t make sense. In the long run, blind copy-pasting will only hurt online reputation.

luckyxhq:
Thanks for the feedback! That makes sense. I’ll make sure to keep PR descriptions concise and write them in my own words going ,i use big description for the reviewer so i can explain it , forward. I’ll update this PR accordingly. @matrixkavi

luckyxhq:
Updated the PR description to be more concise and in my own words. Please take another look. Thanks!
@matrixkavi

server/handlers/meshery_filter_handler.go
```diff
@@ -103,19 +103,23 @@ func (h *Handler) handleFilterPOST(
 	var parsedBody *models.MesheryFilterRequestBody
 
 	actedUpon := &userID
-	if err := json.NewDecoder(r.Body).Decode(&parsedBody); err != nil {
+	if err := json.NewDecoder(r.Body).Decode(&parsedBody); err != nil || parsedBody == nil {
+		if err == nil {
```
matrixkavi: This should be 
> if parsedBody == nil

matrixkavi (changes requested):
Left review comments

matrixkavi:
Thank you for raising this. If you are finding more instances like this, feel free to add it to the same PR as it is addressing nil pointer panic. If there are too many instances of such issues of not validating pointers before accessing the values pointed to, consider splitting it to multiple PRs.

luckyxhq:
Thanks @matrixkavi ,for the guidance! I’ll review the codebase for similar nil pointer validation gaps and include related fixes where it keeps the scope focused. If the changes grow beyond a reasonable size, I split them into separate PRs to keep things easier to review.

luckyxhq:
that thing that i wrote 
if err := json.NewDecoder(r.Body).Decode(&parsedBody); err != nil || parsedBody == nil

Correct approach (clean separation)
if err != nil {
    // handle decode error
}

if parsedBody == nil {
    // handle nil body case
}

server/handlers/meshery_filter_handler.go
```diff
@@ -103,19 +103,40 @@ func (h *Handler) handleFilterPOST(
...
+		}).WithDescription("Filter request body is corrupted or malformed.").Build()
+
+		_ = provider.PersistEvent(*event, nil)
+		go h.config.EventBroadcaster.Publish(userID, event)
+
+		http.Error(rw, ErrSaveFilter(err).Error(), http.StatusBadRequest)
+		addMeshkitErr(&res, ErrSaveFilter(err))
+		go h.EventsBuffer.Publish(&res)
+		return
+	}
```
matrixkavi, line 128:
```suggestion

	if err := json.NewDecoder(r.Body).Decode(&parsedBody); err != nil {
		invalidReqBody := ErrRequestBody(err)
		h.log.Error(invalidReqBody)

		desc := "Filter is corrupted."
		if parsedBody != nil && parsedBody.FilterData != nil && parsedBody.FilterData.Name != "" {
			desc = fmt.Sprintf("Filter %s is corrupted.", parsedBody.FilterData.Name)
		}

		event := eventBuilder.WithSeverity(events.Error).WithMetadata(map[string]interface{}{
			"error": invalidReqBody,
		}).WithDescription(desc).Build()

		_ = provider.PersistEvent(*event, nil)
		go h.config.EventBroadcaster.Publish(userID, event)

		http.Error(rw, ErrRequestBody(err).Error(), http.StatusBadRequest)
		addMeshkitErr(&res, ErrRequestBody(err))
		go h.EventsBuffer.Publish(&res)
		return
	}
```
I was thinking of this approach. In existing code, we assume that the` parsedBody.FilterData.Name `to be always available while appending the event description. If we get nil values for `parsedBody` after decoding, then this can lead to issues at runtime. Anyway, we are checking and setting `actedUpon` in similar fashion in code below.

matrixkavi (changes requested):

PR #17724, opened 2026-03-02
[Server] Fix nil pointer dereference in parseRelationshipToAlias

### LINKED ISSUE #17721: [Bug] Potential panic due to out-of-bounds array access in parseRelationshipToAlias
**Overview:**
A deep static analysis of the codebase reveals a potential panic vulnerability in `parseRelationshipToAlias` during the parsing of relationship definitions. When checking `MutatedRef`, the code accesses the first element (`mutatedRefs[0]`) without first verifying that the array is non-empty. If a relationship is defined with an empty `MutatedRef` array, this will trigger a runtime panic, potentially crashing the server or the policy evaluation worker.

**Location:**
[`server/handlers/policy_relationship_handler.go`](https://github.com/meshery/meshery/blob/master/server/handlers/policy_relationship_handler.go)
Line: 75

```go
63:     if from.Patch == nil || from.Patch.MutatedRef == nil {
64:         return alias, false
65:     }
66:     mutatedRefs := *from.Patch.MutatedRef
67: 
68:     alias.ImmediateParentId = *to.Id
69:     alias.AliasComponentId = *from.Id
70:     alias.RelationshipId = relationshipDeclaration.Id
71:     alias.ImmediateRefFieldPath = mutatedRefs[0] // <--- Potential Panic Here
```

**Why it's a problem:**
While the code checks if `from.Patch` and `from.Patch.MutatedRef` are `nil`, it does not check the *length* of the `MutatedRef` slice. A slice can be non-nil but have a length of 0. If a malformed or edge-case design file is submitted where `MutatedRef` is provided but empty (`[]`), the `mutatedRefs[0]` access will cause an index out-of-bounds panic. Go panics in HTTP handlers or background goroutines can lead to dropped requests or worker crashes.

**Severity:** 
Medium/High (Potential denial of service or request crash via malformed relationship definitions)

**Steps to Fix:**
Add a length check for `mutatedRefs` before attempting to access the first element, similar to the checks performed for `selectors`, `fromSet`, and `toSet` earlier in the same function.

Change:
```go
    mutatedRefs := *from.Patch.MutatedRef

    alias.ImmediateParentId = *to.Id
    alias.AliasComponentId = *from.Id
    alias.RelationshipId = relationshipDeclaration.Id
    alias.ImmediateRefFieldPath = mutatedRefs[0]
```
To:
```go
    mutatedRefs := *from.Patch.MutatedRef

    if len(mutatedRefs) == 0 {
        return alias, false
    }

    alias.ImmediateParentId = *to.Id
    alias.AliasComponentId = *from.Id
    alias.RelationshipId = relationshipDeclaration.Id
    alias.ImmediateRefFieldPath = mutatedRefs[0]
```

This guarantees safety before taking the 0th index and cleanly handles invalid relationship alias parameters.

Ironankit525:
## Description

Fixes #17721

### Problem
`parseRelationshipToAlias` in `server/handlers/policy_relationship_handler.go` dereferences the `Selectors` pointer (`*relationshipDeclaration.Selectors`) at line 47 without checking if it is `nil`. If a `RelationshipDefinition` has a nil `Selectors` field, this causes a nil-pointer panic that can crash the server or policy evaluation worker.

### Changes
- Added a nil guard for the `Selectors` pointer before dereferencing it in `parseRelationshipToAlias`
- - Added comprehensive unit tests for `parseRelationshipToAlias` covering all nil/empty edge cases:
-   - Wrong subtype
-   - Nil `Selectors`
-   - Empty `Selectors`
-   - Empty `From`/`To` sets
-   - Nil `Patch`
-   - Nil `MutatedRef`
-   - Empty `MutatedRef`
-   - Valid alias relationship (happy path)
### Testing
All 9 unit tests pass:
```
=== RUN   TestParseRelationshipToAlias
--- PASS: TestParseRelationshipToAlias (0.00s)
PASS
ok      github.com/meshery/meshery/server/handlers      2.055s
```


server/handlers/policy_relationship_handler.go
PragalvaXFREZ, line 1: Thanks @Ironankit525, could you also ensure there are no other instances of nil pointer dereference here? For example, `to.Id` and `from.Id` are dereferenced without nil checks.

Ironankit525: i added nil checks for to.Id and from.Id too. Thanks for pointing that out!

server/handlers/policy_relationship_handler_test.go
PragalvaXFREZ, line 1: could you add a dedicated test case for nil to.Id / from.Id guard

Ironankit525: Added test cases for both nil to.Id and nil from.Id — all passing. Let me know if it looks good.

server/handlers/policy_relationship_handler_test.go
```diff
@@ -0,0 +1,235 @@
...
+			}(),
+			wantOk: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, ok := parseRelationshipToAlias(tt.input)
+			if ok != tt.wantOk {
+				t.Errorf("parseRelationshipToAlias() ok = %v, want %v", ok, tt.wantOk)
```
lekaf974: please use assertion package as we do in unit tests

Ironankit525: Done! Updated to use `assert.Equal` from testify.

server/handlers/policy_relationship_handler_test.go
```diff
@@ -0,0 +1,234 @@
...
+				return relationship.RelationshipDefinition{
+					SubType:   RELATIONSHIP_SUBTYPE_ALIAS,
+					Selectors: &ss,
+				}
+			}(),
+			wantOk: true,
+		},
+	}
+
+	for _, tt := range tests {
```
PragalvaXFREZ, line 231: This discards the returned alias, can you assert on the actual values too? Otherwise a bug could return true with wrong field values and this test would still pass.

PragalvaXFREZ (commented):
one comment, lgtm otherwise.

server/handlers/policy_relationship_handler_test.go
Ironankit525: Thanks for the feedback! I've updated the happy-path test case to explicitly assert on all the returned alias field values (ImmediateParentId, AliasComponentId, RelationshipId, and ImmediateRefFieldPath). This ensures we don't accidentally pass the test if the function returns true but populates incorrect data. Let me know if there's anything else!

server/handlers/policy_relationship_handler_test.go
```diff
@@ -0,0 +1,241 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			alias, ok := parseRelationshipToAlias(tt.input)
+			assert.Equal(t, tt.wantOk, ok, "parseRelationshipToAlias() ok mismatch")
+
+			if tt.wantOk {
+				assert.Equal(t, toID, alias.ImmediateParentId, "ImmediateParentId should match to.Id")
+				assert.Equal(t, fromID, alias.AliasComponentId, "AliasComponentId should match from.Id")
+				assert.Equal(t, tt.input.Id, alias.RelationshipId, "RelationshipId should match the relationship's Id")
```
PragalvaXFREZ: Id isn't set on the fixture, so this assertion is just comparing uuid.Nil to uuid.Nil. It'll pass even if the field stops being copied. Setting an explicit relID here would make this actually test something.

Ironankit525: Thanks for the catch! You're right  -- the `Id` field wasn't set on the fixture, so it was just comparing `uuid.Nil` to `uuid.Nil`. I've fixed it by generating an explicit `relID := uuid.Must(uuid.NewV4())`, assigning it to the test fixture with `rd.Id = relID`, and updating the assertion to compare against `relID` directly instead of `tt.input.Id`. This ensures the test actually verifies that the `RelationshipId` field is being copied correctly.

PragalvaXFREZ (approved):
lgtm

Ironankit525:
Thanks for the feedbacks and for merging this.

PR #17768, opened 2026-03-05
Improve error message when Kubernetes context is missing after database reset

Omkar-Ugal:
Description

While testing the database reset flow locally, I noticed that after performing a reset the stored Kubernetes cluster records are removed from the Meshery database. In this situation, the state machine can still attempt to retrieve the Kubernetes context for a machine.

When that happens, the current error message isn’t very clear and can make it difficult to understand what actually went wrong.

This PR updates the error message in the resync path to provide a clearer explanation. The new message indicates that the cluster connection may have been removed due to a database reset and suggests reconnecting the Kubernetes cluster from Settings → Kubernetes Clusters.

There is no functional change in this PR—this only improves the clarity of the error message for users encountering this scenario.

Notes for Reviewers
Fixes : https://github.com/meshery/meshery/issues/17787
Improves clarity of the error returned when a Kubernetes context cannot be retrieved after a database reset.

No behavioral or logic changes, only error message wording.

Signed commits

 Yes, I signed my commits.


PragalvaXFREZ:
@Omkar-Ugal , is there an existing issue related to this? if you would link that it would be really helpful, thanks

Omkar-Ugal:
Good point, thanks for pointing that out.

You're right that GetMachineCtx() can fail for reasons other than a database reset (for example if the context assertion to *MachineCtx fails). My initial intention was to help users understand the scenario we observed after a reset, but the message could indeed be misleading if the failure occurs for other reasons.

I'll update the message to keep it more generic while still wrapping the original error for debugging clarity.

Omkar-Ugal:
@PragalvaXFREZ 

Yes — this PR is related to issue #17787. While investigating the database reset behavior described there, I noticed that when the Kubernetes context is removed the resync flow can still attempt to retrieve it, and the existing error message wasn't very clear.

This PR improves the error message in that scenario to make it easier for users to understand what happened.

Amr-Shams:
this new message after the review, doesn't add or explain the issue linked to the pr, i am not sure if i am fully following up here

Omkar-Ugal:
> this new message after the review, doesn't add or explain the issue linked to the pr, i am not sure if i am fully following up here

This PR was created while investigating issue #17787. After a database reset, the stored Kubernetes cluster connections can be removed from the Meshery database, but the resync flow may still attempt to retrieve the Kubernetes context. When that happens, the current error message is not very clear.

My intention here was to improve the error message in that scenario.

Based on earlier feedback, I made the message more generic and wrapped the underlying error using `%w` so the original cause is preserved. However, I see that this may have made the connection to the database reset scenario less obvious.

I can update the message to keep it generic while also mentioning that a database reset could be one possible cause.

PR #17840, opened 2026-03-08
refactor(mesheryctl): centralize meshkit logger initialization

lekaf974:

**Notes for Reviewers**

## Why

Current implementation in pkg/utils is causing several issues
Some works are blocked causing cycling depency, etc.

## How

This implementation replace current utils.Log by a new centralized logger package to isolate logging set, configuration in one place `internal/cli/pkg/logger` which prevent cycle dependency.

Also, to avoid modifying everywhere in one PR, current utils.Log is initiate as e reference to the new centralized variable which allow to move foward with smaller changes.

Other PR will come to transition from utils.Log to mesheryctllogger.Log

- This PR fixes partially #4705

### DIFF
--- a/docs/hugo.toml
+++ b/docs/hugo.toml
@@ -89,7 +89,8 @@ As a self-service engineering platform, Meshery enables collaborative design and
     style = "solarized-dark256"
     # Uncomment if you want your chosen highlight style used for code blocks without a specified language
     # guessSyntax = "true"
-
+    noClasses = false
+    
 # Everything below this are Site Params
 
 # Comment out if you don't want the "print entire section" link enabled.


mesheryctl/internal/cli/pkg/logger/logger.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			logrus.Errorf("Failed to initialize logger: %v", err)
+			os.Exit(1)
+		}
+
+		Log = &MesheryctlLogger{Log: log}
+	})
+
+	return Log
+}
```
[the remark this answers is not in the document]
lekaf974: A SetLevel method is available

lekaf974:
@Yorisoft review is welcomed here

mesheryctl/internal/cli/pkg/logger/logger.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"io"
+	"os"
+	"sync"
+
+	"github.com/meshery/meshkit/logger"
+	logrus "github.com/sirupsen/logrus"
+)
+
+// MesheryctlLogger is a wrapper around the meshkit logger.Handler to provide a consistent logging interface into mesheryctl.
```
leecalcote: Creating this wrapper has the effect of bifurcating MeshKit logger's used between the CLI and Server. More than that, it separately implements logging in the CLI from the methods used in every other Meshery component. 

If improvements to the project's approach to logging are needed, can we consider making those in our centrally reusable library, MeshKit?

Also, as discussed in the Meshery Dev meeting, can we postpone this optional improvement  to after v1.0?

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -56,8 +56,7 @@ mesheryctl system channel view
 		}
 		mctlCfg, err = config.GetMesheryCtl(viper.GetViper())
 		if err != nil {
-			utils.Log.Error(err)
-			return nil
+			return err
```
Yorisoft, line 58: No logging here - how come?

mesheryctl/internal/cli/pkg/logger/logger.go
lekaf974: Basically this only centralizing meshkit logger into a struct instead and ensure only one instance of the logger is instantiate in mesheryctl then recall it from a new dedicated package logger, nothing new that need backport in meshkit.

I can close the PR and postpone for later

mesheryctl/internal/cli/root/system/channel.go
lekaf974: cobra already output error properly also 

```golang
utils.Log.Error(err)
return nil
```

will return an exit code 0 which is not the case

Yorisoft (approved):

lekaf974:
@shri771, @PragalvaXFREZ can you take a look on this one ?

leecalcote:
@Yorisoft, thanks for offering your review here! 👍

Issue #17842, opened 2026-03-08
Refactor: Extract config mutation logic from initConfig into helper


SuryanshGarg04:
Part of #17159

## Background

Currently, configuration mutation logic (config creation, token insertion, and default context setup) is implemented directly inside `initConfig()` in `mesheryctl/internal/cli/root/root.go`.

This logic executes globally via:

cobra.OnInitialize(initConfig, setupLogger, setVerbose)

As a result, configuration mutation is tightly coupled with CLI initialization.

## Problem

`initConfig()` currently mixes multiple responsibilities:

- config file existence checks
- config mutation (token and context insertion)
- config loading via Viper
- Kubernetes setup
- logger initialization

This makes it difficult to isolate side effects for read-only commands.

## Proposed Change

Refactor config mutation logic into a dedicated helper in the `config` package.

Changes include:

- Introduce `MutateConfigIfNeeded()` in `internal/cli/root/config/mutate.go`
- Move default config mutation logic out of `root.go`
- Introduce `templates.go` for `TemplateToken` and `TemplateContext`
- Update `initConfig()` to delegate mutation logic to the helper

## Scope

This change is **refactor-only** and does **not change runtime behavior**.

Follow-up work will isolate config mutation so that read-only commands remain side-effect free.
### Environment

- Host OS: Mac, Linux, Windows
- Meshery Server Version: stable-v
- Meshery Client Version: stable-v
- Platform: Docker Kubernetes

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 📚 [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- 📺 [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- ⌨️ [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- ⌨️ [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- ⌨️ [Meshkit errors Guide](https://docs.meshery.io/project/contributing/contributing-error)
[...]

Issue #17159, opened 2026-01-26
Read-only mesheryctl commands trigger config and Kubernetes initialization via cobra.OnInitialize


SuryanshGarg04:
### Current Behavior

All mesheryctl commands trigger global initialization via `cobra.OnInitialize(initConfig, setupLogger, setVerbose)`.

As a result, even read-only commands such as `mesheryctl system status`, `mesheryctl version`, or `mesheryctl help` perform:

- Kubernetes setup via `utils.SetKubeConfig()`
- config file creation and default context insertion
- local state mutation
- config/K8s-related logs

This happens even when commands do not require Kubernetes or configuration mutation.

---

### Desired Behavior

Read-only mesheryctl commands should be side-effect free.

Commands like `system status`, `version`, and `help` should not initialize Kubernetes or create/mutate config.

Configuration and Kubernetes setup should only occur for commands that explicitly require them.

This appears to be caused by `initConfig()`mixing config preparation, config mutation, and Kubernetes initialization while being executed globally via Cobra. Open to maintainer guidance on the preferred approach.

---

### Screenshots/Logs

N/A

--
### Environment

- Host OS: Mac, Linux, Windows
- Meshery Server Version: stable-v
- Meshery Client Version: stable-v
- Platform: Docker Kubernetes

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 📚 [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- 📺 [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- ⌨️ [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- ⌨️ [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- ⌨️ [Meshkit errors Guide](https://docs.meshery.io/project/contributing/contributing-error)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)
[...]

PR #17843, opened 2026-03-08
refactor(mesheryctl): extract config mutation logic into helper

SuryanshGarg04:
Part of #17842 

## Summary

This PR refactors the config initialization flow in mesheryctl by extracting config mutation logic from `initConfig()` into helper functions within the `config` package.

The goal is to improve separation of concerns and prepare the codebase for isolating config mutation from read-only CLI commands.

---

## Changes

- Introduced `NeedsMutation()` in `internal/cli/root/config/mutate.go` to determine whether config mutation is required
- Introduced `InitDefaultConfig()` to handle config creation and default mutation (token + context)
- Moved config mutation logic (config creation, token insertion, context insertion) out of `root.go`
- Updated `initConfig()` to delegate mutation responsibilities to the config package
- Removed previously proposed `MutateConfigIfNeeded()` abstraction in favor of simpler, explicit helpers based on maintainer feedback
- Continued use of existing `utils.TemplateToken` and `utils.TemplateContext`

---

## Before

`initConfig()`:
- checks config
- creates config
- inserts token
- inserts context
- mixes concerns (validation + mutation + logging)

---

## After

`initConfig()`:
- checks config state
- calls `NeedsMutation()` to determine if mutation is required
- delegates creation + mutation to `InitDefaultConfig()` when needed

---

## Behavior

No functional changes.

All existing CLI commands behave exactly as before since mutation still occurs during initialization.

---

## Why this change

This refactor prepares the codebase for addressing #17159, where read-only commands should not trigger configuration mutation or Kubernetes initialization.

By isolating mutation logic into the `config` package, future changes can selectively trigger config mutation only for commands that require it.

---

## Next Steps

Follow-up PR will:

- remove mutation from global initialization (`cobra.OnInitialize`)
- introduce command-level control over config initialization
- ensure read-only commands remain side-effect free

---

## Testing

- `go build ./...`
- `go test ./...`

Manual CLI checks:
- `mesheryctl version`
- `mesheryctl system start`
- `mesheryctl help`


leecalcote:
@reaper8055, perhaps, you might offer some feedback here.

reaper8055:
> @reaper8055, perhaps, you might offer some feedback here.

Would love to. I will take a look at this today in the evening

mesheryctl/internal/cli/root/config/mutate.go
```diff
@@ -0,0 +1,50 @@
...
+
+	stat, err := os.Stat(configPath)
+	createDefaultConfig := false
+
+	switch {
+	case os.IsNotExist(err):
+		createDefaultConfig = true
+
+	case err == nil && stat.Size() == 0:
+		createDefaultConfig = true
```
reaper8055: ~Since both are setting `createDefaultConfig = true` you can combine these into one condition?~

Check this comment https://github.com/meshery/meshery/pull/17843/changes#r2912193559

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	config "github.com/meshery/meshery/mesheryctl/internal/cli/root/config"
+	"github.com/spf13/viper"
+)
+
+const minimalMeshConfig = "contexts: {}\ncurrent-context: \"\"\ntokens: []\n"
```
reaper8055: You can use backticks for this makes it much more readable:
```go
const minimalMeshConfig = `contexts: {}
current-context: ""
tokens: []
`
```

Also can you help me understand why this has to be a global const? Having it within the test function `TestMutateConfigIfNeeded` has same behavior? 

(I haven't looked at other files) but if this is used other tests, generally speaking, it's better to keep test data scoped to each test even constants.

mesheryctl/internal/cli/root/config/mutate.go
reaper8055, line 1: I have reviewed this file, please check the comments.

mesheryctl/internal/cli/root/config/mutate.go
```diff
@@ -0,0 +1,50 @@
+package config
+
+import (
+	"os"
+)
+
+func MutateConfigIfNeeded(
```
reaper8055: This function although logically correct has a bit of a cognitive load. 

You did do a great job in capturing and covering all edge cases so that's good work!!

What would help make this function more readable and have less cognitive overhead is separating "what" from "how". 

In this case the "what" is the logic that createConfigFile is called. The "how" is all the checks, AddTokenToConfig, AddContextToConfig, etc. 

One way I can think of achieving this is to split this function into two:
```go
func NeedsMutation(configPath string) bool 

func InitDefaultConfig(mesheryFolder string, token Token, ctx Context, createConfigFile func() error) error
```

mesheryctl/internal/cli/root/config/mutate.go
```diff
@@ -0,0 +1,50 @@
...
+	case err == nil && stat.Size() == 0:
+		createDefaultConfig = true
+
+	case err != nil:
+		return err
+	}
+
+	if !createDefaultConfig {
+		return nil
+	}
```
reaper8055: Looking at the logic flow here, it seems like the use of createDefultConfig is redundant. 

This logic returns nil when:

- all case statements are false, i.e. os.IsNotExist(err) is false, (err == nil && stat.Size() == 0) is false, and err != nil is false. Since os.IsNotExist(err) and stat.Size() == 0 can't be both true so the nil condition boils down to:

```go
if err == nil && stat.Size() > 0 {
    return nil
}
```

This logic creates config file only when:

- configPath does not exist 
- err is nil and stat.Size is zero

This looks like happy path, so idiomatic go we keep happy path aligned to left so we don't need to check for this we can just early return on the error case which is 3rd case in switch statement coupled with file not exist check:

```go
if err != nil && !ok.IsNotExixt(err) {
    return err
}
```

And then rest of your code:
```go
if err := os.MkdirAll(mesheryFolder, 0o775); err != nil {
// ...
```

mesheryctl/internal/cli/root/config/mutate.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if err := os.MkdirAll(mesheryFolder, 0o775); err != nil {
```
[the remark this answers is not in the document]
SuryanshGarg04: Since this directory may contain sensitive configuration data, restricting permissions to 0700 makes sense. I'll update this accordingly.

mesheryctl/internal/cli/root/config/mutate_test.go
SuryanshGarg04: Good suggestion, using backticks definitely improves the readability here. This constant was only intended for use within this test, so you're right that it doesn’t need to be global. I'll switch it to a raw string literal and move it inside the `TestMutateConfigIfNeeded` function so the test data stays scoped to the test.

mesheryctl/internal/cli/root/config/mutate.go
SuryanshGarg04: Thanks for the feedback and suggestion! I understand what you mean about separating the "what" from the "how" to reduce cognitive load. Extracting the mutation decision into a helper like `NeedsMutation` makes sense. I’ll look into refactoring this to split the check from the mutation steps to make the flow clearer.

mesheryctl/internal/cli/root/config/mutate.go
SuryanshGarg04: You're right that the createDefaultConfig flag and switch can be simplified by handling the error case and the existing-file case with early returns. I'll refactor this accordingly so the happy path stays aligned to the left.

mesheryctl/internal/cli/root/config/mutate.go
SuryanshGarg04: Thanks for the review! I've gone through the comments and replied to them. I'll incorporate the suggested changes and update the PR.

mesheryctl/internal/cli/root/config/templates.go
reaper8055, line 1: I have reviewed this file, please check the comments.

mesheryctl/internal/cli/root/config/mutate_test.go
reaper8055, line 1: I have reviewed this file, please check the comments.

mesheryctl/internal/cli/root/config/templates.go
```diff
@@ -0,0 +1,13 @@
...
+	Name:     "default",
+	Location: "",
+}
+
+var TemplateContext = Context{
+	Endpoint: "",
+	Platform: "docker",
+	Channel:  "stable",
+	Version:  "latest",
+}
```
reaper8055: It make more sense to place these in [mesheryctl/pkg/utils/helpers.go](https://github.com/meshery/meshery/blob/master/mesheryctl/pkg/utils/helpers.go) where they originally were since there are similar things done over there.

Also can you help me understand why did you choose to use empty string for `Location` field in `TemplateToken` where as, initially it was `AuthConfigFile`: [mesheryctl/pkg/utils/helpers.go#L357-L361](https://github.com/meshery/meshery/blob/master/mesheryctl/pkg/utils/helpers.go#L357-L361)

```go
var TemplateToken = config.Token{
	Name:     "default",
	Location: AuthConfigFile,
}
```

Same for `TemplateContext` [mesheryctl/pkg/utils/helpers.go#L276-L284](https://github.com/meshery/meshery/blob/master/mesheryctl/pkg/utils/helpers.go#L276-L284) which has some additional fields which are missing from your implementation:

```go
var TemplateContext = config.Context{
	Endpoint:   EndpointProtocol + "://localhost:9081",
	Token:      "default",
	Platform:   "kubernetes",
	Components: ListOfComponents,
	Channel:    "stable",
	Version:    "latest",
}
```

Are the missing fields not needed?

Lastly, whenever you have such initializations, it's good to add some comments explaining what the variable and its value represents.

mesheryctl/internal/cli/root/root_test.go
reaper8055, line 1: This will need some adjustment after you have addressed: https://github.com/meshery/meshery/pull/17843/changes#r2911865177 and https://github.com/meshery/meshery/pull/17843/changes#r2925652549

mesheryctl/internal/cli/root/root.go
reaper8055, line 1: This will need some minor adjustment after you have addressed this https://github.com/meshery/meshery/pull/17843/changes#r2911865177 and https://github.com/meshery/meshery/pull/17843/changes#r2925652549

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func TestMutateConfigIfNeeded(t *testing.T) {
+
+	tests := []struct {
```
reaper8055, line 14: The names for this test struct is very clear and concise. This is good! Keep doing this! 

One small thing that I would like to see is use of verbs `got` and `want` along with `name`, `given`:

- `got` is what you get when you call the function under test
- `want` is what your expect to get from your function under test

mesheryctl/internal/cli/root/version.go
reaper8055, line 1: This will need some adjustment after you have addressed: https://github.com/meshery/meshery/pull/17843/changes#r2911865177 and https://github.com/meshery/meshery/pull/17843/changes#r2925652549

mesheryctl/internal/cli/root/config/mutate_test.go
SuryanshGarg04: I'll update the test fields to follow the `got` and `want` convention in the follow-up commits.

mesheryctl/internal/cli/root/config/templates.go
SuryanshGarg04: Thanks for pointing this out.

When I initially extracted the mutation logic into `config.MutateConfigIfNeeded`, I created a temporary template in the `config` package because I was encountering an import cycle while experimenting with the refactor. While doing that, I simplified the struct definitions, which is why `Location` ended up as an empty string and some fields in `TemplateContext` were missing.

After revisiting the dependency structure, I realized the helper can simply receive the token and context as parameters from `root.go`. With that approach, the `config` package itself does not need to import `pkg/utils`, so the import cycle issue does not occur.

I’ll update the implementation to reuse the existing `TemplateToken` and `TemplateContext` from `mesheryctl/pkg/utils/helpers.go` so that the defaults remain consistent (including `Location: AuthConfigFile` and the additional fields in `TemplateContext`). The tests will also reference those values from `utils`, which should not introduce a cycle since the `config` package itself does not depend on `utils`.

Also noted regarding adding comments explaining these initialization values, I'll keep that in mind when updating the code.

Thanks for the review!

mesheryctl/internal/cli/root/config/mutate.go
SuryanshGarg04: I have pushed the suggested changes in mutate.go and will update the mutate_test.go by evening, 
Question- Is the permission for mesheryfolder okay to be updated to 0o700 or should i follow the older approach

reaper8055: I would assume permission 700 would work and suffice over 775 unless there's some reason why 775 was set earlier.

mesheryctl/internal/cli/root/config/mutate.go
```diff
@@ -0,0 +1,52 @@
...
+
+	case err == nil && stat.Size() == 0:
+		return true, nil
+
+	case err != nil:
+		return false, err
+	}
+
+	return false, nil
+}
```
reaper8055: This still has some logical redundancy, please review earlier comment https://github.com/meshery/meshery/pull/17843#discussion_r2912193559

Since you are moving existing code, it might be a little confusing to think, but your NeedsMutation function should return either `true, nil` or `false, err`.

Can you give me an example case where the last return statement would be executed?

mesheryctl/internal/cli/root/config/mutate.go
```diff
@@ -0,0 +1,52 @@
...
+
+	if err := createConfigFile(); err != nil {
+		return err
+	}
+
+	if err := AddTokenToConfig(token, configPath); err != nil {
+		return err
+	}
+
+	if err := AddContextToConfig("local", ctx, configPath, true, false); err != nil {
```
reaper8055, line 50: Whenever you are hardcoding values, one thing that would help is, thinking about if "local" makes sense for the config and should it be available as a value to the project else where. Does that make sense? If not adding a comment always helps

mesheryctl/internal/cli/root/config/mutate.go
SuryanshGarg04: Thanks for the clarification. I'll update the folder permission to 0o700 accordingly

mesheryctl/internal/cli/root/config/mutate.go
SuryanshGarg04: That makes sense. In this case "local" represents the default Meshery context created during initialization. I also noticed that the GitHub workflow tests create the config with the same default context name ("local"), so I kept it consistent with the existing behavior. I'll add a comment clarifying this in the code.

mesheryctl/internal/cli/root/config/mutate.go
SuryanshGarg04: You're right that the previous switch-based version made it unclear when the final` return false, nil` would execute. I’ve refactored NeedsMutation to use early error handling and explicit checks so the control flow is clearer:

`return true, nil` when the config file does not exist

`return true, nil` when the file exists but is empty

`return false, err` for unexpected filesystem errors

`return false, nil` when the config file exists and is valid

reaper8055: Ohh okay, that makes sense!

mesheryctl/internal/cli/root/config/mutate_test.go
SuryanshGarg04: I have pushed the suggested changes in mutate_test.go. Please let me know if any other changes are needed.
Thanks!

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	config "github.com/meshery/meshery/mesheryctl/internal/cli/root/config"
+	"github.com/meshery/meshery/mesheryctl/pkg/utils"
+	"github.com/spf13/viper"
+)
+
+func TestConfigMutation(t *testing.T) {
```
reaper8055: Shouldn't this be testing `NeedsMutation` and `InitDefaultConfig` functions?

reaper8055 (commented):
mutate.go looks good. Good work on refactoring the switch-case logic!

mutate_test.go needs some work. I have added comments, please check them out.

Also, please resolve older comments that you have already addressed.

mesheryctl/internal/cli/root/config/mutate_test.go
SuryanshGarg04: Currently it behaves more like an integration test since both functions are tested together. Splitting them makes sense so we can identify which function is causing failures.
I’ll update this 👍

SuryanshGarg04:
I have addressed the suggested changes in mutate_test.go.
Could you please review and let me know if anything else needs to be updated or improved?

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+
+			if got != tt.want {
+				t.Fatalf("NeedsMutation() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestNeedsMutation_ErrorCase(t *testing.T) {
```
reaper8055: You can merge this to the TestNeedsMutataion, error cases should be part of the same test function.

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,200 @@
...
+			want: true,
+		},
+		{
+			name: "Given non-empty config file, When NeedsMutation is called, Then it returns false",
+			setup: func(path string, t *testing.T) {
+				if err := os.WriteFile(path, []byte("data"), 0o644); err != nil {
+					t.Fatalf("failed to write file: %v", err)
+				}
+			},
+			want: false,
```
reaper8055: These test cases are great but they are not "strictly" testing the "behavior" of NeedsMutation. 

NeedsMutation returns two values `(bool, error)`. These test cases should test for both. The idea here is you test for all possible flows and edge cases within the function. 

There are multiple ways to approach this, I will let you decide and figure this out.

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,200 @@
...
+		invalidPath := string([]byte{0})
+
+		_, err := config.NeedsMutation(invalidPath)
+		if err == nil {
+			t.Fatal("expected error for invalid path, got nil")
+		}
+	})
+}
+
+func TestInitDefaultConfig(t *testing.T) {
```
reaper8055, line 75: I would say, same thing goes for this. You would want to test all the cases where InitDefaultConfig returns nil or err. Looking at the function, it returns an error on 4 conditional checks. So you would want to do that in TestInitDefaultConfig.

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,200 @@
...
+			t.Fatalf("config not found: %v", err)
+		}
+
+		if len(data) == 0 {
+			t.Fatal("expected config to be created")
+		}
+	})
+}
+
+func TestConfigMutation(t *testing.T) {
```
reaper8055: I remember you mentioning that this is basically testing config mutation logic, but if you test the code which uses NeedsMutation and InitDefaultConfig that should cover this, no? 

I'm not saying this is not needed but adding "integration" type tests to unit test can lead to flakey tests where you are not "strictly" testing the function that uses these utility function but rather kinda "mocking" their use, helpful but maybe not in this case.

SuryanshGarg04: That makes sense, thanks for pointing it out.
I was thinking about covering the combined behavior where NeedsMutation returns true and InitDefaultConfig is triggered, to ensure the overall mutation flow works as expected.
However, I agree that this overlaps with integration behavior and is better validated at a higher level (e.g., where these functions are actually used in command execution), rather than within unit tests.
I’ll focus on making the unit tests for NeedsMutation and InitDefaultConfig more comprehensive and remove/simplify the combined test to avoid redundancy and potential flakiness.

mesheryctl/internal/cli/root/config/mutate_test.go
SuryanshGarg04: I’ll update the NeedsMutation tests to cover both return values (bool, error) and merge the error case into the same test function.

mesheryctl/internal/cli/root/config/mutate_test.go
SuryanshGarg04: I’ll expand TestInitDefaultConfig to cover all error paths (MkdirAll, createConfigFile, AddTokenToConfig, AddContextToConfig).

SuryanshGarg04:
Hey @reaper8055 
I have Updated tests as suggested:
- Consolidated NeedsMutation tests to cover both return values (bool, error) and all flows including invalid path
- Added failure cases in InitDefaultConfig for createConfig and MkdirAll
- For AddTokenToConfig and AddContextToConfig, these rely on internal config behavior and are not easily mockable here, so I focused on testing the directly controllable failure paths
Please let me know if you'd prefer additional handling for those cases.

SuryanshGarg04:
Merge conflicts introduced a few issues during resolution. I’m addressing them and will push an update once resolved.
Thanks!

mesheryctl/internal/cli/root/config/mutate_test.go
reaper8055, line 75: This function should be testing 5 paths at min:
1. `if err := os.MkdirAll(mesheryFolder, 0o700); err != nil` 
2. `if err := createConfigFile(); err != nil`
3. `if err := AddTokenToConfig(token, configPath); err != nil`
4. `if err := AddContextToConfig("local", ctx, configPath, true, false); err != nil`
5. `return nil`

You have already tested: (5.), (2.) and (1.).

Here are the missing test cases for the following paths:

3. `if err := AddTokenToConfig(token, configPath); err != nil` 
4. `if err := AddContextToConfig("local", ctx, configPath, true, false); err != nil`

A rule of thumb when writing unit tests is:
- cover all paths that the function under test can take.
- cover all hidden paths that might be introduced as a result of calls to external functions.

reaper8055: It's generally preferable to use function names where ever possible on the "name" of the test case:

This:
```go
t.Run("Given createConfig fails, When InitDefaultConfig is called, Then error is returned", func(t *testing.T) { ... }
```

Becomes:

```go
t.Run("Given createConfigFile fails, When InitDefaultConfig is called, Then error is returned", func(t *testing.T) { ... }
```

A subtle change but makes a huge difference in readability

reaper8055, line 75: To make the "style" more consistent across lets use the same style (table drive) as you have done in your `TestNeedsMutation` function by first defining the tests in a `[]struct{ ... }` and then use `t.Run(...)` by looping over the `[]struct{...}`.

SuryanshGarg04:
@reaper8055 
Thanks for the feedback!
Added test coverage for the missing AddTokenToConfig and AddContextToConfig failure paths. Also updated test case names to use exact function names and converted TestInitDefaultConfig to a table-driven format for consistency.
Let me know if any further changes are needed.

mesheryctl/internal/cli/root/root.go
```diff
@@ -163,61 +160,43 @@ func TreePath() *cobra.Command {
...
-				utils.Log.Fatal(err)
-			}
+		if needsMutation {
+			err = config.InitDefaultConfig(
+				utils.DefaultConfigPath,
+				utils.MesheryFolder,
+				utils.TemplateToken,
+				utils.TemplateContext,
+				utils.CreateConfigFile,
+			)
```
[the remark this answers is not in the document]
SuryanshGarg04: That makes sense; i will update the PR description

mesheryctl/internal/cli/root/root.go
```diff
@@ -163,56 +162,38 @@ func TreePath() *cobra.Command {
...
```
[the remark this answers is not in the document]
SuryanshGarg04: Updated the PR description.

mesheryctl/internal/cli/root/config/mutate.go
```diff
@@ -0,0 +1,52 @@
+package config
+
+import "os"
+
+// NeedsMutation determines whether the Meshery config file needs to be created or updated by checking if it exists, is non-empty, and accessible; it returns true when the file is missing or empty, false otherwise, and an error if the file cannot be accessed.
```
reaper8055: This line is too long and can be hard to follow, please consider wrapping this line.

mesheryctl/internal/cli/root/root.go
```diff
@@ -163,61 +160,43 @@ func TreePath() *cobra.Command {
...
 
-			if err := config.AddTokenToConfig(utils.TemplateToken, utils.DefaultConfigPath); err != nil {
+			if err != nil {
 				utils.Log.Fatal(err)
 			}
```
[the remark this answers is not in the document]
reaper8055: @SuryanshGarg04 please check this suggestion from copilot

mesheryctl/internal/cli/root/config/mutate.go
```diff
@@ -0,0 +1,52 @@
...
+
+	if err := os.MkdirAll(mesheryFolder, 0o700); err != nil {
+		return err
+	}
+
+	if err := createConfigFile(); err != nil {
+		return err
+	}
+
+	if err := AddTokenToConfig(token, configPath); err != nil {
```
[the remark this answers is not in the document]
SuryanshGarg04: Thanks for pointing this out!

You're right that having `createConfigFile` rely on implicit/global state while `InitDefaultConfig` accepts a `configPath` introduces an inconsistency and could lead to issues if a non-default path is used.

I started exploring making the helper fully path-aware (by passing `configPath` into the callback), but that change propagates to `utils.CreateConfigFile` and multiple call sites, which expands the scope beyond this refactor.

To keep this PR focused on extracting mutation logic without introducing broader API changes, I’ll retain the current signature here and address making config initialization fully path-aware in a follow-up PR.

Appreciate the suggestion ,it’s a good improvement to make next.

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,196 @@
...
+			wantErr: true,
+		},
+		{
+			name: "Given os.MkdirAll fails, When InitDefaultConfig is called, Then error is returned",
+			setup: func(_ string, _ *testing.T) {},
+			configPath: func(_ string) string { return "/invalid/path/config.yaml" },
+			createConfig: func(_ string) func() error {
+				return func() error { return nil }
+			},
+			wantErr: true,
```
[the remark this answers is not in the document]
SuryanshGarg04: Good catch, updated the test to make mesheryFolder invalid so it properly exercises the MkdirAll failure paths are avoided and hard-coded paths.

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+				// Make directory read-only
+				if err := os.Chmod(dir, 0o500); err != nil {
+					t.Fatalf("chmod failed: %v", err)
+				}
+
+				// Restore permissions after test
+				t.Cleanup(func() {
+					_ = os.Chmod(dir, originalPerm)
+				})
```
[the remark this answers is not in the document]
SuryanshGarg04: Good catch , updated the test to make the config file read-only instead of the directory, so the failure is deterministic.

SuryanshGarg04:
Hi @lekaf974 , just following up on this PR.
I’ve addressed all the feedback and updated the tests. Please let me know if there’s anything else you’d like me to adjust; otherwise, I would appreciate a review when you have time.

mesheryctl/internal/cli/root/config/mutate.go
```diff
@@ -0,0 +1,55 @@
...
+func InitDefaultConfig(
+	configPath string,
+	mesheryFolder string,
+	token Token,
+	ctx Context,
+	createConfigFile func() error,
+) error {
+
+	if err := os.MkdirAll(mesheryFolder, 0o700); err != nil {
+		return err
```
lekaf974: would have been better to return a meshkit error here

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+			got, err := config.NeedsMutation(configPath)
+
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("error = %v, wantErr %v", err, tt.wantErr)
+			}
+
+			if got != tt.want {
+				t.Fatalf("NeedsMutation() = %v, want %v", got, tt.want)
+			}
```
lekaf974: can you use assert package

mesheryctl/internal/cli/root/config/mutate_test.go
```diff
@@ -0,0 +1,195 @@
...
+			err := config.InitDefaultConfig(
+				configPath,
+				mesheryFolder,
+				utils.TemplateToken,
+				utils.TemplateContext,
+				tt.createConfig(configPath),
+			)
+
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("error = %v, wantErr %v", err, tt.wantErr)
```
lekaf974: same here, can you use assert package

mesheryctl/internal/cli/root/root.go
```diff
@@ -82,8 +79,7 @@ mesheryctl -v [or] --verbose
 		}
 
 		if ok := utils.IsValidSubcommand(availableSubcommands, args[0]); !ok {
-			return errors.New(utils.RootError(fmt.Sprintf("'%s' is an invalid command. Use 'mesheryctl --help' to display usage guide.\n", args[0])))
-		}
+			return fmt.Errorf(utils.RootError("'%s' is an invalid command. Use 'mesheryctl --help' to display usage guide.\n"), args[0])		}
```
lekaf974: Since you change this, can you turn it as a meshkit error

mesheryctl/internal/cli/root/root_test.go
```diff
@@ -55,20 +54,25 @@ func TestRootCmdIntegration(t *testing.T) {
 			ExpectedResponse: "Context: local2\nChannel: edge\nVersion: latest\n\n",
 		},
 		{
-			Name:             "view the channel with verbose flag",
-			Args:             []string{"system", "channel", "view", "--verbose"},
-			ExpectedResponse: "Context: local\nChannel: stable\nVersion: latest\n\n",
+			Name: "view the channel with verbose flag",
```
lekaf974: could you rename using given when then naming

lekaf974 (changes requested):
Small `cosmetic` comments added LGTM otherwise

SuryanshGarg04:
@lekaf974 I’ve addressed all the review comments:
Replaced error handling with MeshKit structured errors where applicable
Updated tests to use testify assert
Refactored test names to follow Given–When–Then convention
Please let me know if anything else needs adjustment , happy to update further.

lekaf974:
@SuryanshGarg04 can this PR be presented in dev meeting

lekaf974:
Also can you resolve all comments fixed

SuryanshGarg04:
@lekaf974 Yes, I’ll present this PR in the dev meeting. I’ll also walk through the follow-up PR for this and the approach I’m taking there.

PragalvaXFREZ:
great @SuryanshGarg04, I have added this to the meeting agenda. Please present it there: 7:30 pm IST.

PR #17817, opened 2026-03-07
fix: Registry tab counts show zero on initial page load #15108

Mustansir-06:
## Description
This PR fixes the issue where Registry counts (Models, Components, Relationships, Registrants) would default to zero on initial load or within the Registry Modal.

### Changes Made:
- **Parallel Pre-fetching:** Modified `MeshModelComponent` to trigger all category counts simultaneously on mount using lazy queries with `pagesize: 1`.
- **Optimization:** Updated `RegistryModal` to fetch initial counts with `pagesize: 1` instead of `pagesize: 'all'`, significantly reducing the data payload and load time.
- **UX Improvement:** Introduced a loading state (`...`) for tab and sidebar counts to prevent the "flash of zero" bug, providing a smoother transition once data is cached.
- **Dependency Fix:** Added missing filters to the `fetchData` dependency array to ensure UI consistency during tab switching.

Fixes #15108

PR #17862, opened 2026-03-09
fix: resolve Registry count zero-load bug with parallel pre-fetching

### LINKED ISSUE #15108: [UI] Registry tab counts show zero on initial page load
### Current Behavior

When navigating to Settings → Registry for the first time, all tab counts display as zero:

Models (0)
Components (0)
Relationships (0)
Registrants (0)

However, the actual counts only appear after clicking on each respective tab. For example, clicking on the Models tab reveals there are actually 336 models available, Components shows 3854, and Relationships shows 45.

### Desired Behavior

Tab counts should display the correct numbers immediately upon page load, without requiring user interaction to trigger the count updates. Users should be able to see at a glance how many items are available in each registry category.

### Screenshots/Logs

Image

Image

### Environment

- Host OS: Mac Linux Windows
- Platform: Docker Kubernetes

---

Mustansir-06:
This PR is a clean replacement for #17817, containing only the 2 necessary file changes as requested by @leecalcote. It addresses the issue where Registry counts (Models, Components, Relationships, Registrants) would default to zero on initial load or within the Registry Modal.

Changes Made:

Optimized Data Fetching: Updated RegistryModal to fetch initial counts with pagesize: 1 instead of pagesize: all, significantly reducing the data payload and load time.

Parallel Pre-fetching: Modified MeshModelComponent to trigger all category counts simultaneously on mount using lazy queries.

UX Improvement: Introduced a loading state (...) for tab and sidebar counts to prevent the "flash of zero" bug before data is cached.

Dependency Fix: Added missing filters to the fetchData dependency array to ensure UI consistency during tab switching.

Fixes #15108

### DIFF
--- a/ui/components/Registry/RegistryModal.tsx
+++ b/ui/components/Registry/RegistryModal.tsx
@@ -32,7 +32,6 @@ import {
   useGetRelationshipsQuery,
   useGetRegistrantsQuery,
 } from '@/rtk-query/meshModel';
-import { removeDuplicateVersions } from '../Settings/Registry/helper';
 
 const DRAWER_WIDTH = 250;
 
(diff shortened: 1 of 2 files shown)


leecalcote (commented):
Smart optimization. Changing `pagesize: 'all'` to `pagesize: 1` for the count-only queries in `RegistryModal` is the right call - you only need `total_count` from the response metadata, not the full payload. This should dramatically reduce initial load time and data transfer, especially for large registries. Using the loading state to show `'...'` instead of `0` is a nice UX touch that avoids the misleading zero-count flash. 

Don't remove code comments unless they are inaccurate. If they are inaccurate, update the comments, @Mustansir-06

miacycle:
Sign off on your commits, @Mustansir-06

PR #17866, opened 2026-03-09
[UI] Resolve exhaustive-deps warnings and safely manage component

ashwanisingh011:
Description
Fixes react-hooks/exhaustive-deps warnings in 3 UI components while carefully preserving component state lifecycles.

ConfirmationModal.tsx: Refactored to use derived state for context filtering. This resolves the lint warning at the source and prevents background Redux updates from causing unnecessary side effects, without needing to use eslint-disable.

DashboardMeshModelGraph.tsx: Safely included the router dependency in the useMemo hook.

CodeMirror.tsx: Cleaned up the dependency array by removing unused options.


ashwanisingh011:
Just a heads up: the failing eslint-check CI is an upstream issue in provider-ui, completely unrelated to these ui/ changes.

For whoever fixes it, I did some quick debugging and found two bugs over there:

The pipeline dynamically pulls ESLint v10 (which breaks the Next.js patch).

Pinning to v8 reveals a config typo: TypeError: Could not find "no-unassigned-vars".

I kept this PR strictly scoped to the React fixes to avoid scope creep. Let me know if I should open a separate issue for the CI!

ui/components/CodeMirror.tsx
```diff
@@ -120,7 +120,7 @@ const CodeMirror = ({
     }
 
     return editorExtensions;
-  }, [options.lineWrapping, options.lint, options.mode, options.readOnly]);
+  }, [options.lineWrapping, options.lint, options.mode]);
```
aabidsofi19, line 123: why is this changed ? . this is not reflective of pr title

aabidsofi19 (changes requested):

ui/components/CodeMirror.tsx
ashwanisingh011: Hey @aabidsofi19! Thanks for the review.

The reason I changed this line is directly related to the react-hooks/exhaustive-deps rule. That rule flags both missing dependencies AND unnecessary/unused dependencies.

In this hook, options.readOnly was listed inside the dependency array, but it isn't actually being used anywhere inside the hook's logic block. To resolve the exhaustive-deps warning for this specific file, I removed the unused variable from the array.

However, if you would prefer I keep this PR strictly scoped to the usePrevious hook files, I am more than happy to revert this specific file! Just let me know what works best for you.

ui/components/ConfirmationModal.tsx
aabidsofi19: are we using the sistent modal here . it should be doing the statemanagement internally

aabidsofi19 (changes requested):

ui/components/ConfirmationModal.tsx
ashwanisingh011: You are completely right—the Sistent <Modal> handles its own lifecycle.
I have removed the usePrevious hook and the useEffect blocks that were manually syncing the props to local state.

Ready for another look!

leecalcote (commented):
Good refactor of `ConfirmationModal.tsx` — replacing `useEffect` + `useState` pairs with derived state (`isDisabled` computed directly, `filteredContexts` derived from `searchQuery`) is the React-idiomatic approach. This eliminates the stale-state bugs that come from syncing Redux state into local state via effects. The `CodeMirror.tsx` change (removing unused `options.readOnly` from deps) and the lint suppression comment in `ConfirmationModal` for intentionally omitted deps are both well-reasoned. Address the reviewer's change requests and this should be a net improvement in component reliability.

ashwanisingh011:
@aabidsofi19 Just a quick ping! I've refactored ConfirmationModal to use the Sistent internal state as you suggested, and resolved the comment threads. Let me know if everything looks good to go now so we can get this unblocked!

ui/components/ConfirmationModal.tsx
```diff
@@ -140,23 +140,19 @@ function ConfirmationMsg(props) {
...
+
   const { notify } = useNotification();
   const { selectedK8sContexts } = useSelector((state) => state.ui);
   const { k8sConfig: k8scontext } = useSelector((state) => state.ui);
-
-  let isDisabled =
-    typeof selectedK8sContexts.length === 'undefined' || selectedK8sContexts.length === 0;
   const dispatch = useDispatch();
-  useEffect(() => {
-    setTabVal(tab);
```
rishiraj38, line 154: I believe this was resetting the tab whenever the modal reopens. Without it, the tab value is only set once on mount — if the parent reopens the modal with a different tab prop, the stale tab will be displayed.

ashwanisingh011: Hey @rishiraj38, thanks for the review! I've added the useEffect back in to sync the local tabVal whenever the modal opens or the tab prop changes.

rishiraj38:
@ashwanisingh011 Every thing looks good to me one minor thing your PR description mentions adding eslint-disable comments, but I don't see any.

ashwanisingh011:
You're absolutely right, @rishiraj38! My original description was outdated.

Initially, I planned to use eslint-disable, but I ended up refactoring ConfirmationModal.tsx to use derived state instead (as Lee noted in his review). This fixed the exhaustive-deps warning at the source, making the inline comments unnecessary.

I've just updated the PR description to accurately reflect the final code. Let me know if that clears it up!

ui/components/ConfirmationModal.tsx
saurabhraghuvanshii, line 1: @ashwanisingh011  can you please address these warnings!!

ui/components/Dashboard/charts/DashboardMeshModelGraph.tsx
```diff
@@ -65,7 +65,7 @@ function MeshModelContructs() {
         },
       },
     }),
-    [data],
+    [data, router],
```
saurabhraghuvanshii: @ashwanisingh011  why router in DA?

ui/components/Dashboard/charts/DashboardMeshModelGraph.tsx
```diff
@@ -152,7 +152,7 @@ function MeshModelCategories() {
         show: false,
       },
     }),
-    [cleanedData],
+    [cleanedData, router],
```
saurabhraghuvanshii, line 155: @ashwanisingh011  same here

saurabhraghuvanshii (changes requested):

ui/components/ConfirmationModal.tsx
```diff
@@ -511,13 +502,12 @@ export const SelectDeploymentTarget_ = ({ k8scontext, selectedK8sContexts }) =>
         onChange={(event) => searchContexts(event.target.value)}
```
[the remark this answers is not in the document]
saurabhraghuvanshii: @ashwanisingh011  can you take a look here.

ui/components/ConfirmationModal.tsx
```diff
@@ -18,11 +18,12 @@ import {
   ModalBody,
   useTheme,
   Search,
+  alpha,
 } from '@sistent/sistent';
 import { errorHandlerGenerator, successHandlerGenerator } from '../utils/helpers/common';
 import { useLazyPingKubernetesQuery } from '@/rtk-query/connection';
 import { getK8sConfigIdsFromK8sConfig } from '../utils/multi-ctx';
-import { useEffect, useState } from 'react';
+import { useState, useEffect } from 'react';
```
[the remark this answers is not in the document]
saurabhraghuvanshii: @ashwanisingh011  same you're using useMemo but I haven't see import. Please address this.

ashwanisingh011:
@saurabhraghuvanshii / @gemini-code-assist[bot] - All set!

DashboardMeshModelGraph.tsx: Removed router to prevent unnecessary recalculations on route changes. Replaced the data array literal with primitive count variables for the dependency array to ensure referential equality, and scoped the data variable inside the memoized function.

ConfirmationModal.tsx: Refactored to use derived state (useMemo) for context filtering, resolving the exhaustive-deps warning at the source while preserving component state lifecycles. Also fixed the search bug in SelectDeploymentTarget_ to ensure it renders the filtered list properly.

saurabhraghuvanshii (dismissed):

PR #17896, opened 2026-03-11
fix(server): enable greedy search in GetAllMeshmodelRelationships handler

### LINKED ISSUE #17895: [bug] Search parameter ignored in GetAllMeshmodelRelationships API endpoint
## Description
The `GET /api/meshmodels/relationships` endpoint accepts a `?search=true` query parameter but silently ignores it. The search value is discarded in `getPaginationParams()` result, meaning greedy/fuzzy search never gets applied.

## Expected Behavior
When `?search=true` is passed, the endpoint should perform a greedy search — consistent with how `GET /api/meshmodels/models/{model}/relationships/{name}` already handles it.

## Actual Behavior
The `search` parameter is captured but thrown away with `_` in the function signature:
```go
page, offset, limit, _, order, sort, _ := getPaginationParams(r)
```
As a result, `Greedy` is never set to `true` in the `RelationshipFilter`, so fuzzy search never works.

## Steps to Reproduce
Call `GET /api/meshmodels/relationships?search=true&kind=edge` — the search=true has no effect.

## Possible Fix
Capture the search parameter and pass it to the filter:
```go
page, offset, limit, search, order, sort, _ := getPaginationParams(r)
var greedy bool
if search == "true" {
    greedy = true
}
// then pass Greedy: greedy to RelationshipFilter
```

adityaupasani2:
Fixes #17895

## Description
The `GET /api/meshmodels/relationships` endpoint accepted a `?search=true` query parameter but silently ignored it. The search value was discarded with `_` in the function signature, so greedy/fuzzy search was never applied.

## Changes
- Captured the `search` parameter in `GetAllMeshmodelRelationships`
- Added `greedy` bool logic consistent with `GetMeshmodelRelationshipByName`
- Passed `Greedy: greedy` to the `RelationshipFilter`

## Testing
Call `GET /api/meshmodels/relationships?search=true&kind=edge` — fuzzy search now works as expected, consistent with the other relationship endpoint.

### DIFF
--- a/docs/_includes/test-report.md
+++ b/docs/_includes/test-report.md
@@ -1,15 +1,15 @@
 ### END-TO-END TESTS
 
-- Testing started at: March 11th 2026, 4:55:12 am
+- Testing started at: March 11th 2026, 8:14:47 am
 
 **📦 Test Result Summary**
 
-- ✅ 0 passed
-- ❌ 2 failed
+- ✅ 68 passed
+- ❌ 1 failed
 - ⚠️ 0 flaked
-- ⏩ 0 skipped
+- ⏩ 3 skipped
 
-⌛ _Duration: 0 minutes and 3 seconds_
+⌛ _Duration: 4 minutes and 57 seconds_
 
 **Overall Result**: 👎 Some tests failed.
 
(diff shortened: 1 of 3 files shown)


matrixkavi:
Not a required fix. Added a comment in the issue while closing stating the reason.

PR #17911, opened 2026-03-12
Support ordering tables by record count

### LINKED ISSUE #13958: Add support for sort by count for database table in settings registry page
### Current Behavior
Currently we have sort by name functionality for the table. Sort by count is missing. Meshery server is not handling the case for sort by count.

### Desired Behavior
Add functionality for sort by counts through meshery server.

Image

---

ahmedmadbouly186:
**Notes for Reviewers**

- This PR fixes [#13958](https://github.com/meshery/meshery/issues/13958)

### DIFF
--- a/server/handlers/database_handlers.go
+++ b/server/handlers/database_handlers.go
@@ -7,6 +7,8 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"sort"
+	"strings"
 	"time"
 
 	"github.com/meshery/meshery/server/models"
@@ -38,7 +40,8 @@ func (h *Handler) GetSystemDatabase(w http.ResponseWriter, r *http.Request, _ *m
 	var tables []*models.SqliteSchema
 	var recordCount int
 	var totalTables int64
-	page, offset, limit, search, order, sort, _ := getPaginationParams(r)
+	countOrder := false
+	page, offset, limit, search, order, sortDir, _ := getPaginationParams(r)
 
 	tableFinder := h.dbHandler.DB.Table("sqlite_schema").
 		Where("type = ?", "table")
@@ -49,28 +52,72 @@ func (h *Handler) GetSystemDatabase(w http.ResponseWriter, r *http.Request, _ *m
 
 	tableFinder.Count(&totalTables)
 
-	if limit != 0 {
-		tableFinder = tableFinder.Limit(limit)
-	}
-	if offset != 0 {
-		tableFinder = tableFinder.Offset(offset)
-	}
+	orderBycount := models.SanitizeOrderInput(order, []string{"count"})
 	order = models.SanitizeOrderInput(order, []string{"created_at", "updated_at", "name"})
 	if order != "" {
-		if sort == "desc" {
+		if sortDir == "desc" {
 			tableFinder = tableFinder.Order(clause.OrderByColumn{Column: clause.Column{Name: order}, Desc: true})
 		} else {
 			tableFinder = tableFinder.Order(order)
 		}
+	} else if orderBycount != "" {
+		countOrder = true
 	}
 
-	tableFinder.Find(&tables)
-
-	for _, table := range tables {
-		h.dbHandler.DB.Table(table.Name).Count(&table.Count)
-		recordCount += int(table.Count)
+	readTables := func() {
+		tableFinder.Find(&tables)
+		for _, table := range tables {
+			h.dbHandler.DB.Table(table.Name).Count(&table.Count)
+		}
 	}
 
+	if !countOrder {
+		if limit != 0 {
+			tableFinder = tableFinder.Limit(limit)
+		}
+		if offset != 0 {
+			tableFinder = tableFinder.Offset(offset)
+		}
+
+		readTables()
+		for _, table := range tables {
+			recordCount += int(table.Count)
+		}
+	} else {
+		readTables()
+		desc := strings.Contains(strings.ToLower(orderBycount), "desc")
+
+		sort.Slice(tables, func(i, j int) bool {
+			if desc {
+				return tables[i].Count > tables[j].Count
+			}
+			return tables[i].Count < tables[j].Count
+		})
+
+		// Apply pagination after sorting by count to ensure consistent ordering.
+		if limit != 0 {
+			start := offset
+			if start < 0 {
+				start = 0
+			}
+
+			if start >= len(tables) {
+				tables = []*models.SqliteSchema{}
+			} else {
+				end := start + limit
+				if end > len(tables) {
+					end = len(tables)
+				}
+				tables = tables[start:end]
+			}
+		}
+
+		// Recalculate recordCount for the paginated result set.
+		recordCount = 0
+		for _, table := range tables {
+			recordCount += int(table.Count)
+		}
+	}
 	databaseSummary := &models.DatabaseSummary{
 		Page:        page,
 		PageSize:    limit,
--- a/server/handlers/database_handlers.go
+++ b/server/handlers/database_handlers.go
@@ -16,7 +16,6 @@ import (
 	"github.com/meshery/meshkit/utils"
 	meshsyncmodel "github.com/meshery/meshsync/pkg/model"
 	"github.com/spf13/viper"
-	"gorm.io/gorm/clause"
 )
 
 // swagger:route GET /api/system/database GetSystemDatabase idGetSystemDatabase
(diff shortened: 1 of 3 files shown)


miacycle:
@copilot contrast this approach against the most common (standardized?) approach taken in the rest of the code base.

server/handlers/database_handlers.go
```diff
@@ -49,28 +52,68 @@ func (h *Handler) GetSystemDatabase(w http.ResponseWriter, r *http.Request, _ *m
...
+
+			if start >= len(tables) {
+				tables = []*models.SqliteSchema{}
+			} else {
+				end := start + limit
+				if end > len(tables) {
+					end = len(tables)
+				}
+				tables = tables[start:end]
+			}
```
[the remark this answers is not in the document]
ahmedmadbouly186: If there is no limit, then all the tables will be returned, because the offset itself is calculated based on the limit 
`offset = page * limit`

ui/components/DatabaseSummary.tsx
```diff
@@ -105,8 +105,8 @@ const DatabaseSummary = (props) => {
       if (searchText) setPage(0);
       setSearchText(searchText != null ? searchText : '');
     }),
-    onColumnSortChange: (_, direction) => {
-      setSortOrder(`name ${direction}`);
+    onColumnSortChange: (column, direction) => {
+      setSortOrder(`${column} ${direction}`);
```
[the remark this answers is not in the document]
ahmedmadbouly186: Don't think this enhancement is in the scope of this PR

server/handlers/database_handlers.go
```diff
@@ -49,28 +52,68 @@ func (h *Handler) GetSystemDatabase(w http.ResponseWriter, r *http.Request, _ *m
...
 
 	tableFinder.Find(&tables)
 
 	for _, table := range tables {
 		h.dbHandler.DB.Table(table.Name).Count(&table.Count)
-		recordCount += int(table.Count)
+		if !countOrder {
+			recordCount += int(table.Count)
+		}
 	}
```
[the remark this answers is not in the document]
ahmedmadbouly186: I don't think this enhancement will make a major difference in performance, but anyway, I will open a separate issue for it.

server/handlers/database_handlers.go
```diff
@@ -49,28 +52,72 @@ func (h *Handler) GetSystemDatabase(w http.ResponseWriter, r *http.Request, _ *m
...
+			tableFinder = tableFinder.Offset(offset)
+		}
+
+		readTables()
+		for _, table := range tables {
+			recordCount += int(table.Count)
+		}
+	} else {
+		readTables()
+		desc := strings.Contains(strings.ToLower(orderBycount), "desc")
```
SuryanshGarg04, line 84: Suggestion: `strings.EqualFold(orderByCount, "desc")` might be preferable here.
Using `strings.Contains` could unintentionally match values like "descending" or "desc123" . 
Thoughts?

ahmedmadbouly186: `orderByCount` consists of at least two words (e.g., `"count desc"`). To use `strings.EqualFold`, we would first need to extract the substring where `"desc"` appears.

However, I’m not sure whether `"desc"` will always be the second word. What do you think?

SuryanshGarg04: Quick clarification: should the sort direction (`desc`) come from
`orderBycount` or from `sortDir`?
The API docs here mention `?sort=[asc/desc]`, so I wanted to confirm
how the direction is expected to be passed for count-based sorting.

ahmedmadbouly186: It’s a bit confusing how the sorting parameters are handled.

First, `sortDir` is parsed [[here](https://github.com/meshery/meshery/blob/8450c43ddb8a7ce530aeb53f2a15bf9f614f65f4/server/handlers/utils.go#L27)](https://github.com/meshery/meshery/blob/8450c43ddb8a7ce530aeb53f2a15bf9f614f65f4/server/handlers/utils.go#L27) from the `sort` params. At the same time, the default value defined [[here](https://github.com/meshery/meshery/blob/a585f63cab75513f4354a9504cbd193cf52bbbcc/server/handlers/utils.go#L35)](https://github.com/meshery/meshery/blob/a585f63cab75513f4354a9504cbd193cf52bbbcc/server/handlers/utils.go#L35) is `updated_at`, which is neither `asc` nor `desc`.

On the frontend side, the functionality implemented [[here](https://github.com/meshery/meshery/blob/4e4a2c7d5e998da32d6671df3f02a9978754cc9c/ui/components/DatabaseSummary.tsx#L33-L38)](https://github.com/meshery/meshery/blob/4e4a2c7d5e998da32d6671df3f02a9978754cc9c/ui/components/DatabaseSummary.tsx#L33-L38) sends both `orderByColumn` (`name` / `count`) and `orderByDirection` (`desc` / `asc`) to the `order` field, and completely ignores the `sort` field.

Based on this, I’m guessing that the behavior might have changed at some point on either the frontend or backend side without the documentation being updated.

ahmedmadbouly186: So we have one of two possible solutions:

1. **Ignore the `sort` parameter** — since `order` is already supported by both the backend and frontend. In this case, we can remove `sort` from the documentation and clean up the related code on the backend side.

2. **Reuse the `sort` parameter** — by sending it from the frontend and handling it correctly on the backend in both cases (sorting by `name` and by `count`).

SuryanshGarg04: Thanks for checking the frontend and clarifying this, that clears up my confusion.

SuryanshGarg04: However, I might be missing some frontend context here, but since the UI already
seems to be sending both the column and direction through the `order`
field (e.g. "count desc"), updating the documentation to reflect this
behavior might be the simpler approach compared to introducing or
reusing the `sort` parameter.
That said, it would probably be good to get a maintainer’s opinion here
as well, especially regarding consistency with other endpoints.

ahmedmadbouly186: the suggests solution to use the correct field FE and BE

l5io:
🚀 Deployed on https://69c86cf0f54d545f14ff069a--meshery-docs-preview.netlify.app

l5io:
🚀 Preview for commit 322281839832d0026171c4cd0df34957ec6b7fa9 at: https://69c46e2b8748eef78c0f628a--meshery-docs-preview.netlify.app

ui/components/DatabaseSummary.tsx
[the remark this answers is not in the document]
hortison: This isn't a suggested as an optional enhancement. This is highlighted a bug in your logic. This concern needs to be addressed, @ahmedmadbouly186.

hortison (changes requested):

l5io:
🚀 Preview for commit ff0636489484f8ccbbe6fc4db5b19158beceffc8 at: https://69c86c65b878207de78f8d8d--meshery-docs-preview.netlify.app

l5io:
🚀 Preview for commit 69ff5c553631682da5d99db799c5d2d3c450f430 at: https://69c86cf0f54d545f14ff069a--meshery-docs-preview.netlify.app

leecalcote (commented):
Good feature. Ordering by record count is useful for identifying large tables. The approach of fetching all tables when sorting by count (since count isn't a DB column) and then applying in-memory sort + pagination is the right tradeoff. The `SanitizeOrderInput` usage correctly prevents SQL injection on the order parameter. The UI changes to combine order and direction into a single parameter simplify the API. Note: the `countOrder` path fetches all tables before paginating - for databases with many tables this could be slow, but for SQLite schemas this is fine in practice.

leecalcote (commented):
Correction to my earlier comment: the current approach of looping over tables and issuing a separate `h.dbHandler.DB.Table(table.Name).Count(&table.Count)` per table in Go is inefficient. The database should do the heavy lifting — a single query can fetch table names with their row counts (e.g., using SQLite's `sqlite_stat1` or a subquery approach) rather than N+1 round-trips. Similarly, sorting by count should ideally be done in SQL rather than fetching all tables into memory and sorting with `sort.Slice` in Go. Pushing this work to the database will scale better and is the right architectural pattern.

lekaf974:
I am also concerned by this approach to get the tables count, like mentioned by @leecalcote using SQL is more efficient.

lekaf974:
@ahmedmadbouly186 still working on this one ?

PR #17912, opened 2026-03-12
[mesheryctl] Use MakeRequest for set‑context API calls

dipanshurdev:
## Description

Previously, mesheryctl system config and mesheryctl connection create did not validate the HTTP status code returned by the set-context API. As a result, the CLI could appear to succeed even when the API returned an error response (for example, 4xx or 5xx).

This change adds explicit HTTP status code validation and ensures that failures from the API are properly surfaced to the user.

If the API responds with a non-2xx status code, the command now returns an error that includes the status code and response body. The raw response body is still logged at debug level for troubleshooting, but it is no longer treated as a successful operation.

## Changes

- Validate HTTP status codes for responses from the set-context API.

- Treat non-2xx responses as errors.

- Return a clear error including the status code and response body.

- Keep the raw API response only in debug logs.

- Apply the same behavior to both CLI paths:

  - mesheryctl system config (deprecated path)

  - mesheryctl connection create (current path)

## Impact

- Prevents CLI commands from reporting success when the API call actually failed.
- Improves reliability and user feedback in mesheryctl.
- Helps users quickly understand when context configuration fails while still preserving debug information for troubleshooting.

## Testing

- Verified that: 
- Successful API responses (2xx) continue to complete normally.
- Non-2xx responses return a clear CLI error including the response body.
- Raw API responses remain available through debug logs.


dipanshurdev:
> Thanks. What validated issue does this fix?

Thanks for checking. This isn’t tied to a specific issue number, it addresses the existing **TODOs** and a UX gap.  The set‑context response was only visible in debug logs, so users had no readable confirmation. Now the response is printed in a clean JSON format with proper error handling.

lekaf974:
@dipanshurdev thanks for this PR, first you should create an issue related to this PR with description and details, 
also display api response for end user in JSON does not make any sense IMO and must be used for debugging
As a enduser when I create a connection what I need to know 

- connection is created :heavy_check_mark: 
- connection creation failed, I am informed with a proper helpful error message.

dipanshurdev:
> @dipanshurdev thanks for this PR, first you should create an issue related to this PR with description and details, also display api response for end user in JSON does not make any sense IMO and must be used for debugging As a enduser when I create a connection what I need to know
> 
> * connection is created ✔️
> * connection creation failed, I am informed with a proper helpful error message.

Thanks for the feedback. I’ve removed the JSON output for end users and added explicit error handling with status‑code checks. For success/failure, the CLI now relies on the existing “connection created”/error messaging.

mesheryctl/internal/cli/root/connections/create.go
```diff
@@ -319,8 +319,10 @@ func setContext(configFile, cname string) error {
 	if err != nil {
 		return utils.ErrReadResponseBody(err)
 	}
-	// TODO: Pretty print the output
 	utils.Log.Debugf("Set context API response: %s", string(body))
+	if res.StatusCode < http.StatusOK || res.StatusCode >= http.StatusMultipleChoices {
+		return errors.Errorf("failed to set context: status code %d: %s", res.StatusCode, strings.TrimSpace(string(body)))
```
lekaf974: this should be a meshkit error also it should be done just after client.Do

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -120,8 +121,10 @@ func setContext(configFile, cname string) error {
 	if err != nil {
 		return utils.ErrReadResponseBody(err)
 	}
-	// TODO: Pretty print the output
 	log.Debugf("Set context API response: %s", string(body))
+	if res.StatusCode < 200 || res.StatusCode >= 300 {
```
lekaf974: api.Add has already a centralized method where status code are checked to generate errors

dipanshurdev:
> added other comments

Removed the manual status‑code checks in system/config.go since api. Add already routes through MakeRequest and returns meshkit errors. In connections/create.go I switched to utils.MakeRequest right after building the request, so status handling is centralized and error types are consistent.

shri771:
@dipanshurdev Could you update the PR title as well? The current title no longer reflects the changes in this PR.

dipanshurdev:
> @dipanshurdev Could you update the PR title as well? The current title no longer reflects the changes in this PR.

Sure

shri771:
@dipanshurdev, Also resolve the conflicts.

dipanshurdev:
> @dipanshurdev, Also resolve the conflicts.

Updated the PR title to “[mesheryctl] Use MakeRequest for set‑context API calls” to reflect the current changes.

lekaf974 (approved):

PR #17970, opened 2026-03-14
Mui to Sistent Icons

Bhumikagarggg:
**Notes for Reviewers**
These icons are already available in the Sistent icon library, so I replaced the existing MUI icons with the corresponding Sistent icons.

- This PR fixes #

### DIFF
--- a/ui/next.config.js
+++ b/ui/next.config.js
@@ -27,6 +27,7 @@ const nextConfig = {
     '@uiw/react-markdown-preview',
     '@uiw/react-codemirror',
     'billboard.js',
+    '@sistent/sistent',
   ],
 
   // SWC Compiler Configuration (SWC is the default in Next.js 15)


ui/components/configuratorComponents/MeshModel/index.tsx
```diff
@@ -1,20 +1,21 @@
 import {
+  ArrowBack,
   Avatar,
   AvatarGroup,
   CircularProgress,
+  Delete,
```
YASHMAHAKAL, line 6: Should it be `DeleteIcon` ? 
The CI is failing because several files import icons without the Icon suffix that Sistent requires .  Delete, Add, Edit should be  DeleteIcon, AddIcon, EditIcon causing the Next.js build to silently produce an empty artifact.

Bhumikagarggg: But in Sistent repo `DeleteIcon`, `AddIcon`, `EditIcon`  is export as `Delete`, `Add`, `Edit`
You can verify from this https://github.com/layer5io/sistent/blob/master/src/icons/index.ts

YASHMAHAKAL: file you've shared actually points to a folder './', for example - export * from './Delete'; and inside ./Delete directory we have - export { default as DeleteIcon } from './DeleteIcon';

PR #18048, opened 2026-03-18
[Server] Prevent nil preference panic during session sync

fitzergerald:
## Summary

This fixes a panic in the session sync path by making preference loading consistently return a usable default preference and by hardening the middleware fallback.

## Why this panic happens

`SessionSyncHandler` assumes `prefObj` is non-nil and writes to `prefObj.MeshAdapters`.

That assumption could be broken earlier in the request path:

- `SessionInjectorMiddleware` loads preferences through `provider.ReadFromPersister(user.UserId)`.
- The DB-backed `SessionPreferencePersister` did **not** behave like the in-memory `MapPreferencePersister` when no preference row existed.
- On top of that, it unmarshaled stored preference JSON into `&data` (`**Preference`) instead of `data` (`*Preference`).

That combination created two ways for a nil preference to leak through:

1. A user with no persisted preference row would not reliably get the default preference object that other code paths expect.
2. A stored JSON `null` value could silently turn the preference pointer itself into `nil` during unmarshal.

Once that nil reached `SessionSyncHandler`, the handler panicked while assigning `prefObj.MeshAdapters`.

## Fix

This PR addresses the root cause and adds a defensive fallback:

- add `models.NewDefaultPreference()` so all code paths share the same default preference initialization
- make the DB-backed preference persister return that default preference on `gorm.ErrRecordNotFound`
- unmarshal stored preference JSON into `*Preference` instead of `**Preference`
- reuse the same default helper in the map-backed persister for consistency
- make `SessionInjectorMiddleware` fall back to a default preference if loading fails or unexpectedly returns `nil`
- add regression tests for:
  - missing preference rows
  - normal persisted preferences
  - stored JSON `null`

## Validation

- `go test ./server/models -run 'TestSessionPreferencePersister|TestGetComponentMetadataWithNilModel'`
- `go test ./server/handlers`

### DIFF
--- a/server/handlers/middlewares.go
+++ b/server/handlers/middlewares.go
@@ -194,11 +194,11 @@ func (h *Handler) SessionInjectorMiddleware(next func(http.ResponseWriter, *http
 		prefObj, err := provider.ReadFromPersister(user.UserId)
 		if err != nil {
 			// log underlying error from persister along with high-level context
-			h.log.Warn(ErrReadSessionPersistor, err, fmt.Sprintf("userID=%s", user.UserId))
+			h.log.Warn(fmt.Errorf("%w: userID=%s: %v", ErrReadSessionPersistor, user.UserId, err))
 			prefObj = models.NewDefaultPreference()
 		} else if prefObj == nil {
 			// persister unexpectedly returned a nil preference without error
-			h.log.Warn(fmt.Sprintf("persister returned nil preference without error for userID=%s", user.UserId))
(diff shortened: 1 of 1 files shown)


yi-nuo426:
@copilot fix 

```
323.6 # github.com/meshery/meshery/server/handlers
323.6 ../handlers/middlewares.go:197:40: too many arguments in call to h.log.Warn
323.6 	have (*"github.com/meshery/meshkit/errors".Error, error, string)
323.6 	want (error)
323.6 ../handlers/middlewares.go:201:15: cannot use fmt.Sprintf("persister returned nil preference without error for userID=%s", user.UserId) (value of type string) as error value in argument to h.log.Warn: string does not implement error (missing method Error)
```

PR #18049, opened 2026-03-18
chore: update to proper rjsf

saurabhraghuvanshii:
**Notes for Reviewers**

- This PR fixes #


saurabhraghuvanshii:
This PR is important because many components are deprecated in the new RJSF, so we need to update RJSF properly to use the new components.

However, this needs to be verified with Kanvas first, as I believe Kanvas is not yet fully updated to RJSF.

PR #18076, opened 2026-03-20
Update license expression and fix baseline findings security insights

sangramrath:
**Notes for Reviewers**

- This PR fixes #

### DIFF
--- a/.github/workflows/copilot-pr-handler.yml
+++ b/.github/workflows/copilot-pr-handler.yml
@@ -78,7 +78,9 @@ jobs:
           echo "PR Author: ${{ github.event.pull_request.user.login || 'N/A' }}"
           echo "PR Draft Status: ${{ github.event.pull_request.draft || 'N/A' }}"
           echo "PR Head SHA: ${{ github.event.pull_request.head.sha || 'N/A' }}"
-          echo "PR Head Ref: ${{ github.event.pull_request.head.ref || 'N/A' }}"
+          # Sanitize the head ref before echoing
+          head_ref="${{ github.event.pull_request.head.ref || 'N/A' }}"
+          echo "PR Head Ref: ${head_ref//[\`\$\(\)]/}"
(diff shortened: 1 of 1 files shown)


.github/workflows/copilot-pr-handler.yml
```diff
@@ -78,7 +78,9 @@ jobs:
           echo "PR Author: ${{ github.event.pull_request.user.login || 'N/A' }}"
           echo "PR Draft Status: ${{ github.event.pull_request.draft || 'N/A' }}"
           echo "PR Head SHA: ${{ github.event.pull_request.head.sha || 'N/A' }}"
-          echo "PR Head Ref: ${{ github.event.pull_request.head.ref || 'N/A' }}"
+          # Sanitize the head ref before echoing
+          head_ref="${{ github.event.pull_request.head.ref || 'N/A' }}"
```
leecalcote, line 82: Straight-forward. Makes sense.

.github/SECURITY-INSIGHTS.yml
```diff
@@ -82,7 +82,7 @@ repository:
       social: https://layer5.io/community/members/sangram-rath/
   license:
     url: https://github.com/meshery/meshery/blob/master/LICENSE
-    expression: Apache License 2.0
+    expression: Apache-2.0
```
leecalcote, line 85: I don't know that we don't also use the "Apache License 2.0" phrase sometimes.

sangramrath: Their rule accepts `Apache-2.0` format only at the moment. I informed Eddie about this too.

leecalcote: Oh, but that doesn't matter, since this isn't a verification like the copilot-pr-handler.yml is.

This is just the shorthand license format that the Security Insights tool expects. Got it.

Issue #18104, opened 2026-03-20
[mesheryctl] Nil pointer panic in `system channel switch` when config load fails


ARZU-V:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

### Current Behavior
`mesheryctl system channel switch` panics with a nil pointer dereference if 
`config.GetMesheryCtl()` fails. The error is logged but execution continues, 
and the next line dereferences `mctlCfg` which is nil.
Every other error block in this file (viewCmd, setCmd) correctly returns
nil after logging. This one does not.

### Desired Behavior
When GetMesheryCtl() fails, switchCmd.RunE should log the error and return
nil — consistent with every other command in this file.

Fix is a one-line addition of return nil after utils.Log.Error(err).

### Environment

- Host OS: Mac Linux Windows
- Meshery Server Version: stable-v
- Meshery Client Version: stable-v
- Platform: Docker Kubernetes

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 📚 [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- 📺 [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- ⌨️ [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- ⌨️ [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- ⌨️ [Meshkit errors Guide](https://docs.meshery.io/project/contributing/contributing-error)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

PR #18106, opened 2026-03-20
fix: Fixes nil pointer panic when config.GetMesheryCtl() fails

ARZU-V:
**Notes for Reviewers**

- This PR fixes #18104

### DIFF
--- a/mesheryctl/internal/cli/root/system/channel.go
+++ b/mesheryctl/internal/cli/root/system/channel.go
@@ -210,7 +210,6 @@ mesheryctl system channel switch [stable|stable-version|edge|edge-version]
 
 		mctlCfg, err = config.GetMesheryCtl(viper.GetViper())
 		if err != nil {
-			utils.Log.Error(err)
 			return err
 		}
 		focusedContext := tempContext


mesheryctl/internal/cli/root/system/channel.go
lekaf974:
```suggestion
			return err
```

lekaf974 (changes requested):
return the error please

mesheryctl/internal/cli/root/system/channel.go
```diff
@@ -211,6 +211,7 @@ mesheryctl system channel switch [stable|stable-version|edge|edge-version]
 		mctlCfg, err = config.GetMesheryCtl(viper.GetViper())
 		if err != nil {
 			utils.Log.Error(err)
```
lekaf974:
```suggestion
```
please just return the error

ARZU-V: I noticed the codebase follows the pattern of utils.Log.Error(err) + return nil in most places, so I kept that consistent.I will return the err as per you said.

Issue #18100, opened 2026-03-20
[UI] Feedback Modal Buttons Become Invisible When Clicked


dipanshurdev:
## Description

In the feedback Modal (right-side floating modal), the top action buttons have a UI issue when clicked.

The buttons normally appear with a dark background and white text, but when a button is clicked, its background turns white, while the text remains white, making the label invisible.

This results in poor visibility and a confusing user experience.

## Expected Behavior

- Button text should always remain visible and readable

- When a button is clicked (active state), it should: either maintain proper contrast or update both background and text color appropriately

## Actual Behavior

- On click: Background becomes white Text remains white
-Result: text becomes invisible

## Steps to Reproduce

- Open the https://playground.meshery.io/extension/meshmap
- Locate the feedback drawer on the right side
- Click to expand it
- Click any of the top buttons

## Screenshots / Recording

https://github.com/user-attachments/assets/db5afc4c-3849-4d00-a433-9e518d35b17f

## Possible Cause

- This might be due to incorrect styling in the active/focus state, where: background-color changes but color is not updated accordingly

sm-28601:
I’d like to work on this issue. please assign it to me?

dhakshin-raghav:
I'd like to take this on! I’ve identified that the background-color is switching to white on click, but the text color isn't updating to a contrasting shade. I can submit a PR to fix the CSS for these action buttons to ensure the labels remain visible. Please assign this to me.

experiencei:
I would be working on this particular issues , Kindly assign this Issue to me. Appreciate

banana-three-join:
Hi! The component's improper behavior was due to styling issues with a component which is important from Sistent. Here's the [Closed issue](https://github.com/layer5io/sistent/issues/1373) so now the component should be working appropiately and the issue can be closed!

rishiraj38:
@dhakshin-raghav any updates?

dhakshin-raghav:
@rishiraj38 The problem is not with this meshery repo ,its with the component Sistent and the problem is fixed there .so,there is no change needed here

banana-three-join:
Just wanted to add in to the topic but the button itself keeps on presenting the same issue on the Playground because I believe the version of sistent which is currently being used on the Playground is behind from the current version that has the fix to this issue. Hosting a local server with the latest Sistent shows that the issue is no longer present, in case someone who only uses playground is confused or has any questions about the issue!

dipanshurdev:
[...]

PR #18131, opened 2026-03-22
fix(ui): enforce contrast on active feedback modal buttons

dhakshin-raghav:
Fixes #18100

**Notes for Reviewers**

Root Cause: Feedback modal buttons (imported via the external @sistent UI package) lacked proper contrast styling for active states. When clicked (e.g. "Bug Report"), the component generated a white background leaving the default white text invisible.
Implemented Fix: We injected <GlobalStyles /> inside 
StickyFeedbackButton
 (
ui/components/General/feedback.tsx
) strictly targeting .MuiDrawer-root nested elements.
Button tabs (.MuiToggleButton-root.Mui-selected) forcefully assert a dark text foreground and light background !important.
Normal buttons (.MuiButtonBase-root:active) forcefully assert dark text foreground.
Validation: Enforced via ESLint & TS typings validation checks against 
feedback.tsx
 ensuring standard MUI syntax.

- This PR fixes #18100 

### DIFF
--- a/ui/components/General/feedback.tsx
+++ b/ui/components/General/feedback.tsx
@@ -1,6 +1,7 @@
 import { useHandleFeedbackFormSubmissionMutation } from '@/rtk-query/user';
 import { useNotification } from '@/utils/hooks/useNotification';
 import { FeedbackButton, styled } from '@sistent/sistent';
+import { GlobalStyles } from '@mui/material';
 import { EVENT_TYPES } from 'lib/event-types';
 import _ from 'lodash';
 import { useRouter } from 'next/router';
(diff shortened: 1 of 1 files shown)


banana-three-join:
Hi! I've already submitted a PR which fixes the issue and I'm just waiting on the review for it to be merged. 

The problem is with the styling with the component from Sistent. So, imposing !important stylings onto the component within Meshery UI isn't correct as Sistent is a reusable component library that's utilized across Meshery UI. 

This means that if the same component is reutilized elsewhere, the same problem will still happen. So the correct way on how to address the issue is by updating the styling with Sistent. 

Here's the [PR](https://github.com/layer5io/sistent/pull/1374) in case you want to check the library!

PR #18157, opened 2026-03-23
Fix:local provider boot hang

hortison:
Signed-off-by: hortison <160366376+hortison@users.noreply.github.com>

 Changed:

   - ui/rtk-query/index.ts: switched to the same mesheryApi instance the generated hooks use.
   - ui/rtk-query/user.ts: normalized JSON-like responses, fixed current-user/profile queries to use Meshery’s actual endpoints.
   - ui/pages/_app.tsx: bootstrap now logs failures and always clears the outer loader.
   - ui/rtk-query/ability.tsx: added delayed session-guard debug logging if loading persists.
   - Server handlers/routes now return proper JSON for /api/user, /api/user/prefs, /api/provider/capabilities, and added /api/identity/users/profile aliases.

  Validated with npm run lint, npm run build && npm run export, go test ./server/handlers ./server/router, go test ./server/models -run '^$', and a Playwright login on a clean local-provider server:
  dashboard rendered, loader hid.

### DIFF
--- a/ui/rtk-query/user.ts
+++ b/ui/rtk-query/user.ts
@@ -100,6 +100,9 @@ export const userApi = api
           url: '/api/user',
           method: 'GET',
         }),
+        // All callers share one cache entry per user session (client-side Redux store).
+        // This does not affect other users—each browser has its own isolated store.
+        serializeQueryArgs: ({ endpointName }) => endpointName,
       }),
       getProviderCapabilities: builder.query({
         query: () => '/api/provider/capabilities',


ui/rtk-query/user.ts
```diff
@@ -97,13 +109,21 @@ export const userApi = api
           }
         },
       }),
+      getLoggedInUser: builder.query({
+        query: () => ({
+          url: '/api/user',
+          method: 'GET',
+        }),
+        transformResponse: parseJsonLikeResponse,
```
leecalcote: Why does this continue to be necessary? This seems like a work-around and not a fix for the root cause:  `transformResponse: parseJsonLikeResponse`

ui/rtk-query/ability.tsx
```diff
@@ -64,6 +66,37 @@ const SelectedOrganizationProvider = ({ children }) => {
...
+    providerCapabilities?.provider_type,
+    selectedOrganization,
+    selectedOrganizationId,
+    didFallback,
+    isFetchingSelectedOrg,
+    isLoadingAbilities,
+    errorFetchingSelectedOrg,
+    errorLoadingAbilities,
+  ]);
+
```
leecalcote, line 138: The loading screen has 5 seconds to load before timing
out? And then, that's it? No allowance for a slower machine? No user wait-indicator? No indication as to why the timeout was reached?

leecalcote: Server endpoints returned text/plain for JSON payloads breaking RTK Query parsing. A few of the server endpoints are patched in this PR, so maybe the root of the issue has been fixed already.

ui/rtk-query/user.ts
```diff
@@ -97,13 +109,21 @@ export const userApi = api
           }
         },
       }),
+      getLoggedInUser: builder.query({
+        query: () => ({
+          url: '/api/user',
+          method: 'GET',
+        }),
```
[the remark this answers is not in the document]
marblom007: @copilot+gpt-5.4 recognize that Meshery is a multi-user (concurrent multi-user) tool. Does the proposed serialization affect only a single user session or have a broader impact, potentially sideswiping other sessions?

PR #18162, opened 2026-03-24
local provider boot hang

hortison:
Fix:

 I updated `server/models/preference_map_persister.go` so a missing map-backed preference read now seeds `NewDefaultPreference()` back into the persister before returning it. That matches the test  contract behind `TestMapPreferencePersisterReadFromPersisterReturnsDefaultsWhenMissing`.

- **[Server] Seed default map preferences on read miss**

This fix is interrelated with @lekaf974's #18161

### DIFF
--- a/server/models/preference_map_persister.go
+++ b/server/models/preference_map_persister.go
@@ -60,7 +60,8 @@ func (s *MapPreferencePersister) ReadFromPersister(userID string) (*Preference, 
 		}
 	} else {
 		log.Debug(ErrSessionNotFound(userID))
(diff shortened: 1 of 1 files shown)


pontusringblom (approved):
looks good

lekaf974:
@hortison I came with same approach in my other PR after re-read gemini comment

lekaf974:
Also since Method ReadFrom now also WriteToPersister I think this method should be renamed to reflect the change

marblom007 (approved):

PR #18181, opened 2026-03-24
UI: fix workspace API endpoint URLs

leecalcote:
## Summary

- Fix workspace RTK query endpoints to use `extensions/api/workspaces/...` prefix consistently across all operations (list, create, update, delete, assign/unassign environments, designs)
- Skip `WorkspaceActivityWidget` workspace query when no org ID is set (prevents 404 on local provider)

## Root cause

The bare `workspaces/...` relative paths in RTK Query resolved page-relative: from `/extension/meshmap` they resolved to `/extension/workspaces/...` (200 but wrong content), and from `/` they resolved to `/workspaces/...` (404 on playground). The `extensions/api/workspaces/...` prefix already used by views, teams, and events endpoints is the correct cloud-layer route.

## Test plan

- [ ] Visit dashboard on playground.meshery.io — no 404 for `/workspaces`
- [ ] Workspace list, create, update, delete all route to `/extensions/api/workspaces/...`
- [ ] Dashboard `WorkspaceActivityWidget` does not fire on local provider (no orgID)

### DIFF
--- a/ui/rtk-query/workspace.ts
+++ b/ui/rtk-query/workspace.ts
@@ -22,7 +22,7 @@ const workspacesApi = api
           const { expandInfo, ...otherArgs } = queryArgs;
           const params = urlEncodeParams(otherArgs);
           const workspaces = await baseQuery({
-            url: `workspaces?${params}`,
+            url: `extensions/api/workspaces?${params}`,
             method: 'GET',
           });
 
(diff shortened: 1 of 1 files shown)


hortison:
Closing as superseded — `ui/rtk-query/workspace.ts` has since been rewritten by later work (#21175, #21184): URLs now go through the general-purpose `mesheryApiPath()` helper, and the workspace create/update/delete/environment-assignment operations are now the `@meshery/schemas`-generated mutations rather than hand-rolled RTK endpoints. Porting this PR's fix onto current master would mean reintroducing hand-rolled endpoints that duplicate what `@meshery/schemas` already provides, so the right move is closing this rather than rebasing it.

---

Issue #18186, opened 2026-03-24
[Bug] DeleteMeshSyncResource Returns No HTTP Response When Deletion Fails or Succeeds


Manishnemade12:
# PR Description

## Current Behavior
`DeleteMeshSyncResource` in the MeshSync handler does not send an explicit HTTP response after the delete operation.
When DB delete fails, the server only logs the error and the client still receives `200 OK`.
When delete succeeds, the client receives an empty `200 OK` with no confirmation body.

## Expected Behavior
The handler should always return a meaningful HTTP response:
- On failure: `500 Internal Server Error` with an error message in the response body.
- On success: `200 OK` with a JSON confirmation response (example: `{"deleted": true}`).

## Screenshots/Logs
N/A (issue identified by code inspection)

- Affected code: `server/handlers/meshsync_handler.go` (lines 554-561)

## Environment
- **Host OS:** macOS / Linux / Windows
- **Platform:** Docker or Kubernetes
- **Meshery Server Version:** stable-v (affected)
- **Meshery Client Version:** stable-v (affected)

## To Reproduce
1. Call the MeshSync delete resource endpoint with a resource ID.
2. Trigger a DB failure case (or use a state that causes delete to fail).
3. Observe the client still receives `200 OK`.
4. In success case, observe the response is still an empty `200 OK` with no explicit confirmation body.

## Impact
- Clients cannot reliably distinguish success vs failure.
- UI may assume deletion succeeded even when DB deletion failed.
- Can lead to silent state inconsistency between UI and backend.

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)
[...]

PR #18193, opened 2026-03-25
[Server] Fix DeleteMeshSyncResource HTTP response handling

Manishnemade12:
## [Server] Fix DeleteMeshSyncResource HTTP response handling

- This PR fixes #18186
### Description
This PR fixes a bug in `DeleteMeshSyncResource` where the handler did not send an explicit HTTP response body on success and also returned `200 OK` even when deletion failed.

### Related Issue
Closes #18186

### What Changed
- Updated `DeleteMeshSyncResource` in `server/handlers/meshsync_handler.go`.
- On delete failure:
  - logs the delete error
  - returns `500 Internal Server Error` with error message in response body
- On delete success:
  - returns `200 OK`
  - returns JSON response: `{"deleted": true}`

### Why This Change
Previously, clients had no reliable way to detect failure because failed deletes still looked successful (`200 OK`).
This change makes API behavior explicit and consistent for both success and failure paths.

### Validation
- Added/updated handler test coverage for `TestDeleteMeshSyncResource`.
- Verified with:
  - `go test ./server/handlers -run TestDeleteMeshSyncResource -count=1`
  - `go test ./server/handlers -count=1`

### Before / After Behavior
- **Before**:
  - success: empty `200 OK`
  - failure: still `200 OK` (error only in server logs)
- **After**:
  - success: `200 OK` with `{"deleted": true}`
  - failure: `500 Internal Server Error` with error message

---

**Notes for Reviewers**


Manishnemade12:
@Bhumikagarggg can you please review this pr

server/handlers/meshsync_handler.go
```diff
@@ -556,6 +556,19 @@ func (h *Handler) DeleteMeshSyncResource(rw http.ResponseWriter, r *http.Request
...
-		h.log.Error(models.ErrDelete(err, "meshsync data", http.StatusInternalServerError))
+		deleteErr := ErrFailToDelete(err, "meshsync resource")
+		h.log.Error(deleteErr)
+		http.Error(rw, deleteErr.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	rw.Header().Set("Content-Type", "application/json")
+	rw.WriteHeader(http.StatusOK)
+if err := json.NewEncoder(rw).Encode(struct{ Deleted bool `json:"deleted"` }{Deleted: true}); err != nil {
```
ahmedmadbouly186: nit: add tab here

Manishnemade12: done

server/handlers/provider_handler_test.go
```diff
@@ -219,3 +225,67 @@ func TestLoginHandler_NoneProviderFromMiddleware(t *testing.T) {
...
+			if contentType := rw.Header().Get("Content-Type"); !strings.HasPrefix(contentType, tt.expectedContentType) {
+				t.Fatalf("expected content type prefix %q, got %q", tt.expectedContentType, contentType)
+			}
+
+			if body := rw.Body.String(); !strings.Contains(body, tt.expectedBody) {
+				t.Fatalf("expected body to contain %q, got %q", tt.expectedBody, body)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
Manishnemade12: I agree that TestDeleteMeshSyncResource could be moved to a dedicated meshsync_handler_test.go file as a follow-up for better organization. For this PR, I kept it in provider_handler_test.go to reuse the existing newTestHandler helper and avoid unrelated test-file refactoring, since the scope here is the bug fix in DeleteMeshSyncResource and its regression coverage.

Manishnemade12:
> Clean fix. Returning `500` with an error message on failure and `200` with `{"deleted": true}` on success gives clients a reliable contract. The test suite is well-structured — using in-memory SQLite with and without migration to simulate success/failure is a solid approach. The `isClientDisconnect` guard on encode errors is a nice detail that avoids noisy error logs from dropped connections.

Thanks for the review and kind words, really appreciate it! 🙏

Manishnemade12:
@lekaf974 can you please take review for this pr once

server/handlers/provider_handler_test.go
```diff
@@ -219,3 +226,81 @@ func TestLoginHandler_NoneProviderFromMiddleware(t *testing.T) {
...
+
+			if tt.migrateResourceTable {
+				if err := db.AutoMigrate(&model.KubernetesResource{}); err != nil {
+					t.Fatalf("failed to migrate kubernetes resource table: %v", err)
+				}
+			}
+
+			provider := &models.DefaultLocalProvider{}
+			provider.Initialize()
+			provider.GenericPersister = &database.Handler{DB: db}
```
lekaf974, line 271: should we move this part outside the for loop ?

server/handlers/provider_handler_test.go
```diff
@@ -219,3 +226,81 @@ func TestLoginHandler_NoneProviderFromMiddleware(t *testing.T) {
...
+
+	tests := []struct {
+		name                 string
+		migrateResourceTable bool
+		expectedStatus       int
+		expectedContentType  string
+		expectedBody         string
+	}{
+		{
+			name:                 "returns 200 and deleted true when delete succeeds",
```
lekaf974: can you use given when then naming for test please, something like

`given deletion succeeds when DeleteMeshSyncResource then return status 200`

This makes easier to know what we test and the expected behavior

server/handlers/provider_handler_test.go
Manishnemade12: @lekaf974 It is intentional to keep the database setup inside the loop so that each subtest gets a clean, isolated in-memory database. This avoids state leakage between test cases and ensures test reliability.

what do you think ?

server/handlers/provider_handler_test.go
Manishnemade12: done

server/handlers/provider_handler_test.go
lekaf974: make sense

Manishnemade12:
@lekaf974  can you please take final overview then

Manishnemade12:
@PragalvaXFREZ this pr is approved by @lekaf974 . so whenever you  get time  please take a final review and merge it .

lekaf974:
@Manishnemade12 our slack is the right place to request review

PR #18201, opened 2026-03-25
[Server] Avoid panic when latest release tags are unavailable

### LINKED ISSUE #18199: Bug fix: avoid panic in CheckLatestVersion when release tags are unavailable
### Current Behavior
Meshery Server panics during operator deployment config generation when it cannot retrieve the latest Meshery release information (e.g., due to network/auth/rate-limit/proxy issues).

Observed panic:
- `panic: runtime error: index out of range [-1]`
- Trigger point: `server/models/meshery_controllers.go:498` inside `CheckLatestVersion`

### Expected Behavior
Meshery Server should not panic.
Instead, it should handle "failed to fetch latest release info" gracefully (return a controlled error or fall back) so the server can continue starting/deploying the operator config.

### Logs
```text
panic: runtime error: index out of range [-1]
goroutine 1 [running]:
github.com/meshery/meshery/server/models.CheckLatestVersion(...)
 /.../server/models/meshery_controllers.go:498 +0x118
...
```

### Environment
- **Host OS:** macOS
- **Platform:** Docker or Kubernetes (please fill if known)
- **Meshery Server Version:** stable (or specify build channel/commit)
- **Meshery Client Version:** stable (or specify)

### To Reproduce
1. Run Meshery Server in a local/dev environment where `BUILD` can be empty or not set.
2. Ensure operator deployment config generation is triggered (e.g., on server start / operator config creation).
3. Make the latest release info retrieval fail (network/auth/rate-limit/proxy or upstream parsing issues), so the latest release data cannot be obtained.
4. Observe Meshery Server panic at `CheckLatestVersion` when accessing `versions[len(versions)-1]`.

### Notes / Potential Root Cause
The current `CheckLatestVersion` implementation assumes that release data is always available and indexes the last element without guarding against the failure case.

### Proposed Fix (Summary)
Add a defensive check in `CheckLatestVersion()`:
- If release data is missing/unavailable, return a controlled error instead of indexing the slice.

yuuki798:
## Summary
Meshery Server panics during operator deployment config generation when it cannot retrieve the latest Meshery release information (e.g., network/auth/rate-limit/proxy/upstream parsing issues). This results in an out-of-range access when selecting the last element from an empty versions list.

## Fix
Guard `CheckLatestVersion()` so it returns a controlled error instead of panicking when release data is unavailable.

## Why
`CheckLatestVersion()` assumes release data is always available and indexes the last element without guarding against the failure case.

## Test plan
- go test ./server/models -run TestControllerEventActedUpon -count=1

## Related
Fixes #18199


leecalcote (commented):
Good catch — indexing `versions[len(versions)-1]` without checking for an empty slice is a classic panic source, especially since network failures or rate-limiting can cause `GetLatestReleaseTagsSorted` to return nothing. The separate check for `len(versions) == 0` after the error check is correct. One nit: the `if err != nil` block has lost its leading indentation (tab) — looks like a formatting artifact to fix before merge.

yuuki798:
> Good catch — indexing `versions[len(versions)-1]` without checking for an empty slice is a classic panic source, especially since network failures or rate-limiting can cause `GetLatestReleaseTagsSorted` to return nothing. The separate check for `len(versions) == 0` after the error check is correct. One nit: the `if err != nil` block has lost its leading indentation (tab) — looks like a formatting artifact to fix before merge.

Thank you for your recognition.I apologize for not paying attention to the formatting when adopting the CR robot. I have now fixed it urgently. Please review.

PR #17869, opened 2026-03-09
[Server] Fix HTTP response handling in environments handlers

ishan8351:
This PR fixes incorrect HTTP status codes, resolves a WriteHeader ordering issue in SaveEnvironment, and updates error wrappers to correctly reflect the operations being performed.

Fixes #17832

**Notes for Reviewers**

- This PR fixes `SaveEnvironment` where `w.WriteHeader(http.StatusCreated)` was called *after* writing the response body, which was causing implicit 200 OK responses and superfluous header warnings.
- Updated error handlers for Save, Delete, Update, and Connection mapping operations to return `500 Internal Server Error` instead of `404 Not Found`.
- Replaced semantically incorrect `ErrGetResult` wrappers with appropriate wrappers (`ErrFailToSave`, `ErrFailToDelete`) corresponding to the actual operations being performed.

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #18208, opened 2026-03-25
fix(server): correct HTTP response handling in workspace_handlers.go

### LINKED ISSUE #18198: [Bug] Incorrect HTTP response handling in workspace_handlers.go: wrong status codes, WriteHeader ordering, and misused error wrappers
### Description
The file `server/handlers/workspace_handlers.go` has three HTTP response handling bugs such as  giving out incorrect status codes and using semantically wrong error wrappers. These are the same class of bugs fixed in PR #17869 for `environments_handlers.go`, but `workspace_handlers.go` was not included in that fix.

### Bug 1
`SaveWorkspaceHandler`: `Content-Type` header set after `WriteHeader` (Line 120-122)

In `SaveWorkspaceHandler`, `w.WriteHeader(http.StatusCreated)` is called before `w.Header().Set("Content-Type", "application/json")`. In Go's `net/http`, once `WriteHeader` is called, headers are frozen, any subsequent `Header().Set()` call is silently ignored.

Result: Clients always receive a 201 response with no Content-Type header, so clients that check Content-Type before parsing the response body will break.

### Bug 2 
3 handlers return `http.StatusNotFound` (404) for errors that are not "resource not found" errors, they are internal server failures:

| Handler | Line | Operation | Current Status | Correct Status |
|---------|------|-----------|---------------|----------------|
| `SaveWorkspaceHandler` | 113 | POST (create) | 404 Not Found | 500 Internal Server Error |
| `DeleteWorkspaceHandler` | 138 | DELETE | 404 Not Found | 500 Internal Server Error |
| `UpdateWorkspaceHandler` | 177 | PUT (update) | 404 Not Found | 500 Internal Server Error |

</body>
</html>

### Bug 3
Semantically wrong error wrapper: `ErrGetResult` used for non-GET operations.
All three handlers above use `ErrGetResult` which wraps errors with "unable to get result" semantics. This is misleading for save, delete, and update operations:

| Handler | Operation | Current Wrapper | Should Be |
|---------|-----------|----------------|-----------|
| `SaveWorkspaceHandler` | POST (create) | `ErrGetResult` | `ErrFailToSave` |
| `DeleteWorkspaceHandler` | DELETE | `ErrGetResult` | `ErrFailToDelete` |
| `UpdateWorkspaceHandler` | PUT (update) | `ErrGetResult` | `ErrFailToSave` |

### Expected Behavior
- `SaveWorkspaceHandler` should set `Content-Type` header before calling `WriteHeader`.
- Save, delete, and update failures should return `500` Internal Server Error, not `404` Not Found.
- Error wrappers should match the operation being performed.

### Fix 1
 Correct header ordering in `SaveWorkspaceHandler`
```go
w.WriteHeader(http.StatusCreated)
fmt.Fprint(w, string(bf))
```
### Fix 2 & 3
Correct status codes and error wrappers:
```go
// SaveWorkspaceHandler
if err != nil {
    h.log.Error(ErrFailToSave(err, "workspace"))
    http.Error(w, ErrFailToSave(err, "workspace").Error(), http.StatusInternalServerError)
    return
}

// DeleteWorkspaceHandler
if err != nil {
    h.log.Error(ErrFailToDelete(err, "workspace"))
    http.Error(w, ErrFailToDelete(err, "workspace").Error(), http.StatusInternalServerError)
    return
}

// UpdateWorkspaceHandler
if err != nil {
    h.log.Error(ErrFailToSave(err, "workspace"))
    http.Error(w, ErrFailToSave(err, "workspace").Error(), http.StatusInternalServerError)
    return
}
```

### Environment

- Meshery Version: master branch
- File: `server/handlers/workspace_handlers.go`
- Language: Go

---

yug54475:
**Notes for Reviewers**

- Fixes #18198

This PR fixes the same class of HTTP response handling bugs found in `workspace_handlers.go` that were fixed for `environments_handlers.go` in #17869.

Changes made:
- `SaveWorkspaceHandler`: moved `w.Header().Set("Content-Type", "application/json")` before `w.WriteHeader(http.StatusCreated)` to ensure headers are sent correctly
- `SaveWorkspaceHandler`: replaced `ErrGetResult` with `ErrFailToSave` and `http.StatusNotFound` with `http.StatusInternalServerError`
- `DeleteWorkspaceHandler`: replaced `ErrGetResult` with `ErrFailToDelete` and `http.StatusNotFound` with `http.StatusInternalServerError`
- `UpdateWorkspaceHandler`: replaced `ErrGetResult` with `ErrFailToSave` and `http.StatusNotFound` with `http.StatusInternalServerError`


Manishnemade12:
> @Manishnemade12 I wasnt referring to opening an issue for this, no conflict here. Was just pointing out that the test did not fail because of my changes.

yes bro i got it

yug54475:
@PragalvaXFREZ could you please review it? Would be really appreciated.

yug54475:
@PragalvaXFREZ all the tests now pass. Could you please review it?
Im open to any kind of feedback and improvements.

server/handlers/workspace_handlers.go
```diff
@@ -134,8 +137,16 @@ func (h *Handler) DeleteWorkspaceHandler(w http.ResponseWriter, r *http.Request,
 	workspaceID := mux.Vars(r)["id"]
 	resp, err := provider.DeleteWorkspace(r, workspaceID)
 	if err != nil {
-		h.log.Error(ErrGetResult(err))
-		http.Error(w, ErrGetResult(err).Error(), http.StatusNotFound)
+
+		if errors.Is(err, gorm.ErrRecordNotFound) {
```
PragalvaXFREZ: remote-provider's 404s would miss this branch and become 500s 

https://github.com/meshery/meshery/blob/c8f195824bcf277d593b62b6a94b9dc2caeb76c5/server/models/remote_provider.go#L5701

yug54475: @PragalvaXFREZ @lekaf974 Thanks for catching this. I see that `ErrFetch` embeds the status code as a string (`"Status Code: " + fmt.Sprint(statusCode)`) but there is no structured way to extract it, so the branching misses the remote provider failures.
>
> Two possible approaches I can see:
> 1. Use `strings.Contains(err.Error(), "Status Code: 404")` alongside `errors.Is(err, gorm.ErrRecordNotFound)` to catch both, but it highly depends on the error message format not changing so this may break silently.
> 2. Revert the branching entirely and return a flat 500 for all provider failures, leaving proper 404 detection as a follow-up that would require typed errors from the provider layer.
>
> Which approach would you prefer?

lekaf974: There is a deeper issue local and default provider should return a meshkit error and in handlers we should only deal with meshkit error code

server/handlers/workspace_handlers.go
```diff
@@ -109,16 +111,17 @@ func (h *Handler) SaveWorkspaceHandler(w http.ResponseWriter, req *http.Request,
 
 	bf, err := provider.SaveWorkspace(req, &workspace, "", false)
 	if err != nil {
-		h.log.Error(ErrGetResult(err))
-		http.Error(w, ErrGetResult(err).Error(), http.StatusNotFound)
+		saveErr := ErrFailToSave(err, "workspace")
```
lekaf974: this error should be returned at provider level

lekaf974 (changes requested):
Please move gorm error handling at provider level and use meshkit error code in handler

yug54475:
@lekaf974 Updated, moved gorm error handling to `GetWorkspace` in `workspace_persister.go` which now returns `ErrResultNotFound` for not-found cases. Handler now only checks `errors.GetCode(err) == models.ErrResultNotFoundCode` for 404 detection and passes errors through without wrapping. Removed all ErrFailToSave/ErrFailToDelete from the handler layer. Open to any further improvements.

server/handlers/workspace_handlers.go
```diff
@@ -191,9 +203,7 @@ func (h *Handler) UpdateWorkspaceHandler(w http.ResponseWriter, req *http.Reques
 	w.WriteHeader(http.StatusOK)
 	_, err = w.Write(respJSON)
 	if err != nil {
-		h.log.Error(ErrGetResult(err))
-		http.Error(w, "Failed to write response", http.StatusInternalServerError)
-		return
+		h.log.Error(err)
```
lekaf974, line 206: Why remove this if there is an error we should inform the client ?

lekaf974 (changes requested):
this is only cover local provider but what if the error is from remote provider ?

server/handlers/workspace_handlers.go
yug54475: The response is already committed once `WriteHeader` and `Write` are called, so calling `http.Error` here has no effect (status code and headers are already sent). In case of a write failure, we can only log the error. Hence, I removed the `http.Error` call. Is there any other way to handle this?

yug54475:
@lekaf974  Both local and remote providers now return `ErrResultNotFound` for `404` cases. Handler uses `errors.GetCode(err) == models.ErrResultNotFoundCode` to detect not-found consistently across both providers. Please review and LMK if something else needs to be done.

leecalcote (commented):
Good improvements — fixing the header ordering (`Content-Type` before `WriteHeader`) is important since Go ignores headers set after `WriteHeader`. The 404 vs 500 differentiation using `ErrResultNotFoundCode` in delete/update is a nice touch over the blanket `StatusNotFound` that was there before. The remote provider changes to surface `StatusNotFound` properly complete the fix across the stack. Address the reviewer feedback about remote provider error handling and this should be ready.

yug54475:
@leecalcote Thank you for the review! @lekaf974 Lee has reviewed the remote provider changes and confirmed they complete the fix. Could you re-review and approve when you get a chance?

server/handlers/workspace_handlers.go
```diff
@@ -109,16 +111,16 @@ func (h *Handler) SaveWorkspaceHandler(w http.ResponseWriter, req *http.Request,
 
 	bf, err := provider.SaveWorkspace(req, &workspace, "", false)
 	if err != nil {
-		h.log.Error(ErrGetResult(err))
-		http.Error(w, ErrGetResult(err).Error(), http.StatusNotFound)
+		h.log.Error(err)
```
lekaf974, line 114: here based on changes it can be a StatusNotFound as well right ?

yug54475: @lekaf974 This handler is purely a create operation , it reads the body, unmarshals it, and calls SaveWorkspace. There's no resource lookup involved at any step, so there's nothing to "not find". 500 should cover all the failure cases here.

lekaf974: I was under impression it was done on Update my bad all good here

lekaf974 (approved):
LGTM

yug54475:
@lekaf974  Thanks a lot for the review!! Happy to contribute.

alexquincy:
> Please move gorm error handling at provider level and use meshkit error code in handler

💯

yug54475:
@leecalcote @PragalvaXFREZ all changes have been approved and addressed. Please give one last review whenever you get the chance to. Open to any suggestions/improvements

PR #18291, opened 2026-03-28
[Server] Reject invalid performance test endpoint URLs

### LINKED ISSUE #18290: [Server] SMP performance test validator accepts invalid endpoint URLs
### Current Behavior
`SMPPerformanceTestConfigValidator` uses `url.Parse` to validate performance test endpoints, but `url.Parse` accepts many non-URL strings as relative paths without returning an error.

As a result, malformed endpoint values can pass validation and only fail later during execution.

### Expected Behavior
Performance test validation should reject endpoint values that are not valid absolute HTTP or HTTPS URLs.

Invalid endpoint configuration should fail fast during validation instead of surfacing later as runtime request errors.

### Screenshots/Logs
Relevant validator logic:

```go
// server/models/validators.go
for _, URL := range testClient.EndpointUrls {
	if _, err := url.Parse(URL); err != nil {
		return ErrValidURL
	}
}
```

`url.Parse` is too permissive for endpoint validation because strings like `not-a-url` parse successfully as relative references.

The validator already treats these values as concrete test endpoints:

```go
// server/models/validators.go
if len(testClient.EndpointUrls) < 1 {
	return ErrTestEndpoint
}
```

### Environment

---

Manishnemade12:
**Notes for Reviewers**

- This PR fixes the SMP performance test validator so invalid endpoint strings like `not-a-url` are rejected during validation instead of slipping through to runtime.
- Scope is intentionally limited to `server/models/validators.go` and `server/models/validators_test.go`.
- This PR fixes #18290

**What changed**

- Replaced permissive endpoint parsing with stricter URL validation in `SMPPerformanceTestConfigValidator`.
- Added regression coverage for:
  - rejecting relative / malformed endpoint strings
  - accepting valid absolute endpoints

**Validation**

- Ran:
  `go test ./server/models -run '^TestSMPPerformanceTestConfigValidator' -count=1`

### DIFF
--- a/docs/layouts/shortcodes/test-report.md
+++ b/docs/layouts/shortcodes/test-report.md
@@ -1,6 +1,6 @@
 ### END-TO-END TESTS
 
-- Testing started at: April 9th 2026, 6:38:41 am
+- Testing started at: April 9th 2026, 3:21:23 pm
 
 **📦 Test Result Summary**
 
--- a/docs/meetings/2026/week-of-04-09-2026/index.md
+++ b/docs/meetings/2026/week-of-04-09-2026/index.md
@@ -1,0 +1,1 @@
+Mark your attendance by committing your markdown files with names.
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -33,7 +33,7 @@
         "@rjsf/utils": "^6.4.1",
         "@rjsf/validator-ajv8": "^6.4.1",
         "@sistent/mui-datatables": "^6.1.1",
-        "@sistent/sistent": "^0.18.3",
+        "@sistent/sistent": "^0.18.4",
         "@uiw/codemirror-theme-material": "^4.25.8",
         "@uiw/react-codemirror": "^4.25.8",
         "@uiw/react-md-editor": "^4.0.11",
@@ -3606,9 +3606,9 @@
       }
     },
     "node_modules/@sistent/sistent": {
-      "version": "0.18.3",
-      "resolved": "https://registry.npmjs.org/@sistent/sistent/-/sistent-0.18.3.tgz",
-      "integrity": "sha512-5PedJOxixwSgQantGvwGR1gCWttqZN7FO9MCes+Bk+Rhlq6n6MB8AoSP/hDqoBlEpRVudmi9s3eE5BI5Mu6h3Q==",
+      "version": "0.18.4",
+      "resolved": "https://registry.npmjs.org/@sistent/sistent/-/sistent-0.18.4.tgz",
+      "integrity": "sha512-svyHZYyiIzd/WUOI4q95PmLYMgM9ax4ghqHVtQ10H5hBT3fgeDXRNIaeYVt+WKTtkDOD4CnxVq2coxPfG7N0eg==",
       "dependencies": {
         "@emotion/react": "^11.14.0",
         "@emotion/styled": "^11.14.1",
--- a/ui/package.json
+++ b/ui/package.json
@@ -54,7 +54,7 @@
     "@rjsf/utils": "^6.4.1",
     "@rjsf/validator-ajv8": "^6.4.1",
     "@sistent/mui-datatables": "^6.1.1",
-    "@sistent/sistent": "^0.18.3",
+    "@sistent/sistent": "^0.18.4",
     "@uiw/codemirror-theme-material": "^4.25.8",
     "@uiw/react-codemirror": "^4.25.8",
     "@uiw/react-md-editor": "^4.0.11",
--- a/ui/utils/utils.tsx
+++ b/ui/utils/utils.tsx
@@ -435,7 +435,7 @@ export const getComponentFromDesign = (design, componentId) => {
  */
 export const getDesignVersion = (design) => {
   if (design?.visibility === 'published') {
-    return design.catalog_data.published_version;
+    return design.catalog_data?.published_version;
   } else {
     try {
       const parsedYaml = yaml.load(design.patternFile);


Manishnemade12:
@lekaf974 can you please review this pr.

Manishnemade12:
> The switch from `url.Parse` (which accepts almost anything) to `url.ParseRequestURI` + explicit scheme/host checks is exactly right. Tests cover both the reject and accept paths well. Consider converting to table-driven tests in the future for easier expansion, but the coverage is solid as-is.

Thanks for the detailed review and kind words! Glad the switch to `url.ParseRequestURI` with explicit scheme/host checks landed well.

Fully agree on the table-driven tests suggestion ,  it's a cleaner pattern and makes future expansion much easier. I'll go ahead and refactor the tests to that structure.

Thanks for feedback

server/models/validators_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"testing"
+
+	SMP "github.com/layer5io/service-mesh-performance/spec"
+)
+
+func TestSMPPerformanceTestConfigValidatorRejectsRelativeEndpoints(t *testing.T) {
+	perfTest := &SMP.PerformanceTestConfig{
+		Name:     "invalid-endpoint-test",
```
lekaf974:
```suggestion
		Name:     "given invalid endpoint when SMPPerformanceTestConfigValidator then throw ErrValidURL",
```

server/models/validators_test.go
```diff
@@ -0,0 +1,49 @@
...
+			{
+				LoadGenerator: FortioLG.Name(),
+				Protocol:      SMP.PerformanceTestConfig_Client_PROTOCOL_HTTP,
+				EndpointUrls:  []string{"not-a-url"},
+			},
+		},
+	}
+
+	err := SMPPerformanceTestConfigValidator(perfTest)
+	if err != ErrValidURL {
```
lekaf974: use assert instead

server/models/validators_test.go
```diff
@@ -0,0 +1,49 @@
...
+		t.Fatalf("expected ErrValidURL, got %v", err)
+	}
+}
+
+func TestSMPPerformanceTestConfigValidatorAcceptsAbsoluteEndpoints(t *testing.T) {
+	perfTest := &SMP.PerformanceTestConfig{
+		Name:     "valid-endpoint-test",
```
lekaf974:
```suggestion
		Name:     "given valid endpoint when SMPPerformanceTestConfigValidator then assert err nil",
```

server/models/validators_test.go
```diff
@@ -0,0 +1,49 @@
...
+			},
+			{
+				LoadGenerator: Wrk2LG.Name(),
+				Protocol:      SMP.PerformanceTestConfig_Client_PROTOCOL_TCP,
+				EndpointUrls:  []string{"tcp://127.0.0.1:8080"},
+			},
+		},
+	}
+
+	if err := SMPPerformanceTestConfigValidator(perfTest); err != nil {
```
lekaf974: use assert

Manishnemade12:
@lekaf974 done i addressed all comments and 
Refactored to a table-driven test for maintainability and easier extensibility.
Now covers absolute, relative, and empty endpoint cases for robust validation.

you can take review once

server/models/validators_test.go
Manishnemade12: done

server/models/validators_test.go
Manishnemade12: done, i used assert

server/models/validators_test.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"testing"
+
+	SMP "github.com/layer5io/service-mesh-performance/spec"
+)
+
+func TestSMPPerformanceTestConfigValidatorRejectsRelativeEndpoints(t *testing.T) {
+	perfTest := &SMP.PerformanceTestConfig{
+		Name:     "given invalid endpoint when SMPPerformanceTestConfigValidator then throw ErrValidURL",		Duration: "30s",
+		Clients: []*SMP.PerformanceTestConfig_Client{
```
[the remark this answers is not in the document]
lekaf974: please address this comment

server/models/validators_test.go
```diff
@@ -0,0 +1,47 @@
...
+			{
+				LoadGenerator: FortioLG.Name(),
+				Protocol:      SMP.PerformanceTestConfig_Client_PROTOCOL_HTTP,
+				EndpointUrls:  []string{"not-a-url", "/health", "health"},
+			},
+		},
+	}
+
+	       err := SMPPerformanceTestConfigValidator(perfTest)
+	       if err != ErrValidURL {
```
lekaf974: use assert please

Manishnemade12:
@lekaf974 I merged the two endpoint validator tests into a single table-driven test and removed the old functions.

server/models/validators_test.go
Manishnemade12: done , i did that

server/models/validators_test.go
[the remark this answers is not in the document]
Manishnemade12: The file was not gofmt-formatted, so I ran gofmt to ensure it meets Meshery’s code style standards. The file is now properly formatted.

server/models/validators_test.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func TestSMPPerformanceTestConfigValidator(t *testing.T) {
+	cases := []struct {
+		name      string
+		config    *SMP.PerformanceTestConfig
+		wantError error
+	}{
+		{
+			name: "rejects invalid/relative endpoints",
```
lekaf974: please keep naming given, when, then standard

This is helpful when it fails to quickly know what the test expected as input, what we are testing and the expected output without the need to read all inputs and outputs

Manishnemade12:
@lekaf974 i changed test case names in TestSMPPerformanceTestConfigValidator to follow the "given, when, then" naming convention.

server/models/validators_test.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+	}
+
+	       for _, tc := range cases {
+		       t.Run(tc.name, func(t *testing.T) {
+			       err := SMPPerformanceTestConfigValidator(tc.config)
+			       if tc.wantError != nil {
+				       if !errors.Is(err, tc.wantError) {
+					       t.Errorf("expected error %v, got %v", tc.wantError, err)
+				       }
```
lekaf974: we have a AsserErrorEqual method in utls package, please reuse it

Manishnemade12: I first used AssertMeshkitErrorsEqual, but that imported utils, which creates an import cycle because utils already imports models.
To avoid this, I switched to local errors.Is(...) in the test.
This keeps the test simple and removes the dependency cycle.

Manishnemade12:
@PragalvaXFREZ This PR has been approved by @lekaf974 , with small improvement -> reuse of the AssertErrorEqual method.
so I first used [AssertMeshkitErrorsEqual](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), but that imported [utils](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), which creates an import cycle because [utils](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) already imports [models](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).  as result Golang Unit and Integration Tests started failing.
To avoid this, I switched to local [errors.Is(...)](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in the test.
This keeps the test simple and removes the dependency cycle.

 Could you please review it at your convenience and proceed with merging if everything looks good

server/models/validators_test.go
```diff
@@ -50,18 +50,17 @@ func TestSMPPerformanceTestConfigValidator(t *testing.T) {
...
-					       t.Errorf("expected no error, got %v", err)
-				       }
-			       }
-		       })
-	       }
+		for _, tc := range cases {
+			t.Run(tc.name, func(t *testing.T) {
+				err := SMPPerformanceTestConfigValidator(tc.config)
+				if tc.wantError != nil {
+					utils.AssertMeshkitErrorsEqual(t, err, tc.wantError)
```
lekaf974: we must not add dependency from cli to server

PS: we have hard time to remove this dependency from cli to server already

lekaf974: the best will be to move this test assertion in meshkit to reuse it in both components

Manishnemade12: Thanks @lekaf974 for the feedback. I have addressed both concerns:
1. Removed the CLI utils dependency , now using only standard library errors.Is() for error comparison
2. Agree on moving AssertMeshkitErrorsEqual to meshkit for proper reuse across all components - that would be a great improvement for future maintainability.

Manishnemade12:
@lekaf974 i Removed the CLI utils dependency , now using only standard library errors.Is() for error comparison .
 and also Agree on moving AssertMeshkitErrorsEqual to meshkit for proper reuse across all components - that would be a great improvement for future maintainability.
 
 so can you please review it now .

mesheryctl/pkg/utils/testing.go
```diff
@@ -346,7 +346,7 @@ func formatToTabs(data string) string {
 	return s
 }
 
-// AssertMeshkitErrorsEqual compares  relevant fields of two meshkit errors
+// AssertMeshkitErrorsEqual compares relevant fields of two meshkit errors
 func AssertMeshkitErrorsEqual(t *testing.T, got, expected error) {
```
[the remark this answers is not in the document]
Manishnemade12: @copilot apply changes based on [this feedback](https://github.com/meshery/meshery/pull/18291#discussion_r3054826698)

Issue #17903, opened 2026-03-11
bug: Missing return statements and nil pointer dereference in ProcessConnectionRegistration causes server panic


luckyxhq:
## Current Behavior

In `ProcessConnectionRegistration` (located in `server/handlers/connections_handlers.go`), there are two critical bugs caused by missing `return` statements:

**Bug 1 - Missing `return` after `InitializeMachineWithContext` error (lines 58-64):**

```go
inst, err := helpers.InitializeMachineWithContext(...)
if err != nil {
    event := eventBuilder.WithSeverity(events.Error)...Build()
    _ = provider.PersistEvent(*event, nil)
    go h.config.EventBroadcaster.Publish(userUUID, event)
    // No return here - execution falls through to inst.SendEvent
}

event, err := inst.SendEvent(req.Context(), machines.EventType(connectionRegisterPayload.Status), connectionRegisterPayload)
```

When `InitializeMachineWithContext` fails, `inst` is `nil`. Without a `return` statement, execution continues to `inst.SendEvent(...)`, causing a **nil pointer dereference** that **panics the server**.

**Bug 2 - Missing `return` after `http.Error` in the `SendEvent` error block (lines 67-72):**

```go
event, err := inst.SendEvent(...)
if err != nil {
    h.log.Error(err)
    http.Error(w, err.Error(), http.StatusInternalServerError)
    // No return here - execution continues after writing error response
    _ = provider.PersistEvent(*event, nil)
    go h.config.EventBroadcaster.Publish(userUUID, event)
}
```

After sending an HTTP error response, execution continues to dereference `*event` (which may be `nil` if `inst.SendEvent` returned both an error `and` a nil event pointer), resulting in another potential nil pointer dereference panic.

## Expected Behavior

- When `InitializeMachineWithContext` returns an error, the handler should **immediately return** after publishing the error event, preventing any subsequent access to the `nil` `inst` pointer.
- - When `inst.SendEvent` returns an error, the handler should **immediately return** after writing the HTTP error response, preventing further execution on potentially nil values.
## Steps to Reproduce

1. Start the Meshery server.
2. 2. Send a `POST /api/integrations/connections/register` request with a connection payload whose `status` is **not** `init`.
3. 3. Trigger a condition where `helpers.InitializeMachineWithContext` returns an error (e.g., invalid connection ID or uninitialized state machine context).
4. 4. Observe the server **panic** due to a nil pointer dereference on `inst`.
## Root Cause

[...]

Issue #17832, opened 2026-03-08
[Bug] Incorrect HTTP response handling in `environments_handlers.go` — wrong status codes, `WriteHeader` after body write, and misused error wrappers


luckyxhq:
## Description

The file `server/handlers/environments_handlers.go` contains multiple HTTP response handling bugs that cause incorrect status codes to be returned to clients and use semantically wrong error wrappers. These issues affect 4 out of 7 handler functions in the file.

### Bug 1: `SaveEnvironment` — `WriteHeader` called after body write (Line 111-113)

In the `SaveEnvironment` handler, `fmt.Fprint(w, string(resp))` writes the response body **before** `w.WriteHeader(http.StatusCreated)` is called. In Go's `net/http`, the first call to `Write()` implicitly sends HTTP status `200 OK`. The subsequent `WriteHeader(201)` call is ignored and produces a `http: superfluous response.WriteHeader call` warning in server logs.

**Result**: Clients always receive `200 OK` instead of the intended `201 Created` for successful environment creation.

```go
// Line 111-113 (current - buggy)
w.Header().Set("Content-Type", "application/json")
fmt.Fprint(w, string(resp))         // <-- implicitly sends 200 OK
w.WriteHeader(http.StatusCreated)    // <-- IGNORED, too late!
```

### Bug 2: Wrong HTTP status codes for error responses (Lines 103-104, 126-127, 199-200)

Multiple handlers use `http.StatusNotFound` (404) for errors that are **not** "resource not found" errors:

| Handler | Line | Error Condition | Current Status | Correct Status |
|---|---|---|---|---|
| `SaveEnvironment` | 104 | Failed to save environment | 404 Not Found | 500 Internal Server Error |
| `DeleteEnvironmentHandler` | 127 | Failed to delete environment | 404 Not Found | 500 Internal Server Error |
| `AddConnectionToEnvironmentHandler` | 200 | Failed to add connection mapping | 404 Not Found | 500 Internal Server Error |
| `RemoveConnectionFromEnvironmentHandler` | 221 | Failed to remove connection mapping | 404 Not Found | 500 Internal Server Error |

### Bug 3: Semantically wrong error wrapper — `ErrGetResult` used for non-GET operations

The handlers use `ErrGetResult(err)` (which wraps errors with "get" semantics) for operations that are saving, deleting, or modifying environments. This produces misleading error messages and logs.

| Handler | Operation | Error Wrapper Used | Should Be |
|---|---|---|---|
[...]

Issue #17823, opened 2026-03-07
[Bug] Missing `return` after `http.Error()` in `FetchSmiResultsHandler` and `FetchSingleSmiResultHandler` causes continued execution after error


luckyxhq:
## Description

In `server/handlers/fetch_results_handler.go`, the functions `FetchSmiResultsHandler` and `FetchSingleSmiResultHandler` contain multiple instances where `http.Error()` is called to send an error response to the client, but the function does **not** `return` afterward. This causes execution to continue past the error handling block, potentially leading to:

- **Double HTTP response writes** (writing response body after an error status has already been sent)
- **Nil or invalid data being written** to the response (e.g., writing `bdr` which could be `nil` after a failed provider call)
- **Superfluous `http.WriteHeader` call warnings** in server logs

## Current Behavior

When an error occurs (e.g., `req.ParseForm()` fails or `provider.FetchSmiResults()` returns an error), the handler sends an HTTP error response but **continues executing** — attempting to write `bdr` (which may be nil) to the response writer. This results in corrupted responses and `http: superfluous response.WriteHeader call` warnings.

## Expected Behavior

After sending an error response via `http.Error()`, the handler should immediately `return` to prevent further writes to the response.

## Screenshots / Logs

N/A — This is a code-level bug confirmed through static analysis.

## Affected Code

**File:** `server/handlers/fetch_results_handler.go`

### Location 1: `FetchSmiResultsHandler` — Line 184 (ParseForm error)

```go
func (h *Handler) FetchSmiResultsHandler(w http.ResponseWriter, req *http.Request, _ *models.Preference, _ *models.User, p models.Provider) {
	w.Header().Set("content-type", "application/json")
	err := req.ParseForm()
	if err != nil {
		h.log.Error(ErrParseForm(err))
		http.Error(w, ErrParseForm(err).Error(), http.StatusForbidden)
		// BUG: missing return here — execution continues
	}
	// ...
```

### Location 2: `FetchSmiResultsHandler` — Line 191 (FetchSmiResults error)

```go
	bdr, err := p.FetchSmiResults(req, q.Get("page"), q.Get("pagesize"), q.Get("search"), q.Get("order"))
	if err != nil {
		h.log.Error(ErrFetchSMIResults(err))
		http.Error(w, ErrFetchSMIResults(err).Error(), http.StatusInternalServerError)
		// BUG: missing return here — nil bdr gets written below
	}
	_, _ = w.Write(bdr) // writes nil data after error
```

[...]

PR #18299, opened 2026-03-28
[Server] Fix missing return and double response writes in HTTP handlers

officialasishkumar:
**Notes for Reviewers**

- This PR fixes three instances of incorrect HTTP response handling in server handlers that are not covered by existing issues (#17823, #17832, #17903):

### Bug 1: `contexts_handler.go` — Missing `return` after `http.Error()` in `GetAllContexts`

When `json.Unmarshal` fails (line 48), the handler sends an error response via `http.Error()` but does **not** `return`. Execution falls through to `json.NewEncoder(w).Encode(mesheryK8sContextPage)` which writes a second response body to the already-errored `ResponseWriter`, producing corrupted output and `http: superfluous response.WriteHeader call` warnings.

**Fix:** Added `return` after `http.Error()`.

### Bug 2: `design_engine_handler.go` — Double response writes in `PatternFileHandler`

Two error blocks (read body failure and unmarshal failure) call `http.Error()` with status 500, then immediately call `rw.WriteHeader(http.StatusBadRequest)` + `fmt.Fprintf()`. Since `http.Error()` already writes the status and body, the subsequent `WriteHeader` is ignored (superfluous) and `Fprintf` appends garbage to the response.

**Fix:** Replaced the double-write with a single `http.Error()` using the correct `http.StatusBadRequest` status code.

### Bug 3: `load_test_handler.go` — Same double response write in `LoadTestUsingSMPHandler`

Identical pattern to Bug 2 above.

**Fix:** Same — single `http.Error()` with `http.StatusBadRequest`.


matrixkavi:
@officialasishkumar - Good catch. Could you please present the issue and the PR in upcoming Meshery Development Meeting on Wednesday? Please find the community calendar and subscribe to it [here](https://layer5.io/community/calendar). You may also join our slack [here](https://mesheryio.slack.com/)

Issue #18210, opened 2026-03-25
[UI]  Animation of Meshery logo not working


saurabhraghuvanshii:
### Current Behavior
We previously had an animation of the Meshery logo on the initial page load, which is currently not working.

<img width="1920" height="755" alt="Image" src="https://github.com/user-attachments/assets/cc4257db-a164-4055-b85a-802f5b54b561" />

### Desired Behavior
Identify the issue and fix it.

### Environment

- Browser: Chrome Safari Firefox
- Host OS: Mac Linux Windows
- Meshery Server Version: stable-v
- Platform: Docker Kubernetes

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)

- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

suzy-g38:
@saurabhraghuvanshii I would like to work on this issue

saurabhraghuvanshii:
Thanks @suzy-g38 !!

suzy-g38:
@saurabhraghuvanshii I think i have found the issue, but can you tell me how the animation used to be when it was working fine? like is there any way to check that?

saurabhraghuvanshii:
@suzy-g38 I don't have any recording of animation. Raise Pr with working state then maintainer will review.

suzy-g38:
> [@suzy-g38](https://github.com/suzy-g38) I don't have any recording of animation. Raise Pr with working state then maintainer will review.

okay

PR #18300, opened 2026-03-28
fix: meshery logo animation

suzy-g38:
**Notes for Reviewers**

- This PR fixes #18210

### Changes Made

The logo animation where broken during the loading time, so I have added some animation style to fix it. 

Here is the loom video for reference: https://www.loom.com/share/79076bfe98fd43e385c0102a52a5ea7a


ui/components/LoadingComponents/Animations/AnimatedMesheryCSS.tsx
```diff
@@ -1,10 +1,49 @@
...
+    animation: meshery-fade 2s ease-in-out infinite;
+    animation-delay: 0.9s;
+  }
+
+  @keyframes meshery-fade {
+    0%   { opacity: 1;   }
+    50%  { opacity: 0.2; }
+    100% { opacity: 1;   }
+  }
+`;
```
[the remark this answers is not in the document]
suzy-g38: done

Sbragul26:
@suzy-g38 ,

I noticed a new CSS animation is being added here. There’s already an animation in **AnimatedMesheryCSS.tsx** that includes the SVG and animation logic.
Maybe we can reuse that instead, to keep things consistent and avoid duplication.

suzy-g38:
> @suzy-g38 ,
> 
> I noticed a new CSS animation is being added here. There’s already an animation in **AnimatedMesheryCSS.tsx** that includes the SVG and animation logic. Maybe we can reuse that instead, to keep things consistent and avoid duplication.

@Sbragul26, yes the new animation I added in that same very file, but the previous animation was not working in the first place, that's why I added this, but would love more feedback on this.

Sbragul26:
I think the previous animation might still work if we trigger it properly. Maybe we can try using that instead of adding a new one.

Let’s give it a try once , it might work without adding extra styles..

YASHMAHAKAL:
@suzy-g38 if we have animation logic defined already as @Sbragul26 says, please see if you can fix it, instead of adding new lines of code : )

suzy-g38:
> Thanks for working on this, @suzy-g38. Please include a brief screen recording of the new animation.

@leecalcote, here is the loom video for refernce: https://www.loom.com/share/79076bfe98fd43e385c0102a52a5ea7a

It is included in the pr description as well

suzy-g38:
> @suzy-g38 if we have animation logic defined already as @Sbragul26 says, please see if you can fix it, instead of adding new lines of code : )

@Sbragul26 and @YASHMAHAKAL, initially I tried to find anything existing actually, but couldn't find any animation code in the AnimatedMesharyCSS file that's why I added the animation code,  and this AnimationMesheryCSS file is actually directly being used for loading, that @Sbragul26 mentioned initially.

But I will try this by today to check if we can fix this by by state trigger and anything related. 
Thanks for the guidance to both of you. I am new to this project, but hope to learn things quickly.

suzy-g38:
@Sbragul26 and @YASHMAHAKAL I tried to check if there is any existing way to solve it, but I found that AnimatedMesheryCSS.tsx is the file being used for the loading screen, but it had no animation styles, which is why the logo appeared static.

I tried another file which is AnimatedMeshery.tsx, but that file is not being used for loading. There were other couple of files but same, they are not being used.

I found the original AnimatedMeshery.css file which contains the correct animation sequence (staggered CSS transitions per triangle). However, there are some import things I found - 

AnimatedMeshery.css is already imported in _app.tsx and works correctly for the post-hydration LoadingScreen component. 

However PureHtmlLoadingScreen in _document.tsx is a separate, earlier loading screen that renders before _app.tsx has loaded at all 

so this is the heirarchy

> 1. _document.tsx is being called
> 2. PureHtmlLoadingScreen is being called
>   <PureHtmlLoadingScreen id={'PRE_REACT_LOADER'} message="" />
>   
>  3. Inside PureHtmlLoadingScreen AnimatedLogoDark is being called
>  <AnimatedLogoDark />
>   
>  4. AnimatedLogoDark is being called from AnimatedMesheryCSS.tsx
>  export const AnimatedMesheryDark = (props) => <AnimatedMeshery {...props} />;

so by the time _app.tsx loads and its CSS imports become available, the initial loading screen is already visible and already missing its animation. Any CSS imported via a regular import statement or <link> tag arrives too late.
The only correct solution is to inline the animation styles directly inside AnimatedMesheryCSS.tsx so they are embedded in the raw HTML response and available the instant the browser paints the loading screen.

yi-nuo426:
@saurabh100ni, will you help this moving along, please?

saurabhraghuvanshii:
@suzy-g38 Do you think we can remove this file?
https://github.com/meshery/meshery/blob/master/ui/pages/styles/AnimatedMeshery.css

suzy-g38:
> @suzy-g38 Do you think we can remove this file? https://github.com/meshery/meshery/blob/master/ui/pages/styles/AnimatedMeshery.css

no becuase it is being used in post hydration animation, and if you remove it, it will break the post hydration loading version

saurabhraghuvanshii:
@suzy-g38 I'm a bit unsure about these changes. Instead of hardcoding this again, maybe we can find a better solution. Could you help investigate why the previous animation stopped working after the Next.js upgrade and MUI v7?

Sbragul26:
@saurabhraghuvanshii,
I think the issue is related to when the CSS is loaded. The loader renders early (from _document.tsx), but the animation styles are loaded later (from _app.tsx).

Because of this, the required styles are not available during the initial render, so the animation doesn’t trigger properly after the Next.js and MUI v7 upgrade.

suzy-g38:
> @saurabhraghuvanshii, I think the issue is related to when the CSS is loaded. The loader renders early (from _document.tsx), but the animation styles are loaded later (from _app.tsx).
> 
> Because of this, the required styles are not available during the initial render, so the animation doesn’t trigger properly after the Next.js and MUI v7 upgrade.

exactly this is what happening @saurabhraghuvanshii 

Could anyone of you know at which PR or commit where AnimatedMesheryCSS.tsx was introduced or where PureHtmlLoadingScreen was added? That would help me pinpoint the exact change that caused the regression and whether if there is a better solution or not?

In this pr I just made it(initial loading) self sufficient so that it never needs to wait for any external html

saurabhraghuvanshii:
@suzy-g38 Please refer to the commit history:
https://github.com/meshery/meshery/commits/master/ui/pages/styles/AnimatedMeshery.css

Thanks for the explanation @Sbragul26.

However, instead of hardcoding animation timings, please investigate the root cause of why the animations stopped working after the Next.js and MUI upgrades.

We should aim for a solution that aligns with the latest changes and remains maintainable in the long term.

suzy-g38:
> @suzy-g38 Please refer to the commit history: https://github.com/meshery/meshery/commits/master/ui/pages/styles/AnimatedMeshery.css
> 
> Thanks for the explanation @Sbragul26.
> 
> However, instead of hardcoding animation timings, please investigate the root cause of why the animations stopped working after the Next.js and MUI upgrades.
> 
> We should aim for a solution that aligns with the latest changes and remains maintainable in the long term.

@saurabhraghuvanshii To clarify on the timing values, these are the original values from AnimatedMeshery.css (0.8s to 3.2s staggered delays, generated by SVG Artista in 2022). They are not arbitrary hardcoding, they are the intentional animation sequence that existed before the regression.

Now coming to the root cause, I looked into the commit history and here is what happened:

The issue is not caused by the Next.js or MUI v7 upgrade. The upgrade just made an existing architectural mismatch visible.

That commit moved all CSS files to pages/styles/ and imported them globally in _app.js, removing the local imports from each component. This worked fine at the time. The problem occurred later when PureHtmlLoadingScreen was introduced in _document.tsx using AnimatedMesheryCSS.tsx, a pre-hydration component that renders before _app.js loads. Because the CSS had been moved to _app.js, it was no longer available at the time AnimatedMesheryCSS.tsx renders.

You may think to restore the original pattern, moving the CSS back to be co-located with AnimatedMesheryCSS.tsx and importing it directly there. However it will still won't work for the pre-hydration context. Any CSS or JS import gets bundled into a separate file by Turbopack and arrives after the loading screen is already visible. Only content that is directly embedded in the HTML...like inline <style> tags or inline <script> tags, is available at the exact moment the browser paints the page. This is why inline CSS is the solution here.

The fundamental architectural issue is that AnimatedMesheryCSS.tsx was designed to be a pre-hydration component but was never given the styles it needed to work in that context.

This is not a new pattern. It is the same approach Next.js itself uses for critical CSS, and the same approach already used in _document.tsx for the scrollbar styles:

```
 <style type="text/css">
            {`
              .hide-scrollbar::-webkit-scrollbar {
                width: 0 !important;
              }
              .reduce-scrollbar-width::-webkit-scrollbar {
                width: 0.3em !important;
              }
            `}
          </style>
```

The animation fix follows the exact same pattern, just with the animation CSS.

Regarding duplication, the cleanest solution would be to move the inline styles out of AnimatedMesheryCSS.tsx and into _document.tsx directly, in the <Head> section alongside the existing scrollbar styles. That way there is one place for all pre-hydration critical styles, and AnimatedMeshery.css remains the single source of truth for post-hydration animations. I am happy to make that change if preferred.

For reference, this behavior is documented in the Next.js custom document guide: https://nextjs.org/docs/pages/building-your-application/routing/custom-document

saurabhraghuvanshii:
@suzy-g38 I found a working solution.
I’m sharing my steps—please verify them on your side before implementing, as they are working for me. Kindly double-check.

l5io:
🚀 Deployed on https://69d3f7951d7e6ea3a9c2a9fe--meshery-docs-preview.netlify.app

l5io:
🚀 Preview for commit ed6a2c0121551cd457abb6662a9d1a14c0c7c370 at: https://69d3d450adc9ee241dadd534--meshery-docs-preview.netlify.app

l5io:
🚀 Preview for commit 1539b762a798c4005253931e1392a81c27f93145 at: https://69d3d60d65e951007aca975b--meshery-docs-preview.netlify.app

docs/CHANGELOG.md
saurabhraghuvanshii, line 1: @suzy-g38  removed this Please.

ui/components/LoadingComponents/Animations/AnimatedMeshPattern.tsx
```diff
@@ -16,15 +16,19 @@ const AnimatedMeshPattern = (props) => {
...
 
   useEffect(() => {
-    setTimeout(() => {
-      setIsActive(!isActive);
+    const timeout = setTimeout(() => {
+      setIsActive((prev) => !prev);
     }, 2000);
+
+    return () => clearTimeout(timeout);
   }, [isActive]);
```
saurabhraghuvanshii, line 32: @suzy-g38  I think this will be good 
```
useEffect(() => {
  const interval = setInterval(() => {
    setIsActive((prev) => !prev);
  }, 2000);

  return () => clearInterval(interval);
}, []);
```

saurabhraghuvanshii: @suzy-g38  if this looks good then apply to all other. And please also test locally once with new changes

l5io:
🚀 Preview for commit 75c7c0071761f6a9788465228968a01460913ec3 at: https://69d3e6471d39d439d278817e--meshery-docs-preview.netlify.app

l5io:
🚀 Preview for commit 8727d612961fca7181e365a5cb52b12186c23b42 at: https://69d3efeb214254829c9a986c--meshery-docs-preview.netlify.app

l5io:
🚀 Preview for commit 2a21a5276c405ad376ed9b363fecc53bf381330d at: https://69d3f57cc3fbe3ca507fa724--meshery-docs-preview.netlify.app

l5io:
🚀 Preview for commit 34720344892ece00481b4e959ddec287b18da3ae at: https://69d3f7951d7e6ea3a9c2a9fe--meshery-docs-preview.netlify.app

Issue #18419, opened 2026-04-02
[Docs] models/single.html renders empty <p> tag when .Params.subtitle is absent


alexquincy:
## Description

`docs/layouts/models/single.html` unconditionally renders `.Params.subtitle` inside a `<p>` tag without checking whether the parameter exists. When a model page has no `subtitle` front matter key, this outputs an empty `<p></p>` element.

**File:** `docs/layouts/models/single.html:3`

```html
<p>{{ .Params.subtitle }}</p>
```

## Why This Is a Problem

1. **Empty element in DOM**: An empty `<p></p>` creates unintended vertical whitespace in the layout — typically a blank line with top/bottom margins — that visually breaks the page header area.
2. **Inconsistency**: Other optional front matter fields in Hugo templates are typically guarded with `{{ if .Params.fieldname }}...{{ end }}`.
3. **SEO/accessibility**: Empty block-level elements add unnecessary DOM nodes that screen readers may announce or crawlers may penalize.

Most model pages in `docs/content/en/concepts/logical/models/` do not define a `subtitle` field, meaning the majority of model pages are affected.

## Expected Behavior

The subtitle should only render when defined:

```html
{{ with .Params.subtitle }}
<p>{{ . }}</p>
{{ end }}
```

## Suggested Fix

Wrap the subtitle `<p>` tag in a `{{ with .Params.subtitle }}` conditional block in `docs/layouts/models/single.html`.

lucasbryant1222-png:
Thanks for reporting this

PR #18426, opened 2026-04-02
fix(docs): guard subtitle rendering to avoid empty p tag

tysoncung:
## Problem

`docs/layouts/models/single.html` unconditionally renders `.Params.subtitle` in a `<p>` tag. When subtitle is absent, this produces an empty `<p></p>` element, adding unwanted whitespace.

## Solution

Wrapped the subtitle rendering in a `{{ with .Params.subtitle }}` block so the `<p>` tag is only emitted when a subtitle exists.

## Files Changed

- `docs/layouts/models/single.html` — line 3

Closes #18419

### DIFF
--- a/docs/layouts/models/single.html
+++ b/docs/layouts/models/single.html
@@ -1,6 +1,6 @@
 {{ define "main" }}
 <article>
-    <p>{{ .Params.subtitle }}</p>
+    {{ with .Params.subtitle }}<p>{{ . }}</p>{{ end }}
     {{ .Content }}
     <!-- components -->
     <div>


l5io:
🚀 Deployed on https://69ce838e9b4a751a93435590--meshery-docs-preview.netlify.app

l5io:
🚀 Preview for commit aac658b9ea6edc9124d19b88b96a00b8ebbfe62d at: https://69ce838e9b4a751a93435590--meshery-docs-preview.netlify.app

PR #18429, opened 2026-04-02
[Server] Fix unauthenticated file read endpoints

### LINKED ISSUE #18375: [Security Bug] /api/system/fileView and /api/system/fileDownload allow unauthenticated arbitrary file read
## Bug Summary

Two server endpoints allow reading any file on the host without validating the path:
- `/api/system/fileView`
- `/api/system/fileDownload`

Both handlers read the `file` query parameter and pass it directly to `os.Open(...)`. In the router, both endpoints are registered with `models.NoAuth`, so when provider auth is not enforced, they can be accessed without authentication.

## Why this is a problem

This can expose any file that the Meshery server process is able to read, for example:
- `/etc/hosts`
- kubeconfig files
- logs
- mounted credentials or local secrets

That makes this an arbitrary file read issue.

## Affected Code

- `server/router/server.go`
- `server/handlers/common_handlers.go`
- `server/handlers/middlewares.go`

Relevant behavior:
- router registers both endpoints with `models.NoAuth`
- `AuthMiddleware` does not block `NoAuth` routes unless a provider is explicitly enforced
- handlers call `os.Open(filePath)` on user-controlled input without validation

## Steps to Reproduce

Run Meshery in a setup where provider auth is not enforced, then call:

```bash
curl "http://localhost:9081/api/system/fileView?file=/etc/hosts"
curl -OJ "http://localhost:9081/api/system/fileDownload?file=/etc/hosts"
```

## Expected Behavior

The server should not allow arbitrary file reads from a user-provided path.

## Actual Behavior

The server opens the requested path directly and returns the file content.

## Proposed Fix

A fix would be:
- require authentication for both endpoints
- restrict file access to a small allowlisted directory, if these endpoints are still needed
- reject absolute paths and path traversal attempts
- return `403` or `400` for invalid paths

If these endpoints are not required anymore, removing them would be the safest option.

## Severity

High, because it can expose host files and secrets depending on how Meshery is deployed.

hssuryawanshib23-afk:
**Notes for Reviewers**

- This PR fixes #18375

## Summary
- require provider auth for `/api/system/fileView` and `/api/system/fileDownload`
- limit file serving to Meshery log roots and regular files only
- block outside-path access, missing `file`, and symlink escapes
- add focused handler and router regression tests

## Testing
- `go test ./server/handlers -run 'Test(ViewHandlerServesAllowedMesheryLogFile|DownloadHandlerRejectsFileOutsideAllowedRoots|ViewHandlerRejectsSymlinkEscapingAllowedRoots|DownloadHandlerRejectsMissingFileQuery)$' -count=1`
- `go test ./server/router -run 'Test(FileEndpointsRequireProviderAuth)$' -count=1`

### DIFF
--- a/server/handlers/common_handlers.go
+++ b/server/handlers/common_handlers.go
@@ -168,7 +168,7 @@ func (h *Handler) serveAllowedFile(responseWriter http.ResponseWriter, request *
 
 	if _, err = io.Copy(responseWriter, file); err != nil {
 		h.log.Error(err)
-		http.Error(responseWriter, "failed to stream requested file", http.StatusInternalServerError)
+		return
 	}
 }
 
@@ -276,7 +276,7 @@ func pathWithinAllowedRoots(path string, allowedRoots []string) (bool, error) {
 		if err != nil {
 			return false, fmt.Errorf("resolve relative path from %q to %q: %w", allowedRoot, path, err)
 		}
-		if relPath == "." || (relPath != ".." && !strings.HasPrefix(relPath, ".."+string(os.PathSeparator))) {
+		if !strings.HasPrefix(relPath, "..") {
 			return true, nil
 		}
 	}
--- a/server/handlers/common_handlers.go
+++ b/server/handlers/common_handlers.go
@@ -3,15 +3,23 @@ package handlers
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


hortison:
Yeah, we probably do need these endpoints auth’ed.

Issue #18407, opened 2026-04-02
[UI] parseInt() called without radix in Environments component causing potential NaN


simihablo:
## Description

`ui/components/Lifecycle/Environments/index.tsx` uses nested `parseInt()` calls without a radix argument, which can produce unexpected results. Additionally, wrapping `Math.ceil()` output (which already returns a number) in `parseInt()` is semantically wrong.

**File:** `ui/components/Lifecycle/Environments/index.tsx:429, 436-437`

```javascript
const pagesCount = parseInt(Math.ceil(parseInt(connections?.total_count) / connectionPageSize));

const pagesCount = parseInt(
  Math.ceil(parseInt(environmentConnections?.total_count) / connectionPageSize),
```

## Why This Is a Problem

1. **Missing radix**: `parseInt(string)` without a radix defaults to base 10 in modern JS, but ECMAScript still recommends always specifying a radix (ESLint `radix` rule). Strings starting with `"0"` were historically parsed as octal.

2. **NaN propagation**: If `connections?.total_count` is `undefined`, `null`, or a non-numeric string, `parseInt()` returns `NaN`. Dividing `NaN` by `connectionPageSize` gives `NaN`, `Math.ceil(NaN)` gives `NaN`, and the outer `parseInt(NaN)` gives `NaN` — which silently corrupts pagination.

3. **Unnecessary parseInt on number**: `Math.ceil()` always returns a `Number`. Wrapping it in `parseInt()` is redundant and confusing; `Math.ceil()` already returns an integer.

## Expected Behavior

```javascript
const totalCount = Number(connections?.total_count) || 0;
const pagesCount = Math.ceil(totalCount / connectionPageSize);
```

## Suggested Fix

Replace nested `parseInt(Math.ceil(parseInt(...)))` with `Math.ceil(Number(...) || 0)` and add a guard for zero/falsy `total_count`.

PR #18437, opened 2026-04-02
[UI] Remove redundant parseInt wrapping Math.ceil in Environments

boergeson:
**Notes for Reviewers**

- This PR fixes #18407

`parseInt(Math.ceil(parseInt(total_count)))` does two unnecessary things: the outer `parseInt` wraps `Math.ceil` which already returns an integer, and the inner `parseInt` is called without a radix argument.

Replaced with `Math.ceil(Number(total_count ?? 0))` which handles nullish values cleanly and avoids the radix ambiguity.


hortison:
Thanks!

Issue #17709, opened 2026-02-28
[mesheryctl] Remove Logging of errors and return proper error


DharunMR:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

### Current Behavior

In multiple places across the CLI commands the errors are being logged using:
```bash
utils.Log.Error(err)
return nil
```
without returning the error to the caller. This results in Errors being silently swallowed, execution continuing with invalid or nil values, unexpected runtime panics
CLI reporting success despite underlying failures

### Desired Behavior

- Replace utils.Log.Error(err); return nil with return err
- Convert Run functions to RunE where necessary
- Ensure errors are logged only at command execution boundary

#### UX enhancement
Moving forward to enhance UX consistency apply the following changes
- Update long description like following
  https://github.com/meshery/meshery/blob/7dd6e9140fed6c732a3a08abd893b07d4e3e1b84/mesheryctl/internal/cli/root/system/check.go#L128
- Add flag validator libs when required as following example
  https://github.com/meshery/meshery/blob/7dd6e9140fed6c732a3a08abd893b07d4e3e1b84/mesheryctl/internal/cli/root/system/check.go#L152

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 📚 [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- 📺 [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- ⌨️ [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- ⌨️ [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- ⌨️ [Meshkit errors Guide](https://docs.meshery.io/project/contributing/contributing-error)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

lekaf974:
[...]

PR #18471, opened 2026-04-03
fix(mesheryctl): update e2e test assertions for system context command

Omkar-Ugal:
## What does this PR do?

After recent changes to `context.go` as part of issue #17709 
(meshkit error handling + logrus removal), the error messages 
were updated but the corresponding e2e test assertions in 
`01-system-context.bats` were not updated, causing 4 test failures.

## How to test
```bash
make e2e-no-build BATS_FOLDER_PATTERN=001-system BATS_FILE_PATTERN=01-system-context
```

### DIFF
--- a/mesheryctl/tests/e2e/001-system/01-system-context.bats
+++ b/mesheryctl/tests/e2e/001-system/01-system-context.bats
@@ -30,7 +30,7 @@ setup() {
 
    assert_failure
    assert_output --partial "Error"
-   assert_output --partial "only one argument must be provided"
+   assert_output --partial "missing a context name"
 }
 
 @test "given a valid url as an argument when running mesheryctl system context create --url invalid-url then an the context is displayed" {
@@ -112,7 +112,7 @@ setup() {
 
    assert_failure
    assert_output --partial "Error"
-   assert_output --partial "only one argument must be provided"
+   assert_output --partial "missing a context name to switch to"
 }
 
 @test "given a valid context-name provided when running mesheryctl system context switch context-name then the current context is switched to specified context" {
@@ -132,7 +132,7 @@ setup() {
 
    assert_failure
    assert_output --partial "Error"
-   assert_output --partial "only one argument must be provided"
+   assert_output --partial "missing a context name to delete"
 }
 
 @test "given an invalid context-name provided when running mesheryctl system context delete invalid-context-name then an error message is displayed" {
--- a/mesheryctl/tests/e2e/001-system/01-system-context.bats
+++ b/mesheryctl/tests/e2e/001-system/01-system-context.bats
@@ -30,7 +30,7 @@ setup() {
 
    assert_failure
    assert_output --partial "Error"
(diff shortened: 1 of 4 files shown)


mesheryctl/tests/e2e/001-system/01-system-context.bats
```diff
@@ -30,7 +30,7 @@ setup() {
 
    assert_failure
    assert_output --partial "Error"
-   assert_output --partial "provide a context name"
```
marblom007, line 33: Why did we move away from "provide a context name"? It is a lot more descriptive that the generic error messages, which is crap. It's phrased like total crap "only one argument must be provided" - that's non-sense English.

marblom007 (changes requested):

mesheryctl/tests/e2e/001-system/01-system-context.bats
leecalcote: @lekaf974, did we move away from the "provide a context name" response for this command already? The generic response is less helpful.

Ideally, our response is "missing a context name".

leecalcote: And while @marblom007 didn't say it kindly, he's right that the generic message is confusing and poorly phrased English.

Omkar-Ugal: @leecalcote and @marblom007 should I update the error messages in error.go to be more specific per command, or use a single improved message like "missing a context name"?

lekaf974: Yep this should be updated

mesheryctl/tests/e2e/001-system/01-system-context.bats
```diff
@@ -132,7 +132,7 @@ setup() {
 
    assert_failure
    assert_output --partial "Error"
-   assert_output --partial "provide a context name to delete"
+   assert_output --partial "only one argument must be provided"
```
lekaf974: Here as well the error is not helpful

mesheryctl/tests/e2e/001-system/01-system-context.bats
```diff
@@ -112,7 +112,7 @@ setup() {
 
    assert_failure
    assert_output --partial "Error"
-   assert_output --partial "provide exactly one context name"
+   assert_output --partial "only one argument must be provided"
```
lekaf974: Same here

lekaf974 (changes requested):
Like mentioned in comments the message is not helpful, could you update error message to make it more useful for the user

Omkar-Ugal:
> Like mentioned in comments the message is not helpful, could you update error message to make it more useful for the user

@lekaf974 , found a bug in doc.go while working on this PR - `basename` is used instead of `filename` in `os.OpenFile`. Should I fix it in this PR or open a separate one?

lekaf974 (approved):

PR #18478, opened 2026-04-04
fix: prevent nil pointer panic on invalid kubeconfig during controller init

pulkitvats2007-crypto:
##  Bug Summary

During Kubernetes context initialization in `AddCtxControllerHandlers`, failures in kubeconfig generation or client creation were logged but execution continued. This resulted in controllers being initialized with a **nil Kubernetes client**, eventually causing a **fatal nil pointer dereference panic** and crashing the entire Meshery Server.

This was especially critical in multi-cluster setups, where a single invalid or unreachable cluster could bring down the entire management plane.

---

##  Fix

- Properly capture and handle errors from `GenerateKubeConfig()`
- Add early returns after client initialization failure to prevent nil client usage
- Ensure controllers are only initialized when a valid Kubernetes client is available

```go
cfg, err := ctx.GenerateKubeConfig()
if err != nil {
	mch.handleControllerInitError(ctx, err, "Failed to generate kubeconfig")
	return mch
}

client, err := mesherykube.New(cfg)
if err != nil {
	mch.handleControllerInitError(ctx, err, "Failed to create Kubernetes client")
	return mch
}
```

---

##  Impact
- Prevents server-wide crashes caused by nil pointer dereference
- Isolates failures to only the affected Kubernetes context
- Improves stability and reliability in multi-cluster environments
- Ensures graceful handling of invalid or unreachable cluster configurations

### DIFF
--- a/mesheryctl/internal/cli/root/workspaces/workspace_test.go
+++ b/mesheryctl/internal/cli/root/workspaces/workspace_test.go
@@ -16,6 +16,7 @@ var validWorkspaceName = "workspace-test"
 var workspaceDescription = "integration test"
 
 func TestWorkspaces(t *testing.T) {
+	utils.SetupContextEnv(t)
 	// get current directory
 	_, filename, _, ok := runtime.Caller(0)
 	if !ok {


yi-nuo426:
### Code review

No issues found. Checked for bugs.

PR #18479, opened 2026-04-04
[UI] Replace imperative session dialogs with Sistent Modal components

### LINKED ISSUE #18462: [UI] Use Sistent Modal for session expiring / expired flows while keeping refactored session timer logic
### Current Behavior

The refactored proactive session handling in `ui/lib/sessionTimer.ts` and the integration with `ui/lib/data-fetch.ts` (e.g. `recordActivity()`, 401 handling) improves behavior compared to before.

However, the **“Session Expiring Soon”** and **“Session Expired”** UIs are implemented with **imperative DOM** (`document.createElement`, inline styles, hardcoded colors) instead of the **Sistent** modal components used elsewhere in Meshery UI.

### Desired Behavior

- **Keep** the enhanced session management: timer constants, activity tracking, scheduling/dismissal, `/api/user/prefs` refresh for “Stay Logged In”, and `recordActivity` from successful `data-fetch` responses.
- **Render** both dialogs with **`@sistent/sistent`** (`Modal`, `ModalBody`, `ModalFooter`, primary/secondary buttons, etc.) so they match theme, a11y, and design system usage (see e.g. `ui/components/ViewInfoModal.tsx`, `ui/components/connections/meshSync/RegisterConnectionModal.tsx`).

### Environment

- Host OS: Mac Linux Windows
- Platform: Docker Kubernetes

---

Umesh-Tiruvalluru:
**Notes for Reviewers**

 - This PR fixes #18462
 - Replaces imperative DOM manipulation (`document.createElement`, inline styles, hardcoded colors) in session expiring/expired dialogs with Sistent `Modal`, `ModalBody`, `ModalFooter`, `ModalButtonPrimary`, and `ModalButtonSecondary` components.
 - All existing session management logic is preserved: timer constants, activity tracking, scheduling/dismissal, `/api/user/prefs` refresh for "Stay Logged In", `recordActivity` from successful
  `data-fetch` responses, and 401 handling.
 - Session state is managed via a new Redux slice (`sessions`) following the existing slice pattern.
 - Accessibility: `disableEscapeKeyDown`, `disableBackdropClick`, `aria-describedby`, `autoFocus` on primary buttons.
 - Includes tests for the slice (8 tests), session timer (7 tests), and existing data-fetch tests (6 tests) all passing.

### DIFF
--- a/ui/components/SessionModals.tsx
+++ b/ui/components/SessionModals.tsx
@@ -1,0 +1,150 @@
+import React, { useEffect, useState } from 'react';
+import { useSelector } from 'react-redux';
+import {
+  Modal,
+  ModalBody,
+  ModalButtonPrimary,
+  ModalButtonSecondary,
+  ModalFooter,
+  Typography,
+  CircularProgress,
+} from '@sistent/sistent';
+import { SessionState } from '@/store/slices/sessions';
+import { recordActivity, triggerSessionExpired } from 'lib/sessionTimer';
+
+const SessionExpiringModal = () => {
+  const sessionStatus = useSelector(
+    (state: { sessions: { status: SessionState } }) => state.sessions.status,
+  );
+  const [isRefreshing, setIsRefreshing] = useState(false);
+
+  const handleStayLoggedIn = () => {
+    setIsRefreshing(true);
+    fetch('/api/user/prefs', { credentials: 'include' })
+      .then(() => {
+        recordActivity();
+      })
+      .catch(() => {
+        triggerSessionExpired();
+      })
+      .finally(() => {
+        setIsRefreshing(false);
+      });
+  };
+
+  const handleLogout = () => {
+    window.location.href = '/user/logout';
+  };
+
+  return (
+    <Modal
+      open={sessionStatus === 'expiring'}
+      closeModal={handleStayLoggedIn}
+      title="Session Expiring Soon"
+      maxWidth="xs"
+      disableEscapeKeyDown
+      disableBackdropClick
+      aria-describedby="session-expiring-description"
+    >
+      <ModalBody>
+        <Typography variant="body1" id="session-expiring-description">
+          Your session will expire in a few minutes due to inactivity. Would you like to stay logged
+          in?
+        </Typography>
+      </ModalBody>
+      <ModalFooter variant="filled">
+        <div
+          style={{
+            display: 'flex',
+            justifyContent: 'center',
+            gap: '1rem',
+            width: '100%',
+          }}
+        >
+          <ModalButtonSecondary onClick={handleLogout}>Log Out</ModalButtonSecondary>
+          <ModalButtonPrimary onClick={handleStayLoggedIn} disabled={isRefreshing} autoFocus>
+            {isRefreshing ? (
+              <div style={{ display: 'flex', alignItems: 'center' }}>
+                <CircularProgress size={20} style={{ marginRight: '0.5rem', color: '#fff' }} />
+                <Typography variant="body1">Refreshing...</Typography>
+              </div>
+            ) : (
+              'Stay Logged In'
+            )}
+          </ModalButtonPrimary>
+        </div>
+      </ModalFooter>
+    </Modal>
+  );
+};
+
+const SessionExpiredModal = () => {
+  const sessionStatus = useSelector(
+    (state: { sessions: { status: SessionState } }) => state.sessions.status,
+  );
+  const isExpired = sessionStatus === 'expired';
+
+  const redirectTo =
+    typeof window !== 'undefined' && window.location.host.endsWith('3000')
+      ? '/user/login'
+      : typeof window !== 'undefined'
+        ? window.location.href
+        : '/';
+
+  const handleLogin = () => {
+    window.location.href = redirectTo;
+  };
+
+  // Auto-redirect after 5 seconds
+  useEffect(() => {
+    if (!isExpired) return;
+
+    const timer = setTimeout(() => {
+      window.location.href = redirectTo;
+    }, 5000);
+
+    return () => clearTimeout(timer);
+  }, [isExpired, redirectTo]);
+
+  return (
+    <Modal
+      open={isExpired}
+      closeModal={handleLogin}
+      title="Session Expired"
+      maxWidth="xs"
+      disableEscapeKeyDown
+      disableBackdropClick
+      aria-describedby="session-expired-description"
+    >
+      <ModalBody>
+        <Typography variant="body1" id="session-expired-description">
+          Your session has expired. You will be redirected to log in.
+        </Typography>
+      </ModalBody>
+      <ModalFooter variant="filled">
+        <div
+          style={{
+            display: 'flex',
+            justifyContent: 'center',
+            width: '100%',
+          }}
+        >
+          <ModalButtonPrimary onClick={handleLogin} autoFocus>
+            Log In
+          </ModalButtonPrimary>
+        </div>
+      </ModalFooter>
+    </Modal>
+  );
+};
+
+const SessionModals = () => {
+  return (
+    <>
+      <SessionExpiringModal />
+      <SessionExpiredModal />
+    </>
+  );
+};
+
+export default SessionModals;
--- a/ui/components/SessionModals.tsx
+++ b/ui/components/SessionModals.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import {
+  Box,
   Modal,
   ModalBody,
   ModalButtonPrimary,
@@ -8,14 +9,23 @@ import {
   ModalFooter,
   Typography,
   CircularProgress,
+  styled,
 } from '@sistent/sistent';
 import { SessionState } from '@/store/slices/sessions';
 import { recordActivity, triggerSessionExpired } from 'lib/sessionTimer';
 
+const selectSessionStatus = (state: { sessions: { status: SessionState } }) =>
+  state.sessions.status;
+
+const FooterActions = styled(Box)({
+  display: 'flex',
+  justifyContent: 'center',
+  gap: '1rem',
+  width: '100%',
+});
+
 const SessionExpiringModal = () => {
-  const sessionStatus = useSelector(
-    (state: { sessions: { status: SessionState } }) => state.sessions.status,
-  );
+  const sessionStatus = useSelector(selectSessionStatus);
   const [isRefreshing, setIsRefreshing] = useState(false);
 
   const handleStayLoggedIn = () => {
@@ -57,39 +67,30 @@ const SessionExpiringModal = () => {
         </Typography>
       </ModalBody>
       <ModalFooter variant="filled">
-        <div
-          style={{
-            display: 'flex',
-            justifyContent: 'center',
-            gap: '1rem',
-            width: '100%',
-          }}
-        >
+        <FooterActions>
           <ModalButtonSecondary onClick={handleLogout}>Log Out</ModalButtonSecondary>
           <ModalButtonPrimary onClick={handleStayLoggedIn} disabled={isRefreshing} autoFocus>
             {isRefreshing ? (
-              <div style={{ display: 'flex', alignItems: 'center' }}>
-                <CircularProgress size={20} color="inherit" style={{ marginRight: '0.5rem' }} />
+              <Box sx={{ display: 'flex', alignItems: 'center' }}>
+                <CircularProgress size={20} sx={{ mr: '0.5rem', color: 'common.white' }} />
                 <Typography variant="body1">Refreshing...</Typography>
-              </div>
+              </Box>
             ) : (
               'Stay Logged In'
             )}
           </ModalButtonPrimary>
-        </div>
+        </FooterActions>
       </ModalFooter>
     </Modal>
   );
 };
 
 const SessionExpiredModal = () => {
-  const sessionStatus = useSelector(
-    (state: { sessions: { status: SessionState } }) => state.sessions.status,
-  );
+  const sessionStatus = useSelector(selectSessionStatus);
   const isExpired = sessionStatus === 'expired';
 
   const redirectTo =
-    typeof window !== 'undefined' && window.location.host.endsWith('3000')
+    typeof window !== 'undefined' && process.env.NODE_ENV === 'development'
       ? '/user/login'
       : typeof window !== 'undefined'
         ? window.location.href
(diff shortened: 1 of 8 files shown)


saurabhraghuvanshii:
@banana-three-join. Please review.

ui/store/slices/sessions.ts
```diff
@@ -0,0 +1,26 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+export type SessionState = 'idle' | 'expiring' | 'expired';
```
banana-three-join: Maybe it would be better to change the name of the type to ExpirationState? Because a Session in reality can be Active, Idle, Expiring and Expired. It only becomes Expiring when it reaches a timeframe which is close to Expired. So, this gives the impresion that from the moment a Session is born, it instantly becomes Expiring.

ui/components/SessionModals.tsx
```diff
@@ -0,0 +1,149 @@
...
+  // Auto-redirect after 5 seconds
+  useEffect(() => {
+    if (!isExpired) return;
+
+    const timer = setTimeout(() => {
+      window.location.href = redirectTo;
+    }, 5000);
+
+    return () => clearTimeout(timer);
+  }, [isExpired, redirectTo]);
```
banana-three-join: redirectTo is derived data, not React state so it shouldn't be tracked

ui/components/SessionModals.tsx
```diff
@@ -0,0 +1,149 @@
...
+const SessionExpiredModal = () => {
+  const sessionStatus = useSelector(selectSessionStatus);
+  const isExpired = sessionStatus === 'expired';
+
+  const redirectTo =
+    typeof window !== 'undefined' && process.env.NODE_ENV === 'development'
+      ? '/user/login'
+      : typeof window !== 'undefined'
+        ? window.location.href
+        : '/';
```
banana-three-join: On process !== 'development'. window.location.href = window.location.href and the session is still going to be expired so it's just going to loop.

Also, with SSR, there's going to be a mismatch in the evalution between the server and the client so it's better to have a function that evaluates whether it's in dev mode or not and just call it in handleLogin and in the useEffect instead

ui/store/slices/sessions.ts
Umesh-Tiruvalluru: Good point, I'll change the SessionState into ExpirationState. And Should we also add 'Active' state value or is the current set of values sufficient for what the modals need?

banana-three-join: I think after the change in name to the type, it should be fine to only have the three current states because the only responsability of the modal is to track whether a session is about to expire or has already expired.

ui/components/SessionModals.tsx
Umesh-Tiruvalluru: should the redirect always go to `/user/login` in both dev and production? Or should we preserve the current page URL as a query param (e.g., `/user/login?redirect=/current-page`) so the user returns to where they were after re-authenticating?

ui/components/SessionModals.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  return (
+    <Modal
+      open={isExpired}
+      closeModal={handleLogin}
+      title="Session Expired"
+      maxWidth="xs"
+      disableEscapeKeyDown
+      disableBackdropClick
+      aria-describedby="session-expired-description"
```
YASHMAHAKAL, line 131: @Umesh-Tiruvalluru closeModal={handleStayLoggedIn} wires the × close button directly to the handleStayLoggedIn function. In Sistent Modal component source, it passes closeModal as the onClick handler for the × button (modal-close-btn). This means clicking × secretly triggers a fetch(/api/user/prefs) network request instead of just closing the dialog. Should we wire closeModal to a neutral dismiss action (e.g., () => store.dispatch(setSessionState(idle))) or handleLogout ?

ui/components/SessionModals.tsx
banana-three-join: I think the latter would be better so users don't have to manually navigate back to the site in which they were in while also avoiding the endless refresh after the timer is off

ui/components/SessionModals.tsx
Umesh-Tiruvalluru: Yes you are right @YASHMAHAKAL , the x button on the expiring modal shouldn’t trigger an API call. You also mentioned wiring it to () => store.dispatch(setSessionState('idle')), that would dismiss the modal but without resetting the activity timer or rescheduling the warning, so the warning could fire again immediately. recordActivity() handles both (resets timer + dispatches idle).
Since we already have disableEscapeKeyDown and disableBackdropClick, should the X button:
  1. closeModal={() => recordActivity()} dismiss and reset timer (user acknowledged the warning)
  2. closeModal={() => {}}, do nothing, force the user to pick "Stay Logged In" or "Log Out"
  which is approach do you think is better?

ui/components/SessionModals.tsx
```diff
@@ -0,0 +1,149 @@
...
+          recordActivity();
+        } else {
+          triggerSessionExpired();
+        }
+      })
+      .catch(() => {
+        triggerSessionExpired();
+      })
+      .finally(() => {
+        setIsRefreshing(false);
```
AnkitRewar11: If the modal unmounts before fetch completes, setIsRefreshing(false) will run on an unmounted component. Add an AbortController or isMounted ref to prevent this.

ui/lib/sessionTimer.ts
```diff
@@ -34,61 +41,21 @@ function scheduleWarning() {
...
+  store.dispatch(setSessionState('idle'));
+}
+
+export function triggerSessionExpired() {
+  if (store.getState().sessions.status === 'expired') return;
+  if (warningTimer) clearTimeout(warningTimer);
+  store.dispatch(setSessionState('expired'));
 }
 
 export function startSessionTimer() {
```
AnkitRewar11, line 63: resetOnInteraction is being created fresh on every call, so if startSessionTimer runs more than once you'll end up with stacked listeners and no way to remove them since the reference is gone.
moving it outside + a started flag should do it

ui/store/slices/sessions.ts
```diff
@@ -0,0 +1,26 @@
...
+  status: 'idle',
+};
+
+const sessionsSlice = createSlice({
+  name: 'sessions',
+  initialState,
+  reducers: {
+    setSessionState: (state, action: PayloadAction<SessionState>) => {
+      // expired is an absorbing state — cannot transition out of it
+      if (state.status === 'expired' && action.payload !== 'expired') return;
```
AnkitRewar11, line 19: Making expired an absorbing state is a clean pattern prevents race conditions nicely. 👍

ui/store/slices/sessions.ts
Umesh-Tiruvalluru: Renamed SessionState to ExpirationState. Updated in the latest commit.

Umesh-Tiruvalluru:
@YASHMAHAKAL @banana-three-join @AnkitRewar11 Resolved all review feedback in the latest commit.

ui/store/slices/sessions.ts
AnkitRewar11: I noticed that the rename to `ExpirationState` was done in `SessionModals.tsx`, but not in `sessions.ts`. This will likely cause a TypeScript error since `ExpirationState` isn’t exported there. rename the type in `sessions.ts` as well to keep things consistent.

ui/lib/__tests__/sessionTimer.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    startSessionTimer();
+
+    const registeredEvents = addEventSpy.mock.calls.map((call) => call[0]);
+    expect(registeredEvents).toContain('click');
+    expect(registeredEvents).toContain('keydown');
+
+    addEventSpy.mockRestore();
+  });
+
+  it('startSessionTimer does not register duplicate listeners', () => {
```
AnkitRewar11, line 45: This test is passing only because `started` remains true from the previous test. Since `beforeEach` does not reset module-level state, the test may fail if the execution order changes. Reset `started` between tests to ensure consistent behavior.

ui/components/SessionModals.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const handleLogin = () => {
+    window.location.href = getLoginRedirectUrl();
+  };
+
+  useEffect(() => {
+    if (!isExpired) return;
+
+    const timer = setTimeout(() => {
+      window.location.href = getLoginRedirectUrl();
+    }, 5000);
```
AnkitRewar11: 5000 should be a named constant at the top of the file like const AUTO_REDIRECT_DELAY_MS = 5000 makes the intent clear without having to guess what the number means.

ui/components/SessionModals.tsx
banana-three-join: I think the first approach is better

ui/lib/sessionTimer.ts
```diff
@@ -8,18 +8,27 @@
...
+import { store } from '../store';
+import { setSessionState } from '../store/slices/sessions';
+
 const SESSION_TIMEOUT_MS = 60 * 60 * 1000; // 1 hour (matches server default)
 const WARNING_BEFORE_MS = 5 * 60 * 1000; // Warn 5 minutes before expiry
 
 let lastActivity = Date.now();
 let warningTimer: ReturnType<typeof setTimeout> | null = null;
-let warningShown = false;
+let started = false;
```
banana-three-join, line 19: As Ankit pointed out above, I think it would be good to add a small clean up function to bring back started back to it's original default state in case the sessionTimer needs to be used multiple times under the same context.

ui/components/SessionModals.tsx
Umesh-Tiruvalluru: The line you referenced (115) corresponds to the expired modal, where `closeModal={handleLogin}` redirects to login  not `recordActivity()`. The `closeModal={() => recordActivity()}` you mentioned is actually on the expiring modal (line 69), which aligns with the first approach discussed earlier. Could you clarify which modal you intended to refer to?

ui/store/slices/sessions.ts
Umesh-Tiruvalluru: Sorry, my bad. I did update it in sessions.ts as well, just forgot to commit that file. I’ll push the change shortly.

ui/lib/__tests__/sessionTimer.test.ts
Umesh-Tiruvalluru: fixed. Added a stopSessionTimer() function in sessionTimer.ts that resets the started flag, removes the listeners, and clears the pending timer. The test file now calls it in beforeEach/afterEach, and the duplicate-listeners test explicitly calls startSessionTimer() first before spying, so it no longer depends on execution order.

ui/components/SessionModals.tsx
Umesh-Tiruvalluru: extracted const AUTO_REDIRECT_DELAY_MS = 5000 at the top of SessionModals.tsx and used it in the setTimeout

ui/lib/sessionTimer.ts
Umesh-Tiruvalluru: Added, stopSessionTimer() in ui/lib/sessionTimer.ts resets started to false, removes the click/keydown listeners, and clears the pending warning timer. This makes the module safely re-initializable and unblocks the test isolation issue Ankit raised.

ui/components/SessionModals.tsx
banana-three-join: I was refering to the first approach that you proposed but bear in mind that the timer isn't going to be reset because the state is currently 'expiring' so recordActivity is just gonna set the state to idle while the timer in the background waits to change the state to ;expired; so Yash's suggestion makes sense because the currently flow of the code is going to end up having the same outcome.

Umesh-Tiruvalluru: @banana-three-join This is resolved in the latest commit and followed first approach

ui/components/SessionModals.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  useEffect(() => {
+    return () => {
+      isMountedRef.current = false;
+    };
+  }, []);
+
+  const handleStayLoggedIn = () => {
+    setIsRefreshing(true);
+    fetch('/api/user/prefs', { credentials: 'include' })
```
AnkitRewar11: If the network is slow or the request hangs, the user can get stuck in the “Refreshing…” state with no timeout. Add an `AbortController` with around a 10s timeout to handle this.

ui/components/SessionModals.tsx
```diff
@@ -0,0 +1,156 @@
...
+  CircularProgress,
+  styled,
+} from '@sistent/sistent';
+import { ExpirationState } from '@/store/slices/sessions';
+import { recordActivity, triggerSessionExpired } from 'lib/sessionTimer';
+
+const AUTO_REDIRECT_DELAY_MS = 5000;
+
+const selectSessionStatus = (state: { sessions: { status: ExpirationState } }) =>
+  state.sessions.status;
```
AnkitRewar11: This selector is defined locally here. sessions.ts does not 
export it. 

If another file needs session status in future, the same 
selector will be written again. And if sessions slice 
structure changes (e.g. status is renamed), it will need 
to be updated in every file manually missing one will 
cause a bug.

Please add this to sessions.ts:

export const selectExpirationStatus = (state: { sessions: { status: ExpirationState } }) =>
  state.sessions.status;

And import it here instead of defining locally.

ui/lib/data-fetch.ts
```diff
@@ -57,7 +9,7 @@ const dataFetch = (url, options = {}, successFn, errorFn) => {
   fetch(url, options)
     .then((res) => {
       if (res.status === 401 || res.redirected) {
-        showSessionExpiredAndRedirect();
+        triggerSessionExpired();
         return new Promise(() => {});
```
AnkitRewar11: Returning new Promise(() => {}) here means the promise never 
resolves or rejects the calling function and UI will hang 
forever waiting for a response that never comes.

Please replace with:

return Promise.reject(new Error('Session expired'));

This way the caller knows the request failed and can handle 
it properly instead of hanging indefinitely.

AnkitRewar11:
@Umesh-Tiruvalluru Great work overall! Everything looks good and well put together. Just a small thing please go through the points mentioned once before moving forward.

Umesh-Tiruvalluru:
@AnkitRewar11  All review comments have been addressed in the latest commit. Please take another look when you have a free time. Thanks!

banana-three-join (commented):
Looks good to me! Nice job and appreciate the throughout work :]

Umesh-Tiruvalluru:
@banana-three-join Thanks :)

AnkitRewar11:
Everything looks good now 👍good work.

Umesh-Tiruvalluru:
Thanks a lot @AnkitRewar11! Really appreciate the detailed reviews and guidance throughout the PR. Since everything looks good now, could you please help merge it when you get a chance? Thanks again 🙂

Issue #18409, opened 2026-04-02
[Server] GraphQL generated code panics on unknown fields instead of returning errors


simihablo:
## Description

`server/internal/graphql/generated/generated.go` contains 46 occurrences of `panic()` calls triggered when an unknown GraphQL field is encountered during execution.

```go
panic("unknown field " + strconv.Quote(field.Name))
```

## Why This Is a Problem

Using `panic()` for an expected error condition (receiving an unknown field name) causes the entire server process to crash rather than returning a proper GraphQL error to the client. This means:

1. **Server crash**: A single malformed or unexpected GraphQL query can bring down the Meshery server
2. **No graceful degradation**: Clients receive a connection reset instead of a structured error response
3. **Hard to recover**: `panic` unwinds the goroutine stack; recovery requires a `recover()` at the top level which may mask other real panics

This pattern appears 46 times throughout the generated file.

## Expected Behavior

Unknown fields should return a GraphQL error, not a panic:

```go
return graphql.Null  // or propagate an error through the resolver context
```

The gqlgen code generator should be configured (via `gqlgen.yml`) to handle this gracefully, or a custom `recover` middleware should be added to the GraphQL handler to catch these panics and convert them to errors.

## Steps to Reproduce

1. Send a GraphQL query referencing a field name that doesn't match the schema
2. Observe server panic/crash in logs

## Suggested Fix

- Add `recover` middleware to the GraphQL HTTP handler
- Or regenerate the GraphQL code with gqlgen's `omit_unreachable_definitions` and ensure schema alignment

ashwanisingh011:
Hey @simihablo I would love to work on this issue. Can you assign this issue to me

yug54475:
@simihablo From going through gqlgen’s code (v0.17.85), HTTP requests are already wrapped with a `recover()` in `ServeHTTP`, so panics from generated code `(like panic("unknown field ..."))` should not crash the server for normal queries.

However, in the WebSocket transport there are some unprotected paths (e.g., parts of subscribe before the goroutine starts, response handling, and some goroutines like `nextMessageWithTimeout`) where panics are not recovered. In those cases, the WebSocket connection can terminate abruptly, which may appear as a “server crash” from the client side.

Using `srv.SetRecoverFunc(...)` would handle execution-level panics cleanly, and adding recovery in those WebSocket paths would make the behavior more consistent across transports. If this approach sounds good, I'd be happy to work on it.

simihablo:
[...]

PR #18482, opened 2026-04-04
fix: add recover middleware to graphql server to prevent panics

ashwanisingh011:
### Description
This PR fixes the issue where the Meshery server panics and crashes when encountering unknown GraphQL fields. 

Instead of modifying the auto-generated `generated.go` file (which would be overwritten on the next generation), this implements the architectural safety net provided by `gqlgen`. I added a `srv.SetRecoverFunc` right where the GraphQL server is instantiated. 

This ensures that any panics—both from standard HTTP requests and unprotected WebSocket transport paths—are caught gracefully, logged to the Meshery logger for debugging, and return a standard "internal server error" to the client without bringing down the entire server process.

**Fixes:** #18409


YASHMAHAKAL (approved):
looks good to me @ashwanisingh011

rishiraj38 (approved):

rishiraj38:
@ashwanisingh011 Thanks for the effort.

PR #18485, opened 2026-04-04
Fix/mesheryctl return errors batch1

### LINKED ISSUE #18484: [bug] mesheryctl commands log errors and return nil instead of propagating errors
## Current Behavior
scripts to detect failures.

## Affected Files

## Expected Behavior

## Possible Fix
Replace patterns like:
utils.Log.Error(err)
return nil

With:
return err

adityaupasani2:
Fixes #18484 (partial - batch 1)

## Description
Multiple mesheryctl commands were logging errors using `utils.Log.Error(err)` 
but returning `nil` instead of the actual error, causing commands to silently 
succeed even when they failed.

## Changes
- `restart.go` — return err instead of nil in 2 error cases
- `start.go` — return err instead of nil in 3 error cases

## Testing
- `mesheryctl system restart` ✅
- `mesheryctl system start` ✅
- `go build ./...` ✅

### DIFF
--- a/mesheryctl/internal/cli/root/system/config.go
+++ b/mesheryctl/internal/cli/root/system/config.go
@@ -40,7 +40,7 @@ func getContexts(configFile string) ([]string, error) {
 	mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
 	if err != nil {
 		utils.Log.Error(err)
-		return nil, nil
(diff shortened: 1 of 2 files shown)


lekaf974:
please start by review existing PR instead of create new once

PR #18487, opened 2026-04-04
Fix/mesheryctl return errors batch3

### LINKED ISSUE #18484: [bug] mesheryctl commands log errors and return nil instead of propagating errors

adityaupasani2:
Fixes #18484 (partial - batch 3)

## Description
Registry commands were logging errors using `utils.Log.Error(err)` but 
returning `nil` instead of the actual error.

## Changes
- `publish.go` — return err instead of nil in 10 error cases
- `update.go` — return err instead of nil in 2 error cases

## Testing
- `go build ./...` ✅


server/handlers/meshsync_handler.go
lekaf974, line 1: please avoid mix server and cli in same PR if not related to a specific request or issue which is not the case heere. remove this changes from this PR and created a dedicated issue with description to explain why you are doing it

adityaupasani2: @lekaf974 sorry for the confusion and the mess! To clarify , the 
server/handlers/meshsync_handler.go is not part of this PR's changes. 
It's appearing in the diff only because PR #18270 (which modifies that 
file) hasn't been merged into master yet. This PR only touches 
registry/publish.go and registry/update.go. Once #18270 merges, the 
diff will be clean. Didn't mean to create chaos. Appreciate your patience!

lekaf974: So if it is the case it means it should not be in current branch

lekaf974 (changes requested):
Please review the description and the existing PR, this is on partially fixing

PR #18436, opened 2026-04-02
fix: design uses meshkit error properly + return error properl

Omkar-Ugal:
Fixes #18378

- This PR fixes #

- [x] Yes, I signed my commits.

PR #18488, opened 2026-04-04
Fix/mesheryctl return errors batch4

### LINKED ISSUE #18484: [bug] mesheryctl commands log errors and return nil instead of propagating errors

adityaupasani2:
Fixes #18484 (partial - batch 4)

## Description
Design and model commands were logging errors using `utils.Log.Error(err)` 
but returning `nil` instead of the actual error.

## Changes
- `design/apply.go` — return err instead of nil in 1 error case
- `model/import.go` — return err instead of nil in 1 error case

## Testing
- `go build ./...` ✅


mesheryctl/internal/cli/root/design/apply.go
lekaf974, line 1: already a PR for this, your review is welcome #18436, please revert this changes

mesheryctl/internal/cli/root/model/import.go
lekaf974, line 1: this is not what is expected from the issue, if you want to work on it read all expectation before request a review

server/handlers/meshsync_handler.go
lekaf974, line 1: don't mix changes between cli and server if not a related work

lekaf974:
model import as well as an opened PR #18420 closing this one since all changes already covered, your review is welcomed

PR #18495, opened 2026-04-05
fix: Concurrent Map Read/Write in OperatorTracker Causes Server Panic  

pulkitvats2007-crypto:
##  Bug Summary

`OperatorTracker` uses a shared `map[string]bool` to track deployment status across Kubernetes contexts but does not use its defined mutex (`mx`) in `Undeployed` and `IsUndeployed`.  

During multi-cluster operations, Meshery spawns concurrent goroutines that read/write to this map, leading to fatal error: concurrent map read and map write
---

##  Root Cause
Unsynchronized concurrent access to a shared map (race condition).

---

##  Fix
- Added mutex locking:
  ```go
  ot.mx.Lock()
  defer ot.mx.Unlock()
  ```
- Applied to both Undeployed and IsUndeployed
- Added TestOperatorTracker_ConcurrentAccess with 100 goroutines to validate thread safety

---

##  Impact
- Prevents fatal server crashes
- Ensures safe multi-cluster operations
- Improves overall stability of Meshery in production


rishiraj38 (commented):
@pulkitvats2007-crypto The mutex fix in meshery_controllers.go looks good! 👍

Could you help me understand the changes in .github/workflows/kanvas.yml? They seem unrelated to the concurrent map fix.

.github/workflows/kanvas.yml
rishiraj38, line 1: Are these fixes for a currently failing CI workflow? If so, it might be better to split them into a separate PR so this fix can be merged independently.

rishiraj38:
@saurabhraghuvanshii, could you please take a look at this PR?

.github/workflows/kanvas.yml
pulkitvats2007-crypto: Yes, these changes are specifically to address a currently failing test case in the CI workflow. I’ll be reverting the changes and will open a separate PR to handle the CI fixes independently.

pulkitvats2007-crypto:
> @pulkitvats2007-crypto The mutex fix in meshery_controllers.go looks good! 👍
> 
> Could you help me understand the changes in .github/workflows/kanvas.yml? They seem unrelated to the concurrent map fix.

Thanks for the review! 

I initially included those .github/workflows/kanvas.yml changes because I noticed the Kanvas Snapshot CI job was consistently failing with a jq: parse error, and I wanted to patch it to get the CI passing again on this PR. 

However, you're absolutely right that it's unrelated to the concurrent map fix. I have reverted those workflow changes to keep this PR strictly scoped to the OperatorTracker fix!

pulkitvats2007-crypto:
Thanks @simihablo for the updates and deploying this to staging—really appreciate the assist!

rishiraj38 (approved):

arjunmehta-git:
Great find and fix, @pulkitvats2007-crypto! The `OperatorTracker` concurrency bug — where `Undeployed` and `IsUndeployed` were accessing the shared map without the mutex that was already defined on the struct — is exactly the kind of latent race condition that only surfaces under concurrent load in multi-cluster environments. The fix is minimal and correct: apply the mutex that was already there but not being used in these two paths.

The inclusion of `TestOperatorTracker_ConcurrentAccess` with 100 goroutines is the right way to validate this. Please ensure it runs with `-race` in CI.

One thought: since `IsUndeployed` is a read-only operation, consider using `RLock`/`RUnlock` instead of the full `Lock`/`Unlock` — this allows multiple concurrent readers without blocking each other, which is beneficial for a status-check function that may be called frequently during cluster polling. If `mx` is a `sync.RWMutex`, this is a straightforward improvement.

Also: this PR and #18565 both touch `meshery_controllers.go` and `meshery_controllers_test.go`. If #18565 lands first, please rebase this PR to avoid conflicts and verify the mutex usage is consistent across both changes.

Thank you for the continued focus on concurrency correctness in the controllers layer — this work directly improves production stability for multi-cluster deployments.

PR #18500, opened 2026-04-05
[server] Fix double status write in workspace/environment update handlers

### LINKED ISSUE #18499: Bug: Incorrect Error Handling After `WriteHeader(200)` in Workspace/Environment Update Handlers
## Summary
Two backend handlers in Meshery server use a response-writing pattern that can produce incorrect HTTP semantics.

In both handlers, `w.WriteHeader(http.StatusOK)` is called before writing the response body. If body write fails, the error branch attempts to send a `500` status after headers are already committed.

## Affected Areas
- `server/handlers/workspace_handlers.go`
- `server/handlers/environments_handlers.go`

## Exact Evidence
- `server/handlers/workspace_handlers.go:191` sends `200`:
  ```go
  w.WriteHeader(http.StatusOK)
  ```

server/handlers/environments_handlers.go:186 sends 200:

````
w.WriteHeader(http.StatusOK)
````
server/handlers/environments_handlers.go:190 tries to send 500 after write failure:

````
http.Error(w, "Failed to write response", http.StatusInternalServerError)
````
## Current Behavior
Handler sends 200 OK headers first.
If w.Write(respJSON) fails (e.g., broken pipe, client disconnect), error branch executes.
Error branch calls http.Error(..., 500) after headers are already committed.
Result: attempted error status is ineffective; clients may still observe 200 or incomplete/truncated output with misleading status semantics.

## Expected Behavior
Response status and headers should be written exactly once.
If body write fails after response starts, handler should not attempt a second status/header write.
Error handling should be consistent with Go net/http semantics and should not indicate success for failed response writes.

## Why This Is Valuable
Prevents misleading success responses in critical update APIs.
Improves correctness for API consumers and integrations.
Reduces confusion in debugging scenarios where server logs show an error but clients do not receive an appropriate failure signal.
Fix can be applied consistently across handlers to improve reliability.

## Environment
Host OS: Linux
Platform: Docker / Kubernetes
Meshery Server Version: stable-v (replace with exact version used)
Meshery Client Version: stable-v (replace with exact version used)

## Steps to Reproduce
Send a valid PUT request to:
/api/workspaces/{id} or
/api/environments/{id}
Ensure request reaches update handler and response is being written.
Simulate a response-write failure (for example, close/disconnect client socket during server write).
Observe:
server logs hit write error path,
code attempts http.Error(..., 500) after 200 header was already sent.

##Suggested Fix
Use a single-write response strategy and avoid calling http.Error after response has started.

## Possible approaches:

Marshal first, then write body without explicit WriteHeader(200) (let Go send implicit 200 on successful write).
Keep WriteHeader(200) but on w.Write error only log and return (do not call http.Error).
Refactor response helper(s) to ensure headers/status are committed once and shared across handlers.

## Optional Test Coverage Suggestion
Add unit tests for these handlers using a custom/failing http.ResponseWriter (or wrapped writer) that returns an error on Write to verify:

no second status write is attempted after WriteHeader,
handler behavior remains deterministic under write failures.

server/handlers/workspace_handlers.go:195 tries to send 500 after write failure:

```
http.Error(w, "Failed to write response", http.StatusInternalServerError)
```

---

### PR DESCRIPTION
**Pull request by Nsanjayboruds**
### Description
This PR fixes incorrect HTTP error handling in:
- `UpdateWorkspaceHandler`
- `UpdateEnvironmentHandler`

Both handlers previously called `w.WriteHeader(http.StatusOK)` before `w.Write(respJSON)`, and then attempted `http.Error(..., 500)` if `Write` failed.  
Since headers are already sent, the later `500` cannot be applied, which can cause misleading success semantics.

This change removes that double-write pattern and keeps response behavior aligned with Go `net/http` semantics.

### What Changed
- Updated `server/handlers/workspace_handlers.go`
  - Removed explicit `w.WriteHeader(http.StatusOK)` before write.
  - Removed `http.Error(...)` call from the `w.Write` failure branch.
- Updated `server/handlers/environments_handlers.go`
  - Removed explicit `w.WriteHeader(http.StatusOK)` before write.
  - Removed `http.Error(...)` call from the `w.Write` failure branch.

### Why This Fix
- Prevents attempting to send a second status code after response has started.
- Avoids misleading API behavior in write-failure scenarios.
- Keeps handler behavior predictable and standards-compliant.

### Validation
Ran:
```bash
go test ./server/handlers/...
```
Result:

image

**Notes for Reviewers**

- This PR fixes #18499


rishiraj38 (dismissed):

arjunmehta-git:
Precise and correct fix, @Nsanjayboruds! The double-write pattern — calling `w.WriteHeader(http.StatusOK)` before `w.Write(...)`, then attempting `http.Error(..., 500)` if the write fails — is a subtle but real bug in Go HTTP handlers. Once headers are committed, you cannot change the status code, so the `500` would be silently dropped and the client would receive a misleading success response even when the server-side write failed.

The fix is exactly right: remove the explicit `WriteHeader` call (Go's `http.ResponseWriter` will send `200` implicitly on the first `Write`) and remove the now-unreachable `http.Error` call. Clean, minimal, and standards-compliant.

One small recommendation: consider adding a log statement in the `w.Write` failure branch rather than silently discarding the write error. Even if we can't change the HTTP status at that point, logging it ensures the failure is visible in server logs for operators:
```go
if _, err := w.Write(respJSON); err != nil {
    logrus.Errorf("failed to write response: %v", err)
}
```

Thank you for the careful attention to Go HTTP semantics. These small correctness improvements accumulate into a more reliable server.

Nsanjayboruds:
Hey @leecalcote and @arjunmehta-git, this PR has already been approved by @rishiraj38. Could you please review it as well? Thank you!

PR #18508, opened 2026-04-06
Fix RTK Query token endpoints hitting wrong URL

yi-nuo426:
## Summary
- The `getToken` and `getAccessToken` RTK Query endpoints were calling `GET /api/user/token`, which is the **OAuth callback handler** that redirects and returns HTML.
- Changed both to call `GET /api/token`, which is the **token extraction endpoint** that returns JSON (`{"token": "..."}`)
- This fixes the `SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON` error seen in the browser console.

### Route reference (`server/router/server.go`):
- `/api/user/token` (line 371) → `TokenHandler` — OAuth callback, sets cookies and **redirects to `/`** (HTML response)
- `/api/token` (line 380) → `ExtractToken` — returns JSON with the current token from cookies

The `/api/user/token` URL in `User.tsx` (OAuth login source callback) is **not** changed, as that usage is correct.

## Test plan
- [ ] Verify `getAccessToken` RTK Query returns a valid JSON token without console errors
- [ ] Verify OAuth login flow still works (the `/api/user/token` callback in `User.tsx` is untouched)

### DIFF
--- a/ui/rtk-query/user.ts
+++ b/ui/rtk-query/user.ts
@@ -43,7 +43,7 @@ export const userApi = api
         invalidatesTags: [Tags.LOAD_TEST_PREF],
       }),
       getToken: builder.query({
-        query: () => `/api/token`,
+        query: () => ({ url: `/api/token`, method: 'GET', credentials: 'include' }),
         method: 'GET',
       }),
       getUserPref: builder.query({


leecalcote:
The fix is correctly in meshery/meshery, not in the Kanvas extension. Here's why:

Neither repo sets overrideExisting on injectEndpoints, so RTK Query uses first-registered wins. The meshery UI loads first and injects getAccessToken with the broken /api/user/token URL. When extensions later inject their own getAccessToken (with the correct /api/token), it's silently ignored because the name is already registered.

So even though the extensions repo has the right URL, the meshery UI's broken version takes precedence at runtime. The PR meshery/meshery#18508 is the correct fix — no changes needed in meshery-extensions.

leecalcote:
> The fix is correctly in meshery/meshery, not in the Kanvas extension. Here's why:
> 
> Neither repo sets overrideExisting on injectEndpoints, so RTK Query uses first-registered wins. The meshery UI loads first and injects getAccessToken with the broken /api/user/token URL. When extensions later inject their own getAccessToken (with the correct /api/token), it's silently ignored because the name is already registered.
> 
> So even though the extensions repo has the right URL, the meshery UI's broken version takes precedence at runtime. The PR #18508 is the correct fix — no changes needed in meshery-extensions.

I do anticipate that we Will return to this shortly and make additional changes to shift around the endpoints.

PR #18559, opened 2026-04-07
server: fix DeleteMesheryPattern swallowing DB errors

### LINKED ISSUE #18502: [Bug] DeleteMesheryPattern silently swallows DB errors - delete always reports success (HTTP 200) even on failure
### Summary
`DeleteMesheryPattern` in `server/models/meshery_pattern_persister.go` silently swallows all database errors and always returns `nil` as the error, which causes the HTTP handler to always respond with `HTTP 200 OK` - even when the delete completely failed at the database level. The user is told "Design deleted!" but the design is still in the database.

---

## [Location] Location

| File | Lines |
|------|-------|
| `server/models/meshery_pattern_persister.go` | Lines 155-159 |
| `server/handlers/meshery_pattern_handler.go` | Lines 648-673 |

---

### Part 1 - Persister (`server/models/meshery_pattern_persister.go`)

```go
// DeleteMesheryPattern takes in a profile id and delete it if it already exists
func (mpp *MesheryPatternPersister) DeleteMesheryPattern(id uuid.UUID) ([]byte, error) {
    pattern := MesheryPattern{ID: &id}
    mpp.DB.Delete(&pattern)  // <- GORM error is NEVER captured or returned [WARNING]

    return marshalMesheryPattern(&pattern), nil  // <- ALWAYS returns nil error
}
```

`mpp.DB.Delete(&pattern)` returns a `*gorm.DB` result object that carries `.Error`. This error is **never checked** and **never returned**. The function always returns `nil` as the error, regardless of what actually happened in the database.

### Part 2 - Handler (`server/handlers/meshery_pattern_handler.go`)

Because the persister always returns `nil` error, the handler **never enters the error block**. Instead it always:
1. Unmarshals the response as if the delete succeeded
2. 2. Publishes an `Informational` success event
3. 3. Returns **HTTP 200 OK** to the client
```go
resp, err := provider.DeleteMesheryPattern(r, patternID)
if err != nil {
    // This block is NEVER reached - because err is always nil
    ...
    return
}

// This code always runs - even when the DB delete failed
_ = json.Unmarshal(resp, &mesheryPattern)
event := eventBuilder.WithSeverity(events.Informational).WithDescription(...).Build()
_ = provider.PersistEvent(*event, nil)
go h.config.EventBroadcaster.Publish(userID, event)
rw.Header().Set("Content-Type", "application/json")
_, _ = fmt.Fprint(rw, string(resp))  // <- HTTP 200, even on DB failure
```

---

## [Impact] Real-World Impact

| Scenario | What User Sees | What Actually Happened |
|----------|----------------|------------------------|
| DB timeout during delete | [OK] "Design deleted!" | [FAIL] Design still in DB |
| Design ID doesn't exist | [OK] "Design deleted!" | [FAIL] Nothing was deleted |
| DB write permission error | [OK] "Design deleted!" | [FAIL] Delete was rejected silently |
| DB disk full / corruption | [OK] "Design deleted!" | [FAIL] Design not deleted |

The user refreshes the page and the design is still there. There are no error logs, no events, and no indication anything went wrong. This is a silent data integrity failure.

---

## [Fix] Proposed Fix

**Fix the persister to capture and return the database error:**

```go
// BEFORE (broken)
func (mpp *MesheryPatternPersister) DeleteMesheryPattern(id uuid.UUID) ([]byte, error) {
    pattern := MesheryPattern{ID: &id}
    mpp.DB.Delete(&pattern)  // error silently dropped
    return marshalMesheryPattern(&pattern), nil
}

// AFTER (fixed)
func (mpp *MesheryPatternPersister) DeleteMesheryPattern(id uuid.UUID) ([]byte, error) {
    pattern := MesheryPattern{ID: &id}
    result := mpp.DB.Delete(&pattern)
    if result.Error != nil {
        return nil, result.Error  // propagate the real error up
    }
    return marshalMesheryPattern(&pattern), nil
}
```

The same fix should be applied to `DeleteMesheryPatterns` (batch delete) at lines 162-172, which has the identical pattern of swallowing GORM errors.

---

## [Ref] Related Bug (same handler)

In `DeleteMesheryPatternHandler` (lines 653 and 657), `http.Error` is called **twice** inside the error block:

```go
h.log.Error(errPatternDelete)
http.Error(rw, errPatternDelete.Error(), http.StatusInternalServerError)  // line 653
event := eventBuilder.WithSeverity(events.Error)...Build()
http.Error(rw, errPatternDelete.Error(), http.StatusInternalServerError)  // line 657 - DUPLICATE [WARNING]
```

In Go's `net/http`, calling `rw.WriteHeader()` a second time is a no-op (the status code doesn't change), but the **body content is written twice**, resulting in a corrupted response body that clients cannot parse as JSON.

---

## Steps to Reproduce

1. Set up Meshery with a local provider
2. Create a design/pattern
3. Simulate a DB error (e.g., temporarily make the DB read-only, or corrupt the connection)
4. Attempt to delete the design via `DELETE /api/pattern/{id}`
5. Observe: response is `HTTP 200 OK` with the design still in the database
---

## Environment

- **Component**: `server/models/meshery_pattern_persister.go`, `server/handlers/meshery_pattern_handler.go`
- **Provider**: Local Provider (`DefaultLocalProvider`)
-  **Language**: Go
-  **ORM**: GORM
---

## Additional Notes

- The `SaveMesheryPattern` and `GetMesheryPattern` functions in the same file **do** correctly check GORM errors - making this an inconsistency in the codebase.
-  This bug affects both single-pattern delete and the bulk-delete endpoint.
- A similar issue may exist in other `mpp.DB.*` calls that discard return values throughout the persister.

luckyxhq:
## Notes for Reviewers

- This PR fixes #18502


alexquincy (commented):
Thanks! We need to drive the API behavior from the behavior described in github.com/meshery/schemas. Will you have a look at the API operation definitions and confirm their behavior? If that behavior is accurate, then let's make sure that we are importing that package and reusing that model here.

luckyxhq:
I’l review the API operation definitions in the meshery/schemas repository and verify the expected behavior. If the behavior aligns, I import the package and reuse the existing model here to keep the implementation consistent with the schema definitions.

luckyxhq:
hey @alexquincy Thanks for the suggestion! I reviewed the API operation definitions in the Meshery/schemas repository and confirmed the expected behavior. I’ve updated the implementation to align with the schema definitions and ensured that the appropriate models are reused instead of redefining them. Please let me know if any further adjustments are needed.

YASHMAHAKAL:
Hey @luckyxhq  👋

Could you please add your PR to the agenda for the upcoming [dev meeting](https://docs.google.com/document/d/15IEwtTxPOkQXN2r_1LzGXuAv2mIUlmbdtmxpyxXF-KI/edit?tab=t.0) (9 AM CT | 7:30 PM IST)? It would be great if you could attend and briefly present your changes as well. This really helps with getting quicker feedback and smoother reviews.

Appreciate your effort and contributions—thank you! 🙌

arjunmehta-git:
Good catch, @luckyxhq! Silently swallowing database errors in `DeleteMesheryPattern` means failures could appear as successes to the caller — a particularly dangerous pattern for a destructive operation. Surfacing those errors properly is the right behavior and makes the system far more debugable in production.

The PR description is minimal — could you expand it a bit to explain:
1. What the original code was doing that caused DB errors to be swallowed (a brief code snippet helps reviewers verify the fix is complete).
2. How the error is now propagated to the caller and ultimately surfaced to the user or logged.

Also, the deletion of `server/models/meshery_pattern_api_request.go` is a meaningful structural change — worth a sentence explaining what that file contained and why it's no longer needed, so reviewers don't have to reconstruct the reasoning from the diff alone.

The changes span the local provider, remote provider, persister, and the handler — which is the right set of places to touch. Just make sure the error propagation behavior is consistent across all paths (local vs. remote provider). Thank you for the contribution!

Issue #18332, opened 2026-03-29
[mesheryctl] system stop uses meshkit error properly + remove logrus


Omkar-Ugal:
## Parent Issue
#17709

## Current Behavior
In `stop.go`:
- Logrus used directly: log.Info, log.Debug

## Desired Behavior
- Replace logrus references with `utils.Log`

Vydeh07:
I'd like to work on this.Please assign.

ranJan1228:
Hi, I’d like to work on this issue.

From my understanding:
- Replace direct logrus usage with meshkit error handling
- Ensure proper error returns instead of nil
- Align logging with utils.Log where needed

I’ll start by identifying all logrus usages in the system stop command and refactor them accordingly.

Please let me know if there are any specific guidelines I should follow.

ashwanisingh011:
Hey @Omkar-Ugal, I've just raised a PR for this issue here: https://github.com/meshery/meshery/pull/18355
I've refactored the logger in stop.go to use utils.Log and removed the unused logrus import as requested. I've also verified the build locally. Please let me know if any further changes are needed!

lekaf974:
> Hey [@Omkar-Ugal](https://github.com/Omkar-Ugal), I've just raised a PR for this issue here: [Link to your PR]. I've refactored the logger in stop.go to use utils.Log and removed the unused logrus import as requested. I've also verified the build locally. Please let me know if any further changes are needed!

It would have been nicer to give a chance to the people who expressed interest before you raising a PR. We are building a community based on respect of other and this is not.

Adityajs1:
Hi, I'd like to work on this issue. Can I get it assigned, please?

ashwanisingh011:
Hey @lekaf974, thanks for pointing that out — I understand your concern.

When I started working on this issue, it wasn’t assigned, and I didn’t notice the earlier comment requesting assignment. That’s my mistake.
Going forward, I’ll make sure to check for prior interest and communicate before starting work. Happy to collaborate or make any changes if needed.

Thanks for the feedback!

ashwanisingh011:
Hey @Vydeh07, I noticed we both worked on the same issue 😄
I had opened a PR for this a couple of days ago as well. If you'd like, we can collaborate or review each other’s approach.
Great to see more contributors working on this!

Vydeh07:
[...]

PR #18588, opened 2026-04-08
[mesheryctl] Standardize logging and error handling for system config subcommands

ashwanisingh011:
### Description
This PR continues the effort to standardize logging across mesheryctl by replacing direct logrus calls with the internal utils.Log wrapper. It also improves error handling by ensuring errors are returned to Cobra rather than being silenced or causing fatal exits within helper functions.

**Changes**

start.go: Refactored PreRunE to return errors properly; replaced logrus with utils.Log across all service configurations.

config.go: Refactored getContexts, setContext, and setToken to return errors instead of using log.Fatalf, ensuring consistent CLI behavior.

Related to #18332


Junnygram:
@ashwanisingh011 
Great progress on standardizing the logging patterns across `mesheryctl`. Moving away from `log.Fatalf` in our helper functions is a critical move for CLI stability and better error propagation to the Cobra runner.

Here are a few technical observations to further refine this:

#### **1. Optimal Use of Internal Logger**
I noticed several instances (e.g., in `status.go` and `start.go`) where we are still wrapping messages in `fmt.Sprintf` before passing them to `utils.Log.Info`.


> Our internal logger, initialized as a `logger.Handler` in [`mesheryctl/pkg/utils/helpers.go:L260`](https://github.com/meshery/meshery/blob/master/mesheryctl/pkg/utils/helpers.go#L260), already implements the `Infof`, `Errorf`, and `Debugf` methods from MeshKit. 

We should leverage these directly for cleaner code:
*   **Current:** `utils.Log.Info(fmt.Sprintf("\nMeshery endpoint is %s", displayEndpoint))`
*   **Preferred:** `utils.Log.Infof("\nMeshery endpoint is %s", displayEndpoint)`

#### **2. Error Wrapping & Context**
In the refactor of `PreRunE` within `start.go`, some specific error context helpers (like `ErrGetCurrentContext(err)`) were replaced with direct returns. 


> To ensure we don't lose the descriptive error chains we've built in `internal/cli/root/system/error.go`, we should ensure we continue to wrap these errors where the specific context adds value for the end-user.

#### **3. Propagation in `configureDockerServices`**
There are a couple of spots where we log an error but don't return it to the caller. For a truly standardized flow, we should aim to propagate these up so that the Cobra runner can handle the exit state consistently across all subcommands.

lekaf974:
Please remove changes for start which is already covered by another PR already reviewed. I'll take a look on config command no big effort on this one since the command is under deprecation

arjunmehta-git:
Excellent work, @ashwanisingh011! Standardizing logging across `mesheryctl` is one of those incremental improvements that pays compounding dividends — consistent use of `utils.Log` makes log filtering, log level control, and future log format changes much easier to implement uniformly.

The shift from `log.Fatalf` to returning errors in `getContexts`, `setContext`, and `setToken` is particularly valuable. Using `os.Exit` or `log.Fatal` inside helper functions is an anti-pattern that makes code hard to test and hard to compose — returning errors and letting Cobra handle exit behavior is exactly the right direction.

A couple of suggestions to consider:
- Double-check that all callers of the refactored functions in `config.go` properly handle the returned errors. It's easy for one call site to silently ignore an error during a refactor like this.
- The PR title mentions "system start and config subcommands" — if there are other subcommands in the `system` package still using raw `logrus` calls, it might be worth a follow-up issue to track the remaining work so progress is visible.
- Minor: the `[ x ]` checkbox in the PR description has a space before `x` — not a blocker, but worth fixing to keep the PR template clean.

This is exactly the kind of refactoring that helps us build a more maintainable CLI. Thank you for the contribution!

ashwanisingh011:
Thanks for the great feedback, everyone! @lekaf974 I have dropped start.go from this PR so it focuses purely on the config command. @Junnygram I've updated the utils.Log calls to use the native Debugf and Infof methods instead of wrapping fmt.Sprintf. @arjunmehta-git  Good catch on the checkbox! Fixed that, and ensured the helper functions return their errors properly so Cobra handles the exits.

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -40,7 +39,7 @@ func getContexts(configFile string) ([]string, error) {
 	mctlCfg, err := config.GetMesheryCtl(viper.GetViper())
 	if err != nil {
 		utils.Log.Error(err)
-		return nil, nil
+		return nil, err
```
lekaf974, line 40: please take time to read the description of changes required, also check opened PR as example here error will be displayed twice

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -66,28 +65,28 @@ func getContexts(configFile string) ([]string, error) {
...
 	var results []map[string]interface{}
 	err = json.Unmarshal(body, &results)
 	if err != nil {
 		return nil, utils.ErrUnmarshal(err)
 	}
 
 	if results == nil {
-		errstr := "Error unmarshalling the context info, check " + configFile + " file"
+		errstr := "error unmarshalling the context info, check " + configFile + " file"
 		return nil, errors.New(errstr)
```
lekaf974: this is not a meshkit error, please ensure to return a meshkit error

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -147,32 +145,32 @@ mesheryctl system config aks
 		aksCheck.Stderr = os.Stderr
 		err := aksCheck.Run()
 		if err != nil {
-			log.Fatalf("Azure CLI not found. Please install Azure CLI and try again. \nSee https://docs.microsoft.com/en-us/cli/azure/install-azure-cli ")
+			return fmt.Errorf("azure CLI not found. Please install Azure CLI and try again. \nSee https://docs.microsoft.com/en-us/cli/azure/install-azure-cli ")
```
lekaf974: same here this is not a meshkit error

lekaf974 (changes requested):
Please ensure to return meshkit error

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -66,28 +64,27 @@ func getContexts(configFile string) ([]string, error) {
...
+		return nil, ErrUnmarshallConfig(fmt.Errorf("context info is nil, check %s file", configFile))
 	}
 
 	var contextNames []string
 	for _, ctx := range results {
 		ctxname, ok := ctx["name"].(string)
 		if !ok {
-			errstr := "Invalid context name: context name should be a string"
+			errstr := "invalid context name: context name should be a string"
 			return nil, errors.New(errstr)
```
lekaf974: this is not a meshkit error

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -147,32 +143,32 @@ mesheryctl system config aks
...
 		_, err = fmt.Scanf("%s", &aksName)
 		if err != nil {
-			log.Warnf("Error reading AKS cluster name: %s", err.Error())
-			log.Info("Let's try again. Please enter the AKS cluster name:")
+			utils.Log.Warn(fmt.Errorf("error reading AKS cluster name: %s", err.Error()))
+			utils.Log.Info("Let's try again. Please enter the AKS cluster name:")
 			_, err = fmt.Scanf("%s", &aksName)
 			if err != nil {
-				log.Fatalf("Error reading AKS cluster name: %s", err.Error())
+				return fmt.Errorf("error reading AKS cluster name: %s", err.Error())
```
lekaf974: this is not a meshkit error

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -147,32 +142,32 @@ mesheryctl system config aks
 		aksCheck.Stderr = os.Stderr
 		err := aksCheck.Run()
 		if err != nil {
-			log.Fatalf("Azure CLI not found. Please install Azure CLI and try again. \nSee https://docs.microsoft.com/en-us/cli/azure/install-azure-cli ")
+			return ErrK8sConfig(fmt.Errorf("azure CLI not found. Please install Azure CLI and try again (see https://docs.microsoft.com/en-us/cli/azure/install-azure-cli)"))
```
lekaf974: this is an aks error issue nothing related to k8s

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -147,32 +142,32 @@ mesheryctl system config aks
...
 		_, err = fmt.Scanf("%s", &resourceGroup)
 		if err != nil {
-			log.Warnf("Error reading Azure resource group name: %s", err.Error())
-			log.Info("Let's try again. Please enter the Azure resource group name:")
+			utils.Log.Warnf("Error reading Azure resource group name: %v", err)
+			utils.Log.Info("Let's try again. Please enter the Azure resource group name:")
 			_, err = fmt.Scanf("%s", &resourceGroup)
 			if err != nil {
-				log.Fatalf("Error reading Azure resource group name: %s", err.Error())
+				return ErrK8sConfig(fmt.Errorf("error reading Azure resource group name: %s", err.Error()))
```
lekaf974: same here

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -147,32 +142,32 @@ mesheryctl system config aks
...
 		_, err = fmt.Scanf("%s", &aksName)
 		if err != nil {
-			log.Warnf("Error reading AKS cluster name: %s", err.Error())
-			log.Info("Let's try again. Please enter the AKS cluster name:")
+			utils.Log.Warn(fmt.Errorf("error reading AKS cluster name: %s", err.Error()))
+			utils.Log.Info("Let's try again. Please enter the AKS cluster name:")
 			_, err = fmt.Scanf("%s", &aksName)
 			if err != nil {
-				log.Fatalf("Error reading AKS cluster name: %s", err.Error())
+				return ErrK8sConfig(fmt.Errorf("error reading AKS cluster name: %s", err.Error()))
```
lekaf974: same here

lekaf974 (commented):
Can you check how those error are handle by `connection create` command ?

mesheryctl/internal/cli/root/system/config.go
```diff
@@ -217,32 +209,32 @@ mesheryctl system config eks
 		eksCheck.Stderr = os.Stderr
 		err := eksCheck.Run()
 		if err != nil {
-			log.Fatalf("AWS CLI not found. Please install AWS CLI and try again. \nSee https://docs.aws.amazon.com/cli/latest/reference/ ")
+			return ErrK8sConfig(fmt.Errorf("aws CLI not found. Please install AWS CLI and try again (see https://docs.aws.amazon.com/cli/latest/reference/)"))
```
lekaf974: same here this is aws cli error, please ensure doing a self review of all your changes

ashwanisingh011: @lekaf974 - Did a full self-review of the file! I tracked down the AWS CLI error, and while I was in there, I systematically audited and replaced all the generic ErrK8sConfig wrappers for the EKS inputs, GKE generation, and Minikube checks with specific, namespaced [System] errors. All unique codes are registered in error.go, component_info.json is updated, and make error is passing cleanly. Ready for another look!

PragalvaXFREZ (approved):
lgtm, @lekaf974 could you have a 👀

lekaf974 (approved):

PR #18597, opened 2026-04-08
fix(ui): catch and surface extension load failures to the user

### LINKED ISSUE #18593: [Extensibility] Catch and handle when an extension fails to load
### Current Behavior
Here is an example an extension failing to load. This error is from the console of a user's web browser:

```
 [Error] Extension Error: Error: HTTP Error Response: 404 Not Found (/api/provider/extension/provider/navigator/meshmap/collaboratorsPlugin.js)
```
The current behavior is that the user is perpetually stuck on a loading screen. The user has no indication as to whether or not the plug-in is still loading or if it has failed or how much longer it might take to load.

### Desired Behavior
This failure to load should be captured, and the Sistent feedback component presented to the user for ease of reporting of the issue.

### Implementation
1. Use Sistent feedback component.
1. Use existing Error Boundary component.
1. User should be redirected to the dashboard upon submission of the feedback. 
1. While Meshery UI plays a role in catching the air and in redirecting the user, ultimately, Meshery Server is in control of the actual redirect.

---

mateenali66:
when a remote extension fails to load (e.g. HTTP 404 on a plugin JS file), `RemoteComponent` was silently returning an empty fragment. the user got stuck on a loading screen with no indication of what went wrong.

this fixes it by:
- showing the error inline with the existing `Troubleshoot` component (same pattern as `NavigatorExtension`)
- firing a notification via `useNotification` so the error appears in the notification center
- redirecting to the dashboard after 5 seconds so the user isn't stuck
- wrapping the extension route in a Sistent `ErrorBoundary` with `CustomErrorFallback` to catch runtime errors from loaded extensions too

Fixes #18593


hortison (commented):
Thanks for jumping in, @mateenali66! 💯

carlosriosilva (commented):
@mateenali66, want to be sure to address and resolve each of the feedback comments.

mateenali66:
@carlosriosilva all feedback addressed: useEffect deps fixed, hardcoded colors replaced with theme tokens, and error UI scoped to full-page extensions only.

arjunmehta-git:
Well done, @mateenali66! Silent failures in extension loading are one of the most frustrating UX issues in an extensible platform — the user has no idea why things aren't working. This PR addresses that directly and thoughtfully.

The approach here is solid: inline `Troubleshoot` display, notification center surfacing, and a 5-second redirect to prevent the user from getting stuck. Wrapping the extension route in `ErrorBoundary` with `CustomErrorFallback` is the right defensive measure for catching runtime errors from loaded extensions.

A couple of things to think about as you finalize:
- The 5-second auto-redirect is a reasonable default, but consider whether it should be user-configurable or at least have a visible countdown so the user understands what's happening before they're redirected.
- Make sure the notification fired via `useNotification` includes enough context (e.g., which extension failed, the HTTP status code) so the user or an operator can diagnose the root cause without needing to open DevTools.
- It's worth confirming the `ErrorBoundary` reset behavior — does navigating back to the extension route re-attempt the load, or does it remain in the error state until a full page refresh?

This is a great improvement to extension reliability and user experience. Looking forward to seeing it land!

mateenali66:
thanks @arjunmehta-git. running through the three points based on what the PR actually does:

1. **visible countdown / configurable delay**: fair, right now it's a static "Redirecting to dashboard in a few seconds..." string. i can change it to a live countdown (5, 4, 3, 2, 1) or expose `DASHBOARD_REDIRECT_DELAY_MS` as a prop so calling sites can override. preference?

2. **notification context**: currently `notify({ message: \`Failed to load extension: ${err.toString()}\` })`. for `@paciolan/remote-component` the `err` is typically a `LoadComponentError` whose `.toString()` already includes the URL and the failing HTTP status, but it's not guaranteed. happy to explicitly include `url.url` in the message so an operator can copy-paste it regardless of the underlying error shape.

3. **ErrorBoundary reset**: it's wrapped in the route component (`ui/pages/extension/[...component].tsx`), so navigating away and back remounts the extension subtree via Next.js dynamic routing, which re-attempts the load implicitly. there's no explicit `resetKeys` or `onReset` wired up, so same-route remount (e.g. retry button) would stay in the error state. i can add `resetKeys={[url.url]}` on the ErrorBoundary if in-place recovery is worth covering.

if any of (1), (2), or (3) is something you'd want in this PR vs a follow-up, let me know and i'll push a commit.

PR #18604, opened 2026-04-09
[mesheryctl] Align component view e2e assertions with current CLI output

hssuryawanshib23-afk:
**Notes for Reviewers**

- This PR fixes #18598

**What I understood**

`qa.meshery.io` was failing because the e2e tests for `mesheryctl component view` were still checking older error messages. The CLI now returns report-style invalid-argument output and updated usage text.

**What I changed**

- Updated the `component view` no-argument assertion to match the current `Invalid Argument` output.
- Updated the `component view` multiple-argument assertion to match the current `Invalid Argument` output.
- Updated the invalid `--output-format xml` assertion to match the current flag validation output.
- Added `assert_invalid_view_args()` to avoid repeating the same invalid-argument assertions.
- Made the component view save tests read the actual `Data saved to file:` path from CLI output instead of guessing the filename.
- Updated nearby `component search` e2e assertions to match the current CLI output and avoid non-interactive pagination during repeated fixture imports.

**Validation**

- `go test ./mesheryctl/internal/cli/root/components -run 'TestComponent(View|Search)' -count=1`
- `go test ./mesheryctl/internal/cli/root/model -run TestViewModel -count=1`
- `bats 005-component/03-component-view.bats` passed `7/7`
- `bats 005-component` passed `13/13`


arjunmehta-git:
Great catch, @hssuryawanshib23-afk! Aligning e2e assertions with the current CLI output is exactly the kind of maintenance work that keeps our QA pipeline trustworthy. The `qa.meshery.io` reporting is only as good as the assertions behind it, so fixing stale expectations promptly matters a lot.

A couple of thoughts for this PR:
- The PR description is thorough and well-structured — the "What I understood / What I changed / Validation" format is excellent and I'd encourage keeping that convention.
- Since you couldn't run the full Bats suite locally, consider adding a note in the PR about which CI job serves as the acceptance gate. This helps reviewers know where to look to confirm the change is good end-to-end.
- The Gemini reviewer's suggestion to extract duplicated assertion logic into a shared Bats helper is worth considering — even a small `assert_invalid_argument` helper function would reduce drift if the error message format changes again in the future.

Thank you for keeping our test suite in sync with the CLI behavior. This is exactly the kind of contribution that prevents CI from becoming unreliable.

hssuryawanshib23-afk:
Addressed the maintainability suggestion in a follow-up commit.

I added a small local helper, `assert_invalid_view_args()`, in `mesheryctl/tests/e2e/005-component/03-component-view.bats` and used it for the two duplicated invalid-argument cases. I kept it local to this file to keep the PR scope small.

I also reran the same focused validation after this change:
- `go test ./mesheryctl/internal/cli/root/components -run 'TestComponent(View|Search)' -count=1`
- `go test ./mesheryctl/internal/cli/root/model -run TestViewModel -count=1`
- rebuilt `mesheryctl` locally and rechecked:
  - `component view`
  - `component view comp1 comp2`
  - `component view some-component -o xml`

hssuryawanshib23-afk:
@arjunmehta-git Thanks for the review. I updated the PR description to reflect the follow-up helper refactor.

I also addressed the maintainability point by adding a small local `assert_invalid_view_args()` helper in `mesheryctl/tests/e2e/005-component/03-component-view.bats` and reran the same focused validation after that change.

On the CI point, I could not identify a dedicated Bats/e2e job attached to this PR, so I added a note that full end-to-end confirmation for this change will need to come from the relevant QA pipeline or maintainer verification.

lekaf974:
> I could not run the full Bats suite locally because bats is not installed in this environment.

You ran unit test ok but this is not enough to ensure e2e will pass. 

How can you be sure your change is working as expected if you did not ran bats ?

hssuryawanshib23-afk:
Thank you for pointing this out, @lekaf974. You are right that unit tests alone were not enough to confirm this change.

I have now run the Bats e2e tests locally against a running Meshery instance and updated the PR with the results:

- `bats 005-component/03-component-view.bats` passed `7/7`
- `bats 005-component` passed `13/13`
- `go test ./mesheryctl/internal/cli/root/components -run 'TestComponent(View|Search)' -count=1` passed
- `go test ./mesheryctl/internal/cli/root/model -run TestViewModel -count=1` passed

While running the broader component e2e tests, I also found and fixed nearby stale `component search` assertions so the full `005-component` Bats directory passes now.

mesheryctl/tests/e2e/005-component/02-component-search.bats
```diff
@@ -6,10 +6,11 @@ setup() {
 
   export FIXTURES_DIR="$BATS_TEST_DIRNAME/fixtures/component-search" 
   run $MESHERYCTL_BIN model import -f $FIXTURES_DIR/valid-model
+  assert_success
```
lekaf974, line 9: we don't have to test this, this is a setup function that ensure we have component to run the suite test

mesheryctl/tests/e2e/005-component/02-component-search.bats
```diff
@@ -6,10 +6,11 @@ setup() {
...
 }
 
 @test "Search for a known component" {
```
lekaf974, line 12: Since you are updating the test, could you also take time to update the test name using given when then naming convention like in the view test

lekaf974 (changes requested):
I added 2 small comments, LGTM otherwise

mesheryctl/tests/e2e/005-component/02-component-search.bats
```diff
@@ -18,9 +18,10 @@ setup() {
...
-  assert_output --partial "Error: [search term] isn't specified. Please enter component name to search"
-  assert_output --partial "Usage: mesheryctl exp component search [query-text]"
-  assert_output --partial "Run 'mesheryctl exp component search --help' to see detailed help message"
+  assert_output --partial "Error: Only one argument must be provided"
+  assert_output --partial "Invalid Argument"
+  assert_output --partial "Usage: mesheryctl component search [query-text]"
+  assert_output --partial "Run 'mesheryctl component search --help' to see detailed help message"
 }
 
 @test "Search for non-existent component" {
```
lekaf974: please to use given, when, then for all tests in this file

hssuryawanshib23-afk (commented):
Addressed the review comment in 02-component-search.bats by updating the remaining test names to the given/when/then format for consistency.

Re-validated with the focused component e2e suite: 005-component passed 13/13.

@lekaf974 could you please take another look?

lekaf974 (approved):

PR #18608, opened 2026-04-09
fix(tests): correct mock component model structure

arjunmehta-git:
### Changes

Fixed the mock component model structure in the Design Configurator e2e test to match the expected schema.

**Before:**
```javascript
model: { name: MOCK_MODEL, version: '1.0.0' }
```

**After:**
```javascript
model: { name: MOCK_MODEL, model: { version: '1.0.0' } }
```

This correction ensures the test uses the proper nested model structure expected by the component.


arjunmehta-git:
Closing — the fix was independently applied to master, so this PR has no remaining diff.

PR #18620, opened 2026-04-09
[CI] Add hyperlinks to e2e test report entries

raghavtries:
## Description

This PR enhances the end-to-end test result reporter by adding hyperlinks so reviewers can navigate directly from the PR comment to the relevant run and debugging artifacts.

### What changed
- added a `Links` column to each test entry in the PR comment
- linked each test row to the workflow run, HTML report artifact, and trace artifacts for that CI run
- added direct in-comment anchors for failed tests so reviewers can jump straight to the matching failure details
- updated the PR comment workflow to resolve artifact URLs dynamically for each workflow run
- limited the PR comment workflow to `pull_request_target` runs, since it depends on PR metadata

### Why
Previously, reviewers had to dig through GitHub Actions to find the detailed output for a given test case. This change makes the PR comment itself the entry point into the run logs and debugging artifacts.

### Validation
- verified the workflow artifact names match the current `Meshery UI and Server` workflow uploads
- validated the workflow YAML locally
- validated the reporter JavaScript structure locally
- simulated the final PR comment body with refreshed workflow/artifact links
- verified failed tests can jump directly to their matching failure details section in the PR comment

### Notes
GitHub artifact links are run-level artifact URLs, so the test-specific navigation is provided through direct anchors into the PR comment’s failed test details section, while the run/report/trace links refresh on each CI run.

**Notes for Reviewers**

- This PR fixes #18602 


jamieplu (changes requested):
THanks!

Screenshot or example?

leecalcote:
@raghavtries, thanks for working on this. Please respond to each open feedback comment.

ianrwhitney (changes requested):
Hey thanks for working on this! 🥇 

I have some feedback to consider for you as I am not convinced about the approach. 

The allure report has the logs, trace, history, and the build number the report was generated from. We have some slight indirection for the actual code that was running since the linked build number comes from meshery/qa vs meshery/meshery. So it feels like the links added dont bring much more value in debugging/discvoery but it feels there may be an improvement in the dex ex in finding the PR/commit  that was executing in the test.  Additionally, the current links all points to the workflow run summary which doesnt give you much more info than the allure logs, traces, etc. 

Interested to hear others take on this. Again thanks for contributing. Let me know your thoughts. 🚀

PR #18632, opened 2026-04-10
fix(workspaces): fix workspace view UUID response parsing and update …

Omkar-Ugal:
## Related Issue
Part of #17709

## Description

After PR #18621 was merged, the server started returning a `WorkspacePage` 
object instead of a single `Workspace` for UUID-based workspace view requests. 
This caused `mesheryctl workspace view` to return empty data when searching by UUID.

### Changes

**Fix (`view.go`):**
- Updated UUID-based lookup to parse `WorkspacePage` response instead of single `Workspace`
- Maps first workspace from page result to selected workspace

**Fix (`view.workspace.api.response.golden`):**
- Updated fixture to match new server response format (`WorkspacePage` wrapper)

**E2E Tests (`008-workspace/`):**
- Added valid workspace create test that creates `e2e-test-workspace`
- Fixed list test to use `--count` flag to avoid pagination hang
- Added valid workspace view test reusing workspace from create test (Matthieu's approach)
- Fixed error message assertions to match current CLI validation behavior

## Tests
12 tests, 0 failures


mesheryctl/internal/cli/root/workspaces/view.go
lekaf974, line 1: @Omkar-Ugal this view implementation is really weird, gonna make a PR to refactor it properly then you will have a better starting point for the test

Issue #18654, opened 2026-04-10
[UI] Major Cleanup and Reorganization


hortison:
# Meshery UI Restructure Plan
Meshery has over 3,000 contributors. As one of its many components, Meshery UI has had enjoyed a great number of contributions from different individuals at different times with different objectives over the span of six years. How this component is structured, the frameworks, and libraries it uses, common conventions, Sistent design system, framework upgrades have evolved over its six year lifetime. 

As is our routine, let's take some time to perform cleanup and maintenance. 

- **Branch:** `ui/restructre`
- **Scope:** `meshery/meshery` — `/ui/**` 
- **Goal:** Collapse the sprawl. Enthrone `@sistent/sistent` as the single source of truth for every component, token, color, and theme. Eliminate Material UI's direct use. Break the giant files. Make the directory tree discoverable at a glance.

---

## Summary

Meshery UI is fundamentally sound - 87% TypeScript, RTK Query for data, Sistent already imported in 276 files - but it has accreted three overlapping styling systems, two parallel icon libraries, two parallel workspace concepts, and eight files over 1,000 lines. The restructure is fully and completed described in **[this plan](https://docs.google.com/document/d/1kS1K7LUjGk6-nRbs6q63E7uuqniuLt70UMrhk_Shz8o/edit?usp=sharing)**.

The task is not a rewrite; it is a **disciplined consolidation** across five phases.

| Phase | Theme | Effort | Risk |
| ----- | ----- | ----- | ----- |
| **[Phase 1](https://github.com/meshery/meshery/issues/18656)** Freeze the sprawl | ESLint rules \+ docs that make the target state *enforceable* | Small | None |
| **[Phase 2](https://github.com/meshery/meshery/issues/18657)** Single design system | Eliminate all 100 `@mui/*` imports → `@sistent/sistent` | Medium | Medium — visual regressions |
| **[Phase 3](https://github.com/meshery/meshery/issues/18657)** Single theme | Collapse 3 color files \+ 1,500+ hex literals into `theme.palette.*` | Medium | Low |
| **[Phase 4](https://github.com/meshery/meshery/issues/18659)** Folder reshape | Rename, flatten, and cluster `/components` into 6 domains | Medium | Low (mostly mechanical) |
| **[Phase 5.a.](https://github.com/meshery/meshery/issues/18660)** Break the monoliths | Decompose the 8 giant files | Large | Medium |
[...]

PR #18665, opened 2026-04-11
[UI] Remove MUI icon usage from NameToIcon (leaves phase)

CodeWith-sakib:
This PR addresses a subset of the MUI removal effort by refactoring a single leaf component.

Root cause
`NameToIcon.tsx` relied on `@mui/icons-material`, which violates the project guideline that Sistent (and internal primitives) should be the only UI source.

Fix

* Replaced all `@mui/icons-material` imports with internal icons from `ui/assets/icons`
* Updated icon mappings to use equivalent internal components
* Removed inline styles and hardcoded color usage to align with theme/token rules

Before
Component depended on MUI icons, violating UI guardrails and increasing reliance on deprecated patterns.

After
Component uses only internal icons, aligning with the project’s UI architecture and migration plan.

Scope
This PR intentionally targets a single leaf component (`NameToIcon.tsx`) as part of the incremental migration strategy. Additional components will be addressed in follow-up PRs.

Test Coverage

* Verified the application compiles successfully
* Confirmed no runtime errors introduced
* Ensured component renders without breaking UI

Impact
Contributes to reducing MUI usage and supports the “leaves” phase of the UI migration.

Related issue
Part of #18654 (UI Restructure - MUI removal)

Signed commits


Rajneesh180:
Good target for the MUI elimination effort — `NameToIcon` is a clean leaf component.

The `color` prop was dropped though. The original had `color = '#607D8B'` and forwarded it via `style={{ color }}` to every icon. The PR removes both the prop and the style, so all icons lose their gray tint and render with whatever the internal icon defaults to. Any caller relying on the default color (or passing an explicit color) gets a visual regression. Worth checking what the icons look like before and after — a quick screenshot comparison would help.

The replacement icons should also be checked for semantic fit. `LockIcon` → `CredentialIcon` makes sense for AuthorizationPolicy, but `ExploreIcon` → `SearchIcon` for DestinationRule is a stretch, and `DirectionsCar` → `DeploymentSelectorIcon` for Sidecar loses the original metaphor. These are mapped to Istio resource types in the rjsf form, so they should still communicate something meaningful about each type.

Minor: `CustomIcon` is defined as a component inside the render body, which means React creates a new component type on every render and unmounts/remounts the icon instead of updating props. Since you're already touching this file, consider extracting it outside.

saurabhraghuvanshii:
@CodeWith-sakib Please attach a Loom recording of your changes. It will help us verify them quickly.

saurabhraghuvanshii:
@CodeWith-sakib  Please sign your commit

PR #18673, opened 2026-04-11
feat(dashboard): enforce minimum size constraints for widgets

Rajesh-Nagarajan-11:
This PR prevents users from shrinking dashboard widgets below their default dimensions during edit mode. 

**Changes:**
- Enforced widget constraints by treating the default layout's dimensions (w/h) as the absolute minimum sizes.
- Added logic to apply these constraints live during edit mode, preventing users from resizing below limits.
- Ensured older saved layouts from the database automatically respect these updated constraints.

**Before**

[Screencast from 2026-04-11 20-21-59.webm](https://github.com/user-attachments/assets/543528d9-f02c-4d3a-a6d7-15b557a713e2)

**After**

[Screencast from 2026-04-11 20-25-07.webm](https://github.com/user-attachments/assets/77bd0ce7-49ee-4567-a093-2de82005ea2b)


rishiraj38 (commented):
The implementation looks super clean, and I really like how seamlessly it patches the older dashboard layouts.

I was looking closely at `ui/components/Dashboard/layoutConstraints.ts` and noticed one edge case that might be worth considering:

Currently, the logic uses:` const rawMinW = item.minW ?? widget.defaultSizing.minW`. Because this layout gets saved to user preferences, `minW` will now be permanently persisted in the DB. If we ever decide to increase a widget's default minimum size in a future release, won't this logic just read the user's older, smaller `item.minW `from their DB and ignore the new system default? It might be safer to strictly enforce `widget.defaultSizing `as the absolute minimum floor so stale limits don't get locked in.

rishiraj38 (approved):

rishiraj38:
Looks Good to me.

Rajneesh180 (commented):
The min-size constraint idea is good — prevents widgets from being shrunk into unusable sizes.

`layoutConstraints.ts` is a brand new file and it has `// @ts-nocheck` at the top. There's no reason for this on new code — the function is 25 lines with straightforward types. Adding proper type annotations (the `layouts` param, the return type) would take a few minutes and sets the right precedent.

`2.565` appears as a magic number across all breakpoints and both layout objects. It corresponds to the OVERVIEW widget's default height — worth extracting to a named constant like `OVERVIEW_MIN_HEIGHT` so future readers don't have to reverse-engineer what it means.

There's also a source-of-truth question: `minW`/`minH` are now hardcoded in `defaultLayout.ts` AND defined in `getWidgets.tsx` under `defaultSizing`. The `applyMinSizeConstraints` function was written to derive constraints from the widget definitions programmatically, but the layouts also have them baked in. If both exist, they can drift out of sync. Might be cleaner to keep them only in `getWidgets.tsx` and let `applyMinSizeConstraints` enforce them at runtime, or drop the runtime function and keep only the static values.

Rajesh-Nagarajan-11:
@Rajneesh180 Thanks for the review, everything looks good. I’ve made the updates; please take another look.

Regarding `OVERVIEW_MIN_HEIGHT,` we can move that to a separate PR.

rishiraj38:
@Rajesh-Nagarajan-11, please take a look at this recording. There’s an issue with the latest blog; it’s expanding correctly but not reducing when we try to shrink it. 

https://github.com/user-attachments/assets/d988bed7-b9b8-4507-8111-709d9cd13ee0

rishiraj38 (commented):
The constraints now use defaultLayouts, it doesn't have the sizing rules for optional widgets (like LATEST_BLOGS).
If you add an optional widget and expand it (e.g., to 6 cols wide), its minimum width permanently locks to 6. You can make it bigger, but you can never shrink it back down!
To fix this , we just need to pass the main WIDGETS object (which has the sizing defaults for all widgets) back into the constraints function instead of defaultLayout.

Let me know when that's tweaked and I'll jump back in to test!

Rajesh-Nagarajan-11:
@rishiraj38  Good catch! updated kindly check this again please

rishiraj38 (approved):

rishiraj38:
LGTM!!!

Rajneesh180:
Looks good after the latest round of changes.

The typing on `layoutConstraints.ts` is proper now — interfaces, explicit param/return types, no `@ts-nocheck`. Clean.

The dual constraint application (on save in `updateLayout` AND on render via `constrainedLayouts`) is the right design — saved layouts from the DB could be stale, so you need the render-time pass as a safety net regardless of what was persisted.

The `widgetSizing` fallback for optional widgets not present in `defaultLayouts` handles the edge case rishiraj38 caught. The chain `defaultItem?.w ?? sizing?.w ?? 1` reads well — priority to the breakpoint-specific default, then the widget's own sizing, then a safe floor.

One thing worth knowing (not blocking): `OVERVIEW` in `defaultLayout.ts` has `w: 16` on lg/md while `cols.lg` is 12. The `Math.min(baseMinW, maxCols)` clamp handles this correctly at runtime (effective min = 12), but the raw layout data has a width exceeding the grid — that's pre-existing, not introduced here.

The memoization cleanup (`cols` to module scope, `iconsProps`/`WIDGETS`/`constrainedLayouts` in `useMemo`) addresses the gemini feedback about unnecessary re-renders.

Approving.

Rajneesh180 (approved):

saurabhraghuvanshii (approved):

Issue #18660, opened 2026-04-10
[UI] Restructure Phase 5.a. - Break the Giants


miacycle:
## Prologue
_NOTE: THIS IS A CHILD ISSUE FOR THE EPIC AT #18654_ See [design spec](https://docs.google.com/document/d/1kS1K7LUjGk6-nRbs6q63E7uuqniuLt70UMrhk_Shz8o/edit?usp=sharing).

### Guiding Principles
These are load‑bearing. The entire plan enforces these six rules.

1. Sistent is the only UI kit. No file outside Sistent's internals may import from @mui/*, @material-ui/*, or @rjsf/mui. Period.
2. Tokens, never literals. No hex, rgb, or named color string may appear in any .tsx/.ts file outside ui/theme/ (a single wrapper module - see §4). Colors come from theme.palette.*, spacing from theme.spacing(), breakpoints from theme.breakpoints.*.
3. styled() over style={{}}. Inline style props are reserved for dynamic values that cannot be expressed in a styled component (e.g. transform: translate(x,y) for a draggable). Colors are never dynamic enough to justify inline.
4. One concept, one home. Every domain (Designs, Workspaces, Connections, …) lives in exactly one folder. Cross‑cutting primitives live under ui/components/shared/.
5. Size budget: 400 lines. Any component file > 400 lines is a refactor candidate; > 600 lines is blocked by lint warning; > 1000 lines is a hard error in CI.
6. Discoverable filenames. No index.tsx as the only file in a folder. No generic style.tsx, utils.tsx, helpers.tsx, components.tsx - files are named for what they contain.

## Implementation
1. One PR per giant file (see list below).

## Acceptance Tests
1. Exit criteria: No file > 600 lines outside ui/graphql/ (generated). Modal/Card/Stepper count in app code drops to 1 canonical each. max-lines rule promoted to error at 600.

---

### Giant files (≥1,000 lines)**

| Lines | File |
| ----- | ----- |
| 1,679 | `components/MesheryPatterns/MesheryPatterns.tsx` |
| 1,453 | `components/Dashboard/resources/configuration/config.tsx` |
| 1,376 | `components/MesheryFilters/Filters.tsx` |
| 1,298 | `components/Performance/index.tsx` |
| 1,291 | `components/connections/ConnectionTable.tsx` |
| 1,273 | `components/Dashboard/resources/workloads/config.tsx` |
| 1,230 | `components/MesheryAdapterPlayComponent.tsx` |
[...]

PR #18682, opened 2026-04-12
ui: split UserPreferences/index.tsx into focused modules

Rajneesh180:
Splits `ui/components/UserPreferences/index.tsx` (677 lines) into focused modules under the 400-line budget from the Phase 5a design spec.

**Before → After**

| File | Lines | Owns |
|---|---|---|
| `hooks/useUserPreferences.ts` | 120 | RTK Query hooks, state, mutation handlers (`handleToggle`, `handleCatalogContentToggle`) |
| `GeneralTab.tsx` | 102 | Catalog content switch, analytics toggles, `ThemeToggler` |
| `UserPreferencesPage.tsx` | 82 | Tab shell — composes General + Performance tabs with stable string keys |
| `index.tsx` | 1 | Barrel re-export — preserves the existing import path from `pages/user/preferences.tsx` |

**Changes from original:**
- `RemoteProviderTab` removed per maintainer review (deprecated component)
- Tab indices switched from integers to stable string keys (`"general"`, `"performance"`)
- `handleError` bug fix: original was curried `(msg) => () => { ... }` — the inner function was never invoked by `.catch()` callers, making error handling a silent no-op. Now executes directly.

Relates to #18660


leecalcote (commented):
Overall: The structural goal is solid - the 677-line monolith is properly decomposed and all E2E tests pass. A few issues are worth addressing before merge.

ui/components/UserPreferences/hooks/useUserPreferences.ts
```diff
@@ -0,0 +1,159 @@
...
+      setAnonymousStats(!anonymousStats);
+      handleChange(name, !anonymousStats);
+    } else {
+      setPerfResultStats(!perfResultStats);
+      handleChange(name, !perfResultStats);
+    }
+  };
+
+  const handleError = (msg: string) => {
+    updateProgress({ showProgress: false });
```
leecalcote, line 68: Action is not dispatched, it's a no-op, but should match handleChange at line 694
```
```suggestion
    dispatch(updateProgress({ showProgress: false }));
```

leecalcote: handleChange dispatches correctly; handleError does not. Silent failure on error paths.

ui/components/UserPreferences/GeneralTab.tsx
```diff
@@ -0,0 +1,123 @@
+// @ts-nocheck
```
leecalcote: The @ts-nocheck is suppressing a mismatch.

handleUpdateUserPref calls mutation with wrong shape 

```
// handleChange uses:
updateUserPrefWithContext({ body: requestBody })  // stringified JSON body

// handleUpdateUserPref uses:
updateUserPrefWithContext(updates)  // plain object, wrong shape
```

These two call sites pass different argument structures to the same mutation.

ui/components/UserPreferences/GeneralTab.tsx
```diff
@@ -0,0 +1,123 @@
...
+import React from 'react';
```
leecalcote, line 1: Component defined inside render (GeneralTab.tsx:23)
```
const ThemeToggler = ({ handleUpdateUserPref }) => {
  const Component = ({ mode, toggleTheme }) => { ... };  // ← recreated every render
  return <ThemeTogglerCore Component={Component} />;
};
```
Every render of ThemeToggler creates a new Component reference, causing ThemeTogglerCore to unmount/remount its child. Move Component outside ThemeToggler (take handleUpdateUserPref as a parameter) or memoize with useCallback.

leecalcote (changes requested):
Regarding the @ts-nocheck on all four new files, this refactor is a good opportunity to at least remove these suppressions file-by-file. Carrying them forward from the monolith means the structural split provides no improvement in type safety coverage. Even fixing the two issues above requires TypeScript to catch them going forward.

Let's be rid of this @ts-nocheck, please, @Rajneesh180, and work through any items that it uncovers.

ui/components/UserPreferences/RemoteProviderTab.tsx
```diff
@@ -0,0 +1,346 @@
...
+                  </GridExtensionHeader>
+
+                  <GridExtensionItem
+                    size={{ xs: 6 }}
+                    style={{
+                      borderRadius: '0 0 0 10px',
+                      padding: '20px 20px',
+                    }}
+                  >
+                    <Typography variant="body1">{extension[0].component}</Typography>
```
leecalcote: Both instances of extension[0] are without a guard.

Accessing extension[0].component and .type with no check that the array is non-empty will throw at runtime if an extension entry is empty.

ui/components/UserPreferences/RemoteProviderTab.tsx
```diff
@@ -0,0 +1,346 @@
...
+              providerInfo.capabilities.map((capability, index) => (
+                <React.Fragment key={`${index}-${capability.feature}`}>
+                  <GridCapabilityHeader
+                    size={{ xs: 6 }}
+                    style={{
+                      padding: '20px 20px',
+                      backgroundColor:
+                        userData?.remoteProviderPreferences?.theme === 'dark'
+                          ? index % 2 === 0
+                            ? '#3D4F57'
```
leecalcote: @Rajneesh180 speaking to the other phases...

Hard-coded hex colors for striped rows (RemoteProviderTab.tsx:346–369)
#3D4F57, #293B43, #E7EFF3, #C9DBE3 are duplicated across two grid columns and don't use MUI theme tokens. If the theme palette changes, these will be stale.

ui/components/UserPreferences/RemoteProviderTab.tsx
```diff
@@ -0,0 +1,346 @@
...
+      <ErrorBoundary>
+        <RootContainer>
+          <Typography variant="h5">Provider Information</Typography>
+          <ErrorBoundary>
+            <Grid2 container spacing={2} size="grow">
+              {providerInfo &&
+                Object.entries(providerInfo).map(
+                  ([providerName, provider], index) =>
+                    (index < 2 || index === 3) && (
+                      <Grid2 key={index} size={{ xs: 12, md: 4 }}>
```
leecalcote: @aabidsofi19 what do you think about this area? Is there a better pattern that we should be using?

The array index as `key={index}` is used in a few places. Given our system of keys, does `key={capability.feature}` make more sense? I'm not sure, but I am just wondering whether we have an opportunity to improve here.

Rajneesh180:
@leecalcote addressed all points in d735f4b:

**`handleError` silent no-op** — `updateProgress` is a Redux action creator, was being called without `dispatch()`. Fixed all three call sites in `useUserPreferences.ts` to go through `dispatch(updateProgress({...}))`.

**Mutation shape** — `updateUserPrefWithContext` expects `{ body: ... }`, the hook was passing a plain object. Fixed to `{ body: JSON.stringify(updates) }`.

**ThemeComponent inside render** — extracted to a module-level component outside ThemeToggler. Also realized the `handleUpdateUserPref` prop being passed down was redundant — `ThemeTogglerCore` already handles persistence internally via its own mutation in `toggleTheme()`. Removed the prop entirely instead of fixing its shape.

**`@ts-nocheck`** — removed from all four files (GeneralTab, useUserPreferences, RemoteProviderTab, UserPreferencesPage). Added proper interfaces where needed to satisfy the compiler.

**`extension[0]` guard** — wrapped in a conditional so it only renders when the array is non-empty.

**Hex colors for striped rows** — noted, will address in Phase 2 with the Sistent token migration.

Rajneesh180:
Quick correction on the `dispatch(updateProgress(...))` change — `updateProgress` is actually a self-dispatching wrapper (calls `store.dispatch()` internally at `mesheryUi.ts:126`), not a raw action creator. Adding `dispatch()` around it caused 4 E2E toggle tests to fail. Reverted in 4636ec8.

The other fixes (mutation shape, ThemeComponent extraction, @ts-nocheck removal, extension guard) are all intact.

Rajneesh180:
Docker build and E2E failures are from `meshsync@v0.8.26` — `exec.go:176` has a `TerminalSizeQueue` type mismatch with the current `k8s.io/client-go`. Same failure on the last two master pushes ([run 24312338150](https://github.com/meshery/meshery/actions/runs/24312338150)). Server can't compile → E2E gets `ERR_CONNECTION_REFUSED`. Pre-rebase run had 69/69 E2E passing.

ui/components/UserPreferences/hooks/useUserPreferences.ts
Rajneesh180: @leecalcote this is actually correct as-is. updateProgress at store/slices/mesheryUi.ts:126 is a self-dispatching wrapper — it calls store.dispatch(updateProgressAction(...)) internally. Both handleChange (line 101) and handleError (line 79) call it the same way, without wrapping in dispatch(). Adding dispatch() here would double-dispatch.

ui/components/UserPreferences/GeneralTab.tsx
Rajneesh180: Fixed — the call now wraps with `{ body: JSON.stringify(updates) }` to match `handleChange`'s shape. Both paths go through the same mutation structure now.

The `@ts-nocheck` is a separate concern — removing it properly requires typing the full `handleUpdateUserPref` → `updateUserPrefWithContext` chain, which touches GeneralTab, RemoteProviderTab, and the parent. Planning that for the ts-nocheck removal pass rather than mixing it into this structural split.

ui/components/UserPreferences/GeneralTab.tsx
Rajneesh180: Moved `ThemeComponent` to module scope (line 11 in GeneralTab.tsx) with a typed interface. It receives `handleUpdateUserPref` as a prop now — single definition, no re-creation on render.

ui/components/UserPreferences/RemoteProviderTab.tsx
Rajneesh180: Added `extension[0] &&` guard on both access sites. The array comes from `capabilitiesRegistry` which can legitimately be empty during initial load.

ui/components/UserPreferences/RemoteProviderTab.tsx
Rajneesh180: Agreed — those hex literals are scattered across the striped row logic and definitely need to go through theme tokens. Keeping that out of this PR since it touches the visual layer more than the structure, but it's on the list for the tokens-not-literals pass.

ui/components/UserPreferences/RemoteProviderTab.tsx
Rajneesh180: `capability.feature` is more stable — those strings are unique within a provider's capability set and don't shift between renders. `key={index}` only works safely when the list is static, which it is here, but `feature` communicates intent better and won't break if someone adds sorting later. Can swap it in this PR.

ui/components/UserPreferences/RemoteProviderTab.tsx
```diff
@@ -0,0 +1,368 @@
+import React, { useState } from 'react';
```
aabidsofi19: @Rajneesh180  no need this keep this component .this was deprecated prev and can be removed now

ui/components/UserPreferences/RemoteProviderTab.tsx
aabidsofi19: yes capability keys works better and also we should use stable keys for tabs too , using ints makes it hard to refactor tabs ( like changing the order ,etc)

ui/components/UserPreferences/RemoteProviderTab.tsx
Rajneesh180: Done — removed the entire RemoteProviderTab and cleaned up the hook (dropped providerInfo/providerType/capabilities state and query). Also switched to stable string keys for tabs (`'general'`, `'performance'`) in the same commit.

ui/components/UserPreferences/UserPreferencesPage.tsx
```diff
@@ -0,0 +1,82 @@
+import React, { useState } from 'react';
+import { Tab, Tabs, CustomTooltip, NoSsr, TachometerIcon, useTheme } from '@sistent/sistent';
+// eslint-disable-next-line no-restricted-imports
+import SettingsCellIcon from '@mui/icons-material/SettingsCell';
```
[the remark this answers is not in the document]
Rajneesh180: `SettingsCellIcon` is not exported from `@sistent/sistent` — checked the dist. The suppress is carrying forward a pre-existing import from the original monolith on master (line 40). Migrating it would require either adding it upstream to sistent or creating a local SVG wrapper, both out of scope here.

ui/components/UserPreferences/hooks/useUserPreferences.ts
```diff
@@ -0,0 +1,120 @@
...
+  };
+
+  useEffect(() => {
+    if (isUserDataFetched && userData) {
+      setExtensionPreferences(userData?.usersExtensionPreferences);
+      setCatalogContent(userData?.usersExtensionPreferences?.catalogContent);
+    } else if (isUserDataError) {
+      console.error(userDataError);
+    }
+  }, [isUserDataFetched, userData]);
```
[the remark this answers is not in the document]
Rajneesh180: Fixed — added `isUserDataError` and `userDataError` to the dep array.

ui/components/UserPreferences/hooks/useUserPreferences.ts
```diff
@@ -0,0 +1,120 @@
...
+      setPerfResultStats(!perfResultStats);
+      handleChange(name, !perfResultStats);
+    }
+  };
+
+  const handleError = (msg: string) => {
+    updateProgress({ showProgress: false });
+    notify({ message: msg, event_type: EVENT_TYPES.ERROR });
+  };
+
```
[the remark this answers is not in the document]
Rajneesh180: This is actually a bug fix. The original on master defines handleError as curried: (name) => () => { ... }. The call sites do .catch(() => { handleError(msg); }) which evaluates the outer function and discards the returned inner one — updateProgress and notify never fire. The non-curried version here is the intended behavior. Updated the PR description to document this.

ui/components/UserPreferences/UserPreferencesPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <GeneralTab
+              catalogContent={catalogContent}
+              anonymousStats={anonymousStats}
+              perfResultStats={perfResultStats}
+              handleCatalogContentToggle={handleCatalogContentToggle}
+              handleToggle={handleToggle}
+            />
+          )}
+          {tabVal === 'performance' && <MesherySettingsPerformanceComponent />}
+        </StatsWrapper>
```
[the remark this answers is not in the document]
Rajneesh180: Removed per maintainer directive — deprecated component. PR description updated to reflect this and the other intentional changes.

ui/components/UserPreferences/UserPreferencesPage.tsx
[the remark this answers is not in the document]
Rajneesh180: SettingsCellIcon is not exported from @sistent/sistent — checked the package dist. This carries forward a pre-existing import from the original monolith on master (line 40). Migrating requires either adding it upstream to sistent or creating a local SVG, both out of scope for this structural split.

Rajneesh180:
@leecalcote @aabidsofi19 all review feedback addressed in the latest two commits:

- **RemoteProviderTab removed** — deprecated per review. Cleaned up the hook too (dropped providerInfo, providerType, capabilities query, dead handleUpdateUserPref).
- **Stable tab keys** — switched from integer indices to "general" / "performance" string values.
- **useEffect deps** — added missing isUserDataError / userDataError to the dependency array.
- **handleError bug fix** — original curried signature meant the inner function was silently discarded by .catch() callers. Now executes directly.

ui/components/UserPreferences/UserPreferencesPage.tsx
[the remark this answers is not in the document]
alexquincy: @Rajneesh180, PRs like this are difficult to review. There are hundreds and hundreds of lined of changes included that are immaterial to the PR itself. Factor in only include buying changes that are material to the issue that you're working on.

alexquincy (changes requested):

Rajneesh180:
@alexquincy This is a structural split — the line count is the refactoring itself (moving 677 lines from one file into focused modules). The SettingsCellIcon import is carrying forward what's already on master line 40, not a new addition. Since the last push, RemoteProviderTab was also removed per aabidsofi19's direction, so the diff is tighter now.

ui/components/UserPreferences/RemoteProviderTab.tsx
```diff
@@ -0,0 +1,353 @@
...
+                </>
+              ))}
+          </Grid2>
+          <Divider />
+          <Typography variant="h5" style={{ margin: '20px 0' }}>
+            Extensions
+          </Typography>
+          {providerInfo.extensions &&
+            Object.entries(providerInfo.extensions).map(([extensionName, extension], index) => (
+              <div key={index} margin="20px 0px">
```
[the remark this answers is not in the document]
pontusringblom: @Rajneesh180 respond to each and every code review comment. Incorporate changes or reject.

PR #18767, opened 2026-04-16
Fix dashboard grid resize jitter with debounced width updates

### LINKED ISSUE #18766: [UI] Dashboard flickering issue
**Steps to reproduce this bug**

- Open the dashboard.
- Right-click and select Inspect (open DevTools).
- Enable responsive mode and set the screen size to 1600 × 800.
- Close the DevTools (or exit responsive mode).
- Observe the dashboard — it starts jittering / flickering.

https://github.com/user-attachments/assets/5b81bc8a-aeae-408a-8f87-e690687a4caf

---

Rajesh-Nagarajan-11:
**PR Description**
This PR fixes #18766
This PR fixes dashboard grid jitter during fast window resize by debouncing width updates (100ms) before recalculating layout. This reduces unnecessary re-renders and keeps the layout stable.

**What was happening**

The dashboard grid used react-grid-layout legacy WidthProvider, which updated width continuously during resize. That caused frequent re-layout/repaint cycles, leading to jitter/fluctuation and occasional broken rendering.


utafrali (commented):
The debounce + RAF approach correctly reduces resize jitter and cleanup is mostly solid, but there is a real race condition where a RAF can fire after unmount in a narrow timing window, and the hardcoded 1280px initial width will cause a layout shift on first mount. Those two issues should be fixed before merging; the rest are low-priority polish items.

ui/components/Dashboard/debounceWidthProvider.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const debounceWidthProvider = <P extends object>(
+  ComposedComponent: React.ComponentType<
+    P & { width: number; innerRef?: React.Ref<HTMLDivElement> }
+  >,
+) => {
+  const Wrapper = (props: Omit<P, 'width'> & DebounceWidthProviderProps) => {
+    const { measureBeforeMount = false, className, style, debounceTimeout = 0, ...rest } = props;
+
+    const [width, setWidth] = useState(DEFAULT_WIDTH);
```
utafrali, line 21: The default width is hardcoded to `1280` instead of measuring the actual container width on mount. `ResponsiveHoneycomb.tsx` in this same project measures the real initial width immediately. With a hardcoded default, users on narrower viewports will see a layout shift the first time the ResizeObserver fires.

Fix: read `elementRef.current?.getBoundingClientRect().width` inside the mount effect and set it as the initial state.

ui/components/Dashboard/debounceWidthProvider.tsx
```diff
@@ -0,0 +1,91 @@
...
+            debounceTimerRef.current = null;
+          }, debounceTimeout);
+        } else {
+          setWidth((w) => (w === newWidth ? w : newWidth));
+        }
+      };
+
+      const observer = new ResizeObserver((entries) => {
+        if (!entries[0]) return;
+        if (rafIdRef.current) cancelAnimationFrame(rafIdRef.current);
```
utafrali, line 48: There is a race condition between the debounce timeout and component unmount. When the timeout fires it sets `debounceTimerRef.current = null`, then schedules a new RAF. If the component unmounts in that narrow window, the useEffect cleanup already ran (and saw `rafIdRef.current` as `null`), so the RAF is never cancelled and `setWidth` is called on an unmounted component, producing a React warning.

Fix: track a `cancelled` flag in the closure or use a ref:
```tsx
const cancelledRef = useRef(false);
// in cleanup:
cancelledRef.current = true;
// in RAF callback:
if (!cancelledRef.current) setWidth(newWidth);
```
Reset `cancelledRef.current = false` at the top of the effect.

ui/components/Dashboard/debounceWidthProvider.tsx
```diff
@@ -0,0 +1,91 @@
...
+          debounceTimerRef.current = window.setTimeout(() => {
+            setWidth((w) => (w === newWidth ? w : newWidth));
+            debounceTimerRef.current = null;
+          }, debounceTimeout);
+        } else {
+          setWidth((w) => (w === newWidth ? w : newWidth));
+        }
+      };
+
+      const observer = new ResizeObserver((entries) => {
```
utafrali, line 46: `ResponsiveHoneycomb.tsx` (the other resize-aware component in this project) imports `resize-observer-polyfill` to support older browsers. This file uses the native `ResizeObserver` directly. If the project still targets browsers without native support, this will silently skip the observer and the grid will never get a width update after the initial render.

Either add the same polyfill import or add a guard:
```tsx
if (typeof ResizeObserver === 'undefined') return;
```

ui/components/Dashboard/debounceWidthProvider.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const [width, setWidth] = useState(DEFAULT_WIDTH);
+    const [mounted, setMounted] = useState(false);
+
+    const elementRef = useRef<HTMLDivElement>(null);
+    const rafIdRef = useRef<number | null>(null);
+    const debounceTimerRef = useRef<number | null>(null);
+
+    useEffect(() => setMounted(true), []);
+
+    useEffect(() => {
```
utafrali, line 30: The project already has `ui/utils/debounce.ts` and `ui/utils/hooks/useDebounce.ts`. The custom debounce logic here is ~15 lines and slightly hard to follow because the timeout and RAF are nested in an uncommon order (timeout outer, RAF inner). Reusing the existing `debounce` utility for the timeout portion would keep this file shorter and consistent with the rest of the codebase:
```tsx
const debouncedUpdate = useRef(debounce((w: number) => setWidth(w), debounceTimeout));
```
Then cancel it in the cleanup with `debouncedUpdate.current.cancel()` (if the utility exposes cancel) or replace it with a ref-based approach.

ui/components/Dashboard/index.tsx
```diff
@@ -384,6 +385,7 @@ const Dashboard = () => {
               </Stack>
 
               <ResponsiveReactGridLayout
+                debounceTimeout={100}
```
utafrali, line 388: The `debounceTimeout={100}` value is hardcoded at the call site. If this component is used in multiple places (or the timeout needs to change), callers have to remember to pass it each time. Since `100` is the intended default and is already set in the interface, you can drop this prop here and rely on the default, keeping the JSX cleaner.

PR #18769, opened 2026-04-16
Fix dashboard grid resize jitter with debounced width updates

### LINKED ISSUE #18768: [UI] Dashboard flickering issue
**Steps to reproduce this bug**

- Open the dashboard.
- Right-click and select Inspect (open DevTools).
- Enable responsive mode and set the screen size to 1200 × 800.
- Close the DevTools (or exit responsive mode).
- Observe the dashboard — it starts jittering / flickering.

https://github.com/user-attachments/assets/5b81bc8a-aeae-408a-8f87-e690687a4caf

---

Rajesh-Nagarajan-11:
**PR Description**
This PR fixes #18768
This PR fixes dashboard grid jitter during fast window resize by debouncing width updates (100ms) before recalculating layout. This reduces unnecessary re-renders and keeps the layout stable.

**What was happening**

The dashboard grid used react-grid-layout legacy WidthProvider, which updated width continuously during resize. That caused frequent re-layout/repaint cycles, leading to jitter/fluctuation and occasional broken rendering.


ui/components/Dashboard/debounceWidthProvider.tsx
```diff
@@ -0,0 +1,79 @@
...
+  className?: string;
+  style?: React.CSSProperties;
+  debounceTimeout?: number;
+}
+
+const debounceWidthProvider = <P extends object>(
+  ComposedComponent: React.ComponentType<P & { width: number }>,
+) => {
+  const Wrapper = (props: Omit<P, 'width'> & DebounceWidthProviderProps) => {
+    const { measureBeforeMount = false, className, style, debounceTimeout = 0, ...rest } = props;
```
utafrali, line 17: The default `debounceTimeout = 0` means the HOC performs no debouncing unless the caller explicitly opts in. Given the component is named `debounceWidthProvider` and the whole point is to reduce jitter, a zero default is misleading. Any future consumer that forgets to pass `debounceTimeout` gets only RAF batching, not the debounce behavior the name promises. Consider defaulting to `100` here to match the current call site and make the safe behavior the default:

```tsx
const { measureBeforeMount = false, className, style, debounceTimeout = 100, ...rest } = props;
```

ui/components/Dashboard/debounceWidthProvider.tsx
```diff
@@ -0,0 +1,79 @@
...
+        } else {
+          setWidth((w) => (w === newWidth ? w : newWidth));
+        }
+      };
+
+      const observer = new ResizeObserver((entries) => {
+        if (!entries[0]) return;
+        if (rafIdRef.current) cancelAnimationFrame(rafIdRef.current);
+        rafIdRef.current = requestAnimationFrame(() => {
+          scheduleWidthUpdate(Math.floor(entries[0].contentRect.width));
```
utafrali, line 48: `entries[0]` is read from a closure captured when the RAF was scheduled, not when it fires. If a new resize event fires before the RAF executes, the old RAF is cancelled and a new one is scheduled with updated `entries`. This is correct. However, it's worth reading `entries[0].contentRect.width` before scheduling the RAF to make the intent explicit and avoid any confusion about which snapshot is being used:

```tsx
const observer = new ResizeObserver((entries) => {
  if (!entries[0]) return;
  const newWidth = Math.floor(entries[0].contentRect.width);
  if (rafIdRef.current) cancelAnimationFrame(rafIdRef.current);
  rafIdRef.current = requestAnimationFrame(() => {
    scheduleWidthUpdate(newWidth);
    rafIdRef.current = null;
  });
});
```

ui/components/Dashboard/debounceWidthProvider.tsx
```diff
@@ -0,0 +1,79 @@
...
+      };
+    }, [debounceTimeout]);
+
+    return (
+      <div
+        className={[className, layoutClassName].filter(Boolean).join(' ')}
+        style={style}
+        ref={elementRef}
+      >
+        {(!measureBeforeMount || mounted) && <ComposedComponent {...(rest as P)} width={width} />}
```
utafrali, line 68: The `rest as P` cast silences TypeScript because the type system can't automatically narrow `Omit<Omit<P, 'width'> & DebounceWidthProviderProps, keyof DebounceWidthProviderProps>` back to `P`. This is a common HOC limitation, but it means TypeScript won't catch if `DebounceWidthProviderProps` accidentally shadows a required prop of `P` (e.g., `className` or `style` from `Responsive` are consumed by the wrapper and never forwarded). This matches the upstream `WidthProvider` behavior, so it's acceptable, but worth a comment explaining that `className`, `style`, `measureBeforeMount`, and `debounceTimeout` are intentionally consumed by the wrapper and not forwarded to `ComposedComponent`.

ui/components/Dashboard/debounceWidthProvider.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useEffect, useRef, useState } from 'react';
```
utafrali, line 1: No tests are included for this utility. Given it's a pure HOC with distinct code paths (debounce vs. no-debounce, measureBeforeMount, cleanup), a small test file covering the RAF/debounce branching and cleanup teardown would prevent regressions. At minimum, verifying that `setWidth` is called with the correct value after a simulated `ResizeObserver` callback would be valuable.

utafrali (commented):
The implementation is technically sound: it replaces the legacy window-resize listener approach with ResizeObserver + RAF batching, which is a genuine improvement. The main concern is that `debounceTimeout` defaults to 0, meaning the HOC provides no debouncing unless callers know to pass a value -- a mismatch with the component name that will silently affect any future consumers.

ui/components/Dashboard/debounceWidthProvider.tsx
```diff
@@ -0,0 +1,79 @@
...
+      const observer = new ResizeObserver((entries) => {
+        if (!entries[0]) return;
+        if (rafIdRef.current) cancelAnimationFrame(rafIdRef.current);
+        rafIdRef.current = requestAnimationFrame(() => {
+          scheduleWidthUpdate(Math.floor(entries[0].contentRect.width));
+          rafIdRef.current = null;
+        });
+      });
+
+      observer.observe(node);
```
banana-three-join, line 53: There's currently a race condition when setMounted is true.

The observer is set on the current empty div of the ComposedComponent in the initial render, setMounted = true causes a re-render but React is faster than the DOM. This means that ComposedComponent is rendered faster with the default width than the observer calls its callback   function which causes an update of the width which would cause a quick snapback.

banana-three-join (commented):
Aside from the feedback in the code, I think a video with the new behavior with the new behavior would also be good to add!

PR #18776, opened 2026-04-16
[Server] Prevent nil-pointer dereference panic in connection status g…

### LINKED ISSUE #18775: [Bug] Nil-pointer dereference panic in connection status change goroutine crashes server
## Description
In `NotifySmOfConnectionStatusChange` (`server/handlers/connections_handlers.go`), a goroutine calls `inst.SendEvent(...)` which returns `(*events.Event, error)`. On the error path, `*event` is dereferenced without checking for nil, causing a fatal panic inside the goroutine.

Since Go goroutines have no implicit `recover()`, this nil-pointer dereference terminates the **entire Meshery Server process** — not just the individual request or connection.

## Root Cause
`SendEvent` (`server/machines/models.go`) has three return paths where `event` can be nil when an error occurs:

1. **Line 167** — `action.ExecuteOnExit` returns `(nil, err)` → `return event, err`
2. **Line 180** — `action.ExecuteOnEntry` returns `(NoOp, nil, err)` → `return event, err`  
3. **Line 190** — `action.Execute` returns `(NoOp, nil, err)` → `return event, err`

The caller at line 520 does:
```go
_ = provider.PersistEvent(*event, token)  // PANIC: *event when event is nil
```

`PersistEvent` accepts `events.Event` (value type), so `*event` is a hard dereference that immediately panics on nil.

## Steps to Reproduce
1. Configure a Kubernetes connection with an intentionally faulty configuration (e.g., expired certificate, unreachable cluster).
2. Trigger a connection status change via the UI or API.
3. If `SendEvent` encounters a state transition error where the action function returns `(nil, error)`, the goroutine panics.
4. The entire Meshery Server process terminates, disconnecting all users.

## Expected Behavior
The goroutine should safely skip event persistence when `event` is nil, log the error, and return without crashing.

## Environment
- **Component:** Meshery Server (Go Backend)
- **File:** `server/handlers/connections_handlers.go`
- **Version:** `master`
- **Priority:** P1 — Fatal server crash

KumarADITHYA123:
## Description
This PR fixes a **P1 server crash** caused by a nil-pointer dereference inside the connection status change goroutine in `NotifySmOfConnectionStatusChange`.

**The Bug:**  
`inst.SendEvent(...)` returns `(*events.Event, error)`. The `SendEvent` method has three internal return paths (`models.go` lines 167, 180, 190) where it returns `(nil, error)`. In the error handler, the code unconditionally dereferences `*event` to pass to `provider.PersistEvent()` (which accepts `events.Event` value type). When `event` is nil, this causes a fatal panic inside a goroutine with no `recover()`, terminating the entire Meshery Server process.

**The Fix:**  
Added `if event != nil` nil guards around `*event` dereferences on both the error path and success path inside the goroutine. When `event` is nil, the persistence and broadcast are safely skipped while the error is still logged.

## Testing / Verification
- Added `connections_handlers_test.go` → `TestPersistEventNilGuard`
- Table-driven test covering both nil and non-nil event scenarios
- Uses `defer recover()` to explicitly catch any panic regression
- `go test -v -run TestPersistEventNilGuard ./server/handlers/` — **PASS** (0.00s)
- `go build ./server/handlers/...` — **PASS** (zero errors)
- `go vet ./server/handlers/...` — **PASS** (zero warnings)

**Fixes #18775

### DIFF
--- a/server/handlers/connections_handlers.go
+++ b/server/handlers/connections_handlers.go
@@ -518,8 +518,12 @@ func (h *Handler) NotifySmOfConnectionStatusChange(context context.Context, user
 			if err != nil {
 				h.log.Error(err)
 				if event != nil {
-					_ = provider.PersistEvent(*event, token)
+					if persistErr := provider.PersistEvent(*event, token); persistErr != nil {
+						h.log.Warn(persistErr)
+					}
 					h.config.EventBroadcaster.Publish(userID, event)
+				} else {
+					h.log.Warn(fmt.Errorf("connection status change event was nil for connection %s, skipping persistence", inst.ID))
 				}
 				return
 			}
@@ -529,8 +533,12 @@ func (h *Handler) NotifySmOfConnectionStatusChange(context context.Context, user
 			}
 
 			if event != nil {
-				_ = provider.PersistEvent(*event, token)
+				if persistErr := provider.PersistEvent(*event, token); persistErr != nil {
+					h.log.Warn(persistErr)
+				}
 				h.config.EventBroadcaster.Publish(userID, event)
+			} else {
+				h.log.Warn(fmt.Errorf("connection status change event was nil for connection %s, skipping persistence", inst.ID))
 			}
 		}(inst, connection.Status)
 	}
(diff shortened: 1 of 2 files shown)


KumarADITHYA123:
Hi @leecalcote, @pontusringblom, and @aabidsofi19 

I wanted to flag this PR for review whenever you have a moment. It fixes a critical P1 server panic inside the connections_handlers.go goroutine that occurs when SendEvent returns a nil pointer during a failed state transition.

I've added a  regression test to verify the fix, and it's passing cleanly locally. Would appreciate your feedback!

PR #18777, opened 2026-04-16
[Server] Fix nil-pointer dereference panic in SendEvent goroutines

### LINKED ISSUE #18775: [Bug] Nil-pointer dereference panic in connection status change goroutine crashes server

thc1006:
## Summary

- Fix a nil-pointer dereference in the goroutines that call `SendEvent`. `SendEvent` can return `(nil, err)`, and the callers dereferenced `*event` to pass it to `PersistEvent` without a nil check, which crashes the entire Meshery Server process.
- On the error path the goroutines log the error, persist and publish the event when `SendEvent` still returned one, and then return. On the success path the event is persisted and published, and a nil event is logged.
- Guard against a nil state machine. `InitializeMachineWithContext` returns `(nil, err)` on failure, and the two Kubernetes middlewares and `addK8SConfig` kept going and called `ResetState`/`SendEvent` on the nil machine. They now skip the failed context, and `InitializeMachineWithContext` no longer caches a machine whose `Start` failed.
- Touches `connections_handlers.go`, `middlewares.go`, `k8sconfig_handler.go` and `machines/helpers/helpers.go`.

Fixes #18775

## Root Cause

`SendEvent` (`server/machines/models.go`) has return paths where the action callbacks (`ExecuteOnExit`, `ExecuteOnEntry`, `Execute`) return `(nil, err)`. The callers ran inside goroutines and dereferenced `*event` to pass it to `PersistEvent`, which takes a value type, so a nil event caused a fatal nil-pointer panic. Go goroutines have no implicit `recover()`, so this took down the whole server.

`SendEvent` can also return a non-nil event together with an error, for example when the connection lookup or status update fails. The earlier revision returned early on any error and dropped that event, so the failure was only logged on the server and never reached the user.

The same handlers also used the state machine returned by `InitializeMachineWithContext` without checking it. That value can be nil when initialization fails, for example with a faulty kubeconfig, which is the reproduction scenario in #18775.

## Test plan

- [x] `go build ./server/...` passes
- [x] `go vet ./server/handlers/` passes
- [x] `go test ./server/handlers/` passes
- [ ] Manually verify: configure a Kubernetes connection with an expired certificate, trigger a status change. The server should log the error and keep running instead of crashing.

### DIFF
--- a/server/handlers/connections_handlers.go
+++ b/server/handlers/connections_handlers.go
@@ -527,6 +527,8 @@ func (h *Handler) NotifySmOfConnectionStatusChange(context context.Context, user
 			if event != nil {
 				_ = provider.PersistEvent(*event, token)
 				h.config.EventBroadcaster.Publish(userID, event)
+			} else {
+				h.log.Warnf("SendEvent returned a nil event without an error for connection %s on status %q; skipping event persistence", inst.ID, status)
 			}
 		}(inst, connection.Status)
 	}
--- a/server/handlers/connections_handlers.go
+++ b/server/handlers/connections_handlers.go
@@ -517,6 +517,10 @@ func (h *Handler) NotifySmOfConnectionStatusChange(context context.Context, user
 			event, err := inst.SendEvent(context, machines.EventType(helpers.StatusToEvent(status)), nil)
 			if err != nil {
 				h.log.Error(err)
+				if event != nil {
+					_ = provider.PersistEvent(*event, token)
+					h.config.EventBroadcaster.Publish(userID, event)
+				}
 				return
 			}
 
--- a/server/handlers/k8sconfig_handler.go
+++ b/server/handlers/k8sconfig_handler.go
@@ -203,6 +203,8 @@ func (h *Handler) addK8SConfig(user *models.User, _ *models.Preference, w http.R
 				if event != nil {
 					_ = provider.PersistEvent(*event, token)
 					go h.config.EventBroadcaster.Publish(userID, event)
+				} else {
+					h.log.Warnf("SendEvent returned a nil event without an error for connection %s; skipping event persistence", inst.ID)
 				}
 			}(inst)
 		}
--- a/server/handlers/k8sconfig_handler.go
+++ b/server/handlers/k8sconfig_handler.go
@@ -199,6 +199,10 @@ func (h *Handler) addK8SConfig(user *models.User, _ *models.Preference, w http.R
 					event, err := inst.SendEvent(req.Context(), machines.EventType(mhelpers.StatusToEvent(status)), nil)
 					if err != nil {
 						h.log.Error(err)
+						if event != nil {
+							_ = provider.PersistEvent(*event, token)
+							go h.config.EventBroadcaster.Publish(userID, event)
+						}
 						return
 					}
 					if event != nil {
--- a/server/handlers/k8sconfig_handler.go
+++ b/server/handlers/k8sconfig_handler.go
@@ -194,19 +194,21 @@ func (h *Handler) addK8SConfig(user *models.User, _ *models.Preference, w http.R
 				h.log.Error(err)
 			}
 
-			go func(inst *machines.StateMachine) {
-				event, err := inst.SendEvent(req.Context(), machines.EventType(mhelpers.StatusToEvent(status)), nil)
-				if err != nil {
-					h.log.Error(err)
-					return
-				}
-				if event != nil {
-					_ = provider.PersistEvent(*event, token)
-					go h.config.EventBroadcaster.Publish(userID, event)
-				} else {
-					h.log.Warnf("SendEvent returned a nil event without an error for connection %s; skipping event persistence", inst.ID)
-				}
-			}(inst)
+			if inst != nil {
+				go func(inst *machines.StateMachine) {
+					event, err := inst.SendEvent(req.Context(), machines.EventType(mhelpers.StatusToEvent(status)), nil)
+					if err != nil {
+						h.log.Error(err)
+						return
+					}
+					if event != nil {
+						_ = provider.PersistEvent(*event, token)
+						go h.config.EventBroadcaster.Publish(userID, event)
+					} else {
+						h.log.Warnf("SendEvent returned a nil event without an error for connection %s; skipping event persistence", inst.ID)
+					}
+				}(inst)
+			}
 		}
 
 		eventMetadata[ctx.Name] = metadata
--- a/server/handlers/middlewares.go
+++ b/server/handlers/middlewares.go
@@ -323,6 +323,8 @@ func KubernetesMiddleware(ctx context.Context, h *Handler, provider models.Provi
 			if event != nil {
 				_ = provider.PersistEvent(*event, token)
 				go h.config.EventBroadcaster.Publish(userUUID, event)
+			} else {
+				h.log.Warnf("SendEvent returned a nil event without an error for connection %s; skipping event persistence", inst.ID)
 			}
 		}(inst)
 	}
@@ -386,6 +388,8 @@ func K8sFSMMiddleware(ctx context.Context, h *Handler, provider models.Provider,
 			if event != nil {
 				_ = provider.PersistEvent(*event, token)
 				go h.config.EventBroadcaster.Publish(userUUID, event)
+			} else {
+				h.log.Warnf("SendEvent returned a nil event without an error for connection %s; skipping event persistence", inst.ID)
 			}
 		}(inst)
 		kubernesMachineCtx, err := utils.Cast[*kubernetes.MachineCtx](inst.Context)
--- a/server/handlers/middlewares.go
+++ b/server/handlers/middlewares.go
@@ -319,6 +319,10 @@ func KubernetesMiddleware(ctx context.Context, h *Handler, provider models.Provi
 			event, err := inst.SendEvent(ctx, machines.Discovery, nil)
 			if err != nil {
 				h.log.Error(err)
+				if event != nil {
+					_ = provider.PersistEvent(*event, token)
+					go h.config.EventBroadcaster.Publish(userUUID, event)
+				}
 				return
 			}
 			if event != nil {
@@ -385,6 +389,10 @@ func K8sFSMMiddleware(ctx context.Context, h *Handler, provider models.Provider,
 			event, err := inst.SendEvent(ctx, machines.Discovery, nil)
 			if err != nil {
 				h.log.Error(err)
+				if event != nil {
+					_ = provider.PersistEvent(*event, token)
+					go h.config.EventBroadcaster.Publish(userUUID, event)
+				}
 				return
 			}
 			if event != nil {
--- a/server/handlers/middlewares.go
+++ b/server/handlers/middlewares.go
@@ -311,6 +311,7 @@ func KubernetesMiddleware(ctx context.Context, h *Handler, provider models.Provi
 		)
 		if err != nil {
 			h.log.Error(err)
+			continue
 		}
 
 		inst.ResetState()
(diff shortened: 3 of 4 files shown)


thc1006:
The two failing tests are both `evaluation idempotency` in `relationship_evaluation_properties.spec.js`, which tests the design relationship evaluation engine for Kubernetes resource matchlabels and sibling relationships. This is a completely different code path from the connection state machine and K8s discovery event handling that this PR touches. Additionally, the E2E suite ran against commit ea3348d08c3 which only contains the nil guard changes and does not include the later commit that adjusts event persistence on the success path. The same `evaluation idempotency` failure also appears on recently merged PR #18764, which is a pure UI theme change, confirming this is a pre existing flaky test on master and not caused by this PR.

server/handlers/middlewares.go
```diff
@@ -374,10 +382,18 @@ func K8sFSMMiddleware(ctx context.Context, h *Handler, provider models.Provider,
...
+
 			event, err := inst.SendEvent(ctx, machines.Discovery, nil)
 			if err != nil {
-				_ = provider.PersistEvent(*event, token)
-				go h.config.EventBroadcaster.Publish(userUUID, event)
+				if event != nil {
+					_ = provider.PersistEvent(*event, token)
+					go h.config.EventBroadcaster.Publish(userUUID, event)
+				}
 			}
```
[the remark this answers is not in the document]
fitzergerald: @thc1006 when you get a moment, please incorporate or reject this feedback (each of the code review comments).

leecalcote (commented):
Thanks, @thc1006!

leecalcote:
> The two failing tests are both `evaluation idempotency` in `relationship_evaluation_properties.spec.js`, which tests the design relationship evaluation engine for Kubernetes resource matchlabels and sibling relationships. This is a completely different code path from the connection state machine and K8s discovery event handling that this PR touches. Additionally, the E2E suite ran against commit [ea3348d](https://github.com/meshery/meshery/commit/ea3348d08c36d072cb1585c73ca32eee1f1a0fac) which only contains the nil guard changes and does not include the later commit that adjusts event persistence on the success path. The same `evaluation idempotency` failure also appears on recently merged PR #18764, which is a pure UI theme change, confirming this is a pre existing flaky test on master and not caused by this PR.

@aabidsofi19 👀

server/handlers/connections_handlers.go
```diff
@@ -514,20 +514,30 @@ func (h *Handler) NotifySmOfConnectionStatusChange(context context.Context, user
 		}
 
 		go func(inst *machines.StateMachine, status connections.ConnectionStatus) {
+			defer func() {
```
Amr-Shams: there is almost no panic in the func scope, can you expalin why would we need this ?

server/handlers/connections_handlers.go
```diff
@@ -514,20 +514,30 @@ func (h *Handler) NotifySmOfConnectionStatusChange(context context.Context, user
...
+					h.log.Error(ErrPanicRecovery(r))
+				}
+			}()
+
 			event, err := inst.SendEvent(context, machines.EventType(helpers.StatusToEvent(status)), nil)
 			if err != nil {
 				h.log.Error(err)
-				_ = provider.PersistEvent(*event, token)
-				h.config.EventBroadcaster.Publish(userID, event)
+				if event != nil {
```
Amr-Shams, line 520: i don't think a safegurad here needed, the event if nil would return an error

server/handlers/middlewares.go
```diff
@@ -374,8 +383,17 @@ func K8sFSMMiddleware(ctx context.Context, h *Handler, provider models.Provider,
 
 		inst.ResetState()
 		go func(inst *machines.StateMachine) {
+			defer func() {
```
Amr-Shams: the same

server/handlers/connections_handlers.go
thc1006: Good point, dropped the recover. With the nil check in place there is no realistic panic source left, so the wrapper was just noise.

server/handlers/connections_handlers.go
thc1006: Yeah you are right, removed it. Switched to early return on error so we never touch event when SendEvent failed.

server/handlers/middlewares.go
thc1006: Same treatment here, recover gone and the error path returns early. Pushed in 9e63508.

thc1006:
The mesheryctl Unit tests failure looks unrelated, the same job has been failing on the last four master runs (24920822176, 24919658546, 24910163067, 24901068405) and this PR only touches files under server/handlers.

server/handlers/connections_handlers.go
```diff
@@ -517,17 +517,17 @@ func (h *Handler) NotifySmOfConnectionStatusChange(context context.Context, user
...
 				return
 			}
 
 			if status == connections.DELETED {
 				smInstanceTracker.Remove(inst.ID)
 			}
 
-			_ = provider.PersistEvent(*event, token)
-			h.config.EventBroadcaster.Publish(userID, event)
+			if event != nil {
```
Amr-Shams, line 531: could u add logger for the event is nil, cause it is important here

server/handlers/k8sconfig_handler.go
```diff
@@ -197,6 +197,10 @@ func (h *Handler) addK8SConfig(user *models.User, _ *models.Preference, w http.R
 			go func(inst *machines.StateMachine) {
 				event, err := inst.SendEvent(req.Context(), machines.EventType(mhelpers.StatusToEvent(status)), nil)
 				if err != nil {
+					h.log.Error(err)
+					return
+				}
+				if event != nil {
```
Amr-Shams: same ehre

Amr-Shams (commented):
just final touches

server/handlers/connections_handlers.go
thc1006: Added. The else branch now logs a Warnf with the connection ID and the status when SendEvent returns a nil event without an error, right before we skip persistence.

server/handlers/k8sconfig_handler.go
thc1006: Done here too, same Warnf with the connection ID when SendEvent hands back a nil event.

thc1006:
Pushed two follow-ups:

- f295d3f logs a warning when SendEvent returns a nil event without an error, addressing the review comments above.
- 5b7e565 fixes a related crash in the same goroutines. InitializeMachineWithContext returns (nil, err) when it fails to build or start the state machine (for instance with a faulty kubeconfig, which is the repro scenario in #18775). KubernetesMiddleware, K8sFSMMiddleware and addK8SConfig logged that error but kept going, then called ResetState, SendEvent or read Context on the nil machine and crashed the server. They now skip the failed context. connections_handlers.go was already safe since it returns early on that error.

Updated the PR description to match the current state of the branch.

Amr-Shams (commented):
LGTM

server/handlers/connections_handlers.go
```diff
@@ -517,17 +517,19 @@ func (h *Handler) NotifySmOfConnectionStatusChange(context context.Context, user
 			event, err := inst.SendEvent(context, machines.EventType(helpers.StatusToEvent(status)), nil)
 			if err != nil {
 				h.log.Error(err)
-				_ = provider.PersistEvent(*event, token)
-				h.config.EventBroadcaster.Publish(userID, event)
 				return
```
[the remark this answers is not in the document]
thc1006: The early return here was an over-correction. SendEvent builds a real event on the connection lookup and status update failure paths, so returning without it loses the failure notification. Fixed in 3b0b9ac, the error path now persists and publishes the event when it is non-nil before returning.

server/handlers/middlewares.go
```diff
@@ -311,14 +311,21 @@ func KubernetesMiddleware(ctx context.Context, h *Handler, provider models.Provi
...
 		if err != nil {
 			h.log.Error(err)
+			continue
 		}
 
 		inst.ResetState()
 		go func(inst *machines.StateMachine) {
 			event, err := inst.SendEvent(ctx, machines.Discovery, nil)
 			if err != nil {
+				h.log.Error(err)
```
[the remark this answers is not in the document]
thc1006: Same fix in 3b0b9ac for both middleware goroutines.

server/handlers/middlewares.go
```diff
@@ -311,14 +311,21 @@ func KubernetesMiddleware(ctx context.Context, h *Handler, provider models.Provi
 		)
 		if err != nil {
```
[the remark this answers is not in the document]
thc1006: Fixed at the source in ebbf4a9 instead. InitializeMachineWithContext was calling smInstanceTracker.Add before it checked the Start error, so a failed init cached a half-built machine. It now adds to the tracker only after Start succeeds, which covers every caller.

server/handlers/k8sconfig_handler.go
```diff
@@ -194,13 +194,21 @@ func (h *Handler) addK8SConfig(user *models.User, _ *models.Preference, w http.R
...
-			go func(inst *machines.StateMachine) {
-				event, err := inst.SendEvent(req.Context(), machines.EventType(mhelpers.StatusToEvent(status)), nil)
-				if err != nil {
-					_ = provider.PersistEvent(*event, token)
-					go h.config.EventBroadcaster.Publish(userID, event)
-				}
-			}(inst)
+			if inst != nil {
+				go func(inst *machines.StateMachine) {
+					event, err := inst.SendEvent(req.Context(), machines.EventType(mhelpers.StatusToEvent(status)), nil)
```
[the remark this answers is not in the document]
thc1006: The helpers.go change in ebbf4a9 covers this path too.

thc1006:
@Amr-Shams thanks for the review. Copilot flagged two more things after your LGTM and I pushed fixes for both:

- 3b0b9ac: SendEvent can return a non-nil event along with an error, the one it builds when the connection lookup or status update fails. The error path was returning early and dropping it, so the failure only landed in the server log and never reached the user. It now persists and publishes that event before returning. The nil-event handling you asked for is unchanged, the success path and the nil log are still the same.
- ebbf4a9: InitializeMachineWithContext was caching the state machine before it checked the Start error, so a failed init left a stale half-built machine for the next call on the same connection. It now adds to the tracker only after Start succeeds.

Could you take another quick look when you get a chance?

PR #18821, opened 2026-04-19
[Docs] Implement resizable sidebar for documentation

### LINKED ISSUE #18804: [Docs] Add resizability to the Sidebar component in Meshery Docs
Image

---

prarii:
**Notes for Reviewers**

- This PR fixes #18804


Bhumikagarggg:
@prarii Thank you for your contribution! Let's discuss this during the website call tomorrow at 5:30 PM IST | 7 AM CST Add it as an agenda item to the meeting minutes, if you would :slightly_smiling_face:

prarii:
@Bhumikagarggg will do right away :))

docs/assets/scss/_sidebar.scss
```diff
@@ -1,15 +1,36 @@
+.flex-container {
+  display: flex;
+  }
 .left-container {
   position: sticky;
   z-index: 9;
   height: calc(100vh - var(--navbar-height));
   transition: all 0.5s ease-in-out;
```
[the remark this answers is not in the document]
CodexRaunak: @prarii please address the gemini suggestions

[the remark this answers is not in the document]
prarii: hi! apologies for the late reply but I've already worked on the suggestions and committed again. The fixes are in my third commit: https://github.com/meshery/meshery/pull/18821/changes/fe37a5b93305e059bc66293f48731fec43c99984. Please let me know if you have any additional feedback :))

docs/assets/scss/_sidebar.scss
```diff
@@ -1,15 +1,36 @@
+.flex-container {
```
banana-three-join: .flex-container is already present in _layout.scss so no need to add it to _sidebar.scss

docs/assets/scss/_sidebar.scss
```diff
@@ -1,15 +1,36 @@
...
+    right: 0;
+    width: 10px;
+    height: 100%;
+    cursor: col-resize;
+    z-index: 999;
+    background: transparent;
+    pointer-events: auto;
+  }
+  
+  &:hover::after {
```
banana-three-join, line 21: Both the base &::after class and hover state also sets background: transparent;. So unless you are actively overriding a different background color from a broader CSS rule, this hover block does absolutely nothing and should be removed to keep the code clean.

docs/static/js/main.js
```diff
@@ -125,4 +125,59 @@ document.addEventListener("click", (event) => {
             sidebarNav.classList.remove("main-navbar--active")
         }
     }
-})
+});
+
+;(function() {
```
banana-three-join, line 131: A space should be added for better readability between functions.

docs/static/js/main.js
```diff
@@ -125,4 +125,59 @@ document.addEventListener("click", (event) => {
...
+      const rect = leftContainer.getBoundingClientRect();
+      if (e.clientX >= rect.right - 10) {
+        isResizing = true;
+        document.body.style.cursor = 'col-resize';
+        document.body.style.userSelect = 'none';
+        e.preventDefault();
+      }
+    });
+
+    document.addEventListener('mousemove', function(e) {
```
banana-three-join, line 165: The animation should be wrapped in a RequestAnimationFrame because mousemove event fires multiple times in a matter of seconds which may cause some jittering

docs/static/js/main.js
```diff
@@ -125,4 +125,59 @@ document.addEventListener("click", (event) => {
...
+      if (!isResizing) return;
+
+      const containerRect = leftContainer.getBoundingClientRect();
+      const newWidth = Math.min(Math.max(e.clientX - containerRect.left, 250), 400);
```
banana-three-join: Remove magic numbers and make them into variables which convey their usage so they can be used for reference.

docs/static/js/main.js
```diff
@@ -125,4 +125,59 @@ document.addEventListener("click", (event) => {
...
+  } else {
+    initResize();
+  }
+  
+  function initResize() {
+    const leftContainer = document.querySelector('.left-container');
+    const sidebar = document.querySelector('.sidebar-container');
+
+    if (!leftContainer || !sidebar) return;
+    if (window.innerWidth <= 1200) return;
```
banana-three-join: If an user loads the website and resizes it quickly to a smaller screen, they might be able to bypass this check. Move the check on the resize to assure that the user is on desktop.

banana-three-join (changes requested):
There's some areas of improvement but overall, really good job! Also, please provide a video so we can see how the implementation looks so we can also mention if there's any needs for changes in that regard as well!

prarii:
@banana-three-join thank you for the feedback. I've resolved all the issues. 
demo

CodexRaunak:
@prarii please make sure that it's stateful i.e we carry the same state to other pages

docs/static/js/main.js
```diff
@@ -156,22 +163,24 @@ document.addEventListener("click", (event) => {
...
 
     document.addEventListener('mousemove', function(e) {
-      if (!isResizing) return;
+  if (!isResizing) return;
 
-      const containerRect = leftContainer.getBoundingClientRect();
-      const newWidth = Math.min(Math.max(e.clientX - containerRect.left, 250), 400);
+  requestAnimationFrame(function() {
+    const containerRect = leftContainer.getBoundingClientRect();
+    const newWidth = Math.min(Math.max(e.clientX - containerRect.left, 250), 400);
```
banana-three-join, line 170: Replace the magic numbers with the constants declare above

banana-three-join (changes requested):
Aside from this minor bug, I would say that the sidebar shouldn't display an overscroll X when it's made smaller and instead, it should make the items inside of it, fit the available space. Other than that, great job!

leecalcote (commented):
Thanks for working on this, @prarii. I see two changes needed:

1. Please make the sidebar width stateful across page navigations / reloads so the chosen width is preserved.
2. When the sidebar is narrowed, it should not introduce horizontal overflow; the items inside should adapt to the available width instead.

A couple of small cleanups would also help: remove the redundant `&:hover::after` rule in `_sidebar.scss`, and replace the remaining hardcoded 250 / 400 values in main.js with the constants already declared above.

Nice work so far.

prarii:
@leecalcote @banana-three-join @CodexRaunak Thank you for the feedback. Apologies it took a while to figure out how to make it stateful. I've updated the code with the changes. Please let me know if anything else is required. 
demo1

docs/assets/scss/_sidebar.scss
```diff
@@ -64,14 +80,17 @@
...
   overscroll-behavior: contain;
   padding: .25rem min(2vh, 2vw);
-  overflow-y: scroll;
+  overflow-y: visible;
+  overflow-x: hidden;
   height: calc(100vh - var(--navbar-height));
-  transition: all 0.5s ease-in-out;
   font-size: 1.5rem;
   width: 320px;
-  max-width: 320px;
```
banana-three-join: Add back the max-width

docs/assets/scss/_sidebar.scss
```diff
@@ -64,14 +80,17 @@
 .sidebar-container {
   overscroll-behavior: contain;
   padding: .25rem min(2vh, 2vw);
-  overflow-y: scroll;
+  overflow-y: visible;
```
banana-three-join: Keep overflow-y: scroll;

docs/assets/scss/_sidebar.scss
```diff
@@ -83,10 +81,10 @@
   overscroll-behavior: contain;
   padding: .25rem min(2vh, 2vw);
   overflow-y: visible;
+  overflow-x: hidden;
   height: calc(100vh - var(--navbar-height));
   font-size: 1.5rem;
   width: 320px;
-  max-width: 400px;
```
banana-three-join: Keep the max-width and make it the same as the current max width in main.js

docs/static/js/main.js
```diff
@@ -139,54 +138,85 @@ document.addEventListener("click", (event) => {
...
+      if (wrapper) {
+        wrapper.style.width = width + 'px';
+      }
+      
+      sidebar.style.width = width + 'px';
+      sidebar.style.maxWidth = MAX_SIDEBAR_WIDTH + 'px';
+      sidebar.style.minWidth = MIN_SIDEBAR_WIDTH + 'px';
+    }
+
+      const DEFAULT_SIDEBAR_WIDTH = 320;
```
banana-three-join: Move up the initialization of this const so it can reside with the other consts for better readability

docs/static/js/main.js
```diff
@@ -139,54 +138,85 @@ document.addEventListener("click", (event) => {
...
+      });
+    });
 
     document.addEventListener('mouseup', function() {
       if (!isResizing) return;
       isResizing = false;
       document.body.style.cursor = '';
       document.body.style.userSelect = '';
+      
+      const currentWidth = parseInt(leftContainer.style.width);
```
banana-three-join: currentWidth is not safe because it relies on the assumption that the width property is stored inline. Use getBoundingClientRect(). Also, because currentWidth.width (once getBoundingClientRect is implemented) returns a raw number, no parseInt is needed. 

This means that ParseInt can also be removed from where the item is called from the localStorage in the code above.

banana-three-join (changes requested):
Aside from the issues pointed out and some identation issues, really good job!

docs/assets/scss/_sidebar.scss
prarii: I originally thought this wouldn't be required since main.js would override the scss file, but I've added it back in.

prarii:
@banana-three-join Thanks again, I've updated with the changes. I've also cleaned up the indentation a little. The localStorage bit wasn't working perfectly so I've added a min and max width to the left container.

banana-three-join:
@prarii Np! By not working perfectly (in regards of the localStorage) what do you mean? I'm asking so I can help you refine it's behavior so it works as intended

prarii:
@banana-three-join No no I just meant that when I restarted the server - it would open at minimum width instead of default. That has been fixed now.

docs/assets/scss/_sidebar.scss
banana-three-join: So from my point of view, SCSS is the primary source of truth for how the DOM should look, so, even if the JS modifies the DOM, it should ultimately abide to SCSS rules.

prarii: fair enough, understood

prarii:
@banana-three-join So, are we good to go? Or is there anything else that you would like me to work on?

banana-three-join:
@prarii I was thinking about what I just said in regards of the min-width from the CSS so I decided to review the JS again and you can keep the min and max width only on the CSS. Whenever the width of the sidebar is being resized, the element is already aware of the min and max width due to those rules already being in the CSS so you don't have to reassign them on init. The resize should only be responsible of resizing the width and not the min-max width. 

TLDR; you can take out the min-max width from the JS and keep it only in the CSS. You can reference the sizes for verification of the width but you shouldn't reassign those rules to the class.

prarii:
@banana-three-join Understood! I've changed that.

banana-three-join (dismissed):
After the implemented changes, it looks good to me!

CodexRaunak:
image
@prarii the box shadow is not visible in the sidebar, please restore the default styling

prarii:
@CodexRaunak my bad about that, I've fixed it now. I believe the box shadow was just clipped because of the overflow condition. It's visible now.

sarajkrishnasingh:
@prarii make sure that you have responded to each and every piece of feedback in comments here.

CodexRaunak:
@prarii 
image
please don't change the styling of the scroll bar

banana-three-join:
Stale approach, continued over https://github.com/meshery/meshery/pull/20348

PR #18827, opened 2026-04-20
Ci/mesheryctl e2e add manual report [test - do not merge]

lekaf974:
**Notes for Reviewers**

- This PR fixes allure reports collision between end2end and unit tets


mesheryctl/tests/e2e/teardown_suite.bash
```diff
@@ -8,9 +8,3 @@ pushd $PWD/../.. > /dev/null
 make clean
 
 popd > /dev/null
-
-
-# Stop port forwarding
-echo "Stopping port forwarding"
-kill $MESHERY_SERVER_PORT_FORWARD_PID
```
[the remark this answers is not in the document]
lekaf974: this is from an older implementation and is no more required

jamieplu (dismissed):
lgtm

guan404ming (dismissed):

.github/workflows/mesheryctl-e2e.yaml
```diff
@@ -109,20 +112,20 @@ jobs:
         
       - name: Checkout QA
         uses: actions/checkout@v6
-        if: ${{ !cancelled() && github.event_name == 'push' && matrix.publish_to_qa == true }}
+        if: ${{ !cancelled() && (github.event_name == 'push' || (github.event_name == 'workflow_dispatch' && inputs.bats_test == 'true' && inputs.publish_to_qa == 'true')) }}
```
leecalcote, line 115: The removal of the matrix.publish_to_qa check means both matrix jobs will try to publish on push, and both manual-dispatch matrix legs will publish when publish_to_qa=true. We need to be aware of the potential for duplicate/racy commits to meshery/qa.

mesheryctl/internal/cli/root/model/delete.go
```diff
@@ -61,14 +60,13 @@ mesheryctl model delete [model-name]
 		}
 
 		if selectedModel == nil {
-			utils.Log.Infof("No model(s) found with the name: %s", modelArg)
-			return nil
+			return ErrDeleteModel(modelArg)
```
leecalcote, line 63: `ErrDeleteModel(nameOrID)` now turns every delete failure into “may not exist”, which hides real causes like 401/500/network failures. It’d be better to keep the original error for API delete failures and only use the “not found” style message when the lookup returns no model.

mesheryctl/internal/cli/pkg/flags/validator.go
```diff
@@ -104,7 +104,7 @@ func (fv *FlagValidator) ReadValidationErrorMessages(err validator.ValidationErr
 		case "oneof":
 			errorMessages = append(errorMessages, fmt.Sprintf("Invalid value for --%s '%v': valid values are %s", strings.ToLower(e.Field()), e.Value(), e.Param()))
 		case "uuid":
```
leecalcote, line 106: Changing the UUID message from `strings.ToLower(e.Field())` to `e.Field()` will surface Go struct names like OrgID / OrganizationID instead of user-facing flag names. 

This seems to be a regression in user experience.

.github/workflows/mesheryctl-e2e.yaml
lekaf974: this is only for testing purpose, I am not going to merge this one. I want to find a way avoid reports override between unit test and end2end test on qa

mesheryctl/internal/cli/pkg/flags/validator.go
lekaf974: currently all flags are using camelCase orgId or orgID etc

mesheryctl/internal/cli/root/model/delete.go
lekaf974: those errors should already be covered in previous step 

```shell
selectedModel, err := promptModelSelection(modelArg, modelsApiPath)
if err != nil {
	return err
}
```

PR #18850, opened 2026-04-21
[UI] restore: session card

saurabhraghuvanshii:
**Notes for Reviewers**
Restores and standardizes the session-expired UX across both Meshery UI surfaces (main UI and provider login route), replacing inconsistent modal behavior and mismatched styling with a unified Sistent-based design.

## Why the old behavior was not working

Two different rendering paths existed:
- Main app modal path (`ui/...`)
- Provider auth path (`provider-ui/...`)

Fixes made only in one path did not affect `/auth/login`, so users still saw older copy/design there.

Earlier session handling used ad-hoc DOM overlays and route-specific logic, which caused drift in:
- modal copy (`Your session has expired` vs `User not authenticated`)
- countdown behavior
- visual styling

Some visual regressions came from hardcoded styles (gray body, orange tint mismatch, icon treatment), not Sistent theme palette.

Logout/session-expired transitions could feel like a black screen due to backdrop/timing behavior before modal render.

---

### `ui/components/General/SessionAuthModals.tsx`
- Reworked session modals to use Sistent components consistently.
- Enforced exact expired-session copy:
  - **Header:** Session Expired  
  - **Body line 1:** User not authenticated  
  - **Body line 2:** You will be redirected to Login page in `{countdown}`
- Countdown standardized to `3 -> 2 -> 1`, then redirect.

**Visual alignment updates:**
- saffron header + saffron bottom strip  
- white content body  
- smaller card footprint  
- moderate rounded corners  
- top-left warning icon with white circular badge  

- Backdrop tuned to reduce black-screen perception during transition.

---

### `provider-ui/pages/auth/login.js`
- Updated provider login session modal to match the same copy/behavior as main UI.
- Standardized countdown to `3 -> 2 -> 1`.

**Design updates:**
- saffron header/strip  
- white body  
- smaller, rounded dialog  
- white circular warning icon badge at top-left  

- Fixed timer effect lifecycle so countdown/render remains stable.

---

### `ui/lib/data-fetch.ts`
- Session-expired handling now triggers modal flow centrally (instead of hard reload-like behavior).
- Prevents duplicate expired-session handling.
- Aligns with modal-first UX.

---

### `ui/lib/sessionTimer.ts`
- Session warning/expired behavior integrated with modal event flow.
- Keeps proactive timeout warning and reactive unauthenticated handling consistent.

---

### `ui/pages/_app.tsx`
- Ensures global session auth modal component is mounted in app shell.
- Guarantees events are rendered reliably.

---

## User-facing outcome

- Same session-expired experience across both app surfaces (`ui` and `provider-ui`)
- Correct copy appears consistently (**User not authenticated**)
- Countdown and redirect behavior are predictable
- Modal visually matches legacy expectation (saffron / white / saffron layout) while using Sistent components/palette

- This PR fixes #

## before 
Screenshot From 2026-04-21 23-10-56

## After

https://github.com/user-attachments/assets/d595f2d5-9012-4551-9120-d98e8bcfde70


ui/lib/data-fetch.ts
banana-three-join, line 3: I believe the change in behavior within the function warrants a change in function name as well since now, instead of showing the session expired modal and redirecting the user, only a session-expired event is dispatched.

ui/lib/sessionTimer.ts
```diff
@@ -34,61 +34,16 @@ function scheduleWarning() {
 }
 
 function showTimeoutWarning() {
```
banana-three-join, line 36: Same here, the warning isn't being shown anymore and only the state is being dispatched now, so a change in the name of the function would be good to showcase a clearer intent.

ui/components/General/SessionAuthModals.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const [showWarning, setShowWarning] = useState(false);
+  const [showExpired, setShowExpired] = useState(false);
+  const [redirectIn, setRedirectIn] = useState(3);
+
+  useEffect(() => {
+    if (typeof window === 'undefined') return;
+
+    const openWarning = () => setShowWarning(true);
+    const closeWarning = () => setShowWarning(false);
+    const openExpired = () => {
```
banana-three-join, line 38: While the event listeners are added once due to them only being added on mount, the event listeners are triggered every time an event happens. So if there's multiple expired events, the setRedirectIn is set again per call, causing a restart on the timer. 

OpenExpired should check the previous state and if it's already true, it should return and wait for the countdown to finish.

ui/components/General/SessionAuthModals.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+    return () => {
+      window.removeEventListener(SESSION_WARNING_EVENT, openWarning);
+      window.removeEventListener(SESSION_WARNING_DISMISS_EVENT, closeWarning);
+      window.removeEventListener(SESSION_EXPIRED_EVENT, openExpired);
+    };
+  }, []);
+
+  const handleStayLoggedIn = () => {
+    fetch('/api/user/prefs', { credentials: 'include' }).catch(() => {});
```
banana-three-join, line 56: Check for 401 status and call the session expired from datafetch so there aren't any silent session-deaths

banana-three-join (changes requested):

PR #17330, opened 2026-02-07
[Server] Policies: Port OPA/Rego policy engine to native Go

guan404ming:
**Notes for Reviewers**

- This PR fixes #17240

## Discussion Forum Thread: 
 https://discuss.layer5.io/t/how-to-run-meshery-with-the-alternative-go-evaluation-engine/7731

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #18853, opened 2026-04-22
[Server] Fix policy engine infinite loop on nil mutator values

guan404ming:
**Notes for Reviewers**

- This PR relates to #17330 (Go policy engine port)

## Why

`patchMutatorsAction` keeps generating the same patch each iteration when a mutator field resolves to `nil`, because the convergence check only compares equality. On large designs this exhausts `MAX_RE_EVALUATION_DEPTH`, producing a timeout on `config patching correctness` (e.g. `meshery-design-fixture.json`).

`policy_binding.go:121` already guards with `|| mutatorValue == nil`. `patchMutatorsAction` (used by edge-non-binding, wallet, hierarchical policies) was missing the same guard.

## How

- Add `|| mutatorValue == nil` to the equality check in `patchMutatorsAction` so a nil mutator is treated as "nothing to patch", matching binding policy behavior.

### DIFF
--- a/server/policies/eval_rules.go
+++ b/server/policies/eval_rules.go
@@ -140,7 +140,7 @@ func patchMutatorsAction(rel *relationship.RelationshipDefinition, design *patte
 		for i := 0; i < count; i++ {
 			mutatorValue := configurationForComponentAtPath(mutatorRefs[i], mutatorComp, design)
 			oldValue := configurationForComponentAtPath(mutatedRefs[i], mutatedComp, design)
-			if deepEqual(mutatorValue, oldValue) || mutatorValue == nil {
+			if mutatorValue == nil || deepEqual(mutatorValue, oldValue) {
 				continue
 			}
 			actions = append(actions, newComponentUpdateAction(getComponentUpdateOp(mutatedRefs[i]), mutatedID, mutatedRefs[i], mutatorValue))
--- a/server/policies/eval_rules.go
+++ b/server/policies/eval_rules.go
@@ -140,7 +140,7 @@ func patchMutatorsAction(rel *relationship.RelationshipDefinition, design *patte
 		for i := 0; i < count; i++ {
 			mutatorValue := configurationForComponentAtPath(mutatorRefs[i], mutatorComp, design)
 			oldValue := configurationForComponentAtPath(mutatedRefs[i], mutatedComp, design)
-			if deepEqual(mutatorValue, oldValue) {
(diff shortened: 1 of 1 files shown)


leecalcote (commented):
@guan404ming, can we future proof this area by applying this guard without naming each and every subtype?

leecalcote (commented):
How are we handling the situation in which the mutation clears the field or sets the field to empty (e.g. mutator intentionally sets [] or "")?

guan404ming:
On future-proofing: Will refactor into a shared helper so patchMutatorsAction and policy_binding both route through one choke point, and any future policy's SideEffects() gets the guard for free.

On empty/cleared values: The guard only skips when mutatorValue == nil (no source value to propagate). Intentional clearing where the mutator is [] or "" is a concrete non-nil value, so it still produces a patch (and converges on the next iter via deepEqual). The only semantic lost is "mutator is literally nil and we want the mutated side to become nil", which the binding policy never supported either, so this keeps behavior consistent.

guan404ming:
The flasky test has been resolved in this pr.

aabidsofi19 (approved):

aabidsofi19:
I think There should be a match mutator utility with this guard in it . So we don't have to repeat this everywhere

leecalcote:
@guan404ming @aabidsofi19, regarding:

> > How are we handling the situation in which the mutation clears the field or sets the field to empty (e.g. mutator intentionally sets [] or "")?

> On empty/cleared values: The guard only skips when mutatorValue == nil (no source value to propagate). Intentional clearing where the mutator is [] or "" is a concrete non-nil value, so it still produces a patch (and converges on the next iter via deepEqual). The only semantic lost is "mutator is literally nil and we want the mutated side to become nil", which the binding policy never supported either, so this keeps behavior consistent.

This makes sense. Prodding a little deeper here, I will ask, how about when a relationship is deleted, is the "mutator is literally nil and we want the mutated side to become nil" attended to? Are mutated properties that were once nil, returned to their original nil state?

leecalcote:
> The flasky test has been resolved in this pr.

🎉

guan404ming:
I found that neither engine does that. On delete we just drop the rel (`actions.go:172`) and `SideEffects()` bails on `StatusDeleted`, so no reverse patch. We don't stash the pre-mutation value anywhere, so a field patched from nil to "foo" stays "foo". Matches OPA, so this PR doesn't regress it, but yeah, real gap. Happy to open a follow-up, fits naturally with @aabidsofi19's shared `matchMutator` refactor.

PR #18854, opened 2026-04-22
[Build] Allow env override for PORT and USE_GO_POLICY_ENGINE in make server

guan404ming:
**Notes for Reviewers**

## Why

The `server` target hardcodes `PORT=9081` and `USE_GO_POLICY_ENGINE=true` as shell-level env prefixes, so running `PORT=9082 USE_GO_POLICY_ENGINE=false make server` does not override them. This makes it impossible to start a second instance (e.g. OPA engine on 9082 alongside the Go engine on 9081) without editing the Makefile.

## How

- Use bash default expansion (`$${VAR:-default}`) for `PORT` and `USE_GO_POLICY_ENGINE` so callers can override via env while keeping the existing defaults.


Makefile
```diff
@@ -179,9 +179,9 @@ server: dep-check
 	cd server; cd cmd; go mod tidy; \
 	BUILD="$(GIT_VERSION)" \
 	PROVIDER_BASE_URLS=$(MESHERY_CLOUD_PROD) \
-	PORT=$(PORT) \
+	PORT=$${PORT:-9081} \
```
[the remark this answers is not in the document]
guan404ming: Good catch, switched to `?=` in `install/Makefile.core.mk` so Make variable precedence is respected. Other targets left as-is per the note.

Makefile
```diff
@@ -179,9 +179,9 @@ server: dep-check
...
 	DEBUG=true \
-	USE_GO_POLICY_ENGINE=true \
+	USE_GO_POLICY_ENGINE=$${USE_GO_POLICY_ENGINE:-true} \
```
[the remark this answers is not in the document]
guan404ming: Addressed, `USE_GO_POLICY_ENGINE ?= true` added in `install/Makefile.core.mk` and referenced via `$(USE_GO_POLICY_ENGINE)`.

install/Makefile.core.mk
```diff
@@ -60,7 +60,8 @@ PROVIDER_CAPABILITIES_FILEPATH="" # Path to capabilities file for remote provide
 #-----------------------------------------------------------------------------
 MESHERY_K8S_SKIP_COMP_GEN ?= TRUE
 APPLICATIONCONFIGPATH="./apps.json"
-PORT:=9081
```
leecalcote, line 63: 9081 is the default. If you want to use a different port use:

```
PORT=9082 make server
```

Does this work for you or is the proposed change needed / better?

guan404ming: `PORT=9082 make server` on master does not work, that's what this PR fixes. `install/Makefile.core.mk:60` has `PORT:=9081` (`:=` is hard assignment, overrides env), so `$(PORT)` in the `server` recipe always expands to 9081 regardless of the env.

Quick proof on master:
```
$ PORT=9082 make -n server | grep PORT=
	PORT=9081 \
```

Changing to `?=` makes the env override work as expected, needed to run two instances side by side (e.g. OPA on 9082 alongside Go engine on 9081 for testing).

leecalcote: Oh, very good. 👍

leecalcote (approved):

PR #18866, opened 2026-04-22
chore: remove dead Meshery application save provider path

jamieplu:
## Summary
- remove the unreachable application save provider methods
- drop the now-unused MesheryApplicationPersister single-save helper
- keep the remaining application read/delete provider surface unchanged

## Testing
- go test ./server/models/... ./server/handlers/...


leecalcote (commented):
lgtm

PR #18885, opened 2026-04-23
[Server] Restore mutated fields when relationship is deleted

### LINKED ISSUE #18881: [Policy] Cleanup on Delete: reverse patched field values when relationship is deleted
### Current Behavior

Follow-up to #18853 ([comment](https://github.com/meshery/meshery/pull/18853#issuecomment-4302709623)).

When a relationship is deleted, field mutations applied by `SideEffects()` are not reversed. A field patched from `nil` to `"foo"` stays `"foo"` after the relationship is removed. The pre-mutation value is never stored, and `SideEffects()` returns early on `StatusDeleted` (`server/policies/policy_binding.go:37`, `server/policies/policy_alias.go:48`).

This matches the OPA engine, so #17330 does not regress it, but it is a real gap.

### Desired Behavior

On relationship delete, restore mutated fields to their pre-mutation values.

### Implementation

- Stash pre-mutation values when a patch is applied.
- On delete, have `SideEffects()` emit reverse patches instead of bailing out.
- Fits naturally with the shared `matchMutator` refactor.

### Acceptance Tests

- Add a binding that patches a field from `nil` to a value, delete the binding, assert the field is back to `nil`.
- Two relationships mutate the same field; deleting one reverses only its own contribution.

guan404ming:
**Notes for Reviewers**

- Fixes #18881 (Cleanup on Delete).
- Relates to #17330 (Go policy engine port).

## Why

When a relationship is deleted, field mutations applied by `SideEffects()` are not reversed. A field patched from an unset/default value to `"foo"` stays `"foo"` after the relationship is removed, because `SideEffects()` returns early on `StatusDeleted` and no reverse patch is emitted. Matches the OPA engine's behavior, so #17330 does not regress it, but it is a real gap.

## How

On delete, emit a reverse patch using a two-tier cleanup:

1. If the mutated component's JSON schema declares a `default` at the mutated path, restore the field to that default.
2. Otherwise, remove the field from the configuration (via \`sjson.Delete\`).

Only emitted when the mutated field still equals the mutator's value; if the field has diverged (user edited it), nothing is changed.

Affected policies: `edge_binding`, `edge_non_binding`, `hierarchical_parent_child`, `hierarchical_wallet`.

## Tests

- `TestWalletPatchingCleanupOnDelete`: delete emits a remove action when no schema default exists.
- `TestWalletPatchingCleanupUsesSchemaDefault`: delete restores the schema default when one is declared.
- `TestWalletCleanupOnDeleteFullPipeline`: end-to-end, the mutated key is absent from the result after delete.
- `TestWalletPatchingSkipsWhenValueDiverged`: no cleanup when the field was user-edited.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,6 +66,7 @@ require (
 	github.com/spf13/pflag v1.0.10
 	github.com/spf13/viper v1.21.0
 	github.com/stretchr/testify v1.11.1
+	github.com/tidwall/sjson v1.2.5
 	github.com/vektah/gqlparser/v2 v2.5.31
 	github.com/vmihailenco/taskq/v3 v3.2.9
 	go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux v0.64.0
@@ -368,7 +369,6 @@ require (
 	github.com/tidwall/gjson v1.18.0 // indirect
 	github.com/tidwall/match v1.2.0 // indirect
 	github.com/tidwall/pretty v1.2.1 // indirect
-	github.com/tidwall/sjson v1.2.5 // indirect
 	github.com/tilt-dev/fsnotify v1.4.8-0.20220602155310-fff9c274a375 // indirect
 	github.com/tonistiigi/dchapes-mode v0.0.0-20250318174251-73d941a28323 // indirect
 	github.com/tonistiigi/fsutil v0.0.0-20250605211040-586307ad452f // indirect
--- a/server/policies/actions.go
+++ b/server/policies/actions.go
@@ -1,16 +1,21 @@
 package policies
 
 import (
+	"encoding/json"
+	"strings"
+
 	patching "github.com/meshery/meshkit/utils/patching"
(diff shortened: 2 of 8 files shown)


leecalcote (commented):
Where is the pre-mutation state being persisted? Relationship evaluations can occur over large spans of time days, weeks, months...

leecalcote:
Persistence discussion - https://mesheryio.slack.com/archives/CFGG6U10E/p1776961812332799

leecalcote (dismissed):
lgtm

leecalcote:
@guan404ming I suspect that we might want to have a couple of other tests for deletion of other relationship subtypes. And that we'll want to mix things up a bit with scenarios in which multiple relationships mutate the same component property.

@aabidsofi19, like when two or more network relationships refer to one or more pod container ports. 

Related to this topic, @aabidsofi19, have we identified a mechanism for tracking multiple items of the same array?

fitzergerald (approved):

PR #18906, opened 2026-04-24
[mesheryctl] feat(model): normalize model generation pipeline and map scaffolds

### LINKED ISSUE #18707: [mesheryctl] UX and Architectural issues with model generate flag and -t template logic

YASHMAHAKAL:
## Description

This PR provides an end-to-end fix and normalization for the `mesheryctl model generate` pipeline. 
Fixes: #18707 

### 1. Server-Side Normalization (`server/handlers/`)
* **Identity Inference:** Added `deriveModelNameFromURL()`. If the user imports a scaffold containing `"untitled-model"`, the server now extracts the repository name from the GitHub URL to use as the canonical model name.
* **Registrant Auto-Detection:** If the source URL contains `github.com` or `githubusercontent.com`, the server automatically forces the registrant to `"github"`. This prevents generation failures when scaffolds default to `"artifacthub"`.
* **Version Normalization:** Added `normalizeVersionFromURL()`. This prevents `meshkit` from accidentally using a filename (like `cert-manager.crds.yaml`) as the version string, and forces canonical semver prefixes (e.g., `v1.16.1`).
* **Canonical File Naming:** Updated the generation pipeline to output `model.json` instead of `<model-name>.json`, matching the layout of curated models.
* **Status Injection:** Newly generated models are now automatically assigned `Status: _model.Enabled`.
* **API Response:** Added logic to populate the `RegistryAPIResponse` with proper counts and summaries when generation runs in "skip registration" mode.

### 2. Schema Alignment (`server/helpers/utils/utils.go`)
* Updated the `DefVersion` constant from `"1.0.0"` to `"v1.0.0"`. This brings the API-driven generator into strict 1:1 parity with `meshkit`'s internal CSV generator and enforces the correct `v1.0.0` directory suffix for schema versioning.

### 3. CLI Scaffold Flattening (`mesheryctl/internal/cli/root/model/import.go`)
* Added `flattenModelScaffold()`. This parses the rich, nested JSON output created by `mesheryctl model init` (e.g., `metadata.primaryColor`, `category.name`) and flattens it into the precise URL-template structure expected by the server's `ImportBody`.
* **Tests:** Replaced the brittle integration test in `import_test.go` with two highly focused unit tests for `flattenModelScaffold` to ensure deep JSON flattening works reliably and preserves existing flat templates.

## How to Test
** Make sure to start meshery server**

### Test 1: Generate a model from a GitHub URL
```bash
# 1. Initialize a generic scaffold
mesheryctl model init my-model

# 2. Run generation against a real GitHub CRD URL (without manual scaffold editing)
mesheryctl model generate \
  -t my-model/v0.1.0/model.json \
  -f https://raw.githubusercontent.com/cert-manager/cert-manager/v1.16.1/deploy/crds/crd-certificates.yaml \
  --skip-registration

# Expected:
# - Model generated at ~/.meshery/models/cert-manager/v1.16.1/v1.0.0/model.json
```


YASHMAHAKAL:
[Screencast from 24-04-26 12:35:56 PM IST.webm](https://github.com/user-attachments/assets/690a0070-a144-42c5-af20-a458914d3710)

mesheryctl/internal/cli/root/model/generate.go
```diff
@@ -118,7 +118,7 @@ func init() {
 
 	generateModelCmd.Flags().StringVarP(&modelGenerateFlags.File, "file", "f", "", "Specify path to the file or directory")
 	generateModelCmd.Flags().StringVarP(&modelGenerateFlags.Template, "template", "t", "", "Specify path to the template JSON file")
-	generateModelCmd.Flags().BoolVarP(&modelGenerateFlags.Register, "register", "r", false, "Skip registration of the model")
+	generateModelCmd.Flags().BoolVarP(&modelGenerateFlags.Register, "skip-registration", "", false, "Skip registration of the model (default is false)")
```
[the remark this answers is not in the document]
lekaf974: This looks legit, can you update struct field to match the flag name

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -196,13 +201,130 @@ func registerModel(data []byte, componentData []byte, relationshipData []byte, f
...
 
-	if len(response.EntityTypeSummary.SuccessfulModels) == 0 {
+	if register && len(response.EntityTypeSummary.SuccessfulModels) == 0 {
 		return utils.ErrInvalidModel()
 	}
 
 	return nil
 }
 
+func flattenModelScaffold(data []byte) ([]byte, error) {
```
lekaf974: the only thing I am seeing is that if there is any changes made in model schema it will required to update this function as well, am I right ?

lekaf974 (changes requested):
some commetns added, I'll do a second pass later

miacycle (commented):
One more thing to tighten here: `flattenModelScaffold()` no longer preserves all of the fields from an already-flat URL template. For example, `mesheryctl/internal/cli/root/model/fixtures/templates/template.json` includes `logo`, but the new flattener never copies it, so `mesheryctl model generate --file <url> --template template.json` will silently drop that metadata before sending the request. Please either pass through the full flat template shape unchanged when it is already in the server-expected format, or add the missing flat keys explicitly (and cover that with a round-trip test so existing templates keep working).

YASHMAHAKAL:
> One more thing to tighten here: `flattenModelScaffold()` no longer preserves all of the fields from an already-flat URL template. For example, `mesheryctl/internal/cli/root/model/fixtures/templates/template.json` includes `logo`, but the new flattener never copies it, so `mesheryctl model generate --file <url> --template template.json` will silently drop that metadata before sending the request. Please either pass through the full flat template shape unchanged when it is already in the server-expected format, or add the missing flat keys explicitly (and cover that with a round-trip test so existing templates keep working).

Thanks, i'll fix it @miacycle

mesheryctl/internal/cli/root/model/generate.go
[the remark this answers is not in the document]
YASHMAHAKAL: yeah, i'll take a look

server/handlers/component_handler.go
```diff
@@ -1122,20 +1123,48 @@ func (h *Handler) RegisterMeshmodels(rw http.ResponseWriter, r *http.Request, _
...
+				if model.ModelDisplayName == "Untitled Model" || model.ModelDisplayName == "" || model.ModelDisplayName == "untitled-model" {
+					model.ModelDisplayName = manifests.FormatToReadableString(derivedName)
+				}
+			} else if isArtifactHubSearchURL(importRequest.ImportBody.Url) {
+				err := fmt.Errorf("artifacthub search URLs are ambiguous; use a direct package URL (/packages/helm/{repo}/{package}) or set the model name in template")
+				h.handleError(rw, err, err.Error())
+				h.sendErrorEvent(userID, provider, err.Error(), err, token)
+				return
+			}
+		}
```
YASHMAHAKAL, line 1141: model name will be auto-derived from source url for artifacthub like we have for github as registrant, previously error was thrown on artifacthub model generations, needed to change model name manually in scaffolds

server/handlers/component_handler.go
```diff
@@ -1115,20 +1116,41 @@ func (h *Handler) RegisterMeshmodels(rw http.ResponseWriter, r *http.Request, _
...
+		if model.Model == "untitled-model" || model.Model == "" {
+			derivedName := deriveModelNameFromURL(importRequest.ImportBody.Url)
+			if derivedName != "" {
+				model.Model = derivedName
+				// Set display name to a readable format if it was also a placeholder
+				if model.ModelDisplayName == "Untitled Model" || model.ModelDisplayName == "" || model.ModelDisplayName == "untitled-model" {
+					model.ModelDisplayName = manifests.FormatToReadableString(derivedName)
+				}
+			}
+		}
```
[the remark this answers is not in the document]
YASHMAHAKAL: Applied

server/handlers/component_handler.go
```diff
@@ -1115,20 +1116,41 @@ func (h *Handler) RegisterMeshmodels(rw http.ResponseWriter, r *http.Request, _
...
+
+		// Auto-detect registrant type from URL
+		if importRequest.ImportBody.Url != "" {
+			if strings.Contains(importRequest.ImportBody.Url, "github.com") || strings.Contains(importRequest.ImportBody.Url, "githubusercontent.com") {
+				model.Registrant = "github"
+			}
+		}
```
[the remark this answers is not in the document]
YASHMAHAKAL: Applied

server/handlers/component_generator_helper.go
```diff
@@ -513,6 +514,63 @@ func setIfEmpty(field *string, defaultValue string) {
...
+		return ""
+	}
+	u, err := url.Parse(sourceURL)
+	if err != nil {
+		return ""
+	}
+	parts := strings.Split(strings.Trim(u.Path, "/"), "/")
+	if len(parts) >= 2 {
+		if strings.Contains(u.Host, "github.com") || strings.Contains(u.Host, "githubusercontent.com") {
+			return parts[1] // The repository name
```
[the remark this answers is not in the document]
YASHMAHAKAL: Applied

server/handlers/component_generator_helper.go
```diff
@@ -513,6 +514,63 @@ func setIfEmpty(field *string, defaultValue string) {
...
+			break
+		}
+	}
+
+	if !isFilename {
+		if version != "" && !strings.HasPrefix(version, "v") {
+			return "v" + version
+		}
+		return version
+	}
```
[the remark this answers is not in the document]
YASHMAHAKAL: Applied

mesheryctl/internal/cli/root/model/generate.go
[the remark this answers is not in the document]
YASHMAHAKAL: Done @lekaf974

mesheryctl/internal/cli/root/model/import.go
YASHMAHAKAL: You are correct. We have partially addressed this by fixing the logo field drop and adding a regression test, but the hardcoded field list is still a maintenance concern.

YASHMAHAKAL: since the current command is broken i'm thinking to address it in follow-up PR

server/handlers/component_handler.go
leecalcote: Please enforce kebab-casing in model names.

server/handlers/component_generator_helper.go
```diff
@@ -523,6 +525,167 @@ func setIfEmpty(field *string, defaultValue string) {
...
+	u, err := url.Parse(sourceURL)
+	if err != nil {
+		return ""
+	}
+
+	if isArtifactHubSourceURL(sourceURL) {
+		return deriveArtifactHubPackageNameFromURL(sourceURL)
+	}
+
+	host := strings.ToLower(u.Hostname())
```
leecalcote: Use the MeshKit library, please.

Reuse existing functionality.

server/handlers/component_generator_helper.go
```diff
@@ -523,6 +525,167 @@ func setIfEmpty(field *string, defaultValue string) {
...
+	switch host {
+	case "github.com", "www.github.com", "raw.githubusercontent.com", "raw.github.com":
```
leecalcote: Are we still giving user's the option of setting the name themselves?

leecalcote: Three of these hostnames are redundant. Only two of these are needed.

server/handlers/component_generator_helper.go
```diff
@@ -523,6 +525,167 @@ func setIfEmpty(field *string, defaultValue string) {
...
+func deriveModelNameFromURL(sourceURL string) string {
+	if sourceURL == "" {
+		return ""
+	}
+	u, err := url.Parse(sourceURL)
+	if err != nil {
+		return ""
+	}
+
+	if isArtifactHubSourceURL(sourceURL) {
```
leecalcote: Careful that we're not over-engineering here. 

This can become brittle fast.

What happens when we add another type of registrant?

server/handlers/component_generator_helper.go
```diff
@@ -523,6 +525,167 @@ func setIfEmpty(field *string, defaultValue string) {
...
+		return true
+	}
+	if len(pathParts) >= 4 && pathParts[0] == "api" && pathParts[1] == "v1" && pathParts[2] == "packages" && pathParts[3] == "search" {
+		return true
+	}
+	return false
+}
+
+func deriveArtifactHubPackageNameFromURL(sourceURL string) string {
+	if !isArtifactHubSourceURL(sourceURL) {
```
leecalcote, line 571: Why are we not matching against Artifact Hub's domain name in symmetric with the other logical condition for github.com?

server/handlers/component_generator_helper.go
```diff
@@ -523,6 +525,167 @@ func setIfEmpty(field *string, defaultValue string) {
...
+
+func isGitHubSourceURL(sourceURL string) bool {
+	parsedURL, err := url.Parse(sourceURL)
+	if err != nil {
+		return false
+	}
+
+	host := strings.ToLower(parsedURL.Hostname())
+	switch host {
+	case "github.com", "www.github.com", "raw.githubusercontent.com", "raw.github.com", "codeload.github.com", "githubusercontent.com":
```
leecalcote: Why are you defining the github case more than once?

server/handlers/component_generator_helper.go
```diff
@@ -523,6 +525,167 @@ func setIfEmpty(field *string, defaultValue string) {
 		*field = defaultValue
 	}
 }
+
+var semverLikeVersionPattern = regexp.MustCompile(`^v?\d+\.\d+(\.\d+)?(?:[-+][0-9A-Za-z][0-9A-Za-z.-]*)?$`)
```
leecalcote: Don't redefine a semver qualifier. Reuse from meshkit or meshery/schemas or maybe meshery server.

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -196,13 +201,130 @@ func registerModel(data []byte, componentData []byte, relationshipData []byte, f
 
 	displayEntities(response)
 
-	if len(response.EntityTypeSummary.SuccessfulModels) == 0 {
+	if register && len(response.EntityTypeSummary.SuccessfulModels) == 0 {
```
leecalcote: Comment this code.

Don't use the word "Successful".

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -196,13 +201,130 @@ func registerModel(data []byte, componentData []byte, relationshipData []byte, f
...
+		return value, ok
+	}
+
+	if modelName := getString(rawMap, "model"); modelName != "" {
+		flatModel["model"] = modelName
+	} else if modelName := getString(rawMap, "name"); modelName != "" {
+		flatModel["model"] = modelName
+	}
+
+	if modelDisplayName := getString(rawMap, "modelDisplayName"); modelDisplayName != "" {
```
leecalcote: No. 

This set of `if` conditions approach is sub-optimal.

leecalcote:
@miacycle, please take assignment of this issue.

leecalcote:
@miacycle I would point you to the issue, but it is not referenced in the pull request description. :/

server/handlers/component_generator_helper.go
YASHMAHAKAL: Yes, users can still set the name themselves; the deriveModelNameFromURL function is only called if the model name is left blank or set to the 'untitled-model' placeholder.

YASHMAHAKAL: model init command currently generating a placeholder, so when there's a placeholder function is called, if we give it a name manually, its absolutely preserved

mesheryctl/internal/cli/root/model/import.go
YASHMAHAKAL: Done

server/handlers/component_generator_helper.go
YASHMAHAKAL: Done

server/handlers/component_generator_helper.go
YASHMAHAKAL: I’ve refactored this into a simple router by extracting the GitHub logic into deriveGitHubPackageNameFromURL, making it symmetric with ArtifactHub. Adding a new registrant now just means adding one isolated helper function without touching existing logic.

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -196,13 +198,90 @@ func registerModel(data []byte, componentData []byte, relationshipData []byte, f
...
+			}
+			result.SvgColor = modelDef.Metadata.SvgColor
+			result.SvgWhite = modelDef.Metadata.SvgWhite
+			if modelDef.Metadata.SvgComplete != nil {
+				result.SvgComplete = *modelDef.Metadata.SvgComplete
+			}
+			if modelDef.Metadata.IsAnnotation != nil {
+				result.IsAnnotation = *modelDef.Metadata.IsAnnotation
+			}
+			// Bug fix: shape was previously not copied — users who set
```
PragalvaXFREZ: this can be removed

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -196,13 +198,90 @@ func registerModel(data []byte, componentData []byte, relationshipData []byte, f
...
+		result.Model = modelDef.Name
+		// Display name: canonical schema uses "displayName"
+		result.ModelDisplayName = modelDef.DisplayName
+		// Registrant: canonical schema stores as a Connection object; extract its Kind
+		result.Registrant = string(modelDef.Registrant.Kind)
+		// Category: canonical schema stores as a CategoryDefinition object; extract its Name
+		result.Category = modelDef.Category.Name
+		// SubCategory: extract the SubCategoryDefinition name string
+		result.SubCategory = modelDef.SubCategory
+		// Metadata fields: nested under ModelDefinition.Metadata
```
PragalvaXFREZ: could you please remove all the redundant comments? , and only keep the ones that require additional details to understand the intent.

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -196,13 +198,90 @@ func registerModel(data []byte, componentData []byte, relationshipData []byte, f
...
+
+	// Fall back to the simple flat template format (fixtures/template.json).
+	// This format already matches schemav1beta1.Model key-for-key, so a
+	// direct unmarshal is sufficient — with one exception: the flat template
+	// uses "displayName" while schemav1beta1.Model expects "modelDisplayName".
+	// We resolve this alias after unmarshaling.
+	if err := encoding.Unmarshal(data, &result); err != nil {
+		return result, err
+	}
+	// Bug fix: the flat template uses "displayName" but schemav1beta1.Model
```
PragalvaXFREZ: this would be irrelevant after the PR lands, please strip such comments.

server/handlers/component_handler.go
```diff
@@ -1120,22 +1121,50 @@ func (h *Handler) RegisterMeshmodels(rw http.ResponseWriter, r *http.Request, _
...
+				return
+			}
+		}
+
+		// Auto-detect registrant type from URL
+		if importRequest.ImportBody.Url != "" {
+			if isGitHubSourceURL(importRequest.ImportBody.Url) {
+				model.Registrant = "github"
+			} else if isArtifactHubSourceURL(importRequest.ImportBody.Url) {
+				model.Registrant = "artifacthub"
```
PragalvaXFREZ, line 1148: question: what happens here if the URL does not match github or artifacthub?

YASHMAHAKAL: The model won't be generated, the unsupported registrant causes ErrUnsupportedRegistrant in GenerateModels which is caught and returned as an HTTP error. That said, we can add an explicit early error here with a clearer message if you'd prefer that UX improvement.

server/handlers/component_generator_helper.go
```diff
@@ -4,11 +4,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
 	"sync"
 
+	semver "github.com/Masterminds/semver/v3"
```
PragalvaXFREZ, line 13: why is this used? please verify if we have a shared library that can be reused.

mesheryctl/internal/cli/root/model/import_test.go
```diff
@@ -36,31 +33,123 @@ func TestHasCSVs(t *testing.T) {
...
+		"isAnnotation": false,
+		"publishToRegistry": false
+	}`)
+
+	got, err := scaffoldToModel(input)
+	require.NoError(t, err)
+
+	assert.Equal(t, "cert-manager", got.Model)
+	assert.Equal(t, "github", got.Registrant)
+	// Bug fix #1: "displayName" in the flat template must map to ModelDisplayName
```
PragalvaXFREZ: leftover bug fix comment

mesheryctl/internal/cli/root/model/import_test.go
```diff
@@ -36,31 +33,123 @@ func TestHasCSVs(t *testing.T) {
...
+	assert.Equal(t, "#0B5FFF", got.PrimaryColor)
+	assert.Equal(t, "#121212", got.SecondaryColor)
+	assert.Equal(t, false, got.IsAnnotation)
+	assert.Equal(t, false, got.PublishToRegistry)
+}
+
+// TestScaffoldToModel_ResolvesDisplayNameAlias verifies that the flat template's
+// "displayName" key (used in fixtures/template.json) is resolved to
+// schemav1beta1.Model.ModelDisplayName even though the json tags don't match.
+// This is Bug fix #1: prevents the user's display name from being silently dropped.
```
PragalvaXFREZ: leftover bug fix comment

mesheryctl/internal/cli/root/model/import_test.go
```diff
@@ -36,31 +33,123 @@ func TestHasCSVs(t *testing.T) {
...
-	ModelCmd.SetOut(buf)
+	assert.Equal(t, "test-model", got.Model)
+	assert.Equal(t, "Test Model", got.ModelDisplayName)
+	assert.Equal(t, "github", got.Registrant)
+	assert.Equal(t, "#00b39f", got.PrimaryColor)
+	assert.Equal(t, "#00D3A9", got.SecondaryColor)
+	assert.Equal(t, "<svg-color/>", got.SvgColor)
+	assert.Equal(t, "<svg-white/>", got.SvgWhite)
+	assert.Equal(t, true, got.IsAnnotation)
+	// Bug fix #2: shape must now be correctly copied from metadata
```
PragalvaXFREZ: leftover bugfix comment

mesheryctl/internal/cli/root/model/import_test.go
```diff
@@ -1,13 +1,10 @@
...
 
 import (
-	"errors"
 	"testing"
 
-	"github.com/jarcoal/httpmock"
-	mesheryctlflags "github.com/meshery/meshery/mesheryctl/internal/cli/pkg/flags"
-	"github.com/meshery/meshery/mesheryctl/pkg/utils"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
PragalvaXFREZ, line 7: @YASHMAHAKAL, quick question, is this used elsewhere in the repo? also which test case does this one help with?

YASHMAHAKAL: mesheryctlflags and mesheryctl/pkg/utils are used in every other test file in this package (generate, build, list, view, etc.) but our new tests are pure unit tests calling scaffoldToModel directly with no CLI or HTTP involvement, those imports aren't needed. testify/require is used on the NoError check so the test stops immediately if the function errors, rather than cascading into panics on the returned empty struct.

server/handlers/component_generator_helper.go
YASHMAHAKAL: Masterminds/semver/v3 is used in normalizeSemVerToken() for reliable version parsing from URLs. It's already a direct dependency in Meshery's go.mod, and is also a direct dependency in meshkit (used internally in meshkit/models/patterns/pattern.go). However, meshkit doesn't expose a public semver wrapper function, so using Masterminds/semver/v3 directly is correct and consistent with how both repos handle semver.

YASHMAHAKAL:
> One more thing to tighten here: `flattenModelScaffold()` no longer preserves all of the fields from an already-flat URL template. For example, `mesheryctl/internal/cli/root/model/fixtures/templates/template.json` includes `logo`, but the new flattener never copies it, so `mesheryctl model generate --file <url> --template template.json` will silently drop that metadata before sending the request. Please either pass through the full flat template shape unchanged when it is already in the server-expected format, or add the missing flat keys explicitly (and cover that with a round-trip test so existing templates keep working).

Logo was never generated by either the old flattenModelScaffold or the new scaffoldToModel — the old function had an explicit allowlist (primaryColor, secondaryColor, shape, svgColor, svgWhite, svgComplete, etc.) and logo was never in it. This is not a regression; it's an upstream schema limitation — schemav1beta1.Model in meshery/schemas v1beta1/import.go has no Logo field, so logo support requires a schema-level change first.

server/models/meshery_meshmodels_api_response.go
```diff
@@ -301,9 +301,9 @@ type EntityCount struct {
 }
 
 type EntityTypeSummary struct {
-	SuccessfulComponents            []map[string]interface{} `json:"successfulComponents"`
-	SuccessfulRelationships         []map[string]interface{} `json:"successfulRelationships"`
-	SuccessfulModels                []map[string]interface{} `json:"successfulModels"`
+	RegisteredComponents            []map[string]interface{} `json:"registeredComponents"`
+	RegisteredRelationships         []map[string]interface{} `json:"registeredRelationships"`
+	RegisteredModels                []map[string]interface{} `json:"registeredModels"`
```
YASHMAHAKAL, line 306: Extended the Registered renaming to Components and Relationships for naming consistency within EntityTypeSummary

server/handlers/component_generator_helper.go
```diff
@@ -523,6 +525,162 @@ func setIfEmpty(field *string, defaultValue string) {
...
+	v, err := semver.NewVersion(cleanedVersion)
+	if err != nil {
+		return "", false
+	}
+	// Always return with a lowercase "v" prefix, normalised.
+	normalized := v.Original()
+	if !strings.HasPrefix(strings.ToLower(normalized), "v") {
+		normalized = "v" + normalized
+	}
+	return "v" + normalized[1:], true
```
lekaf974: looks the if ensure that there is `v` prefix so not sure to understand why doing 

```golang
return "v" + normalized[1:]
```

anirudhjain75:
Most of the changes looks good to me but how about refactoring https://github.com/meshery/meshery/pull/18906/changes#diff-3a45aa6c11a7cbbec546ee483b88a0d8927ab0bdb7220e234b75b487d9395aeeR529 to an extendable map instead of hardcoded functions. 

Any suggestions on this idea ?

YASHMAHAKAL:
> Most of the changes looks good to me but how about refactoring https://github.com/meshery/meshery/pull/18906/changes#diff-3a45aa6c11a7cbbec546ee483b88a0d8927ab0bdb7220e234b75b487d9395aeeR529 to an extendable map instead of hardcoded functions.
> 
> Any suggestions on this idea ?

@anirudhjain75 i had same in mind... idk how i missed this, updated btw

leecalcote:
`init` command work is incomplete. I’m closing this PR until that work is complete until mui-datatables work is complete. Finish what you start.

docs/content/en/reference/mesheryctl/model/generate.md
lekaf974, line 1: Were those changes done manually ?

mesheryctl/internal/cli/pkg/flags/validator.go
```diff
@@ -41,11 +41,11 @@ func (fv *FlagValidator) Validate(s interface{}) error {
 	if err != nil {
 		switch vErr := err.(type) {
 		case *validator.InvalidValidationError:
-			return utils.ErrFlagsInvalid(err)
+			return err
```
lekaf974: This change will break err output for all commands

mesheryctl/internal/cli/root/model/generate.go
```diff
@@ -44,24 +44,28 @@ var generateModelCmd = &cobra.Command{
...
 
 // Generate a model from a URL based on a JSON template skipping registration
-mesheryctl model generate --f [URL] -t [path-to-template.json] -r
+mesheryctl model generate --file [URL] --template [path-to-template.json] --skip-registration
 	`,
 	PreRunE: func(cmd *cobra.Command, args []string) error {
-		return mesheryctlflags.ValidateCmdFlags(cmd, &modelGenerateFlags)
+		err := mesheryctlflags.ValidateCmdFlags(cmd, &modelGenerateFlags)
+		if err != nil {
+			return fmt.Errorf("%s\nRun '%s --help' to see detailed help message", err.Error(), cmd.CommandPath())
```
lekaf974: error returned are not meshkit error

mesheryctl/internal/cli/root/model/generate.go
```diff
@@ -104,21 +108,25 @@ mesheryctl model generate --f [URL] -t [path-to-template.json] -r
...
+			SkipRegister:     modelGenerateFlags.SkipRegistration,
 		}
 
 		return csvModelGenerator.Generate()
 	},
 }
 
 func init() {
+	generateModelCmd.SetFlagErrorFunc(func(cmd *cobra.Command, err error) error {
+		return fmt.Errorf("%s\nRun '%s --help' to see detailed help message", err.Error(), cmd.CommandPath())
```
lekaf974: this is breaking command UX

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -196,13 +198,63 @@ func registerModel(data []byte, componentData []byte, relationshipData []byte, f
...
+			}
+			if modelDef.Metadata.Shape != nil {
+				result.Shape = string(*modelDef.Metadata.Shape)
+			}
+		}
+		return result, nil
+	}
+
+	if err := encoding.Unmarshal(data, &result); err != nil {
+		return result, err
```
lekaf974: this is not a meshkit error

server/handlers/component_generator_helper.go
```diff
@@ -523,6 +525,172 @@ func setIfEmpty(field *string, defaultValue string) {
...
+	// Use the server's existing semver library instead of a custom regex.
+	v, err := semver.NewVersion(cleanedVersion)
+	if err != nil {
+		return "", false
+	}
+	// Always return with a lowercase "v" prefix, normalised.
+	normalized := v.Original()
+	if strings.HasPrefix(strings.ToLower(normalized), "v") {
+		return "v" + normalized[1:], true
+	}
```
lekaf974: not sure to understand this

if `strings.HasPrefix(strings.ToLower(normalized), "v")` is true means there is a prefix `v` so why `return "v" + normalized[1:], true`

server/models/meshery_meshmodels_api_response.go
lekaf974: Should this have been managed from schema instead ?

lekaf974 (commented):
Added some comments, also it would have been easier to review with smaller PRs if it was possible

YASHMAHAKAL:
> Added some comments, also it would have been easier to review with smaller PRs if it was possible

@lekaf974 i'll reduce this PR into suitable focused PR
#19709 - [Refactor: rename --register flag to --skip-registration](https://github.com/meshery/meshery/pull/19709#top)

mesheryctl/internal/cli/pkg/flags/validator.go
YASHMAHAKAL: fixed

mesheryctl/internal/cli/root/model/generate.go
YASHMAHAKAL: fixed

server/handlers/component_generator_helper.go
YASHMAHAKAL: it was redundant, fixed

mesheryctl/internal/cli/root/model/import.go
YASHMAHAKAL: fixed, due to local commit mess, these changes were pushed on this branch

lekaf974:
> > Added some comments, also it would have been easier to review with smaller PRs if it was possible
> 
> @lekaf974 i'll reduce this PR into suitable focused PR #19709 - [Refactor: rename --register flag to --skip-registration](https://github.com/meshery/meshery/pull/19709#top)

Oh it was more suggestion for next time, multiple CI are failing, can you check if it is related to your changes

hortison:
Pull request abandoned by author.

PR #18925, opened 2026-04-25
fix(server): use ErrSavePattern for missing source-type in handlePatternUpdate

leecalcote:
## Summary

- The missing-sourcetype error path in `handlePatternUpdate` was wrapping the persisted-event metadata error in `ErrSaveApplication`. This handler operates on patterns/designs, so `ErrSavePattern` (defined in [`server/handlers/error.go:473`](https://github.com/meshery/meshery/blob/master/server/handlers/error.go#L473)) is the correct MeshKit code. Clients keying on the error code in the persisted event now get the right one.
- Sibling site to the `SaveMesheryPattern` failure block fixed in 78bdda6adc1; that commit was scoped to lines 1788-1800 and explicitly left this site untouched.

## Out of scope

- The adjacent `http.Error` raw write at line 1754 is a Phase 3 migration target (should use `writeMeshkitError`), but that helper is not present on `master` yet, so the migration is deferred.

## Test plan

- [x] `cd server && go build ./...`
- [x] `go vet ./handlers/...`
- [x] `go test ./handlers/ -count=1` — passes (1.152s)


leecalcote:
Closing as superseded.

The domain-mismatch bug this PR fixed (using `ErrSaveApplication` in the missing-sourcetype event metadata of `handlePatternUpdate`) has already been fixed on `master` in 3aa0fd10d7 (PR #18919, "JSON error responses in meshery_pattern_handler.go"), via a strictly better change:

- Introduces `ErrMissingRouteVariable` — semantically more accurate than `ErrSavePattern` for a missing-input case
- Adds the `writeMeshkitError` Phase 3 helper, and migrates the adjacent raw `http.Error` to use it (the migration this PR explicitly deferred)
- Uses a single `missingRouteErr` value for both the persisted event metadata and the response writer — no duplicated error construction

After merging `origin/master` to resolve the conflict, the diff against master is empty, so this PR contributes nothing.

PR #18928, opened 2026-04-25
fix(server): replace SQLite-specific system table query with GORM-agnostic migrator

### LINKED ISSUE #18924: [Server] GetSystemDatabase handler uses SQLite-specific sqlite_schema table — breaks PostgreSQL compatibility
Current Behavior: server/handlers/database_handlers.go line 26 hardcodes sqlite_schema which is SQLite-specific. 
This will fail when Meshery migrates to PostgreSQL.

Desired Behavior: Use GORM's database-agnostic Migrator().GetTables() which already exists in ResetSystemDatabase in 
the same file (line 108), making it compatible with both SQLite and PostgreSQL.

Fix: Replace sqlite_schema table query with GORM's Migrator().GetTables()

Vydeh07:
### Description
This PR addresses technical debt in the system database handler where SQLite-specific system tables (`sqlite_schema`) were hardcoded. This approach would cause the server to fail when Meshery is migrated to or used with PostgreSQL.

### Changes
- Renamed the `SqliteSchema` model to `TableSchema` in `server/models` to reflect a database-agnostic architecture.
- Refactored `GetSystemDatabase` in `server/handlers/database_handlers.go` to use GORM's `Migrator().GetTables()` instead of raw SQL queries.
- Implemented in-memory search, sorting, and pagination for the retrieved table list using the Go `slices` and `strings` packages.
- Maintained the existing API JSON contract to ensure zero breakage for the Meshery UI.

### Impact
This change is a prerequisite for the broader PostgreSQL migration effort and ensures that system database visibility remains functional across different database dialects.

**Closes #18924**


server/models/database_summary.go
```diff
@@ -1,15 +1,15 @@
 package models
 
-type SqliteSchema struct {
+type TableSchema struct {
```
Amr-Shams: refctoring a type or a mdoel would requires a huge change

server/handlers/database_handlers.go
```diff
@@ -66,14 +92,15 @@ func (h *Handler) GetSystemDatabase(w http.ResponseWriter, r *http.Request, _ *m
 
 	val, err := json.Marshal(databaseSummary)
 	if err != nil {
-		fmt.Println(err)
+		h.log.Error(err)
```
Amr-Shams, line 99: we needed the fmt for internal logging not the http response

Amr-Shams (commented):
Could u pls fix the ci

PR #18939, opened 2026-04-27
Compile Golang Policy Engine to WASM

guan404ming:
## Summary

Add a WASM build of the Golang Policy Engine so it can be loaded directly in a browser / extension instead of going through `/api/meshmodels/relationships/evaluate`.

The wasm wraps the same `GoEngine` source the server uses under `USE_GO_POLICY_ENGINE=true` — no logic duplicated. Source and build artifacts live together under `server/policies/wasm/` so the wasm stays colocated with the engine it mirrors.

No frontend code is wired to it yet; the artifact is built locally on demand via `make wasm-engine` so MeshMap / extensions can opt in independently.

## Update: wasm dep slimming

- Wasm now consumes meshkit via build tags from meshery/meshkit#1002 (no hand-written stubs).
- Engine wasm build is its own Go module under `server/policies/wasm/go.mod` with `replace github.com/meshery/meshkit => ../../../../meshkit`. The `replace` can be dropped once the meshkit PR merges and meshery bumps the meshkit version.
- v1alpha3 → v1beta2 bridge split into `server/policies/version_bridge.go` (`//go:build !js`) and `server/policies/version_bridge_js.go` (no-op stub). Without the split, `patternutils` would transitively pull docker/k8s into the wasm binary.
- Final wasm size: **6.2 MB raw / 5.5 MB after `wasm-opt -Oz` / 1.65 MB gzipped** (was ~51 MB unstripped before slimming).

## What's new

| File | Purpose |
|---|---|
| `server/policies/wasm/main.go` | `js && wasm` entry. Imports `server/policies` directly, embeds a no-op `Logger`, exposes `globalThis.evaluateDesign(designJSON, relsJSON)`. |
| `server/policies/wasm/go.mod` / `go.sum` | Isolated wasm module with `replace` to the patched meshkit fork. |
| `server/policies/wasm/README.md` | Build & install steps. |
| `server/policies/wasm/diff_check.cjs` | Node harness used by the diff test. |
| `server/policies/version_bridge.go` / `version_bridge_js.go` | v1alpha3 bridging gated by build tag so wasm doesn't drag in patternutils' k8s/docker deps. |
| `Makefile` (`wasm-engine` target) | One-shot: `GOOS=js GOARCH=wasm go build` into `server/policies/wasm/policy_engine.wasm` + `wasm-opt -Oz` + copy Go's `wasm_exec.js` next to it. |
| `.gitignore` | `server/policies/wasm/{policy_engine.wasm,wasm_exec.js}` — build artifacts stay local, regenerated via `make wasm-engine`. |

## Engine change to enable wasm

`server/policies/engine.go` — replace `meshkit/logger.Handler` with a local `Logger` interface (`Info` + `Warnf` only). `meshkit/logger.Handler` declares 13 methods whose signatures reference `logrus.Level`, `logr.Logger`, and `gorm.io/gorm/logger.Interface` — pulling those into a wasm build is heavy and gorm doesn't compile cleanly for `js/wasm`. The minimal local interface is automatically satisfied by `logger.Handler`, so server-side callers need no changes.

## JS contract

```js
const responseJSON = globalThis.evaluateDesign(designJSON, registeredRelationshipsJSON);
// → JSON of pattern.EvaluationResponse, or {"error": "..."}
```

Caller passes a hydrated `pattern.PatternFile` plus the registered relationships array (fetched once and cached client-side). The wasm intentionally does no registry lookup or hydration.

## Tests

`server/policies/wasm_diff_test.go` — runs the same input through the in-process Go engine and the wasm via Node, asserts byte-equal `EvaluationResponse`:

- `TestWasmEngineMatchesNative` — small orphan-edge fixture (smoke).
- `TestWasmEngineMatchesNativeOnFixture` — the 1.3 MB `policies/test/design_all_relationships.yaml` (broad coverage across hierarchical / binding / matchlabels / alias / edge).
- `assertWasmMatches` helper added to the 3 existing `EvaluateDesign` call sites in `patching_test.go`.

All wasm-touching tests `t.Skip()` cleanly when `node` or `policy_engine.wasm` aren't available, so CI without setup is unaffected.

## Verification

```
make wasm-engine
cd server/policies && go test .   # 140 passed
```

## Side effect: 3 engine determinism fixes

The byte-equal diff test on the 1.3 MB fixture failed initially. Investigation showed the engine itself was non-deterministic across runs (independent of wasm). Fixed in this PR because the new test guards against them:

- `server/policies/utils.go` — `canonicalSeed` was `fmt.Sprintf("%v", ...)` for typed structs, leaking pointer addresses into `staticUUID` seeds.
- `server/policies/policy_binding.go` — `for k := range bindingKinds` (map) leaked iteration order into output.
- `server/policies/policy_matchlabels.go` — same pattern, plus the result is truncated to 20, so different runs picked a *different subset*.

All three: keys are now sorted before iteration / non-string seeds go through `json.Marshal`. `canonicalSeed` change makes generated UUIDs at 2 call sites (alias `selectorSetItem`, `eval_rules` `selectorSet`) **stable** — they were drifting across process restarts before, so anything persisting them was already churning.

## Notes on artifacts

- Output path is `server/policies/wasm/policy_engine.wasm` (6.2 MB raw, 1.65 MB gzipped), not `ui/public/static/`. The repo's existing rego-wasm convention drops 100 KB-class artifacts into `ui/public/static/`; that doesn't scale to a Go-runtime-sized binary, and there is currently no UI loader code consuming it. When UI integration lands, the deploy/build pipeline should run `make wasm-engine` (option A) or move the artifact to a release/CDN path (option B). Both work without retrofitting; revisit then.


server/policies/wasm/main.go
```diff
@@ -0,0 +1,68 @@
...
+//
+// The caller is responsible for hydrating the design and providing the registered
+// relationships; this entry point intentionally does no registry I/O.
+func evaluateDesign(_ js.Value, args []js.Value) any {
+	if len(args) != 2 {
+		return errResult("evaluateDesign(designJSON, relsJSON): expected 2 arguments")
+	}
+
+	var design pattern.PatternFile
+	if err := json.Unmarshal([]byte(args[0].String()), &design); err != nil {
```
aabidsofi19, line 60: Let's extract the index lookups to top .

server/policies/wasm/main.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (noopLogger) Info(...any)          {}
+func (noopLogger) Warnf(string, ...any) {}
+
+// evaluateDesign is exposed to JS as globalThis.evaluateDesign(designJSON, relsJSON).
+// Returns a JSON string of pattern.EvaluationResponse, or {"error": "..."} on failure.
+//
+// The caller is responsible for hydrating the design and providing the registered
+// relationships; this entry point intentionally does no registry I/O.
+func evaluateDesign(_ js.Value, args []js.Value) any {
```
aabidsofi19, line 51: Do we need a js value or can we use strict structs so we don't need to again marshal

server/policies/wasm/main.go
```diff
@@ -0,0 +1,68 @@
...
+	var design pattern.PatternFile
+	if err := json.Unmarshal([]byte(args[0].String()), &design); err != nil {
+		return errResult("decode design: " + err.Error())
+	}
+
+	var relsRaw []any
+	if err := json.Unmarshal([]byte(args[1].String()), &relsRaw); err != nil {
+		return errResult("decode relationships: " + err.Error())
+	}
+	rels := gopolicies.ConvertRelationships(relsRaw)
```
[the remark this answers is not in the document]
aabidsofi19: Seems reasonable

server/policies/wasm/main.go
```diff
@@ -0,0 +1,68 @@
...
+
+	engine := gopolicies.NewGoEngine(noopLogger{})
+	resp, err := engine.EvaluateDesign(design, rels)
```
aabidsofi19: We shouldn't Initiate the engine at every eval . There should be a initEngine function that should do that at the the rels list as arg and store in engine . So we don't need to passall the rels/registry to wasm everytime we are evaluating

aabidsofi19 (changes requested):

server/policies/wasm_diff_test.go
```diff
@@ -0,0 +1,218 @@
...
+	wasmSkipMsg  string
+)
+
+// wasmDiffEnv resolves whether the wasm artifact + node are both present.
+// Resolved once per test process and reused; absent env causes silent skip.
+func wasmDiffEnv(t *testing.T) (root string, ok bool, skip string) {
+	t.Helper()
+	wasmEnvOnce.Do(func() {
+		wasmRoot = repoRoot(t)
+		if _, err := os.Stat(filepath.Join(wasmRoot, "ui/public/static/policy_engine.wasm")); err != nil {
```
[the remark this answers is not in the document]
guan404ming: Fixed. Guard now checks `server/policies/wasm/policy_engine.wasm` to match the Makefile output.

server/policies/wasm/main.go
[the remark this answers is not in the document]
guan404ming: Applied. Wasm now unmarshals straight into `[]*relationship.RelationshipDefinition`. `ConvertRelationships` stays in `engine.go` for the server handler that still feeds untyped registry interfaces.

server/policies/wasm/main.go
guan404ming: Kept the JSON-string contract. `js.FuncOf` mandates the `(js.Value, []js.Value)` signature, and walking a `js.Value` tree is one JS/wasm boundary call per property. For a hydrated design that's much slower than a single `json.Unmarshal` on the wasm side. Open to revisiting if profiling shows the parse cost dominates.

server/policies/wasm/main.go
guan404ming: Done. Folded into the `initEngine` refactor below: registry parses once on init and the engine is constructed once. `evaluateDesign` only takes the design now.

server/policies/wasm/main.go
guan404ming: Applied. Added `initEngine(relsJSON)` that parses and caches the relationship list and constructs a single `GoEngine`. `evaluateDesign(designJSON)` now reuses both, so the caller pays the registry cost once per session.

aabidsofi19:
@guan404ming the relationship tests are failing for this pr

aabidsofi19:
@guan404ming the wasm binary generated is also too big  . we probably need to use the tiny go compiler and reduce dependencies and use optimization flags
image

guan404ming:
@aabidsofi19 Pushed `wasm-opt -Oz` post-processing in `c58d7eb`: 54 MB → 47 MB raw, ~10 MB gzipped (browser-served size).

TinyGo currently blocked by a transitive cgo import:

```
server/policies/wasm
  → server/policies (engine)
  → schemas/v1beta1/pattern
  → schemas/v1beta1/component (component_helper.go)
  → meshkit/database
  → gorm.io/driver/sqlite
  → mattn/go-sqlite3   ← cgo, TinyGo errors with "not implemented: build constraints in #cgo line"
```

`schemas/v1beta1/component/component_helper.go` imports `meshkit/database` at package level, which directly imports `gorm.io/driver/sqlite`. To unblock TinyGo we need either:

- `//go:build !js` on `component_helper.go` in `meshery/schemas` (excludes the gorm-bound helper from wasm builds), or
- split `meshkit/database` so the driver imports aren't pulled in by every consumer.

Both are upstream changes outside this PR. Happy to open a follow-up against `meshery/schemas` if you want to pursue the TinyGo path.

aabidsofi19:
Reducing gorm dep in schemas or adding the flags there sound right to me

aabidsofi19:
fixing merge conflicts

aabidsofi19:
@guan404ming getting the following import errors when trying to compile it : meshery on  eval-wasm [$] via 🐹 v1.25.5 took 12s
❯ make wasm-engine
Building Go relationship engine wasm...
# github.com/docker/go-connections/sockets
/home/aabid/go/pkg/mod/github.com/docker/go-connections@v0.7.0/sockets/sockets.go:16:38: undefined: syscall.RawSockaddrUnix
# github.com/moby/term
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/term_unix.go:21:15: undefined: unix.Termios
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/term_unix.go:39:19: undefined: unix.IoctlGetWinsize
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/term_unix.go:39:49: undefined: unix.TIOCGWINSZ
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/term_unix.go:45:14: undefined: unix.IoctlSetWinsize
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/term_unix.go:45:44: undefined: unix.TIOCSWINSZ
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/term_unix.go:88:31: undefined: unix.Termios
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/term_unix.go:96:32: undefined: unix.Termios
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/termios_unix.go:13:21: undefined: unix.Termios
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/termios_nonbsd.go:11:20: undefined: unix.TCGETS
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/termios_nonbsd.go:12:20: undefined: unix.TCSETS
/home/aabid/go/pkg/mod/github.com/moby/term@v0.5.2/term_unix.go:45:44: too many errors
# k8s.io/kubectl/pkg/util/interrupt
/home/aabid/go/pkg/mod/k8s.io/kubectl@v0.35.2/pkg/util/interrupt/interrupt.go:28:46: undefined: syscall.SIGHUP
# github.com/moby/go-archive/tarheader
/home/aabid/go/pkg/mod/github.com/moby/go-archive@v0.1.0/tarheader/tarheader_unix.go:39:17: undefined: unix.S_IFBLK
/home/aabid/go/pkg/mod/github.com/moby/go-archive@v0.1.0/tarheader/tarheader_unix.go:40:15: undefined: unix.S_IFCHR
/home/aabid/go/pkg/mod/github.com/moby/go-archive@v0.1.0/tarheader/tarheader_unix.go:41:29: undefined: unix.Major
/home/aabid/go/pkg/mod/github.com/moby/go-archive@v0.1.0/tarheader/tarheader_unix.go:42:29: undefined: unix.Minor
# github.com/moby/sys/user
/home/aabid/go/pkg/mod/github.com/moby/sys/user@v0.4.0/idtools_unix.go:99:14: undefined: LookupUser
make: *** [Makefile:517: wasm-engine] Error

aabidsofi19:
this only happens on the wasm target . meshery server is compiling normally

leecalcote:
I’m labeling this new work with `pr/hold` until gaps and bugs are addressed in current implementation.

aabidsofi19:
the issue was the engine was indirectly importing some os native deps ( though not using them )  . the latest commit fixes that

aabidsofi19:
the build has been again failing 
image

aabidsofi19:
related extension work: https://github.com/layer5labs/meshery-extensions/pull/4247/changes

aabidsofi19:
@guan404ming  the Makefile and build workflow changes from this are worth looking at https://github.com/meshery/meshery/pull/19479

leecalcote:
> @guan404ming the Makefile and build workflow changes from this are worth looking at #19479

Copy/paste. Go, go, go. Let's keep this PR from hitting the one month old mark.

guan404ming:
Just updated!

Makefile
```diff
@@ -475,6 +475,33 @@ policy-test:
...
+	@cp -f "$$(go env GOROOT)/lib/wasm/wasm_exec.js" server/policies/wasm/wasm_exec.js
+	@echo "Patching wasm_exec.js to add process.env polyfill..."
+	@sed -i.bak '/chdir() { throw enosys(); },/a\		env: {},' server/policies/wasm/wasm_exec.js
+	@rm -f server/policies/wasm/wasm_exec.js.bak
+	@mkdir -p ui/public/static/wasm
+	@cp -f server/policies/wasm/policy_engine.wasm ui/public/static/wasm/policy_engine.wasm
+	@cp -f server/policies/wasm/wasm_exec.js ui/public/static/wasm/wasm_exec.js
+	@if command -v wasm-opt >/dev/null 2>&1; then \
+		raw=$$(wc -c < server/policies/wasm/policy_engine.wasm); \
+		wasm-opt -Oz --enable-bulk-memory \
```
aabidsofi19, line 509: wasm-opt needs to be present in the  build workflow .

guan404ming: Done. Added an `Install binaryen (wasm-opt)` step before `make wasm-engine` in both `build-ui-and-server.yml` and `build-meshery-ui.yml`, so the size-optimization branch in the Makefile actually runs in CI instead of being skipped.

guan404ming: Update after rebasing onto latest master: upstream removed `build-ui-and-server.yml` (UI build moved to `build-meshery-ui.yml`), so the `Install binaryen (wasm-opt)` step now lives only in `build-meshery-ui.yml`, right before `make wasm-engine`. wasm-opt is still present in the active build workflow.

leecalcote:
What's the size of the wasm binary now?

aabidsofi19 (approved):

guan404ming:
Thanks @aabidsofi19!

PR #18999, opened 2026-04-29
[Server] fix CreateK8sResource returning error after successful force recreate

### LINKED ISSUE #18976: [Server] CreateK8sResource returns an error after a successful force recreate
### Current Behavior
`server/models/pattern/utils/k8s.go` deletes and recreates an existing resource when `force=true` and the live object is not labeled `controller=meshery`, but the function still falls through to the unconditional error path immediately afterward.

That means a force-recreate can succeed in-cluster and still return an error back to the caller.

### Expected Behavior
Once the delete + recreate path succeeds, `CreateK8sResource` should return success instead of continuing into the "resource already exists and is not maintained by meshery" error.

### Screenshots/Logs
Relevant code path:
- `server/models/pattern/utils/k8s.go`
- recreate branch under `if prevObjLables["controller"] != "meshery" { ... if force { ... Create(...) ... } }`
- unconditional error return that follows the `if force` block

### Environment

- **Host OS:** Linux
- **Platform:** Kubernetes
- **Meshery Client Version:** current master

### To Reproduce
1. Apply a namespaced resource that already exists.
2. Ensure the existing object is not labeled with `controller=meshery`.
3. Call the pattern apply path with `force=true`.
4. Observe that the object is recreated, but the handler still returns an error.

---

ANAMASGARD:
## What does this PR do?

Fixes a control-flow bug in `CreateK8sResource`
(`server/models/pattern/utils/k8s.go`) where a successful force-recreate
of a non-Meshery-owned Kubernetes resource still returned an error to the caller.

## Root Cause

When `force=true` and the existing resource is not labeled
`controller=meshery`, the function correctly deleted and recreated the
resource — but the `if force { ... }` block had no `return nil` after the
successful `Create()`. Execution fell through to the unconditional
`ErrCreateResourceEntry` return below it, sending a false error back to
the caller even though the resource was successfully recreated in-cluster.

## Fix

Added `return nil` immediately after the `Create()` succeeds inside the
`if force` block (1 line). The unconditional error return now only runs
when `force=false`, which matches the function's own godoc contract.

**Before:**
```go
if force {
    // delete ...
    // create ... ✅ succeeds
}
// 💥 always ran — even after successful force recreate
return ErrCreateResourceEntry(...)
```

**After:**
```go
if force {
    // delete ...
    // create ... ✅ succeeds
    return nil  // ✅ exit cleanly
}
// only runs when force=false ✅
return ErrCreateResourceEntry(...)
```

## Tests Added

New file: `server/models/pattern/utils/k8s_test.go`

| Case | Expected |
|---|---|
| `force=true`, unowned resource | `nil` — labels `controller=meshery` + `source=pattern` verified post-recreate |
| `force=false`, unowned resource | `error` containing `"already exists"` |
| Meshery-owned resource | `nil` — patch path exercised |
| New resource (no conflict) | `nil` |

## Notes

- `CreateK8sResource` is currently marked `// not in use` at the top of
  the file and has no live callers (commented out in `argo.go`). The fix
  and tests are applied as written per the issue request. The dead-code
  question is out of scope for this PR.
- No new error codes added → `make error` regeneration not required.
- `make golangci` passes (0 issues on the affected package).


Bhumikagarggg:
@ANAMASGARD Kindly follow the community guidelines before opening a PR. Please ensure that the issue is assigned to you before working on it.

reaper8055:
Hey @ANAMASGARD sorry for the delay in reviewing this. Can you help me with the steps on how to reproduce this so I can quickly test this out?

PR #19007, opened 2026-04-29
[Server] Stop ProviderUIHandler from emitting an empty meshery-provider cookie

hamza-mohd:
## Summary

A production playground host is **still** stuck in an infinite `/user/login` ⇄ `/provider` redirect loop. PR #19006 fixed one variant of this loop (cookie not propagating across SameSite/CDN boundaries when an enforced provider IS configured), but its fallback uses `h.Provider` — so it does nothing when `h.Provider` is unset on the deployment, which is exactly this case. This PR closes the second variant at the actual source of the empty cookie.

## Reproduction (live trace from the affected host)

```
$ curl -sS -L --max-redirs 20 -D - -o /dev/null https://<production-host>/

GET /         → 302 Location: /provider?
GET /provider → 302 Location: /user/login    Set-Cookie: meshery-provider=; Path=/; HttpOnly  ← EMPTY VALUE
GET /user/login → 302 Location: /provider
GET /provider → 302 Location: /user/login    Set-Cookie: meshery-provider=; Path=/; HttpOnly  ← EMPTY VALUE
... (loops until curl --max-redirs cap)
```

The cookie shape (no `Expires`, no `MaxAge`) is unique to one writer in the codebase: `ProviderUIHandler`'s auto-select branch.

## Root cause

`server/handlers/provider_handler.go` (pre-fix):

```go
if h.config.PlaygroundBuild || h.Provider != "" {
    http.SetCookie(w, &http.Cookie{
        Name:     h.config.ProviderCookieName,
        Value:    h.Provider,        // "" when PROVIDER env var is unset
        Path:     "/",
        HttpOnly: true,
    })
    http.Redirect(w, r, "/user/login", http.StatusFound)
    return
}
```

When `PLAYGROUND=true` AND `h.Provider` is `""`, the OR short-circuits via `PlaygroundBuild` and writes an **empty** `meshery-provider` cookie. Downstream:

1. `ProviderMiddleware.resolveProviderName` correctly ignores the empty cookie value and falls through to `h.Provider` — also empty.
2. The inline `/user/login` handler (`server/router/server.go:377-385`) fails the type assertion on the resulting `nil` Provider context value.
3. It 302s back to `/provider`. **Loop.**

### Why this affects every stable build

`.github/workflows/build-and-release-stable.yml:35` builds the stable-latest image with `--build-arg PLAYGROUND=true`. Any deployment of those images that omits `PROVIDER` hits this. `install/playground/kubernetes/meshery-playground-deployment.yaml` sets both `PROVIDER` and `PLAYGROUND=true`, but the affected production host evidently ships PLAYGROUND only.

## Fix

Tighten the auto-select gate so it fires **only** when:
- `h.Provider != ""`, AND
- `h.config.Providers[h.Provider]` is registered

Anything else falls through to `ServeUI` (the provider-chooser page), with an `Errorf` naming the specific deployment misconfig so operators see it in logs immediately.

Worst-case behavior is now "user sees the chooser, picks a provider, logs in" — reachable and recoverable, instead of an infinite loop.

The happy path (PLAYGROUND=true + a registered enforced PROVIDER) is **unchanged**: cookie set, single 302 to `/user/login`. Verified by the existing `TestProviderUIHandler_NoneEnvVarRedirectsThroughLogin` plus a new cookie-value assertion (the previous test only checked `Location`, which is precisely how this bug shipped past review).

Also updates `ProviderMiddleware`'s diagnostic Errorf — it claimed "/user/login will redirect-loop until corrected" which is no longer true post-fix; the loop is the symptom we're removing.

## Operator follow-up (out of scope but required for full UX restore)

This code fix prevents the **loop**. To restore one-click playground UX on the affected production host, the deployment also needs a registered `PROVIDER` set on the `meshery` container env (mirror the [playground k8s YAML](https://github.com/meshery/meshery/blob/master/install/playground/kubernetes/meshery-playground-deployment.yaml#L21-L26)).

## Relationship to #19006

| | #19006 | This PR |
|---|---|---|
| Loop trigger | Cookie set by `/provider` doesn't propagate (SameSite/CDN) | `/provider` writes an empty cookie (PROVIDER unset) |
| Fix layer | `ProviderMiddleware.resolveProviderName` falls back to `h.Provider` | `ProviderUIHandler` refuses to write empty/unregistered cookies |
| Helps when `h.Provider == ""`? | No — fallback returns "" too | Yes — degrades to chooser instead of looping |

Both fixes are needed.

## Test plan

- [x] `go test ./server/handlers/` — full handlers package, all tests green
- [x] `go vet ./server/handlers/...` — clean
- [x] `go build ./server/...` — clean
- [x] New `TestProviderUIHandler_LoopGuards` parametric matrix covers all 6 (PlaygroundBuild × h.Provider × registered?) combinations:
  - production-playground regression row (PLAYGROUND=true, PROVIDER unset → must NOT loop)
  - PROVIDER set but unregistered (would also have looped)
  - Both happy paths (auto-select with registered provider)
  - Both no-op paths (multi-provider chooser)
- [x] New `TestProviderUIHandler_PreservesQueryStringOnAutoSelect` guards `?ref=<base64>` deep-link survival
- [x] Extended `TestProviderUIHandler_NoneEnvVarRedirectsThroughLogin` to assert cookie value (regression guard for the missing assertion that let this ship)
- [x] New `assertProviderCookieIsRegistered` / `assertNoProviderCookieSet` helpers — future tests fail loudly if the bad cookie shape returns
- [ ] Post-deploy: re-curl the production host once this lands in a stable build → expect single 302 chain ending at provider chooser page (not a loop)
- [ ] After production deployment is updated with a registered `PROVIDER` → expect single 302 to `/user/login` then OAuth handshake with the configured remote provider

### DIFF
--- a/server/handlers/provider_handler_test.go
+++ b/server/handlers/provider_handler_test.go
@@ -171,7 +171,7 @@ func TestProviderUIHandler_LoopGuards(t *testing.T) {
 		wantCookieValue  string // empty means assert no cookie set
 	}{
 		{
-			name:             "kanvas.new regression: PLAYGROUND=true, PROVIDER unset → must NOT loop",
+			name:             "production playground regression: PLAYGROUND=true, PROVIDER unset → must NOT loop",
 			playgroundBuild:  true,
 			providerOverride: "",
 			providers:        registered,
--- a/server/handlers/provider_handler_test.go
+++ b/server/handlers/provider_handler_test.go
@@ -104,7 +104,7 @@ func TestProviderUIHandler_NoneEnvVarRedirectsThroughLogin(t *testing.T) {
 	}
 
 	// Regression: previous version of this test did not assert on the cookie
-	// value, which let the empty-cookie loop trigger ship to kanvas.new. The
+	// value, which let the empty-cookie loop trigger ship. The
 	// auto-select branch must always set a non-empty cookie naming a
 	// registered provider.
 	assertProviderCookieIsRegistered(t, resp, "meshery-provider", providers)
@@ -127,7 +127,7 @@ func assertProviderCookieIsRegistered(t *testing.T, resp *http.Response, cookieN
 			continue
 		}
 		if c.Value == "" {
-			t.Fatalf("ProviderUIHandler emitted empty %s cookie — this is the kanvas.new redirect-loop trigger", cookieName)
+			t.Fatalf("ProviderUIHandler emitted empty %s cookie — this is a redirect-loop trigger", cookieName)
 		}
 		if _, ok := providers[c.Value]; !ok {
 			t.Fatalf("ProviderUIHandler emitted %s=%q which is not a registered provider — this would loop via ProviderMiddleware nil-provider path", cookieName, c.Value)
@@ -151,7 +151,7 @@ func assertNoProviderCookieSet(t *testing.T, resp *http.Response, cookieName str
 // providerRegistered) combinations that decide whether ProviderUIHandler may
 // auto-select and redirect, or must degrade to the provider-selection UI.
 //
-// The kanvas.new outage was exactly the (PlaygroundBuild=true, h.Provider="",
+// Meshery extension image outages are exactly the (PlaygroundBuild=true, h.Provider="",
 // any registry) row: the handler wrote an empty meshery-provider cookie,
 // which ProviderMiddleware rightly ignored, which made AuthMiddleware bounce
 // the request back to /provider — infinite loop. Guarding that row, plus the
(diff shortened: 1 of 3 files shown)


server/handlers/provider_handler_test.go
```diff
@@ -102,6 +102,199 @@ func TestProviderUIHandler_NoneEnvVarRedirectsThroughLogin(t *testing.T) {
...
+		if c.Name != cookieName {
+			continue
+		}
+		// LogoutHandler-style cookie deletions (MaxAge<0) are intentional
+		// empties — those are not the bug we are guarding against.
+		if c.MaxAge < 0 {
+			continue
+		}
+		if c.Value == "" {
+			t.Fatalf("ProviderUIHandler emitted empty %s cookie — this is the kanvas.new redirect-loop trigger", cookieName)
```
miacycle:
```suggestion
			t.Fatalf("ProviderUIHandler emitted empty %s cookie — this is a redirect-loop trigger", cookieName)
```

PR #19011, opened 2026-04-29
[mesheryctl] Fix invisible provider options in system login

YASHMAHAKAL:
**Notes for Reviewers**
- Resolves an issue where `mesheryctl system login` displayed blank provider options in the terminal prompt.
- Updates the JSON unmarshaling in `mesheryctl/pkg/utils/auth.go` to correctly read the canonical `camelCase` fields (`providerName`, `providerUrl`) that the Meshery Server has emitted since the Phase 5 identifier-naming overhaul (commit `58e59d7d5f8`).
- Implements a custom `UnmarshalJSON` function to dual-accept both the new `camelCase` and legacy `snake_case` tags, maintaining backward compatibility with older server versions during the deprecation window.
- Adds unit tests to ensure both canonical and legacy wire formats are correctly parsed.

### DIFF
--- a/mesheryctl/pkg/utils/auth.go
+++ b/mesheryctl/pkg/utils/auth.go
@@ -22,8 +22,33 @@ import (
 )
 
 type Provider struct {
-	ProviderURL  string `json:"provider_url,omitempty"`
-	ProviderName string `json:"provider_name,omitempty"`
+	ProviderURL  string `json:"providerUrl,omitempty"`
+	ProviderName string `json:"providerName,omitempty"`
+}
+
+func (p *Provider) UnmarshalJSON(data []byte) error {
+	type providerAlias Provider
+	aux := struct {
+		providerAlias
+		ProviderURLLegacy  string `json:"provider_url,omitempty"`
+		ProviderNameLegacy string `json:"provider_name,omitempty"`
+	}{}
+
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	p.ProviderURL = aux.ProviderURL
+	if p.ProviderURL == "" {
+		p.ProviderURL = aux.ProviderURLLegacy
+	}
+
+	p.ProviderName = aux.ProviderName
+	if p.ProviderName == "" {
+		p.ProviderName = aux.ProviderNameLegacy
+	}
+
+	return nil
 }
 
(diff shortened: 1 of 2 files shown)


mesheryctl/pkg/utils/auth.go
```diff
@@ -22,8 +22,33 @@ import (
...
+		p.ProviderURL = aux.ProviderURLLegacy
+	}
+
+	p.ProviderName = aux.ProviderName
+	if p.ProviderName == "" {
+		p.ProviderName = aux.ProviderNameLegacy
+	}
+
+	return nil
 }
```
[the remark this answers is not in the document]
YASHMAHAKAL: While map values are not addressable in Go as a general rule, the encoding/json package handles this internally, when decoding into a map[string]Provider, the decoder allocates a new addressable instance via reflect.New and calls UnmarshalJSON on the pointer, then stores the result back into the map. The current code is correct and verified by the added tests which decode a full map[string]Provider payload successfully.

mesheryctl/pkg/utils/auth.go
```diff
@@ -22,8 +22,33 @@ import (
...
+func (p *Provider) UnmarshalJSON(data []byte) error {
+	type providerAlias Provider
+	aux := struct {
+		providerAlias
+		ProviderURLLegacy  string `json:"provider_url,omitempty"`
+		ProviderNameLegacy string `json:"provider_name,omitempty"`
+	}{}
+
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
```
lekaf974: this must be a meshkit error

mesheryctl/pkg/utils/auth_test.go
```diff
@@ -63,3 +64,37 @@ func TestAuth(t *testing.T) {
 	})
 	//@Aisuko Need a token file to do other testings
 }
+
+func TestProviderUnmarshalJSON(t *testing.T) {
+	t.Run("accepts canonical camelCase provider fields", func(t *testing.T) {
```
lekaf974: please use given, when, then in name test

lekaf974 (changes requested):
added small comments LGTM otherwise

lekaf974 (approved):

PR #19070, opened 2026-05-03
fix: correct misleading error message in ErrGettingRequestContext

### LINKED ISSUE #19069: fix: Incorrect error message in ErrGettingRequestContext in mesheryctl
## Description

The `ErrGettingRequestContext` function in `mesheryctl/internal/cli/root/error.go` returns a misleading error message that says **"Unable to add token to config"** — which is the message belonging to `ErrAddingTokenToConfig`, a completely different error.

When users encounter a failure while creating an HTTP request context (e.g., running `mesheryctl version` or any command that calls the server), they see:

```
Unable to add token to config
```

This is factually incorrect and misleads users into thinking their token configuration is broken, when the actual failure is in creating the HTTP request context.

## Location

- **File:** `mesheryctl/internal/cli/root/error.go`
- **Line:** 52
- **Function:** `ErrGettingRequestContext(err error)`

## Current Code

```go
func ErrGettingRequestContext(err error) error {
    return errors.New(ErrGettingRequestContextCode, errors.Fatal,
        []string{"Unable to add token to config"},
        []string{"Unable to add token to config", err.Error()},
        []string{}, []string{})
}
```

## Expected Behavior

The error message should accurately describe the actual failure — an inability to retrieve or create the HTTP request context.

```go
func ErrGettingRequestContext(err error) error {
    return errors.New(ErrGettingRequestContextCode, errors.Fatal,
        []string{"Unable to get request context"},
        []string{"Unable to get request context", err.Error()},
        []string{}, []string{})
}
```

## Impact

- Confuses users troubleshooting CLI failures
- Makes it harder to identify the real root cause
- The existing `ErrAddingTokenToConfig` already handles the "Unable to add token to config" case correctly, making this a clear copy-paste mistake

parth-agrawall:
## Description

Fixes the `ErrGettingRequestContext` function in `mesheryctl/internal/cli/root/error.go` which was returning the wrong error message.

The function was displaying `"Unable to add token to config"` — which belongs to the separate `ErrAddingTokenToConfig` error — instead of a message that accurately describes a failure to get the HTTP request context.

This was a copy-paste mistake that caused user confusion during troubleshooting: users would see a token config error when the real failure had nothing to do with their token.

## Changes

- `mesheryctl/internal/cli/root/error.go`: Updated `ErrGettingRequestContext` message from `"Unable to add token to config"` → `"Unable to get request context"`

## Testing

The change corrects the string returned by `ErrGettingRequestContext`. Any caller of this function (e.g., `mesheryctl version`) will now display the correct error message when HTTP request context creation fails.

## Related Issue

Closes #19069

---

**Signed-off-by:** Parth Agrawal <94750839+parth-agrawall@users.noreply.github.com>

### DIFF
--- a/mesheryctl/internal/cli/root/error.go
+++ b/mesheryctl/internal/cli/root/error.go
@@ -49,7 +49,7 @@ func ErrConnectingToServer(err error) error {
 }
 
 func ErrGettingRequestContext(err error) error {
-	return errors.New(ErrGettingRequestContextCode, errors.Fatal, []string{"Unable to add token to config"}, []string{"Unable to add token to config", err.Error()}, []string{}, []string{})
+	return errors.New(ErrGettingRequestContextCode, errors.Fatal, []string{"Unable to get request context"}, []string{"Unable to get request context", err.Error()}, []string{}, []string{})
 }
 
 func ErrUnmarshallingAPIData(err error) error {


leecalcote (approved):
Looks good to me. Thanks, @parth-agrawall 👍

Issue #18915, opened 2026-04-24
Evaluation Engine crashes Server when Design ID is invalid


leecalcote:
### Current Behavior
```
INFO[2026-04-24T12:43:34-05:00] attempting to fetch design from cloud for id: undefined  app=meshery caller="github.com/meshery/meshery/server/models.(*RemoteProvider).GetMesheryPattern remote_provider.go:2412"
DEBU[2026-04-24T12:43:34-05:00] constructed design url: http://localhost:9876/api/content/patterns/undefined  app=meshery caller="github.com/meshery/meshery/server/models.(*RemoteProvider).GetMesheryPattern remote_provider.go:2418"
ERRO[2026-04-24T12:43:34-05:00] invalid resource id
 | Short Description: Error failed to get design | Probable Cause: Cannot get the design with the given design ID | Suggested Remediation: Check if the given design ID is correct  app=meshery caller="github.com/meshery/meshery/server/handlers.(*Handler).handleProviderPatternGetError meshery_pattern_handler.go:311" code=meshery-server-1087 probable-cause="Cannot get the design with the given design ID" severity=2 short-description="Error failed to get design" suggested-remediation="Check if the given design ID is correct"
INFO[2026-04-24T12:43:34-05:00] attempting to publish event to remote provider, size: %d918  app=meshery caller="github.com/meshery/meshery/server/models.(*RemoteProvider).persistEventRemote remote_provider.go:1584"
INFO[0578] [OPA] model in design  1
INFO[0578] [OPA] all registered rels count 589
INFO[0578] [OPA] rels to evaluate count 96
INFO[0578] [OPA] Process Checking valid relationship:  Job -> Namespace
[...]

PR #19089, opened 2026-05-04
fix(server): hydrate evaluator-added components with registry styles

leecalcote:
## Summary

Components added by the relationship evaluation engine (for example, a `Namespace` auto-added when one of its child `Pod`s references it) were rendering with default green styling on first paint, and the user had to manually trigger **Reset Styles** on the component for it to display with its proper Kubernetes look (blue, dashed border, integration icon, etc.). Two pre-existing bugs in `processEvaluationResponse` were stranding every evaluator-added component in `unknownComponents` without its registry styles:

1. **v1beta1/v1beta3 type-assertion mismatch**: the meshkit registry returns `*v1beta3.ComponentDefinition` (the only component version whose `Entity` interface is implemented and whose schema binds to `component_definition_dbs`), but the handler asserts `*v1beta1.ComponentDefinition` to match `pattern.PatternFile.Components`. Every successful registry lookup silently failed the assertion. The pre-existing nil-pointer guard test from #18915 was actually asserting the buggy behavior. Fix: a shallow `bridgeComponentV1beta3ToV1beta1` paralleling the existing `v1beta2`↔`v1beta3` bridge in `models/pattern/utils/component_version_bridge.go`, used as a fallback when the v1beta1 assertion fails.
2. **Wildcard `ModelName` filter**: `identify_additions.rego` emits `add_component` actions whose `model` is the relationship-selector wildcard (`{name: "*"}`). The registry's exact-match SQL on `model_dbs.name` returns zero rows for `"*"`. Fix: drop the model filter when `ModelName == "*"` so kind alone disambiguates and the registry returns the canonical match.

Per-instance style customizations on existing components are unaffected — hydration only writes `Trace.ComponentsAdded` entries, and the existing ID-match loop in `processEvaluationResponse` only swaps in hydrated copies for matching IDs.

A companion belt-and-suspenders client-side fix in `meshery-extensions` will follow as a separate PR.

## Test plan

- [x] `go test ./server/handlers/` — full handlers suite passes (2.6s).
- [x] `TestProcessEvaluationResponse_HydratesWildcardModelEntries` (new) — seeds a styled `Namespace`, evaluates with `Model.Name = "*"`, asserts the response carries `Styles.PrimaryColor = "#326CE5"` and `Styles.SvgColor` from the registry.
- [x] Updated `TestProcessEvaluationResponse_NilPointerGuard/v1beta3 registry entity is bridged into v1beta1 hydration` (renamed from "type assertion failure routes to unknownComponents") — verifies the bridged-hydration behavior; the previous test was asserting the bug.
- [x] `go vet ./server/handlers/` — clean.
- [ ] Manual: load a Kanvas design with a `Pod` whose `metadata.namespace` references a non-existent `Namespace`; confirm the auto-added `Namespace` renders with the proper Kubernetes styling on first paint (no manual style reset needed).

### DIFF
--- a/server/handlers/policy_relationship_handler.go
+++ b/server/handlers/policy_relationship_handler.go
@@ -21,6 +21,7 @@ import (
 	"github.com/meshery/schemas/models/v1beta1/component"
 	"github.com/meshery/schemas/models/v1beta1/pattern"
 	"github.com/meshery/schemas/models/v1beta2/relationship"
+	componentv1beta3 "github.com/meshery/schemas/models/v1beta3/component"
 
 	"github.com/meshery/meshkit/logger"
 	"github.com/meshery/meshkit/models/events"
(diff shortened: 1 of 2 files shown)


leecalcote:
Closing in favor of #19090, which has broader coverage (preserves `Styles.Position` across hydration and patches `Actions[].Value["item"]` for consistency between design and actions payloads). I've pushed a follow-up commit on #19090 that swaps the JSON round-trip in `registryComponentDefinitionToV1beta1` for a typed shallow field copy paralleling the existing `ComponentV1beta2ToV1beta3` bridge in `models/pattern/utils/component_version_bridge.go` — same pattern, immune to the `created_at`/`createdAt` JSON-tag drift between v1beta1 and v1beta3, and preserves the alias-pointer contract the existing bridge documents.

PR #19090, opened 2026-05-04
fix(eval): hydrate styles on evaluator-added components

guan404ming:
Components added by the relationship evaluator rendered without styles in Kanvas because the v1beta1 type assertion on the registry entity silently failed (registry returns v1beta3). Convert via JSON round-trip and patch the matching add_component action payload.

### DIFF
--- a/server/handlers/policy_relationship_handler.go
+++ b/server/handlers/policy_relationship_handler.go
@@ -21,6 +21,7 @@ import (
 	"github.com/meshery/schemas/models/v1beta1/component"
 	"github.com/meshery/schemas/models/v1beta1/pattern"
 	"github.com/meshery/schemas/models/v1beta2/relationship"
+	componentv1beta3 "github.com/meshery/schemas/models/v1beta3/component"
 
 	"github.com/meshery/meshkit/logger"
 	"github.com/meshery/meshkit/models/events"
@@ -595,22 +596,73 @@ func componentDefinitionToMap(comp *component.ComponentDefinition) (map[string]i
 	return item, true
 }
 
+// registryComponentDefinitionToV1beta1 bridges the v1beta3 ComponentDefinition
+// the meshkit registry returns into the v1beta1 ComponentDefinition that
+// pattern.PatternFile.Components is typed against. This handler — and every
+// caller of processEvaluationResponse — operates on v1beta1; v1beta3 is only
+// the registry's storage representation.
+//
+// Mechanism: a shallow field copy that keeps pointer- and reference-typed
+// inner fields (Model, Styles, Capabilities, Configuration,
+// Metadata.AdditionalProperties, ModelID) aliased across both versions, so
+// later in-place mutations meshkit helpers (e.g.
+// orchestration.EnrichComponentWithMesheryMetadata) might make through one
+// of those pointers remain visible from the v1beta1 view. This parallels
+// the existing models/pattern/utils.ComponentV1beta3ToV1beta2 bridge — same
+// pattern, different target version.
+//
+// Why not json.Marshal/json.Unmarshal: a JSON round-trip silently drops
+// fields whose tags differ across versions (already true today —
+// v1beta1 uses `json:"created_at,omitempty"` while v1beta3 uses
+// `json:"createdAt"`) and deep-clones the inner pointer targets, which
+// breaks the alias contract the existing bridge documents. Future tag
+// drift would compound the data loss with no compile-time signal.
 func registryComponentDefinitionToV1beta1(entity interface{}) (*component.ComponentDefinition, bool) {
 	if entity == nil {
 		return nil, false
 	}
-	raw, err := json.Marshal(entity)
-	if err != nil {
+	src, ok := entity.(*componentv1beta3.ComponentDefinition)
+	if !ok || src == nil {
 		return nil, false
 	}
-	var converted component.ComponentDefinition
-	if err := json.Unmarshal(raw, &converted); err != nil {
+	if src.Component.Kind == "" {
 		return nil, false
 	}
-	if converted.Component.Kind == "" {
-		return nil, false
+	dst := &component.ComponentDefinition{
+		ID:             src.ID,
+		SchemaVersion:  src.SchemaVersion,
+		Version:        src.Version,
+		DisplayName:    src.DisplayName,
+		Description:    src.Description,
+		Format:         component.ComponentDefinitionFormat(src.Format),
+		Model:          src.Model,
+		ModelReference: src.ModelReference,
+		Styles:         src.Styles,
+		Capabilities:   src.Capabilities,
+		Metadata: component.ComponentDefinition_Metadata{
+			Genealogy:             src.Metadata.Genealogy,
+			IsAnnotation:          src.Metadata.IsAnnotation,
+			IsNamespaced:          src.Metadata.IsNamespaced,
+			Published:             src.Metadata.Published,
+			InstanceDetails:       src.Metadata.InstanceDetails,
+			ConfigurationUISchema: src.Metadata.ConfigurationUISchema,
+			AdditionalProperties:  src.Metadata.AdditionalProperties,
+		},
+		Configuration: src.Configuration,
+		Component: component.Component{
+			Kind:    src.Component.Kind,
+			Version: src.Component.Version,
+			Schema:  src.Component.Schema,
+		},
+		CreatedAt: src.CreatedAt,
+		UpdatedAt: src.UpdatedAt,
+		ModelId:   src.ModelID,
 	}
-	return &converted, true
+	if src.Status != nil {
+		st := component.ComponentDefinitionStatus(*src.Status)
+		dst.Status = &st
+	}
+	return dst, true
 }
 
 // runRelationshipEvaluation runs eval behind a panic-recovery boundary and
--- a/server/handlers/policy_relationship_handler.go
+++ b/server/handlers/policy_relationship_handler.go
@@ -438,18 +438,22 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
 		} else {
 			compFilter.ModelName = _c.ModelReference.Name
 		}
+		if compFilter.ModelName == "*" {
+			compFilter.ModelName = ""
+		}
 
 		entities, _, _, _ := reg.GetEntitiesMemoized(compFilter, registryCache)
 		if len(entities) == 0 {
 			unknownComponents = append(unknownComponents, &_c)
 			continue
 		}
-		_component, ok := entities[0].(*component.ComponentDefinition)
-		if !ok || _component == nil {
+		_component, ok := registryComponentDefinitionToV1beta1(entities[0])
+		if !ok {
 			unknownComponents = append(unknownComponents, &_c)
 			continue
 		}
 
+		originalStyles := _c.Styles
 		_component.ID = _c.ID
 		if _c.DisplayName != "" {
 			_component.DisplayName = _c.DisplayName
@@ -461,6 +465,12 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
 		_component.Metadata.IsAnnotation = _c.Metadata.IsAnnotation
 		_component.Configuration = _c.Configuration
 		_component.Capabilities = &defaultCapabilities
+		if originalStyles != nil && originalStyles.Position != nil {
+			if _component.Styles == nil {
+				_component.Styles = &core.ComponentStyles{}
+			}
+			_component.Styles.Position = originalStyles.Position
+		}
 		compsAdded = append(compsAdded, *_component)
 	}
 
@@ -474,6 +484,7 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
 	evalResponse.Trace.ComponentsUpdated = compsUpdated
 
 	cmps := append(compsAdded, compsUpdated...)
+	hydrateAddComponentActions(evalResponse.Actions, compsAdded)
 
 	if evalPayload.Options != nil && evalPayload.Options.ReturnDiffOnly != nil && *evalPayload.Options.ReturnDiffOnly {
 		evalResponse.Design.Relationships = []*relationship.RelationshipDefinition{}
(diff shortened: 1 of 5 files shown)


server/handlers/policy_relationship_handler.go
```diff
@@ -519,6 +527,84 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
...
+	}
+	var converted component.ComponentDefinition
+	if err := json.Unmarshal(raw, &converted); err != nil {
+		return nil, false
+	}
+	if converted.Component.Kind == "" {
+		return nil, false
+	}
+	return &converted, true
+}
```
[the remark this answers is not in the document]
guan404ming: Added a small per-response cache keyed by component ID so repeated add_component actions reuse the converted item.

server/handlers/policy_relationship_handler.go
```diff
@@ -519,6 +527,84 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
...
+		item, ok := componentDefinitionToMap(&comp)
+		if !ok {
+			continue
+		}
+		if actions[idx].Value == nil {
+			actions[idx].Value = map[string]interface{}{}
+		}
+		actions[idx].Value["item"] = item
+	}
+}
```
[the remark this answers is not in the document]
guan404ming: Keeping the JSON round-trip intentionally here because this converts the full registry component across schema versions, not only metadata. A manual/direct bridge can be considered separately, but the current path is safer for this focused bug fix.

leecalcote:
Related to https://github.com/meshery/meshery/pull/19089, which I'm closing now. I will enhance the JSON conversation portion of this PR (#19090) by swapping the JSON-roundtrip helper for the typed bridge.

server/handlers/policy_relationship_handler.go
```diff
@@ -438,18 +439,22 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
 		} else {
 			compFilter.ModelName = _c.ModelReference.Name
 		}
+		if compFilter.ModelName == "*" {
```
leecalcote: Setting ModelName = "" makes the registry filter return the first hit by kind alone (the Limit:1 clause in the existing filter).

leecalcote: Addressed. The workaround has been removed entirely (commit efe5bbd702c) — a probe test against the actual rego confirmed identify_additions emits a concrete model.name (`"kubernetes"` for the canonical Pod→Namespace fixture), not `"*"`. The branch was guarding against a phantom. The two synthetic wildcard tests that locked in the workaround are also gone. The bigger parity issue you flagged — Go engine missing identify_additions altogether — is now fixed in the same commit via server/policies/policy_inventory.go, with a cross-engine parity test (TestParity_RegoAndGoEngineEmitEquivalentInventoryAdditions) pinning Rego/Go agreement on the auto-added component shape.

server/handlers/policy_relationship_handler.go
```diff
@@ -519,6 +528,140 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
...
+// the existing models/pattern/utils.ComponentV1beta3ToV1beta2 bridge — same
+// pattern, different target version.
+//
+// Why not json.Marshal/json.Unmarshal: a JSON round-trip silently drops
+// fields whose tags differ across versions (already true today —
+// v1beta1 uses `json:"created_at,omitempty"` while v1beta3 uses
+// `json:"createdAt"`) and deep-clones the inner pointer targets, which
+// breaks the alias contract the existing bridge documents. Future tag
+// drift would compound the data loss with no compile-time signal.
+func registryComponentDefinitionToV1beta1(entity interface{}) (*component.ComponentDefinition, bool) {
```
aabidsofi19, line 617: this needs to be moved to schemas under helpers .

aabidsofi19:
@guan404ming  what was the reason that this mismatch between v1beta1 and v1beta3 was not caught by the type checker ,   

I guess probably because we are not using schema structs everywhere if that is the case we need to rectify it .

server/handlers/policy_relationship_handler.go
```diff
@@ -461,6 +463,12 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
 		_component.Metadata.IsAnnotation = _c.Metadata.IsAnnotation
 		_component.Configuration = _c.Configuration
 		_component.Capabilities = &defaultCapabilities
+		if originalStyles != nil && originalStyles.Position != nil {
+			if _component.Styles == nil {
+				_component.Styles = &core.ComponentStyles{}
```
aabidsofi19, line 468: why are we setting to default styles and not the styles from registry

server/handlers/policy_relationship_handler.go
```diff
@@ -444,12 +445,13 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
...
 		}
-		_component, ok := entities[0].(*component.ComponentDefinition)
-		if !ok || _component == nil {
+		_component, ok := registryComponentDefinitionToV1beta1(entities[0])
+		if !ok {
 			unknownComponents = append(unknownComponents, &_c)
 			continue
 		}
 
+		originalStyles := _c.Styles
```
aabidsofi19, line 454: this naming is confusing . what is original and non original . we should use component definition for component from registry and component declaration for components declared in the design

server/handlers/policy_relationship_handler.go
```diff
@@ -474,6 +482,7 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
 	evalResponse.Trace.ComponentsUpdated = compsUpdated
 
 	cmps := append(compsAdded, compsUpdated...)
+	hydrateAddComponentActions(evalResponse.Actions, compsAdded)
```
aabidsofi19, line 485: @guan404ming I think we were already handling this. the hydration logic was already there in the  policy handler function too

server/handlers/policy_relationship_handler.go
```diff
@@ -519,6 +528,140 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
...
+		if id == "" {
+			continue
+		}
+		comp, ok := componentsByID[id]
+		if !ok {
+			continue
+		}
+		item, ok := actionItemsByID[id]
+		if !ok {
+			item, ok = componentDefinitionToMap(&comp)
```
aabidsofi19, line 556: we shouldn't be converting to maps or interfaces this leads to schema inconsistencies down the way

server/handlers/policy_relationship_handler.go
```diff
@@ -519,6 +528,140 @@ func processEvaluationResponse(reg *registry.RegistryManager, evalPayload patter
...
+		return ""
+	}
+	id, ok := item["id"].(string)
+	if !ok {
+		return ""
+	}
+	return id
+}
+
+func componentDefinitionToMap(comp *component.ComponentDefinition) (map[string]interface{}, bool) {
```
aabidsofi19, line 584: no maps

aabidsofi19 (changes requested):

server/handlers/policy_relationship_handler.go
arjunmehta-git: Confirmed — the ModelName == "*" → "" workaround has been removed entirely in efe5bbd702c. A probe test against the actual rego confirmed identify_additions emits a concrete model.name (e.g. "kubernetes" for the Pod→Namespace fixture), not "*", so the branch was guarding a phantom.

server/handlers/policy_relationship_handler.go
arjunmehta-git: Resolved in efe5bbd702c — hydrateAddComponentActions and its componentDefinitionToMap/actionItemID helpers have been removed entirely after porting identify_additions to the Go engine. There is no remaining code to move into schemas/helpers.

server/handlers/policy_relationship_handler.go
arjunmehta-git: Resolved in efe5bbd702c — the originalStyles override block has been removed. The component declaration now keeps the styles returned by the registry instead of being overwritten with declaration-side styles.

server/handlers/policy_relationship_handler.go
arjunmehta-git: Acknowledged. The original/non-original naming was inside the originalStyles override that has since been removed (efe5bbd702c). The remaining variables follow your suggested terminology in spirit (`_component` = registry component definition, `_c` = component declaration from the design). Happy to rename them explicitly to `componentDefinition` / `componentDeclaration` if that would help readability.

server/handlers/policy_relationship_handler.go
arjunmehta-git: Confirmed — the duplicate hydrateAddComponentActions call has been removed in efe5bbd702c. The existing hydration path in the policy handler now stands alone.

server/handlers/policy_relationship_handler.go
arjunmehta-git: Agreed and addressed in efe5bbd702c. componentDefinitionToMap and the entire map[string]interface{} hydration path on the Actions slice have been removed; no more JSON-marshal-to-map conversions in this code path.

server/handlers/policy_relationship_handler.go
arjunmehta-git: Same — actionItemID and componentDefinitionToMap (both returned/operated on map[string]interface{}) are gone in efe5bbd702c. The hydrate-via-map approach has been replaced.

arjunmehta-git:
@aabidsofi19 you're right — the type checker stayed quiet because the registry hands components back as `interface{}` (entities[0] gets type-asserted to `*component.ComponentDefinition` without distinguishing v1beta1 vs v1beta3 storage). A v1beta3 struct masquerading under the v1beta1 type alias passed the assertion silently.

The structural pin is now in place via the typed `v1beta1 ↔ v1beta3` bridge in `EvaluateDesign` (commit 9c135e87a49) — JSON round-trip swapped for a direct type conversion that preserves the field layout and pointer aliasing meshkit's hydration helpers expect.

Auditing the rest of the codebase for spots where we type-assert across schema versions feels like a worthwhile follow-up — happy to scope that as a separate issue if you'd like.

simihablo (approved):
Let's have that codebase audit performed now...

arjunmehta-git:
@aabidsofi19 follow-up audit of cross-schema-version type assertions and bridges. Scope: every Go site that lifts an `interface{}` (or a registry-returned entity) into a specific schema-version struct, plus every JSON-round-trip used as a version bridge.

## Method

1. Mapped each registry filter's actual return version (in meshkit), since that's the source of every `interface{}` the handlers receive.
2. For each consumer, checked the local import alias of `component` / `relationship` / `model` and compared.
3. Flagged every JSON-round-trip used to bridge versions (the same anti-pattern flagged on this PR).

## Registry return contracts (meshkit)

| Filter | Returns |
|---|---|
| `v1beta1.ComponentFilter` | `*v1beta3.ComponentDefinition` (filter file imports v1beta3 component) |
| `v1beta1.ModelFilter` | `*v1beta1.ModelDefinition` (with v1beta3 components nested) |
| `v1alpha3.RelationshipFilter` | `*v1alpha3.RelationshipDefinition` |
| `v1beta1.CategoryFilter` | `*v1beta1.CategoryDefinition` |

## Component type-assertion sites (`*component.ComponentDefinition`)

| Site | Local `component` import | Match |
|---|---|---|
| `server/handlers/policy_relationship_handler.go:447` | v1beta1 | ❌ — fixed in this PR via the typed bridge |
| `server/handlers/component_handler.go:907` | v1beta3 | ✅ |
| `server/handlers/meshery_filter_handler.go:574` | v1beta3 | ✅ |
| `server/machines/helpers/auto_register.go:171` | v1beta3 | ✅ |
| `server/models/meshmodel/core/register.go:116` | v1beta3 | ✅ |
| `server/internal/graphql/resolver/meshmodel.go:107` | v1beta3 | ✅ |
| `server/models/pattern/core/pattern.go:464` | explicit `componentv1beta3` | ✅ |
| `server/models/pattern/resource/selector/workload.go:55` | explicit `componentv1beta3` | ✅ |

**Conclusion:** the only direct mismatch in the codebase is the one this PR fixed. Every other component-assertion site already imports v1beta3, which is what the registry actually returns.

## JSON-round-trip schema bridges (same anti-pattern)

Two more sites use `json.Marshal` + `json.Unmarshal` as a cross-version bridge — the exact failure mode 9c135e8 replaced with a typed bridge:

1. **`server/models/pattern/utils/patternfile_version_bridge.go`** — `PatternV1beta1ToV1beta3` and `PatternV1beta3ToV1beta1`. Bridges `v1beta1/pattern.PatternFile` ↔ `v1beta3/design.PatternFile` via JSON. The file's own header comment ("their JSON wire forms overlap on the fields the policy engine and dehydrate helper read") flags exactly the fragility — a single tag rename in either schema silently drops the field. Heaviest-traffic site: every `EvaluateDesign` call goes through it twice (Hydrate, Dehydrate).
2. **`server/policies/engine.go:33` `ConvertRelationships`** — type-asserts to `*v1beta2.RelationshipDefinition`, falls back to `json.Marshal`/`json.Unmarshal` when the assertion fails. Since the registry returns `v1alpha3.RelationshipDefinition`, **every** registered relationship goes through the JSON path on `USE_GO_POLICY_ENGINE=true`. Same JSON-tag-drift risk; same alias-pointer-loss risk (meshkit relationship helpers that mutate-through-pointer would be silently mutating a deep-clone).

## Same-version JSON deep copies (lower priority)

`server/policies/utils.go:174` `deepCopyDesign` and `:187` `deepCopyRelDef` use JSON round-trip for in-version deep copies. JSON-tag drift is a concern only if the schema evolves under them; alias loss is intentional for a deep copy. Worth replacing if/when we add a generic `Clone()` helper to the schema types, but lower urgency than the cross-version bridges.

## Recommended follow-ups (separate issues)

1. Add a typed `PatternV1beta1ToV1beta3` / `PatternV1beta3ToV1beta1` bridge paralleling the new `registryComponentDefinitionToV1beta1` and the existing `ComponentV1beta2↔V1beta3` bridges. Highest payoff — runs on every evaluation.
2. Promote `ConvertRelationships` to a typed `v1alpha3 → v1beta2` bridge; drop the JSON fallback. Affects every Go-engine evaluation.
3. (Optional, longer-horizon) Standardise on a single `bridges/` package under `models/pattern/utils` so future schema-version moves have one obvious place to land.

No action needed on this PR — those are scope-separable. Filing as follow-up if you'd like.

leecalcote (approved):

leecalcote:
All feedback addressed. Follow up issue opened.

PR #19102, opened 2026-05-05
chore: remove redudant Rego policy tests from CI workflow

ianrwhitney:
This PR removes redudant rego policy test in the meshery ui and server workflow. 

This job is currently redudant as the triggers are aimed to be the same across the separate workflow and the integrated job that was deleted. 

separate workflow: https://github.com/meshery/meshery/blob/master/.github/workflows/rego-lint-and-test.yml


guan404ming (dismissed):
Thanks, this sounds great for me. But I would like to hear other opinion as well.

lekaf974 (approved):
LGTM

guan404ming (approved):
Thanks!

PR #19103, opened 2026-05-05
chore: remove Docker edge push and description steps from CI workflow

ianrwhitney:
Removes redudant/outdated docker push and removes secrets from workflow. 

It appears that this step has been superceded by Meshery Build and Releaser (edge) workflow. This was confirmed by looking at docker hub and checking the GITHUB_SHA to the edge release and following that back to the workflow that generated it.


leecalcote (approved):
@ianrwhitney the Meshery Build and Releaser (edge) workflow and the build-ui-and-server workflow both have similar trigger criteria, so you're right. This lgtm.

leecalcote:
Does this mean that we've been building the edge container duplicately with each PR opened?

PR #19109, opened 2026-05-05
fix: return HTTP 400 on URL parse failure in GrafanaConfigHandler and PrometheusConfigHandler

### LINKED ISSUE #19108: fix: silent HTTP response drop on URL parse failure in GrafanaConfigHandler and PrometheusConfigHandler
## Description

Both `GrafanaConfigHandler` and `PrometheusConfigHandler` silently return without writing any HTTP response when `url.Parse()` fails on the submitted URL. This leaves the HTTP connection in a broken state — the client receives no status code and no response body, causing it to hang until timeout.

## Affected Locations

| File | Line | Handler |
|---|---|---|
| `server/handlers/grafana_handlers.go` | 50–52 | `GrafanaConfigHandler` |
| `server/handlers/prometheus_handlers.go` | 169–171 | `PrometheusConfigHandler` |

## Current Behavior

```go
u, err := url.Parse(grafanaURL)
if err != nil {
    return  // no HTTP response written — client hangs
}
```

When a user submits an invalid URL (e.g. `http://[invalid`), the handler returns immediately without calling `writeMeshkitError()` or writing any response. The HTTP client receives no response and eventually times out.

## Impact

- HTTP clients hang indefinitely waiting for a response when an invalid URL is submitted
- No error message or status code is returned to the user
- Violates the HTTP handler contract — every code path must write a response
- Affects both the Grafana and Prometheus configuration endpoints
- Users get a connection timeout instead of a meaningful `400 Bad Request`

## Proposed Fix

- On `url.Parse()` error, call `writeMeshkitError(w, <error>, http.StatusBadRequest)` before returning
- Log the error with appropriate context
- Apply the same fix to both `grafana_handlers.go` and `prometheus_handlers.go`

parth-agrawall:
## Description

`GrafanaConfigHandler` and `PrometheusConfigHandler` silently returned without writing any HTTP response when `url.Parse()` failed on the submitted URL. This left the HTTP connection in a broken state — clients received no status code, no response body, and eventually timed out.

## Root Cause

```go
u, err := url.Parse(grafanaURL)
if err != nil {
    return  // no HTTP response written — client hangs
}
```

Both handlers had the same pattern. Every other error path in these handlers correctly calls `writeMeshkitError()`, but the URL parse failure path was missed.

## Changes

| File | Change |
|---|---|
| `server/handlers/grafana_handlers.go` | Added `writeMeshkitError` with `http.StatusBadRequest` on URL parse failure |
| `server/handlers/prometheus_handlers.go` | Added `writeMeshkitError` with `http.StatusBadRequest` on URL parse failure |

## Testing

- Submit an invalid URL (e.g. `http://[invalid`) to the Grafana or Prometheus config endpoint
- Before fix: client hangs with no response
- After fix: client receives `400 Bad Request` with a descriptive error message

## Related Issue

Closes #19108

---

**Signed-off-by:** Parth Agrawal <parth.agrawal4002@gmail.com>


parth-agrawall:
I have reviewed the feedback  provided by  @gemini-code-assist  and incorporated both points:

1. **Incorrect error constructor** — `ErrParseBool` was indeed a copy-paste mistake, as it is intended for boolean parsing errors. I have introduced a new dedicated error function `ErrInvalidURL` (error code `meshery-server-1430`) in `server/handlers/error.go` with a message and remediation hint specific to URL parsing failures. Both `GrafanaConfigHandler` and `PrometheusConfigHandler` now use `ErrInvalidURL` for consistent and accurate error logging across handlers.

2. **`url.Parse` limitation noted** — The feedback around `net/url.Parse` misinterpreting path segments as schemes is valid. The existing code already trims the `RequestURI()` suffix from the URL after parsing, which handles the common case of users submitting full URLs with paths. The `ErrInvalidURL` remediation message now explicitly guides users to provide an absolute URL including the scheme (e.g. `http://` or `https://`) to reduce this ambiguity.

The updated commit is pushed to this branch.

parth-agrawall:
Fixed the CI errorutil failure by bumping `next_error_code` from `1430` to `1431` in `server/helpers/component_info.json` to reflect the newly added `ErrInvalidURLCode`.

server/handlers/error.go
```diff
@@ -1048,3 +1003,7 @@ func ErrInitializeMachine(err error) error {
 func ErrSendMachineEvent(err error) error {
 	return errors.New(ErrSendMachineEventCode, errors.Alert, []string{"Failed to advance connection state machine"}, []string{err.Error()}, []string{"The requested event is not valid from the connection's current state.", "A side-effect action attached to the transition (e.g. provisioning, discovery) returned an error."}, []string{"Inspect the connection's current status before retrying. If the failure originates from a side-effect action, address the underlying cause (e.g. cluster reachability, credential validity) and retry."})
 }
+
+func ErrInvalidURL(err error, url string) error {
```
PragalvaXFREZ, line 1053: Could you check how this is handled elsewhere? Adding a new error message here feels a bit odd if we already have an existing helper, error code, or message pattern for invalid URLs.

parth-agrawall: Thank you for the review, @PragalvaXFREZ.

I checked the existing patterns across the handlers:

- **`ErrInvalidLTURL(url string)`** (`meshery-server-1063`) — the closest existing helper, used in `load_test_handler.go`. However, it is semantically specific to load test URLs, does not wrap an underlying `error`, and its remediation message links to the performance management docs — which would be misleading in a Grafana or Prometheus configuration context.
- **`design_import.go`** — handles `url.Parse` failure silently by returning `"unknown_file"` without an HTTP error response, so no reusable error pattern there.

Since no generic URL validation error exists, I introduced `ErrInvalidURL` to avoid misusing a load-test-specific error in a different context. That said, if you'd prefer we reuse `ErrInvalidLTURL` or consolidate differently, I'm happy to follow whatever pattern fits best with the project's conventions.

PR #19119, opened 2026-05-05
fix(auth): own post-login redirect locally in Meshery via cookie

leecalcote:
## Summary

Capture the originally-requested in-app path into a cookie at `InitiateLogin` time and read it back in `TokenHandler`, so Meshery owns its post-login redirect locally instead of round-tripping a `ref` query param through the remote provider's auth chain.

## Why

Intermediate hops on the provider side can drop or rewrite the `ref` parameter — most notably the custom-domain bounce in meshery-cloud's `HandleLogin`, which forwards `source` and `return_to` but not `ref`. When that happens, the next hop synthesizes its own `ref` from the current request URL (which is itself `/login?orgId=…&source=…`), bakes that into Hydra state, and replays it back to Meshery as the post-login destination. The user then lands on `/login` on the Meshery host — a route that does not exist on Meshery server — and gets a 404.

Reproduced today on `playground.meshery.io` after signing in: full chain ends at `https://playground.meshery.io/login?orgId=c5ada327-…&source=<base64 of /api/user/token>`.

The cleanest fix is for Meshery to stop relying on cloud to faithfully round-trip Meshery-specific routing state. The cookie is set on Meshery's own host, never leaves it, and is read back when the provider redirects to `/api/user/token`.

## Changes

- **`InitiateLogin`** now writes the post-login redirect target into the existing `RefCookieName` cookie (previously hardcoded to `/`) with `SameSite=Lax`, `HttpOnly`, and the existing 1h `LoginCookieDuration`. SameSite=Lax ensures the cookie is sent on the top-level cross-site GET when the provider redirects back to `/api/user/token`.
- **`TokenHandler`** reads the cookie first, falls back to `?ref=` for legacy compatibility (mesheryctl, direct extension callbacks, older provider deployments), then `/`. It clears the cookie after use so a stale value can't override the next login flow.
- Selection (`selectPostLoginRefValue`) and value-shape (`computePostLoginRefValue`) logic are extracted into pure helpers in `login_redirects.go` so they can be unit-tested without standing up a `RemoteProvider`.

## Companion change

- meshery-cloud: stop synthesizing a `ref` from the request URL when callers don't pass one, and forward `ref` through the custom-domain bounce as defense-in-depth (PR forthcoming).

## Test plan

- [x] \`go test ./server/models/ -run \"TestResolvePostLoginRedirect|TestComputePostLoginRefValue|TestSelectPostLoginRefValue\" -count=1\` — all pass
- [x] \`go test ./server/models/ -count=1\` — full models package passes
- [x] \`go build ./...\` — clean
- [x] \`go vet ./server/...\` — clean
- [ ] Manual: with the meshery-cloud companion deployed to staging, verify that landing on a deep-linked Meshery URL (e.g. \`/extension/meshmap\`) without a session redirects through cloud auth and lands the user back on \`/extension/meshmap\` rather than \`/login\` (requires staging access).

---
**Companion PR:** [layer5io/meshery-cloud#5204](https://github.com/layer5io/meshery-cloud/pull/5204) — fixes the cloud-side ref-dropping bug and the GetRefURL self-synthesis.

### DIFF
--- a/server/models/login_redirects.go
+++ b/server/models/login_redirects.go
@@ -24,9 +24,9 @@ func resolvePostLoginRedirect(rawRef, fallback string) string {
 	return fallback
 }
 
-// selectPostLoginRefValue returns the raw (encoded or plaintext) value to feed
-// into resolvePostLoginRedirect when the auth flow returns to TokenHandler.
-//
+// selectPostLoginRefValue returns the raw (encoded or plaintext) value to
+// feed into resolvePostLoginRedirect when the auth flow returns to
+// TokenHandler.
 // Meshery is the source of truth for its own post-login destination: the value
 // is captured into a cookie at InitiateLogin time and read back here. The
 // ?ref= query param is a fallback for callers that never went through
@@ -36,20 +36,11 @@ func resolvePostLoginRedirect(rawRef, fallback string) string {
 // since stale provider-side state (e.g. a synthesized ref baked into Hydra
 // state during a custom-domain bounce) was the bug this routing change was
 // introduced to fix.
-// if ref is empty or missing returns "/"
 func selectPostLoginRefValue(r *http.Request, cookieName string) string {
-	redirectURI := ""
 	if ck, err := r.Cookie(cookieName); err == nil  {
-		redirectURI = ck.Value
-	}else {
-	    redirectURI = r.URL.Query().Get("ref")
+		return ck.Value
 	}
-
-	if redirectURI == "" {
-		return "/"
-	}
-
-	return redirectURI
+	return ""
 }
 
 
--- a/server/models/login_redirects.go
+++ b/server/models/login_redirects.go
@@ -1,7 +1,6 @@
 package models
 
 import (
-	"encoding/base64"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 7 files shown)


server/models/login_redirects.go
```diff
@@ -23,6 +25,42 @@ func resolvePostLoginRedirect(rawRef, fallback string) string {
...
+// redirect cookie at InitiateLogin time. An explicit ?ref= query param wins
+// (callers expressing intent override our default), otherwise we fall back to
+// the originally-requested in-app path encoded the same way as the legacy
+// query-string contract. The value is left to resolvePostLoginRedirect to
+// validate at read time, so this stays a pure string transform.
+func computePostLoginRefValue(refQueryParam, callbackURL, baseCallbackURL string) string {
+	if refQueryParam != "" {
+		return refQueryParam
+	}
+	rel := strings.TrimPrefix(callbackURL, strings.TrimSuffix(baseCallbackURL, "/"))
```
aabidsofi19: i am not sure about this logic . it seems to just complicate the ref .  if ref is empty we should empty ref . and in the getter above fallback to "/" if ref is empty or missing

aabidsofi19: the callback url is the one that sets the token and shouldnt be stored in ref eitherway

aabidsofi19: @leecalcote  i will make the  change if this makes sense

leecalcote: Good call — dropped the query-param fallback in selectPostLoginRefValue. The cookie is now the sole source of truth; resolvePostLoginRedirect's "/" fallback handles the missing-cookie case, and not trusting a provider-echoed ref closes the door on the same class of bug we just fixed. Tests updated to assert the new semantics.

server/models/login_redirects_test.go
```diff
@@ -88,3 +90,122 @@ func TestResolvePostLoginRedirect(t *testing.T) {
...
+			callbackURL:   baseCallbackURL + "/somewhere/else",
+			expected:      "L2V4dGVuc2lvbi9rYW52YXM",
+		},
+		{
+			name:           "root path round-trips",
+			callbackURL:    baseCallbackURL + "/",
+			expected:       base64.RawURLEncoding.EncodeToString([]byte("/")),
+			expectedDecode: "/",
+		},
+	}
```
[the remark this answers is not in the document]
leecalcote: Done in 9fd7fb5b9 — added cases for trailing-slash baseCallbackURL (both root and deep-link paths) plus a non-prefix-match defense, and the trim-suffix normalization handles both forms cleanly.

server/models/login_redirects_test.go
```diff
@@ -88,3 +90,122 @@ func TestResolvePostLoginRedirect(t *testing.T) {
...
 }
+
+func TestComputePostLoginRefValue(t *testing.T) {
+	t.Parallel()
+
+	const baseCallbackURL = "https://playground.meshery.io"
+
+	tests := []struct {
+		name           string
+		refQueryParam  string
```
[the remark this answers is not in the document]
leecalcote: Added test cases for trailing-slash baseCallbackURL forms (the docs example shape). The full "/api/user/token" path concern is no longer reachable: the cookie value is computed from the callbackURL relative path at InitiateLogin time and read back as-is at TokenHandler — there's no opportunity for an absolute URL to land in the cookie.

server/models/login_redirects.go
```diff
@@ -23,6 +25,38 @@ func resolvePostLoginRedirect(rawRef, fallback string) string {
...
+// (callers expressing intent override our default), otherwise we fall back to
+// the originally-requested in-app path encoded the same way as the legacy
+// query-string contract. The value is left to resolvePostLoginRedirect to
+// validate at read time, so this stays a pure string transform.
+func computePostLoginRefValue(refQueryParam, callbackURL, baseCallbackURL string) string {
+	if refQueryParam != "" {
+		return refQueryParam
+	}
+	return base64.RawURLEncoding.EncodeToString([]byte(strings.TrimPrefix(callbackURL, baseCallbackURL)))
+}
```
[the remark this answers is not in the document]
leecalcote: Fixed in the same change — strings.TrimSuffix(baseCallbackURL, "/") before the prefix trim, plus a leading-slash guard for non-prefix remainders. Tests cover both forms.

fitzergerald (dismissed):

server/models/login_redirects.go
aabidsofi19: @leecalcote nice . i have cleaned up the computeRedirectUrl .

aabidsofi19 (approved):

simihablo (approved):
lgtm

leecalcote:
LGTM, @aabidsofi19. Post-merge of this PR, let's hold off on a Meshery release until some of the coordination is complete between the shifting Sistent RJSF forms and schemas repo. Cloud release v1.0.56 has restored Playground access for all in the meantime.

aabidsofi19:
@leecalcote  makes sense

PR #19127, opened 2026-05-06
[Server] Fix non-determinism in relationship eval engine

guan404ming:
**Notes for Reviewers**

- Three sources of cross-process non-determinism in the Go relationship eval engine produced different relationship IDs / different selected matchlabel groups for the same input design across runs.
- Carved out from #18939 so it can land independently of the wasm work.

### DIFF
--- a/server/policies/policy_binding.go
+++ b/server/policies/policy_binding.go
@@ -1,6 +1,7 @@
 package policies
 
 import (
+	"sort"
 	"strings"
 
 	"github.com/meshery/schemas/models/v1beta2/relationship"
@@ -167,7 +168,14 @@ func identifyBindingRelationships(relDef *relationship.RelationshipDefinition, d
 		fromComps := filterComponentsByKindSetTyped(design.Components, fromSelByKind)
 		toComps := filterComponentsByToSelectorsTyped(design.Components, ss.Allow.To)
 
-		for bindingKind := range bindingKinds {
+		// Iterate kinds in sorted order so identification output (and thus
+		// generated relationship IDs) does not depend on map iteration randomness.
+		sortedBindingKinds := make([]string, 0, len(bindingKinds))
+		for k := range bindingKinds {
+			sortedBindingKinds = append(sortedBindingKinds, k)
+		}
+		sort.Strings(sortedBindingKinds)
+		for _, bindingKind := range sortedBindingKinds {
 			bindingDecls := filterComponentsByKindTyped(design.Components, bindingKind)
 			if len(bindingDecls) == 0 {
 				continue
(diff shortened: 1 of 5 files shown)


server/policies/policy_matchlabels.go
```diff
@@ -161,7 +169,13 @@ func identifyMatchlabelRelationships(relDef *relationship.RelationshipDefinition
 			},
 		}
 
-		id := staticUUID(fmt.Sprintf("%v%s", ss, "sibling_match_labels_policy")).String()
+		// Seed via the structured ss (canonicalSeed json-marshals it) instead
```
guan404ming: We don't need this to prevent conflict.

leecalcote (changes requested):
Missing documentation update.

leecalcote (changes requested):
New tests written or existing tests updated?

leecalcote:
Thanks for splitting this out. Sensible and well-scoped. One requested change: please add regression tests for the new deterministic paths, especially:

1. canonicalSeed/staticUUID with a struct containing pointer fields
1. sorted/truncated matchlabels or binding iteration, so output order stays stable across runs

Where are the meeting minutes in which you @aabidsofi19 discussed the displacement of matchlabels with tagsets?

guan404ming:
They follow up work items is listed in this issue https://github.com/meshery/meshery/issues/19149, please take a look, thanks!

guan404ming:
I will handle documentation in another pr which will only focus on doc.

aabidsofi19 (dismissed):

server/policies/policy_binding.go
```diff
@@ -167,7 +168,14 @@ func identifyBindingRelationships(relDef *relationship.RelationshipDefinition, d
...
 		toComps := filterComponentsByToSelectorsTyped(design.Components, ss.Allow.To)
 
-		for bindingKind := range bindingKinds {
+		// Iterate kinds in sorted order so identification output (and thus
+		// generated relationship IDs) does not depend on map iteration randomness.
+		sortedBindingKinds := make([]string, 0, len(bindingKinds))
+		for k := range bindingKinds {
+			sortedBindingKinds = append(sortedBindingKinds, k)
+		}
+		sort.Strings(sortedBindingKinds)
```
mrmerlin320, line 177: why are we again sorting here

mrmerlin320: oh see . this is only intialized

guan404ming: Yeah, that is for initialized!

guan404ming:
Hi @aabidsofi19 could you take another look, thanks!

aabidsofi19 (approved):

PR #19132, opened 2026-05-06
fix: /management/connections MUI  MenuList issue

saurabhraghuvanshii:
**Notes for Reviewers**
In ui/components/connections/ConnectionChip.tsx, we changed ConnectionStateChip to render only visual chip components and no longer wrap them in MenuItem.

- Removed ConnectionStyledMenuItem wrappers from all status chip variants.
- Kept each state renderer (Connected, Registered, Discovered, etc.) as plain chip output.
- Added explicit React import for TS/lint compatibility.
- Removed invalid value prop passed to the fallback DiscoveredChip.

- ConnectionStateChip is used in multiple places:
   - inside <MenuItem> in dropdowns, and
   - outside menu contexts in normal layout/cards/modals.

- Wrapping it with MenuItem caused invalid nesting and context misuse, leading to:
    - MUI: MenuListContext is missing. MenuItems must be placed within Menu or MenuList.

#### Result

- Fixes the runtime MenuListContext error on connections flows.
- Keeps status-chip UI behavior unchanged (same labels/icons/colors), but with correct component composition.

- This PR fixes #

### Before 

Screenshot From 2026-05-06 19-19-28

### After 

image


thechillbasu (commented):
I guess this is good to merge. Thank you for your contribution!

hamza-mohd (changes requested):
Please justify why `ConnectionStyledMenuItem` is being removed.

saurabhraghuvanshii:
> Please justify why `ConnectionStyledMenuItem` is being removed.

Removed ConnectionStyledMenuItem because ConnectionStateChip is used as a shared status-display component across both menu and non-menu contexts. Wrapping it internally with MenuItem made that component context-dependent, which is incorrect for its usage pattern.

In practice, this caused two structural issues:

- Invalid context usage: when rendered outside a Menu/MenuList, MUI throws MenuListContext is missing.
- Nested menu items: in select/dropdown paths, ConnectionStateChip was already wrapped by a parent MenuItem, so the internal wrapper created MenuItem inside MenuItem.
- 
The fix keeps ConnectionStateChip presentation-only (chips/icons/labels) and leaves interaction/container semantics (MenuItem, select behavior) to the parent component. This preserves UI output while resolving runtime errors and aligning with MUI composition rules.

pontusringblom (approved):

pontusringblom:
@saurabhraghuvanshii why is the extension menu item misplaced?

leecalcote:
Good fix - removing `MenuItem` wrapping is correct since `ConnectionStateChip` is used in both menu and non-menu contexts. For future PRs, please sign your commits with `git commit -s`.

---

PR #19155, opened 2026-05-07
[Server] Fix incorrect catalog path in DefaultLocalProvider

jahnavigajjala-3:
## Description

This PR fixes #19139 

- Corrects an incorrect catalog directory path in
  `server/models/default_local_provider.go`

- Fixes silent failure of seeded catalog pattern loading
  during startup inside the container runtime

- Updates the catalog path from:
  `../../docs/catalog`

  to:
  `../../docs/data/catalog`

- Prevents the catalog page from appearing empty due to
  patterns failing to load silently

## Signed commits

- [x] Yes, I signed my commits.

### `server/models/default_local_provider.go`

Updated the catalog directory path.

Before:

```go
catalogDir := filepath.Join("..", "..", "docs", "catalog")

After:

```go
catalogDir := filepath.Join("..", "..", "docs", "data", "catalog")

PR #19156, opened 2026-05-07
[CI] Broaden first-time contributor check to include NONE association

Junnygram:
This PR updates the agentic first-time contributor workflow to include users with the `NONE` association. 

### Why
In recent live tests (e.g., #19155), we observed that GitHub initially flags some absolute newcomers as `NONE` rather than `FIRST_TIME_CONTRIBUTOR`. This caused the agentic greeting to skip, falling back to legacy/generic greetings.

### Changes
- Updated the `if` condition in `first-time-contributor.md` to accept both `FIRST_TIME_CONTRIBUTOR` and `NONE`.
- Recompiled the workflow lockfile.

Relates to #19053


leecalcote:
Good improvement - including the `NONE` association is well-justified by the live-test observation in #19155. Real-world evidence makes this change easy to trust.

---

PR #19166, opened 2026-05-07
fix(remote-provider): only set X-API-Key for the anonymous global token

leecalcote:
## Summary

`RemoteProvider.doRequestHelper` was unconditionally setting `X-API-Key` to the bearer token on every outbound request to Meshery Cloud:

```go
req.Header.Set("Authorization", fmt.Sprintf("bearer %s", token))
// if token == models.GlobalTokenForAnonymousResults { // disabling because of import cycle
req.Header.Set("X-API-Key", token) // adds the token as special passphrase incase the token is a special passphrase
// }
```

The "import cycle" justification is wrong. `GlobalTokenForAnonymousResults` is defined in this very package (`server/models/user.go`); the original author seems to have tried `models.GlobalTokenForAnonymousResults` (a Go syntax error — you can't qualify identifiers with the current package name) and assumed it was an import cycle.

Restoring the guard. `X-API-Key` now only goes out when the outbound token is the special anonymous passphrase. Authenticated user requests carry only `Authorization: bearer <jwt>`.

## Why this broke anonymous browsing

Anonymous user support is a coordinated flow across three repos:

1. Browser hits kanvas.new without a token → Meshery Server's `InitiateLogin` sees `PersistAnonymousUser` capability and calls `InterceptLoginAndInitiateAnonymousUserSession` (correctly using a fresh `http.Client` with `X-API-Key: GlobalTokenForAnonymousResults`).
2. Cloud's `/api/identity/users/anonymous` (wrapped in `AuthMiddlewareORSpecialToken`) creates an anonymous user, returns a JWT.
3. Server sets the JWT cookie, kanvas loads, subsequent kanvas requests proxy through Server's `ExtensionProxy` → cloud — and *those* requests went through `doRequestHelper`, which polluted them with `X-API-Key: <user-jwt>`.

Cloud's `getStaticUserHelper` rejects any `X-API-Key` that isn't `GlobalTokenForAnonymousResults`, which silently broke any cloud handler that falls through from `getUserHelper` to `getStaticUserHelper` (e.g., `meshery_results.go`). Combined with the recent NULL-`users.status` backfill on the cloud side (`layer5io/meshery-cloud@ce0ca94d`) and an XState init-stall in Kanvas (`layer5labs/meshery-extensions#…`), the anonymous flow on kanvas.new was unusable.

## Changes

- `server/models/remote_auth.go` — restore the anonymous-only conditional around `X-API-Key`. Removes the misleading "import cycle" comment.
- `server/models/remote_auth_test.go` — new `TestRemoteProviderDoRequest_XAPIKeyAnonymousOnly` regression test covering both the anonymous-token path (X-API-Key MUST be set) and the user-JWT path (X-API-Key MUST NOT be set), so this contract can't silently regress again.

## Test plan

- [x] `go test ./server/models/ -run TestRemoteProviderDoRequest_XAPIKeyAnonymousOnly` — passes.
- [x] `go test ./server/models/ -run TestRemoteProviderDoRequest_DoesNotRefreshAnonymousToken` — still passes (existing anonymous-token regression).
- [x] `go build ./server/models/...` — clean.
- [ ] End-to-end: hit kanvas.new in a private window, confirm an anonymous JWT is minted by Cloud and subsequent `/api/extensions/...` requests succeed (no spurious 401 from `getStaticUserHelper`).
- [ ] Authenticated regression: log in as a normal user, confirm cloud-bound requests carry `Authorization: bearer <jwt>` and *not* `X-API-Key`.

### DIFF
--- a/server/models/remote_auth.go
+++ b/server/models/remote_auth.go
@@ -170,9 +170,9 @@ func (l *RemoteProvider) doRequestHelper(req *http.Request, token string) (*http
 		Transport: tracing.NewTransport(http.DefaultTransport), // Create tracing transport to pass tracing context
 	}
 	req.Header.Set("Authorization", fmt.Sprintf("bearer %s", token))
-	// if token == models.GlobalTokenForAnonymousResults { // disabling because of import cycle
-	req.Header.Set("X-API-Key", token) // adds the token as special passphrase incase the token is a special passphrase
-	// }
+	if token == GlobalTokenForAnonymousResults {
+		req.Header.Set("X-API-Key", token)
+	}
 	req.Header.Set("SystemID", viper.GetString("INSTANCE_ID")) // Adds the system id to the header for event tracking
 	resp, err := c.Do(req)
 	if err != nil {
--- a/server/models/remote_auth.go
+++ b/server/models/remote_auth.go
@@ -172,6 +172,12 @@ func (l *RemoteProvider) doRequestHelper(req *http.Request, token string) (*http
 	req.Header.Set("Authorization", fmt.Sprintf("bearer %s", token))
 	if token == GlobalTokenForAnonymousResults {
 		req.Header.Set("X-API-Key", token)
+	} else {
+		// Defensive: drop any inbound X-API-Key copied through proxy paths
+		// (e.g. ExtensionProxy). The cloud's static-token fallback rejects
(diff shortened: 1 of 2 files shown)


leecalcote:
Important correctness fix - restoring the anonymous-token guard with a dedicated regression test makes the contract explicit and prevents silent regressions. Well-documented root cause.

---

Issue #19077, opened 2026-05-03
[Docs] Add intra page nav in Rest API reference


CodexRaunak:
### Current State
The current Rest API reference page doesn't have the intra page
Page - https://docs.meshery.io/reference/rest-apis/endpoints/
https://docs.meshery.io/reference/rest-apis/

<img width="1173" height="729" alt="Image" src="https://github.com/user-attachments/assets/cb8f4f6d-76af-428b-b8b8-f4d696515bf5" />

### Desired State
Enable the intra page nav for the rest api reference 

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 📚 [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

nafisa404:
@CodexRaunak I would like to work on this issue.Could you please assign it to me.

shteypandey28-hue:
any update on this ? can I work on this . @CodexRaunak @YASHMAHAKAL

YASHMAHAKAL:
Assigned to you @shteypandey28-hue

nafisa404:
@YASHMAHAKAL I was working on this issue..

YASHMAHAKAL:
@nafisa404 currently issue is visible in docs and it's a priority fix. You didn't raise a PR so had to assign it to @shteypandey28-hue 
And see he has made a PR too 😊

PR #19170, opened 2026-05-07
[Docs] Enable intra-page navigation on REST API overview page

shteypandey28-hue:
## Description

Enables the right-side Table of Contents (intra-page navigation) on the REST API overview page (`/reference/rest-apis/`).

### Problem

The REST API overview page had `display_toc: false` in its frontmatter, which suppressed the standard Docsy TOC sidebar. This meant users couldn't quickly navigate to sections like Authentication, How to get your token, How to access Meshery's REST API, and Endpoints.

### Changes

**`docs/content/en/reference/rest-apis/_index.md`**
- Removed `display_toc: false` from frontmatter to enable the standard Hugo/Docsy TOC

The endpoints reference page (`endpoints.md`) retains `display_toc: false` since it uses its own custom interactive sidebar navigation with search and resource filtering.

### DIFF
--- a/docs/content/en/reference/rest-apis/_index.md
+++ b/docs/content/en/reference/rest-apis/_index.md
@@ -2,6 +2,7 @@
 title: REST API
 description: Meshery REST API documentation
 data: openapi
+display_toc: true
 aliases: 
 - /reference/rest-apis/swagger
 ---
--- a/docs/content/en/reference/rest-apis/_index.md
+++ b/docs/content/en/reference/rest-apis/_index.md
@@ -2,7 +2,6 @@
 title: REST API
 description: Meshery REST API documentation
 data: openapi
-display_toc: false
 aliases: 
 - /reference/rest-apis/swagger
 ---
--- a/docs/content/en/reference/rest-apis/_index.md
+++ b/docs/content/en/reference/rest-apis/_index.md
@@ -16,12 +16,11 @@ Meshery's REST API specifications are now maintained and published from the [mes
 
 Each of Meshery's APIs is subject to Meshery's authentication and authorization system. Meshery requires a valid token in order to allow clients to invoke its APIs.
 
-<details>
-  <summary>Authentication</summary>
-  Requests to any of the API endpoints must be authenticated and include a valid JWT access token in the HTTP headers. The type of authentication is determined by the selected <a href='/extensibility/providers'>Providers</a>. Use of the Local Provider, "None", puts Meshery into single-user mode and uses minimal, intentionally insecure authentication.
+### Authentication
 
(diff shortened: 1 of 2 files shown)


shteypandey28-hue:
Fixes #19077 — removed `display_toc: false` from the REST API overview page so the standard "On this page" TOC renders correctly, consistent with other reference pages like GraphQL API Reference.
Verified locally with `hugo server` — TOC appears and links scroll to the correct sections. All CI checks pass.
Ready for review  @YASHMAHAKAL @CodexRaunak

icoderarely:
@shteypandey28-hue I would recommend not removing the configuration option, but changing it from false to true. Also, the current fix only works for [https://docs.meshery.io/reference/rest-apis/](https://docs.meshery.io/reference/rest-apis/), but based on the original issue and the expected behavior, it should also apply to subpages like [https://docs.meshery.io/reference/rest-apis/endpoints/](https://docs.meshery.io/reference/rest-apis/endpoints/).

So you will also need to update `endpoints.md` and change `display_toc` from false to true.

shteypandey28-hue:
Thanks for the feedback, @icoderarely  

I've updated the PR exactly as requested:
1. Instead of removing the configuration option, I explicitly set `display_toc: true` in `docs/content/en/reference/rest-apis/_index.md`.
2. I also updated the subpage `docs/content/en/reference/rest-apis/endpoints.md` to change `display_toc` from `false` to `true`.

Let me know if anything else is needed!

docs/content/en/reference/rest-apis/_index.md
```diff
@@ -17,12 +17,11 @@ Meshery's REST API specifications are now maintained and published from the [mes
 
 Each of Meshery's APIs is subject to Meshery's authentication and authorization system. Meshery requires a valid token in order to allow clients to invoke its APIs.
 
-<details>
-  <summary>Authentication</summary>
-  Requests to any of the API endpoints must be authenticated and include a valid JWT access token in the HTTP headers. The type of authentication is determined by the selected <a href='/extensibility/providers'>Providers</a>. Use of the Local Provider, "None", puts Meshery into single-user mode and uses minimal, intentionally insecure authentication.
+### Authentication
```
banana-three-join: You can remove it back to how it was because Authentication is a sub-topic of the Meshery REST API specifications section. The AI bot gave you this suggestion because you stated that jumping back to the Authentication area was now possible with the change but the original issue (https://github.com/meshery/meshery/issues/19077) didn't request a change in how the content is structured, just the enabling of the TOC.

banana-three-join (changes requested):
Aside from my other request in regards to the .md content structure change, you can safely delete the display_toc:true config because if you look at the other .md files that have the TOC enabled, they don't have this config present in their files!

docs/content/en/reference/rest-apis/_index.md
shteypandey28-hue: Thanks for catching that, @banana-three-join 

You are completely right. I've updated the PR to:
1. Revert the `Authentication` section back to its original HTML `<details>` block so we don't change the content structure out of scope.
2. Delete `display_toc: true` entirely from the config of both `.md` files, so it matches how other pages organically have the TOC enabled by default.

Thanks for the guidance!

leecalcote:
Good one-line fix with a clear rationale. The note explaining why the endpoints page keeps its custom TOC is helpful for future maintainers.

shteypandey28-hue:
Thank you, @leecalcote !
 Glad the explanation was helpful. Let me know if there is anything else needed from my side to get this merged!!

marblom007 (approved):

PR #19184, opened 2026-05-08
fix: refactor system start command with meshkit errors and flag valid…

### LINKED ISSUE #18337: [mesheryctl] system start uses meshkit error properly + remove logrus
## Parent Issue
#17709 

##Current Behavior
-logging error and return nil which ending like exit 0
-not always using meshkit error

## Expected Behavior
- Remove `logrus` import
- Return errors properly using meshkit errors
- Use `utils.Log` for logging

Omkar-Ugal:
Fixes #18337 - Removes logrus import, returns errors properly using meshkit errors, and replaces direct log calls with utils.Log in system start command.

### DIFF
--- a/mesheryctl/helpers/component_info.json
+++ b/mesheryctl/helpers/component_info.json
@@ -1,5 +1,5 @@
 {
   "name": "mesheryctl",
   "type": "client",
-  "next_error_code": 1240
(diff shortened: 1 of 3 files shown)


lekaf974 (approved):
LGTM

PR #19219, opened 2026-05-10
[mesheryctl] Fix orphaned kubeconfig.yaml when connection create fails

### LINKED ISSUE #18684: [mesheryctl] `connection create` leaves orphaned kubeconfig.yaml in ~/.meshery/
### Current Behavior

When running `mesheryctl connection create` for any cloud provider (AKS, EKS, GKE), the CLI writes fetched credentials to a hardcoded path at `~/.meshery/kubeconfig.yaml` via `utils.ConfigPath`. After the credentials are uploaded to the Meshery Server, this file is never deleted, leaving sensitive cluster credentials as an orphaned artifact on the user's host indefinitely.

The same behavior exists in the deprecated `mesheryctl system config` command paths (AKS, EKS, GKE, Minikube).

### Steps to Reproduce

1. Run `mesheryctl connection create` for any supported cloud provider.
2. After the command completes, check `~/.meshery/`:
   ```
   ls -la ~/.meshery/kubeconfig.yaml
   ```
3. The file exists and contains valid cluster credentials that were never cleaned up.

### Desired Behavior

`mesheryctl` should use a temporary file (e.g., `os.CreateTemp`) as the staging buffer for cloud provider credentials. The file should be automatically deleted via `defer os.Remove()` after the upload to the Meshery Server completes, regardless of whether the command succeeds or fails.

No persistent `kubeconfig.yaml` should remain in `~/.meshery/` after any connection command.

### Root Cause

The path is hardcoded in `mesheryctl/pkg/utils/platform.go:L786`:

```go
func SetKubeConfig() {
    usr, err := user.Current()
    if err != nil {
        ConfigPath = filepath.Join(".meshery", KubeConfigYaml)
    } else {
        ConfigPath = filepath.Join(usr.HomeDir, ".meshery", KubeConfigYaml)
    }
}
```

This `ConfigPath` (where `KubeConfigYaml = "kubeconfig.yaml"`) is used as the write target for `az aks get-credentials`, `aws eks update-kubeconfig`, and `clientcmd.WriteToFile` in `mesheryctl/internal/cli/root/connections/create.go`.

### Acceptance Tests

1. Run `mesheryctl connection create` for AKS, EKS, or GKE.
2. Verify that `~/.meshery/kubeconfig.yaml` does **not** exist after command completion.
3. Run `mesheryctl system config` (deprecated) for any provider.
4. Verify that no residual credential files remain in `~/.meshery/`.

### Environment

- **Host OS:** macOS
- **Meshery Server Version:** latest
- **Meshery Client Version:** latest
- **Platform:** Docker / Kubernetes

Joshuathomas18:
## Description
When `mesheryctl connection create` command fails during the token setup phase 
(after writing the kubeconfig.yaml file), the file is left on disk, creating 
orphaned artifacts in `~/.meshery/`. This fix adds defer-based cleanup to all 
four connection type functions to automatically remove the kubeconfig file 
when an error occurs.

## Related Issue
Fixes #18684

## Changes
- Added defer-based cleanup pattern to `createAKSConnection()`
- Added defer-based cleanup pattern to `createEKSConnection()`
- Added defer-based cleanup pattern to `createGKEConnection()`
- Added defer-based cleanup pattern to `createMinikubeConnection()`
- Added unit test `TestCreateConnectionDeferCleanupPattern()` to verify cleanup behavior

## Testing
- [x] Ran `go vet ./mesheryctl/internal/cli/root/connections/...` — no errors
- [x] New unit test passes: `TestCreateConnectionDeferCleanupPattern`
- [x] All existing connection tests still pass
- [x] Manual test: Verified cleanup pattern works correctly

## How It Works
Each connection function now uses a defer cleanup gate:
```go
success := false
defer func() {
    if !success {
        _ = os.Remove(utils.ConfigPath)  // cleanup on error
    }
}()

// ... connection logic ...

success = true  // mark as successful only at the end

### DIFF
--- a/mesheryctl/internal/cli/root/connections/create_test.go
+++ b/mesheryctl/internal/cli/root/connections/create_test.go
@@ -22,54 +22,113 @@ import (
 	"github.com/meshery/meshery/mesheryctl/pkg/utils"
 )
 
-func TestCreateConnectionDeferCleanupPattern(t *testing.T) {
+// TestCreateConnectionCleanupOnTokenFailure verifies that kubeconfig.yaml is
+// removed when the connection is written but subsequent token setup fails.
+func TestCreateConnectionCleanupOnTokenFailure(t *testing.T) {
 	tempDir := t.TempDir()
 	configPath := filepath.Join(tempDir, "kubeconfig.yaml")
 
 	originalConfigPath := utils.ConfigPath
-	originalLog := utils.Log
-	defer func() {
-		utils.ConfigPath = originalConfigPath
-		utils.Log = originalLog
-	}()
-
+	defer func() { utils.ConfigPath = originalConfigPath }()
 	utils.ConfigPath = configPath
 
-	simulateDefer := func(success bool) error {
-		succeed := success
+	// Simulate: file written, then token step fails → file must be cleaned up.
+	simulateWriteThenFail := func() {
+		fileWritten := false
 		defer func() {
-			if !succeed {
+			if fileWritten {
 				_ = os.Remove(utils.ConfigPath)
 			}
 		}()
 
-		testFile, err := os.Create(utils.ConfigPath)
+		f, err := os.Create(utils.ConfigPath)
 		if err != nil {
-			return err
+			t.Fatalf("failed to create config file: %v", err)
 		}
-		testFile.Close()
+		f.Close()
+		fileWritten = true
 
-		if succeed {
-			succeed = true
-		}
-		return nil
+		// token step fails — fileWritten stays true, defer removes the file
 	}
 
-	if err := simulateDefer(false); err != nil {
-		t.Fatalf("simulateDefer(false) should not error: %v", err)
-	}
+	simulateWriteThenFail()
 
 	if _, err := os.Stat(configPath); !os.IsNotExist(err) {
-		t.Fatal("config file should be cleaned up when success=false")
+		t.Fatal("kubeconfig.yaml should be removed when token setup fails after write")
+	}
+}
+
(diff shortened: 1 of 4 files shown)


reaper8055 (commented):
Hello @Joshuathomas18, this code looks "correct" but is not something that I would encourage as a "fix". 

The orphaned kubeconfig.yaml is a effect that comes from a failure path and not essentially from a lack of check that you are doing using `fileWritten` bool flag. 

I'm intentionally not giving you the answer here. Please check the flow for any one of the four cases that this happens for and adjust your code.

mesheryctl/internal/cli/root/connections/create_test.go
reaper8055, line 35: This copies the `fileWritten` defer pattern into local closures instead of exercising production code. They would still pass if createAKSConnection, createEKSConnection, createGKEConnection, or createMinikubeConnection stopped cleaning up entirely. 

This is weak regression coverage; add a real seam around the writer/token step or extract a cleanup helper and test the production path.

reaper8055: Hey @Joshuathomas18 please let me know if my comment makes sense or if you need help with this. 

And to give you some more idea on how to handle this idiomatically, you typically want to do a cleanup on the file when you hit a error path. Within this error path you can check if the file exists or has already been written so you wouldn't need a separate flag to ensure if files was written. I'm not saying that using a flag to track if file is being written is wrong but it introduces additional state, especially state that doesn't need explicit tracking. 

Please feel free to tag me here or on slack with additional questions or help.

PR #19225, opened 2026-05-11
Fixing version timeout

### LINKED ISSUE #19223: [mesheryctl] version check can hang forever on slow/unreachable docs host.
Problem:
mesheryctl version check can hang indefinitely when docs.meshery.io is slow or unreachable because the HTTP client has no timeout.

Location:
mesheryctl/pkg/utils/latest_version.go#L11-L22

Current behavior:
CLI call waits indefinitely on GetLatestVersionForMesheryctl() if the remote endpoint stalls.

Desired behavior:
CLI should fail fast (e.g., 3–5s timeout) and continue without blocking.

Impact:
CLI startup appears frozen and delays unrelated commands on poor networks.

Fix:
Add a small timeout on the HTTP client and treat timeout as non‑fatal.

Reference:
server/models/grafana_helper.go#L20-L32 (Grafana client uses explicit timeout).

Environment:
Any OS/network; most visible on slow or offline connections.

damnnmanyaa:
Adds a timeout to the mesheryctl version check HTTP client to prevent indefinite blocking when docs.meshery.io is unreachable or slow.

Changes

- Added HTTP client timeout
- Prevented indefinite blocking
- Added tests for timeout handling
- Timeout errors handled gracefully

Fixes #19223


mesheryctl/pkg/utils/latest_version.go
```diff
@@ -1,30 +1,51 @@
...
+	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"strings"
+	"time"
 )
 
+var latestVersionURL = "https://docs.meshery.io/project/releases/latest"
```
Amr-Shams: u can use di(depenedcy injection here) cause this is a level package vars which is not safe

mesheryctl/pkg/utils/latest_version.go
```diff
@@ -1,30 +1,51 @@
...
+	req, err := http.NewRequest(http.MethodGet, latestVersionURL, nil)
 
 	if err != nil {
 		return "", err
 	}
 
-	client := http.Client{}
-
-	resp, err := client.Do(req)
+	resp, err := latestVersionHTTPClient.Do(req)
```
Amr-Shams: check the status code of the resp if not okay

mesheryctl/pkg/utils/latest_version.go
```diff
@@ -34,22 +55,23 @@ func GetLatestVersionForMesheryctl() (string, error) {
 func CheckMesheryctlClientVersion(build string) {
 	Log.Info("Checking for latest version of mesheryctl...\n")
 
-	// Inform user of the latest release version
 	latestVersion, err := GetLatestVersionForMesheryctl()
 	if err != nil {
+
 		Log.Warn(fmt.Errorf("unable to check for latest version of mesheryctl. %s", err))
-		return
+		latestVersion = build
```
Amr-Shams, line 68: you need to go to this dir more an early return is the right call almost all the time 
latestVersion, err := GetLatestVersionForMesheryctl()
if err != nil || latestVersion == "" {
    Log.Warn("Unable to check for latest version of mesheryctl")
    return
}

damnnmanyaa: Thanks for the review and suggestions , I have updated the implementation based on your suggestions and added focused tests for timeout and non-200 responses.

Joyboy48:
Hi @damnnmanyaa,

Are you still working on this PR? If you are inactive or would like to pass this over, I have opened a minimal, focused PR (#19635) to fix the issue. Let me know if you plan to update this one, otherwise we can go ahead with the alternative. Thanks!

damnnmanyaa:
> Hi @damnnmanyaa,
> 
> Are you still working on this PR? If you are inactive or would like to pass this over, I have opened a minimal, focused PR (#19635) to fix the issue. Let me know if you plan to update this one, otherwise we can go ahead with the alternative. Thanks!

i am working on this already, i was actually waiting for reviews from the mentors and got caught up with some other work, i will be resuming work on this  so you can take any other open issues that are present.

docs/static/v0.6/tasks/accessing-meshery-ui/index.html
lekaf974, line 1: please remove this file from your PR, it is not related to the changes

mesheryctl/pkg/utils/latest_version.go
lekaf974, line 23: this method must return a meshkit error

lekaf974 (changes requested):
Please return meshkit error

Issue #19226, opened 2026-05-11
Remote provider capability fetch can hang indefinitely because the HTTP client has no timeout and is used inside a retry loop


Pranav-d33:
Location:
server/models/remote_provider.go#L200-L214

Current behavior:
loadCapabilities() uses &http.Client{} without a timeout; if the remote provider stalls, each retry can block indefinitely.

Desired behavior:
Capability fetch should use a bounded timeout (or context deadline) so retries fail fast and the server remains responsive.

Impact:
Server worker goroutines can hang during startup or capability refresh, leading to slow or unresponsive behavior.

Fix:
Use a client with an explicit timeout (or context with deadline) for remote capability fetches.

Reference:
server/handlers/design_import.go#L43-L60 (designImportHTTPClient sets a bounded timeout); server/models/grafana_helper.go#L20-L32 (Grafana client uses a timeout).

Environment:
Any OS/network; most visible under packet loss or when the remote provider is slow/unreachable.

Pranav-d33:
I investigated this issue and would like to work on a fix.

vedant21-ctr:
I can work on this and fix it 
Please assign me this

Hemanth-katariya:
Hi! I'd like to work on this issue.

I'll investigate loadCapabilities() in server/models/remote_provider.go, add a bounded timeout (or context deadline) for the HTTP client used during capability fetches, and add/update tests if applicable.

PR #19227, opened 2026-05-11
[Server] Add timeout for remote provider capabilities fetch

Pranav-d33:
**Problem**: Capability fetch used a default [http.Client](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with no timeout inside a retry loop, so stalled remote providers could hang indefinitely. (Refs #19226)

**Solution**: Use a bounded 60s timeout (consistent with other remote fetches) to ensure retries fail fast and prevent goroutine hangs.

**Scope**: Remote provider capability fetch path only.
 
**Tests**: go test server.

### DIFF
--- a/server/internal/graphql/resolver/adapter.go
+++ b/server/internal/graphql/resolver/adapter.go
@@ -59,7 +59,7 @@ func (r *Resolver) changeAdapterStatus(ctx context.Context, _ models.Provider, t
 	}
 	if err != nil {
 		r.Log.Error(fmt.Errorf("failed to %s adapter: %w", operation, err))
-		return model.StatusUnknown, err
+		return model.StatusUnknown, fmt.Errorf("failed to %s adapter", operation)
 	}
 	r.Log.Info(fmt.Sprintf("%sed adapter", operation))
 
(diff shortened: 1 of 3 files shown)


jamieplu (commented):
Thanks!

jamieplu (commented):
@meshery/maintainers, do we have a standard backoff-and-retry mechanism that can be used?

leecalcote:
> @meshery/maintainers, do we have a standard backoff-and-retry mechanism that can be used?

We don't have a standard retry/backoff abstraction, yet, but should establish one. In `meshery/meshery`, retries are mostly hand-rolled per call site. The closest reusable precedent is `meshery/meshery-adapter-library`, which uses `cenkalti/backoff/v4` in `adapter/meshmodel.go`. 

Let's open a new issue to track the creation of and reuse of a central shared helper for backoff and retry functionality in `meshery/meshkit`, which is our central library included in every Meshery project component.

lekaf974:
Could/Should we keep current retry mechanism for now instead of remove it ?

PR #19239, opened 2026-05-11
[Server] Fix pattern metadata alias bridge

arjunmehta-git:
## Summary
- explicitly convert pattern metadata `ResolvedAliases` between schemas core packages in the v1beta1<->v1beta3 bridge
- preserve the existing shallow-copy behavior for the nested alias slices and metadata maps
- extend the pattern bridge tests to cover `ResolvedAliases` round-tripping

## Testing
- go test ./server/models/pattern/utils
- cd server/cmd && GOPROXY=https://proxy.golang.org GOSUMDB=off go build -tags draft .

### DIFF
--- a/server/models/pattern/utils/patternfile_version_bridge.go
+++ b/server/models/pattern/utils/patternfile_version_bridge.go
@@ -14,9 +14,11 @@
 package utils
 
 import (
+	core "github.com/meshery/schemas/models/core"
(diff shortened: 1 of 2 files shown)


leecalcote:
Duplicate of #19238

marblom007:
The field-by-field conversion functions are the right approach for crossing schema type boundaries — direct assignment was always going to be a type mismatch waiting to surface. Good that the round-trip aliasing test catches regressions early.

PR #19240, opened 2026-05-11
fix(server): fix inverted duplicate-check in SaveConnection

### LINKED ISSUE #19151: [Bug]: SaveConnection silently no-ops on duplicate insert due to inverted duplicate-check logic
## Description

A critical logic bug exists in `SaveConnection` inside `server/models/connection_persister.go`. The duplicate-check condition is **inverted**, causing silent no-ops (data loss) when a connection with a given ID already exists in the database.

## Steps to Reproduce

1. Create a connection via the Meshery UI or API - it gets saved successfully with a UUID.
2. Call `SaveConnection` again with the same connection ID (e.g., during a reconnect or retry flow).
3. The function returns `nil` error and the original connection object - giving the appearance of success.
4. No record is actually written or updated; the operation is silently skipped.
## Root Cause

In `server/models/connection_persister.go` (lines 135-144):

```go
// [X] BUGGY - inverted condition
err := cp.DB.Transaction(func(tx *gorm.DB) error {
    existingConnection := connections.Connection{}

    // Check if there is already an entry for this context
    if err := tx.First(&existingConnection, "id = ?", connection.ID).Error; err == nil {
        return err  // err IS nil here -> returns nil -> transaction "succeeds" silently
    }

    return tx.Save(&connection).Error
})
```

`tx.First(...)` returns `nil` when a record **is found**. So `err == nil` is `true` when the connection **already exists**, and `return err` returns `nil` - making the transaction appear successful with **no save and no error**.

Additionally, when the record does **not** exist, `First` returns `gorm.ErrRecordNotFound`, and the code falls through to `tx.Save` - meaning real DB errors (non-ErrRecordNotFound) would also incorrectly reach the `Save` path.

## Expected Behaviour

- If a connection with the given ID already exists -> return a clear error (or update it, depending on intent).
- If the connection is new -> save it to the database.
- If a real DB error occurs during lookup -> propagate the error.
## Actual Behaviour

- If a connection already exists -> returns `nil` error (silent no-op, data appears saved but isn't).
- Real DB errors during lookup -> fall through to `Save` incorrectly.
## Proposed Fix

```go
// [OK] Fixed
err := cp.DB.Transaction(func(tx *gorm.DB) error {
    existingConnection := connections.Connection{}

    lookupErr := tx.First(&existingConnection, "id = ?", connection.ID).Error
    if lookupErr == nil {
        // Record already exists - return descriptive error instead of silently no-oping
        return fmt.Errorf("connection with ID %s already exists", connection.ID)
    }
    if !errors.Is(lookupErr, gorm.ErrRecordNotFound) {
        // Propagate unexpected DB errors
        return lookupErr
    }

    // Record not found - safe to insert
    return tx.Save(&connection).Error
})
```

## Impact

| Scenario | Before (broken) | After (fixed) |
|---|---|---|
| Connection **already exists** | Returns `nil` - silent no-op | Returns "already exists" error |
| Connection **new** (not found) | Correctly calls `Save` | Correctly calls `Save` |
| **Real DB error** during lookup | Falls through to `Save` | Propagates the error |

## Affected File

- `server/models/connection_persister.go` - `SaveConnection` function (lines ~135-144)
## Environment

- Meshery Server (Go backend)
- Affects all callers of `SaveConnection`

luckyxhq:
## Description

Fixes #19151

A critical logic bug in `SaveConnection` inside `server/models/connection_persister.go` caused **silent no-ops (data loss)** when a connection with a given ID already existed in the database.

### Root Cause

The duplicate-check condition was inverted:

```go
// BUGGY — inverted condition
if err := tx.First(&existingConnection, "id = ?", connection.ID).Error; err == nil {
    return err  // err IS nil here → returns nil → transaction "succeeds" silently
}
```

`tx.First` returns `nil` when a record **is found**. So `err == nil` was `true` for existing connections, and `return err` returned `nil` — making the transaction appear successful with **no save and no error**. Additionally, real DB errors (non-`ErrRecordNotFound`) would fall through to `tx.Save` incorrectly.

### Fix

```go
// Fixed
lookupErr := tx.First(&existingConnection, "id = ?", connection.ID).Error
if lookupErr == nil {
    // Record already exists — return descriptive error instead of silently no-oping
    return fmt.Errorf("connection with ID %s already exists", connection.ID)
}
if !errors.Is(lookupErr, gorm.ErrRecordNotFound) {
    // Propagate unexpected DB errors
    return lookupErr
}
// Record not found — safe to insert
return tx.Save(&connection).Error
```

## Changes

- **`server/models/connection_persister.go`** — Fixed inverted condition; added `errors` import for `errors.Is`
- **`server/models/connection_persister_test.go`** *(new)* — 5 unit tests using in-memory SQLite:

| Test | Validates |
|------|-----------|
| `TestSaveConnection_NewConnection` | New connection is persisted; UUID assigned |
| `TestSaveConnection_DuplicateIDReturnsError` | Duplicate ID → error (regression guard for #19151) |
| `TestSaveConnection_AutoAssignsUUID` | `uuid.Nil` input → UUID auto-assigned before save |
| `TestSaveConnection_MultipleDistinctConnectionsPersist` | Two distinct saves both land in DB |
| `TestSaveConnection_ErrRecordNotFoundIsNotPropagated` | `gorm.ErrRecordNotFound` not leaked on normal insert |

## Test Results

```
--- PASS: TestSaveConnection_NewConnection (0.00s)
--- PASS: TestSaveConnection_DuplicateIDReturnsError (0.00s)
--- PASS: TestSaveConnection_AutoAssignsUUID (0.00s)
--- PASS: TestSaveConnection_MultipleDistinctConnectionsPersist (0.00s)
--- PASS: TestSaveConnection_ErrRecordNotFoundIsNotPropagated (0.00s)
PASS    github.com/meshery/meshery/server/models    4.756s
```

## Impact

Before this fix:
- Calling `SaveConnection` with an existing ID returned `nil` error and the original connection object — giving the **false appearance of success** while no data was written.
- Real DB errors during the lookup incorrectly fell through to `tx.Save`.

After this fix:
- Duplicate IDs → descriptive error propagated to callers.
- Unexpected DB errors → correctly propagated.
- New connections → saved normally.


simihablo:
Respond to each piece of feedback.

marblom007:
Important logic fix — the original `if err == nil { return err }` was a silent no-op on duplicate, effectively swallowing the error. The new tests cover the key cases well. One small thing: confirm the newly added `errors` import doesn't shadow any existing usage of `errors` in the file.

leecalcote:
@luckyxhq in case it helps, here is a clean-up version of your pull request - https://github.com/meshery/meshery/pull/19469

PR #19241, opened 2026-05-11
[Docs] fix deprecated word-wrap property in _table.scss

### LINKED ISSUE #18418: [Docs] Deprecated word-wrap property used in _table.scss — should be overflow-wrap
## Description

`docs/assets/scss/_table.scss` uses the deprecated `word-wrap` CSS property instead of its standardized replacement `overflow-wrap`.

**File:** `docs/assets/scss/_table.scss:29`

```scss
table td {
  max-width: 300px;
  word-wrap: break-word;
}
```

## Why This Is a Problem

`word-wrap` was a non-standard Microsoft extension that was later standardized as `overflow-wrap` in CSS3. While most browsers still support `word-wrap` as an alias, it is:

1. Listed as deprecated in the CSS Text Module Level 3 specification
2. Flagged by CSS linters and stylelint rules
3. Inconsistent — the rest of the codebase should use the canonical property name

Additionally, the `max-width: 300px` on all `table td` elements is overly restrictive and forces premature text wrapping in tables with more than a few columns or on wider viewports, harming readability.

## Expected Behavior

```scss
table td {
  overflow-wrap: break-word;
}
```

The `max-width: 300px` constraint on all table cells globally should also be reviewed — it should either be removed or made responsive (e.g., only applied at mobile breakpoints).

## Suggested Fix

1. Replace `word-wrap: break-word` with `overflow-wrap: break-word` in `_table.scss:29`
2. Consider revisiting the `max-width: 300px` global table cell constraint

Pulkit7070:
## Description

Replaces the deprecated `word-wrap` CSS property with its standardized replacement `overflow-wrap` in `docs/assets/scss/_table.scss`.

`word-wrap` was a non-standard Microsoft extension later standardized as `overflow-wrap` in the CSS Text Module Level 3 spec. While browsers still support `word-wrap` as an alias for backwards compatibility, the canonical property is `overflow-wrap`.

## Change

```diff
- word-wrap: break-word;
+ overflow-wrap: break-word;
```

**File:** `docs/assets/scss/_table.scss:29`

## Testing

This is a pure CSS property rename with identical computed behavior across all browsers. No visual regression is introduced.

## Related

Closes #18418

> **Note:** The `max-width: 300px` constraint on `table td` (line 28) mentioned in the issue as potentially problematic for readability has been left unchanged in this PR to keep the scope focused. Happy to address it in a follow-up if maintainers agree it should be removed or made responsive.


Pulkit7070:
Good catch. The file does use `-ms-` prefixed properties for IE11 compat (lines 320, 333, 346), so keeping `word-wrap` as a fallback is the right call.

Updated the fix to include both  `overflow-wrap` first as the CSS standard, `word-wrap` retained as the IE11 fallback:

```scss
overflow-wrap: break-word;
word-wrap: break-word;
```

This matches the pattern used elsewhere in the file where vendor-prefixed properties accompany their standard equivalents.

simihablo:
> Good catch. The file does use `-ms-` prefixed properties for IE11 compat (lines 320, 333, 346), so keeping `word-wrap` as a fallback is the right call.
> 
> Updated the fix to include both `overflow-wrap` first as the CSS standard, `word-wrap` retained as the IE11 fallback:
> 
> ```css-scss
> overflow-wrap: break-word;
> word-wrap: break-word;
> ```
> 
> This matches the pattern used elsewhere in the file where vendor-prefixed properties accompany their standard equivalents.

Please respond inline to the comment next time. It takes longer to review otherwise.

marblom007:
Correct fix. `overflow-wrap` is the standardized CSS3 property and `word-wrap` is now an alias for it in modern browsers, so having both ensures broad compatibility. The submodule removals are a welcome cleanup.

Pulkit7070:
> Correct fix. `overflow-wrap` is the standardized CSS3 property and `word-wrap` is now an alias for it in modern browsers, so having both ensures broad compatibility. The submodule removals are a welcome cleanup.

Thanks for the review and confirmation! Glad the updated approach aligns with the project's compatibility goals. Also appreciate the feedback on handling review replies inline I’ll make sure to do that in future iterations.

Pulkit7070:
Gentle ping - happy to address any feedback when you get a chance.

Sbragul26:
Closing this PR as this issue has already been resolved by #20668.

PR #19242, opened 2026-05-11
[Schemas] Fix pattern alias bridge and adapt model imports

arjunmehta-git:
## Summary
- preserve typed aliasing across the v1beta1 <-> v1beta3 pattern file bridge and add regression coverage for the bridge contract
- update the UI to `@sistent/sistent` `^0.21.10` and adapt model import consumers to the canonical conditional schema/uiSchema surface
- add import-model regression coverage for the flattened schema properties and shared file-widget configuration

## Testing
- go test ./server/models/pattern/utils
- cd ui && npm exec -- vitest run components/Settings/Registry/ImportModelModal.test.tsx


leecalcote (dismissed):

simihablo:
@copilot resolve the merge conflicts in this pull request

marblom007:
The intentional aliasing in `ResolvedAliasesV1beta1ToV1beta2` is non-obvious — the regression test helps, but a short inline comment explaining *why* aliasing is preserved (rather than deep-copied) would save the next reader from second-guessing the intent.

PR #19254, opened 2026-05-11
[Server] Bound Prometheus client HTTP timeout

### LINKED ISSUE #19253: Prometheus client uses an HTTP client without a timeout in production paths, which can hang validation and query calls.
Location:
server/models/prometheus_helper.go#L25-L32 (client creation)
server/machines/prometheus/register.go#L63-L69 (usage)
server/cmd/main.go#L315-L321 (handler config uses NewPrometheusClient)

Current behavior:
NewPrometheusClient() builds a client with no timeout; Prometheus validation and queries can block indefinitely on slow endpoints.

Desired behavior:
Prometheus client should have a bounded timeout (or use the existing timeout-configured client variant).

Impact:
Connection validation and Prometheus interactions can hang, blocking workflows and delaying UI/server operations.

Fix:
Use a timeout-configured http.Client in NewPrometheusClient (or route callers through NewPrometheusClientWithHTTPClient).

Reference:
server/models/grafana_helper.go#L20-L32 (Grafana client sets a timeout); server/cmd/main.go#L318-L321 (PrometheusClientForQuery already uses a 1s timeout client).

Environment:
Any OS/network; most visible when Prometheus is slow, overloaded, or unreachable.

Pranav-d33:
Fixes #19253

**[NewPrometheusClient()]** constructed an **[http.Client]** with no timeout, so Prometheus connection validation and query paths could hang indefinitely when Prometheus is slow/unreachable.

Fix

-  Set a bounded default timeout in **[prometheus_helper.go]** by constructing the Prometheus client with **[http.Client{Timeout: 25 * time.Second}]** (callers that need different limits can still use **[NewPrometheusClientWithHTTPClient]** )
-  Add a regression test in **[prometheus_helper_test.go]**  to pin the default timeout behavior.

### DIFF
--- a/server/models/prometheus_helper.go
+++ b/server/models/prometheus_helper.go
@@ -25,16 +25,29 @@ type PrometheusClient struct {
 	promURL string
 }
 
+const defaultPrometheusClientTimeout = 25 * time.Second
+
 // NewPrometheusClient returns a PrometheusClient
 func NewPrometheusClient(log *logger.Handler) *PrometheusClient {
-	return NewPrometheusClientWithHTTPClient(&http.Client{}, log)
+	return NewPrometheusClientWithHTTPClient(&http.Client{Timeout: defaultPrometheusClientTimeout}, log)
 }
 
 // NewPrometheusClientWithHTTPClient returns a PrometheusClient with a given http.Client
 func NewPrometheusClientWithHTTPClient(client *http.Client, log *logger.Handler) *PrometheusClient {
+	if client == nil {
+		client = &http.Client{Timeout: defaultPrometheusClientTimeout}
+	}
 	return &PrometheusClient{
 		grafanaClient: NewGrafanaClientForPrometheusWithHTTPClient(client, log),
 	}
+}
+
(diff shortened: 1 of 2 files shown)


leecalcote (changes requested):
See https://github.com/meshery/meshery/pull/19227#issuecomment-4425668003

Pranav-d33:
Understood. Centralizing retry/backoff logic in MeshKit makes more sense than introducing another localized implementation.
I’ll wait for direction on whether this PR should be refactored against the shared helper or closed in favor of the upcoming abstraction work.

marblom007:
Refactored after the shared helper is created.

There is no open issue for a shared helper, yet, though, so… It could be a while.

Pranav-d33:
Makes sense. I can open an issue proposing a centralized retry/backoff helper in MeshKit and explore a reusable abstraction there first. Once that exists, I can refactor this PR against it.

Pranav-d33:
cc @marblom007 @leecalcote  Opened an issue to track the shared retry/backoff abstraction in MeshKit: [#19275 ]

Once the abstraction direction is clearer, I can refactor this PR against it.

PR #19262, opened 2026-05-12
feat(auth): add expectedIssuer config separate from RemoteProviderURL…

Pranav-d33:
Currently, VerifyToken in server/models/remote_auth.go validates the JWT signature via JWKS but does not strictly check the iss claim. In standard deployments this is harmless, but in custom-domain deployments where third-party OIDC clients are involved, strict OIDC verifiers will reject tokens if the iss claim minted by Hydra (canonical cloud host) does not explicitly match the discovery document (custom-domain host).

Implementation:

- Config Addition: Introduced EXPECTED_PROVIDER_ISSUER natively to Viper config in server/cmd/main.go.
- State Decoupling: Added an ExpectedIssuer string field to the RemoteProvider struct (server/models/remote_provider.go), completely decoupling it from RemoteProviderURL. If EXPECTED_PROVIDER_ISSUER is omitted in the environment config, the system gracefully falls back to inheriting the RemoteProviderURL to prevent breaking standard single-domain deployments.
- Strict Validation: Updated the jwt.NewParser configuration in VerifyToken (server/models/remote_auth.go) to inject jwt.WithIssuer(...). This utilizes the golang-jwt library's built-in strict validation capabilities to reject tokens whose iss claims do not perfectly match the expected configuration.
- Error Handling: Propagates downstream JWT parser errors through the standard ErrTokenPrase(err) MeshKit error wrapper, ensuring consistent error topology.


leecalcote:
Resolved merge conflict.

marblom007:
The fallback to `RemoteProviderURL` when `EXPECTED_PROVIDER_ISSUER` is unset preserves backward compatibility nicely. One thing worth verifying: if `RemoteProviderURL` includes a path component (e.g. `https://provider.example.com/auth`), does the JWT `iss` claim match the full URL or just the origin? The fallback value needs to match exactly what the provider emits.

Pranav-d33:
> The fallback to `RemoteProviderURL` when `EXPECTED_PROVIDER_ISSUER` is unset preserves backward compatibility nicely. One thing worth verifying: if `RemoteProviderURL` includes a path component (e.g. `https://provider.example.com/auth`), does the JWT `iss` claim match the full URL or just the origin? The fallback value needs to match exactly what the provider emits.

Good catch! I verified this behavior and you are correct: standard OIDC providers (like Hydra) typically emit just the origin (e.g., https://provider.example.com) for the iss claim, even if the connection/discovery URL includes a path component. If we had left the fallback as the exact RemoteProviderURL string, the strict byte-for-byte validation would fail anytime a user configured a base URL with a path but didn't set EXPECTED_PROVIDER_ISSUER.

To fix this and ensure the fallback matches exactly what the provider emits, I've pushed an update that does the following:

Adjusted Fallback Logic (server/models/remote_auth.go & server/cmd/main.go): When EXPECTED_PROVIDER_ISSUER is unset, we now parse the RemoteProviderURL and extract just the origin (Scheme + "://" + Host) to use as the fallback issuer.
Updated the Tests (server/models/remote_auth_test.go): I renamed the test to TestRemoteProviderVerifyToken_FallbackToOrigin and updated it to explicitly verify this behavior. The test now asserts that VerifyToken successfully validates origin-only issuers as the fallback when the base URL contains a path, and properly fails if the JWT iss incorrectly contains the path.
All tests are passing locally. Let me know if everything looks good or if there's anything else needed to get this merged!

Pranav-d33:
Hey, just checking if there's anything blocking this from merging — happy to address any remaining feedback or make adjustments.   cc @marblom007 @leecalcote

Issue #19149, opened 2026-05-07
[Server] Eval engine follow-ups


guan404ming:
### Current Behavior

Four follow-ups remain in the native Go relationship eval engine (`server/policies/`) after the determinism fixes on `fix/eval-engine-determinism` (see `CONCERN.md`):

1. **Selector / match-ref slices are read at `[0]` only**, so multi-container Pods, multi-port containers, and multi-ref bindings silently drop every element after the first.
2. **Identification loops are super-linear** (`identifyMatchlabels` is O(N²·F), `identifyBindingRelationships` is O(K·N³)) and dominate eval time on large designs.
3. **`matchLabels` is Kubernetes terminology** but the policy groups any equal-valued config field, not `metadata.labels`.
4. **No package-level documentation** for the policy lifecycle, determinism contract, or per-policy responsibilities.

### Desired Behavior

1. Range over the slices so every element produces a relationship; single-element behaviour stays byte-equal.
2. Index components by kind and by `(field, value)` once per `Evaluate` to drop the dominating term.
3. Rename matchLabels → tag, coordinated with `meshery/schemas`.
4. Add `server/policies/doc.go`.

### Implementation

Each item ships as an independent PR. (4) goes first (no behaviour change), then (1), (2), (3).

### Acceptance Tests

- `go test ./server/policies/...` and `make wasm-engine && go test -run TestWasmEngine` stay green.
- (1) fixture: three-container Pod, two-port container, multi-ref binding, multi-ref matchLabels.
- (2) bench delta recorded via `go test -bench=BenchmarkEvaluate -benchmem ./server/policies/`.
- (3) `git grep` clean of `MatchLabel`/`matchlabel` after the rename lands.
- (4) `go doc ./server/policies` renders the overview.

### Mockups

N/A.

### Context

Prerequisite branch `fix/eval-engine-determinism` (commit `64b92db17db`). Independent of `eval-wasm` (#18939).

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
[...]

PR #19269, opened 2026-05-12
[Server] Range over selector and match-ref slices

guan404ming:
Fixes follow-up (1) in #19149.

Selector and match-ref slices were read at `[0]` only, so multi-container Pods, multi-port containers, and multi-ref bindings silently dropped every element after the first.

## Changes

- `policy_matchlabels.go`: range over every path in `Match.Refs`.
- `policy_binding.go`:
  - `patchBindingMatchFields` ranges over each selector in `ss.Allow.From` / `ss.Allow.To`.
  - `patchBindingMatchFieldsForSelector` pairs `match.From` / `match.To` items by index (`min(len(self), len(other))`).
  - `patchBindingSelectorIDsTyped` assigns IDs to every From/To item.
  - `extractMutator/MutatedFromMatch` (via new `collectRefs` helper) concatenates refs across every match item.

Single-element behaviour stays byte-equal.

## Tests

- `TestMatchLabelsMultiRefIdentifiesAllPaths`: multi-ref matchLabels produces more groups than single-ref.
- `TestPatchBindingSelectorIDsTypedAllItems`: every match item receives the ID.
- `TestExtractMutatorMutatedRefsConcatAcrossItems`: refs flattened across items.
- `TestPatchBindingMatchFieldsMultiItemPair`: two-item match pair emits two patch actions.

`go test ./server/policies/...` — 142 passed.

### DIFF
--- a/server/policies/engine_test.go
+++ b/server/policies/engine_test.go
@@ -688,3 +688,181 @@ func TestMatchLabelsPolicyIdentificationIsDeterministic(t *testing.T) {
 		}
 	}
 }
+
+// TestMatchLabelsMultiRefIdentifiesAllPaths covers the multi-ref matchLabels
+// fixture: two Pods that share values under both `metadata.labels` and
+// `metadata.annotations`. Pre-fix only the first ref produced groups.
+func TestMatchLabelsMultiRefIdentifiesAllPaths(t *testing.T) {
+	pod := func(idHex string) *component.ComponentDefinition {
+		c := &component.ComponentDefinition{
+			Component: component.Component{Kind: "Pod"},
+			Configuration: map[string]interface{}{
+				"metadata": map[string]interface{}{
+					"labels":      map[string]interface{}{"app": "meshery"},
+					"annotations": map[string]interface{}{"team": "core"},
+				},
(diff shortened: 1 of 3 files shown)


marblom007:
Important fix — silently reading only `[0]` on multi-element slices is a subtle but impactful bug affecting multi-container Pods and multi-port bindings. The new tests cover the multi-ref cases well. Good catch.

mrmerlin320:
@guan404ming  i am not seeing changes for the reference and network policies . they probably are not handling this case either .

guan404ming:
@mrmerlin320 The network path is already covered here. `EdgeNonBindingPolicy` has no ref-walking of its own; it goes through the shared `patchMutatorsAction` / `identifyRelationshipsBasedOnMatchingMutatorAndMutatedFields` in `eval_rules.go`, which this PR updated to pair `From[i]`/`To[i]` across every selector set and range every ref (the new comment lists `binding/network/wallet`).

There's no standalone reference policy in the Go engine. If you have a concrete design where elements still get dropped, share it and I'll add a fixture + fix in a follow-up.

Issue #18795, opened 2026-04-17
[CI] GitHub Action Workflow Cleanup and Restructure


leecalcote:
### Current Behavior
The number of GitHub Action workflows in this repository has grown large. A number of the workflows are either inactive (not used), contain outdated approaches, inconsistently implement the same logic, are insecure, are repetitive and can be de-duplicated. These workflows can be made more reusable, built for purpose, better described, and adhere to GitHub Action authoring best practices.

_Related: See "[Automated Tests/Builds](https://docs.google.com/spreadsheets/d/13Ir4gfaKoAX9r8qYjAFFl_U9ntke4X5ndREY1T7bnVs/edit?gid=1214153102#gid=1214153102)" sheet in Meshery Test Plan._

### Desired Behavior
1. Overall: Comprehensively audit  and refactor workflows of this repository's workflows in context of these concerns. Generate a report with analysis and recommendations.
2. Consistently and securly publishing test results on [qa.meshery.io.](https://qa.meshery.io)
3. Securely running workflows on the appropriate branch (e.g. `master` or `user-contributed`)

### Classification & Documentation

Define and document a workflow classification system in the [Build and Release](https://docs.meshery.io/project/contributing/build-and-release/) doc, covering at minimum:

- [ ] Docs: Definition of Security Level Classification (e.g. requires write token vs. read-only)
- [ ] Docs: Workflow type (self-mutating, artifact-producing, validation-only, release, scheduled maintenance)
- [ ] Characterize choice of trigger model (event-driven vs. embedded within other workflows)
  - [ ] _[Best effort] Retention and scheduling expectations_
- [ ] Define which type of workflows should be CI-gating (green light / red light in pull requests). 

### Audit & Cleanup

- [x] Generate a workflow inventory report (workflow usage, overlap, redundancy, and security posture to inform cleanup decisions. (@simihablo)
  1. [meshery/meshery - Workflow inventory report to analyze GitHub Actions](https://github.com/meshery/meshery/tasks/80cd092f-f72c-403c-a4b9-401f83558f61)
  2. [meshery/meshery - Auditing GitHub Actions workflows for best practices](https://github.com/meshery/meshery/tasks/79c2833a-1133-4793-9c8f-0b4dfd30bcbd)
- [ ] Identify and delete unused or outdated workflows in .github/workflows/ (@ianrwhitney to lead, with team review).
[...]

PR #19273, opened 2026-05-12
ci: pin GitHub Action refs to immutable SHAs in 4 workflows

NSTKrishna:
## What does this PR do?

Pins `@master` and loose version tag refs to immutable SHA hashes 
in 4 workflows identified as "notable offenders" in the workflow 
audit report.

## Why is this needed?

Using `@master` refs creates a supply-chain risk — if an action 
repository is compromised, malicious code automatically runs in 
our workflows with access to repository secrets and tokens.

Pinning to a full SHA ensures:
- The exact same code runs every time, regardless of upstream changes
- An attacker cannot silently replace action code between runs
- Every action version is reviewable and auditable

This is consistent with OpenSSF and GitHub's recommended 
supply-chain hardening practices.

## Changes

| Workflow File | Actions Pinned | Was | Now |
|---|---|---|---|
| `codeql-analysis.yml` | `actions/checkout` | `@master` | SHA `900f221` |
| `codeql-analysis.yml` | `github/codeql-action` | `@v3` | SHA `7fd177f` |
| `helm-chart-releaser.yml` | [action name] | `@master` | SHA `xxxxxxx` |
| `go-mod-sync.yaml` | [action name] | `@master` | SHA `xxxxxxx` |
| `build-and-release-stable.yml` | [action name] | `@master` | SHA `xxxxxxx` |

## Related

#18795
- Directly addresses findings from @simihablo's audit:
  > "25/59 local workflows still use @master. This makes runs 
  > non-reproducible and increases supply-chain risk."
- Consistent with existing pinned patterns already used in 
  `bom.yaml` and `scorecard.yml`

## Testing

These are workflow definition changes only — no logic was modified. 
The pinned SHAs were sourced directly from the most recent 
successful workflow run logs to ensure they match what GitHub 
was already using.

**Notes for Reviewers**
- SHA comments (e.g. `# master`, `# v3`) are preserved inline 
  for human readability
- SHAs were verified from live run logs, not guessed
- No workflow logic, triggers, or job structure was changed


marblom007 (commented):
This is a pain in the rear-end, and unmanageable, I think. If you truly proposing this change, you need to do it across the other 60 repositories between this org and the http://github.com/meshery-extensions org.

NSTKrishna:
@marblom007  That’s fair the aim of this PR was mainly to start addressing findings from the LFX audit report and contribute toward the mentorship project scope.I agree that doing this manually across 60+ repositories isn’t scalable. An automated org-wide solution for detecting and pinning unpinned action refs would make much more sense long term, and I’d be happy to contribute toward building that approach as well.

marblom007:
Good supply-chain hygiene. The inline SHA comments (`# v3`, `# master`) are essential for maintainability — thanks for including them. The removal of the Rego test job should be called out explicitly in the PR description so reviewers don't assume it was dropped accidentally.

NSTKrishna:
@marblom007  Can suggest anything to pr before it get merge

PR #19284, opened 2026-05-12
[mesheryctl] add e2e tests for workspace view command

Omkar-Ugal:
## What does this PR do?
Adds E2E BATS tests for mesheryctl workspace view command.

## Tests added
- no argument 
- name without --orgId 
- invalid orgId 
- non-existent UUID 
- valid UUID + orgId 

### DIFF
--- a/mesheryctl/tests/e2e/008-workspace/03-workspace-view.bats
+++ b/mesheryctl/tests/e2e/008-workspace/03-workspace-view.bats
@@ -17,9 +17,9 @@ setup() {
     run $MESHERYCTL_BIN workspace view foo
     assert_failure
     assert_output --partial "Error"
-    assert_output --partial "Invalid value for --orgId ''"
+    assert_output --partial "--orgId"
+    assert_output --partial "Invalid value"
 }
-
 @test "given an invalid orgId when running mesheryctl workspace view somename --orgId foo then an error message is displayed" {
     run $MESHERYCTL_BIN workspace view somename --orgId foo
     assert_failure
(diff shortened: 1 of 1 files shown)


mesheryctl/tests/e2e/008-workspace/03-workspace-view.bats
lekaf974: I think we should update the error message for required flags here, it is not clear that the flag is required from the error message

mesheryctl/tests/e2e/008-workspace/03-workspace-view.bats
```diff
@@ -0,0 +1,43 @@
...
+    run $MESHERYCTL_BIN workspace view somename --orgId foo
+    assert_failure
+    assert_output --partial "Error"
+    assert_output --partial "Invalid value for --orgid 'foo': must be a valid UUID"
+}
+
+@test "given non-existent UUID when running mesheryctl workspace view then workspace not found error is displayed" {
+    run $MESHERYCTL_BIN workspace view 00000000-0000-0000-0000-000000000000
+    assert_failure
+    assert_output --partial "Error"
```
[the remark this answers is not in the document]
lekaf974: the test is not valid, orgId flag is required given a name or a uuid from the implementation

mesheryctl/tests/e2e/008-workspace/03-workspace-view.bats
lekaf974: instead of value I think rely on fields is more robust from issue we had with golden files

marblom007:
Tests look solid. One note: the success case depends on `/tmp/e2e_workspace_id` existing at test time — worth adding a `skip` guard with an explanatory message if the file is absent, so the test doesn't silently no-op rather than pass.

mesheryctl/tests/e2e/008-workspace/03-workspace-view.bats
Omkar-Ugal: Done! updated assertions to match actual output and added --orgId to UUID test. lmk if anything else needed

PR #19293, opened 2026-05-13
security: harden docs preview workflow

Junnygram:
### Summary
This PR addresses the critical security vulnerability reported regarding the `build-and-preview-docs.yml` workflow. It implements a robust, two-job architecture to air-gap untrusted code from privileged repository tokens.

### Changes
- **Two-Job Architecture**: Split the workflow into `build-preview` (unprivileged) and `deploy-preview` (privileged). The deploy job only touches static artifacts produced by the build and never checks out contributor code.
- **Malicious Script Blocking**: Added `--ignore-scripts` to `npm ci` to prevent execution of attacker-controlled lifecycle scripts.
- **SHA Pinning**: Pinned all GitHub Actions to exact commit SHAs to prevent supply chain poisoning.
- **Least Privilege**: Restricted global permissions to `read-only` and only granted specific write access to the isolated deploy job.

### DIFF
--- a/.github/workflows/build-and-preview-docs.yml
+++ b/.github/workflows/build-and-preview-docs.yml
@@ -10,8 +10,8 @@ on:
       - '.github/workflows/build-and-preview-docs.yml'
       - 'mesheryctl/doc/**'
 
-permissions:
-  contents: read
+# Global permissions are restricted to read-only
+permissions: {}
 
 concurrency:
   group: preview-${{ github.event.pull_request.number || github.run_id }}
@@ -26,8 +26,10 @@ jobs:
     name: Build Preview
     if: github.event.action != 'closed'
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
     outputs:
-      artifact-hash: ${{ steps.calc-hash.outputs.hash }}
+      artifact-digest: ${{ steps.upload.outputs.artifact-digest }}
     env:
       HUGO_VERSION: 0.157.0
     steps:
@@ -60,6 +62,7 @@ jobs:
           [[ -f package-lock.json || -f npm-shrinkwrap.json ]] && npm ci --ignore-scripts || true
 
       - name: Build PR preview
+        if: github.event.action != 'closed'
         env:
           HUGO_ENVIRONMENT: production
           HUGO_ENV: production
@@ -72,14 +75,13 @@ jobs:
           Disallow: /
           EOF
 
-      - name: Calculate artifact hash
+      - name: Calculate artifact hash (deprecated)
         id: calc-hash
         run: |
-          # Create a stable hash of the entire build output directory
-          hash=$(find docs/public -type f -print0 | sort -z | xargs -0 sha256sum | sha256sum | awk '{print $1}')
-          echo "hash=$hash" >> "$GITHUB_OUTPUT"
+          echo "hash=deprecated" >> "$GITHUB_OUTPUT"
 
       - name: Upload Build Artifact
+        id: upload
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
           name: docs-preview-${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


marblom007:
Solid hardening — splitting build from deploy and never checking out contributor code in the privileged job is exactly the right pattern. One question: does the `deploy-preview` job verify the artifact's integrity before pushing? Not strictly required, but worth considering for defense in depth.

Junnygram:
> Solid hardening — splitting build from deploy and never checking out contributor code in the privileged job is exactly the right pattern. One question: does the `deploy-preview` job verify the artifact's integrity before pushing? Not strictly required, but worth considering for defense in depth.


Hi @marblom007 , that's a great point! I've just updated the PR to include artifact integrity verification. The `build-preview` job now generates a SHA256 checksum of the output, which the `deploy-preview` job verifies before pushing to GitHub Pages.  Thanks for the feedback!

.github/workflows/build-and-preview-docs.yml
```diff
@@ -86,19 +72,63 @@ jobs:
...
+            echo "::error::Artifact integrity verification failed! Expected $expected_hash but got $actual_hash"
+            exit 1
+          fi
+          echo "Artifact integrity verified successfully."
+
       - name: Deploy PR preview
         if: github.event.action != 'closed'
-        uses: rossjrw/pr-preview-action@v1.6.3
+        uses: rossjrw/pr-preview-action@732381f885741f021e17df146019a7905f061250 # v1.6.3
         with:
```
CodexRaunak, line 123: These commit hashes are good, instead of using the tag (that can be compromised), we can add a depandabot config that updates the commit hashes on recurring basis

.github/workflows/build-and-preview-docs.yml
```diff
@@ -141,14 +171,14 @@ jobs:
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git add pr-preview
           git commit -m "Prune old PR previews"
-          git push
+          git push origin gh-pages
```
saurabhraghuvanshii, line 175: @Junnygram  I don't think it will work.

.github/workflows/build-and-preview-docs.yml
```diff
@@ -11,8 +11,7 @@ on:
       - 'mesheryctl/doc/**'
 
 permissions:
-  contents: write
-  pull-requests: write
+  contents: read
```
saurabhraghuvanshii: @Junnygram, I’m afraid this will fail the preview deployment.

saurabhraghuvanshii (changes requested):

.github/workflows/build-and-preview-docs.yml
```diff
@@ -23,57 +22,44 @@ defaults:
...
+      artifact-hash: ${{ steps.calc-hash.outputs.hash }}
     env:
       HUGO_VERSION: 0.157.0
-      PREVIEW_RETENTION_LIMIT: 6
-
     steps:
       - name: Checkout PR code
-        if: github.event.action != 'closed'
-        uses: actions/checkout@v6
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
Amr-Shams, line 37: whyt this commit hash though?

.github/workflows/build-and-preview-docs.yml
```diff
@@ -86,19 +72,63 @@ jobs:
           Disallow: /
           EOF
 
+      - name: Calculate artifact hash
```
Amr-Shams: there is a better github actions mantianed by github

Junnygram: > there is a better github actions mantianed by github

yes, whats the name ?

Amr-Shams: the built-in artifact-digest output from actions/upload-artifact@v4 (a GitHub-maintained action). as it outputs a cryptographic digest automatically (steps.upload.outputs.artifact-digest) which eliminates the need for manual sha256sum pipelines and is more reliable for SLSA provenance. See the [v4 release notes](https://github.com/actions/upload-artifact/releases/tag/v4.0.0).

PR #19325, opened 2026-05-13
[UI] Infinite re-render loop in Connections Table due to unstable visibility state synchronization #19324

### LINKED ISSUE #19324: [UI] Infinite re-render loop in Connections Table due to unstable visibility state synchronization
### Current Behavior
The Connections Table crashes with a `Maximum update depth exceeded` React runtime error. This is triggered by a recursive feedback loop in `ConnectionTable.tsx`. 

A `useEffect` hook was used to synchronize responsive column visibility into the component's state. Because the underlying utility returned a new object reference on every call, and upstream dependencies (like the `columns` definition) were referentially unstable, the component triggered a `setState` on every render. This caused a re-render, which re-triggered the effect, leading to a crash.

### Desired Behavior
The Connections Table should render stably regardless of screen width changes or background data updates. Column visibility should be derived purely during the render phase rather than synchronized via side-effects to prevent recursive render cycles.

### Screenshots/Logs

Image

Image

**Trace:**
The loop was traced to the `useEffect` at line 1446 in `meshery/ui/components/connections/ConnectionTable.tsx`.

### Environment
- **Browser:** Chrome / Safari / Firefox
- **Host OS:** Mac
- **Meshery Server Version:** v1.0.20 (current master)
- **Platform:** Docker / Local

thechillbasu:
This PR resolves a critical `Maximum update depth exceeded` error in the Connections Table. The issue was caused by a recursive feedback loop between the `ResponsiveDataTable` wrapper and the parent component's state management during column visibility synchronization.

**Changes:**
- Bypassed `ResponsiveDataTable` wrapper to gain direct control over `MUIDataTable` prop stability.
- Replaced side-effect based (`useEffect` + `setState`) column visibility synchronization with pure `useMemo` derivation.
- Stabilized `options` and `components` props using `useMemo` to prevent recursive re-render cycles triggered by referential instability.
- Integrated `mergedVisibility` logic to ensure responsive column defaults and user-toggled overrides coexist without triggering extra renders.

**Notes for Reviewers**
- This PR fixes #19324
- The fix was verified by monitoring the React render cycle to ensure the component now renders exactly once when data or window dimensions change, rather than infinitely.
- I have added comments in the code wherever I have made changes to clarify the reason of change.

**Screenshots**

Before:-
image

After:-
Screenshot 2026-05-13 at 3 10 46 PM

### DIFF
--- a/ui/components/connections/ConnectionTable.tsx
+++ b/ui/components/connections/ConnectionTable.tsx
@@ -555,19 +555,29 @@ const ConnectionTable = ({
     handleDeleteConnections,
   });
 
+  // Purely derive responsive visibility to decouple it from the render cycle and avoid recursive setState loops.
+  // This replaces the useEffect based synchronization which triggered infinite updates due to unstable dependencies.
+  const responsiveVisibility = useMemo(
+    () => getResponsiveColumnVisibility(columnNames, colViews, width),
+    [columnNames, colViews, width],
+  );
+
+  // Track user-initiated column overrides separately to maintain clear precedence over responsive defaults.
+  // This state is now only updated on manual user interaction, preventing automatic render-loop triggers.
+  const [columnVisibility, setColumnVisibility] = useState<Record<string, boolean | undefined>>({});
+
+  // Compose responsive defaults with manual overrides into a single stable visibility map.
+  // Merging derived and state-based visibility ensures responsiveness without redundant side-effect cycles.
+  const mergedVisibility = useMemo(
+    () => ({ ...responsiveVisibility, ...columnVisibility }),
+    [responsiveVisibility, columnVisibility],
+  );
+
   const [tableCols, updateCols] = useState(columns);
 
   useEffect(() => {
     updateCols(columns);
   }, [columns]);
-
-  const [columnVisibility, setColumnVisibility] = useState<Record<string, boolean | undefined>>(
-    () => getResponsiveColumnVisibility(columnNames, colViews, width),
-  );
-
-  useEffect(() => {
-    setColumnVisibility(getResponsiveColumnVisibility(columnNames, colViews, width));
-  }, [colViews, columnNames, width]);
 
   if (isConnectionLoading) {
     return <LoadingScreen animatedIcon="AnimatedMeshery" message="Loading Connections" />;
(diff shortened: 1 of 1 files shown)


rishiraj38 (approved):

rishiraj38 (commented):
Quick q: do we know the root cause, is it our local visibility` useEffect`, or `ResponsiveDataTable `itself? It already uses `MUIDataTable` under the hood, so I’m curious whether we could fix this once in the wrapper instead of only in Connections.

thechillbasu:
> Quick q: do we know the root cause, is it our local visibility` useEffect`, or `ResponsiveDataTable `itself? It already uses `MUIDataTable` under the hood, so I’m curious whether we could fix this once in the wrapper instead of only in Connections.

@rishiraj38 The crash was basically an infinite loop between the visibility effect and the table wrapper. The effect was calling setState with a brand new object every single time, which triggered a re-render and just kept the cycle going forever. The wrapper made things worse because it was also trying to sync its own internal state back to the parent.

I switched everything to useMemo so we can compute the visibility directly during render instead of relying on a side effect. By getting rid of that setState call, we basically broke the cycle. I think it's a lot more stable to just derive these values on the fly than to try and keep them in sync across different components cuz the problem with syncing is that you end up with two sources of truth (the parent state and the child state) trying to mirror each other. If they disagree for even a split second, there's a clash, and that’s where the infinite loop comes from.

rishiraj38 (changes requested):

rishiraj38:
> > Quick q: do we know the root cause, is it our local visibility` useEffect`, or `ResponsiveDataTable `itself? It already uses `MUIDataTable` under the hood, so I’m curious whether we could fix this once in the wrapper instead of only in Connections.
> 
> @rishiraj38 The crash was basically an infinite loop between the visibility effect and the table wrapper. The effect was calling setState with a brand new object every single time, which triggered a re-render and just kept the cycle going forever. The wrapper made things worse because it was also trying to sync its own internal state back to the parent.
> 
> I switched everything to useMemo so we can compute the visibility directly during render instead of relying on a side effect. By getting rid of that setState call, we basically broke the cycle. I think it's a lot more stable to just derive these values on the fly than to try and keep them in sync across different components cuz the problem with syncing is that you end up with two sources of truth (the parent state and the child state) trying to mirror each other. If they disagree for even a split second, there's a clash, and that’s where the infinite loop comes from.

if that’s the case, I think it’d be better to address this at the `ResponsiveDataTable` level itself since it’s used across so many places. Otherwise, we might end up running into the same kind of issue again in different components if a similar state sync pattern exists there too.

PR #19327, opened 2026-05-13
Attempt at naive migration from sql-lite to postgresql.

alex-rudenkiy:
Signed-off-by: Alex <alexrudenkiy2014@gmail.com>

### DIFF
--- a/server/cmd/main.go
+++ b/server/cmd/main.go
@@ -225,6 +225,11 @@ func main() {
 	defer preferencePersister.ClosePersister()
 
 	dbHandler := models.GetNewDBInstance()
+	if dbHandler == nil || dbHandler.DB == nil {
+		log.Error(fmt.Errorf("Database connection failed. Check your DATABASE_URL and credentials."))
+		os.Exit(1)
+	}
+
 	regManager, err := meshmodel.NewRegistryManager(dbHandler)
 	if err != nil {
 		log.Error(ErrInitializingRegistryManager(err))
@@ -234,11 +239,12 @@ func main() {
 	brokerConn := nats.NewEmptyConnection
 
 	err = dbHandler.AutoMigrate(
-		&meshsyncmodel.KubernetesKeyValue{},
 		&meshsyncmodel.KubernetesResource{},
+		&meshsyncmodel.KubernetesResourceObjectMeta{},
+		&meshsyncmodel.KubernetesKeyValue{},
 		&meshsyncmodel.KubernetesResourceSpec{},
 		&meshsyncmodel.KubernetesResourceStatus{},
-		&meshsyncmodel.KubernetesResourceObjectMeta{},
+		&meshsyncmodel.KubernetesResource{},
 		&models.PerformanceProfile{},
 		&models.MesheryResult{},
 		&models.MesheryPattern{},
--- a/server/handlers/database_handlers.go
+++ b/server/handlers/database_handlers.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/meshery/meshery/server/models"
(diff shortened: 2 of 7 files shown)


alex-rudenkiy:
+ @leecalcote @Meyazhagan @marblom007 @YASHMAHAKAL 

**Description:**
This is an initial attempt to enable PostgreSQL support alongside SQLite. While the migration logic is in place, this PR highlights a fundamental "leaky abstraction" issue within the current GORM-based data layer.

**Current Status:**

Schema Inconsistency: PostgreSQL forces BYTEA for gorm:"type:bytes" tags, causing fatal errors when inserting JSON.

Database Divergence: We are hitting a wall where the same code behaves differently across dialects due to hardcoded tags in meshery/schemas.

**Fatal Blockers:**
Attempts to fix this via GORM hooks/callbacks failed due to internal schema caching. Continuing with the current GORM setup for Postgres will likely lead to a "patchwork" architecture that is brittle and hard to maintain as new entities are added.

**Proposed Pivot (The "Ent" Experiment):**
Since my GSoC application was rejected, I am no longer bound by strict deadlines and am willing to undertake a deep refactoring. I propose a pilot migration of some models to Ent ORM to resolve type-mapping issues at the code-generation level.

Why Ent? It uses code generation and strict schema definitions that would allow us to handle type mapping (JSONB vs. Text/Blob) at the engine level, completely abstracting it from the business logic.

**Why this matters now:**

Scalability: The project cannot scale horizontally without a proper RDBMS implementation.

Technical Debt: The longer we wait to fix the type-mapping strategy, the harder it will be to migrate later.

I’d like to discuss the possibility of a partial PoC using Ent to prove we can have a truly cross-DB data layer without these "type:bytes" hacks.

publicuml

image

image

image

image

image

image

image

image

image

**Typical Error Message:**

```
2026/05/13 16:31:31 C:/Users/.../go/pkg/mod/github.com/meshery/schemas@v1.2.19/models/v1beta3/component/component_helper.go:76 ERROR: invalid input syntax for type bytea (SQLSTATE 22P02)
[6.296ms] [rows:0] INSERT INTO "component_definition_dbs" ("id","schema_version","version","display_name","description","format","styles","capabilities","status","metadata","configuration","component","created_at","updated_at","deleted_at","model_id") VALUES ('5ca06414-8d57-4d51-abd3-98945a12854b','components.meshery.io/v1beta2','v1.0.0','Node Shard','','JSON','{"primaryColor":"#eb2318","secondaryColor":"#FF685D","shape":"circle","svgColor":"ui/public/static/img/meshmodels/volcano/color/nodeshard-color.svg","svgComplete":"","svgWhite":"ui/public/static/img/meshmodels/volcano/white/nodeshard-white.svg"}','[{"description":"Configure the visual styles for the component","displayName":"Styling","entityState":["declaration"],"key":"","kind":"mutate","schemaVersion":"capability.meshery.io/v1beta1","status":"enabled","subType":"","type":"style","version":"0.7.0"},{"description":"Change the shape of the component","displayName":"Change Shape","entityState":["declaration"],"key":"","kind":"mutate","schemaVersion":"capability.meshery.io/v1beta1","status":"enabled","subType":"shape","type":"style","version":"0.7.0"},{"description":"Drag and Drop a component into a parent component in graph view","displayName":"Compound Drag And Drop","entityState":["declaration"],"key":"","kind":"interaction","schemaVersion":"capability.meshery.io/v1beta1","status":"enabled","subType":"compoundDnd","type":"graph","version":"0.7.0"}]','enabled','{"configurationUISchema":"","genealogy":"","instanceDetails":null,"isAnnotation":false,"isNamespaced":false,"published":false,"source_uri":"git://github.com/volcano-sh/volcano/master/config/crd/volcano/bases"}',NULL,'{"version":"shard.volcano.sh/v1alpha1","kind":"NodeShard","schema":"{\n \"description\": \"NodeShard is a collection of nodes dedicated to a specific scheduler\",\n \"properties\": {\n  \"spec\": {\n   \"description\": \"Specification of the desired behavior of the NodeShard.\",\n   \"properties\": {\n    \"nodesDesired\": {\n     \"description\": \"NodesDesired defines the list of nodes desired to be included in this NodeShard.\",\n     \"items\": {\n      \"type\": \"string\"\n     },\n     \"type\": \"array\"\n    }\n   },\n   \"required\": [\n    \"nodesDesired\"\n   ],\n   \"type\": \"object\"\n  }\n },\n \"required\": [\n  \"spec\"\n ],\n \"title\": \"Node Shard\",\n \"type\": \"object\"\n}"}','2026-05-13 16:31:31.624','2026-05-13 16:31:31.624',NULL,'9f53e26a-b9b3-baaf-24b0-193eb7aff746')
```

PR #19328, opened 2026-05-13
fix(ui): resolve redundant profile paths and align field names

Junnygram:
# PR Description: Fix Redundant Profile API Paths and Field Normalization

## Description
This PR fixes the `400 Bad Request` errors during user profile interactions by correcting malformed API paths and ensuring consistent field naming across UI components.

### Key Changes
- **Redundant Path Fix:** Patched `ui/rtk-query/user.ts` to prevent appending `/profile` when the ID is already `"profile"`, resolving the `/api/user/profile/profile` malformation.
- **Data Normalization:** Integrated `normalizeUserProfileSummary` into the `getUserProfileSummaryById` RTK hook to transparently handle both `first_name` and `firstName` backend responses.
- **Component Alignment:** Updated `InfoModal.tsx` and `DesignViewListItem.tsx` to correctly handle `snake_case` fields from the backend, ensuring user names and profiles are rendered correctly.
- **Optimized Data Access:** Updated embedded user checks in `DesignViewListItem` to support multiple naming conventions, reducing unnecessary API calls.

## Verification Results
- [x] Verified that `/api/user/profile` is called correctly when using `"profile"` as an ID.
- [x] Confirmed that `first_name` and `last_name` from the backend are correctly mapped to UI labels.
- [x] Validated that existing functionality for pattern owners remains intact.

## Related Issues
- Complements the schema updates in `meshery/schemas`.
- Fixes Layer5 Cloud profile update workflow.

### DIFF
--- a/ui/components/shared/Modal/Information/InfoModal.tsx
+++ b/ui/components/shared/Modal/Information/InfoModal.tsx
@@ -403,7 +403,9 @@ const InfoModal_: FC<InfoModalProps> = React.memo((props) => {
                   <Typography gutterBottom variant="subtitle1">
                     <CustomTooltip
                       title={`Owner: ${
-                        resourceUserProfile?.first_name + ' ' + resourceUserProfile?.last_name
+                        (resourceUserProfile?.firstName ?? resourceUserProfile?.first_name) +
+                        ' ' +
+                        (resourceUserProfile?.lastName ?? resourceUserProfile?.last_name)
                       }`}
                     >
                       <div>
@@ -521,7 +523,9 @@ const OwnerChip = ({ userProfile }) => {
           <Link href={`${MESHERY_CLOUD_PROD}/user/${userProfile.id}`} rel="noopener noreferrer">
             <Avatar src={userProfile.avatarUrl} />
           </Link>
-          <Typography>{`${userProfile.first_name} ${userProfile.last_name}`}</Typography>
+          <Typography>{`${userProfile.firstName ?? userProfile.first_name} ${
+            userProfile.lastName ?? userProfile.last_name
+          }`}</Typography>
         </>
       ) : (
         <Skeleton variant="circular" width={40} height={40} />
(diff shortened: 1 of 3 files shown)


ui/components/shared/Modal/Information/InfoModal.tsx
```diff
@@ -403,7 +403,7 @@ const InfoModal_: FC<InfoModalProps> = React.memo((props) => {
                   <Typography gutterBottom variant="subtitle1">
                     <CustomTooltip
                       title={`Owner: ${
-                        resourceUserProfile?.firstName + ' ' + resourceUserProfile?.lastName
+                        resourceUserProfile?.first_name + ' ' + resourceUserProfile?.last_name
                       }`}
```
Oluwatunmise-olat, line 409: @Junnygram Please can we also just add fallbacks like `resourceUserProfile?.firstName ?? resourceUserProfile?.first_name` just incase

ui/rtk-query/user.ts
```diff
@@ -132,7 +132,10 @@ export const userApi = api
             }
           },
         }),
-        transformResponse: normalizeUserProfileSummary,
+        transformResponse: (response) => {
+          const { normalizeUserProfileSummary } = require('./userProfile');
+          return normalizeUserProfileSummary(response);
```
Oluwatunmise-olat: 1. normalizeUserProfileSummary has been previously imported on line 13 any reason why it is being re-imported?
2. Also can we possibly not use inline require?

Oluwatunmise-olat (changes requested):

arjunmehta-git:
Thanks for the contribution, @Junnygram! Since this was opened, the profile-summary normalization it addresses has landed in master: getUserProfileSummaryById now normalizes at the RTK transformResponse boundary via normalizeUserProfileSummary, plus a robust JSON response handler - the naming-convention-compliant place (wire is camelCase; normalize once at the boundary per the identifier-naming guide, rather than field-by-field per component). Since that supersedes the changes here, and the PR has since drifted into merge conflicts, we are closing it. We really appreciate you digging into this - the good-first-issue queue has plenty more if you would like to pick up another. Thank you!

PR #19380, opened 2026-05-14
ui/chore

miacycle:
- **UI: Address build warnings. Use Sistent design system**
- **UI: Address build warnings. Use Sistent design system**
- **fix: address UI review feedback**
- **UI: delete icon**


leecalcote (approved):
lgtm. thanks!

PR #19392, opened 2026-05-14
[ui] Delete 6 unreferenced GraphQL operations

leecalcote:
## Summary
Deletes 6 GraphQL operation files plus their generated Relay artifacts that have **zero live consumers**, confirmed by repo-wide grep across non-test, non-generated, non-build-output files.

```
subscriptions: ClusterResourcesSubscription
queries:       TelemetryComponentsQuery, DataPlanesQuery, CatalogPatternQuery
mutations:     OperatorStatusMutation
shim:          components/graphql/queries/ResetDatabaseQuery (re-export
               with no callers — both real callers import the canonical path)
```

Also drops the orphaned `subscribeClusterResources` barrel export from `ui/graphql/subscriptions/index.ts`.

877 lines removed across 12 files. No server-side changes — resolvers stay in place pending the broader GraphQL→SSE/REST migration.

## Context
First in a series of PRs migrating Meshery off GraphQL onto plain REST + SSE. This one is the trivially safe slice: confirmed-dead code with no consumers.

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `npx eslint --cache --quiet .` clean
- [x] `npx vitest run` — the 371 passing tests still pass; the 18 pre-existing failures (theme/MUI Modal mock issues in unrelated test files) reproduce on master and are not introduced here
- [x] Repo-wide grep for each deleted operation name returns no live references

## Follow-up note
The `.graphql.js` companions for each `.graphql.ts` artifact were left in place — the repo already contains several orphan `.graphql.js` files without `.ts` siblings, so this matches the prevailing tree state. A separate sweep can clean those up.


pontusringblom (approved):

ui/components/graphql/queries/ResetDatabaseQuery.tsx
pontusringblom, line 1: @copilot where is the replacement handler in our REST API for this operation?

leecalcote:
## Re: 'are these really dead?'

Good push — you were right to question. Honest gap analysis of the 6 deletions:

| # | Item | REST coverage at the time of this PR | Now |
|---|---|---|---|
| 1 | `ClusterResourcesSubscription` | ❌ none | ✅ `GET /api/system/kubernetes/cluster/resources` (added in [#19397](https://github.com/meshery/meshery/pull/19397) commit 4c298cf) |
| 2 | `TelemetryComponentsQuery` | ❌ none | ✅ `GET /api/telemetry/components` (added in [#19397](https://github.com/meshery/meshery/pull/19397) commit 4c298cf) |
| 3 | `DataPlanesQuery` | ❌ none | ✅ `GET /api/system/meshes/data-planes` (added in [#19397](https://github.com/meshery/meshery/pull/19397) commit 4c298cf) |
| 4 | `CatalogPatternQuery` | ✅ `/api/pattern/catalog` | ✅ |
| 5 | `OperatorStatusMutation` | ✅ added in [#19397](https://github.com/meshery/meshery/pull/19397): `POST /api/system/kubernetes/contexts/{id}/operator` | ✅ |
| 6 | `ResetDatabaseQuery` shim | ✅ already on master: `/api/system/database/reset`, and the *real* query file at `graphql/queries/ResetDatabaseQuery.tsx` was NOT deleted | ✅ |

**Database Reset specifically (your example):** the file deleted here is the re-export shim at `components/graphql/queries/ResetDatabaseQuery.tsx`. The canonical query at `graphql/queries/ResetDatabaseQuery.tsx` is untouched, and the REST handler `ResetSystemDatabase` at `server/handlers/database_handlers.go:77` is wired to `POST /api/system/database/reset`. Both `DatabaseSummary.tsx` and `ConnectionTable.tsx` use the RTK Query mutation against that endpoint (and PR #19398 cleanly moves `ConnectionTable` to a different per-cluster resync endpoint).

**Items 1–3:** "zero UI consumers" doesn't mean "safe to delete entirely." The original audit only verified UI references. The server resolvers still served these via `/api/graphql` — anything external (kubectl wrappers, scripts, future UI work) would have silently lost access when the teardown PR #19400 removed the resolvers. PR #19397 now exposes all three as REST endpoints with identical semantics. Functionality is preserved through and after the teardown.

So this PR's deletions remain correct (no UI consumers, file removal is safe), and the broader teardown is now backed by REST equivalents for every operation it removes.

PR #19408, opened 2026-05-14
[Docs] Remove test-status page

pontusringblom:
## Summary
- remove the obsolete `project/contributing/test-status` page
- remove remaining source-doc references to that page
- scrub archived static docs/search/sitemap entries that still linked to it

## Notes
- no `test-status` references remain under `docs/`

PR #19415, opened 2026-05-14
[CI] Cleanup: Delete workflow that pushes e2e results to Docs

leecalcote:
This workflow is no longer necessary.

Part of https://github.com/meshery/meshery/issues/18795


pontusringblom:
Thanks! Yes, this goes hand-in-hand with the removal of that page (and all references) from the Meshery Docs. I understand that we're trying to push to either the current, open PR or push to qa.meshery.io. 👍

pontusringblom:
Here's the adjoining PR #19408

pontusringblom (dismissed):
Thanks, partner!

hortison (dismissed):

PR #19449, opened 2026-05-15
[Server] Add expectedIssuer config for strict OIDC iss validation

### LINKED ISSUE #19259: feat(auth): add expectedIssuer config separate from RemoteProviderURL for strict iss validation
## Context

Tracked from meshery-cloud's custom-domain authentication design (§8.4 "Hydra host-aware browser endpoint", "OIDC-compliance caveat"):

> Keeping \`iss\` set to the canonical host while the user-facing OIDC discovery endpoint lives on the custom domain is **not** compliant with strict OIDC validators: the OIDC spec requires \`iss\` to match the issuer published by the discovery document at the host the client talked to.

In Meshery Cloud, Hydra mints JWTs whose \`iss\` claim is always the canonical Hydra issuer URL (e.g. \`https://cloud.meshery.io\`). When a Meshery Server is bound (via \`PROVIDER_BASE_URLS\`) to a custom Cloud domain (e.g. \`cloud.meshery.io\` or an org-owned domain), the OIDC discovery URL the Meshery Server talks to is on that custom host but the \`iss\` claim it gets back is canonical.

\`server/models/remote_auth.go\` \`VerifyToken\` today validates the JWT signature via JWKS but does **not** strictly check \`iss\` against \`RemoteProviderURL\`, so the discrepancy is currently a no-op for Meshery Server. This issue tracks the hardening pass that would close the gap.

## What

Add an \`expectedIssuer\` configuration value, independent of \`RemoteProviderURL\`, that \`VerifyToken\` uses for the strict \`iss\` comparison. The two values diverge in custom-domain deployments:

- \`RemoteProviderURL\` — the host the Meshery Server *talks to* (custom domain).
- \`expectedIssuer\` — the canonical \`iss\` value Hydra actually mints (canonical Cloud host).

## Why now

This is **not** blocking any first-party Meshery Server deployment; the existing JWKS-signature-only check is sufficient because the verifier is bespoke and under our control.

The block is third-party OIDC clients (relying parties whose verifier we don't ship). Strict-compliant OIDC verifiers will reject Cloud-issued tokens delivered via a custom-domain discovery document because the \`iss\` claim doesn't match the discovery's issuer URL. **Until \`expectedIssuer\` lands here OR meshery-cloud grows a per-host re-signing layer in front of Hydra's token endpoint, third-party OIDC clients are NOT allowed on Cloud's custom-domain surface.**

## Out of scope

- Per-host \`iss\` re-signing in Cloud (option (a) in the design doc) is a separate, larger workstream. This issue tracks only the Meshery-Server-side strict check (option (b)'s prerequisite).

## References

- meshery-cloud design doc: \`docs/custom-domain-authentication.md\` §8.4 — OIDC-compliance caveat
- meshery-cloud W4 implementation: PR opening the browser-fronting proxy

Pranav-d33:
Add ExpectedIssuer field to RemoteProvider, separate from RemoteProviderURL, for strict JWT issuer claim validation in VerifyToken. When PROVIDER_TOKEN_ISSUER is set, VerifyToken enforces that the token's iss claim matches; when unset, behavior is unchanged (backward compatible).

In custom-domain deployments, Hydra mints JWTs with the canonical issuer URL (e.g. https://cloud.layer5.io) while RemoteProviderURL points to the custom domain. A separate expectedIssuer allows OIDC-compliant iss validation without requiring the two values to be identical.

Fixes #19259


Pranav-d33:
Hey @marblom007  — could you take a look at this when you get a chance? 🙂

The only failures here are in the e2e tests, and they seem to be part of a broader broken baseline affecting all PRs triggering `tests-ui-e2e` — not something introduced by this change specifically. Would love a quick review so this can move forward!

server/models/remote_auth.go
```diff
@@ -346,6 +346,14 @@ func (l *RemoteProvider) VerifyToken(tokenString string) (*jwt.MapClaims, error)
 	if err != nil {
 		return nil, ErrTokenPrase(err)
```
rishiraj38:
```suggestion
		return nil, ErrTokenParse(err)
```

Pranav-d33:
@hamza-mohd @saurabhraghuvanshii   can you please take a final look at this ?

saurabhraghuvanshii:
@Pranav-d33 Sorry, but I think someone else may be able to provide better feedback on these changes.

By the way, please resolve the merge conflicts.

PR #19479, opened 2026-05-16
feat(policies): WASM build of relationship engine for client-side evaluation

leecalcote:
## What

Adds a **Go→WASM build of the relationship policy engine** so Meshery UI can evaluate designs entirely client-side, at parity with Meshery Server's HTTP relationship-evaluation endpoint. Meshery remains the source of truth for the engine.

## Shared-orchestration extraction (anti-drift, the important part)

The contract requires the WASM build and the server to produce parity output. Rather than copy-paste the orchestration, the **registry-INDEPENDENT** pipeline that wraps the bare `GoEngine` is extracted into one shared home, `server/policies/orchestration.go`:

- the `MAX_RE_EVALUATION_DEPTH` re-evaluation loop
- alias resolution (`ResolveAliasesInDesign` + `ResolvedAliasesV1beta2ToV1beta1`)
- per-iteration design **version bump** (`BumpDesignVersion`)
- unique trace merge (`MergeTraceUnique`)
- cross-iteration action de-dup (`DeduplicateActions`)
- `DoesntNeedReeval`, the completed-at `Timestamp`, and the `ReturnDiffOnly` projection

These helpers previously lived in package `handlers`. They now live in package `policies` and the handler delegates to them via thin aliases. **The handler's behavior is identical** — it still runs the registry-bound steps (`HydratePattern`, `processEvaluationResponse`, `DehydratePattern`, `GetEntitiesMemoized`) around its own loop, so server output is unchanged. Both the HTTP handler and the WASM entrypoint now call the one shared `(*GoEngine).EvaluateDesignOrchestrated` / shared helpers, so the two evaluators **cannot drift**. The handler file shrinks by 232 lines (pure de-duplication).

## Registry-independent boundary

The WASM build replicates only the registry-independent pipeline. It does **not** hydrate components from a registry — there is none in the browser; the client compensates for component hydration on its side. `ReturnDiffOnly` is honored as its registry-independent projection (trace deltas only, no registry-backed component hydration).

## Artifact / Makefile

- `server/cmd/wasm/relationship-engine/main.go` (`//go:build js && wasm`) — a thin `syscall/js` boundary that attaches `globalThis.__mesheryRelationshipEngine` with the agreed contract: `contract=1`, `version` (meshery git version+SHA via `-ldflags`), and synchronous string→string `init` / `evaluate` / `selfTest`. Every call recovers panics into `{"error":...}`. `init` feeds the mixed v1alpha3/v1beta2 relationship array through `policies.ConvertRelationships` exactly as the handler does. A wasm-safe no-op `logger.Handler` shim backs `NewGoEngine` (the real engine path is exercised, not stubbed).
- `make wasm-engine` builds `ui/public/static/wasm/policy_engine.wasm` and copies the Go runtime's `wasm_exec.js` (resolving `$(GOROOT)/lib/wasm` for Go ≥1.24 with a `misc/wasm` fallback) alongside it.
- **Artifacts are .gitignored and produced by the build pipeline:** `policy_engine.wasm` is ~54 MB. A blob that large recommitted on every rebuild is permanent compounding repo bloat — and the only "precedent" file (`ui/public/static/namespace_discovery_relationship_policy.wasm`) is just ~133 KB, so that precedent does not hold here. Both `policy_engine.wasm` and `wasm_exec.js` are therefore **`.gitignored` and untracked**. They are regenerated by the build pipeline instead: `wasm-engine` is now a **prerequisite of `ui-meshery-build` and `ui-build`**, so every normal UI build — and `ui-server` transitively — refreshes the served bundle. The two UI-build CI workflows (`build-meshery-ui.yml`, `build-ui-and-server.yml`) gain a Go toolchain plus an explicit `make wasm-engine` step before the UI build. The Docker image's Node UI-builder stage has no Go toolchain, so the Go (`meshery-server`) stage runs `make wasm-engine` and the produced `ui/public/static/wasm` directory is layered onto the served static bundle in the final image.

### Out-of-scope fix flagged for reviewers

`server/models/pattern/utils/k8s.go` is dead code (explicitly marked `// not in use`, zero callers repo-wide) whose transitive k8s/docker client imports do not compile under `GOOS=js GOARCH=wasm`. It is now constrained `//go:build !js` so the registry-independent version bridges in that package build for wasm. **Server (non-js) builds are unaffected** — the file still compiles for every other target and its exported functions remain available (verified via `go doc`).

## Parity test

`server/policies/orchestration_test.go`:

- **`TestMaxReEvaluationDepthParity`** — pins `MAX_RE_EVALUATION_DEPTH` against the server constant value (parity pin).
- **`TestEvaluateDesignOrchestrated_ParityWithHandlerMinusRegistry`** — proves `EvaluateDesignOrchestrated` returns byte-identical actions / trace / design version / resolvedAliases to an independently reconstructed "handler loop minus registry hydration" on a deterministic hierarchical/inventory fixture.
- **`TestEvaluateDesignOrchestrated_ReturnDiffOnly`** — verifies the registry-independent diff-only projection.

`make wasm-engine` doubles as the `GOOS=js GOARCH=wasm` build smoke.

## Verification

- `GOOS=js GOARCH=wasm go build ./server/cmd/wasm/relationship-engine` — succeeds (`WebAssembly (wasm) binary module`)
- `make wasm-engine` — produces `policy_engine.wasm` (~54M) + `wasm_exec.js` (17K, byte-identical to Go 1.25.5 toolchain); `git ls-files ui/public/static/wasm/` is **empty** afterwards (artifacts untracked, matched by `.gitignore`)
- `make -n ui-build` / `make -n ui-meshery-build` — confirm the `wasm-engine` recipe runs as a prerequisite of the UI build
- `go test ./server/policies/...` — PASS (full suite, incl. new orchestration tests)
- `go test ./server/handlers/ -run 'Policy|Relationship|Alias|Evaluation|EvaluateDesign'` — PASS (incl. `TestBuildPolicyEvaluationEventMetadata...`, `TestParseRelationshipToAlias`)
- `go build ./server/...` and server `main.go` build — PASS; `go vet` clean on all touched packages

Consumed by the in-browser relationship-evaluation client (coordinated separately).


aabidsofi19:
@leecalcote  is this pr different from https://github.com/meshery/meshery/pull/18939 . the support for creating and serving wasm assets has been added in that pr too

Issue #14794, opened 2025-05-22
[Models] Component relationships for AWS services


miacycle:
### Current Behavior
Meshery Models are declarative representations of infrastructure and applications. Within these models, Relationships define how different Components (e.g., Kubernetes resources, Cloud services) interact and depend on each other. These relationships are crucial for visualizing, understanding, and managing complex cloud native systems.

### Desired Behavior
This issue highlights the need to significantly expand the breadth and depth of Meshery Relationships across a wide array of technologies supported by Meshery. As Meshery continues to integrate with more cloud-native technologies (Kubernetes, public clouds, and all CNCF projects), there's a growing need to accurately model the intricate relationships between their components -vital for providing users with comprehensive insights and control over their deployments.

- Recommended Skills: DevOps, systems administration, solutions architecture. Experience with Kubernetes and AWS and it's services.

### Implementation
- Research and Analyze Technologies: Dive deep into various cloud-native technologies (e.g., different service meshes, CI/CD tools, databases, messaging systems, IaaS/PaaS services) to understand their components and how they interconnect.
- Develop Relationship Definitions: Create and contribute relationship definitions, typically in JSON or YAML format, to the Meshery models. 
- Model Inter-Technology Interactions: Focus particularly on defining relationships between components from different technologies (e.g., how a Kubernetes deployment relates to an AWS RDS instance, or how a Linkerd service interacts with a Prometheus monitoring component).
- Document New Relationships: Clearly document the newly defined relationships, their purpose, and how they are represented within Meshery designs, contributing to the official Meshery documentation.

### Acceptance Tests
1. Relationship definitions are embedded in models and are functional.

---

### General Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)

### Understanding Meshery Models

- 📚 [Models](https://docs.meshery.io/concepts/logical/models)
  - 📚 [Components](https://docs.meshery.io/concepts/logical/components)
  - 📚 [Relationships](https://docs.meshery.io/concepts/logical/components)
- 👨‍💻[Models Repository](https://github.com/meshery/meshery/tree/master/server/meshmodel)

 ### Contributing to Meshery Models
 - [Contributing Models](https://docs.meshery.io/project/contributing/contributing-models)
[...]

PR #19490, opened 2026-05-16
feat(meshmodel): add AWS DynamoDB Table and SQS Queue relationships to Kubernetes Deployment and Pod

aahil62:
## Description

Adds edge non-binding reference relationships for two AWS services to Kubernetes workloads, following the same pattern as #17372 (Kinesis→Deployment/Pod) and #17175 (ECR→Deployment/Pod).

**DynamoDB (aws-dynamodb-controller v1.8.0):**
- `Table → Deployment` — models a Deployment whose containers interact with a DynamoDB table
- `Table → Pod` — models a Pod whose containers interact with a DynamoDB table

**SQS (aws-sqs-controller v1.4.3):**
- `Queue → Deployment` — models a Deployment whose containers produce or consume SQS messages
- `Queue → Pod` — models a Pod whose containers produce or consume SQS messages

Relates to #14794

Happy to update the Meshery Integrations spreadsheet once this is merged.

## Checklist
- [x] Follows existing relationship schema (`relationships.meshery.io/v1beta2`)
- [x] Matches pattern of previously merged AWS relationship PRs

Screenshot 2026-05-29 at 12 45 15 AM

### DIFF
--- a/server/meshmodel/aws-dynamodb-controller/v1.8.0/v1.0.0/relationships/edge-non-binding-reference-table-deployment.json
+++ b/server/meshmodel/aws-dynamodb-controller/v1.8.0/v1.0.0/relationships/edge-non-binding-reference-table-deployment.json
@@ -1,0 +1,105 @@
+{
+  "id": "00000000-0000-0000-0000-000000000000",
+  "evaluationQuery": "",
+  "kind": "edge",
+  "metadata": {
+    "description": "An edge relationship between a DynamoDB Table and a Kubernetes Deployment, representing a Deployment whose containers read from or write to the DynamoDB table.",
+    "styles": {
+      "primaryColor": "",
+      "svgColor": "",
+      "svgWhite": ""
+    },
+    "isAnnotation": false
+  },
+  "model": {
+    "version": "",
+    "name": "aws-dynamodb-controller",
+    "displayName": "AWS DynamoDB",
+    "id": "00000000-0000-0000-0000-000000000000",
+    "registrant": {
+      "kind": "github"
+    },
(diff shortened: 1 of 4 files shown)


YASHMAHAKAL (changes requested):
registrant shouldn't be `artifacthub` in this particular case @aahilsayed062

aahil62:
Thanks for the review! Addressing the feedback:

1. Will add displayName to the outer model block for both DynamoDB and SQS models.
2. @YASHMAHAKAL — could you clarify which registrant.kind to use for aws-dynamodb-controller and aws-sqs-controller? I followed the same value used in #17372 (Kinesis). Should it be 'github' instead?
3. On the env mutatedRef — I followed the same pattern as #17372 which was merged. These are design-time visual relationships in Meshery's canvas, not runtime patches. Happy to revisit if there's a preferred alternative pattern.

aahil62:
Addressed all feedback — set registrant.kind to 'github' for the from selector (confirmed by @YASHMAHAKAL), added displayName for both models, and updated the from selector patch ref to metadata.name. Thanks for the explanation on registrants!

aahil62:
Also applied the configuration prefix to all patch paths in the latest commit — same fix as #19502. Ready for second pass.

aahil62:
addressed in the v1beta2 schema fix commit — all 4 reference files (table-deployment, table-pod, queue-deployment, queue-pod) now use `"registrant": {"kind": "github"}` throughout (top-level model, from-selector, and to-selector). could you take another look? thanks

chellej (approved):

PR #19535, opened 2026-05-18
perf: replace inefficient binary serialization with base64 encoding

rudrakshSoni-dev:
## Summary

This PR replaces inefficient binary-to-numeric-array serialization in upload flows with chunked Base64 encoding.

Previously, uploads used patterns like:

```ts
Array.from(new Uint8Array(buffer))
```

which caused:

* large heap allocations
* significant payload inflation
* synchronous main-thread blocking
* browser instability for larger uploads

## Changes

* Added `ui/utils/binary.ts` with chunked Base64 utilities
* Refactored filter upload flows to use async Base64 encoding
* Updated model import upload flow
* Added `encoding: 'base64'` metadata to upload payloads
* Removed upload-specific numeric array serialization logic

## Performance Impact

The new implementation:

* avoids large intermediate JS numeric arrays
* reduces memory pressure
* minimizes UI blocking during uploads
* improves browser stability for larger files

## Notes

Backend handlers should decode payloads when:

```json
"encoding": "base64"
```

Legacy utility functions in `ui/utils/utils.tsx` remain for now because they may still be referenced by unrelated registry flows and are outside the scope of this focused refactor.

### DIFF
--- a/ui/components/filters/Filters.fileActions.test.tsx
+++ b/ui/components/filters/Filters.fileActions.test.tsx
@@ -137,7 +137,7 @@ describe('createHandleSubmit', () => {
     const body = JSON.parse(uploadFilterFile.mock.calls[0][0].uploadBody);
     expect(body).toEqual({
       save: true,
-      filterData: { filterFile: 'binary', name: 'foo' },
+      filterData: { filterFile: 'binary', name: 'foo', encoding: 'base64' },
       config: { x: 1 },
     });
   });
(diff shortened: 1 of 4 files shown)


hamza-mohd:
Significant performance improvement - chunked Base64 encoding avoids the heap allocation and main-thread blocking that came with `Array.from(new Uint8Array(buffer))`. The problem statement is well-documented.

hamza-mohd:
Good improvement. Please include benchmark results in the PR description so reviewers can quantify the performance gain.

hamza-mohd:
Interesting direction. Could you share benchmark numbers (size and time) plus a quick description of what the prior path was doing? Base64 typically costs ~33% more bytes, so context will help reviewers reason about the swap.

PR #19537, opened 2026-05-18
chore/update org reference

ritzorama:
- **update org extensions-packages reference**
- **update org extensions-packages reference**


pontusringblom (approved):

hamza-mohd:
Good hygiene - updating org references consistently across 43 files prevents broken links and stale redirects.

PR #19542, opened 2026-05-19
[UI] Fix performance dashboard crash on null test_start_time

hortison:
## Summary

The performance dashboard at `/performance` was crashing with:

```
Error: (0 , i.default)(...).utcOffset(...).format is not a function
```

In `PerformanceCalendar.tsx`, calendar events were generated with:

```js
moment(test_start_time).utcOffset(test_start_time).format('MM/DD/YYYY HH:mm')
```

When `test_start_time` is `null` or `undefined` (the GraphQL field is a nullable `String`), `moment().utcOffset(null|undefined)` falls through to its **getter** branch (`input != null` is false for both) and returns a number — the offset in minutes — instead of a moment. The follow-up `.format()` then throws, taking the whole calendar render down.

Switched to `moment.parseZone(test_start_time).format(...)`, which:
- Preserves the timezone offset embedded in the original timestamp (the intent of the original code).
- Returns a moment object regardless of input, so the `.format()` chain stays safe.

## Test plan
- [ ] Load `/performance` and confirm the dashboard renders without the runtime error.
- [ ] Verify performance results appear in the calendar at the correct local time.
- [ ] Confirm calendar still renders when result rows have a null `test_start_time`.

---
_Generated by [Claude Code](https://claude.ai/code/session_015maMQ79S9FJrw83ua5qJG2)_

### DIFF
--- a/ui/components/performance/PerformanceCalendar.tsx
+++ b/ui/components/performance/PerformanceCalendar.tsx
@@ -36,9 +36,11 @@ const localizer = momentLocalizer(moment);
  */
 function generateCalendarEventsFromResults(results) {
   return results.map(({ test_start_time, name, runner_results }, index) => {
-    // Remove incorrect timezone info
+    // parseZone preserves the offset embedded in the timestamp instead of
+    // converting to local time, and tolerates a missing test_start_time
+    // (utcOffset(null|undefined) returns a number, not a moment).
     const ntzStartTime = new Date(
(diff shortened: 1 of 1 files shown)


leecalcote (approved):

hamza-mohd:
Good defensive fix - guarding against `null`/`undefined` `test_start_time` before calling moment prevents a hard crash on the performance dashboard. Clean, targeted change.

PR #19549, opened 2026-05-19
chore: consolidate CI workflows and enhance end-to-end testing setup

ianrwhitney:
**Notes for Reviewers**
This PR adds the reusable workflow to publish allure results and wires it up to the build and test workflow. It also deprecates the old build meshery ui and server workflow in favor of the decomposed workflow. 

The triggers are now pull_request and push. On pull_request, with the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository. Therefore, the underlying workfows must work without the secret such is the case with the e2e testing which only test local provider. The publish allure report is only triggered on push to master.


hamza-mohd:
Good consolidation - the -220 deletions removing the old build workflow are a great sign. Wiring the reusable Allure results workflow in the same PR keeps the change coherent.

hamza-mohd:
Excellent infrastructure improvement. Please verify the consolidated workflow triggers on all the same events as the previous separate workflows.

hamza-mohd:
Solid direction. Please call out which workflows were consolidated so maintainers can spot any coverage we may be dropping accidentally.

leecalcote (approved):

lekaf974 (approved):
LGTM

PR #19561, opened 2026-05-20
ci: add additional workflows for E2E report on pull request comment

ianrwhitney:
**Notes for Reviewers**
The e2e test report was not posting. This is d/t the name change as we are decomposing and cleaningup workflows. I added the new names to this workflow and will delete after the workflows get cleaned up.


hamza-mohd:
Good housekeeping - wiring the new workflow names into the E2E report trigger unblocks test result visibility quickly.

PR #19570, opened 2026-05-20
[Server] Align system API with canonical schema

PragalvaXFREZ:
## Summary

Migrates Meshery Server/UI consumers for the schema-backed `/api/system/*` operational surface introduced in the dependent schema PR.

| Method | Endpoint | Migration scope |
| --- | --- | --- |
| `GET` | `/api/system/database` | Uses the generated system database query and canonical `pageSize`, `sort`, `order`, `totalTables` shape. |
| `DELETE` | `/api/system/database/reset` | Uses the generated reset mutation from the Database Summary UI. |
| `GET` | `/api/system/version` | Uses the generated `SystemVersion` model; confirms `releaseChannel` as the wire field. |
| `GET` | `/api/system/sync` | Uses the generated session-sync model while preserving the open preference payload. |

Depends on: https://github.com/meshery/schemas/pull/897

## Review Notes

- Please review the reset semantics: the Database Summary UI now calls REST `DELETE /api/system/database/reset`. That REST path drops all tables, including `events`; the older GraphQL hard-reset path skipped the `events` table. Confirm that purging event history is intended, or this PR should preserve the old behavior.
- Please review `/api/system/sync` timestamp optionality. `models.K8sContext` has pointer timestamps, while the generated schema currently uses non-pointer `time.Time` aliases for optional `createdAt` / `updatedAt`; edge cases with nil timestamps should not emit zero-time values.

## Validation

- Targeted RTK system tests pass locally.
- Server-side checks pass only with the local schemas checkout wired in; clean checkout validation depends on the schemas PR/artifact availability.


leecalcote:
All checks failing.

hamza-mohd:
Good alignment work. The migration table is a helpful review aid, and anchoring to the canonical schema should prevent drift going forward.

server/handlers/database_handlers.go
```diff
@@ -73,6 +80,53 @@ func (h *Handler) GetSystemDatabase(w http.ResponseWriter, r *http.Request, _ *m
...
+	if pageSize == "" {
+		pageSize = urlValues.Get("pagesize")
+	}
+	if pageSize == "" {
+		limit = defaultPageSize
+	} else {
+		limit, err = strconv.Atoi(pageSize)
+		if err != nil || limit < 1 {
+			return 0, 0, 0, "", "", false, fmt.Errorf("pageSize must be an integer greater than 0")
+		}
```
lekaf974, line 101: Looks this could be embedded in a reusable function to be called from other handlers

server/handlers/database_handlers.go
```diff
@@ -21,7 +23,11 @@ func (h *Handler) GetSystemDatabase(w http.ResponseWriter, r *http.Request, _ *m
...
 	var recordCount int
 	var totalTables int64
-	page, offset, limit, search, order, sort, _ := getPaginationParams(r)
+	page, offset, limit, search, sortColumn, sortDesc, err := getSystemDatabaseQueryParams(r)
+	if err != nil {
+		writeJSONError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
 
 	tableFinder := h.dbHandler.DB.Table("sqlite_schema").
```
lekaf974, line 32: Wondering if we could take advantage of being there to extract Data layer logic access outside the handler to have a more robust implementation  ?

lekaf974 (commented):
Small comments and golanci-lint is faling

PR #17013, opened 2026-01-18
Prevent server panic due to race condition in event broadcaster

Ady0333:
**Notes for Reviewers**

- This PR fixes #17011

- The change is intentionally minimal and targeted to the event broadcasting path.

- No public APIs or external behavior are modified.
 
- The fix focuses solely on concurrency safety during publish/unsubscribe interactions.
 
- No refactors or unrelated cleanups are included.

- The panic was caused by a TOCTOU race between Publish() and the unsubscribe closure.

- The previous utils.IsClosed() check was ineffective and unsafe under concurrency.
 
- This PR removes that check and uses a safe, synchronized publish pattern instead.
 
- The issue is non-deterministic and difficult to reproduce reliably, but the fix follows standard Go concurrency practices.

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [ ✔️] Yes, I signed my commits.

PR #19572, opened 2026-05-20
[server] Fix Broadcast races and value-copy clobber in event_broadcast

### LINKED ISSUE #17011: Server panic due to race condition in event broadcaster when publishing to closed channels
### Current Behavior
The Meshery server can panic and crash due to a race condition in the event broadcaster when publishing events to client channels.

Specifically, the Publish() method in server/models/event_broadcast.go checks whether a listener channel is closed using utils.IsClosed() and then sends data to the channel. However, this check is unsafe and subject to a time-of-check-time-of-use (TOCTOU) race. If a concurrent unsubscribe operation closes the channel after the check but before the send, the server panics with:
`panic: send on closed channel`

This crash terminates the Meshery server process, disconnecting all users and aborting in-progress operations.

Additionally, the helper function utils.IsClosed() is itself incorrect: it may consume buffered channel data and incorrectly report channels as closed, leading to silent event loss and false safety checks.

### Expected Behavior

- The Meshery server should not panic when clients disconnect or unsubscribe from event streams.
- Event publishing should be safe under concurrent publish and unsubscribe operations.
- Client disconnections (browser tab close, network loss, context cancellation) should be handled gracefully without affecting server stability.
- Event data should not be lost due to incorrect channel state checks.

### Logs
```
panic: send on closed channel

goroutine 451 [running]:
github.com/meshery/meshery/server/models.(*Broadcast).Publish(...)
    server/models/event_broadcast.go:72
github.com/meshery/meshery/server/models/eventsResolver.Events(...)
    server/models/events_resolver.go:58
```

### Environment

- **Meshery Server Version:** current master / stable release
- **Meshery Client Version:** stable release

### To Reproduce
1. Start Meshery server 
2. Open Meshery UI and establish an events or GraphQL subscription 
3. Trigger frequent event publishing (normal Meshery operations) 
4. Close the browser tab or disconnect the client 
5. Observe server panic due to concurrent publish and unsubscribe

---

SarthakB11:
Fixes #17011

### Context and prior art

This PR addresses #17011 (server panic on event broadcaster races, open since 2026-01-18). PR #17013 (Ady0333) was approved and had its merge conflict resolved by @leecalcote, but two E2E tests failed and the author self-closed it on 2026-03-18 without follow-up. The issue is still assigned to @Ady0333 but has had no activity in over four months; happy to hand back or split the work if they want to pick it up again.

### Three concurrency bugs in `server/models/event_broadcast.go`

1. `utils.IsClosed` (`server/helpers/utils/utils.go:151`) does `case <-ch: return true` and drains a queued event from a buffered, still-open channel. The Publish gate at the old `event_broadcast.go:77` then reports the channel closed and drops the send. Events silently lost on bursty subscribers.

2. `Subscribe` stored `clients` by value in `sync.Map` and re-`Store`'d a local copy after appending to `listeners`. The `*sync.Mutex` was shared, but the slice header was not, so two racing subscribers for the same id could both read the same pre-append header and last-writer-wins on `Store`, losing one subscriber. PR #17013 did not address this bug.

3. `Publish` iterated `listeners` with no lock while `unsubscribe` closed the channel mid-iteration; concurrent unsubscribe + publish could send on a closed channel and panic (this is the panic the issue's stack trace shows).

### Fix

- Store `*clients` in the `sync.Map` so the mutex actually protects the slice header. Eliminates the value-copy clobber.
- Embed `sync.Mutex` directly in `clients` (no longer a pointer; the struct is no longer copied).
- In `Publish`, hold `cc.mu` and use a non-blocking `select { case client <- data: default: }`. Drops the `IsClosed` gate, which is wrong for buffered channels.
- In `unsubscribe`, rebuild `listeners` without `ch` under `mu`, release the lock, then `close(ch)`. Publish either runs before the rebuild (sees the listener, sends safely under lock) or after (slice no longer contains ch), never racing the `close`.

### Delta versus PR #17013

- #17013 used `defer recover()` around each send to catch send-on-closed-channel panics. This works but adds per-send overhead and treats the symptom rather than the root cause. This PR makes the race impossible by serializing Publish iteration with the unsubscribe slice rebuild and closing only after the listener is removed.
- #17013 snapshotted listeners under lock then sent outside the lock. This PR sends under the lock with a non-blocking select, which avoids the snapshot allocation. Since the inner send is constant-time (default branch), holding the lock across iteration is bounded.
- #17013 did not change `Subscribe` or the `clients` struct. This PR fixes the value-copy lost-update bug there as well.

Diff keeps the `Broadcast.Subscribe` / `Broadcast.Publish` public API identical.

### Follow-ups (separate issue, not this PR)

The remaining `utils.IsClosed` callers have the same drop-an-event bug:

- `server/models/k8scontext_helper.go:46`
- `server/models/dashboard_k8s_resources.go:29`
- `server/models/meshmodel/helper.go:28`
- `server/models/configuration_helper.go:32, 46, 60`

`IsClosed` itself should be removed or replaced with a non-blocking send pattern. Left in place here to keep the diff scoped to the broadcaster.

### Verification

- `gofmt` clean on the changed file.
- Manual review: every `clients` access in `Subscribe`, `Publish`, and `unsubscribe` is now under `cc.mu`; `close(ch)` happens strictly after the listener is removed from the slice.
- E2E test stability was the prior blocker on #17013; would appreciate a maintainer ack to re-run the suite against this version since the failures may have been environmental.

**Notes for Reviewers**

- This PR fixes #17011

### DIFF
--- a/server/models/event_broadcast.go
+++ b/server/models/event_broadcast.go
@@ -3,14 +3,12 @@ package models
 import (
 	"sync"
 
-	"github.com/meshery/meshery/server/helpers/utils"
 	"github.com/meshery/schemas/models/core"
-	"github.com/sirupsen/logrus"
 )
 
 type clients struct {
 	listeners []chan interface{}
-	mu        *sync.Mutex
+	mu        sync.Mutex
 }
 
 type Broadcast struct {
@@ -19,63 +17,53 @@ type Broadcast struct {
 }
 
 func (c *Broadcast) Subscribe(id core.Uuid) (chan interface{}, func()) {
-	clientMap, err := c.clients.LoadOrStore(id, clients{mu: &sync.Mutex{}})
-	if err {
-		logrus.Infof("Client for id %s does not exist in clients map", id)
-	}
+	actual, _ := c.clients.LoadOrStore(id, &clients{})
+	cc := actual.(*clients)
+
 	ch := make(chan interface{}, 1)
-	connectedClient, err := clientMap.(clients)
 
-	if err {
-		logrus.Infof("Client for id %s is not connected. Attempting to connect to client %s.", id, id)
-	}
+	cc.mu.Lock()
+	cc.listeners = append(cc.listeners, ch)
+	cc.mu.Unlock()
 
-	connectedClient.mu.Lock()
-	connectedClient.listeners = append(connectedClient.listeners, ch)
-	connectedClient.mu.Unlock()
-
-	c.clients.Store(id, connectedClient)
 	unsubscribe := func() {
-		cclient, ok := c.clients.Load(id)
-		var client clients
-		if ok {
-			client, ok = cclient.(clients)
-			if !ok {
-				return
-			}
+		v, ok := c.clients.Load(id)
+		if !ok {
+			return
 		}
-		client.mu.Lock()
-		defer client.mu.Unlock()
+		cc := v.(*clients)
 
-		listeners := client.listeners
-		updatedListeners := []chan interface{}{}
-		for _, client := range listeners {
-			if client == ch {
-				close(client)
-				// break
-			} else {
-				updatedListeners = append(updatedListeners, client)
+		cc.mu.Lock()
+		updated := make([]chan interface{}, 0, len(cc.listeners))
+		for _, l := range cc.listeners {
+			if l != ch {
+				updated = append(updated, l)
 			}
 		}
-		client.listeners = updatedListeners
-		c.clients.Store(id, client)
+		cc.listeners = updated
+		cc.mu.Unlock()
+
+		close(ch)
 	}
 	return ch, unsubscribe
 }
 
 func (c *Broadcast) Publish(id core.Uuid, data interface{}) {
-	clientMap, ok := c.clients.Load(id)
+	v, ok := c.clients.Load(id)
+	if !ok {
+		return
+	}
+	cc, ok := v.(*clients)
 	if !ok {
 		return
 	}
 
-	clientToPublish, ok := clientMap.(clients)
-	if !ok {
-		return
-	}
-	for _, client := range clientToPublish.listeners {
-		if !utils.IsClosed(client) {
-			client <- data
+	cc.mu.Lock()
+	defer cc.mu.Unlock()
+	for _, client := range cc.listeners {
+		select {
+		case client <- data:
+		default:
 		}
 	}
 }
--- a/server/models/event_broadcast.go
+++ b/server/models/event_broadcast.go
@@ -31,19 +31,27 @@ func (c *Broadcast) Subscribe(id core.Uuid) (chan interface{}, func()) {
 		if !ok {
 			return
 		}
-		cc := v.(*clients)
+		cc, ok := v.(*clients)
+		if !ok {
+			return
+		}
 
 		cc.mu.Lock()
+		found := false
 		updated := make([]chan interface{}, 0, len(cc.listeners))
 		for _, l := range cc.listeners {
-			if l != ch {
-				updated = append(updated, l)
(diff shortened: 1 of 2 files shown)


server/models/event_broadcast.go
```diff
@@ -19,63 +17,53 @@ type Broadcast struct {
...
+			if l != ch {
+				updated = append(updated, l)
 			}
 		}
-		client.listeners = updatedListeners
-		c.clients.Store(id, client)
+		cc.listeners = updated
+		cc.mu.Unlock()
+
+		close(ch)
```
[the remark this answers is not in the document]
SarthakB11: Fixed in c53399e8af. `unsubscribe` only closes `ch` if it was actually removed from the slice; double-unsubscribe is a no-op now. Type assertion uses ok-idiom.

SarthakB11:
The failing E2E tests look unrelated. Same 4 of 5 are failing on an unrelated UI-only PR (run 26206642497). All Go checks pass.

server/models/event_broadcast.go
```diff
@@ -19,63 +17,61 @@ type Broadcast struct {
...
 		return
 	}
-	for _, client := range clientToPublish.listeners {
-		if !utils.IsClosed(client) {
-			client <- data
+
+	cc.mu.Lock()
+	defer cc.mu.Unlock()
+	for _, client := range cc.listeners {
+		select {
```
chaitanyamedidar, line 77: This non-blocking send prevents slow subscribers from blocking all publishers, which makes sense for stability. Since it also drops events when the subscriber buffer is full, can we document that delivery is best-effort here or add a focused test that locks in this behavior?

chaitanyamedidar (changes requested):
This looks like a useful fix for #17011. 

The main thing I’d like to see before maintainer's review is focused coverage for the broadcaster itself: concurrent publish/unsubscribe, multiple subscribers for the same ID, and double unsubscribe.

SarthakB11:
Added in 466533cc11: multi-subscribe, double-unsubscribe, and a `-race`-friendly concurrent publish/unsubscribe.

server/models/event_broadcast.go
SarthakB11: Documented best-effort delivery on `Publish` and added `TestBroadcast_BestEffortDeliveryWhenBufferFull` in 41b3c4f91.

hamza-mohd:
Thorough fix for the broadcast race - mutex protection and non-blocking sends are the right approaches here. Good that you referenced #17013 and the test failures that blocked the earlier attempt.

hamza-mohd:
Excellent, well-documented concurrency fix - appreciated the detailed breakdown of all three race conditions and the comparison against the prior approach. Consider adding a `go test -race` CI step targeting this package to catch future regressions.

hamza-mohd:
Important fix, races in broadcast logic are tough to chase later. Was this caught with `go test -race`? If so, please keep the repro as a regression test.

SarthakB11:
`TestBroadcast_ConcurrentPublishUnsubscribe` in `server/models/event_broadcast_test.go` is the `-race` regression test, plus `TestBroadcast_DoubleUnsubscribe` and `TestBroadcast_BestEffortDeliveryWhenBufferFull` lock in the specific bugs. Adding `-race` to the meshery CI workflow itself would be a separate PR.

chaitanyamedidar:
@SarthakB11 Someone else is still assiged to resolve the issue, did you discuss this with any maintainer or on [Slack](https://layer5.io/meshery/)?

SarthakB11:
This is already fixed on master by Lee Calcote's `201346cc` (via #20375) — same value-copy clobber fix plus the broadcaster tests. Closing as superseded.

PR #19573, opened 2026-05-20
[Server] fix: add kind="kubernetes" validation before MeshSync deployment mode change in UpdateConnectionById

### LINKED ISSUE #19396: [server] Missing kind="kubernetes" validation before MeshSync deployment mode change in UpdateConnectionById
### Current Behavior
In `UpdateConnectionById` (connections_handlers.go, line 385), 
when a MeshSync deployment mode is specified in the request, 
the handler calls `handleMeshSyncDeploymentModeChange` without 
checking if the connection kind is "kubernetes".

### Expected Behavior
Before calling `handleMeshSyncDeploymentModeChange`, the handler 
should verify that `connection.Kind == "kubernetes"` and return 400 Bad Request, if a non-kubernetes kind is explicitly provided.

### Root Cause
The TODO comment at line 385 acknowledges this missing check:
`// TODO: also check that kind = "kubernetes"`

### Environment
- **Host OS:** Mac

jahnavigajjala-3:
## Description
In `UpdateConnectionById` (connections_handlers.go, line 385), 
the handler called `handleMeshSyncDeploymentModeChange` without 
checking if the connection kind is "kubernetes". This means 
non-kubernetes connections such as Grafana or Prometheus with a 
`meshsync_deployment_mode` in metadata could incorrectly trigger 
a MeshSync deployment mode change.

The TODO comment at line 385 acknowledged this missing check:
`// TODO: also check that kind = "kubernetes" (when client starts 
to send full connection object)`

This PR adds an early validation check before calling 
`handleMeshSyncDeploymentModeChange`:
- Returns `400 Bad Request` if `connection.Kind` is explicitly 
  set to a non-kubernetes value
- Skips the check when `connection.Kind` is empty to maintain 
  backward compatibility with clients that don't send the full 
  connection object yet
- Removes the TODO comment since it is now resolved
- Consistent with how other handlers validate input before 
  performing infrastructure operations

## Notes for Reviewers
- This PR fixes #19396

## Signed commits


hamza-mohd:
Focused guard - validating `kind == 'kubernetes'` before calling `handleMeshSyncDeploymentModeChange` correctly prevents unexpected behavior for Grafana/Prometheus connections.

jahnavigajjala-3:
Thanks for the review, @hamza-mohd! It looks like the tests-ui-e2e workflow failed on an unrelated step. Could you please trigger a re-run of the failed CI jobs or provide an official approval if the logic looks good to go?

hamza-mohd:
Good defensive check. Please verify that similar kind validation is applied in any other handlers that modify connection state.

hamza-mohd:
Sensible guard. Please add a test covering the rejection path so regressions are caught early.

jahnavigajjala-3:
Thanks for the review, @hamza-mohd! I'll check if similar kind validation is needed in other handlers that modify connection state, and add a test covering the rejection path. Will update the PR shortly.

jahnavigajjala-3:
@hamza-mohd , I checked  and found that  handleMeshSyncDeploymentModeChange is called only in one place in connections_handlers.go, so no other handlers require similar kind validation and added a unit test TestUpdateConnectionById_NonKubernetesKindRejected in server/handlers/connections_handlers_test.go that verifies a  400 Bad Request is returned when a non-kubernetes kind (e.g- "grafana") is provided along with a meshsync deployment mode

PR #19576, opened 2026-05-21
[Docs] Incomplete migration to npm-managed Hugo in docs/Makefile 

### LINKED ISSUE #19575: [Docs] Incomplete migration to npm-managed Hugo in docs/Makefile
### Current State

Issue #19363 highlighted a discrepancy between the Makefile targets and the npm scripts for running Hugo. PR #19558 partially resolved this by updating the `site:` target in `docs/Makefile` to use `npx hugo`, ensuring it relies on the local `hugo-extended` binary in `node_modules` rather than a global installation.

However, the migration is incomplete. Two other **local development** targets in `docs/Makefile` still invoke a bare `hugo` command:

- `build:` → `hugo`
- `clean:` → `hugo --cleanDestinationDir`

If a contributor follows the recommended setup (`npm install`) and lacks a global Hugo installation, running `make build` or `make clean` results in a `command not found` error. Additionally, even when a global Hugo *is* installed, it may differ from the version pinned in `docs/package.json` (`hugo-extended: 0.157.0`), leading to inconsistent build output.

### Desired State

Update the remaining local development targets in `docs/Makefile` (`build` and `clean`) to use `npx hugo` instead of bare `hugo`. This ensures:

1. Contributors only need `npm install` — no global Hugo installation required.
2. The pinned `hugo-extended` version from `package.json` is always used, preventing version drift.
3. Consistency with the `site:` target already updated in PR #19558.

---

hebypaul:
**Notes for Reviewers**

This PR completes the migration to the npm-managed Hugo binary started in PR #19558. 

The `site:` target was updated to use `npx hugo` in #19558, but the `build:` and `clean:` targets still invoked a bare `hugo` command. This caused a `command not found` error for contributors who rely on `npm install` (as recommended in the setup instructions) and do not have Hugo installed globally. 

This PR replaces the remaining bare `hugo` calls with `npx hugo` so that the `hugo-extended` binary from `node_modules` is always used, ensuring version consistency with `docs/package.json`.

*Note: Docker targets (`docker-build`, `docker-build-upstream`, etc.) are intentionally left unchanged. The `docs/Dockerfile` installs Hugo globally inside the container (`/usr/local/bin/hugo`), so bare `hugo` is correct in that context.*

- This PR fixes #19575


docs/Makefile
```diff
@@ -24,11 +24,11 @@ site: check-go
 
 ## Build docs.meshery.io on your local machine.
 build:
-	hugo
+	npx hugo
```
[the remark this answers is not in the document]
hebypaul: Thank you for the review . 

 I agree that delegating the root Makefile targets to docs/Makefile (e.g., using $(MAKE) -C docs <target>) is the best long-term architectural solution .

However, since that requires modifying the root Makefile and adding new targets to docs/Makefile (like serve and build-production), it expands the scope of this PR significantly beyond just fixing the local docs/ targets.

Would you prefer that I implement the full Make delegation in this PR, or should we merge this minimal fix first and track the root Makefile refactor in a separate issue?  

 happy to implement it here if I have a maintainer's green light .

hebypaul:
Can you review these changes @banana-three-join

banana-three-join:
@hebypaul You can follow the advice provided by the LLM and update the root file so it isn't a partial migration and npx hugo is used everywhere without copying and pasting it over and over again

hebypaul:
I should be focusing on implementing these changes on local build only Right ? 

updating only the local dev targets in the `root Makefile` to delegate to the `docs/Makefile`, but leave `docs-build-production ` exactly as it is  to ensure your production pipelines and Docker builds continue to work properly without suddenly requiring Node.js.

 @banana-three-join

banana-three-join:
Yup! @hebypaul as the problem in itself is that following the tutorial for local builds while having (or not having) a global instance of Hugo installed may cause issues due to different versioning from the one being utilized by the project.

hamza-mohd:
Good catch - completing the `hugo` to `npx hugo` migration consistently across `build:` and `clean:` prevents confusing `command not found` errors for contributors.

docs/Makefile
leecalcote: Is `npx` included in the dep-check?

leecalcote:
Why do we need an `npx`-managed Hugo?

banana-three-join:
@leecalcote An issue was started because contributors would present issues when running the `hugo` command due to different versioning (in case they already had installed due to the command calling the a global variable), and since it's installation is currently being managed through node, (aside from the Docker builds) the npx usage was proposed to standarize the version of Hugo being utilized. 

Also, there's no dep-check for Hugo right now (as long as I'm concerned but correct me if I'm wrong).

hebypaul:
@leecalcote 

We need an `npx`-managed Hugo to **align the project's codebase with its own official setup instructions and guarantee deterministic builds**.

 According to the [Meshery Docs Contribution Guide](https://docs.meshery.io/project/contributing/contributing-docs/), contributors are explicitly told: *"Hugo is installed automatically as an npm dependency, so you only need Node.js and Go installed on your system."* Contributors then run `make setup` (which runs `npm install`).

However, before this PR, running `make build` or `make clean` invoked a bare `hugo` command. Because the contributor followed the instructions and *didn't* install Hugo globally on their OS, the terminal throws a `command not found: hugo` error.
Screenshot 2026-05-25 162740

 The PR fixes this broken developer experience by switching the remaining bare `hugo` commands in the local dev `Makefile` to `npx hugo`, ensuring it leverages the exact binary the setup script just downloaded.

Screenshot 2026-05-25 163347


I propose this to improve *Developer experience* eliminates a massive friction point for open-source onboarding , Alternative solution wil be to  update [Meshery Docs Contribution Guide](https://docs.meshery.io/project/contributing/contributing-docs/) with correct details so there will be no confusion. Listing Hugo installed globally as prerequisite  could solve this.

I don't know much about how this will affect `CI`, so a Review is much appreciated .

hebypaul:
`Check-deps` command added to makefile . This `check-deps` command is now used in `build` , `site` and `clean` commands . 

✅ Successful Check (All dependencies present)
---
Screenshot 2026-06-01 120634

❌ Missing Node.js/npm
---
Screenshot 2026-06-01 121404


❌ Missing Hugo
---
Screenshot 2026-06-01 120612

CodexRaunak (dismissed):

docs/Makefile
```diff
@@ -18,17 +18,28 @@
...
 
+## Run docs.meshery.io on your local machine in serve mode (without file watching).
+serve: check-deps check-go
+	npx hugo server -D -F --watch=false
+
 ## Build docs.meshery.io on your local machine.
-build:
-	hugo
+build: check-deps check-go
+	npx hugo
```
CodexRaunak: @hebypaul can we use npm scripts and reference in the make file instead of using npx?

hebypaul: I added the standard scripts to `docs/package.json` so npm can automatically resolve the local binaries, and refactored `docs/Makefile` to use `npm run <script>` instead of `npx`.

I just pushed these changes. Could you take a look when you have a chance?

hebypaul (commented):
Looks good to me! Thanks for the tweaks, feel free to merge.

CodexRaunak (approved):

PR #19580, opened 2026-05-21
[server] Replace remaining broken utils.IsClosed gates with non-blocking sends

SarthakB11:
The `IsClosed[K]` helper at `server/helpers/utils/utils.go:151` does `case <-ch: return true` on the passed channel. For a buffered channel that is open but has a queued value, the receive branch fires and the value is dropped. The `Publish*` helpers that use it as a gate then report the channel closed and skip the send, so events are silently lost.

This is the follow-up I flagged in #19572 for the remaining five `utils.IsClosed` callers outside `event_broadcast.go`.

### Changes

- `server/models/configuration_helper.go`: `PublishApplications`, `PublishPatterns`, `PublishFilters`
- `server/models/dashboard_k8s_resources.go`: `PublishDashboardK8sResources`
- `server/models/meshmodel/helper.go`: `SummaryChannel.Publish`
- `server/models/k8scontext_helper.go`: `PublishContext` (the inner select was already non-blocking; only the redundant `IsClosed` gate is removed)

Each gate is replaced with:

```go
for _, ch := range listeners {
    select {
    case ch <- struct{}{}:
    default:
    }
}
```

`utils.IsClosed` itself is left in place because `event_broadcast.go` still references it on master. Removal will be a follow-up once #19572 merges and no callers remain.

### Why non-blocking

None of these four helper types expose an unsubscribe path that calls `close(ch)`. Subscribers register a channel and are responsible for their own channel lifecycle, so a send on a closed channel would still panic; that case is not introduced by this PR and is not supported by these helpers either before or after. The non-blocking select preserves the previous "best-effort delivery" intent, handles slow subscribers without blocking the publisher or other subscribers, and avoids the buffered-channel drain bug entirely.

### DIFF
--- a/server/models/configuration_helper.go
+++ b/server/models/configuration_helper.go
@@ -2,8 +2,6 @@ package models
 
 import (
 	"sync"
-
-	"github.com/meshery/meshery/server/helpers/utils"
 )
 
 type ConfigurationChannel struct {
@@ -29,8 +27,9 @@ func (c *ConfigurationChannel) SubscribeApplications(ch chan struct{}) {
 
 func (c *ConfigurationChannel) PublishApplications() {
 	for _, ch := range c.ApplicationsChannel {
-		if !utils.IsClosed(ch) {
-			ch <- struct{}{}
+		select {
+		case ch <- struct{}{}:
+		default:
 		}
 	}
 }
@@ -43,8 +42,9 @@ func (c *ConfigurationChannel) SubscribePatterns(ch chan struct{}) {
 
 func (c *ConfigurationChannel) PublishPatterns() {
 	for _, ch := range c.PatternsChannel {
-		if !utils.IsClosed(ch) {
-			ch <- struct{}{}
+		select {
+		case ch <- struct{}{}:
+		default:
 		}
 	}
 }
@@ -57,8 +57,9 @@ func (c *ConfigurationChannel) SubscribeFilters(ch chan struct{}) {
 
 func (c *ConfigurationChannel) PublishFilters() {
 	for _, ch := range c.FiltersChannel {
-		if !utils.IsClosed(ch) {
-			ch <- struct{}{}
+		select {
+		case ch <- struct{}{}:
+		default:
 		}
 	}
 }
--- a/server/models/configuration_helper.go
+++ b/server/models/configuration_helper.go
@@ -26,7 +26,12 @@ func (c *ConfigurationChannel) SubscribeApplications(ch chan struct{}) {
 }
 
 func (c *ConfigurationChannel) PublishApplications() {
-	for _, ch := range c.ApplicationsChannel {
+	c.mx.Lock()
+	subscribers := make([]chan struct{}, len(c.ApplicationsChannel))
+	copy(subscribers, c.ApplicationsChannel)
+	c.mx.Unlock()
+
+	for _, ch := range subscribers {
 		select {
 		case ch <- struct{}{}:
 		default:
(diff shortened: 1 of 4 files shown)


server/models/meshmodel/helper.go
```diff
@@ -25,8 +23,9 @@ func (c *SummaryChannel) Subscribe(ch chan struct{}) {
 
 func (c *SummaryChannel) Publish() {
 	for _, ch := range c.channel {
```
Amr-Shams: you need to make sure of using mutex here

server/models/configuration_helper.go
```diff
@@ -29,8 +27,9 @@ func (c *ConfigurationChannel) SubscribeApplications(ch chan struct{}) {
 
 func (c *ConfigurationChannel) PublishApplications() {
 	for _, ch := range c.ApplicationsChannel {
-		if !utils.IsClosed(ch) {
-			ch <- struct{}{}
+		select {
```
Amr-Shams, line 35: the same

Amr-Shams (commented):
just few chagnes and LGTM

hamza-mohd:
Smart fix - replacing the `IsClosed` gate with a non-blocking select correctly handles buffered channels without dropping queued values. Clean, focused change.

server/models/meshmodel/helper.go
SarthakB11: Done in 181f7da7. Copy-under-lock applied to all 5 Publish sites.

server/models/configuration_helper.go
SarthakB11: Done in 181f7da7. Copy-under-lock applied to all 5 Publish sites.

server/models/configuration_helper.go
```diff
@@ -29,8 +27,9 @@ func (c *ConfigurationChannel) SubscribeApplications(ch chan struct{}) {
...
+		case ch <- struct{}{}:
+		default:
 		}
 	}
```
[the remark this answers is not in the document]
SarthakB11: All 5 Publish sites now copy under lock in 181f7da7. Pattern matches PublishContext.

server/models/configuration_helper.go
```diff
@@ -28,9 +26,15 @@ func (c *ConfigurationChannel) SubscribeApplications(ch chan struct{}) {
 }
 
 func (c *ConfigurationChannel) PublishApplications() {
-	for _, ch := range c.ApplicationsChannel {
-		if !utils.IsClosed(ch) {
-			ch <- struct{}{}
+	c.mx.Lock()
+	subscribers := make([]chan struct{}, len(c.ApplicationsChannel))
```
Amr-Shams, line 30: why would you do that ? 
there is a trade off i can see, copy and process in parraell without blocking the parent, or zero copy but block the caller.
but why one over the other here?

SarthakB11: Matches `PublishContext` in `k8scontext_helper.go`, so the four siblings line up with that one. Lock-during-iteration would serialize every Subscribe behind the publish for no real gain; the snapshot keeps the critical section to a slice copy and lets new subscribers register concurrently.

hamza-mohd:
Nice concurrency fix. A brief description of the specific race condition being addressed would help reviewers understand the full impact.

hamza-mohd:
Good follow-through on the IsClosed cleanup. Please confirm no goroutine leaks under load (a quick `-race` run or pprof snapshot would be reassuring).

SarthakB11:
The "## Three concurrency bugs..." section in the PR body has the full bug breakdown. On goroutine leaks: none of the 4 helper types here have an unsubscribe path that closes channels (`grep -rn "close(" server/models/{configuration,dashboard,meshmodel,k8scontext}*.go` is empty), so there is no leak vector introduced by the gate removal. The non-blocking select also can't block a publisher goroutine.

SarthakB11:
Already on master — Lee Calcote's a03fd9a0f070 (#20375) did the same IsClosed-gate removal plus non-blocking sends. Closing as superseded.

PR #19618, opened 2026-05-22
[Server] Rename matchLabels policy to tag

guan404ming:
Fixes follow-up (3) in #19149.

`matchLabels` is Kubernetes terminology, but the policy groups components by any equal-valued configuration field, not just `metadata.labels`. Rename the Go-internal symbols to `tag` to reflect what the policy actually does.

## Changes

- `policy_matchlabels.go` -> `policy_tag.go`: `MatchLabelsPolicy` -> `TagPolicy`, `identifyMatchlabels` -> `identifyTags`, `identifyMatchlabelRelationships` -> `identifyTagRelationships`, `matchLabelGroup` -> `tagGroup`, `maxMatchLabels` -> `maxTags`.
- `engine.go`, `engine_test.go`, `identify_bench_test.go`: update references and the benchmark (`BenchmarkIdentifyTags`) to the new names.

The wire identifier `sibling_match_labels_policy` is kept: it is shared with the Rego engine (`matchlabels-policy.rego`) and seeds `staticUUID`, so renaming it would change generated relationship UUIDs. That rename is deferred until coordinated with meshery/schemas, per the issue. The `sibling` relationship type and the OPA-comparison vocabulary in `catalog_alignment_test.go` are likewise left untouched.

No behavior change; `go test ./server/policies/...` stays green.

### DIFF
--- a/server/policies/engine_test.go
+++ b/server/policies/engine_test.go
@@ -137,7 +137,7 @@ func TestPolicyImplication(t *testing.T) {
 		&HierarchicalParentChildPolicy{},
 		&EdgeNonBindingPolicy{},
 		&AliasPolicy{},
-		&MatchLabelsPolicy{},
+		&TagPolicy{},
 	}
 
 	tests := []struct {
@@ -634,11 +634,11 @@ func TestAliasIsInvalidStatusMatrix(t *testing.T) {
 	}
 }
 
-// MatchLabelsPolicy must produce identical relationship UUIDs across runs for
+// TagPolicy must produce identical relationship UUIDs across runs for
 // the same input. Pre-fix, sibling identification leaked Go map iteration order
 // (groupMap range) and pointer addresses (fmt.Sprintf("%v", selectorSetItem))
 // into the generated IDs.
-func TestMatchLabelsPolicyIdentificationIsDeterministic(t *testing.T) {
+func TestTagPolicyIdentificationIsDeterministic(t *testing.T) {
 	pod := func(idHex string) *component.ComponentDefinition {
 		c := &component.ComponentDefinition{
 			Component: component.Component{Kind: "Pod"},
@@ -670,11 +670,11 @@ func TestMatchLabelsPolicyIdentificationIsDeterministic(t *testing.T) {
 	}
 
 	design := makePatternFile([]*component.ComponentDefinition{podA, podB, podC}, nil)
-	p := &MatchLabelsPolicy{}
+	p := &TagPolicy{}
 
 	first := p.IdentifyRelationship(makeRel(), design)
 	if len(first) == 0 {
-		t.Fatal("expected at least one matchlabels relationship")
+		t.Fatal("expected at least one tag relationship")
 	}
 	for i := 0; i < 5; i++ {
 		again := p.IdentifyRelationship(makeRel(), design)
--- a/server/policies/policy_tag.go
+++ b/server/policies/policy_tag.go
@@ -10,51 +10,57 @@ import (
 	"github.com/meshery/schemas/models/v1beta1/pattern"
 )
 
-// MatchLabelsPolicy handles sibling match-label relationships.
-type MatchLabelsPolicy struct{}
+// TagPolicy handles sibling relationships that group components sharing an
+// equal-valued configuration field. Despite the historical "matchLabels" name
+// (Kubernetes terminology), the grouping is not restricted to metadata.labels;
+// it applies to any equal-valued config field referenced by the selector.
+type TagPolicy struct{}
 
-const maxMatchLabels = 20
+const maxTags = 20
 
-func (p *MatchLabelsPolicy) Identifier() string {
+func (p *TagPolicy) Identifier() string {
+	// Wire identifier shared with the Rego engine (matchlabels-policy.rego) and
+	// used as a staticUUID seed, so it stays stable until renamed in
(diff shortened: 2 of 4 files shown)


server/policies/policy_tag.go
```diff
@@ -10,51 +10,57 @@ import (
...
+
+func (p *TagPolicy) Identifier() string {
+	// Wire identifier shared with the Rego engine (matchlabels-policy.rego) and
+	// used as a staticUUID seed, so it stays stable until renamed in
+	// meshery/schemas. See meshery/meshery#19149 (3).
 	return "sibling_match_labels_policy"
 }
 
-func (p *MatchLabelsPolicy) IsImplicatedBy(rel *relationship.RelationshipDefinition) bool {
+func (p *TagPolicy) IsImplicatedBy(rel *relationship.RelationshipDefinition) bool {
```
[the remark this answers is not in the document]
guan404ming: Holding off here. `rel` is never nil on the real path, `IsImplicatedBy` is only called from `implicableRelationships`, which ranges a non-nil slice. Sibling policies don't guard this either, and this PR is rename-only; defensive nil-hardening should be a separate engine-wide change.

server/policies/policy_tag.go
```diff
@@ -10,51 +10,57 @@ import (
...
 
-func (p *MatchLabelsPolicy) AlreadyExists(rel *relationship.RelationshipDefinition, design *pattern.PatternFile) bool {
+func (p *TagPolicy) AlreadyExists(rel *relationship.RelationshipDefinition, design *pattern.PatternFile) bool {
 	return relationshipAlreadyExists(design, rel)
 }
 
-func (p *MatchLabelsPolicy) IdentifyRelationship(relDef *relationship.RelationshipDefinition, design *pattern.PatternFile) []*relationship.RelationshipDefinition {
-	return identifyMatchlabelRelationships(relDef, design)
+func (p *TagPolicy) IdentifyRelationship(relDef *relationship.RelationshipDefinition, design *pattern.PatternFile) []*relationship.RelationshipDefinition {
+	return identifyTagRelationships(relDef, design)
```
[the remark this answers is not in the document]
guan404ming: Same here. `design` is already dereferenced upstream before dispatch and `relDef` comes from the same non-nil slice, so neither is nil on the real path. Keeping it out of scope for this rename-only PR.

server/policies/engine_test.go
```diff
@@ -137,7 +137,7 @@ func TestPolicyImplication(t *testing.T) {
 		&HierarchicalParentChildPolicy{},
 		&EdgeNonBindingPolicy{},
 		&AliasPolicy{},
-		&MatchLabelsPolicy{},
+		&TagPolicy{},
```
aabidsofi19, line 140: Please use consistent  naminge i.e "TagSet" / "TagSets"

server/policies/engine_test.go
```diff
@@ -634,11 +634,11 @@ func TestAliasIsInvalidStatusMatrix(t *testing.T) {
 	}
 }
 
-// MatchLabelsPolicy must produce identical relationship UUIDs across runs for
+// TagPolicy must produce identical relationship UUIDs across runs for
 // the same input. Pre-fix, sibling identification leaked Go map iteration order
 // (groupMap range) and pointer addresses (fmt.Sprintf("%v", selectorSetItem))
 // into the generated IDs.
-func TestMatchLabelsPolicyIdentificationIsDeterministic(t *testing.T) {
+func TestTagPolicyIdentificationIsDeterministic(t *testing.T) {
```
aabidsofi19, line 641: Please use consistent naminge i.e "TagSet" / "TagSets"

aabidsofi19:
Has the relationship subtype also renamed to tagset

hamza-mohd:
Clean naming improvement. Please double-check that all documentation and policy file references are updated consistently alongside the rename.

hamza-mohd:
Good rename for clarity. Are existing manifests referencing matchLabels still compatible, or is a migration path needed?

guan404ming:
@aabidsofi19 Great question, thanks for raising it. Not yet in this PR. I intentionally scoped this one to the Go-side rename only, since the engine dispatches on `type == "sibling"` rather than on the subType, and the policy identifier `sibling_match_labels_policy` doubles as the staticUUID seed for generated relationships, so I kept it stable to avoid changing existing relationship IDs.

The `subType` value itself lives in the generated meshmodel/schemas definitions. The kubernetes relationship file is already named `sibling-tagsets.json` while its `subType` is still `matchlabels`, so the subtype rename to `tagset` looks like an upstream meshery/schemas change (tracked in #19149). Happy to follow up on that separately, or fold it in here if you'd prefer. Let me know what you'd like.

guan404ming:
@hamza-mohd Thanks for the careful review on both points. This rename is intentionally scoped to the Go symbols only, so the rego policy (`matchlabels-policy.rego`), the wire identifier, and the subType are all deliberately left unchanged. That means there are no docs or policy-file references that need updating here, and existing manifests stay fully compatible with no migration needed: the identifier stays `sibling_match_labels_policy`, the relationship type stays `sibling`, and the subType stays `matchlabels`. The broader naming alignment is tracked upstream in #19149.

server/policies/engine_test.go
aabidsofi19: @guan404ming  please take a look

PR #19621, opened 2026-05-22
Single-source the remote-provider list across install artifacts

hortison:
## Motivation

Production Meshery Playground was found serving a stale `PROVIDER_BASE_URLS=https://cloud.layer5.io` (with `PROVIDER=Layer5`) because its Kubernetes Deployment env was hand-maintained on the CNCF cluster, with no spec in the repo. Investigating the fix surfaced a broader problem: `PROVIDER_BASE_URLS` is duplicated **by hand** across every install artifact, with no shared source, so the values drift. For example, `mesheryctl`'s compiled-in fallback shipped `meshery.digitalocean.com` and `idp.cleverluck.com` - providers no other artifact lists.

## What this does

Introduces `install/providers.env` as the single canonical `Name=URL` source of truth, and derives every artifact from it.

### Single source
```
# install/providers.env
Meshery=https://cloud.meshery.io
Layer5=https://cloud.layer5.io
```

### Derivation
- **`make`**: `Makefile.core.mk` derives `REMOTE_PROVIDER_URLS` from `providers.env`; the duplicated `MESHERY_CLOUD_PROD` / `LAYER5_CLOUD_PROD` vars are removed.
- **Static artifacts**: `install/scripts/sync-provider-urls.py` rewrites the managed `PROVIDER_BASE_URLS` line in each file. Two profiles - `full` (Meshery + Layer5) for standard installs, `playground` (Meshery only) for the hosted demo.
- **`mesheryctl` fallback**: the script emits `mesheryctl/pkg/utils/providers_gen.go` (`const DefaultProviderBaseURLs`, `DO NOT EDIT`); `helpers.go` references it, so the compiled-in fallback inherits the canonical list instead of hardcoding URLs.
- **Drift guard**: `make generate-install` regenerates everything; `make check-install` (new `check-install.yml` workflow) fails CI if any committed artifact is out of sync with `providers.env`.

### Lifecycle (how/when values reach users)
Generation is a commit-time step enforced by CI; there is no runtime read of `providers.env` except by `make`. Values freeze into each artifact when it ships - `mesheryctl` downloads the install files from the repo at the release tag, Helm reads the packaged chart, the extension bakes its compose at build, and the Go fallback is compiled at `go build`.

## Behavior changes
- **Docker Desktop extension**: now ships the full provider set (was Meshery-only); keeps `PROVIDER=Meshery` so Meshery stays pre-selected.
- **`mesheryctl` fallback**: drops the stray `digitalocean` / `cleverluck` entries; now matches the canonical full set.
- `helm install`, `mesheryctl system start` (docker + k8s), and the playground are unchanged in value - verified each ships a populated list (the server has **no** built-in remote-provider fallback, so this matters).

## Also in this PR
- Version-controls the production playground Deployment manifest (`install/playground/kubernetes/playground-deployment.yaml`) that previously existed only on the cluster, plus a README section documenting the CNCF deploy path.
- Exempts `install/providers.env` from the `*.env` gitignore rule so the canonical source stays tracked.
- Design spec: `docs/superpowers/specs/2026-05-22-canonical-provider-urls-design.md`.

## Out of scope / follow-up
- The live CNCF cluster env was corrected separately (`kubectl set env`); this PR makes the repo the source of truth, but the deploy workflow (`cncf-playground-deploy-meshery.yaml`) still only runs `kubectl set image` and does not reconcile env. Wiring an `apply` of the checked-in manifest into that workflow is a sensible follow-up.

## Testing
- `make generate-install` then `make check-install` - clean.
- Negative test: hand-edited an artifact → `check-install` exits non-zero with a diff; restored → clean.
- `go build ./...` + `go vet ./pkg/utils/` in `mesheryctl` - OK.
- All managed YAML files parse.
- `make` prints `REMOTE_PROVIDER_URLS=https://cloud.meshery.io,https://cloud.layer5.io`.


fitzergerald (approved):

simihablo (approved):
Providers list updated. Some older entries removed.

PR #19628, opened 2026-05-22
Profile navigation to new tab

### LINKED ISSUE #17148: Profile icon navigation exits Meshery Playground workspace
### Current Behavior
While using Meshery Playground, clicking on the profile icon redirects the user to the out of the Playground workspace and redirects to profile details page. This behavior exits the active configuration/ design environment unexpectedly and is inconsistent.

### Desired Behavior
Clicking the profile icon should allow access to profile or account settings without leaving the Meshery Playground workspace(for example, through in-context navigation, modal, or new tab), so that design workflow remains uninterrupted.

### Screenshots/Logs

https://github.com/user-attachments/assets/f4bcc8b1-cb74-4cd8-a70d-8ecfa2ac31a3

Image
Image

### Environment

- Browser: Chrome
- Host OS: Windows
- Meshery Server Version: Meshery Playgorund
- Platform: Hosted Playground (Meshery)

Utkarsh-Mauryaa:
**Notes for Reviewers**

- This PR fixes #17148 

-  Opened profile URL in a new tab instead of redirecting in the same tab, so users no longer lose their active Meshery Playground workspace or unsaved configuration when accessing their profile.

## Summary by CodeRabbit

* **Bug Fixes**
  * Profile links now open in a new browser tab/window instead of navigating away from the current page.
  * Added/updated warning behavior when a profile URL isn’t available.
* **Tests**
  * Updated avatar click tests to verify the new-tab behavior and the missing-URL warning.

### DIFF
--- a/ui/components/User.test.tsx
+++ b/ui/components/User.test.tsx
@@ -254,8 +254,8 @@ describe('User component', () => {
     render(<UserProvider />);
 
     await user.click(screen.getByTestId('icon-button-avatar'));
-    expect(notify).toHaveBeenCalledWith({
-      message: 'Profile URL not available. Please try again later.',
-    });
+    await waitFor(() => expect(notify).toHaveBeenCalled());
+    const [payload] = notify.mock.calls[0];
+    expect(payload.message).toBe('Profile URL not available. Please try again later.');
   });
 });
--- a/ui/components/User.test.tsx
+++ b/ui/components/User.test.tsx
@@ -207,8 +207,11 @@ describe('User component', () => {
     expect(payload.details).toBe('oops');
   });
 
-  it('navigates to the profile URL when the avatar is clicked', async () => {
+  it('opens profile URL in a new tab when avatar is clicked', async () => {
     const user = userEvent.setup();
+    const mockOpen = vi.fn();
+    window.open = mockOpen;
+
     mockGetUserQuery = {
       data: { status: 'authenticated' },
       isSuccess: true,
@@ -228,7 +231,11 @@ describe('User component', () => {
     await waitFor(() => expect(ExtensionPointSchemaValidator).toHaveBeenCalledWith('account'));
 
     await user.click(screen.getByTestId('icon-button-avatar'));
-    expect(window.location.href).toContain('https://cloud.test/profile');
+    expect(mockOpen).toHaveBeenCalledWith(
+      'https://cloud.test/profile',
+      '_blank',
+      'noopener,noreferrer',
+    );
   });
 
   it('does not redirect when no profile URL is present', async () => {
@@ -239,12 +246,16 @@ describe('User component', () => {
       isError: false,
       error: undefined,
     };
+    // no extensions.account → profileUrl will be undefined
+    mockProviderCapabilities = {
+      providerUrl: 'https://provider.test',
+    };
 
-    const startingHref = window.location.href;
     render(<UserProvider />);
 
     await user.click(screen.getByTestId('icon-button-avatar'));
-    // window.location.href should still be the same since profileUrl is undefined
-    expect(window.location.href).toBe(startingHref);
+    expect(notify).toHaveBeenCalledWith({
+      message: 'Profile URL not available. Please try again later.',
+    });
   });
 });
--- a/ui/components/User.test.tsx
+++ b/ui/components/User.test.tsx
@@ -238,7 +238,7 @@ describe('User component', () => {
     );
   });
 
-  it('does not redirect when no profile URL is present', async () => {
+  it('shows a warning when no profile URL is present', async () => {
     const user = userEvent.setup();
     mockGetUserQuery = {
       data: { status: 'authenticated' },
--- a/ui/components/User.tsx
+++ b/ui/components/User.tsx
@@ -37,7 +37,7 @@ const User = (props) => {
       return;
     }
     notify({
-      message: 'Profile URL not available. Please try again later.',
+      message: 'Please log in to access this profile',
       event_type: EVENT_TYPES.WARNING,
     });
   };
--- a/ui/components/User.tsx
+++ b/ui/components/User.tsx
@@ -33,9 +33,13 @@ const User = (props) => {
   const goToProfile = () => {
     const profileUrl = getProfileUrl();
     if (profileUrl) {
-      window.location = profileUrl;
+      window.open(profileUrl, '_blank', 'noopener,noreferrer');
       return;
     }
+    notify({
+      message: 'Profile URL not available. Please try again later.',
+      event_type: EVENT_TYPES.WARNING,
+    });
   };
 
   useEffect(() => {


Utkarsh-Mauryaa:
Since the implementation now uses a Link component instead of window.open, shouldn't the test case 'navigates to the profile URL when the avatar is clicked' be updated to verify that the Link renders with the correct href, target="_blank", and rel="noopener noreferrer" attributes instead of checking window.location.href?

Should i change the test case in this PR or separate PR?
@saurabhraghuvanshii

hamza-mohd:
Nice UX improvement - please ensure the new-tab behavior applies only to external links and not to in-app routes that benefit from SPA navigation.

hamza-mohd:
Nice quality-of-life. Please confirm middle-click and keyboard Enter behave consistently, and that target=_blank includes rel=noopener.

Utkarsh-Mauryaa:
@hamza-mohd 

Thanks for the review! The change was made only for the profile button, which navigates to an external URL, so the new-tab behavior is intentional in this case. I also verified that middle-click and keyboard Enter behave correctly. `rel="noopener noreferrer"` was already present alongside `target="_blank"`.

Should i change the test case in this PR or separate PR?

saurabhraghuvanshii:
@Utkarsh-Mauryaa I think yes you should also update test case in same PR.

Utkarsh-Mauryaa:
@saurabhraghuvanshii 
Okay! Working on it.

Utkarsh-Mauryaa:
@saurabhraghuvanshii 
I have modified the test case. It should be fine now. Could you please take a look?

Utkarsh-Mauryaa:
https://github.com/user-attachments/assets/89b90649-e20b-493c-9aeb-177c4b5bbee2

saurabhraghuvanshii (dismissed):

ui/components/User.tsx
```diff
@@ -82,13 +76,21 @@ const User = (props) => {
...
         <div data-testid="profile-button">
-          <IconButtonAvatar color={color} aria-haspopup="true" onClick={goToProfile}>
-            <Avatar
-              sx={{ height: 36, width: 36 }}
-              src={isGetUserSuccess ? userData?.avatarUrl : null}
-              imgProps={{ referrerPolicy: 'no-referrer' }}
-            />
-          </IconButtonAvatar>
+          <Link
+            href={profileUrl || '#'}
```
icoderarely: Should we render a link only when profileUrl exists? Using '#' still exposes a clickable/focusable element and may scroll the page to the top. It might be cleaner to conditionally render the link wrapper only when a valid profile URL is available.

ui/components/User.tsx
```diff
@@ -82,13 +76,21 @@ const User = (props) => {
...
+            target="_blank"
+            rel="noopener noreferrer"
+            aria-label="profile link"
```
icoderarely: Something more descriptive like "Open profile" or "Open user profile" be preferable for screen readers

ui/components/User.tsx
```diff
@@ -82,13 +76,21 @@ const User = (props) => {
...
+            style={{ textDecoration: 'none' }}
+          >
+            <IconButtonAvatar color={color} aria-haspopup="true">
```
icoderarely: IconButtonAvatar is built on top of IconButton from @sistent/sistent. Can we confirm what element it renders in the DOM? If it renders a <button> (which is typical for IconButton components), wrapping it with Link would create nested interactive elements (<a><button /></a>). In that case, it may be cleaner to render the IconButton itself as the link (component={Link} / component="a") instead of wrapping it.

ui/components/User.test.tsx
```diff
@@ -227,8 +226,10 @@ describe('User component', () => {
     // Trigger a re-render of the providerCapabilities effect.
     await waitFor(() => expect(ExtensionPointSchemaValidator).toHaveBeenCalledWith('account'));
 
-    await user.click(screen.getByTestId('icon-button-avatar'));
-    expect(window.location.href).toContain('https://cloud.test/profile');
+    const profileLink = screen.getByRole('link', { name: 'profile link' });
+    expect(profileLink).toHaveAttribute('href', 'https://cloud.test/profile');
+    expect(profileLink).toHaveAttribute('target', '_blank');
+    expect(profileLink).toHaveAttribute('rel', 'noopener noreferrer');
```
icoderarely: Could we also assert that the profile avatar/button is wrapped by the rendered link? The current assertions verify the link attributes, but don't verify the avatar is actually associated with that link.

ui/components/User.test.tsx
```diff
@@ -68,16 +68,42 @@ vi.mock('@sistent/sistent', () => ({
...
+          data-testid="icon-button-avatar"
+          href={href}
+          target={target}
+          rel={rel}
+          aria-label={ariaLabel}
+        >
+          {children}
+        </a>
+      );
+    }
```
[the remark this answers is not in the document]
banana-three-join: This hasn't been resolved. 

Sending a "Component" is an approach which should only be utilized when passing down items to a component which will serve as a parent component or when a component will return a modified component off the component being given. In this case, the "Component" prop isn't being used correctly. Its being used only as a conditional and it serves no other purpose within the component so that needs to be avoided.

Also, you're passing down props to a element which do not exist. Revert those changes and make sure the props being defined will serve a purpose by looking at the current usage of the IconButtonAvatar.

ui/components/User.test.tsx
```diff
@@ -68,16 +68,44 @@ vi.mock('@sistent/sistent', () => ({
...
+    return (
+      <button
+        data-testid="icon-button-avatar"
+        data-color={color}
+        data-has-popup={hasPopup}
```
banana-three-join: Custom props will not render in the DOM if they're false. aria-haspopup={hasPopup} is recommended instead.

ui/components/User.test.tsx
```diff
@@ -227,8 +254,11 @@ describe('User component', () => {
     // Trigger a re-render of the providerCapabilities effect.
     await waitFor(() => expect(ExtensionPointSchemaValidator).toHaveBeenCalledWith('account'));
 
-    await user.click(screen.getByTestId('icon-button-avatar'));
-    expect(window.location.href).toContain('https://cloud.test/profile');
+    const profileLink = screen.getByRole('link', { name: 'Open user profile' });
```
banana-three-join: This test does nothing in regards to the opening of the new link. It only makes sure that the item receives the attributes being given to it but it doesn't test the actual link opening when the link is clicked

ui/components/User.tsx
```diff
@@ -78,17 +72,35 @@ const User = (props) => {
...
+    <Avatar
+      sx={{ height: 36, width: 36 }}
+      src={isGetUserSuccess ? userData?.avatarUrl : null}
+      imgProps={{ referrerPolicy: 'no-referrer' }}
+    />
+  );
+
+  const profileLinkProps = profileUrl
+    ? {
+        component: Link as any,
```
banana-three-join: You're passing down a Link component to IconButtonAvatar. The Link component is used for pre-fetching and navigating between routes but the link which is being opened is outside of the internal routes of the Server. Also, while checking the IconButtonAvatar, did you check how it's meant to be used? Because, while we're able to mock up it's behavior, this is done only to add props which will serve to spy on it's behavior while testing the component and we shouldn't deviate from it's intended purpose.

Right now, this the current shape of the IconButtonAvatar:

```
export const IconButtonAvatar = styled(IconButton)(({ theme }) => ({
  padding: 2,
  color: theme.palette.common.white,
}));
```

banana-three-join (changes requested):

ui/components/User.test.tsx
Utkarsh-Mauryaa: aria-haspopup has been removed entirely since IconButtonAvatar is no longer the interactive element. The <a> tag handles interaction now. In the mock, aria-hidden is passed as a variable aria-hidden={ariaHidden} rather than a hardcoded boolean to ensure it renders correctly in the DOM.

ui/components/User.test.tsx
Utkarsh-Mauryaa: Updated the test to attach a click event listener and assert it is called after userEvent.click(profileLink), verifying the link is actually clickable and not just an element with attributes.

ui/components/User.tsx
Utkarsh-Mauryaa: Reviewed IconButtonAvatar, it is a styled IconButton with no polymorphic behavior. Removed Next.js Link completely since the profile URL is external and outside internal app routes. Replaced with a plain <a> tag with target="_blank" and rel="noopener noreferrer". IconButtonAvatar is now used strictly as intended.

When profileUrl exists, IconButtonAvatar receives tabIndex={-1} and aria-hidden="true" to prevent it from being a second interactive element inside the <a> tag, the <a> is the sole focusable and accessible element. When profileUrl is absent, IconButtonAvatar receives disabled instead, making it the sole element and keeping it correctly visible in the tab order and to screen readers.

ui/components/User.tsx
```diff
@@ -78,17 +72,32 @@ const User = (props) => {
...
-              src={isGetUserSuccess ? userData?.avatarUrl : null}
-              imgProps={{ referrerPolicy: 'no-referrer' }}
-            />
-          </IconButtonAvatar>
+          {profileUrl ? (
+            <a
+              href={profileUrl}
+              target="_blank"
+              rel="noopener noreferrer"
+              aria-label="Open user profile"
```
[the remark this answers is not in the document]
Utkarsh-Mauryaa: A native <a target="_blank"> click initiated by the user is never blocked by popup blockers. Browsers treat it as a trusted user gesture. However, I updated the PR description to remove any mention of the warning notification implementation.

ui/components/User.tsx
```diff
@@ -78,17 +72,32 @@ const User = (props) => {
...
+            >
+              <IconButtonAvatar tabIndex={-1} aria-hidden="true">
+                {avatar}
+              </IconButtonAvatar>
+            </a>
+          ) : (
+            <IconButtonAvatar disabled>{avatar}</IconButtonAvatar>
+          )}
```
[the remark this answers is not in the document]
Utkarsh-Mauryaa: The tabIndex={-1} + aria-hidden="true" approach is the accepted workaround for this, as it removes the inner button from tab order and hides it from assistive tech, making the outer <a> the sole interactive element.

saurabhraghuvanshii (dismissed):

ui/components/User.tsx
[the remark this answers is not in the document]
Utkarsh-Mauryaa: Added color prop to IconButtonAvatar.

ui/components/User.tsx
```diff
@@ -30,13 +30,7 @@ const User = (props) => {
...
   };
 
-  const goToProfile = () => {
-    const profileUrl = getProfileUrl();
-    if (profileUrl) {
-      window.location = profileUrl;
-      return;
-    }
-  };
+  const profileUrl = getProfileUrl();
```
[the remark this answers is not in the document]
Utkarsh-Mauryaa: Added the suggested piece of code.

ui/components/User.tsx
```diff
@@ -30,13 +30,16 @@ const User = (props) => {
     return (account || [])?.find((item) => item.title === 'Cloud Account')?.href;
   };
 
-  const goToProfile = () => {
-    const profileUrl = getProfileUrl();
-    if (profileUrl) {
-      window.location = profileUrl;
-      return;
+  const rawProfileUrl = getProfileUrl();
```
banana-three-join: Due to the previous approach of goToProfile (a redirection from the current window to [cloud.layer5.io](https://cloud.layer5.io/account/profile)), I believe it's safe to just directly call getProfileUrl() without the need to create a new URL and whole the try catch. You can just assign the value to profileUrl here.

ui/components/User.tsx
```diff
@@ -78,17 +81,34 @@ const User = (props) => {
     );
   }
 
+  const avatar = (
```
banana-three-join: You don't need to abstract the avatar

ui/components/User.tsx
```diff
@@ -78,17 +81,34 @@ const User = (props) => {
...
         <div data-testid="profile-button">
-          <IconButtonAvatar color={color} aria-haspopup="true" onClick={goToProfile}>
-            <Avatar
-              sx={{ height: 36, width: 36 }}
-              src={isGetUserSuccess ? userData?.avatarUrl : null}
-              imgProps={{ referrerPolicy: 'no-referrer' }}
-            />
-          </IconButtonAvatar>
+          {profileUrl ? (
+            <a
```
banana-three-join: You're wrapping a button with anchor tags that's going to cause unexpected issues. The current IconButtonAvatar is a styled wrapper of `export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>((props, ref)`, remove the duplicity from the new implementation in which you check the profileUrl because aside from the duplicity, users will click on the profile button but they won't understand why they're not being redirected anywhere as there's to clear indicator of why nothing is happening after the click. 

Also, disregard my previous comment about just assigning the value to profileUrl. Bring back goToProfile and instead of redirecting from within the current window,  open a new window on click while passing the `target="_blank" and rel="noopener noreferrer" attributes` on the new tab.

I believe the tests also need to be reverted and updated with the new approach

banana-three-join (changes requested):

ui/components/User.tsx
Utkarsh-Mauryaa: Okay!

ui/components/User.tsx
Utkarsh-Mauryaa: Done!

ui/components/User.tsx
Utkarsh-Mauryaa: Changes made:
- I have removed the duplicity
- Removed the anchor tags so that it doesn't cause unexpected issues.
- If no profileUrl will be present then users will be notified.(Implemented using the notify component).
- Replaced the "does not redirect when no profile URL is present" test with "shows a warning when no profile URL is present." Since we now use window.open(), the profile opens in a new tab, so window.location.href in the current tab no longer changes. The updated test is more meaningful because it verifies that users receive a warning when no profile URL is available.

banana-three-join:
@Utkarsh-Mauryaa Ty. Could you please submit a video of the working change and look into the most recent feedback from the LLM. Also, please make sure that for any future PR, you properly address ALL reviews by stating whether they're valid or not, and not just acknowledging them off as "okay" and "done".

ui/components/User.test.tsx
```diff
@@ -228,23 +231,31 @@ describe('User component', () => {
...
 
-    const startingHref = window.location.href;
     render(<UserProvider />);
 
     await user.click(screen.getByTestId('icon-button-avatar'));
-    // window.location.href should still be the same since profileUrl is undefined
-    expect(window.location.href).toBe(startingHref);
+    expect(notify).toHaveBeenCalledWith({
+      message: 'Profile URL not available. Please try again later.',
+    });
```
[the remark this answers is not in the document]
Utkarsh-Mauryaa: Should we ignore event_type, since none of the other tests validate it? @banana-three-join

Utkarsh-Mauryaa:
https://github.com/user-attachments/assets/0d535e43-fe09-4b70-85cf-5cf06613a081

When profileUrl is present.

Utkarsh-Mauryaa:
https://github.com/user-attachments/assets/a82300d5-1e32-4aee-9c23-54eab017a398

When profileUrl is absent.

banana-three-join (approved):
Ty!

PR #18178, opened 2026-03-24
[mesheryctl] add csv support on relationship generate

PragalvaXFREZ:
**Notes for Reviewers**

- This PR fixes #https://github.com/meshery/meshery/issues/18145

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #19640, opened 2026-05-23
[mesheryctl] add csv support on relationship generate

Omkar-Ugal:
This PR fixes ##18145

Revived @PragalvaXFREZ's PR to add CSV file input support to mesheryctl relationship generate, bringing it to parity with model generate.

## Motive
Previously, `relationship generate` only accepted Google Spreadsheet input. This PR adds `--file/-f` flag to accept local CSV files, making it more accessible for users who want to generate relationship docs without spreadsheet credentials.

## References
- Original PR #18178 by @PragalvaXFREZ 

Screenshot 2026-05-26 160739
Screenshot 2026-05-26 160827


mesheryctl/internal/cli/root/relationships/generate.go
```diff
@@ -48,45 +55,110 @@ type CustomValueRange struct {
...
 	},
 }
 
 func init() {
-	generateCmd.PersistentFlags().StringVar(&relationshipGenerateFlag.SpreadsheetID, "spreadsheet-id", "", "spreadsheet ID for the integration spreadsheet")
-	generateCmd.PersistentFlags().StringVar(&relationshipGenerateFlag.SpreadsheetCred, "spreadsheet-cred", "", "base64 encoded credential to download the spreadsheet")
+	generateCmd.Flags().StringVarP(&relationshipGenerateFlag.File, "file", "f", "", "path to the relationships CSV file")
+	generateCmd.Flags().StringVar(&relationshipGenerateFlag.SpreadsheetID, "spreadsheet-id", "", "spreadsheet ID for the integration spreadsheet")
+	generateCmd.Flags().StringVar(&relationshipGenerateFlag.SpreadsheetCred, "spreadsheet-cred", "", "base64 encoded credential to download the spreadsheet")
+	generateCmd.Flags().StringVarP(&relationshipGenerateFlag.Output, "output", "o", "", "path to the output JSON file (default: docs/data/RelationshipsData.json)")
```
lekaf974: Currently default is `""`

mesheryctl/internal/cli/root/relationships/generate.go
```diff
@@ -48,45 +55,110 @@ type CustomValueRange struct {
...
+	reader := csv.NewReader(f)
+	reader.FieldsPerRecord = -1
+	records, err := reader.ReadAll()
+	if err != nil {
+		return utils.ErrFileRead(err)
+	}
 
+	// First two rows are headers, data starts at row 3
+	if len(records) <= 2 {
+		return ErrEmptyCSVData(fmt.Errorf("no relationship data found in CSV file: %s", filePath))
```
lekaf974: The error does not give any real guidance about the csv file content needed

mesheryctl/internal/cli/root/relationships/generate_test.go
```diff
@@ -2,25 +2,30 @@ package relationships
...
 	"google.golang.org/api/sheets/v4"
 )
 
-func expectedViewFlagError(spreadsheetId string, spreadsheetCred string) error {
-	if spreadsheetCred == "" {
-		return utils.ErrFlagsInvalid(fmt.Errorf("Invalid value for --spreadsheet-cred ''"))
-	}
-
-	return utils.ErrFlagsInvalid(fmt.Errorf("Invalid value for --spreadsheet-id ''"))
+func resetGenerateFlags() {
```
lekaf974: I am pretty sure there is a reusable function doing this is utils package

mesheryctl/internal/cli/root/relationships/generate_test.go
```diff
@@ -29,61 +34,59 @@ func TestGenerateErrorOutput(t *testing.T) {
...
-		ExpectedError    error
+		Name           string
+		Args           []string
+		ExpectError    bool
+		ErrorSubstring string
 	}{
-
 		{
-			Name:           "Generate registered relationships without spreadsheet credentials",
+			Name:           "Generate without any input source",
```
lekaf974: use given when then for the test name

mesheryctl/internal/cli/root/relationships/generate_test.go
```diff
@@ -29,61 +34,59 @@ func TestGenerateErrorOutput(t *testing.T) {
...
 			mesheryctlflags.InitValidators(RelationshipCmd)
 			RelationshipCmd.SetArgs(tt.Args)
 			err := RelationshipCmd.Execute()
 
-			// to validate the expected errors
 			if err != nil {
 				if tt.ExpectError {
-
-					utils.AssertMeshkitErrorsEqual(t, err, tt.ExpectedError)
+					assert.True(t, strings.Contains(err.Error(), tt.ErrorSubstring),
```
lekaf974: please keep the AssertMeshkitErrorsEquals function to assert error

lekaf974 (changes requested):
Please review comments

mesheryctl/internal/cli/root/relationships/generate.go
```diff
@@ -48,45 +55,110 @@ type CustomValueRange struct {
...
+	generateCmd.Flags().StringVar(&relationshipGenerateFlag.SpreadsheetCred, "spreadsheet-cred", "", "base64 encoded credential to download the spreadsheet")
+	generateCmd.Flags().StringVarP(&relationshipGenerateFlag.Output, "output", "o", "", "path to the output JSON file")
+
+	generateCmd.MarkFlagsOneRequired("spreadsheet-id", "file")
+	generateCmd.MarkFlagsMutuallyExclusive("spreadsheet-id", "file")
+	generateCmd.MarkFlagsRequiredTogether("spreadsheet-id", "spreadsheet-cred")
 }
 
-func createJsonFile(resp *sheets.ValueRange, jsonFilePath string) error {
+func generateFromCSV(filePath, outputPath string) error {
```
lekaf974: Naming should be more explicit generate what from CSV ?

mesheryctl/internal/cli/root/relationships/generate.go
```diff
@@ -48,45 +55,110 @@ type CustomValueRange struct {
...
+				VisualizationExample: row[14],
+			})
+		}
+	}
+
+	if len(customResp) == 0 {
+		return ErrEmptyCSVData(fmt.Errorf("no valid relationship rows found in CSV file: %s", filePath))
+	}
+
+	return saveRelationshipsJSON(customResp, outputPath)
```
lekaf974: Hum this save method within generate method is not a good practice, please return the data than save it where it is called to make easier to follow the different steps

mesheryctl/internal/cli/root/relationships/generate.go
```diff
@@ -1,20 +1,27 @@
...
 type cmdRelationshipGenerateFlag struct {
-	SpreadsheetID   string `json:"spreadsheet-id" validate:"required"`
-	SpreadsheetCred string `json:"spreadsheet-cred" validate:"required"`
+	SpreadsheetID   string `json:"spreadsheet-id" validate:"omitempty"`
+	SpreadsheetCred string `json:"spreadsheet-cred" validate:"omitempty"`
+	File            string `json:"file" validate:"omitempty,filepath"`
+	Output          string `json:"output" validate:"omitempty"`
 }
 
+const minRelationshipCSVColumns = 15
```
lekaf974, line 27: I would have add a comment to explain it and perhaps a link to a valid template file

mesheryctl/internal/cli/root/relationships/error.go
```diff
@@ -22,3 +23,14 @@ func ErrEmptySheetData(err error) error {
...
 }
+
+func ErrEmptyCSVData(err error) error {
+	return errors.New(
+		ErrEmptyCSVDataCode,
+		errors.Alert,
+		[]string{"Invalid or empty CSV file provided"},
+		[]string{err.Error()},
+		[]string{"The CSV file must contain valid relationship entries with the required headers."},
+		[]string{"Ensure the CSV file contains two header rows and at least one data row with a minimum of 15 columns. Refer to the Meshery relationship CSV template: \"https://github.com/meshery/meshery/blob/master/mesheryctl/templates/template-csvs/Relationships.csv\""},
```
PragalvaXFREZ, line 34: This is another instance of a broader error formatting pattern that I think we should revisit.

The intent here is good, giving the user more context, but across commands we often end up stacking multiple error strings that say roughly the same thing in different wording. That makes the output feel noisy instead of helpful.

// @lekaf974

PragalvaXFREZ (commented):
Thanks for reviving this PR, @Omkar-Ugal.

When I initially worked on this issue, the motivation was mostly from a UX perspective: why can’t we generate relationships from a CSV?

It has been quite some time since I last ran this command, but from what I remember, it only generates relationship docs. When I presented this at the dev meeting, Aabid suggested that the development of this subcommand might be incomplete.

@lekaf974 , do you have any context on this?

lekaf974 (dismissed):
LGTM

lekaf974:
> Thanks for reviving this PR, @Omkar-Ugal.
> 
> When I initially worked on this issue, the motivation was mostly from a UX perspective: why can’t we generate relationships from a CSV?
> 
> It has been quite some time since I last ran this command, but from what I remember, it only generates relationship docs. When I presented this at the dev meeting, Aabid suggested that the development of this subcommand might be incomplete.
> 
> @lekaf974 , do you have any context on this?

I don't remember but @aabidsofi19 could give some hints perhaps

mesheryctl/internal/cli/root/relationships/generate.go
```diff
@@ -25,7 +36,7 @@ var fetchSheetValues = func(id, cred string) (*sheets.ValueRange, error) {
 	return srv.Spreadsheets.Values.Get(id, "Relationships").Do()
 }
 
-var relationshipsOutputPath = "../docs/_data/RelationshipsData.json"
+var relationshipsOutputPath = "docs/data/RelationshipsData.json"
```
PragalvaXFREZ:
```suggestion
var relationshipsOutputPath = "../docs/data/RelationshipsData.json"
```

recommending this as mostly we run the command from mesheryctl directory.

mesheryctl/internal/cli/root/relationships/generate.go
```diff
@@ -48,45 +59,114 @@ type CustomValueRange struct {
...
+				MetadataDescription:  row[5],
+				Docs:                 row[6],
+				MetadataStyles:       row[7],
+				EvalPolicy:           row[8],
+				SelectorsDenyFrom:    row[9],
+				SelectorsDenyTo:      row[10],
+				SelectorsAllowFrom:   row[11],
+				SelectorsAllowTo:     row[12],
+				CompleteDefinition:   row[13],
+				VisualizationExample: row[14],
```
PragalvaXFREZ, line 153: this is fragile but i think i have seen this pattern else where as well. 

I think this hardcoding has already led to drift with our integration spreadsheet. https://docs.google.com/spreadsheets/d/1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw/edit?gid=1171594589#gid=1171594589

I would suggest we move out from hardcoding the index. 

@Omkar-Ugal , would you please create an issue and work on fixing it? , verify if this drift is present elsewhere as well. 

@YASHMAHAKAL , could you help verify if this command is accurately creating the relationships?

mesheryctl/internal/cli/root/relationships/generate.go
PragalvaXFREZ: thinking about it again, even my recommendation is fragile.  Accounting for the previous behavior i think it's still acceptable, but an enhancement can be made in this aspect as well.

YASHMAHAKAL (commented):
@Omkar-Ugal please update your PR description with information related to your efforts, motive of your PR and other necessary references

mesheryctl/internal/cli/root/relationships/generate.go
YASHMAHAKAL: This command is generating relationship docs but not relationships itself

Omkar-Ugal: @PragalvaXFREZ Checked the codebase hardcoded row index pattern is only present in relationships/generate.go, not found elsewhere in mesheryctl commands.

PragalvaXFREZ (approved):
LGTM, thanks for creating the issue as well.

As a starting point, could you trace how `mesheryctl model generate` avoids positional column access and evaluate whether relationship generate can use a similar approach? The main thing to remove is the brittle row[n] mapping in relationship generation.

Omkar-Ugal:
@PragalvaXFREZ Traced it meshkit already has RelationshipCSVHelper.ParseRelationshipsSheet which handles header-based mapping. Will use that in #19785 .

PragalvaXFREZ:
sounds good, merging this PR. 

> @PragalvaXFREZ Traced it meshkit already has RelationshipCSVHelper.ParseRelationshipsSheet which handles header-based mapping. Will use that in #19785 .

PR #19643, opened 2026-05-23
[server] init subscriber slices with len=0 in helper constructors

### LINKED ISSUE #19581: [server] NewConfigurationHelper and NewSummaryHelper pre-fill subscriber slices with nil channels
### Current Behavior

`NewConfigurationHelper` and `NewSummaryHelper` pre-fill their subscriber slices with nil channels:

- `server/models/configuration_helper.go:16-22` constructs `ApplicationsChannel`, `PatternsChannel`, `FiltersChannel` via `make([]chan struct{}, 10)`. The slice has length 10 of zero-value (nil) channels before any subscriber appends.
- `server/models/meshmodel/helper.go:14-18` constructs `SummaryChannel.channel` the same way: `make([]chan struct{}, 10)`.

The corresponding `Subscribe*` methods then `append` real subscribers onto the slice, so iteration in `Publish*` walks the 10 nil entries first and the real subscribers after.

### Why this is a bug

Sending to a nil channel blocks forever. Before #19580 (now open), the publishers wrapped each send in `if !utils.IsClosed(ch)` which returned `false` for nil channels (the helper checks `if ch == nil { return true }`, but the publishers used `if !IsClosed(ch)`, so the guard short-circuited and the send was skipped — except `IsClosed` also drained buffered channels, so the gate was broken in a different way). Net effect on master today: events get lost or the publisher hangs depending on timing.

After #19580, the publishers use a non-blocking `select { case ch <- struct{}{}: default: }`, which is correct for the real subscribers but silently drops events on the 10 leading nil slots. That is still a behaviour change worth fixing at the root.

### Expected Behavior

Construct subscriber slices with length 0 and let `Subscribe*` grow them via `append`. `NewDashboardK8sResourcesHelper` (`server/models/dashboard_k8s_resources.go:14-18`) and `NewContextHelper` (`server/models/k8scontext_helper.go:14-18`) already do this correctly with `make([]chan struct{}, 0)`.

### Proposed fix

```go
// server/models/configuration_helper.go
func NewConfigurationHelper() *ConfigurationChannel {
    return &ConfigurationChannel{
        ApplicationsChannel: make([]chan struct{}, 0),
        PatternsChannel:     make([]chan struct{}, 0),
        FiltersChannel:      make([]chan struct{}, 0),
    }
}

// server/models/meshmodel/helper.go
func NewSummaryHelper() *SummaryChannel {
    return &SummaryChannel{
        channel: make([]chan struct{}, 0),
    }
}
```

3 LOC change. No semantic ripple because no caller relies on the slice length being 10.

### Context

Spotted while auditing PR #19580 (`utils.IsClosed` follow-up to #17011 / #19572). Out of scope for #19580 because that PR is intentionally scoped to the `Publish*` gate replacement.

SarthakB11:
Fixes #19581.

`NewConfigurationHelper` (`server/models/configuration_helper.go:14-20`) and `NewSummaryHelper` (`server/models/meshmodel/helper.go:12-16`) initialise their subscriber slices with `make([]chan struct{}, 10)`. The slice has length 10 of zero-value (nil) channels before any subscriber appends. `Subscribe*` then appends real subscribers onto the slice, so iteration in `Publish*` walks 10 nil entries first and the real subscribers after.

Sending to a nil channel blocks forever. The pre-existing `utils.IsClosed` guard short-circuited on nil channels (explicit `if ch == nil { return true }`), but the same helper also drains buffered channels of any queued value (`case <-ch: return true`), which is bug #17011's root cause.

### Change

Switch to `make([]chan struct{}, 0)` so `Subscribe*` grows the slice with real channels only.

```go
// server/models/configuration_helper.go
func NewConfigurationHelper() *ConfigurationChannel {
    return &ConfigurationChannel{
        ApplicationsChannel: make([]chan struct{}, 0),
        PatternsChannel:     make([]chan struct{}, 0),
        FiltersChannel:      make([]chan struct{}, 0),
    }
}

// server/models/meshmodel/helper.go
func NewSummaryHelper() *SummaryChannel {
    return &SummaryChannel{
        channel: make([]chan struct{}, 0),
    }
}
```

Matches the existing pattern in `NewDashboardK8sResourcesHelper` (`server/models/dashboard_k8s_resources.go:14`) and `NewContextHelper` (`server/models/k8scontext_helper.go:14`), both of which already use length-zero slices.

### Why a separate PR

The `IsClosed` cleanup chain (#19572 → #19580 → #19585) replaces the broken gate. After that chain lands, the publishers no longer block on nil slots (they hit the `default` branch of the non-blocking select), but they still silently iterate ten no-op slots per publish. Fixing the root cause at construction is independent and correct on master too.

### DIFF
--- a/server/models/configuration_helper.go
+++ b/server/models/configuration_helper.go
@@ -15,9 +15,9 @@ type ConfigurationChannel struct {
 
 func NewConfigurationHelper() *ConfigurationChannel {
 	return &ConfigurationChannel{
-		ApplicationsChannel: make([]chan struct{}, 10),
-		PatternsChannel:     make([]chan struct{}, 10),
-		FiltersChannel:      make([]chan struct{}, 10),
+		ApplicationsChannel: make([]chan struct{}, 0),
+		PatternsChannel:     make([]chan struct{}, 0),
+		FiltersChannel:      make([]chan struct{}, 0),
 	}
 }
 
--- a/server/models/meshmodel/helper.go
+++ b/server/models/meshmodel/helper.go
@@ -13,7 +13,7 @@ type SummaryChannel struct {
 
 func NewSummaryHelper() *SummaryChannel {
 	return &SummaryChannel{
-		channel: make([]chan struct{}, 10),
+		channel: make([]chan struct{}, 0),
 	}
 }
 


SarthakB11:
Master already inits these slices with make([]chan struct{}, 0) via a03fd9a0f070 (#20375). Closing as superseded.

PR #19651, opened 2026-05-24
[Perf] Add Docker BuildKit cache mounts for Go module and build cache in Dockerfile

### LINKED ISSUE #19703: [Docker] Use BuildKit cache mounts for Go module and build cache in Dockerfile
### Current Behavior
The Go build stage in install/docker/Dockerfile runs go clean -modcache (defeats module caching) and compiles with the -a flag (forces full rebuilds). Each docker build redownloads all dependencies and recompiles every package from scratch, adding 2-3 minutes to every CI build.

### Desired Behavior
Docker builds should cache Go modules and compiled objects across builds using BuildKit cache mounts, so that only changed packages are recompiled when source files are modified.

### Implementation
Replace go clean -modcache && go build -a with plain go build. Add --mount=type=cache for /go/pkg/mod and /root/.cache/go-build. Copy go.mod and go.sum before go mod download to allow the module cache to be populated before the source is copied.

### Acceptance Tests
- docker build completes successfully
- Second build with unchanged sources completes in under 30 seconds (cache hit)

shivv23:
**Notes for Reviewers**

- This PR fixes #19703

## Description

The Go build stage in install/docker/Dockerfile currently runs go clean -modcache (defeats module caching) and compiles with -a (forces full rebuilds). This makes repeated docker build runs unnecessarily slow.

### Changes

- Replaced go clean -modcache && go build -a with plain go build
- Added COPY go.mod go.sum ./ before go mod download so the module cache actually works
- Added --mount=type=cache,target=/go/pkg/mod to persist the Go module cache across builds
- Added --mount=type=cache,target=/root/.cache/go-build to persist the Go build cache across builds

### Benefits

- Subsequent docker build runs reuse cached modules and compiled objects instead of redownloading everything
- CI builds benefit even more when agents share a common cache volume

### Testing

- docker build completes successfully
- First build: full compilation (cache miss)
- Subsequent builds: near-instant when Go sources are unchanged

### DIFF
--- a/install/docker/Dockerfile
+++ b/install/docker/Dockerfile
@@ -6,8 +6,20 @@ ARG RELEASE_CHANNEL
 
 RUN adduser --disabled-login appuser
 WORKDIR /github.com/meshery/meshery
+
+# Cache Go module downloads across builds so a new dependency only
+# fetches what's missing instead of redownloading everything.
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    cd server/cmd && \
+    GOPROXY=https://proxy.golang.org GOSUMDB=off go mod download
 ADD . .
-RUN go clean -modcache; cd server; cd cmd; GOPROXY=https://proxy.golang.org GOSUMDB=off go build -ldflags="-w -s -X main.globalTokenForAnonymousResults=$TOKEN -X main.version=$GIT_VERSION -X main.commitsha=$GIT_COMMITSHA -X main.releasechannel=$RELEASE_CHANNEL" -tags draft -a -o /meshery .
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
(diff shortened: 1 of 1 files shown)


shivv23:
All Gemini code review suggestions addressed:

1. **\go mod download\ before \go.mod\/**go.sum copied — This would have made the build fail. Fixed by adding \COPY go.mod go.sum ./\ before the \go mod download\ step, preserving the layer-caching benefit. ✅
2. **Redundant build cache mount** — Removed \--mount=type=cache,target=/root/.cache/go-build\ from the download step; only the module cache (\/go/pkg/mod\) is needed for \go mod download\. ✅

shivv23:
@leecalcote straightforward perf improvement — adds BuildKit cache mounts for Go modules to cut down CI build times. Could use a review.

shivv23:
Hi @leecalcote, this PR is ready for review. It adds Docker BuildKit cache mounts for Go module and build cache, which should significantly speed up Docker builds in CI. Could you please take a look when you get a chance?

PR #19655, opened 2026-05-24
[Server] Return JSON success bodies from event mutation handlers

### LINKED ISSUE #19707: [Server] Event mutation handlers return empty 200 instead of JSON success body
### Current Behavior
Four event mutation handlers (UpdateEventStatus, BulkUpdateEventStatus, BulkDeleteEvent, DeleteEvent) return a bare HTTP 200 with no body on success. The frontend RTK Query expects a JSON body on success, causing response-dispatching issues and making it impossible for the client to distinguish a successful operation from a missing handler.

### Expected Behavior
Each handler should write a JSON success body ({"message": "..."}) with Content-Type: application/json on successful completion, following the same pattern used by other handlers in events_streamer.go.

### Environment
- Host OS: Linux

shivv23:
**Notes for Reviewers**

- This PR fixes #19707

## Description

Four event mutation handlers — UpdateEventStatus, BulkUpdateEventStatus, BulkDeleteEvent, and DeleteEvent — return a bare HTTP 200 with no body (or write nothing to the response writer). The frontend RTK Query expects a JSON body on success, causing response-dispatching issues.

### Response shapes (aligned with v1beta3 event schema)

| Handler | Status | Response body |
|---------|--------|--------------|
| UpdateEventStatus | 200 | `{"message", "eventId", "status"}` |
| BulkUpdateEventStatus | 200 | `{"message", "updated": [UUID]}` |
| BulkDeleteEvent | 200 | `{"message", "deleted": [UUID]}` |
| DeleteEvent | **204** | No body |

The response schemas are defined in `meshery/schemas` at `schemas/constructs/v1beta3/event/api.yml`.

### Changes

All four handlers now write a JSON response body on success via the existing writeJSONMessage helper. DeleteEvent returns 204 No Content per the OpenAPI contract.

### Testing

- go build ./server/handlers/... — passes
- go vet ./server/handlers/ — no issues

### DIFF
--- a/server/handlers/events_streamer.go
+++ b/server/handlers/events_streamer.go
@@ -165,6 +165,7 @@ func (h *Handler) UpdateEventStatus(w http.ResponseWriter, req *http.Request, pr
 		return
 	}
 
+	writeJSONMessage(w, map[string]string{"message": "Event status updated"}, http.StatusOK)
 }
 
 func (h *Handler) BulkUpdateEventStatus(w http.ResponseWriter, req *http.Request, prefObj *models.Preference, user *models.User, provider models.Provider) {
@@ -191,6 +192,7 @@ func (h *Handler) BulkUpdateEventStatus(w http.ResponseWriter, req *http.Request
 		return
 	}
 
+	writeJSONMessage(w, map[string]string{"message": "Event statuses updated"}, http.StatusOK)
 }
 
 func (h *Handler) BulkDeleteEvent(w http.ResponseWriter, req *http.Request, prefObj *models.Preference, user *models.User, provider models.Provider) {
@@ -215,6 +217,8 @@ func (h *Handler) BulkDeleteEvent(w http.ResponseWriter, req *http.Request, pref
 		writeMeshkitError(w, _err, http.StatusInternalServerError)
 		return
 	}
+
+	writeJSONMessage(w, map[string]string{"message": "Events deleted"}, http.StatusOK)
 }
 
 func (h *Handler) DeleteEvent(w http.ResponseWriter, req *http.Request, prefObj *models.Preference, user *models.User, provider models.Provider) {
@@ -228,6 +232,8 @@ func (h *Handler) DeleteEvent(w http.ResponseWriter, req *http.Request, prefObj 
 		writeMeshkitError(w, _err, http.StatusInternalServerError)
 		return
 	}
+
+	writeJSONMessage(w, map[string]string{"message": "Event deleted"}, http.StatusOK)
 }
 
 func getEventFilter(req *http.Request) (*events.EventsFilter, error) {
--- a/server/handlers/events_streamer.go
+++ b/server/handlers/events_streamer.go
@@ -165,7 +165,11 @@ func (h *Handler) UpdateEventStatus(w http.ResponseWriter, req *http.Request, pr
 		return
 	}
 
-	writeJSONMessage(w, map[string]string{"message": "Event status updated"}, http.StatusOK)
+	writeJSONMessage(w, map[string]interface{}{
+		"message": "Event status updated",
+		"eventId": eventID,
+		"status":  status,
+	}, http.StatusOK)
 }
 
 func (h *Handler) BulkUpdateEventStatus(w http.ResponseWriter, req *http.Request, prefObj *models.Preference, user *models.User, provider models.Provider) {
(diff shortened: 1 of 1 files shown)


suzy-g38:
@shivv23 can you mention the issue number? It will be easier

shivv23:
@suzy-g38 yes, I have mentioned the issue no.

PragalvaXFREZ (changes requested):
Thanks, can you verify its status in [meshery/schemas](https://github.com/meshery/schemas)?

If this endpoint is schema backed, the contract change should be introduced there first and then propagated to the consumer here.

shivv23:
Good catch, totally right that these endpoints are schema backed.

I checked `schemas/constructs/v1beta3/event/api.yml` and the response shapes differ from what I initially had:

| Schema operation | Defined response |
|---|---|
| `updateEventStatus` | UpdateEventStatusResponse — `{message, eventId, status}` |
| `bulkUpdateEventStatus` | BulkUpdateStatusResponse — `{updated: [UUID]}` |
| `bulkDeleteEvents` | BulkDeleteResponse — `{deleted: [UUID]}` |
| `deleteEvent` | `204 No Content` |

Pushed a fix (06c27033) that aligns all four handlers to match:

- **UpdateEventStatus** now returns the eventId and status alongside the message.
- **BulkUpdateEventStatus / BulkDeleteEvent** echo back the IDs that were affected (we already have them from the request body since the provider methods return `error` only, not the processed IDs).
- **DeleteEvent** now returns `204 No Content` per the schema instead of 200 with a body.

Updated the PR description to call out the schema mapping too. Thanks for calling this out — wouldn't have caught it otherwise.

shivv23:
@chaitanyamedidar would appreciate a re-review when you get a chance — the schema alignment feedback from last time is addressed.

chaitanyamedidar (commented):
@shivv23 One concern is that the bulk responses currently use IDs from the request body as `updated` / `deleted`. As noted above, the provider methods return only `error`, not the processed IDs, so these fields are reporting requested IDs rather than confirmed affected IDs. That can over-report success if some IDs do not exist or are not changed. We should either return actual affected IDs from the provider layer, or adjust the response contract/name if the intended meaning is requested IDs.

shivv23:
Thanks for the review, @chaitanyamedidar. You are right that the bulk endpoints currently echo back the IDs from the request body rather than confirmed affected IDs from the provider layer. The provider methods (e.g. BulkUpdateEventStatus, BulkDeleteEvent) currently return only error, not the processed IDs, so there is no way to distinguish requested vs actually affected IDs at the handler level without changing the provider interface.

Two paths forward:
1. Update the provider layer to return the set of IDs that were actually modified/deleted, then propagate those to the response.
2. Keep echoing the request IDs but note in the response contract that these reflect the requested set, not the confirmed set.

Which approach do you recommend, or would you accept option 2 as incremental improvement with a follow-up to fix the provider layer?

shivv23:
Hi @leecalcote, this PR is ready for review. It adds JSON success response bodies to event mutation handlers in the server, ensuring consistent API responses. Could you please take a look when you get a chance?

shivv23:
Thanks for flagging this. The response shapes in this PR are already aligned with the v1beta3 event schema defined in meshery/schemas at `schemas/constructs/v1beta3/event/api.yml`. Commit 06c27033 was specifically made to ensure the response body contract matches the canonical schema definitions. The four handler responses (UpdateEventStatus, BulkUpdateEventStatus, BulkDeleteEvent, DeleteEvent) follow the schema exactly - no further alignment is needed.

shivv23:
That is a valid concern. You are right that the provider methods (`UpdateEventStatus`, `BulkUpdateEventStatus`, `BulkDeleteEvent`) return only `error`, not the processed IDs, so the `updated` and `deleted` fields in the response echo the requested IDs rather than confirmed-affected IDs. This is a limitation of the current provider-layer interface. Two approaches: (1) return actual affected IDs from the provider layer (more accurate but requires a bigger refactor), or (2) rename the response fields to `requested` to signal the semantics clearly. I would lean toward approach 1 if you agree - I can add the affected-IDs return to the provider interface in a follow-up. Let me know your preference.

shivv23:
Hi @PragalvaXFREZ and @chaitanyamedidar — commit 06c27033 aligns the response shapes with the v1beta3 event schema in meshery/schemas (\schemas/constructs/v1beta3/event/api.yml\). The four handler responses now follow the canonical schema exactly. All CI is passing (golangci-lint, unit tests, integration tests, E2E). Could you please re-review when you get a chance?

PR #19657, opened 2026-05-24
fix(ui): clear setTimeout on unmount in AnimatedMeshPattern

### LINKED ISSUE #19647: [UI] setTimeout calls not cleaned up in AnimatedMeshPattern causing setState on unmounted component
##  Bug
In `ui/components/shared/LoadingState/Animations/AnimatedMeshPattern.tsx`, 
two `setTimeout` calls inside `useEffect` hooks have no cleanup function. 
When the component unmounts (route change or modal close) before the timers 
fire, `setIsActive` is called on an unmounted component.

## Current Behavior
- Timers continue running after component unmounts
- `setIsActive` called on unmounted component
- React warning in browser console
- Potential memory leak on repeated navigation

## Expected Behavior
Timers should be cancelled when component unmounts using `clearTimeout` 
in the `useEffect` return function.

## Steps To Reproduce
1. Open Meshery UI
2. Navigate to any page that triggers the loading animation
3. Quickly navigate away before animation completes
4. Open browser console and observe React warnings

## Root Cause
```tsx
// No cleanup — timer fires even after unmount
useEffect(() => {
  setTimeout(() => setIsActive(false), 100);
}, []);

useEffect(() => {
  setTimeout(() => setIsActive(!isActive), 2000);
}, [isActive]);
```

## Suggested Fix
```tsx
useEffect(() => {
  const id = window.setTimeout(() => setIsActive(false), 100);
  return () => window.clearTimeout(id);
}, []);

useEffect(() => {
  const id = window.setTimeout(() => {
    setIsActive((prev) => !prev);
  }, 2000);
  return () => window.clearTimeout(id);
}, [isActive]);
```

## Environment
- Area: UI / Loading Animation
- File: `ui/components/shared/LoadingState/Animations/AnimatedMeshPattern.tsx`
- Framework: React

Faisal77666:
## Description

Fixes #19647

## Problem
Two `setTimeout` calls inside `useEffect` hooks in 
`AnimatedMeshPattern.tsx` had no cleanup function. When the component 
unmounts before the timers fire, `setIsActive` is called on an unmounted 
component causing React warnings and memory leaks.

## Root Cause
`useEffect` hooks scheduled `setTimeout` but returned no cleanup so 
`clearTimeout` was never called on unmount.

## Solution
- Stored `setTimeout` return value in variable `id`
- Returned `clearTimeout(id)` as cleanup from both effects
- Used functional `setState` (`prev => !prev`) in second effect 
  to avoid stale closure

## Changes
- `ui/components/shared/LoadingState/Animations/AnimatedMeshPattern.tsx`

### DIFF
--- a/ui/components/shared/LoadingState/Animations/AnimatedMeshPattern.tsx
+++ b/ui/components/shared/LoadingState/Animations/AnimatedMeshPattern.tsx
@@ -15,15 +15,15 @@ const AnimatedMeshPattern = (props: React.SVGAttributes<SVGSVGElement>) => {
   const [isActive, setIsActive] = useState(true);
   const theme = useTheme();
   useEffect(() => {
-    setTimeout(() => {
-      setIsActive(false);
-    }, 100);
+    const id = window.setTimeout(() => setIsActive(false), 100);
+    return () => window.clearTimeout(id);
   }, []);
 
   useEffect(() => {
-    setTimeout(() => {
-      setIsActive(!isActive);
+    const id = window.setTimeout(() => {
+      setIsActive((prev) => !prev);
     }, 2000);
+    return () => window.clearTimeout(id);
   }, [isActive]);
 
   return (


Faisal77666:
Hi @leecalcote  I found this bug while auditing the Meshery UI codebase. Two setTimeout 
calls in AnimatedMeshPattern.tsx had no cleanup function, causing 
setIsActive to be called on an unmounted component when users navigate 
away quickly.Fix is minimal and focused — only the two useEffect blocks are changed, 
nothing else is touched.Regarding the CI failure — it appears to be a kind-registry 
infrastructure issue unrelated to this change, as it is affecting 
multiple PRs. Happy to re-run if needed.

Amr-Shams (commented):
LGTM, while i don't prefer to use the global window but it is fine here

hortison:
> LGTM, while i don't prefer to use the global window but it is fine here

Thanks for this review, @Amr-Shams 👍

hortison (approved):

Issue #19646, opened 2026-05-24
[mesheryctl] fix: goroutine leaks in SSE client, adapter wait routines, and dashboard keep-alive


ishwar170695:
_See [mesheryctl Command Tracker](https://bit.ly/3dqXy1q) for current status of commands._

#### ./internal/cli/root/adapter/adapter.go
```
	timer := time.NewTimer(time.Duration(1200) * time.Second)
	eventChan := make(chan string)

	// Run a goroutine to wait for the response
	go func() {
		for i := range event {
			if strings.Contains(i.Data.Details, query) {
				eventChan <- "successful"
				utils.Log.Infof("%s\n%s\n", i.Data.Summary, i.Data.Details)
			} else if strings.Contains(i.Data.Details, "Error") {
				eventChan <- "error"
				utils.Log.Infof("%s\n", i.Data.Summary)
			}
		}
	}()

	select {
	case <-timer.C:
		return "", ErrTimeoutWaitingForDeployResponse
	case event := <-eventChan:
		if event != "successful" {
			return "", ErrFailedDeployingMesh
		}
	}
```
A background goroutine reads event streams and sends to `eventChan`. The main routine select-blocks on a 1200-second timer or `eventChan`.
If a timeout happens, or the first event is an error, the main function exits. The background goroutine attempts to write to `eventChan `on the next iteration and blocks indefinitely because the receiver is gone, permanently leaking the goroutine.

#### ./pkg/utils/sse-client.go
```
func ConvertRespToSSE(resp *http.Response) (chan Event, error) {
	events := make(chan Event)
	reader := bufio.NewReader(resp.Body)

	go loop(reader, events)

	return events, nil
}
```
 If the caller stops reading from `events`, `loop` blocks indefinitely on `events <- ev` and never exits.

#### ./internal/cli/root/system/dashboard.go
```
ticker := time.NewTicker(10 * time.Second)
				go func() {
					for {
						select {
						case <-signals:
							portforward.Stop()
							ticker.Stop()
							return
						case <-ticker.C:
							keepConnectionAlive(mesheryURL)
						}
					}
				}()
				utils.Log.Infof("Forwarding port %v -> %v", options.podPort, dashboardCmdFlags.Port)
				utils.Log.Infof("Meshery UI available at: %s", mesheryURL)
				utils.Log.Info("Opening Meshery UI in default browser...")
				err = utils.NavigateToBrowser(mesheryURL)
				if err != nil {
[...]

PR #19659, opened 2026-05-24
fix(adapter.go) Prevent adapter event goroutine leak

ishwar170695:
**Notes for Reviewers**
Extracted shared SSE-wait logic into `waitForSSEEvent `helper per reviewer feedback, both `waitForDeployResponse` and `waitForValidateResponse `now delegate to it.
#### Signature change: 
`ConvertRespToSSE` now accepts a `context.Context` as first argument. Both callers (`waitForDeployResponse` and `waitForValidateResponse` in `adapter.go`) have been updated.

- This PR fixes #
This change fixes a part of #19646 and prevents the adapter event reader goroutine from leaking when `waitForDeployResponse` or `waitForValidateResponse` exits early due to timeout or a failed event. The reader now gets a cancellation signal and stops cleanly after the first terminal result. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment and validation status reporting by recognizing success and error events consistently, regardless of capitalization.
  * Operations now stop promptly when completed, cancelled, interrupted, or when the event stream closes.
  * Improved handling of malformed or unexpected event-stream messages to prevent stalled operations.
  * Event-stream processing now closes cleanly after errors, cancellation, or unexpected input.
  * Validation and deployment responses provide more reliable failure status reporting.


ishwar170695:
Thanks for the review. The nested select + done channel was intentional since buffered channel alone doesn't stop the reader goroutine if it's blocked on `resp.Body` reads. The safest approach combines both: `eventChan := make(chan string, 1)` for non-blocking sends, plus cancellation via context to actually stop the reader loop.
Will update with: buffered channel, logs before sends, standardized checks (`i.Data.Details` for both, case-insensitive error detection), and `ConvertRespToSSE(ctx, ...)` so the SSE reader is fully cancellable. Will update all call sites.

mesheryctl/internal/cli/root/adapter/adapter.go
lekaf974, line 1: This looks good but wondering if we could simply this, basically the same thing is duplicated, any thought to apply DRY principle ?

lekaf974 (commented):
Small comment added and nice job

mesheryctl/internal/cli/root/adapter/adapter.go
ishwar170695: @lekaf974 Thanks for the feedback. Are you referring to the duplication between `waitForDeployResponse` and `waitForValidateResponse`?

If so, thinking of extracting the shared SSE-wait logic into a helper like `waitForSSEEvent(ctx, eventCh, query, timeoutErr)` . The only real differences are the URL path, Accept header, and which timeout error gets returned, so those stay in the callers. Does that align with what you had in mind?

ishwar170695: Done. Extracted the shared SSE-wait logic into `waitForSSEEvent(ctx, event, query, successLogFormat, errorLogFormat)`. Both `waitForDeployResponse` and `waitForValidateResponse` now use it, the only differences remaining in each caller are the URL path and which timeout/failure error gets returned.

ishwar170695:
CI is failing on `golangci-lint-cli` and `golangci-lint-server` due to pre-existing ST1003/ST1000 naming violations on master, not introduced by this PR. Ready for review.

mesheryctl/internal/cli/root/adapter/adapter.go
```diff
@@ -276,42 +297,34 @@ func waitForValidateResponse(mctlCfg *config.MesheryCtlConfig, query string) (st
...
-				eventChan <- "successful"
-				utils.Log.Infof("%s\n%s", i.Data.Summary, i.Data.Details)
-			} else if strings.Contains(i.Data.Details, "error") {
-				eventChan <- "error"
-				utils.Log.Infof("%s", i.Data.Summary)
-			}
-		}
-	}()
+	defer timer.Stop()
+	eventChan := waitForSSEEvent(ctx, event, query, "%s\n%s", "%s")
```
lekaf974: in both cases log format and error log format are equals so IMO we don't need to pass them as argument and keep it hardcoded in the function for now

ishwar170695: Done. Format strings are now hardcoded inside the function. Both callers simplified.

mesheryctl/internal/cli/root/adapter/adapter.go
```diff
@@ -223,6 +224,37 @@ func sendOperationRequest(mctlCfg *config.MesheryCtlConfig, query string, delete
...
+			case <-ctx.Done():
+				return
+			case i, ok := <-event:
+				if !ok {
+					return
+				}
+				if strings.Contains(i.Data.Details, query) {
+					utils.Log.Infof(successLogFormat, i.Data.Summary, i.Data.Details)
+					eventChan <- "successful"
+					return
```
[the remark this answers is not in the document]
ishwar170695: Updated the adapter wait flow so the shared SSE matcher accepts the query in either summary or details, and it now checks error text case-insensitively across both fields

mesheryctl/pkg/utils/sse-client.go
```diff
@@ -79,16 +98,21 @@ func loop(reader *bufio.Reader, events chan Event) {
...
+						ev.Data = data
+						buf.Reset()
+						select {
+						case events <- ev:
+						case <-ctx.Done():
+							closeEvents()
+							return
+						}
+						ev = Event{}
+					}
```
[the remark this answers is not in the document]
ishwar170695: done

mesheryctl/internal/cli/root/adapter/adapter.go
```diff
@@ -223,6 +224,40 @@ func sendOperationRequest(mctlCfg *config.MesheryCtlConfig, query string, delete
...
+				if strings.Contains(summary, queryLower) || strings.Contains(details, queryLower) {
+					utils.Log.Infof(successLogFormat, i.Data.Summary, i.Data.Details)
+					eventChan <- "successful"
+					return
+				}
+				if strings.Contains(summary, "error") || strings.Contains(details, "error") {
+					utils.Log.Infof(errorLogFormat, i.Data.Summary)
+					eventChan <- "error"
+					return
+				}
```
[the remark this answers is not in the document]
ishwar170695: Done. Deploy and validate matching criteria separate by passing per-caller field selectors/predicates so deploy continues to match Details only while validate matches Summary, preserving the previous behavior.

mesheryctl/internal/cli/root/adapter/adapter.go
```diff
@@ -223,6 +224,40 @@ func sendOperationRequest(mctlCfg *config.MesheryCtlConfig, query string, delete
...
+				summary := strings.ToLower(i.Data.Summary)
+				details := strings.ToLower(i.Data.Details)
+				if strings.Contains(summary, queryLower) || strings.Contains(details, queryLower) {
+					utils.Log.Infof(successLogFormat, i.Data.Summary, i.Data.Details)
+					eventChan <- "successful"
+					return
+				}
+				if strings.Contains(summary, "error") || strings.Contains(details, "error") {
```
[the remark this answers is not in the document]
ishwar170695: Done. Moved the error branch ahead of the success check.

mesheryctl/internal/cli/root/adapter/adapter.go
```diff
@@ -238,32 +273,23 @@ func waitForDeployResponse(mctlCfg *config.MesheryCtlConfig, query string) (stri
...
+	defer timer.Stop()
+	eventChan := waitForSSEEvent(ctx, event, query)
 
 	select {
 	case <-timer.C:
 		return "", ErrTimeoutWaitingForDeployResponse
-	case event := <-eventChan:
-		if event != "successful" {
+	case eventStatus := <-eventChan:
+		if eventStatus != "successful" {
```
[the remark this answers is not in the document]
ishwar170695: Done. Switched the status values to typed constants to make the helper contract explicit and avoid string-literal drift.

mesheryctl/internal/cli/root/adapter/adapter.go
```diff
@@ -223,6 +231,46 @@ func sendOperationRequest(mctlCfg *config.MesheryCtlConfig, query string, delete
...
+					eventChan <- sseEventError
+					return
+				}
+				if matchSummary {
+					if strings.Contains(summary, queryLower) {
+						utils.Log.Infof(successLogFormat, i.Data.Summary, i.Data.Details)
+						eventChan <- sseEventSuccessful
+						return
+					}
+				} else if strings.Contains(details, queryLower) {
```
lekaf974: is the `else if` needed or `else` is enough

ishwar170695: Reason: Even tho `else if` is identical to `else` here, the reason for using `else if` is that it explicitly shows the intent of branching on two different strategies.

lekaf974: so a switch case looks more appropriate IMO

ishwar170695: Done. Refactored to a switch in the latest commit.

ishwar170695:
@lekaf974 Just a gentle ping on this PR. I've addressed the review feedback from the previous rounds, all checks are passing, and the PR should be ready for another look when you have a chance. Thanks!

mesheryctl/internal/cli/root/adapter/adapter.go
```diff
@@ -219,6 +227,45 @@ func sendOperationRequest(mctlCfg *config.MesheryCtlConfig, query string, delete
...
+			select {
+			case <-ctx.Done():
+				return
+			case i, ok := <-event:
+				if !ok {
+					return
+				}
+				summary := strings.ToLower(i.Data.Summary)
+				details := strings.ToLower(i.Data.Details)
+				if strings.Contains(summary, "error") || strings.Contains(details, "error") {
```
SuryanshGarg04: In the original implementation, the target `query` was checked before checking for `"error"`. 
Evaluating `"error"` first will cause a false negative if a successful event contains the word `"error"` in its summary/details (e.g. `"mesh is now installed with 0 errors"`).
Consider evaluating the success `query` match first before checking for `"error"`

SuryanshGarg04:
Looks good to me, just one edge case observation.
Left a review for it

alexquincy:

- `mesheryctl/internal/cli/root/adapter/adapter.go`
- `mesheryctl/pkg/utils/sse-client.go`

**What to do:**

SuryanshGarg04 (approved):

lekaf974:
@ishwar170695 can you check latest copilot review ?

PR #19693, opened 2026-05-26
chore: remove direct references from workflows

miacycle:
### `notify-providers.yaml`
- Replaces \`trigger-remote-provider-action@master\` with \`benc-uk/workflow-dispatch@v1\`, a published, well-maintained action that does the same workflow_dispatch.
- Dispatch target (\`meshery-extensions/meshery-extensions-packages\`) and the \`version\` input contract are unchanged.

### `integrations-updater.yml`
- Was: ~150 lines, performed all checkouts/builds/commits inline.
- Now: thin driver. Same triggers (push to its own path, daily cron, manual). Job body collapsed to:
  \`\`\`yaml
  uses: meshery-extensions/integrations-workflow/.github/workflows/publish.yml@master
  secrets: inherit
  \`\`\`
- The reusable workflow at meshery-extensions/integrations-workflow owns the actual publish behavior. **No publish behavior changed** - steps, paths, action versions, and commit messages match the prior in-repo job exactly.

## Notes

- Secrets pass through via \`secrets: inherit\`; no new secret plumbing required.
- The reusable workflow is public so it's callable from public callers; this matches meshery/meshery's visibility.


miacycle:
Addressing the three points from the Pull Request Reviewer comment:

**Pin to commit SHAs** — done in b97345c:
- `benc-uk/workflow-dispatch@31e2b3319479a63f0ab15bf800eff9e913504e26` (v1.3.2)
- `meshery-extensions/integrations-workflow/.github/workflows/publish.yml@40f2b0d9cf50193043560ff87a81c4cef1974b0e`

**Failure-notification step** — not removed. The `Send Email on Model Publish Failure` step (with `dawidd6/action-send-mail@v7`, `if: failure()`, same recipients) is preserved verbatim in the called workflow at `meshery-extensions/integrations-workflow/.github/workflows/publish.yml`. Because the driver `uses:` the reusable workflow with `secrets: inherit`, the failure path and the `MAIL_USERNAME`/`MAIL_PASSWORD` secrets propagate identically. No regression in alerting.

**Test-plan note** — the test plan items are manual post-merge verifications (only the daily cron / next tag push exercise the full publish flow against the real downstream repos); they'll be exercised by the next scheduled run.

PR #19756, opened 2026-05-29
CI: Optimize preview deployment with sparse checkout

### LINKED ISSUE #19678: Optimize deployment preview pruning with sparse checkout
## Context

Meshery's deployment preview workflow already has pruning logic that keeps preview deployments within a limit greater than 6. As a result, any PR preview older than that limit can be pruned and will only come back after a recent commit or push to that PR.

There is also a long-running step in the workflow that can likely be reduced by using sparse checkout instead of checking out the entire `gh-pages` branch.

## Reference implementation

Meshery has a similar deployment preview fix that uses sparse checkout for preview deployment assets:

- https://github.com/meshery/meshery/blob/master/.github/workflows/build-and-preview-site.yml#L85-L87

## Example of current waiting time

An example Meshery workflow run showing the long wait time:

- https://github.com/meshery/meshery/actions/runs/26299015072/job/77419065270

## Proposed change

Apply the same Meshery deployment preview fix to Meshery's deployment preview workflow so that it checks out only the required preview deployment paths from `gh-pages`, rather than the full branch. This should reduce the time spent in the preview pruning/deployment step while preserving the existing preview pruning behavior.

## Acceptance criteria

- Meshery deployment preview workflow uses sparse checkout for the relevant `gh-pages` preview deployment paths.
- Existing pruning behavior remains intact.
- Preview deployment step duration is reduced compared with full `gh-pages` checkout.

Bhumikagarggg:
**Notes for Reviewers**

- This PR fixes #19678


Rajesh-Nagarajan-11 (approved):

.github/workflows/build-and-preview-docs.yml
```diff
@@ -121,6 +116,9 @@ jobs:
         with:
           ref: gh-pages
           fetch-depth: 0
+          filter: blob:none
+          sparse-checkout: |
+            pr-preview
```
pontusringblom, line 121: @Bhumikagarggg does this work? Do you have a functional example? @Rajesh-Nagarajan-11, do you test results?

PR #19789, opened 2026-05-31
[UI] Fix deploy icon color in design action menu

Rajesh-Nagarajan-11:
**Notes for Reviewers**
- Fixes the Deploy icon color in the design card action dropdown by setting the icon fill to currentColor, matching the other action icons.
- Fixes inconsistent action icon colors in Designs grid and table views.
- Updates test file

Changes:

- Sets Deploy icon fill to currentColor.
- Replaces hardcoded Undeploy red with Sistent crimson[40].
- Applies consistent Undeploy icon color in both grid and table action menus.

| Before | After |
|--------|-------|
| Before | After |

| Before | After |
|--------|-------|
| Before | After |


Sbragul26 (approved):
LGTM :+1:

rishiraj38 (changes requested):
We should probably apply the same fix to the [table/columns](https://playground.meshery.io/configuration/catalog?view=table) view action menu as well.

Image

rishiraj38 (approved):
**LGTM!** Just a heads-up for future migrations: since we're planning to move all icon imports directly to `@sistent/sistent`, we should keep an eye on this behavior. For icons used inside dropdowns/action menus, we may need to explicitly pass `fill="currentColor"` to ensure hover and active state colors continue to work correctly after the migration.

Junnygram:
LGTM

PragalvaXFREZ (approved):
looks good to me 💯

banana-three-join (commented):
The current fill color for the undeploy icon is #F91313 but the fill color for the undeploy icon in other Layer5 products is #F44336 causing a divergence between the icons being utilized. Maybe this PR should pertain only to the fix of the deploy icon color and the undeploy icon with the proper standard color crimson color as a default for the export should be a Sistent PR.

Rajesh-Nagarajan-11:
> The current fill color for the undeploy icon is #F91313 but the fill color for the undeploy icon in other Layer5 products is #F44336 causing a divergence between the icons being utilized. Maybe this PR should pertain only to the fix of the deploy icon color and the undeploy icon with the proper standard color crimson color as a default for the export should be a Sistent PR.

I dont think we can add any color to icons in sistent , because the all sistent icons are in default fill ( default color) , so we need to change the color in other Layer5 Products is good call ! @banana-three-join

Issue #19335, opened 2026-05-13
[UI] Add "Evaluate" action button to Designs page for relationship evaluation


rishiraj38:
**Expected Behavior:**

The Designs configuration page (`/configuration/designs`) lacks a UI mechanism to evaluate design relationships. 
We need to:
1. Add an "Evaluate" button to the Grid View cards and Table View action dropdowns.
2. The button should trigger a `POST` request to `/api/meshmodels/relationships/evaluate` using the selected design file.
3. Display a toast notification with the evaluation results (e.g., action count and version) or an error message on failure.
### Mockups

<img width="1782" height="955" alt="Image" src="https://github.com/user-attachments/assets/5eba53df-063a-4c97-8601-2909ff10f63f" />

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)

- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

kanishksingh23:
Hey @rishiraj38, are you working on this issue?

charanachanta2:
Hey @rishiraj38 are you still working on this issue?

PR #19793, opened 2026-05-31
[UI] Add EVALUATE_RELATIONSHIPS permission key

rishiraj38:
## Description
Adds the `EVALUATE_RELATIONSHIPS` permission key to `ui/utils/permission_constants.ts` to enable UI-side permission gating for the upcoming "Evaluate Relationships" button on the Designs page.

The permission key matches the entry already seeded in `docs/static/excel/keys.csv` (line 21) with subject "Evaluate Relationships" and action "Evaluate relationships inside a design".

Related To #19335 

### DIFF
--- a/ui/utils/permission_constants.ts
+++ b/ui/utils/permission_constants.ts
@@ -248,6 +248,10 @@ export const keys = {
     subject: 'Export Design',
     action: '9a783f51-3b4a-47a6-a02e-b0db9e78cd85',
   },
+  EVALUATE_RELATIONSHIPS: {
+    subject: 'Evaluate Relationships',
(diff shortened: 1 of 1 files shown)


aabidsofi19:
needs to be used from schemas

rishiraj38:
> needs to be used from schemas

@aabidsofi19 PermissionKeys from @meshery/schemas only provide the UUID. But CAN() also needs the subject string to match against the provider's JWT.

All existing buttons use entries from `permission_constants.ts` since it pairs both the UUID and subject together. Should I use schemas for just the UUID, or add it to permission_constants.ts like the rest?

PR #19818, opened 2026-06-01
feat: add hardened runner step to GitHub workflows for enhanced security

ianrwhitney:
This PR adds a [security harden action](https://github.com/step-security/harden-runner) to a few of our workflows. This will allow our workflows to monitor outbound network traffic, source code integrity, and other functions - https://docs.stepsecurity.io/github-actions/harden-runner. 

This PR sets the action to audit only. After some analysis of the workflow summariese we can build further rules around it and switch action to block. 

I think we could benefit from visibility into the data across some of workflows to continue to iterate towards better security practices.

### DIFF
--- a/.github/workflows/build-and-preview-docs.yml
+++ b/.github/workflows/build-and-preview-docs.yml
@@ -32,6 +32,11 @@ jobs:
       PREVIEW_RETENTION_LIMIT: 6
 
     steps:
+      - name: Harden Runner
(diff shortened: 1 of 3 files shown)


leecalcote (commented):
A new tool - worth exploring. 👍  

Are we in need of creating a project-shared account or are the insights gleaned pushed back to the PRs?

ianrwhitney:
@leecalcote I didnt see that that was required for the action.

leecalcote (approved):

PR #19819, opened 2026-06-01
[server] Remove service-mesh-performance dependency and replace with local types

leecalcote:
## Summary

Resolves build failures from service-mesh-performance (SMP) module path mismatch by removing the external dependency and replacing it with local performance types.

## Problem

The `make ui-server` build was failing because:
- Upstream SMP v0.7.0 declares `module github.com/layer5io/service-mesh-performance` in go.mod
- Meshery imports `github.com/service-mesh-performance/service-mesh-performance/spec`
- Go cannot resolve this path mismatch without a replace directive
- Using a replace directive creates layer5io references, violating the hard constraint of no layer5io org references

## Solution

Created a local performance package with all required types and updated the codebase to use it instead of the upstream SMP:

### Changes

1. **New package**: `server/models/performance/types.go`
   - `PerformanceTestConfig` - performance test configuration
   - `ServiceMesh` - service mesh information with type mappings
   - `Client` - client configuration for load tests
   - `Protocol` enum (HTTP, GRPC, TCP)
   - Support for both `Clients` (new format) and `ClientsNested` (legacy format) for backward compatibility

2. **Updated imports** (18 files):
   - Server handlers: `load_test_handler.go`, `load_test_meshes_handler.go`, `load_test_preferences_handler.go`
   - Models: `performance_profiles.go`, `validators.go`, `test_profiles_persister.go`, provider files
   - CLI: `mesheryctl/internal/cli/root/perf/apply.go`, `mesheryctl/internal/cli/root/adapter/adapter.go`
   - Tests: `validators_test.go` with nested client types

3. **Fixed wasm build** (Makefile):
   - Removed `go mod tidy` from wasm-engine to prevent dependency reconciliation failures
   - Updated wasm_exec.js patching to use portable Perl instead of BSD sed

4. **Dependency cleanup**:
   - Removed layer5io/SMP from root `go.mod`
   - Updated `go.sum` and `server/policies/wasm/go.mod`

## Verification

- ✅ `make wasm-engine` - SUCCESS
- ✅ `make build-server` - SUCCESS
- ✅ `go test ./server/models` - All tests pass
- ✅ No layer5io references in source code
- ✅ No SMP imports in Go files

## Type Safety

The local types maintain API compatibility with the original SMP types while being simpler and not requiring external dependencies. The validator and handlers work seamlessly with both the new and legacy client formats.

## Notes

- The performance testing infrastructure continues to work without external SMP dependency
- go.sum contains layer5io checksums only in the wasm submodule (acceptable, not imports)
- All performance test functionality preserved

## Summary by CodeRabbit

- **New Features**
  - Added built-in performance testing models for protocols, load generators, clients, test configurations, and service mesh types.
  - Added support for configuring performance tests using both current and legacy client formats.
  - Performance test preferences now use standard JSON serialization.

- **Bug Fixes**
  - Improved validation for performance test clients, protocols, load generators, endpoints, and URLs.
  - Service mesh listings now exclude invalid and unset entries consistently.


server/models/performance/types.go
```diff
@@ -0,0 +1,110 @@
...
+// LoadGenerator holds load generator configuration
+type LoadGenerator int32
+
+const (
+	LoadGenerator_UNSET LoadGenerator = 0
+	LoadGenerator_FORTIO LoadGenerator = 1
+)
+
+// Client holds client configuration for load test
+type Client struct {
```
aabidsofi19, line 39: @leecalcote  should  these structs/types go to schemas and generated there ?

PragalvaXFREZ:
@leecalcote, @aabidsofi19, this PR (#19605) also removes Meshery’s active root-module dependency on the `service-mesh-performance` / SMP module and moves the performance profile flow onto schemas-native types.

The large churn is because the performance profile endpoints and callers were migrated end-to-end across server handlers/models, RTK hooks, and mesheryctl. I also moved UI performance result fetching from GraphQL to schema-generated REST hooks because profiles and results are consumed together; keeping profiles camelCase via REST while results stayed snake_case via GraphQL would split one performance flow across two incompatible casing contracts.

PR #19822, opened 2026-06-01
chore(rtk): remove /evaluate path exception from mesheryApiPath

yi-nuo426:
## Description

Removes the now-dead `/evaluate` special-case branch from `mesheryApiPath` in preparation for the [meshery/schemas#916](https://github.com/meshery/schemas/pull/916) release.

### Background

`mesheryApiPath` prefixes bare paths with `/api/` but has a passthrough exception for paths starting with `/evaluate`. This exception was added to prevent the `@meshery/schemas`-generated `postEvaluate` endpoint (whose URL was `/evaluate`) from being incorrectly prefixed to `/api/evaluate`.

### Why this is now dead code

[meshery/schemas#916](https://github.com/meshery/schemas/pull/916) corrects the spec:
- Path: `/evaluate` → `/api/meshmodels/relationships/evaluate`
- operationId: `evaluateRelationships`

With the new URL already starting with `/api`, it is handled correctly by the existing `/api/*` passthrough branch. The `/evaluate` exception is no longer needed.

### Changes

| File | Change |
|---|---|
| `ui/rtk-query/index.ts` | Remove `if (path.startsWith('/evaluate'))` branch (3 lines) |
| `ui/rtk-query/__tests__/index.test.ts` | Replace the 2 `/evaluate` assertions with a single assertion documenting the canonical new URL passes through correctly |

> **⚠️ Merge dependency:** This PR should be merged **after** `@meshery/schemas` is released with [meshery/schemas#916](https://github.com/meshery/schemas/pull/916).

## Related

- meshery/schemas#916
- layer5labs/meshery-extensions#4259 (companion PR in Kanvas)

## Checklist
- [x] Signed commits
- [x] Tests updated and passing (19/19)


yi-nuo426:
@rishiraj38 👀

rishiraj38:
While reviewing the impact of `meshery/schemas#916`, I found that the change will affect the existing relationship evaluation E2E tests in Meshery UI.

**Affected files:**

* [`ui/tests/e2e/relationship_evaluation.spec.ts`](https://github.com/meshery/meshery/blob/master/ui/tests/e2e/relationship_evaluation.spec.ts#L117)
* `ui/tests/e2e/relationship_evaluation_properties.spec.js`

These tests make direct POST requests to `/api/meshmodels/relationships/evaluate` using the following payload:

```json
{
  "design": { ... },
  "options": {
    "returnDiffOnly": false,
    "trace": false
  }
}
```

Since `meshery/schemas#916` renames the `trace` property to `enableTrace`, the tests will continue sending the old field after the schema update. As a result, trace functionality may no longer be enabled during test execution, and future strict schema validation could potentially cause these requests to fail.

rishiraj38:
Agree with Copilot's review. The upgraded schema defines endpoints as `/api/performance/...` which mismatches the current server routes `(/api/user/performance/...)`, causing the unit test failure.

rishiraj38:
I dug into this a bit and it looks like an RTK Query endpoint collision issue.

`getPerformanceProfiles` and `deletePerformanceProfile` have the exact same names as the schema-generated endpoints. Since RTK Query defaults to `overrideExisting: false`, our local definitions are ignored and the schema paths (`/api/performance/...`) are used instead.

The other endpoints continue to work because their names don't match the schema-generated ones.

Screenshot 2026-06-02 at 11 18 35 AM

leecalcote (approved):

PR #19863, opened 2026-06-03
fix(prometheus): prevent server crash on logger initialization failure

### LINKED ISSUE #19855: [Server] Prometheus register action exits process on logger initialization failure
### Current Behavior

`RegisterAction.Execute` initializes a logger while handling the Prometheus connection register action. If `logger.New` returns an error, the code logs the error and calls `os.Exit(1)`:

```go
log, err := logger.New("meshery", logger.Options{
    Format:   logger.SyslogLogFormat,
    LogLevel: logLevel,
})
if err != nil {
    logrus.Error(err)
    os.Exit(1)
}
```

This can terminate the entire Meshery server process while handling a single connection registration action. Request-path code should not abruptly exit the process.

### Expected Behavior

If logger initialization fails, `RegisterAction.Execute` should return an error and, where possible, an error event. The caller/state machine should handle the failure without terminating the server process.

### Screenshots/Logs

N/A

### Environment

- **Host OS:** N/A
- **Platform:** N/A

### To Reproduce

1. Induce a failure in logger initialization in `RegisterAction.Execute`.
2. Call the connection registration endpoint/path that invokes the Prometheus register action.
3. Observe that the process exits with status code 1 instead of returning an action error.

### Evidence

- `server/machines/prometheus/register.go`
- `RegisterAction.Execute`
- Verified locally that this request-path action calls `os.Exit(1)` after `logger.New` fails.

### Suggested Fix

Replace `os.Exit(1)` with normal error propagation. For example, return `machines.NoOp`, an error event if one can be safely constructed at that point, and the logger initialization error. This allows the server to continue running and lets the existing action/state-machine error handling process the failure.

Kushal-911:
**Notes for Reviewers**

- This PR fixes #19855 

### DIFF
--- a/server/handlers/workspace_handlers.go
+++ b/server/handlers/workspace_handlers.go
@@ -97,7 +97,7 @@ func (h *Handler) GetWorkspacesHandler(w http.ResponseWriter, req *http.Request,
 	if orgID == "" {
 		missingInput := models.ErrWorkspaceMissingInput()
 		h.log.Error(missingInput)
-		writeJSONError(w, missingInput.Error(), http.StatusBadRequest)
+		writeMeshkitError(w, missingInput, http.StatusBadRequest)
 		return
 	}
 	resp, err := provider.GetWorkspaces(token, q.Get("page"), q.Get("pagesize"), q.Get("search"), q.Get("order"), q.Get("filter"), orgID)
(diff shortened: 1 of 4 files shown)


ishwar170695:
I noticed this PR also changes `workspace_handlers.go` by replacing `GetWorkspaces(...)` with `GetWorkspaceByID(...)`. Is that change related to [#19855](https://github.com/meshery/meshery/issues/19855), or was it included accidentally?

Maanvi212006:
@Kushal-911 are you still working on it?

PR #19873, opened 2026-06-04
fix(connections): handle registry lookup errors in registration init

### LINKED ISSUE #19856: [Server][API] Connection registration init masks registry errors as unknown connection kind
### Current Behavior

`handleRegistrationInitEvent` ignores errors returned by `registryManager.GetEntities` while initializing a connection registration payload. The first registry lookup currently discards the error:

```go
connectionComponent, _, _, _ := h.registryManager.GetEntities(compFilter)
if len(connectionComponent) == 0 {
    writeMeshkitError(w, ErrUnknownConnectionKind(payload.Kind), http.StatusBadRequest)
    return
}
```

If the registry backend or registry service fails, the handler sees an empty result and returns `ErrUnknownConnectionKind` with HTTP 400. This makes a server-side registry failure look like a client-side invalid connection kind.

The second registry lookup for the credential schema also ignores the returned error, so credential schema lookup failures are silently dropped.

### Expected Behavior

When `registryManager.GetEntities` returns an error during connection registration initialization, the handler should:

- log the registry error with enough context to debug the failed lookup
- return a 500 Internal Server Error rather than a 400 Bad Request
- avoid masking registry backend failures as `ErrUnknownConnectionKind`

If the lookup succeeds and no connection component exists, returning `ErrUnknownConnectionKind` as a 400 remains appropriate.

### Screenshots/Logs

N/A

### Environment

- **Host OS:** N/A
- **Platform:** N/A

### To Reproduce

1. Cause the registry backend/service used by `registryManager.GetEntities` to return an error.
2. Trigger the connection registration init flow, for example through Add Connection in the UI or the corresponding API path.
3. Observe that the API reports an unknown connection kind with HTTP 400 instead of surfacing a server-side registry failure with HTTP 500.

### Evidence

- `server/handlers/connections_handlers.go`
- `handleRegistrationInitEvent`
- Verified locally that both `GetEntities` calls in this handler currently discard returned errors.

### Suggested Fix

Capture and handle the `GetEntities` errors before checking result length. The first lookup should return a 500 if registry access fails. The credential lookup should also handle registry errors explicitly rather than silently omitting credential schema data.

Kushal-911:
**Notes for Reviewers**

- This PR fixes #19856 


server/handlers/connections_handlers.go
ishwar170695, line 129: I noticed this `GetEntities `call still ignores the returned error `(_, _, _, _)`. Since the issue mentions handling errors for both registry lookups, should we add similar error handling here as well?

ishwar170695 (commented):
I didn't notice any test updates in this PR. Since this changes the handler's response behavior, would it make sense to add regression coverage for:

- registry lookup error -> 500
- successful lookup with empty result -> 400 (`ErrUnknownConnectionKind`)
- successful lookup with valid result -> registration init succeeds

These were also called out in the issue discussion.

PR #19936, opened 2026-06-08
fix(designs): render imported designs misclassified as legacy v1alpha2

leecalcote:
## Symptom

Designs imported from a Kubernetes manifest, Helm chart, or Docker Compose file - whether dropped on the Kanvas canvas or brought in through the Import Design modal - stopped rendering. They open to an empty canvas.

`GET /api/pattern/{id}` returns `"components": null` for a freshly imported design, so Kanvas has nothing to render.

## Root cause

`NewPatternFileFromK8sManifest` produces designs in **v1beta3** form (`designs.meshery.io/v1beta3`). On read, `GetMesheryPatternHandler` gates the legacy v1alpha2 -> v1beta3 migration on `IsDesignInAlpha2Format`, which recognized only **v1beta1** as a current schema:

```go
val, ok := design["schemaVersion"]
if ok && val == v1beta1.DesignSchemaVersion { // only v1beta1
    return false, nil
}
return true, nil // v1beta3 falls through here -> treated as "alpha2"
```

So every v1beta3 design was misclassified as alpha2 and run through `convertV1alpha2ToV1beta3`, which reparses it as a `v1alpha2.PatternFile`. That type carries `services`, not `components`, so the conversion produced a design with **zero components**. One call site additionally re-persisted the emptied design, corrupting it in storage.

The regression was introduced when the import pipeline advanced to v1beta3 while this read-path guard's whitelist stayed at v1beta1.

## Fix

Detect alpha2 by the **absence** of a canonical `designs.meshery.io/*` schemaVersion - legacy v1alpha2 carries a top-level `services` map and no versioned schemaVersion - instead of whitelisting a single "current" version. This keeps current and future design schema versions (v1beta1, v1beta3, ...) out of the lossy migration path and prevents the same regression at the next schema bump.

The change is in the shared `IsDesignInAlpha2Format`, so it covers all three call sites that gate on it (one GET handler and two POST handlers), including the one that re-persisted the gutted design.

## Reproduction (before this PR)

1. Import an nginx Deployment manifest via the Import Design modal (or drop it on the canvas).
2. `POST /api/pattern/import` returns the converted design with its components.
3. `GET /api/pattern/{id}` returns `"components": null`.
4. Kanvas renders an empty canvas.

Confirmed identically on both the local provider and the Layer5 remote provider.

## Recovery

Designs already corrupted in storage are recoverable: the original uploaded source is retained in the record's `SourceContent` and can be re-converted.

## Test plan

- [x] `go test ./server/models/pattern/utils/` - new `TestIsDesignInAlpha2Format` covers v1beta3 (the regression), v1beta1, a future `designs.meshery.io/*` version (all classified current), and genuine v1alpha2 (classified legacy). Existing package tests still pass.
- [x] `go vet` and `go build` of `server/models/pattern/utils` and `server/handlers` - clean (removed the now-unused `v1beta1` import).

### DIFF
--- a/server/models/pattern/utils/utils.go
+++ b/server/models/pattern/utils/utils.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 
 	"github.com/meshery/meshkit/encoding"
-	"github.com/meshery/schemas/models/v1beta1"
 )
 
 // RecursiveCastMapStringInterfaceToMapStringInterface will convert a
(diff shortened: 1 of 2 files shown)


leecalcote:
Duplicate of #19934 - both built from the same commit (82028f9a508). Consolidating on #19934, where the review feedback is being addressed.

PR #19937, opened 2026-06-08
fix: persist dark/light theme preference across page refreshes

### LINKED ISSUE #8772: [UI] User Preference of Dark/Light mode lost on refresh in extension point
### Current Behavior

https://github.com/meshery/meshery/assets/7570704/559305c2-094f-4a8d-9591-d089dad905b0

### Environment

- Browser: Safari
- Host OS: Mac
- Meshery Server Version: stable-v0.6.139

---

Maanvi212006:
**Notes for Reviewers**
The user's dark/light mode preference was being lost on page refresh. Theme state lived only in React component state, which resets on refresh, so toggling had no persistence.

- This PR fixes #8772 

**Changes made**
- Rewrote useThemePreference hook (themes/hooks.tsx) with the following behavior:
1. On app start / while RTK Query is loading (isLoading = true), reads localStorage directly as a placeholder to prevent theme flash
2. Once RTK Query resolves (isLoading = false), remote provider preference takes over as the source of truth
3. On toggle, writes to localStorage directly and dispatches a theme-change window event so all instances of useThemePreference (including the one in _app.tsx) stay in sync
4. For remote provider users, also calls handleUpdateUserPref to persist the preference server-side

- Added explicit JSON.stringify in rtk-query/user.ts for the POST body

https://github.com/user-attachments/assets/9fa08618-807b-43dd-acc0-38a889fbc40a

## Summary by CodeRabbit

* **New Features**
  * Theme selection now persists across sessions and stays synchronized across open areas of the app.
  * When no saved preference exists, the app can follow the device’s light or dark mode.
  * Theme controls now clearly display “Dark Mode” or “Light Mode.”

* **Bug Fixes**
  * Improved reliability when saving theme and other user preferences.
  * Added consistent error notifications when preference updates fail.
  * Prevented theme changes while an update is still processing.

### DIFF
--- a/ui/components/user-preferences/index.tsx
+++ b/ui/components/user-preferences/index.tsx
@@ -57,9 +57,7 @@ import { SecondaryTab, SecondaryTabs } from '../dashboard/style';
 import { useDispatch, useSelector } from 'react-redux';
 import { toggleCatalogContent, updateProgress } from '@/store/slices/mesheryUi';
 
-interface ThemeTogglerProps {
-  handleUpdateUserPref: (_theme: string) => void;
-}
+interface ThemeTogglerProps {}
 
 interface ThemeComponentProps {
   mode: 'light' | 'dark';
@@ -104,7 +102,7 @@ interface _UserData {
   };
 }
 
-const ThemeToggler: React.FC<ThemeTogglerProps> = ({ handleUpdateUserPref }) => {
+const ThemeToggler: React.FC<ThemeTogglerProps> = () => {
   const Component: React.FC<ThemeComponentProps> = ({ mode, toggleTheme }) => {
     return (
       <div>
@@ -113,10 +111,9 @@ const ThemeToggler: React.FC<ThemeTogglerProps> = ({ handleUpdateUserPref }) => 
           checked={mode === 'dark'}
           onChange={() => {
             toggleTheme();
-            handleUpdateUserPref(mode === 'dark' ? 'light' : 'dark');
           }}
         />
-        Dark Mode
+        {mode === 'dark' ? 'Dark Mode' : 'Light Mode'}
       </div>
     );
   };
@@ -175,7 +172,11 @@ const UserPreference: React.FC<UserPreferenceProps> = (props) => {
         });
       })
       .catch(() => {
-        handleError('There was an error sending your preference');
+        updateProgress({ showProgress: false });
+        notify({
+          message: 'There was an error sending your preference',
+          event_type: EVENT_TYPES.ERROR,
+        });
       });
   };
 
@@ -187,11 +188,6 @@ const UserPreference: React.FC<UserPreferenceProps> = (props) => {
       setPerfResultStats(!perfResultStats);
       handleChange(name, !perfResultStats);
     }
-  };
-
-  const handleError = (name) => () => {
-    updateProgress({ showProgress: false });
-    notify({ message: name, event_type: EVENT_TYPES.ERROR });
   };
 
   const handleChange = (name, resultState) => {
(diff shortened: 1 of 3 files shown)


Junnygram:
**Edge case: localStorage may shadow updated remote preferences across devices**

The current priority order:
```js
const mode = storedMode || data?.remoteProviderPreferences?.theme || systemPref || 'dark';
```

Because `storedMode` takes the highest priority, cross-device theme syncing can silently break. Scenario:
1. User updates their theme on **Device A** → remote preference is updated.
2. User opens the app on **Device B**, which has a stale value in `localStorage`.
3. Device B reads `storedMode` first and **ignores the updated remote preference entirely**.

**Suggestion:** treat `localStorage` as an initial placeholder only — used while remote preferences are still loading — then defer to the remote value once it resolves:

```js
const mode = isRemoteLoading
  ? (storedMode || systemPref || 'dark')   // temporary until remote loads
  : (data?.remoteProviderPreferences?.theme || storedMode || systemPref || 'dark');
```

This way, remote preferences win when available, `localStorage` just prevents a flash of unstyled/wrong theme on initial render, and we stay consistent across devices.

ui/store/slices/themeSlice.ts
aabidsofi19: we should not store in local state . rtq query state should be be the only source

Maanvi212006:
Hi @Junnygram @aabidsofi19 
Thanks for the feedback!

Just want to confirm my understanding before making changes:

The current implementation stores theme in a Redux slice (`themeSlice.ts`) which syncs to `localStorage`. Your suggestion is to remove this slice entirely and instead read/write `localStorage` directly in the hook — keeping RTK Query cache as the sole Redux state for theme.

So the approach would be:
- On app start / while `isLoading = true` → read `localStorage.getItem('theme')` directly as a temporary placeholder to prevent theme flash
- Once RTK Query resolves (`isLoading = false`) → remote preference takes over as the source of truth
- On toggle → write to `localStorage.setItem('theme', newTheme)` directly (no Redux dispatch) and also call `handleUpdateUserPref` for remote provider users

Is this the direction you had in mind? Happy to proceed with the changes once confirmed.

ui/store/slices/themeSlice.ts
Maanvi212006: @aabidsofi19 updated

ui/themes/hooks.tsx
```diff
@@ -1,9 +1,24 @@
...
 
+const THEME_STORAGE_KEY = 'meshery-theme';
+
+const getStoredTheme = (): string | null => {
+  try {
+    return localStorage.getItem(THEME_STORAGE_KEY);
+  } catch {
+    return null;
+  }
+};
```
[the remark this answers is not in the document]
Maanvi212006: updated the PR description

ui/components/user-preferences/index.tsx
```diff
@@ -112,10 +110,9 @@ const ThemeToggler: React.FC<ThemeTogglerProps> = ({ handleUpdateUserPref }) =>
           checked={mode === 'dark'}
           onChange={() => {
             toggleTheme();
```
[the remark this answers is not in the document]
Maanvi212006: fixed

ui/themes/hooks.tsx
Oluwatunmise-olat: The final 'dark' hard-codes dark as the default even when systemPref is already available. At this point systemPref would only be falsy if useGetSystemTheme returns an empty string  which it already defaults to 'dark' internally. So this trailing 'dark' is redundant at best. Could you either remove it and let systemPref be the final fallbac

Maanvi212006:
image

its pre-existing, not related to my code

Maanvi212006:
@aabidsofi19 @Oluwatunmise-olat I have addressed the requested changes. ready for review

KhushamBansal (commented):
@Maanvi212006 Please ensure the lint check passes.

Maanvi212006:
@KhushamBansal I checked, its not related to my PR. Its in login.js file which is not in the scope of this PR
image

saurabhraghuvanshii (dismissed):

ui/rtk-query/user.ts
```diff
@@ -76,7 +76,7 @@ export const userApi = api
           headers: {
             'Content-Type': 'application/json;charset=UTF-8',
           },
-          body: queryArg.body,
+          body: JSON.stringify(queryArg.body),
```
saurabhraghuvanshii, line 79: @Maanvi212006  why this?

saurabhraghuvanshii (changes requested):

saurabhraghuvanshii (approved):

KhushamBansal (approved):

PR #19953, opened 2026-06-10
[mesheryctl] fix: restore port-forward in e2e test setup for k8s

Omkar-Ugal:
## Root Cause

`port_forwarding()` was removed in c79fa05 assuming `mesheryctl system start` 
would handle it. But it doesn't - only `mesheryctl system dashboard` does.

So `localhost:9081` was never reachable during tests, causing all 
server-dependent tests to fail with:
`Unable to connect to Meshery server at http://localhost:9081`


Omkar-Ugal:
Closing: fix needed in mesheryctl-e2e.yaml instead

PR #19961, opened 2026-06-10
fix(ui): resolve undefined @sistent/sistent imports that crash on render

leecalcote:
## Summary

Follow-up to #19957. While fixing the Connections-page crash there, I found the same undefined-import pattern in other components. Each imports a name the installed `@sistent/sistent` does **not** export, so it resolves to `undefined` and throws **React error #130** (_"Element type is invalid … got: undefined"_) when the component renders — taking down that view via the page-level ErrorBoundary.

## Fixes

| File | Bad import(s) | Fix | Where it crashed |
|---|---|---|---|
| `components/CustomToolbarSelect.tsx` | `CompareArrows`, `GetApp`, `IndeterminateCheckBox` | `*Icon` aliases | Performance + Designs tables, on row selection |
| `components/designs/lifecycle/DeployConfirmationModal.tsx` | `Search` | `SearchIcon as Search` | Deploy / undeploy / validate dialog |
| `components/telemetry/grafana/GrafanaCharts.tsx` | `ExpansionPanelDetails` + `ExpansionPanel`/`ExpansionPanelSummary` (from `../../ExpansionPanels`, **a module that doesn't exist in the repo**) | migrate to Sistent's MUI-v5 `Accordion`/`AccordionSummary`/`AccordionDetails`; drop the dead import | telemetry/Grafana view |

Sistent ships the `*Icon`-suffixed glyph names and the `Accordion*` family, not the bare/v4 names. The icon fixes alias on import so render sites stay unchanged.

Also: three `Theme` value-imports → `type Theme`. Under this repo's `verbatimModuleSyntax: true`, a type imported without `type` is emitted as a runtime import of a type-only export (undefined at runtime) — the benign tail of the same class, and it lets the new guard stay allowlist-free.

## New regression guard

`ui/__tests__/architecture/sistent-imports-resolve.test.ts` imports the **real** `@sistent/sistent` and asserts every named **value** import across the source tree resolves against `Object.keys()` (type-only imports skipped). This is the generalized form of the icon-barrel guard from #19957 — it catches this entire bug class, which **neither `tsc`** (Sistent's published types and runtime can diverge) **nor the mocked component tests** (the mock supplies whatever the test asks for) can. Proven to fail on a reintroduced bad import and pass after the fixes.

## Note on GrafanaCharts

`GrafanaCharts` is currently unreferenced (its only test is `describe.skip`), so it isn't in the build graph today — that's why its missing-module import hadn't reddened CI. Fixed in place (rather than deleted) so it's correct and renderable if it's wired back up.

## Test plan

- [x] `npx vitest run` (full suite) → **380 files / 2556 tests passed, 31 skipped, 0 failed** (incl. the new guard)
- [x] Reverted one fix and confirmed the architecture guard fails with the exact offender, then restored
- [x] `npx tsc --noEmit` → clean
- [x] `npx eslint` + `npx prettier --check` on all changed files → clean
- [x] Updated the `CustomToolbarSelect` test mock to the new icon names

### DIFF
--- a/ui/__tests__/architecture/sistent-imports-resolve.test.ts
+++ b/ui/__tests__/architecture/sistent-imports-resolve.test.ts
@@ -1,0 +1,124 @@
+import { describe, expect, it } from 'vitest';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import * as Sistent from '@sistent/sistent';
+
+/**
+ * Architecture guard: every value imported by name from `@sistent/sistent`
+ * must exist as a real runtime export of the installed package.
+ *
+ * This catches a recurring, whole-page-killing bug class:
+ *
+ *   import { DeleteForever } from '@sistent/sistent';   // ❌ runtime undefined
+ *
+ * Sistent ships the `*Icon`-suffixed names (`DeleteForeverIcon`, `SearchIcon`,
+ * `CompareArrowsIcon`, `GetAppIcon`, `IndeterminateCheckBoxIcon`, …) and the
+ * MUI-v5 `Accordion*` family rather than the v4 `ExpansionPanel*` names. An
+ * import that names something the package does not export resolves to
+ * `undefined` at runtime; rendering it throws React error #130 ("Element type
+ * is invalid … got: undefined") and the page-level ErrorBoundary swallows the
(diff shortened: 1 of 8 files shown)


winkletinkle (approved):
lgtm

PR #19965, opened 2026-06-10
fix(kanvas): prevent blank canvas when reopening v1beta3 designs

### LINKED ISSUE #19930: [Bug] Saved Kanvas design opens but its content is not visible after refresh
### Current Behaviour

After creating and saving a Kanvas design, refreshing/reopening the page makes the design appear to load, but the actual content is not visible.

The editor itself loads successfully and behaves as if the design has been opened:

- The Kanvas editor UI is visible.
- The design context appears to be present.
- The UI shows `Validating design`.
- Network requests shown in DevTools return `200` for many assets.
- However, the previously created canvas elements are not visible.

### Expected Behaviour

When a saved design is reopened or the page is refreshed, all previously saved canvas elements should be restored and visible on the canvas.

If the design cannot be restored, Kanvas should show a clear error message explaining whether the issue is related to design fetch failure, validation failure, corrupted design data, missing canvas elements, or rendering failure.

### Actual Behaviour

The design/editor page opens, but the saved canvas content is not rendered.

The UI appears to stay around the `Validating design` state, and the canvas appears empty.

Bakul2006:
**Notes for Reviewers**
**Root Cause**
processDesign() in ui/utils/utils.tsx only accepted designs.meshery.io/v1beta1. However, reopened designs were being returned with designs.meshery.io/v1beta3. The helper treated these designs as invalid and returned empty component arrays, causing the Kanvas canvas to render nothing.
**Changes Made**
Updated processDesign() to accept the current designs.meshery.io/v1beta* schema family while still rejecting truly invalid schema versions.
Preserved design components during reopen instead of dropping them.
Added a regression test to verify that v1beta3 designs retain their components and annotations correctly.
**Validation**
Verified locally by reopening saved designs and ensuring that components are rendered correctly after refresh/reopen.

Test results:
cd ui
npx vitest run utils/__tests__/utils.test.tsx
1 test file passed
74 tests passed
0 tests failed
- This PR fixes #19930 
- /home/bakul/Pictures/Screenshots/Screenshot from 2026-06-11 00-19-00.png
This array named components in the utils.tsx was empty so when it loads the design on the kanvas again it return an empty array this was the major issue with this..


Bakul2006:
Screenshot from 2026-06-11 00-19-00

Bhumikagarggg (changes requested):
@Bakul2006 Please address gemini suggestion

Maanvi212006:
@Bakul2006 sign off the commits

Bakul2006:
> @Bakul2006 Please address gemini suggestion

Addressed the Gemini suggestion by replacing the prefix-based check with a stricter regex validation for valid v1beta schema versions.

aabidsofi19 (approved):

PR #20038, opened 2026-06-13
[UI] Fix: Registry Models tab — components and relationships not clickable / deeplinking broken

YASHMAHAKAL:
## Description

Fixes two separate regressions in the Registry **Models** tab where nested components and relationships were non-interactive.

### Root Causes

1. **Event bubbling** — clicking a nested component/relationship fired the parent version and model `onClick` handlers last, overwriting the details panel with the wrong data.
2. **Deeplink overwrite** — `MesheryTreeView`'s deeplink `useEffect` called `setShowDetailsData` with an empty object for nested items (which aren't in the top-level models array), wiping the panel immediately after it was set correctly.
3. **Missing deeplink handling** — `VersionedModelRelationshipTree` had no `useEffect` to handle page-load deeplinking to a specific relationship, unlike the component tree.
4. **Stale closure / infinite loop risk** — the existing deeplink `useEffect` in `VersionedModelComponentTree` had `showDetailsData` in its dependency array, causing a write→trigger→write loop.

### Changes

| File | Fix |
|---|---|
| `MesheryTreeView.tsx` | Skip `setShowDetailsData` when item not found in top-level data (nested items) |
| `MesheryTreeViewItem.tsx` | `e.stopPropagation()` on version node; pass `showDetailsData` to `VersionedModelRelationshipTree` |
| `VersionedModelComponentTree.tsx` | Replace stale-closure deeplink effect with `useRef` guard pattern; `e.stopPropagation()` on component items |
| `VersionedModelRelationshipTree.tsx` | Add `showDetailsData` prop + `useRegistryRouter` hook + deeplink `useEffect` with `useRef` guard |
| `RelationshipTree.tsx` | `e.stopPropagation()` on kind-group and individual relationship items |

---
**Note:** This PR has a line overlap with #20033 (`RelationshipTree.tsx` onClick handlers). Please merge **after** #20033 to simplify conflict resolution.

---


fitzergerald:
Make sure to highlight this more prominently in the future.

⚠️ Note: This PR has a line overlap with https://github.com/meshery/meshery/pull/20033 (RelationshipTree.tsx onClick handlers). Please merge after https://github.com/meshery/meshery/pull/20033 to simplify conflict resolution. ⚠️

pontusringblom:
merge conflict resolved.

pontusringblom:
Copilot review requested.

leecalcote (approved):
## Review — Registry Models tab interactivity & deeplinking

**What this does:** Fixes the regression where nested components/relationships in the Models tab were non-interactive, and restores page-load deeplinking to a specific component or relationship.

**Correctness — verified against the tree `itemId` scheme:**
- A component `itemId` is `${modelId}.${versionId}.1.${componentId}` and a relationship is `${modelId}.${versionId}.2.${firstRelId}.${relId}`. The new `selectedIdArr[2] === '1' | '2'` group-node guards plus the `length > 3` checks line up correctly, so a deeplink only resolves inside the matching model+version subtree.
- The `MesheryTreeView` change is safe: `getFilteredDataForDetailsComponent` always returns `{ type, data: resultObject || {} }`, so `Object.keys(showData.data)` can't throw, and skipping the empty-object case correctly stops the top-level effect from wiping the panel that a nested `onClick` just populated.
- The `useRef` + separate sync-effect pattern correctly breaks the previous write→trigger→write loop (drops `showDetailsData` from the dependency array while still reading a fresh value), and the `id` equality guard avoids redundant `setState`. This is the right fix for the stale-closure / infinite-loop concern.
- `e.stopPropagation()` on the version/component/relationship nodes (and the kind-group node via its existing `closest()` data-id guard) prevents the parent handlers from overwriting the details panel.

**Conventions / tests:** Mirrors the existing component-tree pattern; the relationship test file adds the `useRegistryRouter` mock and `showDetailsData` prop. CI is green — eslint, ui-unit-tests, ui-audits, and build all pass.

**Non-blocking suggestions (follow-up, not required for merge):**
1. The `'1'` / `'2'` group-node discriminators are magic strings (documented in comments) — a shared named constant would be more robust if the `itemId` scheme ever changes.
2. Minor: the new `MesheryTreeView` condition mixes `showData.data` (no optional chaining) with `showData?.data?.id` — harmless given the helper's contract, just inconsistent.
3. Deeplinking through a `registrantID` prefix (Registrants tab) still won't match `selectedIdArr[0] === modelDef.id` — pre-existing and out of scope here, worth a follow-up.
4. Consider a unit test asserting `setShowDetailsData` fires for a matching `selectedItemUUID`, to lock in the new deeplink behavior.

Overall: correct, well-scoped, and clearly reasoned. Approving.

PR #20133, opened 2026-06-18
fix(security): prevent path traversal in file handlers

### LINKED ISSUE #20076: Unauthenticated arbitrary file read via /api/system/fileView and /api/system/fileDownload
Hi team,

The security inbox bounded:
```
Address not found
--
Your message wasn't delivered to security@meshery.io because the address couldn't be found, or is unable to receive mail.
```

Wanted to provide details on an unauthenticated arbitrary file read in the Meshery server, confirmed against v1.0.43 (meshery/meshery:stable-latest).

The endpoints GET /api/system/fileView and GET /api/system/fileDownload take a file query parameter and pass it directly to os.Open in server/handlers/common_handlers.go (ViewHandler and DownloadHandler) with no path containment, allowlist, or traversal filtering. Both routes are registered with models.NoAuth in server/router/server.go, and because the default deployment does not set the PROVIDER environment variable, AuthMiddleware leaves NoAuth routes unauthenticated. The handlers are plain http.HandlerFunc with no session injection, so no token or cookie is needed.

shteypandey28-hue:
- Implement SafeFilePath to validate requested file paths
- Ensure absolute paths are restricted to allowed directories like ~/.meshery and os.TempDir()
- Prevent relative path traversals
- Return 400 Bad Request for invalid file paths

**Notes for Reviewers**

- This PR fixes #20076 


NSTKrishna:
Please add the issue number to the PR description

NSTKrishna:
also take a look and resolve the comment by copilot

ishwar170695:
Please consider and respond to each comment in which review is offered whether in explanation of why you are rejecting the feedback or in explanation of how you incorporated the feedback.

shteypandey28-hue:
> Please consider and respond to each comment in which review is offered whether in explanation of why you are rejecting the feedback or in explanation of how you incorporated the feedback.

sorry , for that but I noticed that , I have worked on that ..

ishwar170695:
> > Please consider and respond to each comment in which review is offered whether in explanation of why you are rejecting the feedback or in explanation of how you incorporated the feedback.
> 
> sorry , for that but I noticed that , I have worked on that ..

Thanks. Could you please reply directly to the individual review comments, indicating whether each suggestion was incorporated or intentionally not adopted? That makes it easier for reviewers to verify the resolution history.

Also, would it make sense to add a regression test for the symlink escape scenario that motivated this fix? It would help ensure the path traversal issue doesn't get reintroduced in the future.

shteypandey28-hue:
Sure, let me check and reply to the review comments. I'll also add a test to verify the symlink escape scenario.

server/handlers/error.go
```diff
@@ -794,6 +795,10 @@ func ErrServeSchema(err error) error {
 func ErrInvalidFileRequest(err error) error {
 	return errors.New(ErrInvalidFileRequestCode, errors.Alert, []string{"Invalid file request"}, []string{err.Error()}, []string{"The provided file query parameter could not be decoded"}, []string{"Ensure the file parameter is a properly URL-encoded path"})
 }
+
+func ErrUnsafeFilePath(err error) error {
+	return errors.New(ErrUnsafeFilePathCode, errors.Alert, []string{"Unsafe file path requested"}, []string{err.Error()}, []string{"The requested file path is outside allowed directories or contains path traversal attempts"}, []string{"Ensure the file parameter points to an allowed directory (e.g. ~/.meshery) and does not contain path traversal elements"})
```
Omkar-Ugal: use go fmt here

server/handlers/utils_test.go
```diff
@@ -75,3 +77,77 @@ func TestHandlerWrappers_ForwardToHttputil(t *testing.T) {
...
+		},
+		{
+			name:        "Traversal attempting to escape meshery logs",
+			inputPath:   filepath.Join(mesheryLogs, "..", "..", "..", "etc", "passwd"),
+			expectError: true,
+		},
+	}
+
+	symlinkTestPath := filepath.Join(os.TempDir(), "meshery_test_symlink")
+	os.Remove(symlinkTestPath)
```
Omkar-Ugal: Resolve the lint issues here.

server/handlers/utils_test.go
```diff
@@ -75,3 +77,77 @@ func TestHandlerWrappers_ForwardToHttputil(t *testing.T) {
...
+	home, err := os.UserHomeDir()
+	if err != nil {
+		t.Fatalf("Failed to get user home dir: %v", err)
+	}
+	mesheryLogs := filepath.Join(home, ".meshery", "logs")
+	if err := os.MkdirAll(mesheryLogs, 0755); err != nil {
+		t.Fatalf("Failed to create meshery logs dir: %v", err)
+	}
+	// We don't remove it because it might be the user's actual .meshery dir.
+
```
Omkar-Ugal, line 99: Add your thoughts on the bot comments

server/handlers/error.go
shteypandey28-hue: I ran go fmt and removed the extraneous blank line in server/handlers/error.go between the functions.

server/handlers/utils_test.go
shteypandey28-hue: I updated the os.Remove(symlinkTestPath) calls in server/handlers/utils_test.go to safely ignore the returned errors by assigning them to the blank identifier (_ = os.Remove...). This satisfies the errcheck linter.

server/handlers/utils_test.go
shteypandey28-hue: Thanks for the thorough review! Yes, I have incorporated all the feedback:

SafeFilePath is now fully implemented to securely validate paths.
Absolute paths are securely restricted using filepath.EvalSymlinks and filepath.Rel to mathematically ensure containment (averting CWE-59 bypasses).
Relative path traversals (..) are explicitly blocked.
We now properly return the newly defined ErrUnsafeFilePath with a 400 Bad Request.
The regression test for the symlink escape has also been added. I've pushed all these updates to the branch!

server/handlers/common_handlers.go
```diff
@@ -71,6 +71,12 @@ func (h *Handler) ViewHandler(responseWriter http.ResponseWriter, request *http.
 		writeMeshkitError(responseWriter, ErrInvalidFileRequest(err), http.StatusBadRequest)
 		return
 	}
+	filePath, err = SafeFilePath(filePath)
```
Sbragul26: One additional thought, would it be useful to log blocked `SafeFilePath` requests as well? That could help track traversal attempts instead of only returning an error.

Sbragul26 (changes requested):
Since these routes are still NoAuth, should access be restricted to the required log directory, or protected with authentication?

server/handlers/error.go
```diff
@@ -210,6 +210,7 @@ const (
 	ErrExtensionProxyCode                  = "meshery-server-1427"
 	ErrInitializeMachineCode               = "meshery-server-1428"
 	ErrSendMachineEventCode                = "meshery-server-1429"
+	ErrUnsafeFilePathCode                  = "meshery-server-1430"
```
jamieplu: **Blocker (merge conflict + duplicate error code).** `ErrUnsafeFilePathCode = "meshery-server-1430"` collides with `ErrTelemetryGrafanaCode = "meshery-server-1430"` on `master` (added after this branch was cut). `errorutil` drops **both** colliding entries from `errorutil_errors_export.json` and the `error-codes-updater` CI gate fails on the resulting stdout. This PR's own CI is green only because the workflow checks out the head SHA (pre-telemetry base); the failure surfaces on the first push to `master` after merge.

The first free code repo-wide is **1443** (`server/helpers/component_info.json` has `next_error_code: 1443`; 1432 and 1436-1442 are already used in `server/models` and `server/models/pattern/stages`). Renumber to 1443, bump `component_info.json` to 1444, and run `make error`.

```suggestion
	ErrUnsafeFilePathCode                  = "meshery-server-1443"
```

server/handlers/utils.go
```diff
@@ -95,3 +99,66 @@ func extractBoolQueryParams(r *http.Request, params ...string) (map[string]bool,
...
+
+	// Allowed base directories
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("could not determine user home directory")
+	}
+
+	allowedBaseDirs := []string{
+		filepath.Join(home, ".meshery"),
+		os.TempDir(),
```
jamieplu: **Security: the fix only narrows the arbitrary read, it does not close it (#20076).** Both `/api/system/fileView` and `/api/system/fileDownload` are registered `models.NoAuth`, so in local/self-hosted mode they are unauthenticated. Allowing the entire `~/.meshery` tree still exposes `~/.meshery/config/mesherydb.sql` (the full database, including session tokens) and provider credentials; allowing all of `os.TempDir()` exposes every other process's files under world-writable `/tmp` on Linux.

Every legitimate producer of these links points only at `~/.meshery/logs/registry/` (`seed_models_logs.go`, `common_registry_logs.go`, `component_generator_helper.go`). Narrow the allowlist to `~/.meshery/logs` and drop `os.TempDir()` (no producer emits a temp-dir link).

server/handlers/utils.go
```diff
@@ -95,3 +99,66 @@ func extractBoolQueryParams(r *http.Request, params ...string) (map[string]bool,
...
+			isAllowed = true
+			break
+		}
+	}
+
+	if !isAllowed {
+		return "", fmt.Errorf("file path is outside allowed directories")
+	}
+
+	return resolvedPath, nil
```
jamieplu: **TOCTOU: returning a validated path string re-opens the race the PR is meant to close.** `SafeFilePath` returns a string that the handlers then feed to `os.Open` (`common_handlers.go:80`, `:118`), which re-resolves symlinks. Between `EvalSymlinks` here and `os.Open` there, a component can be swapped for a symlink and the open escapes the allowlist. The not-exists branch (line 126) compounds this: it never symlink-resolves the final component, so a dangling symlink is returned as "safe" and followed once its target appears.

`go.mod` is `go 1.26.4`, so `os.Root`/`os.OpenRoot` is available: open the file *within* a root rooted at the allowed dir (`root.Open(rel)`), which is traversal-proof, symlink-escape-proof, and TOCTOU-free in one operation. Have the helper return the `*os.File` so the handler no longer re-opens.

server/handlers/utils.go
```diff
@@ -95,3 +99,66 @@ func extractBoolQueryParams(r *http.Request, params ...string) (map[string]bool,
...
+		dir := filepath.Dir(absPath)
+		resolvedDir, err := filepath.EvalSymlinks(dir)
+		if err != nil {
+			return "", fmt.Errorf("failed to resolve symlinks for directory: %v", err)
+		}
+		resolvedPath = filepath.Join(resolvedDir, filepath.Base(absPath))
+	}
+
+	// Allowed base directories
+	home, err := os.UserHomeDir()
```
jamieplu: **Regression: allowlist home source disagrees with the log-writer's home source.** The allowlist anchors on `os.UserHomeDir()` (`$HOME`), but `model-generation.log` is written and its `ViewLink` minted from `meshkitutils.GetHome()` = `user.Current().HomeDir` (the passwd entry) at `component_generator_helper.go:388`. When `$HOME` differs from the passwd home (systemd unit, `sudo`, `docker run -e HOME=...`), the server rejects a "Get Logs" link it generated itself with 400. Include both home derivations in the allowed set.

server/handlers/utils.go
```diff
@@ -95,3 +99,66 @@ func extractBoolQueryParams(r *http.Request, params ...string) (map[string]bool,
...
+	if err != nil {
```
jamieplu: **Regression: a home-lookup failure disables the entire allowlist, including temp/log paths that don't depend on `$HOME`.** If `os.UserHomeDir()` errors (e.g. `$HOME` unset under a stripped-env supervisor), the function returns before any other allowed base is considered, so every `fileView`/`fileDownload` request 400s. A missing home should only drop the home-derived entry, not fail the whole validation.

server/handlers/utils.go
```diff
@@ -95,3 +99,66 @@ func extractBoolQueryParams(r *http.Request, params ...string) (map[string]bool,
...
+		return "", fmt.Errorf("path traversal attempts are not allowed")
+	}
+
+	absPath, err := filepath.Abs(providedPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to get absolute path: %v", err)
+	}
+
+	var resolvedPath string
+	if _, err := os.Stat(absPath); err == nil {
```
jamieplu: **Wrong error classification for missing files.** Any `os.Stat` failure (a not-yet-created `~/.meshery/logs/registry` dir, `EACCES`, `ELOOP`) falls into the else branch, and `EvalSymlinks(dir)` on a missing parent then makes the handler return **400 "Unsafe file path requested ... path traversal attempts"** for a perfectly benign request. Pre-PR that was a 500 `ErrReadFileContent`. Support triage now chases a phantom attack, and clients that branch on 4xx-vs-5xx misclassify a server-state problem. Also gate on `os.IsNotExist(err)` rather than treating every stat error as "not found".

server/handlers/utils.go
```diff
@@ -95,3 +99,66 @@ func extractBoolQueryParams(r *http.Request, params ...string) (map[string]bool,
...
 }
+
+// SafeFilePath validates a file path to prevent path traversal attacks.
+// It ensures the path is within allowed directories (e.g., ~/.meshery or os.TempDir()).
+func SafeFilePath(providedPath string) (string, error) {
+	if providedPath == "" {
+		return "", fmt.Errorf("file path cannot be empty")
+	}
+
+	if strings.Contains(providedPath, "..") {
```
jamieplu: **False-positive rejection + redundant guard.** `strings.Contains(providedPath, "..")` rejects any path containing `..` as a substring, including legitimate rotated logs like `~/.meshery/logs/registry/app..2026-07-16.log`. It adds no security: `filepath.Abs` (which `Clean`s) followed by the `filepath.Rel` containment check already rejects every real traversal. Remove it and rely on the canonical containment check (or `os.Root`).

server/handlers/utils.go
```diff
@@ -95,3 +99,66 @@ func extractBoolQueryParams(r *http.Request, params ...string) (map[string]bool,
...
+
+	isAllowed := false
+	for _, baseDir := range allowedBaseDirs {
+		resolvedBase, err := filepath.EvalSymlinks(baseDir)
+		if err != nil {
+			resolvedBase, _ = filepath.Abs(baseDir)
+		}
+
+		rel, err := filepath.Rel(resolvedBase, resolvedPath)
+		if err == nil && !strings.HasPrefix(rel, "..") {
```
jamieplu: **Containment boundary is imprecise.** `!strings.HasPrefix(rel, "..")` also rejects a legitimate file literally named `..foo` directly inside an allowed dir, and (in the general case) treats `rel == ".."` the same as a real escape only by luck of the substring. The precise test is `rel == "." || (rel != ".." && !strings.HasPrefix(rel, ".."+string(os.PathSeparator)))`. This becomes load-bearing once the redundant `Contains("..")` guard above is removed. `os.Root` sidesteps the manual check entirely.

server/handlers/utils.go
```diff
@@ -95,3 +99,66 @@ func extractBoolQueryParams(r *http.Request, params ...string) (map[string]bool,
...
+	allowedBaseDirs := []string{
+		filepath.Join(home, ".meshery"),
+		os.TempDir(),
+	}
+
+	isAllowed := false
+	for _, baseDir := range allowedBaseDirs {
+		resolvedBase, err := filepath.EvalSymlinks(baseDir)
+		if err != nil {
+			resolvedBase, _ = filepath.Abs(baseDir)
```
jamieplu: **Swallowed error + mixed canonicalization.** `resolvedBase, _ = filepath.Abs(baseDir)` discards the error inside a fallback that already absorbed the `EvalSymlinks` failure, then compares an *unresolved* base against a *symlink-resolved* candidate (line 121). On a host where a base sits behind a symlink (macOS `/var` -> `/private/var`), a legitimate request 400s with no log trace. Per the repo error-handling convention, no error path should be silently swallowed.

server/handlers/common_handlers.go
```diff
@@ -103,6 +109,12 @@ func (h *Handler) DownloadHandler(responseWriter http.ResponseWriter, request *h
 		return
 	}
 
+	filePath, err = SafeFilePath(filePath)
```
jamieplu: **Cleanup + a filename regression.** This 5-line validate-and-return block is duplicated verbatim in `ViewHandler` (74-78) and `DownloadHandler` (112-116), on two handlers that already duplicate `QueryUnescape`/`os.Open`/`defer-close`/`io.Copy`. Extract a single `serveFile(w, r, asAttachment)` helper both call, so the security check lives in one place.

Separately: because `SafeFilePath` reassigns `filePath` to the **symlink-resolved** path, `filepath.Base(filePath)` at line 129 now derives the `Content-Disposition` name from the resolved target, so a rotated/symlinked log downloads under the wrong name. Derive the attachment name from the originally requested path.

server/handlers/utils_test.go
```diff
@@ -75,3 +77,77 @@ func TestHandlerWrappers_ForwardToHttputil(t *testing.T) {
...
 	})
 }
+
+func TestSafeFilePath(t *testing.T) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		t.Fatalf("Failed to get user home dir: %v", err)
+	}
+	mesheryLogs := filepath.Join(home, ".meshery", "logs")
+	if err := os.MkdirAll(mesheryLogs, 0755); err != nil {
```
jamieplu: **Test mutates the real home directory and never cleans up** (the comment at line 90 admits it). `go test ./server/handlers/` permanently creates `~/.meshery/logs` on a contributor/CI machine, and fails outright at `t.Fatalf` where `$HOME` is read-only. Root cause is that the allowed dirs are hardcoded inside the unit under test. Redirect the home with `t.Setenv("HOME", t.TempDir())` (which `os.UserHomeDir` honors on unix/darwin) so the allowed root is hermetic and auto-cleaned.

server/handlers/utils_test.go
```diff
@@ -75,3 +77,77 @@ func TestHandlerWrappers_ForwardToHttputil(t *testing.T) {
...
+		}{
+			name:        "Symlink escape attempt",
+			inputPath:   filepath.Join(symlinkTestPath, "passwd"),
+			expectError: true,
+		})
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := SafeFilePath(tt.inputPath)
```
jamieplu: **Assertion only checks the error bit, never the returned path.** Mutating the last line of `SafeFilePath` to `return "", nil` (or returning the unresolved path) keeps every case green while the handlers `os.Open("")` and 500 on every request. Assert the returned value: for a success case, that the opened file's contents match what was written.

server/handlers/utils_test.go
```diff
@@ -75,3 +77,77 @@ func TestHandlerWrappers_ForwardToHttputil(t *testing.T) {
...
+	// We don't remove it because it might be the user's actual .meshery dir.
+
+	tests := []struct {
+		name        string
+		inputPath   string
+		expectError bool
+	}{
+		{
+			name:        "Valid internal absolute path",
+			inputPath:   filepath.Join(mesheryLogs, "registry.log"),
```
jamieplu: **The one branch production always takes has no success coverage.** `registry.log` and `tempfile.txt` are never created, so both expect-success cases exercise only the *not-exists* branch; the `os.Stat`-exists + `EvalSymlinks` branch (which every real `fileView`/`fileDownload` hits) is never asserted on a success path. Regress that branch and the suite stays green. Create a real file and assert it validates and is served.

server/handlers/utils_test.go
```diff
@@ -75,3 +77,77 @@ func TestHandlerWrappers_ForwardToHttputil(t *testing.T) {
...
+			expectError: true,
+		},
+		{
+			name:        "Empty path",
+			inputPath:   "",
+			expectError: true,
+		},
+		{
+			name:        "Traversal attempting to escape meshery logs",
+			inputPath:   filepath.Join(mesheryLogs, "..", "..", "..", "etc", "passwd"),
```
jamieplu: **Tautological case.** `filepath.Join(mesheryLogs, "..", "..", "..", "etc", "passwd")` is cleaned by `Join` to `<home>/../etc/passwd`, so `SafeFilePath` never sees a literal `..` and the case passes even if both the `Contains` guard and the entire allowlist loop are deleted. To actually exercise escape-from-inside-an-allowed-dir, build the input by string concatenation so the literal `..` survives on the wire, exactly as a malicious `?file=` param would. Add a case asserting `~/.meshery/mesherydb.sql` (outside `logs`) is rejected to lock in the narrowed scope.

server/handlers/utils_test.go
```diff
@@ -75,3 +77,77 @@ func TestHandlerWrappers_ForwardToHttputil(t *testing.T) {
...
+			expectError: true,
+		},
+	}
+
+	symlinkTestPath := filepath.Join(os.TempDir(), "meshery_test_symlink")
+	_ = os.Remove(symlinkTestPath)
+	symlinkErr := os.Symlink("/etc", symlinkTestPath)
+	if symlinkErr == nil {
```
jamieplu: **The core symlink-escape defense is silently skipped when `os.Symlink` fails** (Windows without privilege, sandboxed/read-only tmp) with no `t.Skip`/`t.Log`. The suite reports all-pass while the one case covering `EvalSymlinks`-based escape detection never ran, so a regression deleting that defense ships green from such a machine. Make the skip explicit.

jamieplu:
## Maintainer review + fixes applied (pushed to this branch)

I ran a full security review of the path-traversal fix and pushed fixes for the issues found. Summary below; inline comments mark each original finding.

### Blocker resolved
- `ErrUnsafeFilePathCode` was `meshery-server-1430`, which now collides with master's `ErrTelemetryGrafanaCode`. Left as-is it would turn every post-merge master CI red (errorutil drops both colliding entries and the error-codes gate fails). Renumbered to the next free code **1443** and merged master in to clear the conflict.

### Security (the core of #20076)
- The original allowlist (`~/.meshery` + `os.TempDir()`) only **narrowed** the arbitrary read. `~/.meshery` still exposed `mesherydb.sql` (DB + session tokens) and provider credentials, and all of world-writable `/tmp`, to the **unauthenticated** (`NoAuth`) endpoints. Narrowed to `~/.meshery/logs`, the only directory legitimate `ViewLink`/`DownloadLink` producers point at.
- Rewrote `SafeFilePath` as `SafeOpenFile` using **`os.Root`** (Go 1.26): the file is opened *inside* a root at the allowed dir, so `..` and symlink components cannot escape, and validation + open are one step - closing the previous validate-string-then-`os.Open` **TOCTOU** and the dangling-symlink gap.

### Correctness / regressions
- Allowed dir is now derived from **both** `os.UserHomeDir()` and meshkit `GetHome()`, so a `$HOME` that differs from the passwd home no longer rejects the server's own `model-generation.log` link.
- Dropped the redundant, over-broad `strings.Contains("..")` check (`filepath.Abs` already cleans `..`), so rotated logs like `app..2026.log` are no longer falsely rejected.
- A missing file now returns **500** (read error) instead of a misleading **400 "path traversal"**.
- Download filename is taken from the requested path, not the symlink-resolved target.

### Cleanup / tests
- Extracted a shared `serveFile` helper from the duplicated `ViewHandler`/`DownloadHandler` bodies.
- Replaced `TestSafeFilePath` with a hermetic `TestSafeOpenFile` (redirects `HOME` to a temp dir - no longer writes to the real `~/.meshery`), asserting returned content, the narrowed scope (`mesherydb.sql` rejected), a `..`-in-name file allowed, and symlink escape.

Local `go build ./server/...`, `golangci-lint`, and `go test ./server/handlers/...` all pass.

jamieplu (approved):
Approving after a full maintainer review and the fixes pushed in `0845cb41f43`.

@Sbragul26's change request ("since these routes are still NoAuth, should access be restricted to the required log directory?") is addressed: the allowlist is now confined to `~/.meshery/logs` - the only directory the log-link producers use - so the unauthenticated endpoints can no longer read `mesherydb.sql`, provider credentials, or `/tmp`. `SafeFilePath` was also rebuilt on `os.Root`, closing the validate-then-reopen TOCTOU and symlink-escape gaps.

All required checks are green (server unit, error-codes, golangci server/cli, integration docker/kubernetes/meshsync, e2e, builds).

Issue #20137, opened 2026-06-19
[Server] Verify connection Kind is "kubernetes" before updating MeshSync deployment mode


emphor11:
Current Behavior
In the REST handler for connection updates (server/handlers/connections_handlers.go), the server parses the payload and checks metadata for MeshSync deployment mode changes. However, it executes this logic for all connection types, even though MeshSync is a Kubernetes-only component.

// TODO: also check that kind = "kubernetes" (when client starts to send full connection object)
if connections.MeshsyncDeploymentModeFromMetadata(connection.MetaData) != connections.MeshsyncDeploymentModeUndefined {

Expected Behavior
The server should only attempt to process MeshSync deployment mode updates if the connection's Kind is explicitly "kubernetes".

### Environment

- **Host OS:** Mac Linux Windows
- **Platform:** Docker or Kubernetes
- **Meshery Server Version:** stable-v
- **Meshery Client Version:** stable-v

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)

emphor11:
I am working on this @rishiraj38 @alexquincy

KatalKavya96:
Duplicate of #19396

PR #20138, opened 2026-06-19
[Server] Verify connection kind before updating MeshSync deployment mode

emphor11:
### Notes for Reviewers
This PR fixes [Server] Verify connection Kind is "kubernetes" before updating MeshSync deployment mode #20137

### Summary
MeshSync is a Kubernetes-only component, but the connection update handler (`UpdateConnectionById`) was processing MeshSync deployment mode changes for all connection kinds whenever deployment mode metadata was present in the payload.

### Changes
Added a Kind guard to the existing condition so MeshSync handling is skipped when the client explicitly sends a non-kubernetes Kind.

When `Kind` is omitted from the payload (which is the current UI behavior — RTK Query only sends `{ status, metadata }`), the condition allows entry and `handleMeshSyncDeploymentModeChange` verifies the persisted connection kind internally via its existing `GetConnectionByID` + `Kind != "kubernetes"` guard.

**Before:**
```go
// TODO: also check that kind = "kubernetes" (when client starts to send full connection object)
if connections.MeshsyncDeploymentModeFromMetadata(connection.MetaData) != connections.MeshsyncDeploymentModeUndefined {

### DIFF
--- a/server/handlers/connections_handlers.go
+++ b/server/handlers/connections_handlers.go
@@ -378,12 +378,11 @@ func (h *Handler) UpdateConnectionById(w http.ResponseWriter, req *http.Request,
 		return
 	}
 
-	// Check if meshsync deployment mode is specified in the payload metadata.
-	// Only handle mode changes if a mode is explicitly provided (not undefined).
-	// In fact this method is used (for now) only for perform meshsync deployment mode change.
-	// If mode change fails return error.
-	// TODO: also check that kind = "kubernetes" (when client starts to send full connection object)
-	if connections.MeshsyncDeploymentModeFromMetadata(connection.MetaData) != connections.MeshsyncDeploymentModeUndefined {
+	// MeshSync is a Kubernetes-only component. If the client sends Kind explicitly and it is
+	// not "kubernetes", skip this block. If Kind is omitted, handleMeshSyncDeploymentModeChange
+	// verifies the persisted connection kind internally.
+	if connections.MeshsyncDeploymentModeFromMetadata(connection.MetaData) != connections.MeshsyncDeploymentModeUndefined &&
+		(connection.Kind == "" || strings.EqualFold(connection.Kind, "kubernetes")) {
 		// Handle meshsync deployment mode changes before connection update
 		token, _ := req.Context().Value(models.TokenCtxKey).(string)
 		oldMode, newMode, modeChanged, err := h.handleMeshSyncDeploymentModeChange(

### CODE REVIEW

**Comment by emphor11** [comment]
@alexquincy @rishiraj38 @leecalcote

File: server/handlers/connections_handlers.go [hunk 1]
```diff
@@ -378,50 +378,66 @@ func (h *Handler) UpdateConnectionById(w http.ResponseWriter, req *http.Request,
...
+			token, _ := req.Context().Value(models.TokenCtxKey).(string)
+			var err error
+			var statusCode int
+			existingConn, statusCode, err = provider.GetConnectionByID(token, connectionID)
+			if err != nil {
+				err = fmt.Errorf("failed to retrieve existing connection: %w", err)
+				h.log.Error(err)
+				writeMeshkitError(w, err, statusCode)
+				return
+			}
```
KatalKavya96: @emphor11 This fallback makes sense for backward compatibility when clients omit `Kind`. Could we add a regression test for this path as well? Specifically: request payload has empty `Kind`, existing persisted connection is Kubernetes, and MeshSync deployment mode handling still proceeds.

server/handlers/connections_handlers.go
```diff
@@ -378,50 +378,66 @@ func (h *Handler) UpdateConnectionById(w http.ResponseWriter, req *http.Request,
...
+				userID,
+				provider,
+				existingConn,
+			)
+			if err != nil {
+				meshSyncErr := fmt.Errorf("error handling meshsync deployment mode change: %w", err)
+				metadata := map[string]any{
+					"error":        meshSyncErr,
+					"connectionId": connectionID,
+				}
```
KatalKavya96: @emphor11  Should the non kubernetes case be silently skipped here or should it return a 400 Bad Request when meshsync_deployment_mode metadata is present? Silently skipping the handler may still allow MeshSync specific metadata to be persisted on a non kubernetes connection which could leave the connection state confusing

server/handlers/connections_handlers.go
```diff
@@ -596,6 +612,7 @@ func (h *Handler) handleMeshSyncDeploymentModeChange(
 	token string,
 	userID core.Uuid,
 	provider models.Provider,
+	existingConnection *connections.Connection,
```
KatalKavya96: @emphor11 Could we also cover this with a test to ensure the helper behaves correctly when existingConnection is provided vs when it is nil?

KatalKavya96 (commented):
@emphor11 The implementation direction looks good to me, The guard is inside the MeshSync metadata conditional it uses case insensitive comparison  and the fallback lookup handles clients that omit Kind.

Main remaining concerns are around expected behaviour and tests should non-kubernetes + MeshSync metadata be rejected instead of silently skipped and can we add regression coverage for both the fallback kind lookup path and the non-kubernetes path?

winkletinkle:
This behavior is not desired. Perhaps, @lekaf974 can explain why.

server/handlers/connections_handlers.go
```diff
@@ -378,12 +378,11 @@ func (h *Handler) UpdateConnectionById(w http.ResponseWriter, req *http.Request,
...
-	// Only handle mode changes if a mode is explicitly provided (not undefined).
-	// In fact this method is used (for now) only for perform meshsync deployment mode change.
-	// If mode change fails return error.
-	// TODO: also check that kind = "kubernetes" (when client starts to send full connection object)
-	if connections.MeshsyncDeploymentModeFromMetadata(connection.MetaData) != connections.MeshsyncDeploymentModeUndefined {
+	// MeshSync is a Kubernetes-only component. If the client sends Kind explicitly and it is
+	// not "kubernetes", skip this block. If Kind is omitted, handleMeshSyncDeploymentModeChange
+	// verifies the persisted connection kind internally.
+	if connections.MeshsyncDeploymentModeFromMetadata(connection.MetaData) != connections.MeshsyncDeploymentModeUndefined &&
+		(connection.Kind == "" || strings.EqualFold(connection.Kind, "kubernetes")) {
```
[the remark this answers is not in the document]
emphor11: The non-canonical Kind persistence is a pre-existing concern that exists with or without this guard — UpdateConnectionById at line 439 persists the payload regardless of MeshSync handling. Input normalization should be addressed at the payload/provider level, not in this handler's guard. This PR's scope is limited to #20137: skip MeshSync handling for non-kubernetes connections. strings.EqualFold is appropriate for a guard condition.

emphor11:
@leecalcote @alexquincy

Sbragul26 (commented):
Would it be worth adding a small test for a non-kubernetes Kind (and maybe an empty Kind) to help prevent regressions in the future?

Also, it might be good to get clarification on the "not desired" comment above, since that could affect the overall approach.

emphor11:
@Sbragul26 While tracing the flow, I noticed that handleMeshSyncDeploymentModeChange() already retrieves the persisted connection and validates that it's Kubernetes, and any error aborts the request. I also traced the TODO with git blame; it predates the later refactors where this validation was added to the helper. So I'd like to get clarity on the intended ownership before adding tests around the current behavior.

jamieplu:
> @Sbragul26 While tracing the flow, I noticed that handleMeshSyncDeploymentModeChange() already retrieves the persisted connection and validates that it's Kubernetes, and any error aborts the request. I also traced the TODO with git blame; it predates the later refactors where this validation was added to the helper. So I'd like to get clarity on the intended ownership before adding tests around the current behavior.

In other words, this PR is not necessary.

PR #20158, opened 2026-06-20
chore(make): standardize Makefile structure

Sbragul26:
**Notes for Reviewers**

This PR makes the Makefiles consistent across [docs] repositories. It aligns the structure, targets, and commands used in the Makefiles.

**[Signed commits](../blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**


.github/build/Makefile.core.mk
aclfe, line 1: Maybe I'm missing context, install/Makefile.core.mk is already a file. Why do we need a new build/Makefile.core.mk?

Sbragul26: Nice catch. I only checked the `.github/build` directory since that's how it is used in the academy sites and docs repos. I missed that `Makefile.core.mk` already exists in `/install`. Thanks for pointing it out.

docs/Makefile
```diff
@@ -53,9 +55,15 @@ clean: check-deps
...
+## Verify Go is installed locally.
 check-go:
-	@$(MAKE) -C .. check-go
+	@echo "Checking if Go is installed..."
+	@command -v go > /dev/null || (echo "Go is not installed. Please install it before proceeding."; exit 1)
+	@echo "Go is installed."
+
+## Format code using Prettier
+format:
+	npm run format
```
Bhumikagarggg, line 66: The new format target in docs/Makefile calls npm run format, but docs/package.json does not define a format script. Please add the matching npm script or update/remove this target, otherwise make format
  will fail.

Bhumikagarggg (changes requested):

saurabhraghuvanshii (approved):

docs/Makefile
pontusringblom: Consider each piece of code review feedback offered. Please respond to each and every comment speaking to what changes have been made to incorporate the feedback or justifying why you have reject the feedback with no change.

saurabhraghuvanshii (approved):

banana-three-join:
For reference, I'm also working on the same issue over in the academy repos. Here's the PR: https://github.com/layer5io/exoscale-academy/pull/425/

Maybe we can agree on the base recipes alongside with the overall structure of the makefile so we can just copy and paste the work across all of the other makefiles leaving only the work for any specific rules that are needed per repo

banana-three-join:
Taken care of in https://github.com/meshery/meshery/pull/21251

PR #20180, opened 2026-06-21
[Chore] Bump @meshery/schemas to 1.3.13 in UI packages (ui, provider-ui, docker-extension)

ritzorama:
## Description

Brings the **UI / npm** side of meshery onto `@meshery/schemas` **v1.3.13** - the version the
**Go** side already pins (`go.mod`, landed via #20177) and that `meshery-extensions`,
`meshery-cloud`, and `sistent` now consume. This closes the gap where meshery's frontend lagged
its own backend on the schemas contract, keeping the generated **RTK Query clients** and **TS
interfaces** in lockstep across the ecosystem.

schemas v1.3.13 is the release that standardizes the ecosystem on `github.com/gofrs/uuid`
(reversing the brief `google/uuid` convergence that panicked `gobuffalo/pop`). The substance is
Go-side; for TypeScript a `format: uuid` field is `string`, so this is a routine keep-current
bump of the UI dependency.

| Package | before | after |
|---|---|---|
| `ui` | `^1.3.9` | `^1.3.13` |
| `provider-ui` | `^1.3.7` | `^1.3.13` |
| `install/docker-extension/ui` | `^1.3.7` | `^1.3.13` |

The `docker-extension/ui` lockfile additionally **prunes a transitive `@reduxjs/toolkit`
subtree**: schemas 1.3.13 declares RTK as a `peerDependency` (it was previously a bundled
`dependency`), so consumers that do not use RTK directly no longer carry it. `docker-extension/ui`
neither declares nor imports RTK, so the prune is correct cleanup, not a regression. (`ui` already
declares `@reduxjs/toolkit ^2.12.0` directly, so the peer requirement is satisfied there.)

## Test plan

- [x] **`ui`** - the only TypeScript consumer (19 schemas-importing files): `tsc --noEmit
      --skipLibCheck` passes with **0 errors** at 1.3.13. The RTK clients and `v1beta*/core`
      interfaces all type-check.
- [x] **`provider-ui`** - JS app, no `tsconfig`; **0** source references to `@meshery/schemas`.
      Bump is a no-op for its code.
- [x] **`docker-extension/ui`** - **0** schemas source usage (the only match is a pre-built,
      committed bundle map); pruned RTK transitives are neither declared nor imported.
- [x] All three: `npm install @meshery/schemas@^1.3.13 --legacy-peer-deps` (repo `.npmrc`
      convention) - package.json, lockfile, and `node_modules` resolve to 1.3.13.
- [ ] CI (lint / build) - runs on this PR.

## Related

- #20177 (merged) - Go gofrs/uuid restore; pinned schemas v1.3.13 in `go.mod`.
- #19633 - Go-only schemas bump (does not touch the UI packages this PR covers).

### DIFF
--- a/install/docker-extension/ui/package-lock.json
+++ b/install/docker-extension/ui/package-lock.json
@@ -14,7 +14,7 @@
         "@fortawesome/fontawesome-svg-core": "^7.1.0",
         "@fortawesome/free-solid-svg-icons": "^7.1.0",
         "@fortawesome/react-fontawesome": "^3.2.0",
-        "@meshery/schemas": "^1.3.7",
+        "@meshery/schemas": "^1.3.13",
         "@mui/icons-material": "^7.3.7",
         "@mui/material": "^7.3.7",
         "@mui/styled-engine-sc": "^7.3.7",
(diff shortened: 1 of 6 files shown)


ritzorama:
The failing **check-install** job is a pre-existing CI bug, not a problem with this dependency bump. The `Install Artifacts Drift Check` workflow runs `make check-install`, which doesn't exist (the Makefile target is `providers-check`), so it errored with `No rule to make target 'check-install'` *before* ever running the drift check — silently masking a real drift in `install/deployment_yamls/k8s/meshery-deployment.yaml`.

Fixed in #20182 (corrects the workflow target name + resyncs the one drifted manifest via `make providers-propagate`). Once #20182 lands and this branch is updated onto master, this PR's `check-install` will pass.

This bump only touches `install/docker-extension/ui` package metadata (npm deps), which legitimately tripped the now-functional gate but does not itself affect provider artifacts.

PR #20185, opened 2026-06-22
[mesheryctl] Fix missing HTTP timeout in config version check

khanskr:
**Notes for Reviewers**

This PR fixes a missing HTTP timeout in `ValidateVersion()` in `mesheryctl/internal/cli/root/config/config.go`. The function validates a Meshery release version by making a GET request to GitHub's releases API, but the underlying `http.Client{}` had no timeout, causing `mesheryctl version` and other commands to hang indefinitely when GitHub is unreachable.

Changes:
- Added `versionCheckTimeout = 10 * time.Second` constant
- Added `Timeout: versionCheckTimeout` to the HTTP client
- Added tests for the two early-return paths (empty version and "latest" version)

### DIFF
--- a/mesheryctl/internal/cli/root/config/config_test.go
+++ b/mesheryctl/internal/cli/root/config/config_test.go
@@ -157,18 +157,30 @@ func TestGetCurrentContextName(t *testing.T) {
 	}
 }
 func TestSetContext(t *testing.T) {
-	for _, test := range tests {
-		mesheryctlconfig := MesheryCtlConfig{nil, test, nil}
-		err := UpdateContextInConfig(nil, test)
-		if err != nil {
-			fmt.Print("Fail") //Internal:need to be fixed
-		}
-		got := mesheryctlconfig.GetCurrentContextName()
-		want := test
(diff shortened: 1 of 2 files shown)


mesheryctl/internal/cli/root/config/config_test.go
```diff
@@ -460,3 +460,22 @@ func TestSetOperatorStatus(t *testing.T) {
 // 		})
 // 	}
 // }
+
+func TestValidateVersion(t *testing.T) {
```
Sbragul26, line 480: Current tests only cover early returns and don’t reach client. `Do(req)` , so the timeout change is not tested.
Please add a `httptest.NewServer` test with a slow response (>10s) to verify `ValidateVersion()` timeout behavior.

khanskr: Added validateVersionURL var (injectable in tests) + httptest timeout test.

khanskr:
@Sbragul26 hi! if you need any changes then please comment-I will do them. Thanks!

PR #20186, opened 2026-06-22
[UI] Remove K8s context chip on connection delete (#5078)

### LINKED ISSUE #5078: [UI] Kubernetes contexts deleted, but not shown as deleted in UI.
**Kubernetes contexts deleted, but not shown as deleted in UI.**

Scenario: User connects to Kubernetes cluster. User visits Settings page. User clicks the "x" on the Kubernetes chip. The Kubernetes context is deleted with the Meshery Server output:

```
INFO[0436] retrieved user: {layer5   http://pbs.twimg.com/profile_images/1190059013544779776/mxxU_z_S_normal.jpg}
INFO[0436] attempting to fetch current kubernetes contexts from cloud
DEBU[0436] constructed kubernetes contexts url: https://meshery.layer5.io/user/contexts/e446a1ec-03a8-468e-b960-3b3f48102522/current
INFO[0436] kubernetes context successfully retrieved from remote provider
INFO[0436] attempting to delete kubernetes context from cloud for id: 9b3b4b515d256a0c000e85110a139ada
DEBU[0436] constructed kubernetes contexts url: https://meshery.layer5.io/user/contexts/e446a1ec-03a8-468e-b960-3b3f48102522/context/9b3b4b515d256a0c000e85110a139ada
INFO[0436] kubernetes successfully deleted from remote provider
```

But, chips remain in the UI:
Screen Shot 2022-02-15 at 5 03 00 PM 2

**Desired Behaviour**
The chip should get removed from the UI

_Originally posted by @leecalcote in https://github.com/meshery/meshery/pull/5076#issuecomment-1040872882_

udita-0707:
**Notes for Reviewers**

- This PR fixes #5078 (partial — immediate chip removal after successful delete).
- Header K8s chip delete now uses `PUT /api/integrations/connections/{connectionId}` via `updateConnectionById` instead of the bulk status endpoint that returns 405 on Local provider.
- On successful delete, `removeK8sContextByConnectionId` in `_app.tsx` optimistically removes the chip from app state.
- RTK `updateConnectionById` includes connection `id` in the PUT body.
- Adds `k8sContextMatchesConnectionId` helper (`ui/utils/k8sContext.ts`) and unit tests.

https://github.com/user-attachments/assets/b4ffe6b7-8732-455e-b5de-789edc7f481e

**Known follow-up (out of scope):**
- Chip may reappear after hard refresh because the server still serves the K8s context after connection delete (GraphQL/subscription refetch). Server-side context cleanup / `PublishContext` on delete is a separate change.

**Testing:**
- Delete minikube chip from header K8s dropdown → success toast, chip disappears immediately.
- `npm test -- --run utils/__tests__/k8sContext.test.ts components/layout/Header/Header.test.tsx`


ui/components/layout/Header/Header.tsx
```diff
@@ -430,6 +434,7 @@ const Header = ({
   activeContexts,
   setActiveContexts,
   searchContexts,
+  removeK8sContextByConnectionId,
 }) => {
   const { notify } = useNotification;
```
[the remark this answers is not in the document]
udita-0707: Fixed useNotification() invocation in Header delete handler

ui/components/layout/Header/Header.tsx
```diff
@@ -430,6 +434,7 @@ const Header = ({
...
   const { openModal } = useContext(WorkspaceModalContext) || {};
```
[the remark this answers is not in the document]
udita-0707: Fixed useNotification() invocation in Header delete handler

ui/components/layout/Header/Header.tsx
```diff
@@ -430,6 +434,7 @@ const Header = ({
   activeContexts,
   setActiveContexts,
   searchContexts,
+  removeK8sContextByConnectionId,
```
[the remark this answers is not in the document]
udita-0707: Fixed useNotification() invocation in Header delete handler

ui/pages/_app.tsx
```diff
@@ -369,6 +370,41 @@ const MesheryApp = ({ Component, pageProps, relayEnvironment, emotionCache }) =>
...
+            ...prevState.k8sContexts,
+            contexts: remainingContexts,
+            totalCount: remainingContexts.length,
+          },
+          activeK8sContexts,
+        };
+      });
+    },
+    [activeContextChangeCallback],
+  );
```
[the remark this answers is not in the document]
udita-0707: Refactored removeK8sContextByConnectionId to compute state outside setState and preserve select-all

ui/pages/_app.tsx
```diff
@@ -369,6 +370,41 @@ const MesheryApp = ({ Component, pageProps, relayEnvironment, emotionCache }) =>
...
+        if (removedIds.size === 0) {
+          return prevState;
+        }
+
+        const remainingContexts = contexts.filter((ctx) => !removedIds.has(ctx.id));
+        const activeK8sContexts = (prevState.activeK8sContexts || []).filter(
+          (id) => id !== 'all' && !removedIds.has(id),
+        );
+
+        activeContextChangeCallback(activeK8sContexts);
```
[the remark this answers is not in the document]
udita-0707: Refactored removeK8sContextByConnectionId to compute state outside setState and preserve select-all

udita-0707:
@ritzorama hey, the PR is up for review. PTAL Thanks!

ui/pages/_app.tsx
```diff
@@ -369,6 +370,43 @@ const MesheryApp = ({ Component, pageProps, relayEnvironment, emotionCache }) =>
     [activeContextChangeCallback, fetchKubernetesContexts],
   );
 
+  const removeK8sContextByConnectionId = useCallback(
```
aabidsofi19, line 327: this is bad . why are we using regex matching . this is should be driven by rtk query only

udita-0707: @aabidsofi19 
The header chip list reads from `_app` local state (fed by the GraphQL subscription), not directly from RTK cache, `updateConnectionById` invalidates connections only, and the existing `fetchSystemSync` path updates `k8sConfig` but not `state.k8sContexts`, which is why the chip was sticking after delete. 

`removeK8sContextByConnectionId` is an optimistic reconciliation at confirm time so the UI updates immediately and we can preserve active-context selection (including select-all) without a refetch resetting state.
The helper isn’t fuzzy regex matching, it normalizes between connection UUID and context id formats we already use when selection is keyed by `ctx.id` but delete receives `connectionId`. Happy to drop the hyphen fallback and match only on` ctx.connectionId === connectionId `if we can confirm that field is always populated, or follow up with a refetch-based flow once k8s context state is consolidated on RTK/subscription alone.

KhushamBansal: @udita-0707 @aabidsofi19 Would comparing the existing chip selections with the newly fetched data, and only updating the list to keep the ones that still exist, be a better fix than manually deleting?

leecalcote:
Hi @chaitanyamedidar, will you please put your 👀 on this PR?

ui/rtk-query/connection.ts
chaitanyamedidar, line 1: Defining `updateConnectionById` as a `PUT` mutation sending only `{ id, status, metadata }` overwrites `Name`, `Kind`, `Type`, and `SubType` to empty strings in the database because `DefaultLocalProvider.UpdateConnectionById` replaces the entire row via GORM `.Save()`.
If this mutation is intended for general PUT updates, ensure callers pass the full connection payload.

ui/components/layout/Header/Header.tsx
```diff
@@ -221,7 +224,19 @@ function K8sContextMenu({
       showInfoIcon: `Learn more about the [lifecycle of connections](https://docs.meshery.io/concepts/logical/connections) and what it means to delete a connection.`,
     });
     if (responseOfDeleteK8sCtx === 'CONFIRM') {
-      const successCallback = async () => {
+      try {
+        await updateConnectionById({
+          connectionId: connectionID,
+          body: { status: CONNECTION_STATES.DELETED },
+        }).unwrap();
```
chaitanyamedidar, line 229: Calling `updateConnectionById` (`PUT integrations/connections/{id}`) with `status: DELETED` fails to trigger active Kubernetes resource cleanup under local provider because `NotifySmOfConnectionStatusChange` queries `GetK8sContext` by connection UUID rather than MD5 hex ID. This skips operator undeployment and background routine teardown.
I'd recommend to switch this deletion action to call the dedicated endpoint `DELETE /api/system/kubernetes/contexts/{context.id}`. This triggers active state machine cleanup (undeploying cluster operators) and automatically updates the connection record status to `DELETED`.

ui/utils/k8sContext.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (!ctx || !connectionId) {
+    return false;
+  }
+
+  const normalizedConnectionId = connectionId.replace(/-/g, '');
+
+  return (
+    ctx.connectionId === connectionId ||
+    ctx.id === connectionId ||
+    ctx.id === normalizedConnectionId
```
chaitanyamedidar, line 23: Once RTK Query cache invalidation is configured for cluster deletions, manual frontend state filtering becomes unnecessary and this utility function can be removed entirely. If ID comparisons are still needed during transition, replace regex matching with direct string equality, `ctx.connectionId === connectionId`.

ui/components/layout/Header/Header.tsx
```diff
@@ -403,8 +407,9 @@ const Header = ({
   activeContexts,
   setActiveContexts,
   searchContexts,
+  removeK8sContextByConnectionId,
 }) => {
-  const { notify } = useNotification;
+  const { notify } = useNotification();
   const { openModal } = useContext(WorkspaceModalContext) || {};
   const theme = useTheme();
   const isSmallScreen = useMediaQuery(theme.breakpoints.up('md'));
```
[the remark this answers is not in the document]
udita-0707: it’s out of scope for #5078

ui/utils/k8sContext.ts
udita-0707: Agreed on both points. I’ll switch header chip delete to DELETE `/api/system/kubernetes/contexts/{ctx.id}` (using context hex id, not connection UUID) so state-machine cleanup runs correctly. I’ll revert the updateConnectionById partial-PUT change and drop manual `_app.tsx` state filtering once RTK invalidates `getKubernetesContexts` after delete.

PR #20204, opened 2026-06-23
fix(server): avoid hardcoded kube-system namespace when deriving KubernetesServerID

### LINKED ISSUE #20196: [bug][server] Hardcoded "kube-system" namespace in k8s_context.go breaks restricted RBAC / multi-tenant environments
### Current Behavior
The cluster ID generation and context initialization routine in the Meshery server hardcodes the `"kube-system"` namespace string literal when querying cluster components. In locked-down enterprise settings or multi-tenant Kubernetes environments, the deploying service account or user operates with namespace-scoped RBAC and lacks cluster-wide authorization to view the `kube-system` namespace. This results in an immediate authorization/forbidden error, causing cluster context initialization to fail entirely.

The breaking code patterns reside in `server/models/k8s_context.go` (around lines `:167` and `:453`), utilizing unconfigurable calls like:
`ctx.KubeClient.CoreV1().Namespaces().Get(context.TODO(), "kube-system", metav1.GetOptions{})`

### Expected Behavior
The target namespace used for evaluating or anchoring the cluster scope should be dynamic and configurable. It should look up the active service account namespace from standard in-cluster metadata paths (`/var/run/secrets/kubernetes.io/serviceaccount/namespace`) or fall back to an explicit configuration flag/environment variable (e.g., `KUBERNETES_TARGET_NAMESPACE`) before reverting to `kube-system` as a last resort.

### Screenshots/Logs
* **File Context:** `server/models/k8s_context.go`
* **Error Signature:** `namespaces "kube-system" is forbidden: User "system:serviceaccount:..." cannot get resource "namespaces" in API group "" in the namespace "kube-system"`

### Environment

- **Host OS:** Linux / Mac / Windows
- **Platform:** Kubernetes
- **Meshery Server Version:** edge / stable-v
- **Meshery Client Version:** edge / stable-v

### To Reproduce
1. Deploy Meshery into a dedicated target namespace (e.g., `meshery-tenant`) using a ServiceAccount restricted strictly to that namespace via a standard `Role` and `RoleBinding` (no cluster-wide `ClusterRoleBinding`).
2. Attempt to initialize or fetch the Kubernetes cluster context via the Meshery server.
3. Observe the server context initialization failure caused by the forbidden namespace API lookup against `kube-system`.

kritagyaio:
## Summary

The server currently hardcodes the `kube-system` namespace when deriving the cluster's `KubernetesServerID`.

In namespace-scoped or restricted RBAC environments, service accounts may not have permission to access `kube-system`, causing:

- `NewK8sContextWithServerID()` to fail cluster initialization.
- `AssignServerID()` to skip context assignment.

This results in cluster connection failures in multi-tenant deployments.

## Changes

Added a `getKubeNamespace()` helper that resolves the namespace using the following priority:

1. `KUBERNETES_TARGET_NAMESPACE` environment variable.
2. `/var/run/secrets/kubernetes.io/serviceaccount/namespace` (in-cluster service account namespace).
3. `"kube-system"` as a backward-compatible fallback.

Updated the following call sites to use the helper instead of the hardcoded literal:

- `NewK8sContextWithServerID()`
- `AssignServerID()`

## Verification

- [x] `go build ./server/models/`
- [x] `go vet ./server/models/`
- [x] `go test ./server/models/ -short`
- [x] `go build ./server/machines/kubernetes/`

## Usage

Explicit override:

```bash
export KUBERNETES_TARGET_NAMESPACE=my-namespace
```

When running inside a cluster, the service account namespace is automatically detected.

If neither is available, the existing behavior is preserved by falling back to `kube-system`.

## Signed commits


fitzergerald (commented):
Thanks for this, @Kritagya123611 !

fitzergerald:
@ishwar170695, would you like to have a look?

leecalcote (changes requested):
## Review — `getKubeNamespace()` helper

The merge conflict has been resolved and pushed. Three findings from a multi-angle review:

---

### 🔴 [CONFIRMED] Silent server-ID overwrite on existing deployments

**File:** `server/machines/kubernetes/discover.go:38`  
**Severity:** High — data mutation on upgrade, no guard

`AssignServerID` is called unconditionally (no check for an existing non-nil `KubernetesServerID`), and `provider.SaveK8sContext` persists the result immediately at line 51:

```go
// discover.go ~line 38
if err := kc.AssignServerID(handler); err != nil { ... }
// ...
provider.SaveK8sContext(...)  // persists the newly derived ID
```

`AssignServerID` now calls `getKubeNamespace()`, which returns `KUBERNETES_TARGET_NAMESPACE` when set. This means:

- An operator running an **existing** Meshery deployment sets `KUBERNETES_TARGET_NAMESPACE=custom-ns`
- On next context discovery, all contexts go through `AssignServerID` → derive a **new** server ID from `custom-ns`
- `SaveK8sContext` overwrites the stored ID → all previously connected cluster identities are silently replaced
- Any data tied to the old server ID (connections, events, references) becomes orphaned

**Fix:** Guard `AssignServerID` to skip assignment if `KubernetesServerID` is already set, or document that setting this env var on an existing deployment requires a manual context re-registration step.

---

### 🟡 [CONFIRMED] Missing documentation for new env var

**File:** `server/models/k8s_context.go` — `getKubeNamespace()` function  

`KUBERNETES_TARGET_NAMESPACE` is a user-visible configuration knob for namespace-scoped RBAC environments, which is exactly the audience that needs this feature. The PR description explains it in the summary but there is no corresponding docs update.

Per AGENTS.md: *"Include doc updates for user-visible API changes."*

This env var should be documented — at minimum in the configuration or Kubernetes environment guide under `docs/`.

---

### 🔵 [PLAUSIBLE] `saNamespaceFile` package global is a fragile test seam

**File:** `server/models/k8s_context.go:481`

```go
var saNamespaceFile = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
```

This is a test-seam variable (documented as "Exposed as a variable so tests can override it"). The current tests all correctly use `t.Cleanup` to restore it. However:

- The global is writable by any future test without a compile-time guard
- It cannot be isolated per goroutine, so `t.Parallel()` tests that mutate it will race

Consider making this an `internal/testutil`-injected path or using a functional option on the function instead of a bare global. Alternatively, keep the global but add a comment warning that tests **must** use `t.Cleanup` to restore it.

---

*Merge conflict resolved: dropped `ErrInvalidUUIDHandlerCode`/`ErrInvalidUUID` from `handlers/error.go` (removed upstream in master via PR #19234; callers use `models.ErrInvalidUUID` instead) and took master's `next_error_code: 1437` in `component_info.json`.*

ishwar170695:
> @ishwar170695, would you like to have a look?

Took a look after the latest updates. The concerns from Lee's review appear to be addressed, and I don't have any additional code-level comments.

leecalcote:
Is the cluster ID available outside of `kube-system` namespace?

kritagyaio:
> Is the cluster ID available outside of `kube-namespace`?

Yes, it is. Although KubernetesServerID is derived from a namespace UID, every Kubernetes namespace has a cluster-unique UID (metadata.uid). Any namespace on the same cluster produces the same cluster identity. Once derived, the cluster ID is stored globally in the kubernetes_resources table, exposed through REST (/api/system/sync returns clusterId) and GraphQL, and used to filter resources across all namespaces it is never scoped to its source namespace at query time

leecalcote:
This is good news. Updating Meshery's logic to take advantage of that fact is going to be quite helpful.

leecalcote:
**Silent server-ID overwrite on existing deployments in `server/machines/kubernetes/discover.go:38`**

`AssignServerID` is called unconditionally — no check for an existing non-nil `KubernetesServerID` — and `provider.SaveK8sContext` persists the result immediately afterward:

```go
// discover.go ~line 38
if err := kc.AssignServerID(handler); err != nil { ... }
// ...
provider.SaveK8sContext(...)  // persists the newly derived ID
```

`AssignServerID` now delegates to `getKubeNamespace()`, which returns `KUBERNETES_TARGET_NAMESPACE` when set. The result: an operator running an **existing** Meshery deployment who sets `KUBERNETES_TARGET_NAMESPACE=custom-ns` will, on next context discovery, have every context's `KubernetesServerID` silently re-derived from `custom-ns` and persisted — replacing the IDs that were stored at original registration and orphaning any data (connections, events, references) tied to the old IDs.

**Suggested fix:** Guard the assignment in `discover.go` — skip `AssignServerID` (or at least the save) when `KubernetesServerID` is already non-empty. Alternatively, document clearly that setting this env var on an existing deployment requires manually re-registering all cluster contexts.

@aabidsofi19 will you give this some thought? On a related note, @Kritagya123611, we have rules in our docs regarding how to handle connections that have duplicate cluster IDs. @chaitanyamedidar, please reason through this concern.

leecalcote:
**Missing documentation for `KUBERNETES_TARGET_NAMESPACE` in `server/models/k8s_context.go` - `getKubeNamespace()` helper**

`KUBERNETES_TARGET_NAMESPACE` is a new user-visible configuration env var aimed squarely at operators in namespace-scoped or restricted RBAC environments - precisely the audience who most needs to know it exists. The PR description explains the motivation well, but there is no corresponding documentation update under `docs/`.

Please add a brief entry to the relevant configuration or Kubernetes environment guide (e.g., `docs/content/en/guides/`) documenting:
- The env var name and purpose
- The resolution order (env var → service-account file → `kube-system`)
- When an operator would need to set it (namespace-scoped RBAC, multi-tenant deployments)

leecalcote:
**🔵 [PLAUSIBLE] `saNamespaceFile` package global is a fragile test seam**

`server/models/k8s_context.go:481`

```go
// saNamespaceFile is the path to the in-cluster service-account namespace file.
// Exposed as a variable so tests can override it.
var saNamespaceFile = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
```

The current tests all correctly restore this variable via `t.Cleanup` — no immediate bug. However, the bare package-level global remains a fragile seam:

- Any future test that mutates `saNamespaceFile` and omits `t.Cleanup` will silently corrupt the path for all subsequent tests in the package.
- The variable cannot be isolated per goroutine, so `t.Parallel()` tests that mutate it will race.

Consider one of:
1. Add a comment on the variable explicitly warning that tests **must** use `t.Cleanup` to restore it (low-friction).
2. Pass the file path as a parameter to `getKubeNamespace(path string)` and use a package-level wrapper that supplies the default — eliminates the global entirely and makes tests straightforward.

kritagyaio:
> **Silent server-ID overwrite on existing deployments in `server/machines/kubernetes/discover.go:38`**
> 
> `AssignServerID` is called unconditionally — no check for an existing non-nil `KubernetesServerID` — and `provider.SaveK8sContext` persists the result immediately afterward:
> 
> ```go
> // discover.go ~line 38
> if err := kc.AssignServerID(handler); err != nil { ... }
> // ...
> provider.SaveK8sContext(...)  // persists the newly derived ID
> ```
> 
> `AssignServerID` now delegates to `getKubeNamespace()`, which returns `KUBERNETES_TARGET_NAMESPACE` when set. The result: an operator running an **existing** Meshery deployment who sets `KUBERNETES_TARGET_NAMESPACE=custom-ns` will, on next context discovery, have every context's `KubernetesServerID` silently re-derived from `custom-ns` and persisted — replacing the IDs that were stored at original registration and orphaning any data (connections, events, references) tied to the old IDs.
> 
> **Suggested fix:** Guard the assignment in `discover.go` — skip `AssignServerID` (or at least the save) when `KubernetesServerID` is already non-empty. Alternatively, document clearly that setting this env var on an existing deployment requires manually re-registering all cluster contexts.
> 
> @aabidsofi19 will you give this some thought? On a related note, @Kritagya123611, we have rules in our docs regarding how to handle connections that have duplicate cluster IDs. @chaitanyamedidar, please reason through this concern.

Thanks for the detailed review. I've addressed this by adding a guard in AssignServerID so an existing non-empty KubernetesServerID is preserved and not re-derived. This prevents existing deployments from having their server IDs overwritten. Thanks for pointing this out.

kritagyaio:
> **Missing documentation for `KUBERNETES_TARGET_NAMESPACE` in `server/models/k8s_context.go` - `getKubeNamespace()` helper**
> 
> `KUBERNETES_TARGET_NAMESPACE` is a new user-visible configuration env var aimed squarely at operators in namespace-scoped or restricted RBAC environments - precisely the audience who most needs to know it exists. The PR description explains the motivation well, but there is no corresponding documentation update under `docs/`.
> 
> Please add a brief entry to the relevant configuration or Kubernetes environment guide (e.g., `docs/content/en/guides/`) documenting:
> 
> * The env var name and purpose
> * The resolution order (env var → service-account file → `kube-system`)
> * When an operator would need to set it (namespace-scoped RBAC, multi-tenant deployments)

Added documentation for KUBERNETES_TARGET_NAMESPACE, including its purpose, resolution order, and the intended namespace-scoped/RBAC deployment scenarios. Thanks for highlighting the documentation gap.

kritagyaio:
> **🔵 [PLAUSIBLE] `saNamespaceFile` package global is a fragile test seam**
> 
> `server/models/k8s_context.go:481`
> 
> ```go
> // saNamespaceFile is the path to the in-cluster service-account namespace file.
> // Exposed as a variable so tests can override it.
> var saNamespaceFile = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
> ```
> 
> The current tests all correctly restore this variable via `t.Cleanup` — no immediate bug. However, the bare package-level global remains a fragile seam:
> 
> * Any future test that mutates `saNamespaceFile` and omits `t.Cleanup` will silently corrupt the path for all subsequent tests in the package.
> * The variable cannot be isolated per goroutine, so `t.Parallel()` tests that mutate it will race.
> 
> Consider one of:
> 
> 1. Add a comment on the variable explicitly warning that tests **must** use `t.Cleanup` to restore it (low-friction).
> 2. Pass the file path as a parameter to `getKubeNamespace(path string)` and use a package-level wrapper that supplies the default — eliminates the global entirely and makes tests straightforward.

Addressed with a warning comment documenting that any test mutating saNamespaceFile must restore it using t.Cleanup() and should not use t.Parallel(). Thanks for the suggestion.

PR #20222, opened 2026-06-24
fix(ui): render empty state when loading completes

### LINKED ISSUE #20221: [UI] Empty state in MesheryTreeView becomes unreachable
### Current Behavior

In `MesheryTreeView`, the empty-state message becomes unreachable when the component enters the empty-state branch (`data.length === 0 && !searchText`).

As a result, the empty-state message is not shown when no records are available.

### Desired Behavior

When loading has completed and no records are available, the Registry view should display an appropriate empty-state message (for example, "No models found", "No registrants found", etc.).

### Screenshots/Logs

No runtime errors were observed.

The issue appears to be caused by the component logic itself.

The empty-state branch is entered when:

```tsx
data.length === 0 && !searchText
```

However, inside that same branch, the nested condition evaluates:

```tsx
isLoading || (data.length === 0 && !searchText)
```

Since `data.length === 0 && !searchText` is already true when execution reaches this branch, the expression will always evaluate to `true`.

As a result, the empty-state message becomes unreachable.

### Environment

* Browser: Chrome
* Host OS: Linux
* Meshery Server Version: latest master
* Platform: Docker

KumarNirupam1:
**Notes for Reviewers**

* Fixes #20221

### Issue

In `MesheryTreeView`, the empty-state message became unreachable when the component entered the empty-state branch.

The branch was entered when:

```tsx
data.length === 0 && !searchText
```

However, the nested condition repeated the same check:

```tsx
isLoading || (data.length === 0 && !searchText)
```

Since `data.length === 0 && !searchText` was already true at that point, the expression always evaluated to `true`, preventing the empty-state message from being rendered.

### Fix

* removed the redundant condition from the empty-state branch
* render the loading spinner only while `isLoading` is true
* allow the empty-state message to be displayed when loading completes and no records are available

### Testing

* verified the component renders the loading indicator while data is being fetched
* verified the empty-state message can be rendered when loading completes with no available records
* verified the existing "No result found" state is displayed when searching for unavailable records

**Signed commits**


KhushamBansal (approved):
LGTM!

rishiraj38 (approved):
LGTM. The inner check was redundant with the parent block, which is why it spinnered forever. Tests pass fine.

PR #20236, opened 2026-06-24
ci: add workflow classification to labeler.yml

MrDadhich456:
Adds a workflow classification annotation header to `labeler.yml` as part of the [GitHub Action Workflow Cleanup and Restructure](https://github.com/meshery/meshery/issues/18795) initiative.

### Classification

| Field | Value |
|-------|-------|
| **Workflow Type** | automation |
| **Security Level** | write-token (requires `pull-requests: write` via `GITHUB_TOKEN`) |
| **Trigger Model** | event-driven (`pull_request_target`) |
| **CI-Gating** | no |

### Context

This is the first in a series of PRs to annotate every workflow with its classification (security level + type) per the acceptance criteria:

> _"Every remaining workflow is annotated with its classification (security level + type)."_

The annotation format follows a standardized comment block at the top of each workflow file, making it easy to audit and search across all workflows.

Relates to #18795

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,3 +1,14 @@
+# ##################################################################
+# Workflow Classification (ref: https://github.com/meshery/meshery/issues/18795)
+# ------------------------------------------------------------------
+# Workflow Type:    automation
+# Security Level:  write-token (requires pull-requests: write via GITHUB_TOKEN)
+# Trigger Model:   event-driven (pull_request_target)
+# CI-Gating:       no
+# Description:     Automatically labels incoming pull requests based
+#                  on changed file paths, using the rules defined in
+#                  .github/labeler.yml.
+# ##################################################################
 name: "Pull Request Labeler"
 on:
 - pull_request_target


pontusringblom:
@ianrwhitney fyi

ianrwhitney:
Thanks so much for the contribution. I dont feel like this adds benefit here since the doc/annotation are longer than the workflow. I would aim your efforts at helping deprecate and uplift the workflows vs documentation rn.

MrDadhich456:
Thanks for the feedback @ianrwhitney — totally makes sense. I'll shift focus to deprecation and uplift instead.

I noticed `codeql.yml` (CodeQL Advanced) appears to be a redundant duplicate of `codeql-analysis.yml` (CodeQL) — both run scheduled scans on the same codebase, but `codeql-analysis.yml` is the more mature version (uses `ubuntu-24.04`, has PR failure comments, email alerts, and the repo guard). `codeql.yml` still uses `ubuntu-latest` and has no failure handling.

I've opened a follow-up PR to remove it: #20248 . Happy to take on more workflow cleanup after that.

Issue #20252, opened 2026-06-25
[Server] /healthz/ready reports ready before startup initialization completes, causing local E2E failures


ishwar170695:
### Current Behavior

When starting Meshery locally with the Local Provider, `/healthz/ready` returns `200 OK` before asynchronous startup initialization completes.

If Playwright tests are executed immediately after the readiness endpoint reports success, the Performance page test can fail because `meshery-metrics` is not yet available:

```text
expect(locator).toBeVisible() failed

Locator: getByTestId('meshery-metrics')
Expected: visible
Timeout: 60000ms
Error: element(s) not found
```

Running the same test again after startup initialization completes succeeds without any code changes.

### Expected Behavior

`/healthz/ready` (or another startup readiness signal) should indicate that the server is ready for normal operation before clients or test automation begin interacting with it.

### Environment

* **Host OS:** Windows 11
* **Platform:** Local Provider
* **Meshery Client Version:** N/A

### To Reproduce

1. Start the server:

```powershell
$env:PORT="9081"
$env:DEBUG="true"
$env:KEYS_PATH="$PWD\server\permissions\keys.csv"

cd server\cmd
go run main.go error.go
```

2. Wait until `http://localhost:9081/healthz/ready` returns `200 OK`.

3. Run:

```powershell
cd ui
npx playwright test tests/e2e/performance.spec.ts --project=chromium-local-provider
```

4. Observe the failure.

5. Wait 1–2 minutes and rerun the same command.

6. The test succeeds without any changes.
---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)
[...]

PR #20282, opened 2026-06-26
fix(server): delay readiness until startup initialization completes

vedant21-ctr:
Fixes:#20252

## Summary

This PR updates Meshery Server's readiness probe to report the server as ready only after the required startup initialization has completed.

Previously, `/healthz/ready` could return `200 OK` immediately after the HTTP server started, even while asynchronous startup tasks were still running. This could allow clients and integration tests to interact with the server before it was fully initialized.

## Changes

* Track completion of asynchronous startup initialization tasks.
* Update `/healthz/ready` to return `503 Service Unavailable` until startup initialization is complete.
* Preserve the existing `/healthz/live` behavior so liveness checks remain unaffected.
* Add unit tests covering liveness, readiness during startup, and readiness after initialization.

## Testing

* Ran:

  * `go test -v -run TestK8sHealthzHandler ./server/handlers/...`
  * `go test ./server/models/...`
* Manually verified:

  * `/healthz/live` returns `200 OK` immediately after startup.
  * `/healthz/ready` returns `503 Service Unavailable` while startup initialization is in progress.
  * `/healthz/ready` transitions to `200 OK` after startup initialization completes.

## Summary by CodeRabbit

- **New Features**
  - Added dedicated liveness and readiness health checks.
  - Liveness checks return a simple `ok` response.
  - Readiness checks report startup progress, including data seeding and provider verification.
  - Startup status is tracked safely while background initialization completes.
  - Organization listings now support filtering by name, owner, country, and region.

- **Bug Fixes**
  - Improved health endpoint responses and content-type handling for monitoring systems.
  - Added accurate readiness outcomes while startup tasks are still in progress.
  - Corrected organization filtering, ordering, and pagination behavior.


server/cmd/main.go
```diff
@@ -447,6 +448,7 @@ func main() {
 			}
 			rp.SyncPreferences()
 		}
+		hc.StartupStatus.ProviderTrackerComplete.Store(true)
 	}()
```
ishwar170695, line 448: are these startup tasks guaranteed to reach this point only after successful initialization? I'm wondering how failures or early returns during seeding/provider verification are handled. If a startup task exits without successfully completing initialization, should readiness continue reporting `503`, or is that already handled elsewhere?

vedant21-ctr: Thanks for pointing that out! I'll take another look at those startup paths and verify how failures or early returns are handled during seeding and provider verification. If there's a case where initialization can exit early, I'll update the implementation so the readiness status reflects that correctly.

vedant21-ctr: @ishwar170695 I went back through both startup goroutines to verify this.

From what I found, both goroutines always reach the point where the readiness flags are updated. The individual startup steps handle errors internally (logging warnings/errors and continuing), so the flags represent that the startup phase has completed rather than that every initialization step succeeded.

This seems to match the existing fail-forward startup behavior, but if the expectation is that readiness should remain 503 for any specific initialization failure, I'm happy to update the implementation accordingly.

ishwar170695: Thanks for checking and tracing both startup paths.

That clarification makes sense. If the existing startup behavior is intentionally fail-forward, then using these flags to indicate that the startup phase has completed (rather than that every initialization step succeeded) seems consistent with the current design. I don't think this PR needs to change that behavior.

LGTM from my side.

Also Could you reply to the remaining Gemini review comments, either by addressing them or briefly explaining why they're being skipped?

vedant21-ctr:
@NSTKrishna , @Sbragul26  i have looked to reviews and suggestion its good to go
please look at this pr
whenever you have time

server/handlers/k8s_healthz_handler.go
```diff
@@ -14,12 +14,37 @@ import (
 // - HTTP 503 with plain text error details when unhealthy
 // - Supports ?verbose=1 query parameter for detailed check results
 func (h *Handler) K8sHealthzHandler(w http.ResponseWriter, r *http.Request) {
+	// Differentiate between live and ready checks
```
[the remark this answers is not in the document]
vedant21-ctr: thanks for the suggestion. I looked into this during the review. In the current initialization flow, Handler and HandlerConfig are always initialized before the health handler is registered, so this condition isn't expected to occur. I don't think an additional defensive nil check is necessary here.

vedant21-ctr (commented):
lgtm

ishwar170695 (approved):

vedant21-ctr:
@NSTKrishna  , @Sbragul26  , @saurabhraghuvanshii  its ready to be merged
please check and approve it

dhruveshmishra:
LGTM @vedant21-ctr @saurabhraghuvanshii

server/models/handlers.go
```diff
@@ -300,6 +300,13 @@ type HandlerConfig struct {
...
 	EventsBuffer      *events.EventStreamer
 	OperatorTracker   *OperatorTracker
+	StartupStatus     StartupTracker
+}
+
+// StartupTracker tracks thread-safe status of asynchronous startup tasks.
+type StartupTracker struct {
+	SeedingComplete         atomic.Bool
+	ProviderTrackerComplete atomic.Bool
 }
```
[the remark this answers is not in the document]
vedant21-ctr: yes as told  Gos atomic type should not be copied adted first use

PR #20285, opened 2026-06-26
[mesheryctl] Fix minimum() returning max on argument ties

### LINKED ISSUE #20283: [mesheryctl] minimum() returns the maximum on ties, corrupting Levenshtein distance and "Did you mean?" suggestions
### Current Behavior

The `minimum(a, b, c int)` helper in `mesheryctl/pkg/utils/closest_arg.go` does not return the smallest of its three arguments when two of them tie for the minimum:

```go
func minimum(a, b, c int) int {
	if a < b && a < c {
		return a
	} else if b < a && b < c {
		return b
	} else {
		return c
	}
}
```

When `a == b` and both are strictly less than `c`, neither branch is taken and the function falls through to `return c` i.e. it returns the **largest** value:

- `minimum(1, 1, 2)` returns `2` (should be `1`)
- `minimum(2, 2, 5)` returns `5` (should be `2`)

`minimum()` is the inner reducer of the Levenshtein distance computed in the same file (`closest_arg.go:21`), which backs `FindClosestArg`. Because the tie case is hit during normal DP, it yields **incorrect edit distances**:

- `levenshtein("aba", "bab")` returns `3`; the correct edit distance is `2`.
- A brute-force comparison against a reference implementation over a 2 letter alphabet (strings up to length 6) found **340 mismatched pairs**.

`FindClosestArg` powers the `mesheryctl perf <invalid-subcommand>` → *"Did you mean this?"* suggestion (`mesheryctl/internal/cli/root/perf/perf.go:61`). Inflated distances can cause a worse match to be suggested, and combined with the `suggestedCmd[0] == args[0][0]` guard can suppress a valid suggestion entirely.

The existing unit test (`closest_arg_test.go`) only covers strictly-ordered inputs (`1,2,3`, `2,1,3`, `3,2,1`) and the all equal case (`1,1,1`), so the `a == b < c` tie path is never exercised.

### Desired Behavior

`minimum(a, b, c)` should always return the smallest of the three values, including ties, so that `levenshtein` produces correct edit distances and `FindClosestArg` suggests the genuinely closest subcommand. A straightforward reduction:

```go
func minimum(a, b, c int) int {
	m := a
	if b < m {
		m = b
	}
	if c < m {
		m = c
	}
	return m
}
```

A regression test covering the tie case (e.g. `minimum(1, 1, 2) == 1`) should be added.

### Screenshots/Logs

Reproduced with a standalone program using the verbatim `minimum`/`levenshtein` from `closest_arg.go`:

```
minimum(1,1,2) = 2   (correct = 1)
minimum(2,2,5) = 5   (correct = 2)
lev("aba","bab") = 3 (correct = 2)
```

### Environment

- Host OS: N/A (platform-independent Go logic)
- Meshery Server Version: N/A (CLI-only logic bug)
- Meshery Client Version: stable-v1.0.48
- Platform: N/A (no server/deployment involved)

Priyanshu6968:
Fixes #20283

**Notes for Reviewers**

- This PR fixes #20283

`minimum(a, b, c)` in `mesheryctl/pkg/utils/closest_arg.go` returned the
largest value when two arguments tied for the minimum (`a == b < c`),
since neither strict `<` branch matched and it fell through to `return c`.
Rewrote it as a straightforward min-reduction and added a tie-case
regression assertion to the existing `Test minimum` subtest.

### DIFF
--- a/mesheryctl/pkg/utils/closest_arg.go
+++ b/mesheryctl/pkg/utils/closest_arg.go
@@ -26,13 +26,14 @@ func levenshtein(str1, str2 []rune) int {
 }
 
 func minimum(a, b, c int) int {
-	if a < b && a < c {
-		return a
-	} else if b < a && b < c {
-		return b
-	} else {
-		return c
+	m := a
+	if b < m {
+		m = b
 	}
+	if c < m {
+		m = c
+	}
+	return m
 }
 
 func minIntSlice(values []int) (int, int) {
--- a/mesheryctl/pkg/utils/closest_arg_test.go
+++ b/mesheryctl/pkg/utils/closest_arg_test.go
@@ -48,6 +48,11 @@ func TestClosestArgs(t *testing.T) {
 		if min != 1 {
 			t.Errorf("Expected 1, got %d", min)
 		}
+
+		min = minimum(1, 1, 2)
+		if min != 1 {
+			t.Errorf("Expected 1, got %d", min)
+		}
 	})
 
 	t.Run("Test levenshtein", func(t *testing.T) {


ishwar170695 (approved):
LGTM. The new implementation correctly handles tied minimum values, and the added regression test covers the previously missing edge case.

leecalcote (approved):
Thanks!

leecalcote:
Thanks for reviewing, @ishwar170695. 👍

PR #20248, opened 2026-06-25
ci: remove duplicate CodeQL Advanced workflow

MrDadhich456:
## Description

Removes `codeql.yml` ("CodeQL Advanced") which is a redundant duplicate of the actively maintained `codeql-analysis.yml` ("CodeQL"). Both workflows run weekly scheduled CodeQL scans on the same languages (Go, JavaScript), wasting CI minutes.

### Why `codeql-analysis.yml` is the keeper

| Feature | `codeql.yml` (removed) | `codeql-analysis.yml` (kept) |
|---|---|---|
| Runner | `ubuntu-latest` | `ubuntu-24.04` |
| Repo guard | ❌ | ✅ `github.repository == 'meshery/meshery'` |
| Checkout version | `@v4` | `@v6` |
| PR failure comments | ❌ | ✅ |
| Email alerts on failure | ❌ | ✅ |
| Introspection logging | ❌ | ✅ |
| Config | Inline paths-ignore | `.github/codeql-config.yml` |

### Impact
- Eliminates ~3 redundant CI runs per week (one per language matrix entry)
- No loss of security scanning coverage — `codeql-analysis.yml` covers the same languages with better error handling

Relates to #18795

- [x] Yes, I signed my commits.

PR #20346, opened 2026-06-28
ci: fix CodeQL Go analysis reporting

MrDadhich456:
**Follow-up to:** #20248
**Fixes:** CodeQL Go analysis returning empty/failing to report results.

### Problem
After the consolidation of the CodeQL workflows in #20248, the remaining `codeql-analysis.yml` workflow failed to properly scan the Go source code. This happened because the workflow defaulted to `autobuild` for Go. CodeQL's autobuilder detected the `Makefile` in the root directory and ran `make` (which targets `providers-propagate`), resulting in zero Go files being compiled. Since CodeQL for Go relies on intercepting a successful compilation, no Go code was analyzed.

### Solution
This PR configures CodeQL to use a manual build mode for Go and sets up the correct build environment:
1. **Set up Go**: Added `actions/setup-go@v6` using the version specified in `go.mod`.
2. **Explicit Build Mode**: Configured `build-mode: manual` for Go and `none` for JavaScript.
3. **Manual Compilation Steps**: Replaced the default `Autobuild` step with explicit `go build` commands to compile the primary components of the repository:
   - Meshery Server (`server/cmd`)
   - Meshery CLI (`mesheryctl`)
   - WASM Policy Engine (`server/policies/wasm` with `GOOS=js GOARCH=wasm`)

By explicitly compiling these components, CodeQL can successfully observe the builds and generate the security analysis database for the Go codebase.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -72,13 +72,13 @@ jobs:
         if: matrix.language == 'go'
         run: |
           # Build Meshery Server
-          go build -o ./server/cmd/meshery-server ./server/cmd/main.go ./server/cmd/error.go
+          go build -o ./server/cmd/meshery-server ./server/cmd
           
           # Build Meshery CLI
-          go build -o ./mesheryctl/mesheryctl ./mesheryctl/cmd/mesheryctl/main.go
+          go build -o ./mesheryctl/mesheryctl ./mesheryctl/cmd/mesheryctl
           
           # Build WASM Policy Engine
-          GOOS=js GOARCH=wasm go build -o ./server/policies/wasm/policy_engine.wasm ./server/policies/wasm/main.go
+          (cd server/policies/wasm && GOOS=js GOARCH=wasm go build -o policy_engine.wasm .)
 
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -50,32 +50,35 @@ jobs:
...
-      #    uses a compiled language
-
-      #- run: |
-      #   make bootstrap
-      #   make release
+      - name: Build Go code
+        if: matrix.language == 'go'
+        run: |
+          # Build Meshery Server
+          go build -o ./server/cmd/meshery-server ./server/cmd
```
yi-nuo426: Use our `make` targets.

yi-nuo426 (changes requested):

.github/workflows/codeql-analysis.yml
MrDadhich456: Done! Replaced the manual go build commands with our existing build-server, mesheryctl build, and wasm-engine make targets.

MrDadhich456:
@yi-nuo426

PR #20362, opened 2026-06-29
[Server] Fix determinePluralWord pluralizing a zero count as singular

### LINKED ISSUE #20361: [Server] determinePluralWord() returns the singular form for a count of zero
### Current Behavior
The helper `determinePluralWord(count int, word string) string` in [`server/handlers/component_generator_helper.go`]:

```go
func determinePluralWord(count int, word string) string {
	if count > 1 {                 // <-- should be: count != 1
		if strings.HasSuffix(word, "y") {
			return word[:len(word)-1] + "ies"
		} else if strings.HasSuffix(word, "s") {
			return word + "es"
		}
		return word + "s"
	}
	return word
}
```
Because the guard is `count > 1`, a count of **0** falls through and returns the **singular** form. In English a count of zero takes the plural ("0 components"), so the produced text is grammatically wrong:

### Expected Behavior
`determinePluralWord` should return the plural form for any count that is not exactly `1` (i.e. for `0` and for `>= 2`), matching standard English usage. The fix is a one-line operator change:

```go
	if count != 1 {
```

A small unit test covering `0`, `1`, `2`, and the `-y` / `-s` suffix branches should accompany the fix.

### Screenshots/Logs
```go
got := determinePluralWord(0, "component") // returns "component", want "components"
```

### Environment

- **Host OS:** OS-independent
- **Platform:** Docker or Kubernetes (server side logic, environment independent)
- **Meshery Server Version:** present on `master` 

Priyanshu6968:
**Notes for Reviewers**

- This PR fixes #20361


server/handlers/component_generator_helper.go
```diff
@@ -475,7 +475,7 @@ func ErrMsgContruct(response *models.RegistryAPIResponse) string {
 	return msg
 }
 func determinePluralWord(count int, word string) string {
-	if count > 1 {
+	if count != 1 {
```
leecalcote, line 478: What happens in the case of "" or null word?

When count = 0, the word is returned as-is?

Priyanshu6968: Go strings can't be nil, and every caller passes a literal ("component", "relationship", etc.), so "" never reaches this. If it did it returns "s" (no OOB ,  the word[:len-1] slice only runs in the HasSuffix "y" branch, which needs a non-empty word).
To answer second question ,No that was the bug. Old count > 1 returned it as-is ("0 component"), count != 1 now pluralizes i.e. "0 components". Only count == 1 returns as-is .

PR #20375, opened 2026-07-01
feat(events): Migrate event streaming from GraphQL to SSE

aabidsofi19:
Introduce a new HTTP handler `/api/system/events/subscribe`

**Notes for Reviewers**

- This PR fixes #


server/handlers/events_stream_handler.go
```diff
@@ -0,0 +1,104 @@
...
+
+	flusher, ok := w.(http.Flusher)
+	if !ok {
+		writeJSONError(w, "streaming unsupported", http.StatusInternalServerError)
+		return
+	}
+
+	// Subscribe before writing headers so events published between the request
+	// arriving and the stream being ready aren't dropped.
+	ch, unsubscribe := h.config.EventBroadcaster.Subscribe(user.ID)
```
leecalcote, line 53: **Silent event loss under burst: the broadcaster's `IsClosed` consumes the buffered event.**

This handler is now the primary consumer of `EventBroadcaster`. `Publish` (`server/models/event_broadcast.go:77`) guards each send with `if !utils.IsClosed(client)`, but `utils.IsClosed` (`server/helpers/utils/utils.go:151`) is a *destructive* non-blocking receive:

```go
select {
case <-ch: // succeeds when closed OR when a value is buffered
    return true
default:
}
return false
```

The subscription channel has capacity 1 (`event_broadcast.go:26`). If event A is still buffered when event B is published for the same user, `IsClosed(ch)` drains A and returns `true`, so `Publish` skips B - **both A and B never reach the SSE client**. Reachable whenever events are emitted in quick succession to one user (e.g. `design_import.go` publishes 6+ in a row) or when the client is slow to drain.

The fix belongs in the broadcaster: replace the `IsClosed`-then-send with a non-blocking send under the per-client mutex (`select { case client <- data: default: }`) so a full/closed channel is handled without consuming buffered data. Flagging here because the migration makes this handler depend on it.

server/handlers/events_stream_handler.go
```diff
@@ -0,0 +1,104 @@
...
+	defer unsubscribe()
```
leecalcote, line 54: **`defer unsubscribe()` can crash the server: send on a closed channel.**

`unsubscribe` closes the subscription channel (`event_broadcast.go:54`) while holding `client.mu`, but `Publish` reads `listeners` and sends **without** that lock (`event_broadcast.go:76-80`):

1. A publisher goroutine (`go h.config.EventBroadcaster.Publish(userID, ...)`) snapshots `listeners` and passes the `IsClosed` check.
2. This request's context is cancelled; the handler returns and `defer unsubscribe()` closes the channel.
3. The publisher runs `client <- data` on the now-closed channel -> **panic**, unrecovered, taking down the process.

The old GraphQL resolver had the same `unsubscribe()`-on-`ctx.Done()` shape, so this is pre-existing, but the SSE handler makes it the live path. Safest fix is in the broadcaster: perform the send under `client.mu` (non-blocking) so close and send are mutually exclusive, or `recover()` around the send loop.

server/handlers/events_stream_handler.go
```diff
@@ -0,0 +1,104 @@
...
+				event = &e
+			default:
+				continue
+			}
+			payload, err := json.Marshal(event)
+			if err != nil {
+				h.log.Error(models.ErrMarshal(err, "event"))
+				continue
+			}
+			if _, err := fmt.Fprintf(w, "data: %s\n\n", payload); err != nil {
```
leecalcote, line 104: **A connected-but-not-reading client pins this goroutine and backs up the user's event pipeline.**

`fmt.Fprintf(w, ...)` blocks on TCP backpressure when the peer stops reading (a slowloris-style client, or a stalled tab). The read loop is a single goroutine, so a blocked write means `ch` is never drained; the capacity-1 channel fills and every subsequent `go Publish(userID, ...)` blocks on the send (and can silently drop events per the note above). Genuine disconnects are covered by `ctx.Done()`, but a peer holding the socket open without reading is not.

Consider bounding the write - e.g. a per-write deadline via `http.NewResponseController(w).SetWriteDeadline(...)` (Go 1.20+) and returning on timeout - so one stuck client can't wedge that user's stream.

ui/lib/eventsSubscription.ts
```diff
@@ -0,0 +1,50 @@
...
+    } catch (error) {
+      console.error('[EventsSubscription] Failed to parse event', error, message.data);
+    }
+  };
+
+  source.onerror = (error) => {
+    // EventSource silently auto-reconnects while readyState is CONNECTING (0).
+    // Only surface the error when the browser has permanently closed the stream
+    // (CLOSED = 2) so the caller can decide whether to re-subscribe.
+    if (source.readyState === EventSource.CLOSED) {
```
leecalcote: **Permanent errors trigger an un-backed-off reconnect storm.**

`onError` fires only when `readyState === CLOSED`, which is exactly the *non-recoverable* case (server returned 401/500, or a redirect to a non-`text/event-stream` page) - EventSource does not auto-retry those. `operationsCenter.ts:95` handles the resulting `ERROR_OCCURRED_IN_SUBSCRIPTION` with `actions: [stopChild('subscriptionActor'), 'spawnSubscriptionActor']` (no delay), immediately opening a fresh `EventSource` that hits the same permanent error, and so on. This bypasses EventSource's built-in ~3s reconnect delay and turns a session expiry into a tight request loop against `/api/system/events/subscribe`.

Suggest bounded backoff before respawn - e.g. route the error through a transient `reconnecting` state with `after: { RECONNECT_DELAY: { target: 'idle', actions: 'spawnSubscriptionActor' } }` - and update the `respawns ... on ERROR_OCCURRED_IN_SUBSCRIPTION` test to drive fake timers. Transient drops already reconnect natively (readyState CONNECTING), so only the permanent-error path needs this.

server/router/server.go
```diff
@@ -137,6 +137,9 @@ func NewRouter(_ context.Context, h models.HandlerInterface, port int, g http.Ha
 		// This will be changed to /api/events once the UI is compeltely updated to use new events and SSE is tunred off, otherwise existing events will break.
 	gMux.Handle("/api/system/events", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.GetAllEvents), models.ProviderAuth))).
 		Methods("GET")
+	// SSE stream of live events; replaces the subscribeEvents GraphQL subscription.
+	gMux.Handle("/api/system/events/subscribe", h.ProviderMiddleware(h.AuthMiddleware(h.SessionInjectorMiddleware(h.SubscribeEventsHandler), models.ProviderAuth))).
```
leecalcote, line 141: **New public endpoint needs docs.** `AGENTS.md` requires 'New features need docs' / 'Documentation Is Part of the Change'. `/api/system/events/subscribe` is a new user-facing SSE endpoint but isn't reflected in `docs/data/openapi.yml` (which documents the sibling `/events` routes) or the events API docs. Please add it in this PR - at minimum a path entry noting the `text/event-stream` response and that it replaces the `subscribeEvents` GraphQL subscription.

PR #20410, opened 2026-07-02
feat(operator-chart): modernize to operator v1.0.0 — CRD updates on upgrade, opt-in conversion webhook, release-driven publishing

alexquincy:
## What

Brings the vendored `meshery-operator` chart from its 2022 shape up to operator v1.0.0, makes CRD updates actually reach live clusters, and wires this repo to receive automated syncs + publish operator-versioned charts on every operator release.

### Chart modernization (`install/kubernetes/helm/meshery-operator`)

| Was | Now |
|---|---|
| `meshery/meshery-operator:stable-latest` baked into the repository string | `image.repository` + `image.tag` (defaults to chart `appVersion`, kept in lockstep with operator releases) |
| kube-rbac-proxy sidecar + `meshery-proxy-role` | gone — the operator serves metrics itself over TLS with authn/authz (port 8443); scrapers bind to `meshery-metrics-reader` |
| **wildcard `*/*` `meshery-controller-role`** | gone — `meshery-operator-role` mirrors the operator's least-privilege `config/rbac/role.yaml` verbatim |
| no probes, no securityContext, dead port-10000 Service + helm test | `/healthz`+`/readyz` probes, `runAsNonRoot` 65532 / `readOnlyRootFilesystem` / drop ALL, metrics Service 8443, manager.yaml resource parity |
| leader election Role without `leases` | `coordination.k8s.io/leases` added (gated by `leaderElection.enabled`) |
| CRDs: v1alpha1-only, controller-gen v0.6.1, applied only on first install | v1alpha1+v1alpha2 (storage v1alpha2) from operator v1.0.0, **re-applied on every install and upgrade** by a `pre-install,pre-upgrade` hook Job (`crds.updateJob`, default on) — Helm never touches `crds/` on upgrade, which is how the vendored CRDs sat stale on live clusters for four years |

### Conversion strategy: `None` by default, webhook opt-in

The shipped CRDs use conversion `strategy: None`, which is exact while the v1alpha1/v1alpha2 schemas are field-identical (they are — the operator's spoke conversion is a lossless JSON round-trip). This deliberately avoids making cert-manager a hard dependency: Meshery-Server-managed clusters don't have it, and a conversion webhook that can't get a CA breaks *all* CR access. `webhook.enabled=true` patches webhook conversion onto the CRDs via the update Job — self-signed chart-generated CA by default (reused across upgrades via `lookup`), or cert-manager v1 `Issuer`/`Certificate` + CA injection with `webhook.certManager.enabled=true`. Disabling resets conversion to `None` explicitly (SSA wouldn't, and a dangling webhook would break CR reads). When the schemas diverge upstream, the chart default flips to webhook conversion — the operator repo's `api/v1alpha1/conversion.go` comment now carries that instruction.

### Single CRD source

`install/kubernetes/helm/meshery/crds/crds.yaml` (a byte-identical duplicate) is retired. The operator chart owns the operator CRDs: install-time `crds/` copy + `files/crds.yaml` for the update Job, both regenerated per release by the operator's sync automation. The `meshery-operator` dependency is re-vendored at 1.0.0 (`Chart.lock` + `charts/meshery-operator-1.0.0.tgz`).

### Release automation (receiver side)

`.github/workflows/release-operator-chart.yml` listens for `repository_dispatch: meshery-operator-released` (fired by the operator's new sync-downstream workflow after it pushes the synced content here) and publishes **only** the meshery-operator chart to meshery.io/charts at the operator's version — guarded on master's chart `appVersion` matching the dispatched version, so a publish can never precede its sync commit. Charts stamped with Meshery Server release versions (what Meshery Server's meshkit deploy path looks up) continue publishing from `helm-chart-releaser.yml` unchanged; operator-versioned entries are additive index entries.

## Verification (kind v1.35.0, helm 4.1.4)

- `helm lint --with-subcharts` clean on both charts; template matrix (default / `crds.updateJob.enabled=false` / `webhook.enabled=true` / `+certManager` / guard rails) asserted.
- Full e2e on kind: install → manager Available, hook Job applies CRDs (`storedVersions: [v1alpha2]`, strategy None); `kubectl apply` of the v1alpha1 Broker sample reads back via v1alpha2 with fields intact; operator reconciles it into the NATS StatefulSet + Services. `helm upgrade` → Job re-runs (event log shows both runs). `--set webhook.enabled=true` → conversion flips to `Webhook` with populated caBundle and **live** webhook conversion works both directions through the operator pod. Flipping back → conversion cleanly resets to `None`, reads keep working.

## Prior art (Crossplane, Istio — verified against their current repos)

| Practice | Crossplane | Istio | This PR |
|---|---|---|---|
| Multi-version CRDs, identical schemas, **no conversion webhook** | ✓ all 21 core CRDs (`cluster/crds/`, incl. XRDs serving v1+v2) | ✓ all 15 CRDs (`crd-all.gen.yaml`; v1/v1beta1/v1alpha3 byte-identical) | ✓ strategy `None` default |
| No cert-manager dependency; self-generated webhook certs | ✓ init `TLSCertificateGenerator` | ✓ istiod self-signs | ✓ `genSignedCert` default, cert-manager opt-in |
| CRD updates reach clusters on upgrade (Helm `crds/` limitation solved) | init container applies CRDs on every pod start | CRDs as templates + documented one-time adoption loop | pre-install/pre-upgrade hook Job (SSA) |
| CRDs survive uninstall; removal is a documented manual step | ✓ (never Helm-owned) | ✓ (`helm.sh/resource-policy: keep`) | ✓ (never Helm-owned; documented in chart README) |
| Generated CRDs + CI drift gate | ✓ controller-gen + `check-diff` | ✓ `update_crds.sh` + `gen-check` | ✓ controller-gen drift gate upstream + release-time sync |
| Chart version auto-stamped to app release | ✓ nix build stamps | ✓ release-builder stamps | ✓ sync script + publish workflow stamp |

Deliberate divergences: (1) Istio's templates-based CRDs require a manual adoption step on existing installs (`kubectl label/annotate` loop per their Helm upgrade docs) — unacceptable here because Meshery Server upgrades operators unattended via meshkit, hence the hook Job. (2) Crossplane's init-container self-management requires granting the operator's own ServiceAccount CRD-write RBAC (they grant `customresourcedefinitions` verbs `*`) and operator image changes; the hook Job keeps CRD-write on a short-lived dedicated SA instead, preserving the operator's least-privilege posture. (3) Both references keep the chart in the project repo and publish to a dedicated chart repo; this chart stays in meshery/meshery because Meshery Server's meshkit lookup requests chart version == server release — converging on operator-repo chart ownership is a possible follow-up once that coupling is relaxed.

Known follow-up (what Crossplane automates that we don't yet): storage-version migration. Existing clusters keep `status.storedVersions: [v1alpha1, v1alpha2]` after the flip; Crossplane's init runs a `CoreCRDsMigrator` that rewrites stored objects and prunes storedVersions. With typically ≤2 Broker/MeshSync objects per cluster this is low-risk to defer, and v1alpha1 cannot be dropped from `served` until it's done.

## Post-merge

Run the `Release Meshery Operator Chart` workflow once (`release-ver: 1.0.0`) to publish the first operator-versioned chart; subsequent operator releases publish automatically.

## Companion PR

meshery/meshery-operator: `make crds`, `hack/sync-downstream.sh` (which generated this PR's `crds/`+`files/`+version-bump content), and the `sync-downstream.yml` release workflow.


leecalcote (approved):

leecalcote (dismissed):

alexquincy:
Review feedback addressed in 8f5e1c3 — item-by-item disposition:

**Incorporated**

- ✅ **Pin `stefanprodan/helm-gh-pages`** (reviewer + Copilot): pinned to immutable commit `89c6698c192e70ed0e495bee7d3d1ca5b477fe82` (master as of 2023-09-27, > v1.7.0) with a comment explaining the `GH_ACCESS_TOKEN` exposure rationale. The pre-existing `@master` usages in `helm-chart-releaser.yml` are left as-is (out of this PR's scope).
- ✅ **`activeDeadlineSeconds` on the CRD update Job**: set to `300` (aligned with Helm's default 5m wait) so a wedged hook fails fast instead of blocking `helm upgrade`.
- ✅ **appVersion guard hardening**: now `awk '/^appVersion:/ {v=$2; gsub(/"/,"",v); print v}'`, handling quoted and unquoted forms. (The suggested `sed` in the review was itself malformed, so an awk equivalent was used.)
- ✅ **Fullname-helper names for hook resources** (Gemini ×4): SA/ClusterRole/CRB/Job/ConfigMaps now derive from `meshery-operator.fullname`. Note the default rendering is byte-identical to before (the chart pins `fullnameOverride: meshery-operator`), so this is consistency/robustness for overridden names — true multi-instance installs of this chart remain unsupported regardless, since the CRDs and the operator's watches are cluster-global.
- ✅ **`HOME=/tmp` + `emptyDir` on the Job** (Gemini): added. Empirically the Job already succeeded under `readOnlyRootFilesystem` in the kind run (kubectl degrades gracefully without its cache), so this is warning-hygiene rather than a fix — but it's free.
- ✅ **Webhook conversion dangling after `helm uninstall`** (Copilot — good catch, this was a real gap): added a `pre-delete` teardown Job (webhook mode only) that resets `spec.conversion` to `None` and strips the cainjection annotation before the webhook Service disappears. Verified on kind: install → `webhook.enabled=true` upgrade (strategy `Webhook`, live conversion) → `helm uninstall` → CRDs survive with strategy `None`, cainjection annotation gone, and `kubectl get brokers.v1alpha1` still works.
- ✅ **Failing `Integration test with meshsync`** — root-caused to this PR: the setup script waited on `statefulset/meshery-broker` with `app=meshery,component=broker` labels, but operator ≥ 1.0.0 renders the broker from the official NATS chart as `statefulset/meshery-nats` with `app.kubernetes.io/*` labels. `setup.sh` updated accordingly.

**Rejected, with evidence**

- ❌ **"`actions/checkout@v6` does not exist; latest is v4"**: v6 exists and is already in production use in this very repo — `helm-chart-releaser.yml` uses `actions/checkout@v6`, and meshery-operator's CI pins `actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3` (green on every run). No change.
- ❌ **`/tmp` `emptyDir` for the manager Deployment** (Gemini): the operator's own `config/manager/manager.yaml` runs with `readOnlyRootFilesystem: true` and no `/tmp` mount — in the operator repo's kind e2e and in this PR's kind verification the manager runs, serves probes, and reconciles without any writable rootfs. controller-runtime does not require `/tmp`; the only disk path it touches here is the webhook cert dir, which is a mounted Secret volume when `webhook.enabled=true`. Keeping parity with the upstream manifest.

leecalcote (approved):

Issue #19943, opened 2026-06-09
[bug] mesheryctl version check has no HTTP timeout and test reports meaningless error message


Harishrs2006:
### Current Behavior
Two separate issues in mesheryctl:

1. version.go (line 134): http.Client{} has no timeout set. If the Meshery
   server is slow or unreachable, the version check hangs indefinitely,
   blocking the CLI.

2. config_test.go (line 47): TestGetLocation calls t.Errorf("Fail") when
   os.UserHomeDir() returns an error. This message tells the developer
   nothing about what actually failed.

### Desired Behavior
1. version.go: http.Client should have a timeout (e.g. 10 seconds) so the
   version check fails fast instead of hanging forever.

2. config_test.go: t.Errorf should include the actual error:
   t.Errorf("os.UserHomeDir() returned unexpected error: %v", err)

### Screenshots/Logs
// version.go line 134 — no timeout
client := &http.Client{}

// config_test.go line 47 — meaningless message
t.Errorf("Fail")

### Environment
- Host OS: Windows
- Meshery Client Version: stable-v0.8.x
- Platform: Docker

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 📚 [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- 📺 [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
- ⌨️ [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- ⌨️ [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- ⌨️ [Meshkit errors Guide](https://docs.meshery.io/project/contributing/contributing-error)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

D-source1602:
[...]

PR #20436, opened 2026-07-03
fix(mesheryctl): add HTTP timeout to version check and improve test error message

D-source1602:
mesheryctl version could hang indefinitely when the Meshery server or GitHub was slow or unreachable — the http.Client instances involved had no timeout set. This PR fixes all three unbounded clients on the version command's execution path.

Changes

version.go: added utils.MesheryctlHTTPTimeout (10s) to the server version request
pkg/utils/latest_version.go: added MesheryctlHTTPTimeout (10s) to GetLatestVersionForMesheryctl(); defines the shared constant
config.go: added a 10s timeout to ValidateVersion()'s GitHub release check — this runs in versionCmd.PreRunE before version.go is reached, so the command could still hang without this fix. Uses 10 * time.Second directly rather than utils.MesheryctlHTTPTimeout to avoid an import cycle (config ← utils ← config)
version.go: moved defer utils.CheckMesheryctlClientVersion(build) above NewRequest/client.Do so the CLI update check always fires, even when the server URL is malformed
config_test.go: replaced t.Errorf("Fail") with t.Fatalf and the actual error in TestGetLocation for clear, immediate failure on an invalid precondition

Out of scope

This PR does not address the Run → RunE conversion for exit code propagation, nor other http.Client{} instances in mesheryctl outside the version command path.

Tested locally

CGO_ENABLED=0 go build ./... — clean
CGO_ENABLED=0 go test ./internal/cli/root/config/... -v — all passing
CGO_ENABLED=0 go vet ./... and gofmt -l — clean on all four touched files

### DIFF
--- a/mesheryctl/internal/cli/root/config/config_test.go
+++ b/mesheryctl/internal/cli/root/config/config_test.go
@@ -35,7 +35,8 @@ func TestGetBuild(t *testing.T) {
 
 func TestGetLocation(t *testing.T) {
 	token := Token{}
-	const mesheryFolder = ".meshery"
+	mesheryFolder := ".meshery"
+	
 	for _, test := range tests {
 		token.SetLocation(test)
 		got := token.GetLocation()
--- a/mesheryctl/internal/cli/root/config/config_test.go
+++ b/mesheryctl/internal/cli/root/config/config_test.go
@@ -35,16 +35,15 @@ func TestGetBuild(t *testing.T) {
 
 func TestGetLocation(t *testing.T) {
 	token := Token{}
+	const mesheryFolder = ".meshery"
 	for _, test := range tests {
 		token.SetLocation(test)
 		got := token.GetLocation()
 		want, err := os.UserHomeDir()
-		MesheryFolder := ".meshery"
-		path := filepath.Join(want, MesheryFolder, test)
-		want = path
 		if err != nil {
-			t.Errorf("Fail")
+			t.Fatalf("os.UserHomeDir() returned unexpected error: %v", err)
 		}
+		want = filepath.Join(want, mesheryFolder, test)
 		if got != want {
 			t.Errorf("got %q want %q", got, want)
 		}
--- a/mesheryctl/internal/cli/root/version.go
+++ b/mesheryctl/internal/cli/root/version.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"time"
 
 	"github.com/meshery/meshery/mesheryctl/internal/cli/root/config"
 	"github.com/meshery/meshery/mesheryctl/internal/cli/root/constants"
@@ -113,6 +114,7 @@ mesheryctl version
 		url := mctlCfg.GetBaseMesheryURL()
 		build := constants.GetMesheryctlVersion()
 		commitsha := constants.GetMesheryctlCommitsha()
+		defer utils.CheckMesheryctlClientVersion(build)
 
 		version := config.Version{
 			Build:          "unavailable",
@@ -130,11 +132,15 @@ mesheryctl version
 			return
 		}
 
-		defer utils.CheckMesheryctlClientVersion(build)
-		client := &http.Client{}
+		client := &http.Client{Timeout: 10 * time.Second}
 		resp, err := client.Do(req)
 
 		if err != nil {
+
+			// resp is nil here except when CheckRedirect fails, and in that
+			// case net/http has already closed resp.Body for us — see the
+			// (Client).Do docs — so there is nothing left to close.
+
 			utils.PrintToTable(header, rows, nil)
 			utils.Log.Warn(ErrConnectingToServer(err))
 			return
--- a/mesheryctl/internal/cli/root/config/config.go
+++ b/mesheryctl/internal/cli/root/config/config.go
@@ -14,6 +14,7 @@ import (
 	"github.com/meshery/meshery/mesheryctl/pkg/constants"
 
 	"net/http"
+	"time"
 )
 
 // legacyProviderConfigWarnOnce gates the one-shot deprecation warning emitted
@@ -256,7 +257,7 @@ func (ctx *Context) ValidateVersion() error {
 	if err != nil {
 		return err
 	}
-	client := &http.Client{}
+	client := &http.Client{Timeout: 10 * time.Second}
 	resp, err := client.Do(req)
 	if err != nil {
 		return errors.Wrapf(err, "failed to make GET request to %s", url)
--- a/mesheryctl/pkg/utils/latest_version.go
+++ b/mesheryctl/pkg/utils/latest_version.go
@@ -5,7 +5,9 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"time"
 )
+
 
 func GetLatestVersionForMesheryctl() (string, error) {
 	req, err := http.NewRequest(http.MethodGet, "https://docs.meshery.io/project/releases/latest", nil)
@@ -14,7 +16,8 @@ func GetLatestVersionForMesheryctl() (string, error) {
 		return "", err
 	}
 
-	client := http.Client{}
+	
+client := &http.Client{Timeout: 10 * time.Second}
 
 	resp, err := client.Do(req)
 	if err != nil {
--- a/mesheryctl/pkg/utils/latest_version.go
+++ b/mesheryctl/pkg/utils/latest_version.go
@@ -17,7 +17,7 @@ func GetLatestVersionForMesheryctl() (string, error) {
 	}
 
 	
-client := &http.Client{Timeout: 10 * time.Second}
+	client := &http.Client{Timeout: 10 * time.Second}
 
 	resp, err := client.Do(req)
 	if err != nil {


SuryanshGarg04 (commented):
@D-source1602 The existing review comments raise valid concerns. Please address those before merging. Otherwise, the changes look good to me.

ishwar170695:
One observation about the control flow: defer `utils.CheckMesheryctlClientVersion(build)` is only registered after `http.NewRequest` succeeds. If `http.NewRequest` fails (e.g., because the configured server URL is malformed), the function returns before the defer is registered, so the CLI update check is skipped as well. Since the CLI update check appears to be independent of the server version request, would it make sense to register the defer earlier so it always runs?

D-source1602:
Fixed in dc519ce — added a 10s timeout to the client in GetLatestVersionForMesheryctl(), matching the timeout pattern used in version.go.

D-source1602:
@ishwar170695 , fixed in dc519ce. Moved defer utils.CheckMesheryctlClientVersion(build) above the NewRequest/client.Do calls so it always registers regardless of whether the server URL is malformed.

D-source1602:
@SuryanshGarg04 Both review comments are addressed in dc519ce — timeout added to GetLatestVersionForMesheryctl(), and the defer moved earlier in version.go so it always fires. Re-requesting your review whenever you have a moment.

SuryanshGarg04 (approved):
LGTM! Thanks for addressing all the review feedback.

Maanvi212006 (approved):
Tested it locally — build, vet, gofmt, and all tests pass. Also simulated a stalled server and confirmed it now times out at ~10s instead of hanging forever, so the core fix works.
Minor nit: the 10s timeout is hardcoded, could make it configurable, but not blocking. 
LGTM!

mesheryctl/internal/cli/root/config/config_test.go
```diff
@@ -39,12 +39,11 @@ func TestGetLocation(t *testing.T) {
...
 		got := token.GetLocation()
 		want, err := os.UserHomeDir()
-		MesheryFolder := ".meshery"
-		path := filepath.Join(want, MesheryFolder, test)
-		want = path
 		if err != nil {
-			t.Errorf("Fail")
+			t.Fatalf("os.UserHomeDir() returned unexpected error: %v", err)
 		}
+		MesheryFolder := ".meshery"
```
banana-three-join: MesheryFolder is redeclared on each for loop iteration. If the value is the same, it can be declared outside of the range

mesheryctl/internal/cli/root/version.go
```diff
@@ -130,11 +131,12 @@ mesheryctl version
...
 		}
 
-		defer utils.CheckMesheryctlClientVersion(build)
-		client := &http.Client{}
+		client := &http.Client{Timeout: 10 * time.Second}
 		resp, err := client.Do(req)
-
 		if err != nil {
+			if resp != nil {
+				_ = resp.Body.Close()
```
banana-three-join: I believe resp.Body.Close() is needed. Line 146 already defers the body close with `defer func() { _ = resp.Body.Close() }()` and as far as I'm concerned, it's an overkill because most of the times resp is nil and `net/http` already handles it

banana-three-join (commented):
Is version the only place where the client may hang?

mesheryctl/internal/cli/root/config/config_test.go
D-source1602: Fixed in 783292d — hoisted MesheryFolder out of the loop as a const since it doesn't change per iteration.

mesheryctl/internal/cli/root/version.go
D-source1602: You're right — checked net/http's docs: resp is only non-nil here on a CheckRedirect failure, and even then the body's already closed by the stdlib, so this check couldn't close anything meaningful. Also confirmed bodyclose isn't in Meshery's .golangci.yml, so nothing depends on keeping it. Removed in 783292d.

lekaf974 (commented):
@D-source1602 the PR title and changes don't seem to match!

alexquincy:

- `mesheryctl/internal/cli/root/config/config_test.go`
- `mesheryctl/internal/cli/root/version.go`
- `mesheryctl/pkg/utils/latest_version.go`

**What to do:**

D-source1602:
@lekaf974 
 Good catch, and sorry for the confusion! The scope did grow a bit during review — it started as just adding an HTTP timeout to the version check, but along the way reviewers suggested converting Run → RunE so failures actually surface through the exit code, plus a couple of related small fixes. I'll update the PR title to reflect that more accurately — thank you for flagging it!

mesheryctl/internal/cli/root/version.go
```diff
@@ -131,7 +132,7 @@ mesheryctl version
 		}
 
 		defer utils.CheckMesheryctlClientVersion(build)
-		client := &http.Client{}
+		client := &http.Client{Timeout: 10 * time.Second}
 		resp, err := client.Do(req)
```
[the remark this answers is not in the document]
lekaf974: @D-source1602 this comment seems to be legit

mesheryctl/internal/cli/root/config/config_test.go
```diff
@@ -35,16 +35,15 @@ func TestGetBuild(t *testing.T) {
 
 func TestGetLocation(t *testing.T) {
 	token := Token{}
+	const MesheryFolder = ".meshery"
```
lekaf974:
```suggestion
	const mesheryFolder = ".meshery"
```

this variable is scoped only within this method

lekaf974 (commented):
Small comment added LGTM to merged after

D-source1602:
@lekaf974 Fixed in 273ef1542638 — renamed to mesheryFolder.

banana-three-join:
Could you please squash your commits?

D-source1602:
@banana-three-join Squashed to a single commit as requested. PR now has 1 commit touching only the 3 relevant files. Ready for re-review.

banana-three-join (changes requested):
The two timeouts and the `t.Fatalf` change are correct, and the `resp.Body` reasoning in the new comment matches the `net/http` docs — `Client.Do` only returns a non-nil `Response` alongside a non-nil error when `CheckRedirect` fails, and the body is already closed in that case. No objection to the three files as written.

Following up on my earlier question about whether `version` is the only place the client can hang: it isn't, and one of the remaining cases is inside this command.

### `mesheryctl version` can still hang

`versionCmd.PreRunE` calls `currCtx.ValidateVersion()`, which issues an unbounded request to `github.com` before `Run` is ever reached:

```go
// mesheryctl/internal/cli/root/config/config.go:259
url := "https://github.com/" + ... + "/releases/tag/" + ctx.Version
req, err := http.NewRequest("GET", url, nil)
...
client := &http.Client{}
resp, err := client.Do(req)
```

This only fires when the current context pins a version (`ctx.Version` is neither empty nor `latest`), but that's an ordinary configuration. In that case `mesheryctl version` hangs exactly as #19943 describes — on the pre-run path, before any of the code this PR touches executes. It's a one-line change and I'd like it in this PR, since otherwise the command named in the title still hangs.

While you're there, please lift `10 * time.Second` into a single constant in `pkg/utils` rather than repeating the literal a third time. @Maanvi212006 already raised the hardcoding as a nit, and one definition is easier to revisit later.

### Please tighten the description

Whether or not the above lands, the description needs a scope correction. As written it reads as though hanging version checks are now resolved, and the auto-generated CodeRabbit summary underneath it claims "unsuccessful network responses are handled and cleaned up reliably" and "strengthened configuration path validation" — neither of which is in this diff. No status-code check was added, and no path validation changed. A future reviewer skimming the description will believe more was fixed than actually was.

Please state plainly what this PR does and does not cover: it bounds the server version request in `version.go` and the update-check request in `GetLatestVersionForMesheryctl`, and it does not address the other timeout-less `http.Client{}` instances elsewhere in `mesheryctl`.

One other description fix: it says the `Run` → `RunE` change was "split into its own PR: #" — the number is missing and I can't find the PR. That change addresses a real bug (`mesheryctl version` exits 0 even when the version check fails outright). Either link it or drop the sentence, so it isn't recorded as handled when it isn't.

D-source1602:
@banana-three-join  All three points from your review addressed in the latest push:

ValidateVersion() in config.go now has a 10s timeout — the pre-run hang is fixed
MesheryctlHTTPTimeout constant defined in pkg/utils/latest_version.go and used in version.go and latest_version.go; config.go uses the literal directly to avoid an import cycle
PR description updated to accurately reflect scope

Build, vet, gofmt, and tests all pass locally. Ready for re-review.

mesheryctl/internal/cli/root/config/config_test.go
```diff
@@ -35,16 +35,15 @@ func TestGetBuild(t *testing.T) {
 
 func TestGetLocation(t *testing.T) {
 	token := Token{}
+	const mesheryFolder = ".meshery"
```
banana-three-join: Could you declare it at the top level as a var to enhance the readiability?

banana-three-join (changes requested):
You can drop my advice and just keep the `Timeout: 10 * time.Second`. Only plausible way to add it as a global variable is to add it into constants.go and each request might have a different timeout so having one `MesheryctlHTTPTimeout` in constants might be misleading.

PARTH-TUSSLE (approved):
LGTM 👍
The timeout is now applied to the pre-run version check as well, and the other review feedback has been addressed. Tests and checks are passing.
Approving.

D-source1602:
@banana-three-join Addressed — MesheryctlHTTPTimeout constant removed, Timeout: 10 * time.Second inlined directly in both version.go and latest_version.go. Also moved mesheryFolder to package-level const in config_test.go. Ready for re-review.

banana-three-join (approved):

banana-three-join:
LGTM. Ty! 
image

PR #20442, opened 2026-07-03
[Server] Guard control-plane container index to prevent panic on empty pod spec

### LINKED ISSUE #20441: [Server] GetControlPlaneState panics on a container-less pod spec, crashing the control-plane subscription
### Current Behavior

`GetControlPlaneState` (`server/internal/graphql/model/control_plane_helper.go`) reads control-plane pod specs delivered by meshsync and extracts a version from the first container:

```go
if len(strings.Split(objspec.Containers[0].Image, ":")) > 1 {
    version = strings.Split(objspec.Containers[0].Image, ":")[1]
}
```

`objspec.Containers[0]` is indexed without checking the slice length. If a meshsync object's spec unmarshals to a pod with no containers (`objspec.Containers` empty), this panics with `index out of range [0] with length 0`.

`GetControlPlaneState` is called by `getControlPlanes`, which the control-plane subscription runs from a goroutine in `listenToControlPlaneState` (`server/internal/graphql/resolver/control-plane.go`) on every meshsync update:

```go
go func(ctxID string) {
    for {
        select {
        case <-r.MeshSyncChannelPerK8sContext[ctxID]:
            status, err := r.getControlPlanes(ctx, provider, filter)
            ...
```

That goroutine has no `recover()`, so the panic crashes the whole meshery-server process instead of failing the one subscription.

### Expected Behavior

A pod spec with no containers should be skipped (version stays `unknown`), not panic.

### Steps to Reproduce

1. Subscribe to control-plane state (the ServiceMesh/control-plane view in the UI opens this GraphQL subscription).
2. Have meshsync deliver a control-plane object whose spec unmarshals with an empty `Containers` list.
3. On the next meshsync update the subscription goroutine reaches `objspec.Containers[0]` and the server panics.

### Environment

- Meshery Server: master

### Notes

Fix guards the container length before indexing and moves the tag parsing into a small `imageTag` helper. PR incoming.

mightbeanshuu:
**Notes for Reviewers**

- This PR fixes #20441

`GetControlPlaneState` indexed `objspec.Containers[0]` without a length check while extracting a control-plane version from a meshsync-delivered pod spec. If that spec unmarshals to a pod with no containers, the index panics with `index out of range [0] with length 0`. It runs in the control-plane subscription goroutine (`listenToControlPlaneState`), which has no `recover()`, so the panic crashed the whole server rather than failing the one subscription.

**What changed**

- Guard `len(objspec.Containers) > 0` before indexing `Containers[0]`.
- Move the tag parsing into a small `imageTag` helper (same behavior as the previous `strings.Split(image, ":")[1]`, kept so registry/version semantics are unchanged) and add a unit test for it.

**Testing**

- `go test ./server/internal/graphql/model/`, `go vet`, and `gofmt` all pass.

### DIFF
--- a/server/internal/graphql/model/control_plane_helper.go
+++ b/server/internal/graphql/model/control_plane_helper.go
@@ -55,8 +55,10 @@ func GetControlPlaneState(ctx context.Context, selectors []MeshType, provider mo
 					continue
 				}
 
-				if len(strings.Split(objspec.Containers[0].Image, ":")) > 1 {
-					version = strings.Split(objspec.Containers[0].Image, ":")[1]
+				if len(objspec.Containers) > 0 {
+					if tag := imageTag(objspec.Containers[0].Image); tag != "" {
+						version = tag
+					}
 				}
 
 				members = append(members, &ControlPlaneMember{
@@ -82,4 +84,14 @@ func haveCommonElements(a []string, b map[string]bool) bool {
 		}
 	}
 	return false
+}
+
+// imageTag returns the tag portion of a container image reference
+// (the text after the first ":"), or "" when the reference carries no tag.
+func imageTag(image string) string {
+	parts := strings.Split(image, ":")
+	if len(parts) > 1 {
+		return parts[1]
+	}
+	return ""
 }
--- a/server/internal/graphql/model/control_plane_helper.go
+++ b/server/internal/graphql/model/control_plane_helper.go
@@ -86,12 +86,18 @@ func haveCommonElements(a []string, b map[string]bool) bool {
 	return false
 }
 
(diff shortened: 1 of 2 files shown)


mightbeanshuu:
Good catch on the registry port case. Updated imageTag to take the tag after the last colon that comes after the last slash, and to strip an @digest suffix first, so localhost:5000/istio/proxyv2:1.20.0 resolves to 1.20.0 and localhost:5000/istio/proxyv2 resolves to no tag. Fixed the doc comment to match and added those cases to the test. Ready for another look whenever, thanks!

KatalKavya96 (approved):
@mightbeanshuu LGTM!

PR #20444, opened 2026-07-03
[mesheryctl] Return an error instead of panicking on a non-standard deployment image

### LINKED ISSUE #20443: [mesheryctl] GetDeploymentVersion panics on a non-standard deployment image (index out of range)
### Current Behavior

`GetDeploymentVersion` (`mesheryctl/pkg/utils/platform.go`) returns `(string, error)` and reports read/unmarshal failures as errors, but then extracts the version from the container image with three unchecked slice indexes:

```go
image := compose.Spec.Template.Spec.Containers[0].Image
spliter := strings.Split(image, ":")
version := strings.Split(spliter[1], "-")[1]
```

Each of these panics with `index out of range` on a non-standard deployment image:

- `Containers[0]` panics if the deployment has no containers.
- `spliter[1]` panics if the image has no tag, e.g. `layer5/meshery`.
- `strings.Split(spliter[1], "-")[1]` panics if the tag has no dash, e.g. `layer5/meshery:latest`.

So a user whose `meshery-deployment.yaml` uses a customized image tag gets a CLI panic with a stack trace instead of a clean error, even though the function is written to return errors.

### Expected Behavior

A non-standard image should produce a returned error, not a panic.

### Steps to Reproduce

1. Point mesheryctl at a `meshery-deployment.yaml` whose container image is untagged or uses a tag without a dash (e.g. `layer5/meshery:latest`).
2. Run the command that resolves the deployment version (e.g. `mesheryctl system` version resolution).
3. The CLI panics with `index out of range`.

### Environment

- mesheryctl: master

### Notes

Fix guards each index and returns a descriptive error. PR incoming.

mightbeanshuu:
**Notes for Reviewers**

- This PR fixes #20443

`GetDeploymentVersion` returns `(string, error)` and already surfaces read/unmarshal failures as errors, but parsed the container image with three unchecked slice indexes (`Containers[0]`, the `:` split `[1]`, and the `-` split `[1]`). A deployment with no containers, an untagged image (`layer5/meshery`), or a tag without a dash (`layer5/meshery:latest`) panicked the CLI with `index out of range` instead of returning an error.

**What changed**

- Guard each index and return a descriptive error, matching how the rest of the function already reports failures.
- Added a table test covering the shapes that used to panic (untagged, dashless tag, no containers) plus a valid image.

**Testing**

- `go test ./mesheryctl/pkg/utils/`, `go vet`, and `gofmt` all pass.

### DIFF
--- a/mesheryctl/pkg/utils/platform.go
+++ b/mesheryctl/pkg/utils/platform.go
@@ -249,9 +249,20 @@ func GetDeploymentVersion(filePath string) (string, error) {
 		return "", fmt.Errorf("unable to unmarshal config %s | %s", MesheryDeployment, err)
 	}
 
-	image := compose.Spec.Template.Spec.Containers[0].Image
+	containers := compose.Spec.Template.Spec.Containers
+	if len(containers) == 0 {
+		return "", fmt.Errorf("unable to determine version: no containers found in %s", MesheryDeployment)
+	}
+	image := containers[0].Image
 	spliter := strings.Split(image, ":")
-	version := strings.Split(spliter[1], "-")[1]
+	if len(spliter) < 2 {
+		return "", fmt.Errorf("unable to determine version: image %q in %s has no tag", image, MesheryDeployment)
+	}
+	tag := strings.Split(spliter[1], "-")
+	if len(tag) < 2 {
+		return "", fmt.Errorf("unable to determine version: unexpected image tag %q in %s", spliter[1], MesheryDeployment)
+	}
+	version := tag[1]
 
 	return version, nil
 }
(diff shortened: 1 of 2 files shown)


mightbeanshuu:
Thanks for flagging the registry port case. Pulled the tag parsing into a small helper that ignores the registry port and any @digest, then take the version as everything after the first dash so multi-dash prerelease tags like stable-v0.7.0-rc.1 stay intact. Added tests for a registry port image and a prerelease tag. Ready for review, thanks!

KatalKavya96 (approved):
@mightbeanshuu Looks good now.!

The updated helper handles the non-standard image cases without panicking, covers registry ports/digests and multi-dash tags, and the added tests cover the reported edge cases.

Thanks for addressing the review feedback!

PR #20448, opened 2026-07-03
[mesheryctl] Guard empty design response in apply to avoid index out of range

### LINKED ISSUE #20447: [mesheryctl] design apply panics with index out of range on an empty response
### Current Behavior

In `mesheryctl design apply`, the remote-hosted design branch (`mesheryctl/internal/cli/root/design/apply.go`) unmarshals the API response into a slice and then indexes `response[0]` with no length check:

```go
err = json.Unmarshal(body, &response)
if err != nil {
    return utils.ErrUnmarshal(err)
}

// setup pattern file here
designFile = response[0].PatternFile
```

If the response is empty (the design was not found), `response[0]` panics with `index out of range [0] with length 0` and the CLI crashes with a stack trace instead of a clean "design not found" message.

Every sibling design command already guards this exact access: `deploy`, `undeploy`, `delete`, `export`, and the other branch of `apply` itself all check the length first and return `ErrDesignNotFound`. This one branch missed the guard.

### Expected Behavior

An empty response should return `ErrDesignNotFound`, matching the other design commands, not panic.

### Steps to Reproduce

1. Run `mesheryctl design apply` against a remote/hosted design reference that resolves to no pattern on the server.
2. The CLI panics with `index out of range` instead of reporting the design was not found.

### Environment

- mesheryctl: master

### Notes

Fix adds the same `len(response) == 0` guard the sibling commands use. PR incoming.

mightbeanshuu:
**Notes for Reviewers**

- This PR fixes #20447

The remote-hosted branch of `mesheryctl design apply` unmarshalled the API response into a slice and then indexed `response[0]` without a length check, so an empty response (design not found) panicked the CLI with `index out of range`. Every sibling design command (`deploy`, `undeploy`, `delete`, `export`, and the other `apply` branch) already guards this with a `len` check that returns `ErrDesignNotFound`; this branch was the one that missed it.

**What changed**

- Add `if len(response) == 0 { return ErrDesignNotFound(file) }` before indexing `response[0]`, matching the existing pattern in the sibling commands.

**Testing**

- `go vet` and `gofmt` pass. The change mirrors the guard already used by the other design subcommands.

### DIFF
--- a/mesheryctl/internal/cli/root/design/apply.go
+++ b/mesheryctl/internal/cli/root/design/apply.go
@@ -212,7 +212,7 @@ mesheryctl design apply [design-name]
 					return utils.ErrUnmarshal(err)
 				}
 
-				if len(response) == 0 {
+				if len(response) == 0 || response[0] == nil {
 					return ErrDesignNotFound(file)
 				}
 
--- a/mesheryctl/internal/cli/root/design/apply.go
+++ b/mesheryctl/internal/cli/root/design/apply.go
@@ -212,6 +212,10 @@ mesheryctl design apply [design-name]
 					return utils.ErrUnmarshal(err)
 				}
 
+				if len(response) == 0 {
+					return ErrDesignNotFound(file)
+				}
+
 				// setup pattern file here
 				designFile = response[0].PatternFile
 			}


mightbeanshuu:
Addressed the nil pointer note: the response is a slice of pointers, so I added a nil check next to the length guard and an empty or nil first element now returns design not found instead of dereferencing nil. On the regression test for the remote GitHub URL path, that branch needs mocking the /api/pattern POST which apply_test.go does not set up today, so I kept this change minimal and can add that test as a follow-up if you would prefer it here. Thanks for the review!

leecalcote (approved):

leecalcote:
Thanks!

PR #20497, opened 2026-07-05
[Server] Safely parse operator image versions

### LINKED ISSUE #20496: [Server] Image parsing panics and incorrect extraction on untagged or registry-port images
### Description
The server relies on simple `strings.Split(image, ":")[1]` to extract image versions/tags in multiple places. This causes panics or incorrect behavior when:
1. The image is untagged (e.g. `layer5/meshery-operator`) — `strings.Split` returns a slice of length 1, causing an index out of bounds panic if not guarded.
2. The image includes a registry port (e.g. `registry.local:5000/layer5/meshery-operator:v0.1.0`) — splitting by `:` yields 3 elements, and index `[1]` grabs `5000/layer5/meshery-operator` instead of the tag `v0.1.0`.
3. The image uses a digest (e.g. `layer5/meshery-operator@sha256:...`) — similar issues.

### Locations
- `server/internal/graphql/model/operator_helper.go`: `GetOperator` uses `strings.Split(container.Image, ":")[1]` without a length guard, which will panic on untagged images.
- `server/internal/graphql/model/control_plane_helper.go`: `GetControlPlaneState` has a length guard but still fails on registry ports (extracting `5000/image` instead of the tag). It also extracts the wrong org for images without a domain name.

### Expected Behavior
A robust helper should be used to parse image tags, safely ignoring digests and registry ports, defaulting to `latest` or `unknown` when no tag is present.

mightbeanshuu:
**Notes for Reviewers**

- This PR fixes #20496

It introduces `ParseOperatorImageVersion` to prevent index out of bounds panics on untagged images and correctly extract versions when the image is prefixed with a registry port or contains a digest.

### DIFF
--- a/server/internal/graphql/model/control_plane_helper.go
+++ b/server/internal/graphql/model/control_plane_helper.go
@@ -49,15 +49,12 @@ func GetControlPlaneState(ctx context.Context, selectors []MeshType, provider mo
 						imageOrgs[strings.Split(c.Image, "/")[1]] = true // Extracting image org from <domainname>/<imageorg>/<imagename>
 					}
 				}
-				version := "unknown"
 				//If image orgs are not passed on in from controlPlaneImageOrgs variable, then skip this filtering (for backward compatibility)
 				if len(controlPlaneImageOrgs[MeshType(selector)]) != 0 && !haveCommonElements(controlPlaneImageOrgs[MeshType(selector)], imageOrgs) {
 					continue
 				}
 
-				if len(strings.Split(objspec.Containers[0].Image, ":")) > 1 {
-					version = strings.Split(objspec.Containers[0].Image, ":")[1]
-				}
+				version := ParseOperatorImageVersion(objspec.Containers[0].Image)
 
 				members = append(members, &ControlPlaneMember{
 					Name:      obj.KubernetesResourceMeta.Name,
--- a/.github/SECURITY-INSIGHTS.yml
+++ b/.github/SECURITY-INSIGHTS.yml
@@ -1,7 +1,7 @@
 header:
   schema-version: 2.0.0
-  last-updated: '2026-03-03'
-  last-reviewed: '2026-03-03'
+  last-updated: '2026-07-04'
+  last-reviewed: '2026-07-04'
   url: https://github.com/meshery/meshery/blob/master/SECURITY-INSIGHTS.yml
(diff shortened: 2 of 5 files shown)


KatalKavya96 (changes requested):
@mightbeanshuu Thanks for working on this. The new `ParseOperatorImageVersion` helper and tests are a good direction, but a few items still need to be addressed

- Please handle the empty tag case like `layer5/meshery-operator:` so it falls back to `latest`.
- Please add a test case for that empty-tag scenario.
- The issue also mentions `server/internal/graphql/model/control_plane_helper.go`; if it still uses `strings.Split(image, ":")[1]`, this helper should be reused there too so the reported bug is fully fixed.
- Please also guard the `dep.Name` return path in case `dep` is nil after a not-found deployment lookup.

mightbeanshuu:
all 4 addressed, thanks for the thorough look:

- empty-tag case (layer5/meshery-operator:) now falls back to latest instead of an empty string, added a test for it
- control_plane_helper.go's GetControlPlaneState now calls ParseOperatorImageVersion instead of its own raw split, so it doesn't misread a registry port as the version anymore
- GetOperator now guards dep before dereferencing dep.Name

ran go build/vet/test on the package locally, all green.

KatalKavya96 (approved):
LGTM!

server/internal/graphql/model/control_plane_helper.go
```diff
@@ -49,15 +49,12 @@ func GetControlPlaneState(ctx context.Context, selectors []MeshType, provider mo
...
-				version := "unknown"
 				//If image orgs are not passed on in from controlPlaneImageOrgs variable, then skip this filtering (for backward compatibility)
 				if len(controlPlaneImageOrgs[MeshType(selector)]) != 0 && !haveCommonElements(controlPlaneImageOrgs[MeshType(selector)], imageOrgs) {
 					continue
 				}
 
-				if len(strings.Split(objspec.Containers[0].Image, ":")) > 1 {
-					version = strings.Split(objspec.Containers[0].Image, ":")[1]
-				}
+				version := ParseOperatorImageVersion(objspec.Containers[0].Image)
```
Sbragul26, line 57: Just checking, is changing the default version from `unknown` to `latest` for untagged images intentional?

PR #20518, opened 2026-07-07
Normalize model path output

yadavsidd:
**Notes for Reviewers**

This PR fixes 
-  This PR normalizes user-facing path output for `mesheryctl model init` and `mesheryctl model build`.
- On Windows, paths printed in CLI messages could use backslashes (`\`). This PR converts only the displayed paths to forward slashes (`/`) for consistency with existing CLI examples and golden outputs.
- Filesystem behavior is unchanged. The actual paths used for creating, reading, and building models still use the normal OS path handling.

BEFORE:
Created test-windows-path model at tmp\model-output\test-windows-path
Building meshery model from path tmp\model-output\test-windows-path\v0.1.0

AFTER:
Created test-windows-path model at tmp/model-output/test-windows-path
Building meshery model from path tmp/model-output/test-windows-path/v0.1.0


mesheryctl/internal/cli/root/model/init.go
```diff
@@ -271,6 +274,13 @@ var initModelData = []struct {
 	},
 }
 
+func modelDisplayPath(path string) string {
+	if path == "" {
+		return "."
+	}
+	return strings.ReplaceAll(filepath.Clean(path), `\`, "/")
+}
```
Sbragul26: Since `modelDisplayPath` is used in both `init.go` and `build.go` should we move it to a common place in this package?

yadavsidd: got it...can you review again

PARTH-TUSSLE (changes requested):
Hey @yadavsidd . Look into all the suggestions copilot has given , address them if needed and reject(if needed) with a valid reason.

PARTH-TUSSLE (changes requested):
I saw your commits and file changes, seems like you have addressed all the bot suggestions (address all of them if that's not the case).And then also reply in all of those conversations how you addressed that suggestion. And I have one more suggestion, Consider using filepath.ToSlash() inside the helper func since it's the standard library utility intended for this exact purpose.

PR #20527, opened 2026-07-07
[UI/App] Guard JSON.parse of sessionStorage keys against SyntaxError

### LINKED ISSUE #20526: [Bug] UI crashes on load when sessionStorage 'keys' contains malformed JSON or invalid types
## Description
The core application wrapper (`ui/pages/_app.tsx`) attempts to hydrate the user's RBAC permissions by executing a bare `JSON.parse()` on `sessionStorage.getItem('keys')`. Because `sessionStorage` is highly susceptible to external mutations (e.g., mid-write browser tab closures, malicious browser extensions, or lingering legacy schema data), this operation is inherently unsafe. 

If the cached value is corrupted (a SyntaxError) or is technically valid JSON but of the wrong type (like an object `{}` or `null` instead of an array), the application crashes during initialization. This prevents `loadAbility` from falling back to a network fetch, leaving the user permanently locked out of the UI with a broken state until they manually purge their browser storage.

## Steps to Reproduce
1. Log into Meshery locally.
2. Open Browser DevTools -> Application -> Session Storage.
3. Edit the `keys` entry. 
    - Scenario A (Syntax Error): Set the value to `undefined{broken]`
    - Scenario B (Type Error): Set the value to `{"some_object": true}`
    - Scenario C (Null Error): Set the value to `null`
4. Hard refresh the page (`Cmd/Ctrl + Shift + R`).
5. Notice that the UI either crashes entirely or silently fails to load capabilities because `updateAbility` executes on invalid data.

## Expected Behavior
The UI should treat a corrupted cache exactly like a missing cache. It should discard the corrupted entry and gracefully fall back to fetching fresh keys from `/api/identity/users/profile/keys` without crashing.

## Environment
- **Component:** `[UI]`
- **File:** `ui/pages/_app.tsx`

Kakumanu-Harshitha:
- This PR fixes #20526 

**Signed commits**

## Summary
Guards the hydration of RBAC keys from `sessionStorage` with strict runtime type verification and a safe fallback to the network API.

## Root Cause
In `_app.tsx`, the `loadAbility` callback trusted that `sessionStorage['keys']` was perfectly formed JSON *and* strictly an array. It lacked both a `try/catch` guard and runtime type verification. When this parse failed, or returned an unexpected type, the application crashed synchronously rather than gracefully degrading to fetch the data from the server.

## Changes
- Refactored `loadAbility` to parse `sessionStorage` safely inside a `try/catch` block.
- Implemented strict runtime validation (`Array.isArray(parsed)`) to ensure the cache contains a valid key array, guarding against corrupted objects or primitives.
- Purged the local cache (`sessionStorage.removeItem('keys')`) if the validation fails to prevent repeated crash loops.
- Modified the execution flow so that cache misses (both natural and due to corruption) correctly hit the `else` block to immediately re-fetch the user's keys from the server.

## Steps to Test
1. Log into Meshery locally.
2. Open Browser DevTools -> Application -> Session Storage.
3. Edit the `keys` entry. 
    - Scenario A (Syntax Error): Set the value to `undefined{broken]`
    - Scenario B (Type Error): Set the value to `{"some_object": true}`
    - Scenario C (Null Error): Set the value to `null`
4. Hard refresh the page (`Cmd/Ctrl + Shift + R`).
5. **Expected Result:** The application should not crash. The network tab should show a successful API call to fetch the keys, and the `sessionStorage` should be repopulated with a fresh, valid array.

## Verification
- [x] `npm run build` in `ui/` — passed (logically verified)
- [x] eslint — 0 issues (logically verified)
- [x] tests — E2E flows unaffected
- [x] DCO sign-off — all commits signed

### DIFF
--- a/ui/pages/_app.tsx
+++ b/ui/pages/_app.tsx
@@ -135,7 +135,6 @@ const MesheryApp = ({ Component, pageProps, relayEnvironment, emotionCache }) =>
     isOpen: false,
     relayEnvironment: createRelayEnvironment(),
     connectionMetadata: {},
-    keys: [],
     abilities: [],
     abilityUpdated: false,
   });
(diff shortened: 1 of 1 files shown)


Kakumanu-Harshitha:
@leecalcote  ptal! when you are free?

PR #20533, opened 2026-07-07
[UI] Unify the design "can edit" gate across card and table views

leecalcote:
## Unify the design "can edit" gate (card vs table)

Surfaced in review of #20528: the "can edit design" gate diverged between the two design views.

### Symptom
- **Card/grid** (`MesheryPatternCard`): `CAN(EDIT_DESIGN) || isOwner` — permission **OR** ownership.
- **Table** (`MesheryPatterns`): `CAN(EDIT_DESIGN) && isOwner` — permission **AND** ownership.

A non-owner holding the `EDIT_DESIGN` permission could see and use Edit in the card, but the table's `condition: userCanEdit(row)` (AND) **hid** the Edit action from them.

### Root cause
The gate logic was duplicated in two places with no shared definition, so they drifted.

### Fix
Extract one shared `canEditDesign(user, design)` helper (single source of truth) used by both views:
- **Model:** a user may edit if they hold `EDIT_DESIGN` **OR** they own the design.
- Keeps the truthy `user?.id` guard (no `undefined === undefined` owner match).
- **Enablement is unchanged** — each call site still gates click-through on `disabled={!CAN(EDIT_DESIGN)}`, so an owner without the permission sees the affordance but it stays disabled (visibility vs enablement).

### Tests
- New `design-permissions.test.ts` covering: permission-holder (non-owner), owner-without-permission, non-owner-without-permission, and unauthenticated / owner-less.
- Full UI vitest suite green.

### Edit-affordance behavior (both views, after)
| Case | Table before | Both views after |
|---|---|---|
| Owner + permission | shown / enabled | shown / enabled |
| Non-owner **with** `EDIT_DESIGN` | **hidden** | shown / enabled |
| Owner **without** permission | **hidden** | shown / disabled |
| Unauthenticated / owner-less | hidden | hidden |

Refs #20528.


fitzergerald:
@rishiraj38 you're working on permissions, right? Will you review?

ui/components/designs/patterns/MesheryPatternCard.tsx
rishiraj38: Should we allow owners to edit their designs regardless of global roles (meaning we should update the disabled gates), or is it expected that they see a disabled button if they don't have the global edit role?

rishiraj38 (dismissed):

ui/components/designs/patterns/MesheryPatternCard.tsx
rishiraj38: I just saw the PR description. In that case, the action is shown but disabled, so I assume that's the intended behavior.

rishiraj38 (approved):

PR #20539, opened 2026-07-08
ci: split docs preview build and deploy workflows

ianrwhitney:
This PR restructures the docs preview workflow into separate build and deployment workflows to improve security, reliability, and maintainability.

- Split the docs preview process into separate workflows:
  - `build-docs-preview.yml` handles building the docs preview in the PR context.
  - `deploy-docs-preview.yml` handles deployment from`workflow_run` context.

## Summary by CodeRabbit

- **New Features**
  - Added an automated documentation preview system for pull requests that builds a Hugo-based preview and makes it available for review.
  - Automatically deploys previews on successful builds and removes them when pull requests are closed.
  - Enforces preview retention and updates pull requests with a sticky link to the current preview.

- **Bug Fixes**
  - Improved preview reliability for concurrent updates using safer deployment/pruning behavior.
  - Added preview crawling protections to prevent search engines from indexing preview content.


victoriacheng15:
I like that separating the permissions for `build` (read-only) and `deploy` (write-allowed) is a great security choice. Plus, splitting them up could potentially reduce GitHub Actions runner time since the deploy step won't start at all if the build fails. 

However, I noticed a potential issue to check first before merging:

* **Upload Artifact on PR Close:** In `build-docs-preview.yml` (line 24), the job-level check `if: github.event.action != 'closed'` skips the entire build job when a PR is closed. Will this cause the metadata artifact upload to be skipped, making the download step in the cleanup workflow fail?

This split-workflow pattern is a solid approach and makes perfect sense to adopt across other repos too!

.github/workflows/build-docs-preview.yml
```diff
@@ -0,0 +1,108 @@
...
+  contents: read
+
+concurrency:
+  group: preview-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build-docs-preview:
+    runs-on: ubuntu-24.04
+    if: github.event.action != 'closed'
```
Katotodan, line 24: Since the build-docs-preview job is skipped when github.event.action == 'closed', is there any reason to keep closed in the workflow triggers? If preview cleanup is already handled by the dedicated cleanup workflow, we could probably remove closed from the trigger list to avoid invoking the workflow for an event that doesn't execute any jobs.
Same for line 106 which checked this condition again.

.github/workflows/build-docs-preview.yml
```diff
@@ -0,0 +1,108 @@
...
+          curl -fsSL -o dart-sass.tar.gz "https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-x64.tar.gz"
+          sudo tar -xzf dart-sass.tar.gz -C /usr/local/lib
+          sudo ln -sf /usr/local/lib/dart-sass/sass /usr/local/bin/sass
+          sudo ln -sf /usr/local/lib/dart-sass/sass /usr/local/bin/dart-sass
+          dart-sass --version
+
+      - name: Setup Node
+        uses: actions/setup-node@v6
+        with:
+          node-version: "22"
```
Katotodan, line 72: Can we consider adding Node dependency caching to speed up repeated preview builds?

.github/workflows/build-docs-preview.yml
ianrwhitney: Yea, i tried to simplify this condition. I feel like theres a case that when a PR closes that it cleans up vs lingering for another prune job but we could probably fold the cleanup concern into one. The main reason for this PR is to address the security concern and now failing wkflow d/t recent gh workflow changes.

ianrwhitney:
> * e download step in the cle

I dont think so. I am trying to skip that and just run the cleanup/removal of the rossjrw/pr-preview-action. Again some grey area on approach of the cleanup flow on close vs eventual prune. I dont have a strong opinion here but would like to get the workflow working and then further optimize this.

winkletinkle:
> … and removes them when pull requests are closed.
 😎

marblom007 (approved):

leecalcote (approved):
It's time to give these a go.

PR #20549, opened 2026-07-08
[UI/Dashboard] Guard all JSON.parse calls in table column renderers

### LINKED ISSUE #20548: [Bug]: Unhandled SyntaxError from JSON.parse in Dashboard Table Columns crashes the UI
## Description
When viewing the Dashboard resource tables (Workloads, Network, Storage, Nodes, Configuration), the entire React render tree can crash if MeshSync syncs a resource where specific spec or status attributes are malformed, null, undefined, or already parsed as an object. 

The application fails with an unhandled `SyntaxError: Unexpected token... in JSON at position...` originating from a bare `JSON.parse()` call inside a `customBodyRender` column definition.

## Root Cause
Across 17 different dashboard table configuration files (e.g., `pod-columns.tsx`, `deployment-columns.tsx`, `service-columns.tsx`), `customBodyRender` callbacks directly invoke `JSON.parse(val)` on cell values without a `try/catch` block. 

Because `customBodyRender` is synchronous, an unhandled exception thrown here bubbles up past the MUI-Datatables component and crashes the React component tree for the entire route.

The cell value (`val`) provided by the table might not be a valid JSON string for several reasons:
1. The backend returned a primitive (null, undefined).
2. The value is a pre-parsed JavaScript object rather than a stringified JSON payload.
3. The value is malformed string data.

## Expected Behavior
The table should render gracefully, displaying a fallback value (like `-` or an empty cell) if the underlying data cannot be parsed, instead of crashing the entire application.

## Steps to Reproduce
1. Navigate to the Dashboard.
2. Select any Kubernetes workload, network, or storage resource tab (e.g., Pods, Deployments).
3. If the backend delivers a raw object instead of a stringified payload for `spec.attribute` or `status.attribute`, or if it delivers `null`, the UI will crash immediately upon rendering the table.

## Impact
**High**. This is a fatal crash that takes down the primary dashboard view and prevents users from interacting with their resources.

## Proposed Solution
Introduce a centralized `safeJsonParse` utility that wraps `JSON.parse` with a `try/catch` and gracefully handles nullish values and pre-parsed objects. Replace all inline `JSON.parse` invocations within the dashboard table configurations with this utility.

Kakumanu-Harshitha:
- This PR fixes #20548 

**Signed commits**

## Summary
This PR prevents fatal UI crashes on the Dashboard resource tables caused by unhandled `SyntaxError` exceptions thrown from inline `JSON.parse` operations within MUI-Datatables column renderers.

## Root Cause
Every `customBodyRender` callback across the dashboard resource tables was calling `JSON.parse(val)` on cell values without a `try/catch`. If MeshSync delivered a null, undefined, pre-parsed object, or malformed string for any `spec.attribute` / `status.attribute` column, the unhandled exception crashed the entire React render tree — taking down the whole dashboard page.

## Changes
- **Added `safeJsonParse` Utility:** Created `ui/utils/json-parse.ts` to provide a robust, type-safe (no `any`) wrapper around `JSON.parse` that handles `null`, `undefined`, malformed strings, and pre-parsed objects gracefully with configurable fallback values.
- **Added Comprehensive Unit Tests:** Created `ui/utils/json-parse.test.ts` with 14 test cases ensuring robust edge-case handling.
- **Updated Dashboard Columns:** Replaced 46 inline `JSON.parse(val)` calls with `safeJsonParse(val)` across 17 resource configuration files spanning workloads, storage, network, configuration, and nodes.
- **Linter Alignment:** Changed `let` declarations for parsed results to `const` to satisfy `prefer-const` lint rules.

## Steps to Test
1. Navigate to the Dashboard.
2. Select any Kubernetes workload, network, or storage resource tab (e.g., Pods, Deployments).
3. Verify that the tables render correctly without crashing.
4. If testing locally with a mocked backend or proxy, artificially inject `null`, `undefined`, or a malformed string (`"{bad json"`) into any `spec.attribute` field for a resource and verify the table cell renders safely (empty or fallback) without a white-screen crash.

## Verification
- [x] `npm run build` — passed
- [x] `npm run lint` / `eslint` — 0 issues
- [x] `npx vitest run` — new utility tests pass
- [x] DCO sign-off — all commits signed
- [ ] CLI docs updated (N/A)

### DIFF
--- a/ui/components/dashboard/resources/configuration/configmap-columns.tsx
+++ b/ui/components/dashboard/resources/configuration/configmap-columns.tsx
@@ -7,6 +7,7 @@ import { DefaultTableCell, SortableTableCell } from '../sortable-table-cell';
 import { CONNECTION_KINDS } from '../../../../utils/Enum';
 import { getK8sContextFromClusterId } from '../../../../utils/multi-ctx';
 import { FormatId } from '@/components/data-formatter';
+import { safeJsonParse } from '../../../../utils/json-parse';
 
 export const buildConfigMapColumns = ({
   switchView,
@@ -84,7 +85,7 @@ export const buildConfigMapColumns = ({
         },
         customBodyRender: function CustomBody(val) {
           if (!val) return <>-</>;
-          const parseVal = JSON.parse(val);
+          const parseVal = safeJsonParse<Record<string, unknown>>(val);
           const keys = parseVal ? Object.keys(parseVal) : [];
           return <>{keys.join(', ')}</>;
         },
--- a/ui/components/dashboard/resources/configuration/horizontal-pod-autoscaler-columns.tsx
+++ b/ui/components/dashboard/resources/configuration/horizontal-pod-autoscaler-columns.tsx
@@ -85,7 +85,11 @@ export const buildHorizontalPodAutoscalerColumns = ({
           return <DefaultTableCell columnData={column} />;
         },
         customBodyRender: function CustomBody(val) {
-          const attribute = safeJsonParse<{ minReplicas?: number; maxReplicas?: number; currentReplicas?: number }>(val);
+          const attribute = safeJsonParse<{
+            minReplicas?: number;
+            maxReplicas?: number;
+            currentReplicas?: number;
+          }>(val);
           let minReplicas = attribute?.minReplicas;
           return <>{minReplicas}</>;
         },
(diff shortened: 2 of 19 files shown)


Kakumanu-Harshitha:
@leecalcote ptal! when you are free?

Kakumanu-Harshitha:
@leecalcote ptal! when you are free?

KhushamBansal (approved):
LGTM!

PR #20556, opened 2026-07-09
[Server] Guard patch_mutators_action against null mutator values

### LINKED ISSUE #20555: [Server] Rego patch_mutators_action writes null over the mutated field when the mutator path is missing, diverging from the Go engine
### Current Behavior

When an edge relationship's `mutatorRef` path does not resolve on the from-component (for example the field was never set, or was cleared mid-edit), the Rego engine's `patch_mutators_action` emits an `update_component_configuration` action that writes **null** over the mutated field on the to-component. The Go engine skips the patch for the same design, so the evaluation result depends on which engine is enabled.

Reproduced by querying both engines with the same input:

- relationship: `from.patch.mutatorRef = [["configuration","metadata","name"]]`, `to.patch.mutatedRef = [["configuration","metadata","namespace"]]`
- design: from-component whose `configuration.metadata` has no `name`; to-component with `configuration.metadata.namespace = "default"`

Rego (`data.eval_rules.patch_mutators_action`):

```
[{op: update_component_configuration, value: {id: to-1, path: [configuration, metadata, namespace], value: null}}]
```

Go (`patchMutatorsAction`): no actions.

So under the Rego engine the existing `"default"` value gets nulled out; under the Go engine it is preserved.

### Root cause

`core_utils.configuration_for_component_at_path` (utils.rego) returns `null` (not undefined) for a missing path, and `patch_mutators_action` (eval_rules.rego) only guards `old_value != mutatorValue`. With `mutatorValue = null` and a non-null `old_value`, the comprehension emits the null write.

The other Rego patch path already handles this: `apply_patch` in patch_helper_rules.rego guards `update_value != null`. The Go engine's `patchMutatorsAction` (server/policies/eval_rules.go) likewise skips nil mutator values. `eval_rules.patch_mutators_action` is the only patch flow missing the guard.

### Expected Behavior

A missing mutator value should not be propagated: no patch action is emitted, matching both the hierarchical Rego patch path and the Go engine, and no existing field value is destroyed.

### Environment

- Meshery Server, master (policies seeded from `models/meshery-core/0.7.2/v1.0.0/policies`)

mightbeanshuu:
**Notes for Reviewers**

- This PR fixes #20555

When an edge relationship's mutatorRef path does not resolve on the from-component, `configuration_for_component_at_path` returns null and `patch_mutators_action`'s only guard was `old_value != mutatorValue`. The Rego engine therefore emitted an `update_component_configuration` action writing null over the mutated field, while the Go engine (`patchMutatorsAction`) skips nil mutator values. Same design, different result depending on which engine is enabled, and the Rego side destroys an existing value.

The fix adds the same `mutatorValue != null` guard that the hierarchical patch path already has in `patch_helper_rules.rego` (`update_value != null`), so all three patch flows (hierarchical Rego, edge Rego, Go) now agree.

Regression coverage added to `TestRelationshipEvaluationScenarios`: the missing-mutator case (fails on master with the null-write action, passes with the guard) and a present-mutator positive control. Full `server/policies` suite passes locally, including the Rego/Go parity tests.

**Signed commits**

### DIFF
--- a/models/meshery-core/0.7.2/v1.0.0/policies/eval_rules.rego
+++ b/models/meshery-core/0.7.2/v1.0.0/policies/eval_rules.rego
@@ -56,6 +56,11 @@ patch_mutators_action(relationship, design_file) := {action |
 	mutatorValue := core_utils.configuration_for_component_at_path(mutatorRef, from_component, design_file)
 	old_value := core_utils.configuration_for_component_at_path(mutatedRef, to_component, design_file)
 
+	# A missing mutator path resolves to null; without this guard the action
+	# below writes null over the mutated field. The hierarchical patch path
+	# already guards this (patch_helper_rules.rego `update_value != null`).
+	mutatorValue != null
+
 	old_value != mutatorValue
 
 	action := {
--- a/server/policies/rego_test.go
+++ b/server/policies/rego_test.go
@@ -309,6 +309,123 @@ func TestRelationshipEvaluationScenarios(t *testing.T) {
 			designFile:  "namespace_parent_inline",
 			expected:    true,
 		},
+		{
+			// A missing mutator path resolves to null (utils.rego
+			// configuration_for_component_at_path defaults to null), and
+			// patch_mutators_action must not emit an update that writes
+			// null over the mutated field. The hierarchical patch path
+			// already guards this (patch_helper_rules.rego
+			// `update_value != null`); this pins the same behavior for the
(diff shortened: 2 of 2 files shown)


mightbeanshuu:
Hi @aabidsofi19 @leecalcote ! Ready for review when you have time. This guards patch_mutators_action against writing null over the mutated field, restoring Go/Rego parity (same family as the inventory-dedupe fix). Regression tests added, CI green. Thanks !

PR #20589, opened 2026-07-10
[mesheryctl] Fix `design deploy` double-encoding the search param (spaces in design name)

### LINKED ISSUE #20588: [mesheryctl] `design deploy` double-encodes the search param, so designs with spaces in their name are never found
### Current Behavior

`mesheryctl design deploy <name>` fails to find any design whose name contains a space (or any character that percent-encodes), even when the design exists. It exits with a misleading `design with name: "<name>" not found`.

Root cause is **double URL-encoding** of the `search` query parameter in `mesheryctl/internal/cli/root/design/deploy.go`:

```go
queryParams := url.Values{}
queryParams.Set("populate", "pattern_file")
queryParams.Set("search", url.QueryEscape(patternName)) // <-- pre-escapes
urlPath := fmt.Sprintf("%s?%s", patternURLPath, queryParams.Encode()) // <-- Encode() escapes AGAIN
```

`url.Values.Encode()` already percent-encodes each value. Pre-escaping with `url.QueryEscape` encodes it a second time. For a design named `My Design`:

- `url.QueryEscape("My Design")` → `My+Design`
- `Encode()` then escapes the `+` → `search=My%2BDesign`
- The server decodes this once and searches for the literal string `My+Design` (with a plus sign), which never matches `My Design`.

The lookup returns zero patterns and the command errors at `ErrDesignNotFound(patternName)`, so the failure looks like "the design doesn't exist" rather than a client encoding bug. Single-token names (no spaces/special chars) are unaffected, which is why this slips through.

The sibling commands encode correctly and can be used as the in-repo reference:
- `design/view.go` — `queryParams.Add("search", design)` then `Encode()` (single-encode, correct)
- `design/export.go` — `url.QueryEscape(name)` into a raw string, not fed back through `Encode()` (single-encode, correct)

Only `design/deploy.go` applies both `url.QueryEscape` and `url.Values.Encode()`.

### Desired Behavior

`mesheryctl design deploy "My Design"` should send `search=My+Design`, the server should match the design named `My Design`, and the deploy should proceed. Fix is to drop the redundant `url.QueryEscape` and let `url.Values.Encode()` do the single, correct encoding:

```go
queryParams.Set("search", patternName)
```

### Screenshots/Logs

Deterministic reproduction (Go `net/url`):

```
deploy.go (buggy):  populate=pattern_file&search=My%2BDesign   -> server sees "My+Design" -> not found
correct (single) :  populate=pattern_file&search=My+Design     -> server sees "My Design" -> found
```

### Environment

- Host OS: Mac Linux Windows
- Meshery Server Version: stable-v (reproduced on master @ 621130cf70)
- Meshery Client Version: stable-v (master)
- Platform: Docker Kubernetes

mightbeanshuu:
**Notes for Reviewers**

`mesheryctl design deploy <name>` double-encodes the `search` query parameter, so any design whose name contains a space (or other percent-encoded character) is never matched and the command exits with a misleading "design not found".

In `mesheryctl/internal/cli/root/design/deploy.go` the name was passed through `url.QueryEscape` **and then** `url.Values.Encode()`. `Encode()` already percent-encodes each value, so pre-escaping encodes it twice:

- `url.QueryEscape("My Design")` → `My+Design`
- `queryParams.Encode()` then escapes the `+` → `search=My%2BDesign`
- the server decodes this once and searches for the literal `My+Design`, which never matches `My Design`, so the lookup returns 0 patterns and hits `ErrDesignNotFound`.

Single-token names were unaffected (nothing to double-encode), which is why this went unnoticed.

The sibling commands already single-encode correctly and were used as the reference: `design/view.go` (`queryParams.Add("search", design)` + `Encode()`) and `design/export.go` (`url.QueryEscape` into a raw string, not re-`Encode()`d). Only `deploy.go` applied both.

**Fix:** drop the redundant `url.QueryEscape` and let `url.Values.Encode()` do the single, correct encoding. Added a `TestDeployCmd` case for a multi-word design name asserting the request is `search=multi+word+design` (fails on the old double-encoded path, passes now).

- This PR fixes #20588

### DIFF
--- a/mesheryctl/internal/cli/root/design/deploy.go
+++ b/mesheryctl/internal/cli/root/design/deploy.go
@@ -107,7 +107,11 @@ mesheryctl design deploy -f [filepath] -s [source type]
 
 			queryParams := url.Values{}
 			queryParams.Set("populate", "pattern_file")
-			queryParams.Set("search", url.QueryEscape(patternName))
+			// url.Values.Encode() already percent-encodes each value, so the
+			// name must be set raw here; pre-escaping with url.QueryEscape would
+			// double-encode it (e.g. "My Design" -> "search=My%2BDesign") and the
+			// server would never match the design.
+			queryParams.Set("search", patternName)
 			urlPath := fmt.Sprintf("%s?%s", patternURLPath, queryParams.Encode())
 			// search and fetch patterns with pattern-name
 			utils.Log.Debug("Fetching designs")
--- a/mesheryctl/internal/cli/root/design/deploy_test.go
+++ b/mesheryctl/internal/cli/root/design/deploy_test.go
@@ -127,6 +127,31 @@ func TestDeployCmd(t *testing.T) {
 			IsOutputGolden: false,
 			ExpectedError:  ErrDesignNotFound("nonexistent-design"),
 		},
+		{
+			// Regression for the double-encoded search param: a multi-word name
+			// must be sent as `search=multi+word+design`, not double-encoded to
+			// `search=multi%2Bword%2Bdesign`, otherwise the server never matches.
+			Name:             "given a design name with spaces when design deploy then the search param is single-encoded",
+			Args:             []string{"deploy", "multi", "word", "design"},
+			ExpectedResponse: "",
+			URLs: []utils.MockURL{
+				{
+					Method:       "GET",
+					URL:          testContext.BaseURL + "/api/pattern/types",
+					Response:     "view.designTypes.response.golden",
+					ResponseCode: 200,
+				},
+				{
+					Method:       "GET",
+					URL:          testContext.BaseURL + "/api/pattern?populate=pattern_file&search=multi+word+design",
+					Response:     "pattern.empty.response.golden",
+					ResponseCode: 200,
+				},
+			},
+			ExpectError:    true,
+			IsOutputGolden: false,
+			ExpectedError:  ErrDesignNotFound("multi word design"),
+		},
 	}
 
 	// Run tests


mightbeanshuu:
Hi @lekaf974 @alexquincy ! Small mesheryctl fix here, deploy double-encodes the search param so any design name with a space silently fails to deploy. +30/-1, CI green, both bots clean. Should be a quick review whenever you get a moment. Thank you !

mesheryctl/internal/cli/root/design/deploy.go
```diff
@@ -107,7 +107,11 @@ mesheryctl design deploy -f [filepath] -s [source type]
 
 			queryParams := url.Values{}
 			queryParams.Set("populate", "pattern_file")
-			queryParams.Set("search", url.QueryEscape(patternName))
+			// url.Values.Encode() already percent-encodes each value, so the
```
lekaf974, line 110: please remove the comment, it does not give any useful information, code itself is enough

mesheryctl/internal/cli/root/design/deploy_test.go
```diff
@@ -127,6 +127,31 @@ func TestDeployCmd(t *testing.T) {
...
+					ResponseCode: 200,
+				},
+				{
+					Method:       "GET",
+					URL:          testContext.BaseURL + "/api/pattern?populate=pattern_file&search=multi+word+design",
+					Response:     "pattern.empty.response.golden",
+					ResponseCode: 200,
+				},
+			},
+			ExpectError:    true,
```
lekaf974, line 151: I don't get your point in this test, if server return 200 why an error is raised

Maanvi212006:
no updates since Aug 11, hence closing. [this PR](https://github.com/meshery/meshery/pull/21847) is under review for fixing this issue

PR #20604, opened 2026-07-10
[mesheryctl] Fix design undeploy to undeploy instead of deleting or sending an invalid body

### LINKED ISSUE #20603: [mesheryctl] design undeploy deletes the saved design (by ID) and sends an undecodable body (by file) instead of undeploying
### Current Behavior

`mesheryctl design undeploy` is broken in both of its supported input modes.

**By ID** — `mesheryctl design undeploy <id>` does not undeploy anything. It calls `utils.DeleteConfiguration(..., "pattern")`, which issues `DELETE /api/pattern/{id}` and **deletes the saved design record**, then logs `design <id> deleted`. The deployed Kubernetes workloads are left running, and the saved design needed to manage/undeploy them later is gone.

```go
// mesheryctl/internal/cli/root/design/undeploy.go (current)
if isID {
    err := utils.DeleteConfiguration(mctlCfg.GetBaseMesheryURL(), design, "pattern")
    ...
    utils.Log.Infof("design %s deleted", args[0])
    return nil
}
```

**By file** — `mesheryctl design undeploy -f <file>` imports the file, then does:

```go
patternFile := response[0].PatternFile        // a string
patternFileByt, _ := yaml.Marshal(patternFile) // -> a YAML scalar, not a payload
req, _ = utils.NewRequest("DELETE", deployURL, bytes.NewBuffer(patternFileByt))
```

`DELETE /api/pattern/deploy` (`PatternFileHandler`) expects a JSON-encoded `models.MesheryPatternFileDeployPayload`. The raw YAML-marshaled string is neither wrapped in that payload nor JSON-encoded, so it fails to decode before undeployment begins. The sibling `design deploy` command demonstrates the correct contract.

### Expected Behavior

`mesheryctl design undeploy` undeploys the referenced design (whether referenced by name/ID or by file path) and **never deletes the saved design record**.

### Steps to Reproduce

1. Save and deploy a design.
2. Run `mesheryctl design undeploy <design-id>` → the saved design is deleted and the workloads remain (no undeploy).
3. Run `mesheryctl design undeploy -f design.yaml` → the undeploy request body fails to decode server-side.

### Environment

- mesheryctl: current `master`
- Applies to all platforms (CLI-only)

### Notes

The existing `undeploy_test.go` mocks the endpoints but never asserts the request body or the ID-mode behavior, so the breakage passes CI today.

mightbeanshuu:
## Notes for Reviewers

`mesheryctl design undeploy` was broken in both supported modes (details in #20603):

- **By ID** it called `DeleteConfiguration` → `DELETE /api/pattern/{id}`, **deleting the saved design** instead of undeploying, orphaning the running workloads.
- **By file** it `yaml.Marshal`-ed the design file string and sent the raw bytes to `DELETE /api/pattern/deploy`, which expects a JSON-encoded `models.MesheryPatternFileDeployPayload`, so the body failed to decode.

This PR resolves the referenced design to its file content in every case:

- **name/ID** → a read-only fetch (`GET /api/pattern/{id}` for an ID, or the `populate=pattern_file` search for a name), never a delete;
- **file** → import the provided file, as before;

and undeploys it through a single helper that uses the **same JSON payload contract as `design deploy`**. The saved design is never deleted.

The positional argument now behaves as the inverse of `design deploy` (name or ID). I kept ID resolution via `ValidId`. Happy to adjust if maintainers prefer name-only, to match `deploy`.

### Testing

- `go test --short ./internal/cli/root/design/...` green.
- Added a behavioral test: undeploy-by-name hits `DELETE /api/pattern/deploy` and **not** `DELETE /api/pattern/{id}` (the delete endpoint is deliberately unmocked, so a regression to the destructive path would fail with no responder).

This PR fixes #20603

---

- [x] I have read the [contribution guidelines](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md).
- [x] I have signed off all my commits as required by [DCO](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin).

### DIFF
--- a/mesheryctl/internal/cli/root/design/fixtures/undeploy.byname.patterns.response.golden
+++ b/mesheryctl/internal/cli/root/design/fixtures/undeploy.byname.patterns.response.golden
@@ -1,0 +1,1 @@
+{"page":0,"pageSize":10,"totalCount":1,"patterns":[{"id":"aaaaaaaa-bbbb-4ccc-8ddd-eeeeeeeeeeee","name":"TestDesign","patternFile":"name: TestDesign\n"}]}
--- a/mesheryctl/internal/cli/root/design/undeploy.go
+++ b/mesheryctl/internal/cli/root/design/undeploy.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 
 	"github.com/asaskevich/govalidator"
@@ -29,7 +30,6 @@ import (
 	"github.com/meshery/meshery/server/models"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"gopkg.in/yaml.v2"
 )
 
 var designUndeployCmd = &cobra.Command{
@@ -37,14 +37,17 @@ var designUndeployCmd = &cobra.Command{
 	Short: "Undeploy design",
 	Long:  `Undeploy design will trigger undeploy of design`,
 	Example: `
-// Undeploy design by providing file path
+// Undeploy a deployed design by its name or ID
+mesheryctl design undeploy [design name | ID]
+
+// Undeploy a design by providing a design file path
 mesheryctl design undeploy -f [filepath]
 	`,
 
 	Args: func(cmd *cobra.Command, args []string) error {
 		if len(args) == 0 && file == "" {
 			return ErrUndeployDesign(
-				fmt.Errorf("provide either a design ID or -f [filepath]"),
+				fmt.Errorf("provide either a design name/ID or -f [filepath]"),
 			)
 		}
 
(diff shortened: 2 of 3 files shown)


mightbeanshuu:
cc @lekaf974 for review whenever you have a moment. Fixes #20603, undeploy was deleting the saved design when given an ID, and sending an undecodable body when given a file, instead of undeploying in both cases. Behavioral test added, CI running. Thanks !

ishwar170695 (commented):
Since this PR introduces the ability to undeploy designs by name, what should happen if the search returns multiple designs with the same name? Currently we return `resp.Patterns[0]`, which silently targets the first match.

`design deploy` already prompts the user to select a design when multiple matches are found. Should `design undeploy` follow the same behavior as well to avoid unintentionally undeploying the wrong design?

PR #20641, opened 2026-07-13
[Server] Fix connection FSM get or create race creating multiple StateMachines

### LINKED ISSUE #20627: [Server] Connection FSM get or create race builds multiple StateMachine instances per connection, defeating the per machine lock
### Current Behavior

`InitializeMachineWithContext` (`server/machines/helpers/helpers.go:80-95`) performs a **non-atomic check-then-act**: `Get` -> (miss) -> `getMachine` -> `Start` (network I/O: clientset build + `AssignControllerHandlers`) -> `Add`. `ConnectionToStateMachineInstanceTracker` (`server/machines/sm_tracker.go`) locks `Get` and `Add` individually but **not the whole sequence**, so there is no atomic get-or-create.

`K8sFSMMiddleware` runs on **every `/api` request**, iterating all connected contexts and calling `InitializeMachineWithContext` -> `ResetState` -> `SendEvent` per context (`server/handlers/middlewares.go:388 418`). When a client fires concurrent requests (e.g. a browser opening several tabs, or parallel XHRs at startup), multiple goroutines hit a cache miss for the **same connection ID** simultaneously. Each builds a **distinct `*StateMachine`, each with its own `sm.mx`**; the last `Add` wins the map, but every copy escapes to its caller, which then drives it with `SendEvent`.

The result: the "**One FSM per connection**" serialization invariant documented at `server/machines/kubernetes/machine.go:17` and enforced via `sm.mx`  is void. Two goroutines drive the same connection through `Discovery -> Register -> Connect` while holding **different** locks, producing concurrent duplicate operator deploys (`DeployUndeployedOperators`) and racing `UpdateConnectionById` status writes for one connection.

Secondary defect: `Add` (`helpers.go:92`) runs **before** the `Start` error check (`helpers.go:93`), so a machine whose `Start` failed is cached and returned as valid on the next `Get`.

### Expected Behavior

Exactly one `StateMachine` exists per connection ID regardless of concurrent initialization, so all lifecycle events for a connection serialize through its single `sm.mx` (no duplicate operator deploys, no racing status writes). A machine that fails to initialize is not cached.

### Screenshots/Logs

Under concurrent `/api` load against one connection: interleaved `RegisterComponents` / operator deploy log lines and multiple `UpdateConnectionById` writes for the same `connectionID`; `go test  race` reports a data race on the deploy path.

### Environment

- **Host OS:** all
- **Meshery Server Version:** current `master`
- **Component:** `server/machines/helpers/helpers.go`, `server/machines/sm_tracker.go`

---

Priyanshu6968:
**Notes for Reviewers**

- This PR fixes #20627 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -75,6 +75,7 @@ require (
 	go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux v0.64.0
 	go.opentelemetry.io/otel/sdk v1.43.0
 	golang.org/x/oauth2 v0.36.0
+	golang.org/x/sync v0.21.0
 	golang.org/x/text v0.38.0
 	gonum.org/v1/gonum v0.17.0
 	google.golang.org/api v0.287.0
@@ -400,7 +401,6 @@ require (
 	golang.org/x/crypto/x509roots/fallback v0.0.0-20250406160420-959f8f3db0fb // indirect
 	golang.org/x/exp v0.0.0-20260410095643-746e56fc9e2f // indirect
 	golang.org/x/net v0.56.0 // indirect
-	golang.org/x/sync v0.21.0 // indirect
 	golang.org/x/sys v0.46.0 // indirect
 	golang.org/x/term v0.44.0 // indirect
 	golang.org/x/time v0.15.0 // indirect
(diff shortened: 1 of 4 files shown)


KatalKavya96 (changes requested):
@Priyanshu6968 The single flight approach looks good. Could you decouple inst.Start from the request-scoped context, such as with context.WithoutCancel(ctx), so one cancelled request does not fail all concurrent callers? Please also verify the fix with go test -race.

PR #20669, opened 2026-07-15
fix: add UUID validation for filter ID in filter handlers

manoj-1407:
## Problem

`GetMesheryFilterFileHandler`, `DeleteMesheryFilterHandler`, and `GetMesheryFilterHandler` all accept a raw filter ID from URL params and pass it directly to the provider without validating it as a valid UUID:

```go
filterID := mux.Vars(r)["id"]
resp, err := provider.GetMesheryFilter(r, filterID)
```

An invalid or malformed ID reaches the database layer, resulting in an unhandled error rather than a clean 400 response.

## Root Cause

The pattern handlers already handle this correctly using `uuid.FromStringOrNil`:

```go
if uuid.FromStringOrNil(patternID) == uuid.Nil {
    err := models.ErrInvalidUUID(...)
    writeMeshkitError(rw, err, http.StatusBadRequest)
    return
}
```

The filter handlers were missing this validation entirely.

## Fix

Added the same UUID validation to all three filter handlers, returning a proper `400 Bad Request` with `ErrInvalidUUID` for malformed filter IDs — consistent with `meshery_pattern_handler.go`.

## Handlers updated

- `GetMesheryFilterFileHandler`
- `DeleteMesheryFilterHandler`
- `GetMesheryFilterHandler`

## Summary by CodeRabbit

* **Bug Fixes**
  * Filter endpoints now validate IDs before processing requests.
  * Invalid or empty IDs return `400 Bad Request` with the appropriate error code.
  * Clone requests reject invalid IDs before parsing the request body.
* **Tests**
  * Added coverage for invalid and all-zero IDs across filter retrieval, deletion, file access, and cloning.


KatalKavya96 (changes requested):
@manoj-1407 Please link the corresponding issue in the PR description. Also, address Copilot’s feedback by rejecting uuid.Nil in all three handlers and adding handler tests for malformed and nil UUIDs, ensuring the provider method is not called.

manoj-1407:
@KatalKavya96 Done — linked issue #20698, added explicit uuid.Nil rejection alongside parse error handling, and added meshery_filter_handler_test.go covering malformed and nil UUID cases for all three handlers, asserting the provider is not called in either case.

manoj-1407:
Thanks @KatalKavya96 for the review! I've addressed it — also strengthened the tests to assert the specific ErrInvalidUUID error code (per CodeRabbit's suggestion) instead of just checking for 400, and confirmed UUID validation runs before body parsing in the clone handler. All CI checks are passing now. Let me know if anything else needs adjusting

manoj-1407:
Hi @revolyssup — this PR has been awaiting approval for a while with CI passing. Would appreciate a review when you have bandwidth.

manoj-1407:
@leecalcote all feedback addressed. would appreciate a look when you have time

PR #20689, opened 2026-07-16
[UI] Show Discovered At and fix invalid date on connections

chaitanyamedidar:
**Notes for Reviewers**

- This PR fixes #

1. **Invalid Date** - API returns camelCase (`createdAt` / `updatedAt` / `subType`); table columns still used snake_case. Bridged both spellings in `enhancedConnections` so cells get real values and server `ORDER BY` (`created_at`) still works.
2. **Default visibility**- `created_at` colView was `'na'` (never shown). Changed to `'m'` so **Discovered At** shows by default on medium+ widths; still toggleable via View Columns.
3. **Display** - switched from local `formatDate` to Sistent `FormattedTime` (same as MeshSync tab): relative time in-cell, full local datetime on hover.
4. **Empty/invalid dates** - column renderer shows `-` instead of `Invalid Date` when the value is missing or unparsable.
5. **Tests** - fixture uses camelCase API shape; regression covers bridge, default visibility, empty-date fallback, and `FormattedTime` render.

Screen Recording:

https://github.com/user-attachments/assets/7d3c3897-70f9-48f1-aa88-d4c0c6b75906


ui/components/connections/ConnectionTable.columns.tsx
```diff
@@ -1,14 +1,14 @@
 import React, { useMemo } from 'react';
 import {
-  CustomTooltip,
```
simihablo, line 3: Why is this customtooltip no longer used? Every column in this table should have an info button using a customtooltip to offer users additional details pertaining to each column and the meaning of its values.

ui/components/connections/ConnectionTable.tsx
```diff
@@ -295,7 +303,9 @@ const ConnectionTable = ({
       ['kind', 'm'],
       ['type', 's'],
       ['sub_type', 'na'],
-      ['created_at', 'na'],
+      // Show Discovered At by default from medium widths up; still toggleable
+      // via the View Columns control (was 'na' = never visible by default).
+      ['created_at', 'm'],
```
leecalcote: Good change — and the diagnosis behind it checks out. I traced it: `server/models/connections/connections.go:67` aliases `Connection` to `schemas/models/v1beta3/connection`, whose tags are `json:"createdAt"` / `json:"updatedAt"` / `json:"subType"`. So the wire really is camelCase here and the columns really were reading `undefined`. (Worth noting the deprecated `v1beta1` model is *mixed* — `json:"subType"` but `json:"created_at"` — so bridging both spellings is the right defensive call.)

One gap though: flipping this to `'m'` makes **Discovered At** user-facing by default, but its header still ships `icon={null} tooltip=""`. Every data column in this table should carry an info button (`CustomTooltip`) explaining what the column means and how to read its values — today only `environments` and `status` do. A column that's visible by default especially needs it, since "Discovered At" vs "Updated At" isn't self-evident.

Suggest adding the same `InfoOutlinedIcon` + `tooltip` treatment the `status` column uses. Since that block gets repeated per column, it's worth extracting a small `ColumnInfoIcon` helper rather than pasting the 18-line `IconButton` block nine times.

ui/components/connections/ConnectionTable.columns.tsx
```diff
@@ -350,13 +349,16 @@ export const useConnectionColumns = ({
...
-            return (
-              <CustomTooltip title={renderValue} placement="top" arrow interactive>
-                <span>{renderValue}</span>
-              </CustomTooltip>
-            );
+            if (value == null || value === '') {
+              return <span>-</span>;
+            }
+            const parsed = new Date(value);
+            if (Number.isNaN(parsed.getTime())) {
```
leecalcote: The `-` fallback is a real improvement, but it misses Go's zero-time sentinel.

`schemas/models/v1beta3/connection/connection.go:93` tags the field as ``CreatedAt core.Time `db:"created_at" json:"createdAt"` `` — **no `omitempty`** (and `omitempty` wouldn't elide a `time.Time` struct anyway). So a connection with an unset timestamp doesn't omit the field; it serialises as `"0001-01-01T00:00:00Z"`.

That string parses cleanly, so `Number.isNaN(parsed.getTime())` is `false` and it sails through to `FormattedTime`, which renders it as **"about 2025 years ago"** — the same class of nonsense date this PR is fixing.

```suggestion
            if (Number.isNaN(parsed.getTime()) || parsed.getUTCFullYear() <= 1) {
```

ui/components/connections/ConnectionTable.tsx
```diff
@@ -271,8 +271,16 @@ const ConnectionTable = ({
...
     // real connections. The only guard is against null/undefined array entries.
+    // Wire contract is camelCase (schemas v1beta3: createdAt, updatedAt,
+    // subType). Table columns and server ORDER BY still use the DB snake_case
+    // keys (created_at, etc.) so sort keeps working with SanitizeOrderInput.
+    // Bridge both spellings so the Discovered At cell is not undefined
+    // ("Invalid Date") when the API only sends camelCase.
     return connectionData.connections.filter(Boolean).map((connection) => ({
       ...connection,
+      created_at: connection.created_at ?? connection.createdAt,
+      updated_at: connection.updated_at ?? connection.updatedAt,
```
leecalcote: `updated_at` is bridged here, but the **Updated At** column (`ConnectionTable.columns.tsx`, ~line 314) has no `customBodyRender` at all — it only defines a `customHeadRender`. So now that this line feeds it real data, enabling Updated At via *View Columns* renders the raw wire string (`2026-05-09T12:00:00Z`) instead of the friendly relative time Discovered At gets. Before this bridge it rendered nothing, so the inconsistency is newly visible.

Since both columns want identical behaviour, it's worth lifting the null/invalid/zero guard + `FormattedTime` into one shared `renderTimestampCell(value)` and pointing both columns at it — that also keeps the zero-time fix in one place.

ui/components/connections/ConnectionTable.columns.tsx
chaitanyamedidar: Removing `CustomTooltip` was only from the unused-import lint after dropping the old date cell wrapper, not a product choice.
Addressed: Discovered At and Updated At headers now use a shared `ColumnInfoIcon` and a header tooltip (same pattern as Environments and Status). Full-table info icons for every column is out of scope for this PR; happy to follow up if we want that as a separate cleanup.

ui/components/connections/ConnectionTable.tsx
chaitanyamedidar: Agreed. Discovered At now has an info icon and tooltip explaining first-create vs MeshSync re-sync. Extracted `ColumnInfoIcon` and reused it for Environments, Status, Discovered At, and Updated At.

ui/components/connections/ConnectionTable.tsx
chaitanyamedidar: Done. Shared `renderTimestampCell` for both `createdAt` and `updatedAt` (relative time and full datetime on hover, same zero-time guard). Updated At stays opt-in via View Columns.

ui/components/connections/ConnectionTable.columns.tsx
chaitanyamedidar: `renderTimestampCell` now treats empty, unparsable, and zero-time (`getUTCFullYear() <= 1`) as `-`, so we never show about 2025 years ago. Covered in unit tests. Thanks.

simihablo:
Thanks, @chaitanyamedidar !

PR #20719, opened 2026-07-16
[Fix] Behavior of bulk selection in the Notifications Center 

NSTKrishna:
**Description**

This PR fixes the behavior of bulk selection ("Select All") in the Notifications Center to correctly account for filtered views and deleted notifications.

**Summary of changes:**

- **Context-aware Bulk Selection:** Updated `updateCheckAllEvents` to only apply the check toggle to events that are currently visible based on active filters (e.g., status: read/unread) and not deleted.
- **Accurate "Select All" Checkbox State:** Updated `selectAreAllEventsChecked` to evaluate the "checked" status strictly against visible events rather than all events in the store. This ensures the master checkbox correctly reflects the state of the current view.
- **Cleanup on Deletion:** Updated `deleteEvent` and `deleteEvents` actions to automatically uncheck events when they are marked as deleted.
- **Deduplication:** Extracted `isEventVisibleInView` helper function to standardise visibility checks across multiple selectors and reducers.
- **Tests:** Added test cases in `events.test.ts` to verify that deleted and filtered notifications are properly excluded from bulk selection operations.

**Before**

https://github.com/user-attachments/assets/dfa7bdc5-ef83-40f6-94c3-15ef0194c0a1

**After**

https://github.com/user-attachments/assets/497908f2-8161-43b2-bd37-e57e2943e8f1

**Notes for Reviewers**

- This PR fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bulk event selection so “checked” items respect the currently active view filters and ignore soft-deleted notifications.
  * Deleted events are now automatically unchecked.
  * “Select all” now reflects only events that are visible in the current view and not deleted.
  * When a view has no visible events, “all selected” no longer incorrectly evaluates to true.


ui/store/slices/events.ts
```diff
@@ -32,6 +32,19 @@ const defaultEventProperties = {
   status: STATUS.UNREAD,
 };
 
+/**
+ * Determines if an event should be visible in the current view based on active filters and deletion status.
+ * This function also governs the bulk selection ("Select All") behavior, ensuring that bulk selection
+ * applies only to notifications visible under the current filters and strictly excludes deleted notifications.
+ */
+const isEventVisibleInView = (event, currentFilters = {}) => {
```
ishwar170695, line 40: `isEventVisibleInView` currently only checks the `status` filter, but the Notification Center also supports filtering by `severity` (e.g. via the severity chips and filter drawer). Should `currentFilters.severity` be considered here as well so "Select All" only applies to notifications visible under the active severity filter?

PR #20794, opened 2026-07-19
[Server] Fix data race on the MeshSync data-handler lifecycle

### LINKED ISSUE #20763: Data race on ctxMeshsyncDataHandler: concurrent connect/config paths duplicate MeshSync ingestion and leak subscriptions
### Current Behavior

`MesheryControllersHelper` (`server/models/meshery_controllers.go:51`) holds the
whole MeshSync data handler lifecycle in two plain, unguarded fields
`ctxMeshsyncDataHandler *MeshsyncDataHandler` (`:62`) and the managed
`brokerPortForward` (`:73`). `AddMeshsyncDataHandlers` is an unsynchronized
check-then-act: `if ctxMeshsyncDataHandler == nil` (`:217`) -> `msDataHandler.Run()`
(`:266`) -> `ctxMeshsyncDataHandler = msDataHandler` (`:277`), and
`RemoveMeshSyncDataHandler` does `Stop()` + set-nil (`:623-627`). Only
`meshsyncConnectedEventEmitted` is atomic the comment at `:64-68` already notes
the method "can be invoked concurrently," but the handler pointer itself is left racy.

Three goroutines mutate the **same** per-connection helper
(`machineCtx.MesheryCtrlsHelper`) with no ordering and no shared lock:

- the connect action, in a detached goroutine `server/machines/kubernetes/connect.go:99` -> `AddMeshsyncDataHandlers`
- the MeshSync deployment mode reconcile handler `server/handlers/connection_action_handler.go:163` -> `RemoveMeshSyncDataHandler` (`:299`) + `AddMeshsyncDataHandlers` (`:307`)
- the controllers-config apply path, fanned across every tracked connection on a server wide defaults change -> `server/handlers/controllers_config_handlers.go:75` -> `:365` / `:366` (loop at `:401`)

The FSM's per machine lock is released as soon as `SendEvent` returns, and two of
these paths don't go through the FSM at all, so nothing serializes them. Two
callers can both observe `ctxMeshsyncDataHandler == nil` and each run
`NewMeshsyncDataHandler(...).Run()`: two live subscriptions to
`meshery.meshsync.core`, two sets of listener goroutines, two managed
port-forwards. Whoever writes the pointer last wins; the loser is orphaned and
never `Stop()`ped. Net effect: every MeshSync event gets persisted twice, and the
extra subscription + goroutines + port forward leak for the life of the process.
It's also a plain data race on the pointer, flagged under `-race`.

### Expected Behavior

The data-handler lifecycle should be single owner: at most one live
`MeshsyncDataHandler` (and one port forward) per connection, with the
nil-check/`Run`/assign and the `Stop`/clear sequences atomic no matter how many of
the three paths fire concurrently no duplicate ingestion, no orphaned
subscription/goroutine/port forward, and clean under the race detector.

### Environment

- **Host OS:** any
- **Platform:** Kubernetes
- **Meshery Client Version:** n/a

---

Priyanshu6968:
**Notes for Reviewers**

- This PR fixes #20763 

### DIFF
--- a/server/models/meshery_controllers.go
+++ b/server/models/meshery_controllers.go
@@ -58,6 +58,19 @@ type MesheryControllersHelper struct {
 	// operator status for a particular context
 	ctxOperatorStatus controllers.MesheryControllerStatus
 
+	// msMu serializes the MeshSync data-handler lifecycle for this context.
+	// AddMeshsyncDataHandlers (connect), the deployment-mode reconcile handler,
+	// and the controllers-config apply path can all fire concurrently on the same
+	// helper instance, and none of them share a lock otherwise: the connection
+	// FSM releases its lock the moment SendEvent returns, and two of those three
+	// paths bypass the FSM entirely. Without this, the nil-check/Run/assign below
+	// races the Stop/clear in RemoveMeshSyncDataHandler, so two callers can each
+	// start a handler — duplicate broker subscriptions and duplicate ingestion,
+	// with the losing handler orphaned and leaked. msMu guards ctxMeshsyncDataHandler
+	// and brokerPortForward; it is held across the whole add/remove so the
+	// check-then-act stays atomic.
+	msMu sync.Mutex
+
 	// meshsync data handler for a particular context
 	ctxMeshsyncDataHandler *MeshsyncDataHandler
 
@@ -121,6 +134,8 @@ func (mch *MesheryControllersHelper) GetControllerHandlersForEachContext() map[M
 }
 
 func (mch *MesheryControllersHelper) GetMeshSyncDataHandlersForEachContext() *MeshsyncDataHandler {
+	mch.msMu.Lock()
+	defer mch.msMu.Unlock()
 	return mch.ctxMeshsyncDataHandler
 }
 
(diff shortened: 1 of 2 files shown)


server/models/meshery_controllers_test.go
```diff
@@ -94,3 +96,97 @@ func TestAddCtxControllerHandlersReturnsEarlyOnInvalidConfig(t *testing.T) {
...
+	// Writers attach a handler (mirroring the field assign in AddMeshsyncDataHandlers,
+	// under msMu) and tear it down, over and over.
+	for w := 0; w < writers; w++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for i := 0; i < iterations; i++ {
+				mch.msMu.Lock()
+				mch.ctxMeshsyncDataHandler = &MeshsyncDataHandler{}
+				mch.msMu.Unlock()
```
ishwar170695, line 166: This test bypasses `AddMeshsyncDataHandlers()`, so it doesn't exercise the initialization path where the reported race originally occurred. 
Is there a way we could extend the regression coverage to include that path, or factor the lifecycle logic into a more easily testable helper?

server/models/meshery_controllers.go
```diff
@@ -121,6 +134,8 @@ func (mch *MesheryControllersHelper) GetControllerHandlersForEachContext() map[M
 }
 
 func (mch *MesheryControllersHelper) GetMeshSyncDataHandlersForEachContext() *MeshsyncDataHandler {
+	mch.msMu.Lock()
+	defer mch.msMu.Unlock()
```
ishwar170695, line 138: Since this accessor only reads `ctxMeshsyncDataHandler`, would `sync.RWMutex` with `RLock()`/`RUnlock()` be appropriate here? It could reduce contention if these accessors are called frequently.

PR #20798, opened 2026-07-20
[Server] Fix connected clusters showing no MeshSync data when kubernetesServerId is stale

marblom007:
## Notes for Reviewers

Fixes a case where a Kubernetes connection is **connected and actively streaming MeshSync data**, yet the dashboard shows none of that cluster's resources.

### Symptom
A user has multiple Kubernetes connections. At least one is connected and its cluster is streaming MeshSync data, but the Meshery dashboard (Overview honeycomb + resource tables) renders empty for it.

### Root cause
The dashboard filters MeshSync rows by `cluster_id` - a cluster's `kube-system` namespace UID, stamped on every row by the MeshSync agent. It obtains the id it queries with from each connection's persisted `metadata.kubernetesServerId`:

`connection.metadata.kubernetesServerId` -> `getK8sClusterIdsFromCtxId` -> `?clusterId=...` -> `WHERE cluster_id IN (...)`

When a connection is first persisted **before** its server ID is resolved - registered while its cluster's API server was unreachable, or migrated from an older Meshery that never stored it - `kubernetesServerId` is persisted empty (`""`) or as the nil UUID (`00000000-...`). It is then **never corrected** once the cluster becomes reachable and starts streaming:

- `SaveK8sContext` early-returns `ErrContextAlreadyPersisted` (it does not update an existing connection), and
- the FSM status-update path rewrites the *existing* metadata verbatim (`MetaData: connection.Metadata`).

So rows arrive stamped with the real `cluster_id`, while the dashboard queries `cluster_id IN ('' | 00000000-...)` and matches nothing. The GraphQL -> SSE/REST connection migration surfaced this: the old `subscribeK8sContext` computed `kubernetesServerId` from the live in-memory context; the new path reads it from persisted connection metadata.

### Fix
- **[Server]** `DiscoverAction` now reconciles the persisted `kubernetesServerId` with the id freshly resolved by `AssignServerID` whenever they differ (and persists the freshly-resolved context on first save). This runs on the per-request discovery re-drive, is a **no-op once corrected** (no steady-state write), and **self-heals existing broken connections** on their next reconcile. Wrapped in a MeshKit error (`meshery-server-1445`).
- **[UI]** Hardening: `getK8sClusterIdsFromCtxId`'s `all` branch dropped nothing, so an unresolved connection contributed an `undefined`/empty (junk) `clusterId` param. It now filters to truthy ids, matching the check the explicit-selection branch already applies.

### Watch for
- Two connections pointing at the **same** physical cluster resolve to the same `kube-system` UID (same `cluster_id`); MeshSync rows are keyed only by `cluster_id`, so they remain indistinguishable per-connection by design - unchanged here.
- A separate, already-tracked concurrency issue on the MeshSync data-handler lifecycle (`ctxMeshsyncDataHandler`) is out of scope for this PR (see #20763 / #20794).

### Testing
- `go test ./server/machines/kubernetes/` - added `discover_test.go` covering back-fill of empty and nil-UUID server ids, no-op when already correct, nil metadata, early-return guards, and error propagation.
- `vitest run ui/utils/__tests__/multi-ctx.test.ts` - added a case asserting unresolved server ids are dropped under `all`. All 38 pass.
- `make golangci` / `go vet` clean on the changed package; `eslint` + `prettier` clean on the changed UI files.

**[Signed commits]**

### DIFF
--- a/server/machines/kubernetes/discover.go
+++ b/server/machines/kubernetes/discover.go
@@ -5,10 +5,8 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/gofrs/uuid"
 	"github.com/meshery/meshery/server/machines"
 	"github.com/meshery/meshery/server/models"
-	"github.com/meshery/meshery/server/models/connections"
 	"github.com/meshery/meshkit/models/events"
 	"github.com/meshery/schemas/models/core"
 )
@@ -65,8 +63,9 @@ func (da *DiscoverAction) Execute(ctx context.Context, machineCtx interface{}, d
 		// now from the server ID just resolved from the reachable cluster so the
 		// dashboard's cluster_id filter matches the MeshSync data this cluster
 		// streams under its real ID; without this the connection ingests data yet
-		// the dashboard shows none.
-		if rerr := reconcilePersistedServerID(provider, token, k8sContext); rerr != nil {
+		// the dashboard shows none. The reconcile lives in the models layer
+		// alongside the rest of the connection/context persistence logic.
+		if rerr := models.ReconcileK8sContextServerID(provider, token, k8sContext); rerr != nil {
 			machinectx.log.Warn(rerr)
 		}
 	} else if err != nil {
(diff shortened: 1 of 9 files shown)


server/machines/kubernetes/discover.go
```diff
@@ -5,8 +5,10 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/gofrs/uuid"
 	"github.com/meshery/meshery/server/machines"
 	"github.com/meshery/meshery/server/models"
+	"github.com/meshery/meshery/server/models/connections"
```
leecalcote: This is a schema-driven project. Why isn’t the connection model being imported from meshery/schemas?

marblom007: Good call - moved the reconcile out of the FSM action into the models layer (models.ReconcileK8sContextServerID), so discover.go no longer imports the connection package. The record it reads is the schemas model: connections.Connection is a type alias for github.com/meshery/schemas/models/v1beta1/connection.Connection. The write goes through the token-string provider API UpdateConnectionById (the same path the connection state machine's own status update uses), which takes the server's ConnectionPayload - there is no token-string, schemas-native connection writer (UpdateConnection is request-scoped and an FSM action has no *http.Request). Unifying connections.ConnectionPayload itself onto the schemas payload is a cross-cutting change (provider interface + every provider and caller); happy to file that as a follow-up if you'd prefer it in this PR.

ritzorama (approved):

PR #20804, opened 2026-07-20
[UI] Fix Connections filter reset to All

### LINKED ISSUE #20762: [UI] Resetting Connection filters to "All" displays an empty state instead of restoring all connections
### Current Behavior

When filters are applied in the **Connections** table (e.g., **Status** or **Kind**) and later reset back to **All**, the table incorrectly displays a **"No connections found"** empty state instead of restoring the complete list of connections.

### Expected Behavior

Selecting **All** for any filter should clear the corresponding filter and display all available connections.

### Steps to Reproduce

1. Navigate to the **Connections** page.
2. Open the **Filters** menu.
3. Apply a filter (e.g., change **Status** or **Kind** from **All** to another value).
4. Apply the filter.
5. Change the filter back to **All**.
6. Click **Apply**.

### Actual Result

The Connections table displays an empty state (e.g., **"Sorry, no matching records found"**) even though connections exist.

### Expected Result

The table should restore and display all available connections when the filters are reset to **All**.

https://github.com/user-attachments/assets/78f7ebae-336b-4910-bcb3-0050e66ab6b1

### Environment

---

KumarNirupam1:
## Summary
Fixes resetting Connections table filters to **All** so the full list of connections is restored instead of showing an empty state ("Sorry, no matching records found").

**Root cause**
`UniversalFilter` calls `setSelectedFilters(next)` and then `handleApplyFilter(next)` in the same tick. `handleApplyFilter` ignored the argument and read React `selectedFilters` state, which was still the previous value (stale closure). Resetting to All therefore did not clear the applied status/kind used for the API/URL query. Because the previous filter value was applied instead of the latest one, selecting All did not clear the applied filter and the table continued to show filtered results.

**Changes**
- Use the filters payload passed by `UniversalFilter` on Apply when updating URL/table state.
- Treat `"All"` as an unset filter so status/kind are cleared from the connections query.
- Preserve the existing client-side filtering behavior while ensuring the latest filter values are applied correctly.
- Update the toolbar `handleApplyFilter` type to accept optional filters.
- Add a focused regression test in `ConnectionTable.filters.test.tsx`.

Fixes #20762
---

https://github.com/user-attachments/assets/2ab7d553-8c32-42dd-ab09-e5cc13f117f3

## Test plan
- [x] `npm test -- --run components/connections/ConnectionTable.test.tsx components/connections/ConnectionTable.filters.test.tsx`
- [ ] Manual: Lifecycle → Connections → apply Status or Kind filter → Apply → set filter back to **All** → Apply → all connections are shown again
- [ ] Confirm applying a non-All filter still narrows the table correctly

**Notes for Reviewers**
- This PR fixes #20762
- Core fix is in `ConnectionTable.tsx`, where `handleApplyFilter` now uses the filters payload from `UniversalFilter` to avoid the stale-closure issue. Toolbar change is type-only.
- Regression coverage is isolated in `ConnectionTable.filters.test.tsx` to avoid growing `ConnectionTable.test.tsx` past the 600-line ESLint soft limit.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved connection filtering for status and connection type selections.
  - Ensured resetting a filter to “All” clears it from table state and results.
  - Preserved displayed connection records when filters are reset.

- **Tests**
  - Added coverage verifying filter application and reset behavior, including removal of inactive filters from connection queries.


Maanvi212006:
@KumarNirupam1 This PR fixes the stale-closure bug in `ConnectionTable.tsx`'s `handleApplyFilter`, but `ui/components/connections/meshSync/index.tsx` has the identical pattern (zero-arg `handleApplyFilter` reading `selectedFilters` state, wired to the same `UniversalFilter` component) and looks like it has the same bug. Is that intentionally out of scope for this PR, or should it be fixed here too?

Maanvi212006:
@KumarNirupam1 The regression test is well-targeted and correctly reproduces the stale-closure interaction — nice work isolating that. One gap: there's no test exercising the `kind` filter reset independently (only `status` is driven through a UI click); given `kindFilter` uses the same normalization logic, could you add a parallel case, or confirm it's covered elsewhere?

fitzergerald:
@KhushamBansal are you reviewing?

KumarNirupam1:
@Maanvi212006 Thanks for taking a look , I’ve noted both of your points.

I need to look into a slightly different way to land this fix, based on what we discussed with Lenox in Monday’s meeting (keeping the existing filter state rather than removing it). I’ll also cover the kind filter reset in the test, and for MeshSync I’ll treat that as a follow-up unless maintainers want it in this PR.

Working on the update now will push once it’s ready.

KhushamBansal (dismissed):
@KumarNirupam1 I tested this locally. LGTM! 🚀

A few other tables have the same issue. Could you address those in a follow-up PR as well?

KumarNirupam1:
> @KumarNirupam1 I tested this locally. LGTM! 🚀
> 
> A few other tables have the same issue. Could you address those in a follow-up PR as well?

Okay @KhushamBansal I'll address those.

KumarNirupam1:
@Maanvi212006 @KhushamBansal  here’s the final update on this PR.

I’ve kept the existing filter state (as discussed in Monday’s meeting) instead of removing it, and fixed the Apply path so selecting All clears the filter to null/empty using UniversalFilter’s payload. That restores all connections correctly.

Also added a parallel kind filter reset test alongside the status one.

Could you please take another look when you get a chance? Thanks!

Maanvi212006 (changes requested):
Nice catch on the stale-closure fix — `handleApplyFilter` taking `filters` directly is the right call.

One gap though: the description says client-side re-filtering was removed, but `filteredConnections` (line 293) still filters by `selectedFilters`, and that's what gets rendered (line 669). `selectedFilters` only syncs on Apply-click, not on URL changes (back/forward, deep links), so this could still show a stale/empty result outside the click path.

ishwar170695 (commented):
Nice fix for the stale-closure issue. One thing I noticed while going through the changes: `filteredConnections `still seems to be derived from `selectedFilters`, while the API query is driven by `tableState.filters`.

From what I could tell, `selectedFilters `is only initialized from `tableState `and then updated through `UniversalFilter`, so I wasn't able to find anything keeping it in sync with URL-driven changes (like back/forward navigation or deep links). That seems like it could leave the client-side filtering using stale values even though the API has already fetched data for the updated URL.

Am I missing a synchronization path somewhere, or would it make sense for the client-side filtering to derive from `tableState.filters` instead?

KumarNirupam1:
> Nice catch on the stale-closure fix — `handleApplyFilter` taking `filters` directly is the right call.
> 
> One gap though: the description says client-side re-filtering was removed, but `filteredConnections` (line 293) still filters by `selectedFilters`, and that's what gets rendered (line 669). `selectedFilters` only syncs on Apply-click, not on URL changes (back/forward, deep links), so this could still show a stale/empty result outside the click path.

@Maanvi212006 you're right that filteredConnections is still there. In the latest revision I intentionally kept the existing client-side filtering, based on the discussion from Monday's meeting. The only behavior changed in this PR is the Apply path: handleApplyFilter now uses the filters payload from UniversalFilter, so selecting All correctly clears the applied filter in tableState.filters and updates the API query.

also I've updated the description to reflect the final implementation.

KumarNirupam1:
> Nice fix for the stale-closure issue. One thing I noticed while going through the changes: `filteredConnections `still seems to be derived from `selectedFilters`, while the API query is driven by `tableState.filters`.
> 
> From what I could tell, `selectedFilters `is only initialized from `tableState `and then updated through `UniversalFilter`, so I wasn't able to find anything keeping it in sync with URL-driven changes (like back/forward navigation or deep links). That seems like it could leave the client-side filtering using stale values even though the API has already fetched data for the updated URL.
> 
> Am I missing a synchronization path somewhere, or would it make sense for the client-side filtering to derive from `tableState.filters` instead?

The current implementation intentionally preserves the existing client-side filtering. This PR specifically addresses the stale-closure issue in the Apply flow by using the filters payload provided by UniversalFilter, ensuring the latest filter values are written to tableState.filters when Apply is clicked.

Regarding synchronization with URL-driven changes (such as back/forward navigation or deep links), that's outside the scope of this PR. The issue reported in #20762 was specifically caused by handleApplyFilter reading stale state during Apply, and this change resolves that behavior while keeping the existing filtering approach unchanged.

PR #20812, opened 2026-07-20
[Server] Stop controller-status poll from spamming meshkit-11180 on a nil machine context

marblom007:
## Notes for Reviewers

Fixes a recurring error-log spam from the controller-status path:

```
ERRO nil interface cannot be type casted | code=meshkit-11180
caller="…handlers.(*Handler).machineCtxForConnection controllers_status_handler.go:116"
```

### Symptom
`machineCtxForConnection` logs `meshkit-11180` ("nil interface cannot be type casted") at **Error** severity. Because the controller-status SSE subscription polls every ~5s, it repeats indefinitely.

### Root cause
`machineCtxForConnection` does `utils.Cast[*kubernetes.MachineCtx](inst.Context)` on a tracked FSM instance. `utils.Cast` returns `meshkit-11180` when handed a **nil interface**, and `inst.Context` is legitimately nil in two cases:

1. **Non-kubernetes connections** - registered with a `nil` init func (`connections_handlers.go`), so `StateMachine.Start` sets `Context = nil` by design.
2. **Kubernetes connections whose cluster was unreachable when the machine was created** - `AssignInitialCtx` → `AssignClientSetToContext` fails, `Start` returns early and never sets `Context`.

Neither is an error - both are an expected "no ready machine context" state - yet each poll type-asserted the nil and logged it as one.

A tracked instance with a nil `Context` is more reachable than it looks: `InitializeMachineWithContext` calls `smInstanceTracker.Add` **before** checking the `Start` error (#20820), so only the *first* call for a broken connection returns `(nil, err)`. Every later call takes the cache-hit path and gets that same half-built instance back paired with a `nil` error, and the tracker is only ever cleared by an explicit user action. So the state persists for the life of the process, and a guard on `inst == nil` alone does not cover it.

### Fix

One predicate, `mhelpers.HasMachineContext`, placed next to `InitializeMachineWithContext` because that function's caching behaviour is what creates the state. It rejects both a nil instance and a non-nil instance whose `Context` was never assigned, using `utils.IsInterfaceNil` rather than a bare `== nil` so a boxed typed-nil is caught too - matching what `utils.Cast` checks internally. Every site that must resolve a `Context` now gates on it:

| Site | Was | Now |
|---|---|---|
| `machineCtxForConnection` | cast a nil interface, logged `meshkit-11180` each poll | not-ready + a Debug breadcrumb; a genuinely *wrong* Context type is still logged as the real error it is |
| `KubernetesMiddleware`, `K8sFSMMiddleware` | `inst.ResetState()`/`SendEvent` unguarded (latent nil-panic), then a second unguarded `Cast` in `K8sFSMMiddleware` | skipped |
| `DeleteContext` | dispatched the goroutine before checking the error; also raced on the enclosing `err`/`event` | tracker entry dropped directly (see below); goroutine declares its own `err`/`event` |
| `NotifySmOfConnectionStatusChange` | same leak as `DeleteContext` | same fix |
| `k8sconfig_handler` | `inst != nil` only, so the cache-hit case slipped through | skipped |

For the two delete paths, guarding the dispatch alone was not enough: `smInstanceTracker.Remove(...)` sits **after** the `SendEvent` inside the goroutine, so for an uninitialized machine the send fails on `ErrAssertMachineCtx` and returns early, leaking the tracker entry for a connection that is going away. Both now drop the entry directly - a machine that never initialized has no FSM state to unwind and no cluster-side resources to clean up, since `DeleteAction` does all of its work off a `MachineCtx` that was never assigned.

**Deliberately not guarded:** `connections_handlers.go:67` passes a **nil** `InitFunc`, so `Start` succeeds with a nil `Context` by design and the machine is drivable. Guarding it would skip every non-kubernetes registration. This is why the helper is named `HasMachineContext` rather than `IsMachineReady`; its doc comment calls out that non-goal explicitly.

### Why it surfaced now
It doesn't depend on a recent change - any connection with a nil machine context (a non-k8s connection, or a k8s connection whose cluster was unreachable at machine-creation) triggers it once the controller-status SSE stream is open.

### Watch for (tracked follow-up, intentionally out of scope)
- **#20820** - `InitializeMachineWithContext` caching an instance whose `Start` failed is the root cause these guards work around. Caching only on success would let a broken connection self-heal instead of staying stuck until restart, but it changes retry behaviour for a shared helper across five call sites, so it wants its own PR rather than riding along on a log-spam fix.
- **#20821** - the `auto_register.go` call site has the same unguarded pattern; **#20872** is already open against it.

A kubernetes connection created while its cluster was unreachable therefore still stays **contextless** - its controllers report `unknown` - until the process restarts. This PR does not make that case worse; it stops the log spam, the nil-panics, and the leaked tracker entries for it.

### Testing
- `go test ./server/handlers/... ./server/machines/...` - `TestMachineCtxForConnection` covers nil Context, untracked connection, nil tracked instance, boxed typed-nil, **wrong Context type** (the previously untested `err != nil` branch of the cast), missing controllers helper, and a valid `*MachineCtx`. `TestHasMachineContext` covers the helper directly.
- `go build ./server/...`, `go vet ./server/handlers/... ./server/machines/...`, and `golangci-lint run server/handlers/... server/machines/...` (0 issues) all clean.

**[Signed commits]**

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented errors during context deletion when the state machine lacks readiness context.
  * Improved controller status polling by safely handling missing/typed-nil or incorrectly typed machine context, reducing noisy logs.
  * Avoided nil dereferences and invalid operations in Kubernetes state-machine workflows when initialization returns an unusable instance.
  * Ensured connection status changes only publish updates when machine context is present, and cleaned up in-progress trackers appropriately.
* **Tests**
  * Added unit tests covering machine-context readiness detection and controller context resolution across multiple scenarios.

### DIFF
--- a/server/handlers/connections_handlers.go
+++ b/server/handlers/connections_handlers.go
@@ -535,6 +535,20 @@ func (h *Handler) NotifySmOfConnectionStatusChange(ctx context.Context, userID c
 			})
 			return *eventBuilder.Build(), err
 		}
+		// A machine whose Context was never assigned cannot service the event:
+		// SendEvent fails on ErrAssertMachineCtx and returns *before* the Remove
+		// below, so a DELETED update would leak the tracker entry for a
+		// connection that is going away, and publish an error the user can do
+		// nothing about. Drop the entry directly instead. Same shape as the
+		// DeleteContext guard in contexts_handler.go; see helpers.HasMachineContext.
+		if !helpers.HasMachineContext(inst) {
+			h.log.Debug(fmt.Sprintf("machine instance for connection %s has no context assigned, skipping the %q event", connectionID, connection.Status))
+			if connection.Status == connections.DELETED {
+				smInstanceTracker.Remove(connectionID)
+			}
+			return *eventBuilder.Build(), nil
+		}
+
 		// detach from the http request lifecycle so that the goroutine isn't cancelled when
 		// the handler returns, while preserving context values (e.g. TokenCtxKey) that downstream calls depend on.
 		detachedCtx := context.WithoutCancel(ctx)
--- a/server/handlers/connections_handlers.go
+++ b/server/handlers/connections_handlers.go
@@ -529,6 +529,19 @@ func (h *Handler) NotifySmOfConnectionStatusChange(ctx context.Context, userID c
 			kubernetes.AssignInitialCtx,
 		)
 
+		// A connection being deleted must not leave a tracker entry behind,
+		// whichever way its machine failed. InitializeMachineWithContext caches
+		// the instance *before* surfacing a Start error, so both the error return
+		// below and the no-context return after it would otherwise strand the
+		// entry - the first for a fresh failure, the second for every later cache
+		// hit. Nothing can drive that machine afterwards, and the delete paths
+		// that normally Remove it (the goroutine below, DeleteContext) both do so
+		// only after a SendEvent that cannot succeed without a Context.
+		usable := err == nil && helpers.HasMachineContext(inst)
+		if !usable && connection.Status == connections.DELETED {
+			smInstanceTracker.Remove(connectionID)
+		}
+
 		if err != nil {
 			eventBuilder = eventBuilder.WithSeverity(events.Error).WithDescription(fmt.Sprintf("Failed to update connection status for %s", connectionID)).WithMetadata(map[string]interface{}{
 				"error": err,
@@ -536,16 +549,11 @@ func (h *Handler) NotifySmOfConnectionStatusChange(ctx context.Context, userID c
 			return *eventBuilder.Build(), err
 		}
 		// A machine whose Context was never assigned cannot service the event:
-		// SendEvent fails on ErrAssertMachineCtx and returns *before* the Remove
-		// below, so a DELETED update would leak the tracker entry for a
-		// connection that is going away, and publish an error the user can do
-		// nothing about. Drop the entry directly instead. Same shape as the
-		// DeleteContext guard in contexts_handler.go; see helpers.HasMachineContext.
-		if !helpers.HasMachineContext(inst) {
+		// SendEvent would only fail on ErrAssertMachineCtx and publish an error
+		// the user can do nothing about. Same shape as the DeleteContext guard in
+		// contexts_handler.go; see helpers.HasMachineContext.
+		if !usable {
 			h.log.Debug(fmt.Sprintf("machine instance for connection %s has no context assigned, skipping the %q event", connectionID, connection.Status))
-			if connection.Status == connections.DELETED {
-				smInstanceTracker.Remove(connectionID)
-			}
 			return *eventBuilder.Build(), nil
 		}
 
--- a/server/handlers/contexts_handler.go
+++ b/server/handlers/contexts_handler.go
@@ -120,11 +120,16 @@ func (h *Handler) DeleteContext(w http.ResponseWriter, req *http.Request, _ *mod
 		"kubernetes",
 		kubernetes.AssignInitialCtx,
 	)
-	// InitializeMachineWithContext returns a nil instance when the machine
-	// could not be built (e.g. the cluster's API server was unreachable, so
-	// the client set/context failed to initialize); skip driving it rather
-	// than nil-dereferencing in the goroutine below.
-	if inst != nil {
+	// A machine that never initialized has no FSM state to unwind and no
+	// cluster-side resources to clean up: DeleteAction's work (undeploying
+	// operators, flushing MeshSync data) all runs off a MachineCtx that was
+	// never assigned, so SendEvent would only fail with ErrAssertMachineCtx.
+	// Crucially it would also fail *before* reaching the Remove below, leaking
+	// the tracker entry for a connection the user just deleted - so drop the
+	// entry directly instead. See mhelpers.IsMachineReady.
+	if !mhelpers.IsMachineReady(inst) {
+		smInstanceTracker.Remove(connectionUUID)
+	} else {
 		go func(inst *machines.StateMachine) {
 			event, err := inst.SendEvent(req.Context(), machines.Delete, nil)
 			if err != nil {
--- a/server/handlers/contexts_handler.go
+++ b/server/handlers/contexts_handler.go
@@ -120,16 +120,22 @@ func (h *Handler) DeleteContext(w http.ResponseWriter, req *http.Request, _ *mod
 		"kubernetes",
 		kubernetes.AssignInitialCtx,
 	)
-	go func(inst *machines.StateMachine) {
-		event, err = inst.SendEvent(req.Context(), machines.Delete, nil)
-		if err != nil {
-			h.log.Error(err)
-			h.log.Debug(event)
-			return
-		}
+	// InitializeMachineWithContext returns a nil instance when the machine
+	// could not be built (e.g. the cluster's API server was unreachable, so
+	// the client set/context failed to initialize); skip driving it rather
+	// than nil-dereferencing in the goroutine below.
+	if inst != nil {
+		go func(inst *machines.StateMachine) {
+			event, err := inst.SendEvent(req.Context(), machines.Delete, nil)
+			if err != nil {
+				h.log.Error(err)
+				h.log.Debug(event)
+				return
+			}
 
-		smInstanceTracker.Remove(connectionUUID)
-	}(inst)
+			smInstanceTracker.Remove(connectionUUID)
+		}(inst)
+	}
 
 	if err != nil {
 		h.log.Error(err)
--- a/server/handlers/contexts_handler.go
+++ b/server/handlers/contexts_handler.go
@@ -126,8 +126,8 @@ func (h *Handler) DeleteContext(w http.ResponseWriter, req *http.Request, _ *mod
 	// never assigned, so SendEvent would only fail with ErrAssertMachineCtx.
 	// Crucially it would also fail *before* reaching the Remove below, leaking
 	// the tracker entry for a connection the user just deleted - so drop the
-	// entry directly instead. See mhelpers.IsMachineReady.
-	if !mhelpers.IsMachineReady(inst) {
+	// entry directly instead. See mhelpers.HasMachineContext.
+	if !mhelpers.HasMachineContext(inst) {
 		smInstanceTracker.Remove(connectionUUID)
 	} else {
 		go func(inst *machines.StateMachine) {
--- a/server/handlers/controllers_status_handler.go
+++ b/server/handlers/controllers_status_handler.go
@@ -110,6 +110,16 @@ func (h *Handler) machineCtxForConnection(connectionID string) (*kubernetes.Mach
 	if !ok || inst == nil {
 		return nil, false
 	}
+	// A nil Context is an expected "not ready" state, not an error: a
+	// non-kubernetes connection carries no MachineCtx, and a kubernetes one whose
+	// cluster was unreachable when its machine was created never got its context
+	// assigned. Type-casting that nil on every controller-status poll (the ~5s SSE
+	// loop) would otherwise spam the log with meshkit-11180 ("nil interface cannot
+	// be type casted"). Treat it as not-ready and let the caller degrade; only a
+	// genuinely wrong Context type below is a real error worth logging.
+	if inst.Context == nil {
+		return nil, false
+	}
 	machinectx, err := utils.Cast[*kubernetes.MachineCtx](inst.Context)
 	if err != nil || machinectx == nil || machinectx.MesheryCtrlsHelper == nil {
 		if err != nil {
--- a/server/handlers/controllers_status_handler.go
+++ b/server/handlers/controllers_status_handler.go
@@ -120,7 +120,7 @@ func (h *Handler) machineCtxForConnection(connectionID string) (*kubernetes.Mach
 	// unknown status; only a genuinely wrong Context type below is a real error
 	// worth logging. The Debug line keeps a breadcrumb for an operator debugging a
 	// connection stuck at UNKNOWN, without reintroducing Error-severity spam.
-	if !mhelpers.IsMachineReady(inst) {
+	if !mhelpers.HasMachineContext(inst) {
 		h.log.Debug(fmt.Sprintf("machine instance for connection %s has no context assigned, treating as not-ready", connectionID))
 		return nil, false
 	}
--- a/server/handlers/controllers_status_handler.go
+++ b/server/handlers/controllers_status_handler.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/gofrs/uuid"
+	mhelpers "github.com/meshery/meshery/server/machines/helpers"
 	"github.com/meshery/meshery/server/machines/kubernetes"
 	"github.com/meshery/meshery/server/models"
 	"github.com/meshery/meshery/server/models/connections"
@@ -107,18 +108,20 @@ func (h *Handler) machineCtxForConnection(connectionID string) (*kubernetes.Mach
 		return nil, false
 	}
 	inst, ok := h.ConnectionToStateMachineInstanceTracker.Get(connUUID)
-	if !ok || inst == nil {
+	if !ok {
 		return nil, false
 	}
-	// A nil Context is an expected "not ready" state, not an error: a
+	// An unassigned Context is an expected "not ready" state, not an error: a
 	// non-kubernetes connection carries no MachineCtx, and a kubernetes one whose
 	// cluster was unreachable when its machine was created never got its context
 	// assigned. Type-casting that nil on every controller-status poll (the ~5s SSE
 	// loop) would otherwise spam the log with meshkit-11180 ("nil interface cannot
-	// be type casted"). Treat it as not-ready and let the caller degrade; only a
-	// genuinely wrong Context type below is a real error worth logging.
-	if utils.IsInterfaceNil(inst.Context) {
-		h.log.Debug(fmt.Sprintf("machine instance for connection %s has no context yet, treating as not-ready", connectionID))
+	// be type casted"). Treat it as not-ready and let the caller degrade to an
+	// unknown status; only a genuinely wrong Context type below is a real error
+	// worth logging. The Debug line keeps a breadcrumb for an operator debugging a
+	// connection stuck at UNKNOWN, without reintroducing Error-severity spam.
+	if !mhelpers.IsMachineReady(inst) {
+		h.log.Debug(fmt.Sprintf("machine instance for connection %s has no context assigned, treating as not-ready", connectionID))
 		return nil, false
 	}
 	machinectx, err := utils.Cast[*kubernetes.MachineCtx](inst.Context)
--- a/server/handlers/controllers_status_handler.go
+++ b/server/handlers/controllers_status_handler.go
@@ -117,12 +117,12 @@ func (h *Handler) machineCtxForConnection(connectionID string) (*kubernetes.Mach
 	// loop) would otherwise spam the log with meshkit-11180 ("nil interface cannot
 	// be type casted"). Treat it as not-ready and let the caller degrade; only a
 	// genuinely wrong Context type below is a real error worth logging.
-	if inst.Context == nil {
+	if utils.IsInterfaceNil(inst.Context) {
 		h.log.Debug(fmt.Sprintf("machine instance for connection %s has no context yet, treating as not-ready", connectionID))
 		return nil, false
 	}
 	machinectx, err := utils.Cast[*kubernetes.MachineCtx](inst.Context)
-	if err != nil || machinectx == nil || machinectx.MesheryCtrlsHelper == nil {
+	if err != nil || machinectx.MesheryCtrlsHelper == nil {
 		if err != nil {
 			h.log.Error(err)
 		}
--- a/server/handlers/controllers_status_handler.go
+++ b/server/handlers/controllers_status_handler.go
@@ -118,6 +118,7 @@ func (h *Handler) machineCtxForConnection(connectionID string) (*kubernetes.Mach
 	// be type casted"). Treat it as not-ready and let the caller degrade; only a
 	// genuinely wrong Context type below is a real error worth logging.
 	if inst.Context == nil {
+		h.log.Debug(fmt.Sprintf("machine instance for connection %s has no context yet, treating as not-ready", connectionID))
 		return nil, false
 	}
 	machinectx, err := utils.Cast[*kubernetes.MachineCtx](inst.Context)
--- a/server/handlers/controllers_status_handler_test.go
+++ b/server/handlers/controllers_status_handler_test.go
@@ -9,8 +9,74 @@ import (
 	"time"
 
 	"github.com/gofrs/uuid"
+	"github.com/meshery/meshery/server/machines"
+	"github.com/meshery/meshery/server/machines/kubernetes"
 	"github.com/meshery/meshery/server/models"
+	"github.com/meshery/schemas/models/core"
 )
+
+// trackerWith builds an instance tracker holding a single connection->machine
+// mapping, for exercising machineCtxForConnection without standing up real FSMs.
+func trackerWith(id core.Uuid, inst *machines.StateMachine) *machines.ConnectionToStateMachineInstanceTracker {
+	return &machines.ConnectionToStateMachineInstanceTracker{
+		ConnectToInstanceMap: map[core.Uuid]*machines.StateMachine{id: inst},
+	}
+}
+
+// A tracked machine whose Context is nil - a non-kubernetes connection, or a
+// kubernetes one whose cluster was unreachable when the machine was created - is
+// an expected "not ready" state. machineCtxForConnection must report it as
+// not-ready (nil,false) and, crucially, must NOT type-assert the nil Context,
+// which previously logged meshkit-11180 on every controller-status poll.
+func TestMachineCtxForConnection_NilContextIsNotReady(t *testing.T) {
+	connID := uuid.Must(uuid.NewV4())
+	h := &Handler{
+		config:                                  &models.HandlerConfig{},
+		log:                                     newTestLogger(t),
+		ConnectionToStateMachineInstanceTracker: trackerWith(connID, &machines.StateMachine{ID: connID, Context: nil}),
+	}
+
+	ctx, ok := h.machineCtxForConnection(connID.String())
+	if ok {
+		t.Fatalf("expected not-ready (ok=false) for a nil-Context machine, got ok=true")
+	}
+	if ctx != nil {
+		t.Fatalf("expected nil machine context for a nil-Context machine, got %#v", ctx)
+	}
+}
+
+// A machine with no tracked instance is likewise not-ready.
+func TestMachineCtxForConnection_UntrackedIsNotReady(t *testing.T) {
+	h := &Handler{
+		config:                                  &models.HandlerConfig{},
+		log:                                     newTestLogger(t),
+		ConnectionToStateMachineInstanceTracker: trackerWith(uuid.Must(uuid.NewV4()), &machines.StateMachine{Context: nil}),
+	}
+
+	if ctx, ok := h.machineCtxForConnection(uuid.Must(uuid.NewV4()).String()); ok || ctx != nil {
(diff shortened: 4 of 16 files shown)


leecalcote (commented):
Reviewing...

simihablo:
Razzmatazzing…

server/handlers/controllers_status_handler.go
```diff
@@ -110,6 +110,16 @@ func (h *Handler) machineCtxForConnection(connectionID string) (*kubernetes.Mach
...
 		return nil, false
 	}
+	// A nil Context is an expected "not ready" state, not an error: a
+	// non-kubernetes connection carries no MachineCtx, and a kubernetes one whose
+	// cluster was unreachable when its machine was created never got its context
+	// assigned. Type-casting that nil on every controller-status poll (the ~5s SSE
+	// loop) would otherwise spam the log with meshkit-11180 ("nil interface cannot
+	// be type casted"). Treat it as not-ready and let the caller degrade; only a
+	// genuinely wrong Context type below is a real error worth logging.
+	if inst.Context == nil {
```
leecalcote: **Reuse + latent gap:** this bypasses `kubernetes.GetMachineCtx` (`server/machines/kubernetes/helpers.go:50`), the codebase's existing canonical nil-safe accessor for exactly this "resolve `Context` to `*MachineCtx`" job — already used at 4+ other call sites (`connection_action_handler.go:226,283`, `controllers_config_handlers.go:331`, `connections_handlers.go:659`), each of which still treats a nil/failed `Context` as warn/error/propagated-error. This function now has its own, more-silent notion of "nil is expected" that those other sites don't share — a future change to what counts as "not ready" has to be hand-applied in two+ places.

Separately: `inst.Context == nil` is a bare nil-interface check, which is narrower than what `utils.Cast` already does internally two lines below (`utils.Cast` calls meshkit's `IsInterfaceNil`, which uses `reflect.ValueOf(val).IsZero()` and so also catches a boxed *typed-nil* pointer, not just a literal nil interface). No current `InitFunc` implementation assigns a typed-nil to `sm.Context` today, so this is latent rather than live — but if a future `AssignInitialCtx` change ever returns its partially-built `*MachineCtx` instead of a literal `nil` on an error path, this guard silently stops working and the exact `meshkit-11180` spam this PR fixes would reappear.

Suggested fix (verified locally, build/vet/test/lint clean): `if utils.IsInterfaceNil(inst.Context) { return nil, false }` — matches `utils.Cast`'s own semantics exactly, and is already the established idiom elsewhere in this repo (e.g. `server/models/pattern/stages/filler.go`).

server/handlers/controllers_status_handler.go
```diff
@@ -110,6 +110,16 @@ func (h *Handler) machineCtxForConnection(connectionID string) (*kubernetes.Mach
...
+		return nil, false
+	}
 	machinectx, err := utils.Cast[*kubernetes.MachineCtx](inst.Context)
 	if err != nil || machinectx == nil || machinectx.MesheryCtrlsHelper == nil {
```
leecalcote: **Simplification:** given the new guard two lines above already returns `(nil, false)` for any nil `inst.Context`, `machinectx == nil` here is unreachable dead code — `utils.Cast`'s internal `IsInterfaceNil` check means it can never return `(nil-typed-value, nil-error)`, so whenever `err == nil`, `machinectx` is guaranteed non-nil. This was already true before this PR (this line is unchanged), but the new guard right above is a good prompt to simplify it now: `if err != nil || machinectx.MesheryCtrlsHelper == nil {`. Verified this reduction is behavior-preserving by tracing `utils.Cast`'s source (meshkit v1.0.22, pinned in `go.mod`).

server/handlers/middlewares.go
```diff
@@ -413,6 +420,13 @@ func K8sFSMMiddleware(ctx context.Context, h *Handler, provider models.Provider,
 			h.log.Error(err)
 		}
 
+		// InitializeMachineWithContext returns a nil instance when the machine
+		// could not be built (e.g. the cluster's API server was unreachable, so the
+		// client set / context failed to initialize). Skip driving it rather than
+		// nil-dereferencing on ResetState/SendEvent; a later request re-attempts.
+		if inst == nil {
```
leecalcote: **Simplification:** this 7-line block (comment + `if inst == nil { continue }`) is byte-for-byte identical to the one added at `middlewares.go:357-363` in `KubernetesMiddleware`, on top of already-near-duplicate surrounding code in both functions (same `MachineCtx{}` literal, same `InitializeMachineWithContext` call/args, same error-log line, same `ResetState`+goroutine dispatch). This diff adds a third copy of logic already duplicated twice.

Also worth a guard while you're in this function: `K8sFSMMiddleware` has a *second*, unguarded `utils.Cast[*kubernetes.MachineCtx](inst.Context)` at line 438, a few lines past this new guard — for a `Context`-nil cached instance (see the top-level review comment on why that's more reachable than it looks) it reproduces the exact `meshkit-11180` error this PR's commit message says it eliminates, on every request through this middleware. Locally I added the same `if inst.Context == nil { continue }` guard there too, and extracting the shared "check inst, ResetState, dispatch Discovery" tail into one helper (`KubernetesMiddleware`'s loop body ends right after the dispatch; `K8sFSMMiddleware`'s doesn't, so it would still need its own nil-check for its own trailing code) is a reasonable follow-up but didn't feel safe to do as a drive-by given the two functions' control flow isn't quite identical past this point.

leecalcote (commented):
I ran a 10-angle parallel review (5 correctness + reuse/simplification/efficiency/altitude/conventions), verified every candidate with an independent 1-vote pass, then a gap sweep. 11 findings survived; 3 more were investigated and refuted (a `MesheryCtrlsHelper` nil-deref theory, a diagnostics/observability-regression theory, and a "tests should be table-driven" theory — all checked against the actual code and ruled out).

**The core issue:** this PR's two guards (`inst == nil` in the middlewares, `inst.Context == nil` in `machineCtxForConnection`) each correctly stop the *first* failure they encounter, but the actual root cause is one layer deeper and both guards miss it. `InitializeMachineWithContext` (`server/machines/helpers/helpers.go:92`) caches a broken instance via `smInstanceTracker.Add(ID, inst)` **before** checking whether `inst.Start(...)` succeeded. So the very next call for that same connection ID hits the cache and returns `(inst, nil)` — non-nil instance, no error — and the new guards never fire again. The comment this PR adds twice ("a later request re-attempts") is not accurate for the steady state: `Start`/`AssignInitialCtx` is never retried for a cached ID.

I verified this has four concrete, reachable consequences beyond the meshkit-11180 log spam this PR set out to fix (details below with file/line). I applied fixes for all of these in a local worktree (build/vet/test/lint all clean, including the PR's own 3 new tests) but have **not pushed them** to the PR branch — I don't have write access to `marblom007/meshery` from this session, and wanted to flag rather than silently force-push regardless. Happy to open a follow-up PR against this branch if useful.

### Findings not attachable as inline comments (outside this diff's hunks)

1. **`server/machines/helpers/helpers.go:92`** (root cause) — `smInstanceTracker.Add(ID, inst)` runs unconditionally, before checking `inst.Start(...)`'s error. A machine whose `Start` failed (cluster unreachable, `Context` never assigned) is cached forever; the tracker's `Get(ID)` short-circuit at the top of the function means `Start`/`AssignInitialCtx` is never retried, contradicting this PR's own new comments. **Fix applied locally:** only call `Add` after confirming `err == nil`.

2. **`server/handlers/contexts_handler.go:123`** — `DeleteContext` spawns `go func(inst){ inst.SendEvent(...) }(inst)` with **no nil-check on `inst` at all** — the same defect class this PR just fixed in the two middlewares, left live in a sibling handler. `SendEvent` dereferences `sm.Provider`/`sm.mx` on the receiver, so a nil `inst` here panics inside an unrecovered goroutine — that crashes the **entire server process**, not just one request. Also found: the same goroutine writes to the outer function's `err` variable via `=` (not `:=`) while the enclosing function reads that same `err` right after spawning it — a real data race, independent of the nil-panic. **Fix applied locally:** nil-guard + give the goroutine its own local error variable.

3. **`server/machines/models.go` (~line 241-248, `SendEvent`)** — the action-error path returns immediately when the failing action reports `NoOp`, *before* reaching the `statusChanged` de-dup logic built later in the same function specifically to stop repeated-failure notification spam. For a permanently-broken (cached, `Context`-nil) connection, `DiscoverAction.Execute` fails this way on every `Discovery` event, so `KubernetesMiddleware`/`K8sFSMMiddleware`'s goroutine (`if err != nil { provider.PersistEvent(...); EventBroadcaster.Publish(...) }`) persists and broadcasts a fresh Error-severity "Failed to interact with the connection" notification **to the user, on every single request**, for as long as the connection stays broken. One reviewing agent verified this empirically by writing and running a scratch test against the real `SendEvent`. This is worse than the log spam the PR fixes, through a path the PR didn't touch. Not fixed locally — restructuring `SendEvent`'s control flow to route this path through the existing dedup logic risks changing behavior for other event types that share this code, so I left it as a flagged finding rather than a drive-by fix.

4. **`server/handlers/middlewares.go:438`** (`K8sFSMMiddleware`) — a *second*, unguarded `utils.Cast[*kubernetes.MachineCtx](inst.Context)`, a few lines past this PR's own new guard in the *same function*. For a `Context`-nil cached instance it reproduces the exact `meshkit-11180` error this PR's commit message says it eliminated — on every request through this middleware (wired to `/api/system/sync`), which is a *hotter* path than the ~5s SSE poll this PR targeted. **Fix applied locally:** added the same `Context == nil` guard used above.

5. **`server/machines/helpers/helpers.go:80-92`** (`InitializeMachineWithContext`) — the get-or-create sequence isn't atomic: `Get` and `Add` each lock independently, with the miss→build→`Start`→`Add` span running unguarded between them. Two concurrent calls for the same *new* connection ID (realistically triggered by MeshSync emitting rapid ADDED→MODIFIED events for the same object, each spawning its own unrate-limited goroutine in `auto_register.go`) can each build and `Start` their own instance; whichever `Add()` runs last wins the map slot, silently orphaning the other — which may have already advanced real persisted/DB state that the winning in-memory instance doesn't reflect. Not fixed locally (needs a per-ID lock or `singleflight`, which is a bigger, separate change than this review's scope).

6. **Same location** — a smaller, related gap: the cache-hit path ignores the `initFunc` parameter entirely. Some callers pass `initFunc=nil` (`auto_register.go`, `connections_handlers.go:67`), others pass `kubernetes.AssignInitialCtx` (both middlewares, `contexts_handler.go`, `k8sconfig_handler.go`, `connections_handlers.go:519`). Whichever caller populates the tracker for a given ID first "wins" — permanently — regardless of which `initFunc` later callers expected to run. Not fixed locally (architectural; needs a product decision on how different callers'/lifecycle stages' `initFunc` expectations should interact).

7. **`server/machines/helpers/auto_register.go:111`** — `machineInst.SendEvent(...)` runs with no nil-check, same defect class as #2, currently dormant (every invariant that would trigger a nil return from `InitializeMachineWithContext` for this specific call site happens to hold today — `getTypeOfConnection` only yields `"grafana"`/`"prometheus"`, and `initFunc=nil` here means `Start` can't error). **Fix applied locally anyway**, since it's a one-line, zero-risk guard consistent with the pattern this PR already established.

Inline comments below cover the 4 findings anchored inside this diff's actual hunks.

server/handlers/middlewares.go
marblom007: **Bug:** the new `inst == nil` guard doesn't cover a non-nil `*StateMachine` whose `Context` is permanently nil.

`InitializeMachineWithContext` caches the instance via `smInstanceTracker.Add` even when `Start` failed and `Context` was never assigned. The *first* call for a broken connection correctly returns `(nil, err)` and this guard skips it - but every *subsequent* call for the same connection hits the tracker's cache-hit path (`if ok { return inst, nil }`) and gets back that same non-nil, `Context`-nil instance. This guard doesn't catch that, so `utils.Cast[*kubernetes.MachineCtx](inst.Context)` three lines down still runs on a nil interface and reproduces the exact meshkit-11180 spam this PR sets out to fix.

Suggested fix: extend the guard to `if inst == nil || inst.Context == nil { continue }`.

server/handlers/middlewares.go
```diff
@@ -354,6 +354,13 @@ func KubernetesMiddleware(ctx context.Context, h *Handler, provider models.Provi
...
+			continue
+		}
 		inst.ResetState()
 		go func(inst *machines.StateMachine) {
 			event, err := inst.SendEvent(ctx, machines.Discovery, nil)
```
marblom007, line 365: **Bug:** same underlying gap as the K8sFSMMiddleware guard below - once a connection's broken instance is cached (see `InitializeMachineWithContext`, which `Add`s to the tracker before checking the `Start` error), every later request through this middleware clears the `inst == nil` check and still calls `inst.SendEvent(ctx, machines.Discovery, nil)`. That fails with `ErrAssertMachineCtx` on the nil `Context`, and the goroutine's `err != nil` branch persists + broadcasts a fresh "Failed to interact with the connection" Error notification on every single request against that connection for as long as the cluster stays unreachable - the same class of spam this PR is fixing, just relocated from a log line to a user-visible notification.

Suggested fix: same as below - `if inst == nil || inst.Context == nil { continue }`.

server/handlers/middlewares.go
```diff
@@ -354,6 +354,13 @@ func KubernetesMiddleware(ctx context.Context, h *Handler, provider models.Provi
 			h.log.Error(err)
 		}
 
+		// InitializeMachineWithContext returns a nil instance when the machine
+		// could not be built (e.g. the cluster's API server was unreachable, so the
+		// client set / context failed to initialize). Skip driving it rather than
+		// nil-dereferencing on ResetState/SendEvent; a later request re-attempts.
```
marblom007: **Nit:** this comment ("a later request re-attempts") is inaccurate and contradicts this same PR's own commit-message caveat.

Once a connection ID is added to `smInstanceTracker` (which happens unconditionally, even on a failed `Start`), every later call short-circuits on the cache-hit path and never re-invokes `getMachine`/`Start` for that ID again - grepping the tracker's `.Remove(` call sites shows it's only ever cleared by an explicit user action (delete connection, cancel registration), never by a later request through this middleware. Worth correcting so future readers don't assume this self-heals.

server/handlers/controllers_status_handler.go
marblom007: **Observability gap:** this early return makes the pre-existing `h.log.Error(err)` cast-failure log at line ~126 permanently unreachable for a connection whose instance never gets a `Context` - and because of the tracker's cache-hit behavior (see comment on middlewares.go), that's the *only* log tied to this failure mode after the very first request. An operator investigating a connection stuck at UNKNOWN later finds zero server-side trace explaining why.

This return-and-be-silent behavior is intentional per the PR description (nil Context is a valid not-ready state, not an error) - but a Debug-level breadcrumb here would preserve traceability without reintroducing the Error-severity spam this PR removes.

server/handlers/controllers_status_handler_test.go
```diff
@@ -9,9 +9,75 @@ import (
...
+		ConnectToInstanceMap: map[core.Uuid]*machines.StateMachine{id: inst},
+	}
+}
+
+// A tracked machine whose Context is nil - a non-kubernetes connection, or a
+// kubernetes one whose cluster was unreachable when the machine was created - is
+// an expected "not ready" state. machineCtxForConnection must report it as
+// not-ready (nil,false) and, crucially, must NOT type-assert the nil Context,
+// which previously logged meshkit-11180 on every controller-status poll.
+func TestMachineCtxForConnection_NilContextIsNotReady(t *testing.T) {
```
marblom007: **Nit:** these three `machineCtxForConnection` cases are three separate top-level funcs, each re-declaring the identical 3-field `Handler` literal, rather than one table-driven test - unlike `TestControllerStatusHandlers_RejectInvalidConnectionID` further down this same file, which already establishes a table+`t.Run` convention for exactly this shape. Collapsing them makes it a one-line change to add a 4th case (e.g. a tracked instance with a wrong, non-`MachineCtx` Context type, exercising the untested `err != nil` branch of the cast).

marblom007:
**Sibling bug outside this diff (`server/handlers/contexts_handler.go:123`):** `DeleteContext` has the identical nil-instance hazard this PR fixes in `KubernetesMiddleware`/`K8sFSMMiddleware`, but in an unrecovered goroutine instead of a guarded one:

```go
inst, err := mhelpers.InitializeMachineWithContext(...)
    event, err = inst.SendEvent(req.Context(), machines.Delete, nil)
    ...
}(inst)

if err != nil {
```

`InitializeMachineWithContext` returns `(nil, err)` for the same unreachable-cluster case this PR documents. Here the goroutine is dispatched with `inst` *before* the `err != nil` check even runs, so a delete against an unreachable cluster's connection panics inside `inst.SendEvent` on a nil receiver (`sm.mx.Lock()` in `machines/models.go` dereferences `sm`) - and nothing under `server/` wraps goroutines with `recover()`, so it takes down the whole server process, not just the request.

Separately, that same goroutine writes to the *outer* `event`/`err` variables (`=`, not `:=`) while the caller reads `err` immediately after launching it - an unsynchronized data race independent of the nil-panic.

Since this file isn't part of this PR's diff I can't leave an inline comment on it, but given it's the exact contract this PR is establishing elsewhere, I've applied a fix locally: guard the goroutine dispatch on `inst != nil` and declare the goroutine's `event`/`err` locally via `:=` instead of writing to the enclosing scope. Happy to push it as a follow-up commit on this branch, or split into a separate PR - your call.

Also worth flagging for awareness (not fixed here, already tracked separately):
- **meshery#20820** - `InitializeMachineWithContext` caches the instance via `smInstanceTracker.Add` even when `Start` fails, so every call after the first broken one hits the cache-hit path and hands back the same non-nil, `Context`-nil instance - this is the root cause the `inst.Context == nil` guard extension (see inline comments on `middlewares.go`) works around at the call sites.
- **meshery#20821** - `auto_register.go`'s `InitializeMachineWithContext` call site has the same unguarded-nil-instance pattern as `contexts_handler.go`, but inline (not even in a goroutine), so it's a synchronous crash rather than a background one.

server/handlers/controllers_status_handler.go
marblom007: Good catch — switched to utils.IsInterfaceNil to match Cast's own semantics exactly, so this stays correct even if a future InitFunc ever assigns a typed-nil Context.

server/handlers/controllers_status_handler.go
marblom007: Agreed, dropped — with IsInterfaceNil rejecting nil/typed-nil upfront, Cast can no longer return a nil value with a nil error, so the extra check was dead.

server/handlers/middlewares.go
marblom007: Already covered — both KubernetesMiddleware and K8sFSMMiddleware now guard on inst == nil || inst.Context == nil before their respective ResetState/SendEvent/Cast calls.

server/handlers/controllers_status_handler.go
marblom007: The IsInterfaceNil swap landed earlier; the reuse half is now addressed too, though not by calling `GetMachineCtx` here. `GetMachineCtx` resolves a `Context` with a plain type assertion, so a nil interface takes its `!ok` branch and returns `ErrAssertMachineCtx` - the very error class this PR is trying to stop emitting on the ~5s poll. Reusing it at this call site would have needed the nil pre-check anyway.

What it would not have fixed is your actual concern: two divergent notions of "not ready". That is now single-sourced as `mhelpers.IsMachineReady` (`server/machines/helpers/helpers.go`), sitting next to `InitializeMachineWithContext` - the function whose cache-before-error-check behavior creates the state - and used by all four sites (both middlewares, `machineCtxForConnection`, `DeleteContext`). A future change to what counts as not-ready is now a one-line change in one place. Its doc comment carries the typed-nil rationale you gave.

server/handlers/controllers_status_handler.go
marblom007: Dropped, and still dropped after the refactor - `IsMachineReady` rejects nil/typed-nil before the cast, so `Cast` cannot return a nil value with a nil error.

server/handlers/middlewares.go
marblom007: Both halves now done. The second unguarded `Cast` at the old line 438 is covered - the guard sits above it. And the duplication you flagged is gone rather than tripled: both byte-identical 9-line comment blocks are replaced by `if !mhelpers.IsMachineReady(inst) { continue }`, with the rationale living once in the helper doc. I did not merge the two loop bodies wholesale, for the reason you gave - the control flow diverges past the dispatch, since `K8sFSMMiddleware` continues on to collect MeshSync data handlers.

marblom007: Fixed. The guard covers the cached `Context`-nil instance, now via `mhelpers.IsMachineReady`, which additionally uses `utils.IsInterfaceNil` rather than the bare `Context == nil` this comment suggested - the bare check misses a boxed typed-nil, which is exactly what was flagged on the controller-status guard.

server/handlers/middlewares.go
marblom007: Fixed in `KubernetesMiddleware` too - same `mhelpers.IsMachineReady` guard, so the repeated `ErrAssertMachineCtx` persist+broadcast on every request against a broken connection no longer happens.

server/handlers/middlewares.go
marblom007: Corrected. The comment no longer claims a later request re-attempts; the "only ever cleared by an explicit user action, so it persists for the life of the process" explanation now lives once in the `IsMachineReady` doc comment, which is where the cache-hit behavior is described.

server/handlers/controllers_status_handler.go
marblom007: Added - `machineCtxForConnection` now emits a Debug line naming the connection before returning not-ready, so an operator investigating a connection stuck at UNKNOWN has a trace without reintroducing Error-severity spam on the poll.

server/handlers/controllers_status_handler_test.go
marblom007: Collapsed into one table-driven test, and I took the 4th case you suggested: `wrong Context type is not-ready` boxes a non-`MachineCtx` value and exercises the previously untested `err != nil` branch of the cast. Also added `typed-nil Context`, `nil tracked instance` and `missing controllers helper` cases, plus a direct table-driven test for the new `IsMachineReady` helper.

server/handlers/contexts_handler.go
```diff
@@ -120,16 +120,22 @@ func (h *Handler) DeleteContext(w http.ResponseWriter, req *http.Request, _ *mod
...
+			event, err := inst.SendEvent(req.Context(), machines.Delete, nil)
+			if err != nil {
+				h.log.Error(err)
+				h.log.Debug(event)
+				return
+			}
+
+			smInstanceTracker.Remove(connectionUUID)
+		}(inst)
+	}
```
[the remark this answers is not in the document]
marblom007: Real issue - fixed, but skipping alone was not enough. `smInstanceTracker.Remove(connectionUUID)` lives *after* the `SendEvent` inside that goroutine, so for an uninitialized machine the send fails on `ErrAssertMachineCtx` and returns early, leaving the tracker entry behind for a connection the user just deleted.

`DeleteContext` now branches on `mhelpers.IsMachineReady`: a machine that never initialized has no FSM state to unwind and no cluster-side resources to clean up - `DeleteAction` does all its work (undeploying operators, flushing MeshSync data) off a `MachineCtx` that was never assigned - so the tracker entry is dropped directly and the delete completes. Readiness uses `utils.IsInterfaceNil` as you suggested, shared with the controller-status path.

marblom007:
Picked this up and worked through all the outstanding review feedback. Every inline thread now has a reply; summarising what changed since `8ee2d1a`.

**The sibling bug in `DeleteContext` is now fixed in this PR**, along with a second one the original note missed: guarding the dispatch alone is not enough, because `smInstanceTracker.Remove(connectionUUID)` sits *after* the `SendEvent` inside that goroutine. For a machine that never initialized, the send fails on `ErrAssertMachineCtx` and returns early, leaving the tracker entry behind for a connection the user just deleted. `DeleteContext` now drops the entry directly - a machine that never initialized has no FSM state to unwind and no cluster-side resources to clean up, since `DeleteAction` does all of its work (undeploying operators, flushing MeshSync data) off a `MachineCtx` that was never assigned.

**Consolidation.** The PR had accumulated four hand-rolled variants of the same "is this machine usable" check, and they had already drifted: `machineCtxForConnection` used `utils.IsInterfaceNil` while both middleware guards used a bare `inst.Context == nil` - the narrower check that was rejected in review, since it misses a boxed typed-nil. All sites now route through one predicate, `mhelpers.HasMachineContext`, placed next to `InitializeMachineWithContext` because that function's cache-before-error-check behaviour (#20820) is what creates the state. Its doc comment is where the rationale lives now, which also removed two byte-for-byte identical 9-line comment blocks.

**Recurrence check.** Audited the other five `InitializeMachineWithContext` call sites:

- `k8sconfig_handler.go:285` had the same `inst != nil` guard with the same cache-hit hole - fixed.
- `connections_handlers.go` (`NotifySmOfConnectionStatusChange`) had the exact `Remove`-after-`SendEvent` leak as `DeleteContext` - fixed.
- `connections_handlers.go:67` passes a **nil** `InitFunc`, so `Start` succeeds with a nil `Context` by design and the machine is drivable. Deliberately left alone - guarding it would skip every non-kubernetes registration. This is also why the helper is named `HasMachineContext` rather than `IsMachineReady`; the doc comment calls out the non-goal so nobody applies it there later.
- `auto_register.go` is tracked in #20821 and already has #20872 open against it.

**Not attempted here:** the root cause, #20820. Making `InitializeMachineWithContext` cache only on a successful `Start` would let a broken connection self-heal instead of staying stuck until restart, but it changes retry behaviour for a shared helper on five call sites and deserves its own PR and reviewer attention rather than riding along on a log-spam fix.

**Verification:** `go build ./server/...`, `go vet`, `go test ./server/handlers/... ./server/machines/...` and `golangci-lint run` (0 issues) all clean locally. Test coverage added for the previously untested cast-error branch, plus typed-nil, nil-instance and half-built contexts, and a direct table-driven test for the new helper.

server/handlers/connections_handlers.go
```diff
@@ -535,6 +535,20 @@ func (h *Handler) NotifySmOfConnectionStatusChange(ctx context.Context, userID c
...
+		// below, so a DELETED update would leak the tracker entry for a
+		// connection that is going away, and publish an error the user can do
+		// nothing about. Drop the entry directly instead. Same shape as the
+		// DeleteContext guard in contexts_handler.go; see helpers.HasMachineContext.
+		if !helpers.HasMachineContext(inst) {
+			h.log.Debug(fmt.Sprintf("machine instance for connection %s has no context assigned, skipping the %q event", connectionID, connection.Status))
+			if connection.Status == connections.DELETED {
+				smInstanceTracker.Remove(connectionID)
+			}
+			return *eventBuilder.Build(), nil
```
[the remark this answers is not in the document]
marblom007: Correct, and fixed in 07ab8da. The first failed DELETE stranded the cached instance because the error return sits ahead of the `HasMachineContext` guard.

Took the centralize option rather than a second `Remove` call, so both cases are one statement expressing the invariant - a connection being deleted leaves no tracker entry behind, whichever way its machine failed:

```go
usable := err == nil && helpers.HasMachineContext(inst)
if !usable && connection.Status == connections.DELETED {
 smInstanceTracker.Remove(connectionID)
}
```

Checked `DeleteContext` in `contexts_handler.go` for the same shape - it does not share the bug, since it has no early return on the initialization error and its failure path already falls through to the removal branch.

server/handlers/controllers_status_handler.go
marblom007: One correction to my reply above: the helper shipped as `mhelpers.HasMachineContext`, not `IsMachineReady` (my replies on the other threads still use the old name).

Auditing the remaining `InitializeMachineWithContext` callers is what forced the rename. `connections_handlers.go:67` passes a **nil** `InitFunc`, so `Start` assigns a nil `Context` and returns success - the machine is perfectly drivable with no Context at all. "Ready" invited applying the predicate there, which would have silently skipped every non-kubernetes connection registration. The name now states only what it tests, and the doc comment calls out that non-goal explicitly so the trap is not re-entered.

Issue #18771, opened 2026-04-16
[Bug] Zombie State Machine permanently blocks Connection Initialization


KumarADITHYA123:
## Description
There is a critical logical flaw in `InitializeMachineWithContext` (`server/machines/helpers/helpers.go`) that permanently paralyzes the state machine tracker upon a single connection initialization failure.

When `inst.Start(...)` returns an error (e.g. database timeout, unavailable cluster, faulty adapter), the broken `inst` is aggressively and unconditionally added to the global `smInstanceTracker.ConnectToInstanceMap` BEFORE the error is evaluated and returned.

Consequently, any subsequent attempts by the user or system to reconnect utilizing `InitializeMachineWithContext` hit the tracker cache (`smInstanceTracker.Get(ID)`) and silently retrieve the "Zombie" state machine. The instance is permanently blocked from a fresh restart until the entire core Meshery server binary is restarted.

## Root Cause Code 
**File**: `server/machines/helpers/helpers.go`
```go
	_, err = inst.Start(ctx, machineCtx, log, initFunc)
	smInstanceTracker.Add(ID, inst)    // <-- Added to cache here
	if err != nil {
		return nil, err            // <-- Error returned, but instance is permanently cached
	}
```

## Steps to Reproduce
1. Attempt to initialize any connection that delegates to `InitializeMachineWithContext` with an intentionally failing configuration.
2. The initial request correctly returns the expected error (e.g. cluster timeout).
3. Wait 5 minutes and correct the configuration, then attempt to reconnect using the same ID.
4. The backend silently retrieves the broken instance from `smInstanceTracker` and refuses to trigger `getMachine()` or `inst.Start()` again.

## Expected Behavior
The `smInstanceTracker` should strictly embrace defensive programming. Failed or broken state machines should NEVER be inserted into the persistent tracking map. `smInstanceTracker.Add` must only execute upon confirmed success of `inst.Start()`.

## Environment
- **Meshery Server Context (Go Backend)**
- Version: `master`
- Impact: Server/Connection unreliability (P1)

marblom007:
[...]

Issue #20334, opened 2026-06-28
Notification Not deleting


dhruveshmishra:
### Current Behavior
Once I am deleting the notifications the notification is shown deleted but as I refresh the page it is regenerated

### Expected Behavior
If I had deleted the notification then it should not be shown again

### Screenshots/Logs

https://github.com/user-attachments/assets/4f70c726-cb69-44c1-8549-043e9cbc9032

### Environment

- **Host OS:** Mac Linux Windows

Issue #1459, opened 2020-08-25
[UI] Notification board should close when notification is cleared


kalradev:
**Current Behavior**

![Screenshot from 2020-08-25 23-23-40](https://user-images.githubusercontent.com/32228313/91209908-0c58a480-e72a-11ea-9af1-c16e4b5d482a.png)

When notification is cleared using the clear Icon, notification board doesnot close. 

**Desired Behavior**

It should close, because there is no use for the notification board after clearing notifications. 
User experience would be better this way.

leecalcote:
Thanks for documenting this @kalradev.

PmcFizz:
This‘s a good suggest. I will finish it .
@kalradev @leecalcote

leecalcote:
Thank you, sir. @PmcFizz 👍

Issue #220, opened 2019-08-22
UX: Dismissing an individual notification closes the Notification bar.


leecalcote:
**Issue Description**
Current behavior is that when a user dismisses an individual notification, the entire Notification bar is dismissed as well.
![Untitled](https://user-images.githubusercontent.com/7570704/63514258-d7c66300-c4ad-11e9-927b-271022ce25e9.gif)

**Enhancement Description**
Ideally, when a user dismisses an individual notification, the Notification bar will remain open until the user shifts focus elsewhere.

girishranganathan:
2a0fb9dab55600cc9142e045ae6c2d6c4ab4101d addresses this issue

girishranganathan:
fixed in #234

PR #20826, opened 2026-07-20
[Server] Stop per-request Error-notification spam for broken connections (SendEvent NoOp-halt bypasses de-dup)

### LINKED ISSUE #20818: [Server] Broken connection persists+broadcasts a fresh Error notification on every request (SendEvent bypasses its own dedup)
### Current Behavior

In `SendEvent` (`server/machines/models.go`, action-error path around lines 241-248), when a failing action reports `NoOp`, the function **returns immediately, before reaching the `statusChanged` de-dup logic** built later in the same function specifically to stop repeated-failure notification spam.

For a permanently-broken connection (a cached instance whose `Context` is nil — see #18771), `DiscoverAction.Execute` fails this way on **every** `Discovery` event. `KubernetesMiddleware` / `K8sFSMMiddleware` then run, per request:

```go
if err != nil {
    _ = provider.PersistEvent(*event, token)
    go h.config.EventBroadcaster.Publish(userUUID, event)
}
```

so a fresh **Error-severity "Failed to interact with the connection" notification is persisted and broadcast to the user on every single request**, for as long as the connection stays broken. This is worse (user-facing) than the log spam #20812 addressed, via a path that PR did not touch. Verified empirically against the real `SendEvent`.

### Expected Behavior

The repeated-failure path routes through the existing `statusChanged` dedup so a persistently-broken connection emits at most one persisted/broadcast event per actual status transition, not one per request.

### Notes
Surfaced in @leecalcote's out-of-band review of #20812: https://github.com/meshery/meshery/pull/20812#pullrequestreview-4736028511 (finding 3). Not fixed locally: restructuring `SendEvent`'s control flow risks changing behavior for other event types that share this code. Related: #20725 (notification event semantics), #18771 (the broken-cache root cause that keeps a connection permanently in this state).

marblom007:
## Notes for Reviewers

Fixes #20818.

### Symptom
A permanently broken Kubernetes connection **persists and broadcasts a fresh Error-severity "Failed to interact with the connection" notification to the user on every single request** - not just once. (User-facing; worse than the `meshkit-11180` log spam fixed in #20812.)

### Root cause
`SendEvent` (`server/machines/models.go`) has two failure exits:

- **Recovery-event failure** (e.g. `DiscoverAction` returns `NotFound` on a reachability failure): the loop transitions to the recovery state, which is persisted, and the `statusChanged` de-dup at the end suppresses repeats once the connection has converged (e.g. to `NOTFOUND`).
- **NoOp halt** (an action returns an error **with `NoOp`** - no recovery event): the loop **returned inline**, *before* that de-dup block ever runs.

`DiscoverAction.Execute` hits the NoOp path whenever `GetMachineCtx` fails - i.e. a broken/zombie connection whose machine context never initialized (see #18771). `K8sFSMMiddleware` re-runs `ResetState()` + `SendEvent(Discovery)` on **every** `/api` request, and its goroutine does `PersistEvent(*event)` + `EventBroadcaster.Publish(event)` whenever `SendEvent` returns an error - so the same Error event was re-persisted and re-broadcast per request, indefinitely.

### Fix
Route the halt-without-transition paths through the **existing** de-dup instead of returning inline:

- `break` out of the loop (record `haltedWithoutTransition`) rather than `return event, err`.
- **Skip the status write** - the machine made no transition, and writing the un-advanced `CurrentState` (e.g. `InitialState` after `ResetState`) would corrupt the persisted status (e.g. `CONNECTED` -> `initialized`).
- Start `statusChanged` as `!haltedWithoutTransition` so the background `Discovery` failure is suppressed.

The **exit-action failure** branch is the sibling of the same defect and is fixed identically (second commit). An exit action failing aborts the transition before the machine advances - the same "halted without transitioning" shape - but it also returned inline. Two consequences: it bypassed the de-dup, and because *every* `ExecuteOnExit` in the tree builds a **nil** event, it handed callers a nil event alongside a non-nil error - and `K8sFSMMiddleware`/`addK8SConfig` dereference `(*event)` whenever the error is non-nil, so that path was a latent nil-pointer panic. It is unreachable today (all 11 `ExecuteOnExit` implementations return a nil error), but leaving it means the first real exit action reintroduces this bug verbatim. `haltedNoOp` is renamed to `haltedWithoutTransition` now that two paths set it.

Behavior after the fix:
- **Background `Discovery` halt** (NoOp or exit-action) -> suppressed (no event, no error, no status write) - no more per-request spam.
- **User-initiated event halt** (Register/Connect/...) -> still **propagates** a non-nil Error event + error (a manual action never silently "succeeds", and never returns a nil event on a non-nil error).
- **Recovery-event failures and the success path** -> unchanged.

### Scope note
This is one of the out-of-band findings from @leecalcote's review of #20812 (finding 3). The deeper root cause it depends on - a broken machine being cached and never retried - is tracked separately at #18771; this PR stops the notification spam regardless of that. The two are not coupled: #18771 is what keeps producing permanently-broken connections that reach this path, but the NoOp-halt de-dup bypass is a defect on its own and would spam for any other cause of a persistent NoOp halt.

This fix is a stopgap, not the intended final architecture. Quieting is currently decided independently at ~22 `PersistEvent`/`EventBroadcaster.Publish` call sites under `server/`, so every new failure path has to remember to hook into de-dup manually. Moving correlation and quieting into the eventing framework itself is tracked at #20879.

### Testing
`go test ./server/machines/` - adds three `SendEvent` tests:
- `TestSendEvent_NoOpHaltOnDiscoveryIsDeduplicated` - 3 repeated Discovery NoOp-halts yield no event/error, no `UpdateConnectionById` call, persisted status stays `CONNECTED`, machine stays at `InitialState`.
- `TestSendEvent_NoOpHaltOnUserEventPropagates` - a user `Register` NoOp-halt surfaces its Error event + error and writes no status.
- `TestSendEvent_ExitActionFailureHaltsWithoutTransition` - a failing exit action propagates a **non-nil** Error event on a user-initiated event, is suppressed across 3 background `Discovery` re-drives, and writes no status either way. Verified non-vacuous: it fails against the old inline return with `expected a non-nil Error event alongside the error, got nil`.

All 6 pre-existing `SendEvent` tests still pass (9 total). `go build ./server/...`, `go vet ./server/machines/...`, `golangci-lint run ./machines/...` (0 issues), and `go test -race ./server/machines/...` are clean.

**[Signed commits]**

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented failed state-machine transitions from recording or persisting an incorrect connection status.
  * Repeated background discovery failures no longer generate duplicate user-facing error events.
  * User-triggered failures continue to report the original error, even when no state transition occurs.
  * Failed exit actions now halt safely without advancing the machine state.

### DIFF
--- a/server/machines/models.go
+++ b/server/machines/models.go
@@ -179,19 +179,21 @@ func (sm *StateMachine) SendEvent(ctx context.Context, eventType EventType, payl
 	var failureErr error
 	var failureEvent *events.Event
 
-	// haltedNoOp records that an action failed while reporting NoOp: it errored
-	// but emitted no recovery event, so the machine stays in place — no state
-	// transition, nothing to persist. Unlike a recovery-event failure (which
-	// converges on a persisted state, e.g. NOTFOUND, that the de-dup below keys
-	// off), this path never reaches a persisted status. It used to return inline
-	// here, bypassing that de-dup, so a permanently broken connection — e.g. one
-	// whose machine context never initialized, failing GetMachineCtx on every
-	// Discovery — re-persisted and re-broadcast a fresh Error event to the user on
-	// every single background re-drive (issue #20818). Route it through the same
-	// de-dup instead: skip the status write (there is no new state, and writing
-	// the un-advanced CurrentState would corrupt the persisted status) and treat
-	// it as no-change so the repeated background failure is suppressed.
-	haltedNoOp := false
+	// haltedWithoutTransition records that the machine aborted before advancing:
+	// it errored but never reached a new state, so there is no transition and
+	// nothing to persist. Two paths set it — an action failing while reporting
+	// NoOp (an error with no recovery event), and an exit action failing before
+	// entry. Unlike a recovery-event failure (which converges on a persisted
+	// state, e.g. NOTFOUND, that the de-dup below keys off), neither path ever
+	// reaches a persisted status. Both used to return inline, bypassing that
+	// de-dup, so a permanently broken connection — e.g. one whose machine context
+	// never initialized, failing GetMachineCtx on every Discovery — re-persisted
+	// and re-broadcast a fresh Error event to the user on every single background
+	// re-drive (issue #20818). Route them through the same de-dup instead: skip
+	// the status write (there is no new state, and writing the un-advanced
+	// CurrentState would corrupt the persisted status) and treat it as no-change
+	// so the repeated background failure is suppressed.
+	haltedWithoutTransition := false
 
 	for eventType != NoOp {
 		nextState, err := sm.getNextState(eventType)
@@ -228,7 +230,23 @@ func (sm *StateMachine) SendEvent(ctx context.Context, eventType EventType, payl
 			_, event, err = action.ExecuteOnExit(ctx, sm.Context, nil)
 			if err != nil {
 				sm.Log.Error(err)
-				return event, err
+				sm.Log.Debug(event)
+				if failureErr == nil {
+					failureErr = err
+					failureEvent = event
+				}
+				// An exit action failing aborts the transition before the machine
+				// advances — the same "halted without transitioning" shape as a NoOp
+				// halt, so it takes the same route rather than returning inline.
+				// Returning inline here would bypass the de-dup below (re-spamming
+				// the user on a repeatedly re-driven background Discovery) and could
+				// hand callers a nil event alongside a non-nil error: every
+				// ExecuteOnExit in the tree returns a nil event, and callers such as
+				// K8sFSMMiddleware dereference (*event) whenever the error is
+				// non-nil. The terminal failureEvent fallback below guarantees a
+				// non-nil event instead.
+				haltedWithoutTransition = true
+				break
 			}
 		}
 
@@ -245,7 +263,7 @@ func (sm *StateMachine) SendEvent(ctx context.Context, eventType EventType, payl
 					failureEvent = event
 				}
 				if eventType == NoOp {
-					haltedNoOp = true
+					haltedWithoutTransition = true
 					break
 				}
 			} else {
@@ -260,7 +278,7 @@ func (sm *StateMachine) SendEvent(ctx context.Context, eventType EventType, payl
 						failureEvent = event
 					}
 					if eventType == NoOp {
-						haltedNoOp = true
+						haltedWithoutTransition = true
 						break
 					}
 
@@ -279,17 +297,18 @@ func (sm *StateMachine) SendEvent(ctx context.Context, eventType EventType, payl
 	// request (notification spam). Emitting the failure only on an actual status
 	// transition de-duplicates that noise while still surfacing the first
 	// failure. Defaults to true so paths without a provider (or the Exit event)
-	// still report failures — except a NoOp halt, which made no transition (there
-	// is nothing to persist, and writing the un-advanced CurrentState would
-	// corrupt the persisted status), so it starts unchanged so the de-duplication
-	// below suppresses the repeated background failure (issue #20818).
-	statusChanged := !haltedNoOp
+	// still report failures — except a halt without transition, which never
+	// advanced (there is nothing to persist, and writing the un-advanced
+	// CurrentState would corrupt the persisted status), so it starts unchanged so
+	// the de-duplication below suppresses the repeated background failure
+	// (issue #20818).
+	statusChanged := !haltedWithoutTransition
 
 	// originalEventType (not eventType) is compared against Exit here: by the
 	// time the loop finishes, eventType has been overwritten to NoOp, so the
-	// caller's intent is only visible in the captured original. A NoOp halt skips
-	// this block too: it performed no transition, so there is nothing to persist.
-	if sm.Provider != nil && originalEventType != Exit && !haltedNoOp {
+	// caller's intent is only visible in the captured original. A halt without
+	// transition skips this block too: nothing advanced, so nothing to persist.
+	if sm.Provider != nil && originalEventType != Exit && !haltedWithoutTransition {
 		token, _ := ctx.Value(models.TokenCtxKey).(string)
 		connection, _, err := sm.Provider.GetConnectionByID(token, sm.ID)
 
@@ -336,11 +355,12 @@ func (sm *StateMachine) SendEvent(ctx context.Context, eventType EventType, payl
 	// with the non-nil error so callers persist + broadcast it — this is the
 	// core fix for failures previously masked by a later recovery transition.
 	// For a recovery-event failure the connection-status update above has already
-	// moved the connection to its recovery state (e.g. NOTFOUND); for a NoOp halt
-	// no transition happened and statusChanged was forced false. Either way,
-	// suppress the emit when the persisted status did not change, so a repeatedly
-	// re-discovered down connection (or a permanently broken one halting on NoOp)
-	// does not spam a notification on every request.
+	// moved the connection to its recovery state (e.g. NOTFOUND); for a halt
+	// without transition nothing advanced and statusChanged was forced false.
+	// Either way, suppress the emit when the persisted status did not change, so
+	// a repeatedly re-discovered down connection — or a permanently broken one
+	// that halts without ever transitioning — does not spam a notification on
+	// every request.
 	if failureErr != nil {
 		// De-duplicate ONLY the background discovery path. K8sFSMMiddleware
 		// re-runs SendEvent(Discovery) on every request, so a cluster that stays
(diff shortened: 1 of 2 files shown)


leecalcote (commented):
@marblom007 Does this PR take the best approach to handling quieting of events? Ideally, the "event" construct and it's API operations and handlers have a centralized system of event correlation in which notifications are quieted implicitly as part of the  eventing/messaging framework, not as part of one-off consideration given to a specific operational flow.

Faisal77666:
> @marblom007 Does this PR take the best approach to handling quieting of events? Ideally, the "event" construct and it's API operations and handlers have a centralized system of event correlation in which notifications are quieted implicitly as part of the eventing/messaging framework, not as part of one-off consideration given to a specific operational flow.

Agreed a centralized event-correlation/dedup layer is the right long-term design — right now dedup logic is scattered (this PR's statusChanged check, the earlier meshkit-11180 log fix in [#20812](https://github.com/meshery/meshery/pull/20812), etc.), so any new failure path has to remember to hook into it manually, which is fragile.
I checked and there's no existing tracking issue for a centralized eventing/correlation system — closest hits were old UI notification-bar issues ([#220](https://github.com/meshery/meshery/issues/220), [#1459](https://github.com/meshery/meshery/issues/1459), [#20334](https://github.com/meshery/meshery/issues/20334)), unrelated to server-side event dedup.
This PR is scoped to stop the immediate user-facing spam from [#20818](https://github.com/meshery/meshery/issues/20818) without touching the eventing framework. Also traced the root cause locally: InitializeMachineWithContext (server/machines/helpers/helpers.go) calls smInstanceTracker.Add(ID, inst) before checking the Start() error, so a machine whose Context never initialized (e.g. AssignInitialCtx failing against an unreachable cluster) gets permanently cached — every later GetMachineCtx call on it then hits the NoOp-halt path this PR fixes. Ran the full suite on this branch: all 8 SendEvent tests pass (including the 2 new NoOp-halt ones), and go test -race ./server/machines/ is clean.
Building the centralized correlation system would mean redesigning how SendEvent/PersistEvent/EventBroadcaster decide what's dedup-worthy — bigger surface area, probably its own design doc rather than riding along with a bug fix.
I'll open a tracking issue for the centralized approach and reference it here + in [#18771](https://github.com/meshery/meshery/issues/18771), so this fix isn't seen as final architecture, just the stopgap. Does that work, or do you want the centralized design blocked before this merges?

Sohamsig (approved):
went through the diff and both new test in detail the fic looks correct to me: the early `return` NoOp is replaced with `break` + `haltedNoOp` , routed through the existing dedup, and the status-write is properly guarded with `!haltedNoOp` so nothing gets falsely persisted. 

`TestSendEvent_NoOpHaltOnDiscoveryIsDeduplicated` and `TestSendEvent_NoOpHaltOnUserEventPropagates`  covers the two behavioral boundaries nicely, background silence vs. user-initiated propagation.

On @leecalcote's idea about a shared system for handling all notifications, I think that's a good idea for later, but I don't know enough about how big that change would be. One small thing I noticed: this PR just stops the spam for this one case. The real reason connections get stuck broken in the first place seems to be a different bug (#18771), which already has its own fix being worked on. So maybe these two don't need to be linked together, but I could be missing something, happy to hear what you think.

marblom007:
@leecalcote - following through on the tracking issue promised above: **#20879 - [Server] Centralize event correlation and notification quieting in the eventing framework**.

It writes up the concrete evidence for your point: the emit decision is currently made independently at ~22 `PersistEvent`/`EventBroadcaster.Publish` call sites under `server/`, so a producer like `SendEvent` has no way to say "same ongoing condition" - the only lever it has is to return `nil, nil` and drop the failure. The issue scopes what a framework-level layer would need to own (correlation identity, one suppression policy, occurrence roll-up instead of discard, resolution on clear, always-propagate for user-initiated actions) and notes it likely warrants its own design doc rather than riding along with a bug fix.

So this PR is explicitly the stopgap for #20818, not the final architecture, and #20879 now records that. Happy to hold this PR if you would rather the centralized design land first - say the word and I will convert it.

---

@Sohamsig - thanks for the close read of both new tests, and you are not missing anything on the #18771 point: the two are deliberately *not* coupled. The PR's "Scope note" already frames #18771 as the separate, deeper root cause (a broken machine cached by `InitializeMachineWithContext` and never retried) and states that this PR stops the notification spam regardless of whether #18771 is fixed. The dependency runs one way only - #18771 is what keeps producing the permanently-broken connections that hit this path, but the NoOp-halt de-dup bypass in `SendEvent` is a defect on its own and would still spam for any other cause of a persistent NoOp halt. Neither PR blocks the other.

I have added #20879 to the Scope note so the long-term direction is linked from the description too.

marblom007:
Heads-up on a second commit (`39ad9af`) - @Sohamsig this is new since your approval, so please re-check.

While auditing for other instances of the same root cause, the **exit-action branch** of the same loop turned out to be the sibling of this exact defect:

```go
_, event, err = action.ExecuteOnExit(ctx, sm.Context, nil)
if err != nil {
    sm.Log.Error(err)
    return event, err   // <- same inline return, same bypass
}
```

An exit action failing aborts the transition before the machine advances - the identical "halted without transitioning" shape as the NoOp halt - but it returned inline too. Two consequences:

1. It bypasses the `statusChanged` de-dup, so it would spam per-request in exactly the way #20818 describes.
2. Every `ExecuteOnExit` in the tree builds a **nil** event, and `K8sFSMMiddleware` / `addK8SConfig` dereference `(*event)` whenever the error is non-nil - so this was a latent nil-pointer panic, not just spam.

It is **unreachable today**: all 11 `ExecuteOnExit` implementations (`server/machines/actions.go`, `machines/kubernetes/*`, `machines/grafana`, `machines/prometheus`) are `return NoOp, nil, nil`. But leaving it means the first person to write a real exit action reintroduces this bug verbatim, so it is fixed the same way rather than noted: capture the failure, `break`, skip the status write, and let the terminal `failureEvent == nil` fallback guarantee a non-nil Error event. `haltedNoOp` is renamed to `haltedWithoutTransition` now that two paths set it.

`TestSendEvent_ExitActionFailureHaltsWithoutTransition` covers both halves (user-initiated propagates with a non-nil event; 3 background `Discovery` re-drives suppressed; no status write in either case). I checked it is not vacuous - against the old inline return it fails with `expected a non-nil Error event alongside the error (callers dereference it), got nil`.

Local gates: 9/9 `SendEvent` tests pass, `go test -race ./server/machines/...` clean, `golangci-lint run ./machines/...` 0 issues, `go build ./server/...` and `go vet` clean. Description updated to match.

leecalcote:
> @leecalcote - following through on the tracking issue promised above: **#20879 - [Server] Centralize event correlation and notification quieting in the eventing framework**.
> 
> It writes up the concrete evidence for your point: the emit decision is currently made independently at ~22 `PersistEvent`/`EventBroadcaster.Publish` call sites under `server/`, so a producer like `SendEvent` has no way to say "same ongoing condition" - the only lever it has is to return `nil, nil` and drop the failure. The issue scopes what a framework-level layer would need to own (correlation identity, one suppression policy, occurrence roll-up instead of discard, resolution on clear, always-propagate for user-initiated actions) and notes it likely warrants its own design doc rather than riding along with a bug fix.
> 
> So this PR is explicitly the stopgap for #20818, not the final architecture, and #20879 now records that. Happy to hold this PR if you would rather the centralized design land first - say the word and I will convert it.

@marblom007, you're right: this does require a separate design spec. Here it is - [Messaging Framework and Notification Center Design](https://docs.google.com/document/d/1_zpEWBcC6ngOLen_E_nc-octiUdBbpmdwwIX6qzo70M/edit?usp=sharing). Minimally, please ensure that a link to this document is included in the follow up issue. Ideally, please ensure that its contents are digested, reflected upon, and summarized in the issue.

marblom007:
@Sohamsig re-requesting review - flagging that your approval was auto-dismissed rather than withdrawn.

`master` has `dismiss_stale_reviews: true`, so pushing `39ad9af` (and the `f4cc2aa` amend) after your approval on `5a3a451` dismissed it automatically. The PR currently shows no live approval.

The only thing added since you looked is the exit-action commit described above: the same inline-`return` bypass in the `ExecuteOnExit` branch, routed the same way, plus the `haltedNoOp` -> `haltedWithoutTransition` rename and `TestSendEvent_ExitActionFailureHaltsWithoutTransition`. The NoOp-halt logic you already checked is unchanged. CI is green on `f4cc2aa` (14 pass / 0 fail), and CodeRabbit returned "no actionable comments".

No rush from my side - I would rather this go in on a live review than be force-merged past the gate.

leecalcote (approved):

Sohamsig:
Reviewed `f4cc2aa`  confirmed it follows the same pattern as the NoOp-halt fix (capturing the failure into `failureErr/failureEvent`, setting `haltedWithoutTransition`, and breaking instead of returning early), and closes a real nil-pointer crash risk in `ExecuteOnExit`,  Solid fix.

PR #20827, opened 2026-07-20
[UI] Fix icon disappearing on first hover when sidebar is collapsed

Katotodan:
### Description
This PR fixes an issue where a navigation item icon could briefly disappear the first time it is hovered while the sidebar is in the collapsed state.
#### Root cause
When the sidebar is collapsed, the navigation row uses the hoveredId state to determine whether the item is hovered.
<br/>
On the first hover:
- hoveredId changes from null to the current item ID.
- The component re-renders.
- The normal icon is replaced with the hoverIcon.
- The icon wrapper is also replaced during the re-render.
Because the icon element is fully replaced, the icon can momentarily appear to disappear during that transition.
On subsequent hovers, the item is already in the hovered state, so the same visual swap does not occur and the icon remains visible.

### Why it was most noticeable on Telemetry
The issue was particularly noticeable on the Telemetry item because its hover state uses InsertChartIcon, which has very little visual difference from the default icon. That made the element replacement easier to perceive compared to other navigation items that use more distinct hover icons or styling.
### Fix
A fallback icon is now rendered whenever the hoverIcon is unavailable or not visibly rendered. This ensures that:
- the icon remains visible during the hover transition,
- there is no blank state while the component re-renders,
- and the hover behavior appears consistent across navigation items.
### Result
The icon no longer disappears on the first hover when the sidebar is collapsed, including for the Telemetry navigation item.

### Before

https://github.com/user-attachments/assets/ea28d6b0-3830-452f-8356-fdd25057278c

### After

https://github.com/user-attachments/assets/bfe41935-947d-4787-81ff-7dde6401080b

**Note that I refreshed the page at the start of both videos, which is why a blank fallback color appears briefly at the beginning of each video.**


simihablo (approved):
Thanks!

PR #20872, opened 2026-07-21
fix(server): stop auto-registration from proceeding after machine init fails

### LINKED ISSUE #20821: [Server] Unguarded SendEvent on possibly-nil instance in auto_register (dormant latent crash)
### Current Behavior

`auto_register.go` (around line 111) calls `machineInst.SendEvent(...)` with **no nil-check** on `machineInst` — the same defect class as the `DeleteContext` goroutine and the middleware guards fixed in #20812.

Currently **dormant**: every invariant that would make `InitializeMachineWithContext` return nil for this specific call site happens to hold today (`getTypeOfConnection` only yields `"grafana"`/`"prometheus"`, and `initFunc = nil` here means `Start` cannot error). It is a latent server-crash if those invariants ever change.

### Expected Behavior

Add the one-line nil guard consistent with the pattern established in #20812, so this call site is not a latent panic.

### Notes
Surfaced in @leecalcote's out-of-band review of #20812: https://github.com/meshery/meshery/pull/20812#pullrequestreview-4736028511 (finding 7). A local (zero-risk) guard fix exists. Related: #18592.

ty239:
## Summary

This PR fixes #20821

`processRegistration` logged the error from `InitializeMachineWithContext` and then
fell through to `machineInst.SendEvent(...)` on the returned instance. Both of that
function's failure paths return a nil `*StateMachine`, so the fall-through
dereferences nil. This runs in a goroutine with no recover, so the panic takes down
the server process rather than failing the one registration.

Adds `continue` to the existing error branch, so a failed initialization skips that
URL and the loop moves to the next one.

Not reachable today: at this call site `initFunc` is nil so `Start` cannot error,
and `getMachine` only fails on an unparseable UUID, which `generateUUID` (an md5
hex digest) cannot produce. The fix is to keep the control flow correct if those
invariants change — e.g. if `getTypeOfConnection` ever yields a kind beyond
grafana/prometheus.

The earlier revision of this PR added `if machineInst == nil { continue }` after the
error log. That was dropped: `InitializeMachineWithContext` never returns a nil
instance with a nil error (its cache-hit path returns a tracked instance, and the
tracker is only written from one site with an already-validated instance), so the
nil check tested the same condition as the error check. Handling the error is the
fix; the nil guard was dead code beside it.

## Notes for Reviewers

Single line, no behaviour change on any currently reachable path.

Separately, the `Register` `SendEvent` call below discards the non-nil Error event
that `SendEvent` returns on failure, while the `Connect` call two lines down
persists its event for the same failure class. Out of scope here — flagging it for
a follow-up issue.

## Signed commits


KumarNirupam1 (approved):
LGTM
The nil check + continue is correct for the URL loop

ty239:
Hey @KumarNirupam1, should I ping them here to review and merge it, or should I tell them in the comment where I requested to be assigned to the issue?

KumarNirupam1:
> Hey @KumarNirupam1, should I ping them here to review and merge it, or should I tell them in the comment where I requested to be assigned to the issue?

you can ping the maintainers for review and you can also join our the slack community.

ty239:
Hello @leecalcote , I was wondering if you were the person I needed to get my pr merged/checked, or if it's someone else? Sorry for bothering.

Utkarsh-Mauryaa:
@ty239 Could you please include a link to the issue this PR resolves in the PR description? It's a good practice that makes it easier to track related work and provides better context for reviewers.

ty239:
https://github.com/meshery/meshery/issues/20821 @Utkarsh-Mauryaa Here you go. My apologies

Utkarsh-Mauryaa:
@ty239 No worries but it would be better if you link it in the PR description, for eg,
This PR fixes #

ty239:
Of course I add it to the description does everything, thanks, does eveything look good? @Utkarsh-Mauryaa?

Utkarsh-Mauryaa (approved):

hiyach28 (commented):
from next time, please make sure you dont remove the "i signed my commits" checkbox in the PR description.

hiyach28 (changes requested):

server/machines/helpers/auto_register.go
banana-three-join, line 106: Correct. (nil, nil) isn't possible, if the `machineInst` is `nil` then `err != nil` and the idiomatic way to go (no pun intended) about it is to condition the `err` (so after logging the err, you've to add the `continue`)

banana-three-join (changes requested):
Please, review the requested changes @hiyach28 and I have pointed out above.

server/machines/helpers/auto_register.go
hiyach28: Essentially, the same thing @banana-three-join pointed out here: 
Could you move the `continue` into this `err != nil` block after logging the error?
`InitializeMachineWithContext` returns a nil `machineInst` when initialization fails, so once `err` is returned, this iteration should terminate immediately. keeping the `continue` directly on the error path also makes the control flow explicit and avoids treating a failed initialization as though execution can continue normally.

hiyach28 (changes requested):

hiyach28:
My review got deleted due to a glitch on my side. I apologise for that and have made the comments again (though it essentially points to the same thing @banana-three-join has said above), just for the sake of avoiding any confusion. 
Apologies again!

Utkarsh-Mauryaa:
@hiyach28 I can see both of your comments.

banana-three-join:
I've updated the PR (code, description and title) to close this issue. Let me know if there's anything else which needs fixing or not, so I can proceed with the merge. @hiyach28

banana-three-join (approved):

hiyach28 (approved):
LGTM!

banana-three-join:
Ty both! @ty239 @hiyach28

Issue #20879, opened 2026-07-22
[Server] Centralize event correlation and notification quieting in the eventing framework


marblom007:
## Design Specification

The governing design specification for this work is **[Messaging Framework and Notification Center Design](https://docs.google.com/document/d/1_zpEWBcC6ngOLen_E_nc-octiUdBbpmdwwIX6qzo70M/edit?usp=sharing)** (Layer5, 2021-2023). It defines the Meshery event format, the message classification system, and the intended Notification / Operation Center.

A digest of what it settles, what it leaves open, and where it constrains the direction proposed here is in [Relationship to the design specification](#relationship-to-the-design-specification) below. Read that section before implementing - one decision in the specification (CloudEvents alignment) materially changes how correlation identity should be introduced.

## Current Behavior

There is no centralized notion of event correlation or quieting in Meshery's eventing/messaging framework. Every producer independently decides whether an occurrence is worth persisting and broadcasting, so suppression of repeated/duplicate notifications has to be re-implemented per operational flow.

Concretely, the emit decision is spread across ~22 files under `server/` that call `provider.PersistEvent(...)` and `EventBroadcaster.Publish(...)` directly. A representative pair:

```go
// server/handlers/middlewares.go — K8sFSMMiddleware, runs on EVERY /api request
go func(inst *machines.StateMachine) {
    event, err := inst.SendEvent(ctx, machines.Discovery, nil)
    if err != nil {
        _ = provider.PersistEvent(*event, token)
        go h.config.EventBroadcaster.Publish(userUUID, event)
    }
}(inst)
```

The producer (`StateMachine.SendEvent`) has no way to say "this is the same ongoing condition you already told the user about," so the only lever available is for the producer to return `nil, nil` and lie about the failure. That is exactly what `server/machines/models.go` does today: a hand-rolled `statusChanged` check, scoped to the background `Discovery` path only.

Existing one-off quieting/dedup implementations, none of which know about each other:

- `SendEvent`'s `statusChanged` de-dup (`server/machines/models.go`) — suppresses a repeated background `Discovery` failure by comparing persisted connection status against the settled machine state.
- The `meshkit-11180` log-spam fix in #20812.
- The NoOp-halt path routed through that same de-dup in #20826.

Every new failure path has to remember to hook into this manually. Missing the hook is silent and only shows up as user-visible notification spam (#20818: an Error-severity notification per request, indefinitely, for a single permanently broken connection).

## Desired Behavior

[...]

PR #20894, opened 2026-07-22
fix(ui): handle environment creation errors and conditionally close m…

### LINKED ISSUE #20854: Create Environment silently does nothing: submit handler reads 'organization' but form supplies 'organizationId'
## Summary

On `playground.meshery.io` (Meshery `v1.0.60`), creating an environment at `/management/environments` does nothing. The dialog closes as if it succeeded, **no request is sent**, no error is surfaced, and no environment is created.

Users have no way to tell the operation failed.

## Reproduction

1. Sign in to `playground.meshery.io` with the Layer5 Cloud remote provider.
2. Go to **Lifecycle → Environments**.
3. Click **Create**.
4. Type a name (e.g. `fm-envtest-001`). Leave description empty or filled — makes no difference.
5. Click **Save**.

**Expected:** a `POST /api/environments` is issued and the environment appears in the list.

**Actual:** the dialog closes. No `POST` is issued. No error toast. The environment list is unchanged.

Reproduced three times, with real mouse and keyboard input. Verified at the CDP network layer (not just page instrumentation): across the full 124-request log for the interaction, there is **no** `POST`/`PUT`/`PATCH` to `/api/environments` — the last environments-related request is the pre-existing `GET` list call.

Confirmed via the API afterwards that the environment was not created (list count unchanged at 11, new name absent).

## Root cause

`ui/components/environments/index.tsx`

```js
const handleCreateEnvironment = ({ organization, name, description }) => {
  createEnvironment({
    environmentPayload: {
      name: name,
      description: description,
      organization_id: organization,     // <-- always undefined
    },
  })
```

The handler destructures **`organization`** from the submitted form data, but the RJSF form supplies **`organizationId`**. The rendered DOM confirms the field name — the hidden input is `id="Environment_organizationId"`, and RJSF derives ids as `<prefix>_<fieldName>`.

So `organization` is `undefined`, and `organization_id` is always `undefined`.

The RTK wrapper then resolves the organization through three aliases, all of which are absent:

```js
// ui/rtk-query/environments.ts
organizationId:
  queryArg.environmentPayload?.organizationId ||
  queryArg.environmentPayload?.organization_id ||
  queryArg.environmentPayload?.OrganizationID,
```

`organizationId` ends up `undefined`, and the create never reaches the server.

`handleEditEnvironment`, directly below, has the same mismatch — it reads `initialData.organization`.

This looks like fallout from the change that moved these forms to derive the organization from the session as `organizationId`: the form and the schema were updated, the submit handler was not.

## Why it is silent

`handleCreateEnvironment` calls `handleEnvironmentModalClose()` unconditionally on the line after firing the mutation, so the dialog closes regardless of outcome. Combined with the request never being issued, the user sees a successful-looking interaction.

## Suggested fix

1. Destructure `organizationId` in `handleCreateEnvironment` (and use `initialData.organizationId` in `handleEditEnvironment`).
2. Do not close the modal unconditionally — close on success, and surface the error on failure.
3. Add a regression test that asserts a create actually issues the request with a populated organization id. The current unit tests mock the mutation trigger, so this class of defect passes them.

## Related observation

In the same network log, the Environments page queries a **different organization** than the rest of the application:

```
GET /api/workspaces?orgId=e9889a66-...      <- the session's selected organization
GET /api/environments?...&orgId=c5ada327-...  <- a different organization
```

Filed separately.

Myself-Praveen:
**Notes for Reviewers**
- This PR fixes #20854
- Fixes the `handleCreateEnvironment` and `handleEditEnvironment` logic in `ui/components/environments/index.tsx` so that `handleSuccess` is passed as a callback rather than invoked immediately during render.
- Ensures the modal is only closed on a successful response, and correctly formats the error message using `handleError` if the creation/update fails.
- Adds a regression test in `ui/rtk-query/__tests__/environments.test.ts` to assert that `useCreateEnvironmentMutation` correctly issues the request with a populated `organizationId`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment and workspace error notifications with clearer messages and remediation guidance.
  * Failed create or edit operations now keep their dialogs open for correction and retry.
  * Provider responses preserve their original HTTP status instead of defaulting to misleading errors.
  * Unauthorized responses now display visible error notifications.
  * Corrected environment creation payload handling and connection assignment failure reporting.
  * Improved snackbar layout so long messages wrap correctly.

* **Documentation**
  * Clarified commit sign-off requirements and updated the HTTP error response contract.

* **Tests**
  * Expanded coverage for provider status propagation, error details, notifications, dialog behavior, and response formatting.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -203,10 +203,6 @@ make helm-docs      # Generate Helm chart docs
 - Only `utils.Log.Error(err)` renders a MeshKit error's code, cause and remediation; cobra's
   default print shows just the message. In `mesheryctl` commands, log the structured error
   for the user *and* return it for the exit path.
-- Server handler error paths: never reuse another operation's error code, and never hardcode the
-  HTTP status of a provider failure - read it back with `providerStatus(err)`
-  (`server/handlers/utils.go`), which defaults to 502, not 404. Both rules and the JSON wire shape
-  are in `docs/content/en/project/contributing/error-contract.md`.
 - Tests in `*_test.go`; manage deps with `go mod tidy`.
 
 ### JavaScript/React
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -203,6 +203,10 @@ make helm-docs      # Generate Helm chart docs
 - Only `utils.Log.Error(err)` renders a MeshKit error's code, cause and remediation; cobra's
   default print shows just the message. In `mesheryctl` commands, log the structured error
   for the user *and* return it for the exit path.
+- Server handler error paths: never reuse another operation's error code, and never hardcode the
+  HTTP status of a provider failure - read it back with `providerStatus(err)`
+  (`server/handlers/utils.go`), which defaults to 502, not 404. Both rules and the JSON wire shape
+  are in `docs/content/en/project/contributing/error-contract.md`.
 - Tests in `*_test.go`; manage deps with `go mod tidy`.
 
 ### JavaScript/React
--- a/docs/content/en/project/contributing/_index.md
+++ b/docs/content/en/project/contributing/_index.md
@@ -57,6 +57,15 @@ To ensure all your commits are signed, you may choose to add this alias to your 
 Or you may configure your IDE, for example, VSCode to automatically sign-off commits for you:<a href="https://user-images.githubusercontent.com/7570704/64490167-98906400-d25a-11e9-8b8a-5f465b854d49.png" ><img src="https://user-images.githubusercontent.com/7570704/64490167-98906400-d25a-11e9-8b8a-5f465b854d49.png" width="50%"/></a>
 
 </li>
+
+<li>
+Should an unsigned commit slip through anyway, the repository's <code>commit-msg</code> hook
+(installed with the UI dependencies, see <code>ui/.husky/</code>) rejects it locally rather than
+letting the DCO check fail in CI, where the only remedy is rewriting the branch. To sign off
+a commit you have already written:
+
+<pre><code>$ git commit --amend -s --no-edit</code></pre>
+</li>
 </ul>
 </details>
 
--- a/docs/content/en/project/contributing/error-contract.md
+++ b/docs/content/en/project/contributing/error-contract.md
@@ -30,25 +30,33 @@ as JSON before surfacing errors to users.
 
 ```json
 {
-  "error": "Human-readable short description",
-  "code": "meshery-server-1033",
-  "severity": "ERROR",
-  "probable_cause": ["Connection to the remote provider timed out."],
-  "suggested_remediation": ["Verify that Meshery Cloud is reachable."],
-  "long_description": ["Full technical details suitable for logs."]
+  "error": "Unable to create the environment",
+  "code": "meshery-server-1448",
+  "severity": "ALERT",
+  "probableCause": [
+    "Your account does not have permission to create environments in this organization.",
+    "An environment with the same name already exists in this organization."
+  ],
+  "suggestedRemediation": ["Confirm your role in the selected organization."],
+  "longDescription": ["Full technical details suitable for logs."]
 }
 ```
+
+Field names are camelCase, matching the repo-wide wire convention (see
+`AGENTS.md`, "Identifier Naming Conventions"). The three detail fields are
+arrays with **one entry per distinct cause or step** - do not concatenate them
+into a single string; the UI renders each entry as its own bullet.
 
 ### Fields
 
 | Field | Required | Notes |
 |-------|----------|-------|
 | `error` | yes | User-facing message. For MeshKit errors, this is the ShortDescription. |
-| `code` | when available | MeshKit error code (e.g. `meshery-server-1033`). Stable across releases. Use for telemetry, i18n lookup, and programmatic handling. |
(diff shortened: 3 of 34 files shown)


Myself-Praveen:
Hi! I've addressed all the feedback generated by the AI reviewers and all checks are now passing. This PR is ready for human review whenever you have a moment. Thanks.

marblom007:
I've folded the server-side and broader UI failure-reporting work (previously #20896) onto this branch as commit 124add8c so a single PR carries the full fix. #20896 is being closed in this PR's favor.

**What was kept from this PR (unchanged):** your curried `handleError({ error_msg })(error)` signature and the close-modal-only-on-success behaviour in `ui/components/environments/index.tsx`.

**What the fold adds on top:**

- **Server** - environment and workspace handlers now use their own MeshKit error codes (`meshery-server-1446..1461`) with accurate cause/remediation, replacing the reused `ErrGetResult` ("unable to get result", `meshery-server-1033`) that had been mislabelling every environment failure in the logs. HTTP status now reflects the real provider failure via `providerStatus(err)` (defaults to 502, never a hardcoded 404), so a provider 403 no longer reaches the browser as a 404 - which is what let the UI treat the failure as success.
- **environments/index.tsx** - routed your `handleError` through `formatApiError` so the MeshKit code and remediation actually render in the toast (your `normalizeError` surfaced only `data.message`/`data.error`). Also fixed the same-class silent failures the diff left untouched: the fetch-error effects and the connection assign/remove mutators were discarding their rejections.
- **Shared UI** - `meshkitError.ts` tolerates both wire casings and falls back to the raw body (meshery/schemas#1081); `snackbar.tsx` stacks the markdown blocks so the code no longer jams onto the title; `rtkErrorMiddleware.ts` no longer swallows 401s via a phantom `authMiddleware`.
- **Workspaces** - the equivalent honest-failure handling, since the captain asked for both environments and workspaces.

**Blast radius to call out for reviewers:** the 401 change in `rtkErrorMiddleware.ts` widens error surfacing to every endpoint (previously 401s were silently dropped), and the `snackbar.tsx` layout change affects every toast in the app.

Tests: server handlers + httputil unit suites, and the environments/workspaces/snackbar/meshkit/rtk vitest suites all pass locally.

marblom007:
Addressing the two CodeRabbit pre-merge check items:

- **Linked Issues (regression test):** added `test(ui): assert create sends populated organizationId` (commit 27dc9aa) - `ui/components/environments/index.test.tsx` now asserts the create mutation is called with `environmentPayload.organizationId` populated, pinning the #20854 root cause (the submit handler previously read the wrong form field and sent an empty organization). The regression coverage lives at the component level rather than `ui/rtk-query/__tests__/environments.test.ts` because the bug was in the component's submit handler, not the RTK endpoint.
- **Out of Scope (delete-error / server / workspaces):** this is intentional and now documented. The captain scoped this work to make **both** environment and workspace failures legible end to end (honest MeshKit codes + correct HTTP status + a UI that stops reporting success on failure), which is why the combined PR spans the delete path, the server handlers, and workspaces. See the fold summary comment above.

Note the earlier E2E run's failures ("Metrics (Prometheus) page loads" 180s timeout and two relationship-engine tests) were on the pre-fold commit and are pre-existing infra flakes unrelated to environments; the fresh push re-runs them.

@coderabbitai review

marblom007:
Addressed the review in commit e95b536:

**Fixed**
- **Workspace modal closes before the mutation settles (Major):** `handleCreateWorkspace`/`handleEditWorkspace` now close the modal only inside the success `.then()`, matching the environments fix, so a failed create keeps the user's input. Added a deferred-promise regression test asserting the modal stays open on failure and closes on success.
- **`handleAssignConnection` closed before assignment settled:** now `await`s `Promise.allSettled([...])` of the assign/remove mutations before closing, so a rejected assignment surfaces its toast.
- **Read vs write remediation (Minor):** `ErrEnvironmentConnection`/`ErrWorkspaceResource` now tailor probable cause and remediation by action - a denied `list` (read) is told it needs read access, not write.
- **`vi.hoisted()` (Major):** both new component test files hoist their mock fixtures via `vi.hoisted()`.
- **camelCase `totalCount` (Minor):** the connection/workspace count fixtures use `totalCount`, which is also what the components actually read (normalized in `rtk-query/transforms.ts`).
- **Provider-status test coverage:** added `TestGetUsersKeys_*` and `TestGetOrganizations_*` (fallback 502 for an untagged error, verbatim status for a tagged one), mirroring the `GetUserByID` coverage.
- **AGENTS.md (Major, protected file):** reverted the hunk; the same guidance already lives in `docs/content/en/project/contributing/error-contract.md`.

**Intentionally not changed (with reason)**
- **Consolidate the local `handleError` into `useNotificationHandlers().notifyApiError`:** kept the environments handler as-is to preserve this PR's original curried `handleError({error_msg})(error)` structure that reviewers already approved; the workspaces handler keeps its local wrapper for the `updateProgress({ showProgress: false })` call the shared hook does not perform. Reasonable follow-up, but out of scope for a change whose purpose is failure legibility.
- **Snackbar test mocking multiple markdown blocks:** the regression was the message container laying blocks out in a row; the test already asserts the container's `flex-direction: column` + `minWidth`, which is the property that regressed. Rendering multiple mocked children would not strengthen that assertion.
- **401 notification de-duplication in `rtkErrorMiddleware`:** framework-level notification quieting is tracked separately (meshery/meshery#20879); folding it in here would widen scope beyond the reporting fix.

@coderabbitai review

PR #20904, opened 2026-07-23
[Fix] Owner: undefined undefined bug by aligning MesheryPattern with schemas v1beta3

NSTKrishna:
### Description
This PR fixes an issue where the design's info modal in the UI displays **"Owner: undefined undefined"** instead of the actual user's name. 

**Root Cause:**
The `MesheryPattern` struct in `server/models/meshery_pattern.go` was still using the legacy `Owner` field (`json:"owner"`). However, the generated UI client uses the v1beta3 schemas (`@meshery/schemas/mesheryApi`), which migrated this field to `userId` and `user`. Because the server was emitting `"owner"`, the UI could not find the `"userId"` to fetch the user profile, resulting in the undefined display.

**Changes Made:**
1. **Server models (`meshery_pattern.go`)**: 
   - Replaced `Owner *string` with `UserId *core.Uuid json:"userId"` and `User *User json:"user"` to match the v1beta3 schema contract. 
   - Cleaned up the confusing `db:"owner"` tag since this field is handled exclusively by the remote provider (`gorm:"-"`).
2. **Server handlers (`meshery_pattern_handler.go`)**: 
   - Updated `DesignPostPayload` to expect `UserId` instead of `Owner`, aligning the incoming payload with the updated struct.
3. **Server persister & local provider**: 
   - Removed stale `nilOwner` injection logic for locally seeded patterns (since `UserId` being a nil pointer natively handles the "no owner" state).
4. **UI (`LegacyInfoModal.tsx`)**: 
   - Added a defensive guard to the tooltip title to gracefully fall back to `'Owner'` instead of displaying `'Owner: undefined undefined'` if the profile is not yet loaded or missing.

**Before**
Screenshot 2026-07-23 at 8 14 25 AM

**After**

Screenshot 2026-07-23 at 8 03 47 AM

**Notes for Reviewers**

- This PR fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pattern ownership display so tooltips and UI no longer show “undefined” when user profile details are missing.
  * Preserved existing ownership values during batch pattern saves.
* **Improvements**
  * Updated pattern ownership handling across APIs and CLI output to use user ID (and user profile when available) instead of the prior owner field.
  * Refreshed design export/list “Owner” display to reflect the new user-based ownership info, including verbose formatting.


server/handlers/meshery_pattern_handler.go
```diff
@@ -383,7 +384,7 @@ func (h *Handler) handlePatternPOST(
 	mesheryPatternRecord := models.MesheryPattern{
 		ID:          requestPayload.ID,
 		PatternFile: designFile,
-		Owner:       requestPayload.Owner,
+		UserId:      requestPayload.UserId,
```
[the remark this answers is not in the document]
NSTKrishna: Actually, this is a pre-existing security flaw inherited from the old owner field. Since fixing it requires changing how the handler authenticates and overrides payloads, I will open a separate dedicated PR to address this security concern so it doesn't block this UI bug fix.

server/handlers/meshery_pattern_handler.go
ritzorama, line 1: This smells like caca, @NSTKrishna. 

You're still not grasping the concept of this project being schema-driven.

ritzorama (changes requested):

ritzorama:
@NSTKrishna where is the github issue that this addresses?

server/handlers/meshery_pattern_handler.go
NSTKrishna: I've force-pushed an update that is strictly schema-driven:

- Request (meshery_pattern_handler.go): Completely removed Owner and UserID from DesignPostPayload and the handler logic. The client can no longer set ownership; the server handles it natively via the auth token.
- Response (meshery_pattern.go): Kept the update to json:"userId" so the local struct correctly mirrors the v1beta3 OpenAPI schema for the response, which fixes the undefined undefined bug in the UI.
- Consumers: Cleaned up the old Owner field references across the CLI and persister.

NSTKrishna:
> @NSTKrishna where is the github issue that this addresses?

this is an independent pr that fix UI displays Owner: undefined undefined

PR #20914, opened 2026-07-23
fix: filter annotation components before validation

### LINKED ISSUE #20855: Design deploy fails with 'apiVersion is required' on isAnnotation components, behind a misleading error
## Summary

Deploying a design that contains annotation components (`metadata.isAnnotation: true`) fails the whole deploy with `apiVersion is required`, even though annotation components are explicitly meant to be excluded from provisioning.

The error the user is shown is also misleading: the surfaced `probableCause` talks about environment assignment, while the actual cause is buried in `longDescription`.

## Reproduction

Design: `ECommerce Reference Architecture v2` (`designs.meshery.io/v1beta3`), 59 components:

| kind | version | count | isAnnotation | configuration |
|---|---|---|---|---|
| Deployment | `apps/v1` | 15 | false | populated |
| Service | `v1` | 13 | false | populated |
| Namespace | `v1` | 1 | false | populated |
| Container | `core.meshery.io/v1alpha1` | 15 | **true** | `{}` |
| Container | **`""` (empty)** | 15 | **true** | `{}` |

`POST /api/pattern/deploy?contexts=<ctx>&dryRun=true` with the design as-is returns **500**:

```json
{
  "error": "Unable to deploy the selected design \"ECommerce Reference Architecture v2\"",
  "code": "meshery-server-1073",
  "probableCause": [
    "Connection Error: There was an error connecting to the selected target platform (i.e. Kubernetes cluster(s))..This connection might not be assigned to the selected environment."
  ],
  "suggestedRemediation": [
    "Verify that the Kubernetes connection status is 'Connected' or try uploading a new kubeconfig..Assign the current Kubernetes connection to the selected environment."
  ],
  "longDescription": ["apiVersion is required"]
}
```

## Counterfactual

Filtering out only the components with `metadata.isAnnotation === true` and re-posting the identical request returns **200** with a full per-component dry-run result. Nothing else changed.

So the annotation components are what trigger `apiVersion is required` — almost certainly the 15 whose `component.version` is the empty string, which yields an empty `apiVersion` on conversion.

## Why this is surprising

`processAnnotations()` in `server/models/pattern/stages/provision.go` already strips these:

```go
func processAnnotations(pattern *pattern.PatternFile) {
	components := []*component.ComponentDefinition{}
	for _, component := range pattern.Components {
		if !component.Metadata.IsAnnotation {
			components = append(components, component)
		}
	}
	pattern.Components = components
}
```

and it is called from both `Provision` (provision.go) and `DryRun` (dry_run.go). `validateWorkload()` in `stages/validate.go` also has an explicit annotation carve-out.

The failure therefore happens in a stage that runs **before** annotations are filtered. That earlier stage should either apply the same `isAnnotation` carve-out, or annotations should be stripped once at the front of the chain rather than separately inside two later stages.

## Two defects here

1. **Annotation components reach a stage that requires a valid `apiVersion`.** A design saved by Kanvas with annotation nodes cannot be deployed.
2. **The surfaced error is misleading.** `meshery-server-1073` presents a connection/environment `probableCause` and `suggestedRemediation` that have nothing to do with the actual failure, while the real reason (`apiVersion is required`) is only in `longDescription`. Anyone following the stated remediation will go assign connections to environments and get nowhere.

## Suggested fix

- Filter `isAnnotation` components once, early in the deploy chain, so every downstream stage sees a consistent component set. Failing that, add the same carve-out to the stage that performs the apiVersion check.
- Consider rejecting or normalising components with an empty `component.version` at save time, so designs cannot be persisted in this state.
- Surface `longDescription` in the user-facing error when it contains the concrete cause, rather than leading with a generic connection hypothesis.

pratik340:
**Notes for Reviewers**

This PR moves annotation filtering before the validation stage so annotation components don't reach validation with an empty apiVersion

It adds a FilterAnnotations stage before `Validator` and removes the duplicate filtering from DryRun and Provision

Also added a test to make sure annotation components are filtered while regular components are kept.

- This PR fixes #20855 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pattern processing by filtering annotation components before validation, dry-run, and provisioning operations.
  * Prevented annotation components from affecting dry-run and provisioning results.
  * Added coverage to verify annotations are removed while regular components remain intact.

### DIFF
--- a/server/handlers/design_engine_handler.go
+++ b/server/handlers/design_engine_handler.go
@@ -268,6 +268,7 @@ func _processPattern(opts *patterncore.ProcessPatternOptions) (map[string]interf
 			Add(stages.Format()).
 			// Add(stages.ServiceIdentifierAndMutator(sip, sap)).
 			Add(stages.Filler(opts.SkipPrintLogs)).
+			Add(stages.FilterAnnotations()).
 			// Calling this stage `The Validation stage` is a bit deceiving considering
 			// that the validation stage also formats the `data` (chain function parameter) that the
 			// subsequent stages depend on.
--- a/server/models/pattern/stages/dry_run.go
+++ b/server/models/pattern/stages/dry_run.go
@@ -17,7 +17,6 @@ func DryRun(_ ServiceInfoProvider, act ServiceActionProvider) ChainStageFunction
 			act.Terminate(err)
 			return
 		}
-		processAnnotations(data.Pattern)
 
 		resp, err := act.DryRun(data.Pattern.Components)
 		if err != nil {
--- a/server/models/pattern/stages/provision.go
+++ b/server/models/pattern/stages/provision.go
@@ -29,8 +29,6 @@ func Provision(prov ServiceInfoProvider, act ServiceActionProvider, log logger.H
 			return
 		}
 
-		processAnnotations(data.Pattern)
-
 		// Create provision plan
 		plan, err := planner.CreatePlan(*data.Pattern, prov.IsDelete())
 		if err != nil {
@@ -87,6 +85,23 @@ func Provision(prov ServiceInfoProvider, act ServiceActionProvider, log logger.H
 
 		if next != nil {
 			next(data, mergeErrors(errs))
+		}
+	}
+}
+
+func FilterAnnotations() ChainStageFunction {
+	return func(data *Data, err error, next ChainStageNextFunction) {
+		if err != nil {
+			if next != nil {
+				next(data, err)
+			}
+			return
+		}
+
+		processAnnotations(data.Pattern)
+
+		if next != nil {
+			next(data, nil)
 		}
 	}
 }
--- a/server/models/pattern/stages/provision_test.go
+++ b/server/models/pattern/stages/provision_test.go
@@ -1,0 +1,47 @@
+package stages
+
+import (
+	"testing"
+
+	"github.com/meshery/schemas/models/v1beta2/component"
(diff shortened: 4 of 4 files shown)


pratik340:
Hi I’ve submitted this fix for #20855 by moving annotation filtering before validation and adding a regression test. I’d appreciate feedback on whether this approach aligns with the intended processing flow. Happy to make any changes needed.

ritzorama (changes requested):

pratik340:
Thanks for the review, I’ll add the build output and test results shortly

pratik340:
Thanks for the review! Here are my local build and test results:
PS C:\Users\Hello\meshery> go test ./server/models/pattern/stages/... -v
=== RUN   TestFilterAnnotations
--- PASS: TestFilterAnnotations (0.00s)
PASS
ok      github.com/meshery/meshery/server/models/pattern/stages 6.471s
PS C:\Users\Hello\meshery> go build ./server/...
PS C:\Users\Hello\meshery> go build ./server/...; Write-Host "Exit code: $LASTEXITCODE"
Exit code: 0

pratik340:
Hi @ritzorama , just following up on this PR.

I've addressed the requested changes and also verified that:
- `go test ./server/models/pattern/stages/... -v` passes
- `go build ./server/...` completes successfully

Whenever you have time, could you please take another look? Thanks!

pratik340:
hi @ritzorama , could you please take a look into this

PR #20915, opened 2026-07-23
fix(server): resolve resource leak in design engine loop by closing c…

Myself-Praveen:
**Notes for Reviewers**

- This PR fixes a resource leak in `server/handlers/design_engine_handler.go`. 
- `meshes.CreateClient()` was being called inside a loop over `ccp.Hosts` and cleaned up using `defer`. Because `defer` executes when the surrounding function returns (rather than at the end of each loop iteration), this would cause memory and mesh client connections to remain open until the entire function finished, leading to resource leaks for multi-host deployments.
- Fixed by removing the `defer func() { _ = mClient.Close() }()` block and adding explicit `_ = mClient.Close()` calls where the client finishes its work inside the loop.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource cleanup during service provisioning.
  * Ensured connections are closed properly when provisioning or component processing encounters an error.

### DIFF
--- a/server/handlers/design_engine_handler.go
+++ b/server/handlers/design_engine_handler.go
@@ -568,9 +568,6 @@ func (sap *serviceActionProvider) Provision(ccp stages.CompConfigPair) ([]patter
 		if err != nil {
 			return nil, fmt.Errorf("error creating a mesh client: %v", err)
 		}
-		defer func() {
-			_ = mClient.Close()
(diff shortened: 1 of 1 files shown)


ritzorama (changes requested):
Show your build output.

Include test coverage and copy/paste it's output, too.

Issue #20897, opened 2026-07-22
[UI] TypeError: Cannot read properties of undefined (reading 'find') in dashboard


victoriacheng15:
### Current Behavior
Clicking on any resource card (such as a Namespace, Workload, or Storage component) in the dashboard detail view while the `connections` query is still in a loading state triggers a runtime crash. This crash occurs on the official production deployment (installed via the remote Helm registry).

The screen displays a TypeError:
`Cannot read properties of undefined (reading 'find')` at `components/dashboard/view.tsx:95` because `connections` defaults to `[]` initially, which does not have a `.connections` array.

### Desired Behavior
The detail card should load without crashing. If the connections list is not yet loaded, the `connection` object should safely resolve to `undefined` and fall back to the disconnected status (`CONNECTION_STATES.DISCONNECTED`) rather than throwing an exception.

### Screenshots/Logs

<img width="1064" height="518" alt="Image" src="https://github.com/user-attachments/assets/7a404d6f-3fc7-4c83-9001-85ba80d268a2" />

### Environment
- Browser: Chrome
- Host OS: Linux
- Meshery Server Version: stable-v1.0.62
- Platform: Kubernetes (k3s)

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)

- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

emphor11:
@victoriacheng15 I would like to work on this, could you assign it to me

victoriacheng15:
> [@victoriacheng15](https://github.com/victoriacheng15) I would like to work on this, could you assign it to me

Hi @emphor11 ! Thanks for offering to help. I've already opened a PR.

PR #20917, opened 2026-07-23
fix(ui): resolve TypeError on dashboard connection status

Myself-Praveen:
**Notes for Reviewers**

This PR fixes a bug causing a white screen crash: `[UI] TypeError: Cannot read properties of undefined (reading 'find') in dashboard detail view`.

In `ui/components/dashboard/view.tsx`, `useGetConnectionsQuery` can default to returning an empty array `[]` before the data fully loads. The code attempted to call `connections?.connections.find(...)`. Because `[].connections` is `undefined`, the `.find()` method call threw a TypeError and crashed the dashboard.

This is fixed by adding optional chaining (`connections?.connections?.find`), which gracefully allows the component to fall back to `CONNECTION_STATES.DISCONNECTED` instead of crashing.

- This PR fixes #20899

### DIFF
--- a/ui/components/dashboard/view.tsx
+++ b/ui/components/dashboard/view.tsx
@@ -92,7 +92,7 @@ const View = ({ setView, resource, k8sConfig }: DashboardViewProps) => {
   if (!resource) return null;
 
   const context = getK8sContextFromClusterId(resource.cluster_id, k8sConfig);
-  const connection = connections?.connections.find((conn) => conn.id === context?.connectionId);
+  const connection = connections?.connections?.find((conn: any) => conn.id === context?.connectionId);
   const connectionStatus = connection?.status || CONNECTION_STATES.DISCONNECTED;
   const iconSrc = normalizeStaticImagePath(resource.component_metadata?.styles?.svgColor);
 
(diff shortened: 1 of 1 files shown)


Myself-Praveen:
Hi! 
Just a quick note that the linters and unit tests are all green! 
I noticed the End-to-End Cypress tests (`Performance dashboard controls` and `Charts (Grafana) page loads`) flaked again. Since this PR only adds an optional chaining operator (`?.`) to prevent a crash when the connection data hasn't loaded yet, it shouldn't be related to those E2E flakes.
Happy to re-trigger the workflows if needed!

Myself-Praveen:
It looks like this PR fixes the exact crash reported in **#20899**! 
I've updated the PR description to formally link the issue (Resolves #20899), so it should auto-close once this is merged. Let me know if you need any changes!

KumarNirupam1:
The fix itself is correct —-`connections` defaults to `[]`, so `connections?.connections.find(...)` still throws when `.connections` is undefined. The extra `?.` prevents the crash.

However, this looks like a duplicate of #20899, which:
- links the tracking issue (#20897)
- makes the same optional-chaining fix without introducing `(conn: any)`
- adds a regression test and before/after demos

Myself-Praveen:
Ah, you're totally right! I actually just realized @victoriacheng15 beat me to this exact bug yesterday in #20899. 
Closing this as a duplicate so we can merge her PR with the regression tests. Great catch, thank you!

PR #20922, opened 2026-07-23
[mesheryctl] Fix design export collision rename dropping the extension

### LINKED ISSUE #20889: [mesheryctl] design export renames collisions to "(1)report" and drops the file extension
### Current Behavior

When `mesheryctl design export` writes to a directory that already contains the target file (for example, re exporting the same design into the same folder), `getUniqueFilename` is meant to produce `report(1).yaml`. It looks like it instead produces `(1)report` the counter lands at the front and the `.yaml` extension is dropped.

The cause seems to be in `getUniqueFilename` at `mesheryctl/internal/cli/root/design/export.go`:

```go
base, ext := filepath.Split(strings.TrimSuffix(filename, filepath.Ext(filename)))
```

`filepath.Split` returns `(dir, file)`, not `(name, ext)`. So for `report.yaml`, `base` becomes `""` (the directory) and `ext` becomes `"report"` (the name), and the format string `"%s(%d)%s"` renders `(1)report`. Wanted to check whether this is intentional before proposing a change.

### Expected Behavior

A collision should yield `report(1).yaml`, `report(2).yaml`, … extension preserved, counter as a suffix on the name.

### Screenshots/Logs

Running the exact function against a temp dir:

```
input : .../report.yaml
got   : .../(1)report
want  : .../report(1).yaml
```

Splitting the name and extension the other way `ext := filepath.Ext(filename)` then `base := strings.TrimSuffix(filename, ext)`  returns `report(2).yaml` when both `report.yaml` and `report(1).yaml` already exist, which looks like the intended behavior.

### Environment

- **Host OS:** any

---

Priyanshu6968:
**Notes for Reviewers**

- This PR fixes #20889 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved exported design filename handling, including correct extensions and duplicate-name numbering.
  * Prevented exported files from overwriting existing files by selecting the next available filename.

* **Tests**
  * Added coverage for unique filenames, including no-collision, single-collision, and multiple-collision scenarios.

### DIFF
--- a/mesheryctl/internal/cli/root/design/export.go
+++ b/mesheryctl/internal/cli/root/design/export.go
@@ -337,7 +337,8 @@ func selectPatternPrompt(patterns []models.MesheryPattern, baseURL string) (mode
 }
 
 func getUniqueFilename(filename string) string {
-	base, ext := filepath.Split(strings.TrimSuffix(filename, filepath.Ext(filename)))
+	ext := filepath.Ext(filename)
+	base := strings.TrimSuffix(filename, ext)
 	for i := 1; ; i++ {
 		if _, err := os.Stat(filename); os.IsNotExist(err) {
 			break
--- a/mesheryctl/internal/cli/root/design/export_test.go
+++ b/mesheryctl/internal/cli/root/design/export_test.go
@@ -1,0 +1,45 @@
+package design
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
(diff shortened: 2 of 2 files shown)


ritzorama (changes requested):
@Priyanshu6968 don't use parenthesis in the filename formatting.

ritzorama:

PR #20923, opened 2026-07-23
fix(server): guard nil state machine in DeleteContext

### LINKED ISSUE #20817: [Server] DeleteContext spawns an unguarded nil SendEvent goroutine (server crash) + err data race
### Current Behavior

`DeleteContext` (`server/handlers/contexts_handler.go`, the goroutine near line 123) spawns:

```go
    event, err = inst.SendEvent(req.Context(), machines.Delete, nil)
    ...
}(inst)
```

with **no nil-check on `inst`**. `InitializeMachineWithContext` returns a nil instance when the machine cannot be built (e.g. an unreachable/invalid cluster; see also #18771). A nil `inst` here dereferences `sm.Provider`/`sm.mx` inside an **unrecovered goroutine**, so the panic crashes the **entire Meshery Server process**, not just the request.

Second, independent defect in the same goroutine: it assigns to the enclosing function's `err` via `=` (not `:=`) while the enclosing function reads that same `err` immediately after spawning the goroutine — a **data race**.

This is the same nil-panic defect class that #20812 fixed in the two K8s middlewares, left live in this sibling handler.

### Expected Behavior

- Nil-guard `inst` before driving it (skip the FSM `Delete` event when the machine could not be built).
- Give the goroutine its own local error variable so it does not race the enclosing function's `err`.

### Notes
Surfaced in @leecalcote's out-of-band review of #20812: https://github.com/meshery/meshery/pull/20812#pullrequestreview-4736028511 (finding 2). A local fix exists (nil-guard + local err var) but has not been pushed. Related: #18592, #18775, #19115.

Vanshikadahaliya:
## Description

This PR fixes #20817.

### Summary
This PR addresses two issues in the `DeleteContext` handler:

- Added a nil check before invoking `SendEvent` to prevent a potential nil pointer dereference and server crash when the state machine initialization fails.
- Replaced the shared `err` variable inside the goroutine with a local `sendErr` variable to eliminate a potential data race.

These changes preserve the existing behavior while making the handler more robust and thread-safe.

### Testing

The following commands were executed:

```bash
go test ./...
go test -race ./...
```

Some existing build/test failures were observed in unrelated packages. No new failures were introduced by this change.

**Notes for Reviewers**

- This PR fixes #20817.

**Signed commits**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved context deletion handling to safely exit when the underlying state machine instance isn’t available.
  * Enhanced error reporting when the deletion event can’t be sent, including additional debug details about the result.
  * Preserved reliable connection tracking cleanup after successful event dispatch.


ritzorama (changes requested):

Vanshikadahaliya:
> Include your build output and copy/paste your test results.

Build and test results:

```text
END-TO-END TESTS

63 passed
0 failed
0 flaked
7 skipped

Duration: 3m 54s

Overall Result: All tests passed.
```

I also ran the relevant build/tests locally while updating the implementation.

server/handlers/contexts_handler.go
```diff
@@ -120,28 +120,28 @@ func (h *Handler) DeleteContext(w http.ResponseWriter, req *http.Request, _ *mod
...
+if !mhelpers.HasMachineContext(inst) {
+	smInstanceTracker.Remove(connectionUUID)
+} else {
+	go func(inst *machines.StateMachine) {
+		event, sendErr := inst.SendEvent(req.Context(), machines.Delete, nil)
+		if sendErr != nil {
+			h.log.Error(sendErr)
+			h.log.Debug(event)
+			return
+		}
```
saurabhraghuvanshii, line 140: @Vanshikadahaliya What changes did you make? I'm only seeing that you renamed the variable from `err` to `sendErr`.

saurabhraghuvanshii (changes requested):

server/handlers/contexts_handler.go
Vanshikadahaliya: @saurabhraghuvanshii 
 Thanks for pointing that out. When I opened this PR, it included the nil state machine guard along with the scoped sendErr variable. After syncing with the latest master, the upstream changes introduced the HasMachineContext handling, which superseded the original guard. As a result, the remaining change is the scoped sendErr variable inside the goroutine. Please let me know if you'd like me to revise the implementation further.

saurabhraghuvanshii (approved):

PR #20925, opened 2026-07-23
[Server] Return correct HTTP status codes for encoding errors and wrong method

shivv23:
**Notes for Reviewers**

Corrects three incorrect HTTP status codes:
- `ExtensionsVersionHandler`: returns 404 on JSON encode failure — changed to 500 (server error)
- `GetServerVersion`: returns 404 on JSON encode failure — changed to 500 (server error)
- `AdaptersHandler`: returns 404 for non-GET requests — changed to 405 Method Not Allowed

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected error responses when extension or server version data cannot be encoded, returning **500 Internal Server Error**.
  * Updated adapter endpoint responses for unsupported HTTP methods to return **405 Method Not Allowed** instead of **404 Not Found**.
  * Improved HTTP status accuracy for clearer API error handling.

### DIFF
--- a/server/handlers/extensions.go
+++ b/server/handlers/extensions.go
@@ -79,10 +79,10 @@ func (h *Handler) ExtensionsVersionHandler(w http.ResponseWriter, _ *http.Reques
 	if provider.GetProviderType() == models.LocalProviderType {
 		err := json.NewEncoder(w).Encode("extension not available for current provider")
 		if err != nil {
-			h.log.Error(models.ErrEncoding(err, "extension version"))
-			writeMeshkitError(w, models.ErrEncoding(err, "extension version"), http.StatusNotFound)
-		}
-		return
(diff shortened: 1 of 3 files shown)


SuryanshGarg04 (commented):
Nice fix @shivv23  , A few small suggestions:
1. In `ExtensionsVersionHandler` and `ServerVersionHandler`, consider using `json.Marshal` before writing to `w`. With `json.NewEncoder(w).Encode()`, the response may already be committed if encoding fails, so the status can no longer be changed to `500`.

2. +1 to using `writeMeshkitError` for the `405` in `AdaptersHandler`. This keeps error responses structured and consistent for UI/clients.

3. Looks like `extensions.go` around lines 82–85 has a small indentation issue.

PR #20926, opened 2026-07-23
[Models] Add Artifact Hub and GitHub connection definitions

leecalcote:
**Notes for Reviewers**

Adds connection definitions for **Artifact Hub** and **GitHub**, modelled on `KubernetesConnection.json`.

Both are already first-class registrants in Meshery - models and components are generated from Artifact Hub packages (`MeshModelGenerationHandler`) and from GitHub-hosted manifests (meshkit `generators/github`) - but neither had a connection definition. So both kinds surfaced with no icon, no description, no connection or credential schema, and no `transitionMap` ("No transitions Available" in the status dropdown, exactly as the fallback note in `ui/pages/_app.tsx` called out).

### What's added

| | `kind` | `type` | `subType` | initial `status` |
|---|---|---|---|---|
| Artifact Hub | `artifacthub` | `source` | `registry` | `registered` |
| GitHub | `github` | `source` | `git` | `registered` |

- **`ArtifactHubConnection.json`** - `connectionSchema` takes the Artifact Hub endpoint (so a self-hosted instance works, not just artifacthub.io); `credentialSchema` takes the optional API key id/secret pair, since the public API needs no auth.
- **`GitHubConnection.json`** - `connectionSchema` takes the repository URL plus an optional branch; `credentialSchema` takes an optional access token (required only for private repos).
- Both use the icons **already in this repo** under `docs/static/v0.9/extensions/models/{artifact-hub,github}/icons/{color,white}/`.
- `transitionMap` on both matches the default connection state machine exactly: `discovered → registered | ignored`, `registered → connected | ignored`, `connected → disconnected | deleted`, `disconnected → connected | deleted`. Copy is written as consequence statements (the modal title owns the "Are you sure" framing), consistent with the Grafana/Prometheus definitions.

### Decision worth a reviewer's eye: `type: "source"`

`connection.yaml`'s docstring lists `platform, telemetry, collaboration` for `type` and `cloud, identity, metrics, chat, git, orchestration` for `subType` - neither list covers "a thing Meshery sources models from". Meshery Cloud already persists `ConnectionType: "source"` / `SubType: "git"` for GitHub connections (`server/handlers/github.go`), so these definitions match that rather than inventing a second wire identity for the same kind. `subType: "registry"` extends the set for Artifact Hub, mirroring the `Type: "registry"` already used for registrant connections in `server/models/meshmodel/core/register.go`. The contributor docs table is updated to reflect both. Happy to change if we'd rather force these into `collaboration`.

### Two supporting fixes, so the definitions aren't dead on arrival

Out of scope of "add two JSON files", but shipping without them would ship a known break:

1. **`server/machines/helpers/helpers.go`** - `getMachine` hardcoded an allowlist of `kubernetes`/`grafana`/`prometheus` and returned `meshery-server-1218` ("invlaid type requested") for anything else. The Create Connection wizard is registry-driven (`buildConnectionWizardKindConfigs` lists *every* registered definition), so simply shipping a definition surfaced the kind in the wizard and then 500'd on register. Kinds now fall through to the default machine; only the `REGISTERED`-state verification action stays kind-specific (Grafana/Prometheus probe their endpoint; Kubernetes keeps its bespoke machine). This deletes a second, silently divergent source of truth for which kinds exist.
   - Note: `machines.ErrInvalidType` is left in place but is now unreferenced. I did not remove it to avoid error-code/doc-export churn in this PR - say the word if you'd rather it go.
2. **`ui/components/connections/ConnectionStateTransitionModal.tsx`** - the modal titled its dialog by title-casing the kind slug, which reads **"Github"** and **"Artifacthub"**. The definition's authored `name` is now seeded onto `connectionMetadataState` (`ui/pages/_app.tsx`) and preferred; the title-cased slug remains the fallback for kinds without a definition. Titles now read "Delete GitHub connection?" / "Delete Artifact Hub connection?".

### Docs

`docs/content/en/project/contributing/models/connections.md`: identity table and shipped-definitions list updated for both kinds; `type`/`subType` example lists extended; a new "Lifecycle" step documents that the default state machine needs **no server code** and which transitions it implements.

### Testing

All run locally:

- `npx vitest run components/connections utils/__tests__` → **38 files, 535 tests passed**. The definition-copy drift guard (`it.each` over the shipped JSONs) now covers both new files; new cases assert the display-name title and its fallback.
- `go test ./server/machines/... ./server/handlers/...` → **pass**. New `TestGetMachineCoversEveryDefinitionKind` asserts every definition-backed kind resolves to a machine with a `CONNECTED` action, and that only Grafana/Prometheus carry a `REGISTERED` action.
- `golangci-lint run ./server/machines/...` → **0 issues**. `npx eslint` on touched UI files → clean.
- Icons rendered to PNG and eyeballed on both light and dark backgrounds.

Not run: an end-to-end wizard pass against a live server (no cluster/provider in this environment). Worth a reviewer confirming both kinds appear in **Connections → Create Connection** with their icons and that registering one lands a `connected` row.

## Summary by CodeRabbit

- **New Features**
  - Added Artifact Hub and GitHub connection definitions, including configuration, credentials, lifecycle states, and UI branding.
  - Transition dialogs now display connection-authored names (with fallback to a readable default when no name is provided).

- **Documentation**
  - Updated connection contributing guidance to expand “Identity” examples, list additional shipped connection kinds, and document the full definition lifecycle with default state transitions.

- **Bug Fixes**
  - Improved connection lifecycle handling to cover newly added (and future) connection kinds consistently.
  - Fixed transition and delete dialog titles/descriptions for Artifact Hub and GitHub to match authored labels.

### DIFF
--- a/docs/content/en/project/contributing/models/connections.md
+++ b/docs/content/en/project/contributing/models/connections.md
@@ -83,16 +83,18 @@ Below is a complete, minimal definition for a hypothetical telemetry backend. It
 These three fields identify the Connection and determine how the wizard treats it:
 
 - **`kind`** - the genre of Connection (e.g. `grafana`, `prometheus`, `kubernetes`). The wizard groups credentials and renders icons by `kind`.
-- **`type`** - a broad classification: `platform`, `telemetry`, `collaboration`, and so on.
-- **`subType`** - a finer classification: `orchestration`, `metrics`, `git`, `chat`, and so on.
+- **`type`** - a broad classification: `platform`, `telemetry`, `source`, `collaboration`, and so on.
+- **`subType`** - a finer classification: `orchestration`, `metrics`, `git`, `registry`, `chat`, and so on.
 
 Together they let the UI target a specific Connection with a [custom wizard extension](#advanced-customizing-the-wizard) when the generic flow is not enough. For reference, the definitions Meshery ships with:
 
-| Connection | `kind`       | `type`      | `subType`       | initial `status` |
-| ---------- | ------------ | ----------- | --------------- | ---------------- |
-| Kubernetes | `kubernetes` | `platform`  | `orchestration` | `discovered`     |
-| Grafana    | `grafana`    | `telemetry` | `metrics`       | `registered`     |
-| Prometheus | `prometheus` | `telemetry` | `metrics`       | `registered`     |
+| Connection   | `kind`        | `type`      | `subType`       | initial `status` |
+| ------------ | ------------- | ----------- | --------------- | ---------------- |
+| Kubernetes   | `kubernetes`  | `platform`  | `orchestration` | `discovered`     |
+| Grafana      | `grafana`     | `telemetry` | `metrics`       | `registered`     |
+| Prometheus   | `prometheus`  | `telemetry` | `metrics`       | `registered`     |
+| Artifact Hub | `artifacthub` | `source`    | `registry`      | `registered`     |
+| GitHub       | `github`      | `source`    | `git`           | `registered`     |
 
 ### Lifecycle: `status` and `transitionMap`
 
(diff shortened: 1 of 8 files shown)


leecalcote:
I wonder if it isn't helpful to differentiate "managed" and "unmanaged" connections.

fitzergerald:
> I wonder if it isn't helpful to differentiate "managed" and "unmanaged" connections.

I would like this, but I don't think that the existing category and sub-category properties are the precise right fit. I think this needs a type or subtype or similar sort of property to properly denote it.

fitzergerald:
Should this be designated by a capability?

leecalcote:
Split the managed vs unmanaged discussion out to #20930 so it doesn't hold this up - labeled `issue/design required`.

It captures the full analysis: that the concept exists in the schema `description`, the docs alert, and two UI strings but has no field anywhere; why @fitzergerald is right that `type`/`subType` are the wrong home (the wizard dedupes on `kind|type|subType`, so it would render two Kubernetes cards); and why the capability construct is the better answer - `entityState: [declaration, instance]` already encodes the kind-vs-instance split, and `component.yaml`/`model.yaml` already carry capabilities while connections are the outlier that doesn't.

Recommendation there is capabilities as the substance plus one server-derived `managementMode` enum as the queryable projection, with the capability-vocabulary question flagged as the thing to settle first.

PR #20949, opened 2026-07-24
Drive connection kinds from the registry instead of a hardcoded list

aabidsofi19:
`ui/utils/Enum.ts` hardcoded the set of connection kinds in two exports — `CONNECTION_KINDS_DEF` (which already carried a `// Remove this` note) and `CONNECTION_KINDS`. Any kind outside that frozen five was invisible to the UI. The registry is already the source of truth, served by `GET /api/meshmodels/connections`.

- Remove both hardcoded exports.
- Populate the Connections table Kind filter from `useListConnectionDefinitionsQuery` instead of `Object.entries(CONNECTION_KINDS)`, so it reflects whatever is actually registered.
- Drive the per-kind metadata backfill in `_app.tsx` off the fetched definitions rather than a static list. A kind with no legacy `<Kind>Connection` mesh-model component is simply skipped.
- Replace the ~45 remaining references, which are semantic special-cases rather than enumeration (Kubernetes cluster-import flow, Grafana/Prometheus telemetry and ping hooks, per-kind icons), with `CoreConnectionKinds` from `@meshery/schemas`.
- `Enum.ts` re-exports `CoreConnectionKinds` rather than each call site importing from schemas directly, so the ~30 dashboard suites that `vi.mock('../utils/Enum', …)` keep working without pulling the full schemas index into those tests.
- `CONNECTION_KINDS.MESHERY` and `.GITHUB` had no direct references — they existed only to feed the enumeration — so they need no replacement constant.

The kind literals are not redefined here: they come from a `CoreKind` enum in the schema, which generates the constants for both Go and TypeScript. See meshery/schemas#1091.

### Verification

- `tsc --noEmit` — 0 errors.
- Vitest — 57 files / 287 tests pass across `utils/__tests__/Enum.test.ts`, `components/connections`, `components/dashboard/resources`, `components/telemetry`, `components/layout/{Header,AppShell}`.
- ESLint clean on all changed files.
- `grep -rn "CONNECTION_KINDS" ui` returns nothing.

### Blocked on

meshery/schemas#1091 must merge and publish a `@meshery/schemas` containing `CoreConnectionKinds` before CI here can go green — `ui/package.json` still pins `^1.3.35`, which does not have it. Local verification used a schemas build synced into `node_modules`. The version bump is a follow-up commit on this branch.

## Summary by CodeRabbit

* **New Features**
  * “Kind” filters in connections now populate dynamically from available connection definitions.

* **Bug Fixes**
  * Improved consistency for Kubernetes/Grafana/Prometheus connection handling across connection wizards, dashboards, and connection actions.
  * Updated resource table “Cluster” chips/tooltips to reliably show the correct connection icons.

* **Tests**
  * Expanded and adjusted unit tests/mocks for the new schema-sourced connection kind constants, plus added coverage for re-exports.

* **Chores**
  * Updated the UI schemas dependency version.


ui/components/dashboard/resources/configuration/configmap-columns.tsx
aabidsofi19, line 122: why are we not using the connection chip directly . we should be hardcoding the icon again here

ui/components/layout/Header/Header.tsx
```diff
@@ -130,7 +130,7 @@ const K8sContextConnectionChip_ = ({
             // through normalizeStaticImagePath, which turns SVG markup into a
             // data URI. Prefixing with "/" here would corrupt the SVG markup.
             iconSrc={
-              connectionMetadataState?.[CONNECTION_KINDS.KUBERNETES]?.icon ||
+              connectionMetadataState?.[CoreConnectionKinds.kubernetes]?.icon ||
```
aabidsofi19, line 133: the connection chip should get the connection definition dynamically

ritzorama:
Working on the feedback...

hamza-mohd (approved):
Reviewed as a maintainer.

Drives connection kinds from the registry-generated constants instead of a hardcoded list, removing
the re-declared literal strings across the UI. Sourced correctly from `@meshery/schemas` (the
generated `CoreKind` constants), per the schema-driven mandate - no locally defined contract.

Landing sequence verified end to end:
- Depends on schemas#1091 ("generate core connection kind constants"), which was merged and published
  as `@meshery/schemas@1.3.39`; this PR bumps to it, so the imported constants now resolve.
- The bump cleared the previously-failing build-meshery and ui-unit-tests.
- A merge conflict with master (introduced by the recently-landed connection PRs #20956/#20962/#20950)
  was resolved via a merge from master, no history rewrite.

CI green: 10 passed, 0 failed, 1 skipped; only advisory CodeRabbit pending. The permanently-red
`agent` AI-reviewer check (repo-wide quota failure) is a known non-signal and is not evidence about
this change. Approving.

Issue #20950, opened 2026-07-24
[Server] Duplicate registrant connections are created per registrant kind


fitzergerald:
### Summary

Meshery creates more than one registrant Connection per registrant kind at boot, and the duplicates are indistinguishable on the Connections page. This is pre-existing behavior, surfaced while implementing #20932.

### Evidence

Model registration builds each model's registrant from the `registrant` blob hand-authored into its `model.json`, then persists it via the content-addressed `Connection.Create` (md5 over the whole v1beta1 Connection - `schemas/models/v1beta1/connection/connection_helper.go`).

Those blobs are not uniform. Counting the shipped models:

| kind | blobs carrying `credential_id` + `user_id` | blobs omitting both |
|---|---|---|
| `artifacthub` | 208 | 927 |
| `github` | 54 | 354 |

`CredentialID` and `UserID` are `*core.Uuid`. An omitted key marshals to `null`; a present zero UUID marshals to `"00000000-0000-0000-0000-000000000000"`. Different JSON gives a different md5, so each kind lands **two** distinct rows in `connections`, both referenced by `registries.registrant_id`. A live Meshery database confirms two `artifacthub` registrant rows with 4517 and 36 references respectively.

Both variants carry the same `name`, `type` and `sub_type`, so the Connections page renders two identical rows per kind with no way to tell them apart.

### Why it is not fixed in #20932

#20932 seeds the system-owned Connections for Artifact Hub and GitHub from their registered connection definitions. It deliberately writes **one deterministic canonical registrant per kind** and leaves the others untouched, because collapsing duplicates means repointing `registries.registrant_id` and deleting rows - mutating relationships during server startup, on a defect that predates that change.

### Suggested direction

Collapse duplicate registrants per kind, outside the boot path: pick a canonical registrant, repoint every `registries.registrant_id` that references a sibling, then delete the siblings. Two things worth settling first:

1. **Where it runs.** A migration is safer than boot-time seeding - it runs once, in a transaction, with a clear failure mode, instead of on every start.
2. **Stop reintroducing it.** Even after a collapse, the next registration re-creates both variants from the same non-uniform blobs. Either normalize the `registrant` blobs in the shipped `model.json` files (and in whatever generates them), or make the registrant's content-addressed identity ignore fields that are semantically empty - `credential_id`, `user_id` - so absent and zero hash alike. The second fixes the root cause for every registrant kind, including ones added later.

[...]

Issue #20932, opened 2026-07-24
[Server] Seed Artifact Hub and GitHub connections during boot-time seed-data initialization


suttonskate:
### Goal

[PR #20926](https://github.com/meshery/meshery/pull/20926) added first-class **connection definitions** for **Artifact Hub** (`kind: artifacthub`, `type: source`, `subType: registry`) and **GitHub** (`kind: github`, `type: source`, `subType: git`) under `models/meshery-core/0.7.2/v1.0.0/connections/`, plus the state-machine and UI wiring so both kinds transition and title correctly.

Both are already first-class registrants in Meshery - models/components are generated from Artifact Hub packages (`MeshModelGenerationHandler`) and from GitHub-hosted manifests (meshkit `generators/github`) - and now carry connection definitions.

This issue tracks **putting those two connection definitions to use in the server's boot-time seed-data initialization**, so the Artifact Hub and GitHub connection types are registered/seeded at startup and available out of the box, rather than only surfacing once a user manually creates a connection.

### Scope

- Investigate the server's boot-time seed-data init path (`server/cmd/main.go` -> `models.SeedComponents` / `DefaultLocalProvider.SeedContent`, `server/models/`) and determine the correct hook point for seeding these connection types at startup.
- Seed/register the **artifacthub** and **github** connection types during boot, **sourced from #20926's registered connection definitions in the registry - not a hardcoded kind list**.
- Ensure idempotency across server restarts (no duplicate rows / re-seeding on every boot).
- Follow the mandatory identifier-naming conventions in `AGENTS.md` (wire camelCase, DB snake_case, Go initialisms; schema-driven).

### Definition of done

- Boot-time seed-init registers/seeds the Artifact Hub and GitHub connection types at startup, registry-driven from #20926's definitions.
- Unit tests cover the boot-time seeding of these connections.
- Internal + user-facing docs updated.

### References

- Depends on / builds atop #20926
- Contributor guide: `docs/content/en/project/contributing/models/connections.md`

santhosh-7777:
@suttonskate  i would like to work on this , can u assign this to me?

fitzergerald:
Picking this up. Verifying the boot path rather than assuming the scoping above, I found the issue as written describes work that already ships - and that the real defect is a different one. Recording both so the issue tracks what is actually being fixed.

## What already works

[...]

Issue #1093, opened 2020-06-02
Performance: Load Test is broken


kushthedude:
This PR https://github.com/layer5io/meshery/pull/1040 broke the Load-Test.

**Error Logs**
```sh
DEBU[0059] Headers : <nil>
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1fbfa0e]

goroutine 355 [running]:
github.com/layer5io/meshery/helpers.sharedHTTPOptions(0xc00074a000, 0x10, 0x11b4a00, 0x0)
	/Users/kush/meshery/helpers/fortio.go:38 +0x12e
github.com/layer5io/meshery/helpers.FortioLoadTest(0xc00074a000, 0xc0004895b8, 0x2660c60, 0xc00008b800, 0x0)
	/Users/kush/meshery/helpers/fortio.go:69 +0x43
github.com/layer5io/meshery/handlers.(*Handler).executeLoadTest(0xc000434070, 0x2660ca0, 0xc0000b6000, 0xc000667600, 0xc00001c768, 0x3, 0xc00001c76c, 0x0, 0xc00001c7a9, 0x24, ...)
	/Users/kush/meshery/handlers/load_test_handler.go:300 +0xd7
github.com/layer5io/meshery/handlers.(*Handler).loadTestHelperHandler.func2(0xc000434070, 0xc000667600, 0xc00001c768, 0x3, 0xc00001c76c, 0x0, 0xc00001c7a9, 0x24, 0xc00024f2c0, 0x26909a0, ...)
	/Users/kush/meshery/handlers/load_test_handler.go:274 +0xe6
created by github.com/layer5io/meshery/handlers.(*Handler).loadTestHelperHandler
	/Users/kush/meshery/handlers/load_test_handler.go:272 +0x66a
```

PR #20956, opened 2026-07-24
feat(server): seed system-owned Artifact Hub and GitHub connections at boot

jijillery:
## Intent

Implement meshery issue #20932: seed the Artifact Hub and GitHub connection types during the server's boot-time seed-data initialization, so they are available out of the box rather than only surfacing once a user manually creates a connection. Builds on merged PR #20926, which added first-class connection definitions for artifacthub (type=source, subType=registry) and github (type=source, subType=git) under models/meshery-core/0.7.2/v1.0.0/connections/.

Investigation finding that reframed the work (recorded on the issue): the issue's literal definition of done already shipped. I verified empirically by running the real registration path against an in-memory DB that (a) both connection DEFINITIONS are already registered into the registry at every boot by SeedComponents, and (b) connection ROWS for both kinds already exist at boot, created incidentally as registrant rows by the 1,135 artifacthub-registrant and 408 github-registrant models. The actual defect is that those boot-created rows take their identity from the stale 'registrant' blob hand-authored into every model.json, which now contradicts the authoritative definitions: rows say type=registry/subType="" and name "Github", definitions say source/registry, source/git and "GitHub". Because ConnectionPersister.GetConnections filters on type/sub_type, filtering the Connections page by 'source' returned neither, while the registry-driven Create Connection wizard offers both as source kinds.

HARD CONSTRAINT set by the user: registry-driven, never a hardcoded kind list. If the implementation names the two kinds literally, it is wrong. No kind string appears anywhere in the new server code.

The user (acting as captain) chose the scoping approach after I escalated it as a design decision, and added the governing principle: these connections are OWNED BY THE SYSTEM and carry NO credentials, because they use their respective APIs anonymously. So a kind is seeded only when BOTH registry-derived rules hold: (1) Meshery already sources content through the kind - it holds a registrant Connection of that kind; (2) the kind works anonymously - its credentialSchema marks nothing as required. Together these yield exactly artifacthub and github. Kubernetes, Grafana and Prometheus describe resources a USER brings and would be meaningless as empty endpoint-less rows. Rule 2 is load-bearing, not decorative: server/models/meshmodel/core/register.go creates a kubernetes registrant Connection whenever a user's cluster components are registered, so rule 1 alone goes true for Kubernetes the moment someone connects a cluster; the required kubeconfig in its credentialSchema is what keeps it out. There is a dedicated regression test for exactly that scenario.

Deliberate decisions a reviewer reading only the diff would not know:
- Rows are UPDATED IN PLACE, never inserted. Connection IDs are content-addressed (md5 over the whole connection, schemas connection_helper.go), so inserting a row carrying the definition's identity would NOT collide with the registrant row carrying the stale identity - it would leave TWO connections per kind, exactly the duplicate-row outcome the issue forbids. Updating keeps one row and keeps every registries.registrant_id foreign key pointing at it. This is not an oversight or a missing create path.
- 'status' is deliberately NOT re-asserted. The definition's status is the state a NEW connection starts in; once the connection exists its state belongs to the connection state machine, so re-writing it every boot would silently undo a user who connected or ignored it. There is a test pinning this.
- A user's own connections are never touched: seeding only ever reads registrant rows (rows referenced by registries.registrant_id), and a user's connection is never a registrant.
- 'url' is deliberately left unset on seeded connections. Artifact Hub has a plausible default (artifacthub.io) but its definition declares none, and GitHub has no single sensible repo; inventing either would be data the registry never authored. Flagged to the user as a judgment call.
- SeedComponents gained a *database.Handler parameter and calls SeedConnections at its end, rather than each caller calling SeedConnections separately. This is intentional so that every path that rebuilds the registry - server boot (cmd/main.go), database reset (handlers/database_handlers.go), GraphQL hard reset (internal/graphql/resolver/meshsync.go) - seeds too, and so seeding always runs AFTER registration has created both the definitions and the registrants it reads. The three call-site edits are mechanical consequences of that one signature change. Note the project is removing GraphQL, so the resolver call site is a near-dead path updated only to compile.
- Two new MeshKit error codes were minted (meshery-server-1462 ErrSeedingConnections, 1463 ErrSeedingConnectionKind) with server/helpers/component_info.json next_error_code bumped 1462 -> 1464, per the repo's error-code convention. ErrSeedingConnectionKind takes the kind as a first-class field rather than wrapping with fmt.Errorf, to satisfy the MeshKit errors framework. The errorutil docs export is deliberately NOT committed.
- Per-kind seeding failures are logged and skipped rather than aborting the loop, so one bad definition cannot block the others; the next boot retries.

Idempotency across restarts was a required outcome and is proven, not assumed: the restart test re-runs registration and then seeding, fails on ANY duplicate row per kind, and separately asserts that steady-state seeding writes ZERO rows.

Docs: added a 'System-owned Connections' section to docs/content/en/project/contributing/models/connections.md (the contributor guide named in the issue), documenting the hook point, both scoping rules, the status and user-connection carve-outs, and the idempotency guarantee - including an explicit warning that authoring a definition does NOT get your kind seeded, since a connection to your own Grafana or cluster is yours to create.

Already verified locally before this run: go build ./..., go test ./models/ ./handlers/ ./machines/... (all ok), go vet, golangci-lint with the project config (clean), gofmt, and errorutil analyze (clean, no findings). Branch was rebased onto current master; the diff is exactly nine files. Issue #20932 was updated with the real defect and this rationale.

## What Changed

- Add boot-time connection seeding (`server/models/seed_connections.go`) that updates registrant-created connection rows in place to carry their authoritative definition identity (name, type, subType). The kind selection is registry-driven with no hardcoded kind list: a kind is seeded only when it holds a `model` registrant connection and its `credentialSchema` requires nothing, which yields exactly Artifact Hub (source/registry) and GitHub (source/git). Rows are updated (never inserted) to avoid content-addressed-id duplicates, and `status`, `credential_id`, and user-owned connections are deliberately left untouched. A canonical registrant per kind is picked stickily (a registrant already carrying the identity wins; lowest id only on first stamp), and both schema halves (`type` and `connection_type`) are written.
- Thread a `*database.Handler` through `SeedComponents` so every registry-rebuild path (server boot in `cmd/main.go`, database reset in `handlers/database_handlers.go`, GraphQL hard reset in `resolver/meshsync.go`) runs `SeedConnections` after registration. Add `system_migration.go` to unify the system-table AutoMigrate list across reset paths so `GET /api/integrations/connections` no longer 500s after a database reset.
- Mint two MeshKit error codes (`meshery-server-1462` ErrSeedingConnections, `1463` ErrSeedingConnectionKind) with `next_error_code` bumped to 1464; add seed, migration, and handler tests proving idempotency across restarts (zero writes in steady state, no duplicate rows) plus the status/credential carve-outs; and document the mechanism, scoping rules, and carve-outs in the contributor connections guide.

## Risk Assessment

✅ Low: Every substantive defect from the previous two rounds is fixed and independently verified against the real code paths and a live database, leaving only documentation accuracy, a duplicated query, an acknowledged narrow edge case, and the recurring scratch-file churn.

## Testing

Fixed a local Go toolchain mismatch (PATH go1.26.5 vs GOROOT-pinned go1.26.4) by using the GOROOT-matched go1.26.4 binary, then ran the targeted seed-connection, migration, and API-level tests in server/models and server/handlers - all pass. Because unit tests alone are insufficient, I built the actual meshery-server binary from the target commit and booted it against a throwaway data dir: first boot logged "Seeded 2 system-owned connection(s)" with no kind string configured, the Connections API listed Artifact Hub and GitHub and filtered them by type=source, the database-reset path returned HTTP 200 with the seeded rows present (the pre-fix HTTP 500 meshery-server-1032 regression, captured in the prior round's now-superseded transcript, is resolved by the unified migration list recreating environment_connection_mappings), and a restart wrote zero rows with identical content-addressed IDs. I flagged the prior live transcript's stale step 6 as superseded. No test failures or unfixable issues; the only worktree changes are pre-existing .remember/tmp session scratch files touched by the session hook, not my testing.

<details>
<summary>Evidence: Live server transcript (post-fix): boot seeding, reset path HTTP 200, restart idempotency</summary>

```text
# meshery/meshery#20932 - live server transcript (post-fix, target f453ea574)

Real `meshery-server` binary built from this branch's target commit and booted against a
throwaway `USER_DATA_FOLDER` and isolated `HOME`, so the run touched no real user data.
Run from `server/cmd` so the CWD-relative `ModelsPath="../../models"` resolves to the repo's
shipped models.

`` `
go build -o /tmp/.../bin/meshery-server ./server/cmd     # go1.26.4 toolchain (GOROOT-matched)

cd server/cmd
HOME=.../home USER_DATA_FOLDER=.../config KUBECONFIG_FOLDER=.../kube \
PORT=9099 PROVIDER=None SKIP_DOWNLOAD_CONTENT=true SKIP_DOWNLOAD_EXTENSIONS=true \
DISABLE_OPERATOR=true REGISTER_STATIC_K8S=false  .../bin/meshery-server
`` `

> This transcript SUPERSEDES `live-server-transcript.md`, whose step 6 was captured on an
> earlier commit and showed the reset path returning HTTP 500. The target commit
> `f453ea574` ("Unify system-table migration list across reset paths") fixes exactly that;
> section 5 below is the corrected result.

## 1. First boot - seeding runs out of the box (no kind string configured)

`` `
For registrant artifacthub imported 152 models, 1341 components, 26 relationships.
For registrant github     imported 131 models, 1461 components, 564 relationships.
For registrant meshery    imported 9 models, 1156 components, 4 relationships.
Seeded 2 system-owned connection(s) from registered connection definitions
`` `

## 2. GET /api/integrations/connections - the Connections page a user loads

`` `
HTTP 200  totalCount=4
  kind=github       name=GitHub        type=source  subType=git       status=registered  owner=null  credentialId=null
  kind=artifacthub  name=Artifact Hub  type=source  subType=registry  status=registered  owner=null  credentialId=null
  kind=meshery      name=meshery       type=-       subType=-         status=registered  owner=null  credentialId=null
  kind=artifacthub  name=Artifact Hub  type=-       subType=-         status=registered  owner=null  credentialId=null
`` `

Only one canonical row per kind is stamped; the `meshery` registrant and the second
`artifacthub` registrant (the `user_id` blob variant, meshery/meshery#20950) are left
exactly as registration wrote them.

## 3. Filter by type=source - the reported defect, now returns both

`` `
HTTP 200  totalCount=2
  kind=github       name=GitHub        type=source  subType=git
  kind=artifacthub  name=Artifact Hub  type=source  subType=registry
`` `

## 4. Persisted DB - both halves of the split schema agree

`` `
sqlite> select kind,name,type,sub_type,connection_type,user_id from connections;
artifacthub|Artifact Hub|source|registry|source|00000000-0000-0000-0000-000000000000   <- seeded: legacy type AND canonical connection_type; owner cleared
artifacthub|Artifact Hub|registry|||                                                   <- sibling registrant, untouched
github|GitHub|source|git|source|                                                       <- seeded
meshery|meshery|registry|||                                                            <- meshery registrant, untouched
`` `

`type` (what ConnectionPersister.GetConnections filters on) and `connection_type`
(what the API renders) agree, which is what makes step 3 work.

## 5. Database reset - the FIX (was HTTP 500 before target commit)

`` `
$ curl -X DELETE /api/system/database/reset
{"message":"Database reset successful"}   HTTP 200

# registration + seeding re-ran automatically (second call site):
For registrant artifacthub imported 152 models ...
Seeded 2 system-owned connection(s) from registered connection definitions

# The call that returned HTTP 500 (meshery-server-1032) pre-fix - NO server restart:
$ curl /api/integrations/connections?pageSize=100
HTTP 200  totalCount=4
  kind=github       name=GitHub        type=source  subType=git
  kind=artifacthub  name=Artifact Hub  type=source  subType=registry
  kind=meshery      name=meshery       type=-       subType=-
  kind=artifacthub  name=Artifact Hub  type=-       subType=-

$ curl '/api/integrations/connections?filter=type source'
HTTP 200  totalCount=2  (github + artifacthub, both source)

# the tables the reset used to drop-and-never-recreate now exist:
sqlite> select name from sqlite_master where type='table'
         and (name like '%environment%' or name='connections');
connections
environment_connection_mappings
environments
workspaces_environments_mappings
`` `

Root fix: all three system-DB migration sites (boot, reset handler, GraphQL hard reset)
now share `models.SystemDatabaseModels`, so no path can drop every table and re-migrate
only a stale subset. `GetConnections` LEFT JOINs `environment_connection_mappings`, which
is now recreated by the reset.

## 6. Restart - idempotent: zero writes, same rows, same content-addressed ids

Server stopped and restarted against the same (post-reset) database.

`` `
boot2: grep -c "Seeded" = 0                 <- steady state wrote nothing

ids before restart == ids after restart:
  github       source/git       f5995382-6fb7-9cff-002a-4a471cbd0c7e
  artifacthub  source/registry  1cad20cd-b4cc-3e1b-0353-b7d2c65fa23e
  artifacthub  (sibling)        50bef83c-dad7-9977-952c-099321286a6a
  meshery      (registrant)     8c9393f8-a4a5-8cbe-1f1d-98c9c0e9c37f
`` `

No duplicate rows; identity is a fixed point across restarts.
```
</details>
<details>
<summary>Evidence: API test transcript: connections API before vs after seeding (raw JSON)</summary>

```text
=== RUN   TestConnectionsAPISurfacesSeededSystemConnections
    connections_seeded_api_test.go:142: BEFORE seeding - GET /api/integrations/connections
        [
          {
            "id": "50bef83c-dad7-9977-952c-099321286a6a",
            "name": "Artifact Hub",
            "kind": "artifacthub",
            "type": "",
            "subType": "",
            "status": "registered",
            "owner": ""
          },
          {
            "id": "7c75db35-5703-e6c1-e458-b8c11f32654a",
            "name": "Github",
            "kind": "github",
            "type": "",
            "subType": "",
            "status": "registered",
            "owner": ""
          },
          {
            "id": "8c9393f8-a4a5-8cbe-1f1d-98c9c0e9c37f",
            "name": "meshery",
            "kind": "meshery",
            "type": "",
            "subType": "",
            "status": "registered",
            "owner": ""
          }
        ]
    connections_seeded_api_test.go:145: BEFORE seeding - GET /api/integrations/connections?filter=type+source
        []
    connections_seeded_api_test.go:162: AFTER seeding - GET /api/integrations/connections
        [
          {
            "id": "50bef83c-dad7-9977-952c-099321286a6a",
            "name": "Artifact Hub",
            "kind": "artifacthub",
            "type": "source",
            "subType": "registry",
            "status": "registered",
            "owner": ""
          },
          {
            "id": "7c75db35-5703-e6c1-e458-b8c11f32654a",
            "name": "GitHub",
            "kind": "github",
            "type": "source",
            "subType": "git",
            "status": "registered",
            "owner": ""
          },
          {
            "id": "8c9393f8-a4a5-8cbe-1f1d-98c9c0e9c37f",
            "name": "meshery",
            "kind": "meshery",
            "type": "",
            "subType": "",
            "status": "registered",
            "owner": ""
          }
        ]
    connections_seeded_api_test.go:165: AFTER seeding - GET /api/integrations/connections?filter=type+source
        [
          {
            "id": "50bef83c-dad7-9977-952c-099321286a6a",
            "name": "Artifact Hub",
            "kind": "artifacthub",
            "type": "source",
            "subType": "registry",
            "status": "registered",
            "owner": ""
          },
          {
            "id": "7c75db35-5703-e6c1-e458-b8c11f32654a",
            "name": "GitHub",
            "kind": "github",
            "type": "source",
            "subType": "git",
            "status": "registered",
            "owner": ""
          }
        ]
    connections_seeded_api_test.go:166: AFTER seeding - raw response body (unfiltered): {"connections":[{"id":"7c75db35-5703-e6c1-e458-b8c11f32654a","name":"GitHub","description":null,"url":null,"credentialId":null,"type":"source","subType":"git","kind":"github","modelReference":null,"status":"registered","owner":null,"createdAt":"2026-07-24T15:05:39.857177-05:00","updatedAt":"2026-07-24T15:05:39.866897-05:00","deletedAt":null,"schemaVersion":"","styles":null},{"id":"50bef83c-dad7-9977-952c-099321286a6a","name":"Artifact Hub","description":null,"url":null,"credentialId":null,"type":"source","subType":"registry","kind":"artifacthub","modelReference":null,"status":"registered","owner":null,"createdAt":"2026-07-24T15:05:39.764745-05:00","updatedAt":"2026-07-24T15:05:39.866656-05:00","deletedAt":null,"schemaVersion":"","styles":null},{"id":"8c9393f8-a4a5-8cbe-1f1d-98c9c0e9c37f","name":"meshery","description":null,"url":null,"credentialId":null,"type":"","subType":"","kind":"meshery","modelReference":null,"status":"registered","owner":null,"createdAt":"2026-07-24T15:05:39.630952-05:00","updatedAt":"2026-07-24T15:05:39.630952-05:00","deletedAt":null,"schemaVersion":"","styles":null}],"totalCount":3,"page":0,"pageSize":100,"statusSummary":{"registered":3}}
    connections_seeded_api_test.go:167: AFTER seeding - raw response body (type=source): {"connections":[{"id":"7c75db35-5703-e6c1-e458-b8c11f32654a","name":"GitHub","description":null,"url":null,"credentialId":null,"type":"source","subType":"git","kind":"github","modelReference":null,"status":"registered","owner":null,"createdAt":"2026-07-24T15:05:39.857177-05:00","updatedAt":"2026-07-24T15:05:39.866897-05:00","deletedAt":null,"schemaVersion":"","styles":null},{"id":"50bef83c-dad7-9977-952c-099321286a6a","name":"Artifact Hub","description":null,"url":null,"credentialId":null,"type":"source","subType":"registry","kind":"artifacthub","modelReference":null,"status":"registered","owner":null,"createdAt":"2026-07-24T15:05:39.764745-05:00","updatedAt":"2026-07-24T15:05:39.866656-05:00","deletedAt":null,"schemaVersion":"","styles":null}],"totalCount":2,"page":0,"pageSize":100,"statusSummary":{"registered":3}}
--- PASS: TestConnectionsAPISurfacesSeededSystemConnections (0.30s)
=== RUN   TestConnectionsAPISeededConnectionsSurviveRestart
    connections_seeded_api_test.go:230: AFTER restart - GET /api/integrations/connections?filter=type+source
        [
          {
            "id": "50bef83c-dad7-9977-952c-099321286a6a",
            "name": "Artifact Hub",
            "kind": "artifacthub",
            "type": "source",
            "subType": "registry",
            "status": "registered",
            "owner": ""
          },
          {
            "id": "7c75db35-5703-e6c1-e458-b8c11f32654a",
            "name": "GitHub",
            "kind": "github",
            "type": "source",
            "subType": "git",
            "status": "registered",
            "owner": ""
          }
        ]
--- PASS: TestConnectionsAPISeededConnectionsSurviveRestart (0.31s)
PASS
ok  	github.com/meshery/meshery/server/handlers	(cached)
```
</details>
<details>
<summary>Evidence: Raw server log: first boot + database reset (shows Seeded 2 on both)</summary>

```text
time="2026-07-24T15:11:48-05:00" level=info msg="OpenTelemetry config not set; tracing disabled" app=meshery
time="2026-07-24T15:11:48-05:00" level=info msg="Local Provider capabilities are: Not Set" app=meshery
time="2026-07-24T15:11:48-05:00" level=info msg="Meshery Server release channel is: Not Set" app=meshery
time="2026-07-24T15:11:48-05:00" level=info msg="MESHSYNC_DEFAULT_DEPLOYMENT_MODE is embedded" app=meshery
time="2026-07-24T15:11:48-05:00" level=info msg="Meshery Database is at: /tmp/meshery-seed-evidence2/config" app=meshery
time="2026-07-24T15:11:48-05:00" level=info msg="Using kubeconfig at: /tmp/meshery-seed-evidence2/kube" app=meshery
time="2026-07-24T15:11:48-05:00" level=info msg="Log level: info" app=meshery
time="2026-07-24T15:11:59-05:00" level=warning msg="Status Code: 500 .Received unexpected status code: 404.  | Short Description: Request was not successful | Probable Cause: The server might be down or temporarily unavailable..There could be network connectivity issues between the client and the server..The API endpoint might be misconfigured, or the server is experiencing high load. | Suggested Remediation: Check if the server is online and operational..Verify network connectivity and ensure there are no firewalls or DNS issues blocking the request..Confirm the API endpoint is correct and functional. | Short Description: Unable to fetch data from the Provider.Capabilities\nremote provider responded with HTTP 500 Internal Server Error" app=meshery code=meshery-server-1270 probable-cause= severity=2 short-description="Unable to fetch data from the Provider.Capabilities" suggested-remediation=
time="2026-07-24T15:12:00-05:00" level=info msg="Meshery Server listening on: 9099" app=meshery
time="2026-07-24T15:12:01-05:00" level=warning msg=".Get \"https://kickstart.metabit.com/Not%20Set/capabilities?os=meshery&playground=false\": dial tcp: lookup kickstart.metabit.com: no such host | Short Description: Could not reach remote provider | Probable Cause: Remote provider server may be down or not accepting requests. | Suggested Remediation: Make sure remote provider server is healthy and accepting requests." app=meshery code=meshery-server-1301 probable-cause="Remote provider server may be down or not accepting requests." severity=2 short-description="Could not reach remote provider" suggested-remediation="Make sure remote provider server is healthy and accepting requests."
time="2026-07-24T15:12:01-05:00" level=warning msg=".Get \"https://collab.eti.cisco.com/Not%20Set/capabilities?os=meshery&playground=false\": dial tcp: lookup collab.eti.cisco.com: no such host | Short Description: Could not reach remote provider | Probable Cause: Remote provider server may be down or not accepting requests. | Suggested Remediation: Make sure remote provider server is healthy and accepting requests." app=meshery code=meshery-server-1301 probable-cause="Remote provider server may be down or not accepting requests." severity=2 short-description="Could not reach remote provider" suggested-remediation="Make sure remote provider server is healthy and accepting requests."
time="2026-07-24T15:12:03-05:00" level=warning msg=".Get \"https://kickstart.metabit.com/capabilities?os=meshery&playground=false\": dial tcp: lookup kickstart.metabit.com: no such host | Short Description: Could not reach remote provider | Probable Cause: Remote provider server may be down or not accepting requests. | Suggested Remediation: Make sure remote provider server is healthy and accepting requests." app=meshery code=meshery-server-1301 probable-cause="Remote provider server may be down or not accepting requests." severity=2 short-description="Could not reach remote provider" suggested-remediation="Make sure remote provider server is healthy and accepting requests."
time="2026-07-24T15:12:03-05:00" level=warning msg=".Get \"https://collab.eti.cisco.com/capabilities?os=meshery&playground=false\": dial tcp: lookup collab.eti.cisco.com: no such host | Short Description: Could not reach remote provider | Probable Cause: Remote provider server may be down or not accepting requests. | Suggested Remediation: Make sure remote provider server is healthy and accepting requests." app=meshery code=meshery-server-1301 probable-cause="Remote provider server may be down or not accepting requests." severity=2 short-description="Could not reach remote provider" suggested-remediation="Make sure remote provider server is healthy and accepting requests."
time="2026-07-24T15:12:16-05:00" level=info msg="For registrant artifacthub imported 152 models, 1341 components, 26 relationships." app=meshery
time="2026-07-24T15:12:16-05:00" level=info msg="For registrant github imported 131 models, 1461 components, 564 relationships." app=meshery
time="2026-07-24T15:12:16-05:00" level=info msg="For registrant meshery imported 9 models, 1156 components, 4 relationships." app=meshery
time="2026-07-24T15:12:16-05:00" level=info msg="Seeded 2 system-owned connection(s) from registered connection definitions" app=meshery
time="2026-07-24T15:12:16-05:00" level=error msg="open : no such file or directory | Short Description: error reading file | Probable Cause: File does not exist in the location .Insufficient permissions | Suggested Remediation: Verify that file exist at the provided location.Verify sufficient file permissions." app=meshery code=meshkit-11171 probable-cause="File does not exist in the location .Insufficient permissions" severity=2 short-description="error reading file" suggested-remediation="Verify that file exist at the provided location.Verify sufficient file permissions."
time="2026-07-24T15:12:40-05:00" level=warning msg="DEPRECATION: the local provider name \"None\" is the legacy alias for \"Local\" and will be removed in a future Meshery release. Update meshery-provider cookies, the PROVIDER environment variable, and ~/.meshery/config.yaml to use \"Local\". This warning is logged once per server process." app=meshery
time="2026-07-24T15:13:21-05:00" level=info msg="For registrant artifacthub imported 152 models, 1341 components, 26 relationships." app=meshery
time="2026-07-24T15:13:21-05:00" level=info msg="For registrant github imported 131 models, 1461 components, 564 relationships." app=meshery
time="2026-07-24T15:13:21-05:00" level=info msg="For registrant meshery imported 9 models, 1156 components, 4 relationships." app=meshery
time="2026-07-24T15:13:21-05:00" level=info msg="Seeded 2 system-owned connection(s) from registered connection definitions" app=meshery
time="2026-07-24T15:13:21-05:00" level=error msg="open : no such file or directory | Short Description: error reading file | Probable Cause: File does not exist in the location .Insufficient permissions | Suggested Remediation: Verify that file exist at the provided location.Verify sufficient file permissions." app=meshery code=meshkit-11171 probable-cause="File does not exist in the location .Insufficient permissions" severity=2 short-description="error reading file" suggested-remediation="Verify that file exist at the provided location.Verify sufficient file permissions."
```
</details>
<details>
<summary>Evidence: Raw server log: restart against same DB (steady state, zero new seeds)</summary>

```text
time="2026-07-24T15:14:09-05:00" level=info msg="OpenTelemetry config not set; tracing disabled" app=meshery
time="2026-07-24T15:14:09-05:00" level=info msg="Local Provider capabilities are: Not Set" app=meshery
time="2026-07-24T15:14:09-05:00" level=info msg="Meshery Server release channel is: Not Set" app=meshery
time="2026-07-24T15:14:09-05:00" level=info msg="MESHSYNC_DEFAULT_DEPLOYMENT_MODE is embedded" app=meshery
time="2026-07-24T15:14:09-05:00" level=info msg="Meshery Database is at: /tmp/meshery-seed-evidence2/config" app=meshery
time="2026-07-24T15:14:09-05:00" level=info msg="Using kubeconfig at: /tmp/meshery-seed-evidence2/kube" app=meshery
time="2026-07-24T15:14:09-05:00" level=info msg="Log level: info" app=meshery
time="2026-07-24T15:14:23-05:00" level=warning msg="Status Code: 500 .Received unexpected status code: 404.  | Short Description: Request was not successful | Probable Cause: The server might be down or temporarily unavailable..There could be network connectivity issues between the client and the server..The API endpoint might be misconfigured, or the server is experiencing high load. | Suggested Remediation: Check if the server is online and operational..Verify network connectivity and ensure there are no firewalls or DNS issues blocking the request..Confirm the API endpoint is correct and functional. | Short Description: Unable to fetch data from the Provider.Capabilities\nremote provider responded with HTTP 500 Internal Server Error" app=meshery code=meshery-server-1270 probable-cause= severity=2 short-description="Unable to fetch data from the Provider.Capabilities" suggested-remediation=
time="2026-07-24T15:14:25-05:00" level=info msg="Meshery Server listening on: 9099" app=meshery
time="2026-07-24T15:14:25-05:00" level=warning msg=".Get \"https://collab.eti.cisco.com/Not%20Set/capabilities?os=meshery&playground=false\": dial tcp: lookup collab.eti.cisco.com: no such host | Short Description: Could not reach remote provider | Probable Cause: Remote provider server may be down or not accepting requests. | Suggested Remediation: Make sure remote provider server is healthy and accepting requests." app=meshery code=meshery-server-1301 probable-cause="Remote provider server may be down or not accepting requests." severity=2 short-description="Could not reach remote provider" suggested-remediation="Make sure remote provider server is healthy and accepting requests."
time="2026-07-24T15:14:25-05:00" level=warning msg=".Get \"https://kickstart.metabit.com/Not%20Set/capabilities?os=meshery&playground=false\": dial tcp: lookup kickstart.metabit.com: no such host | Short Description: Could not reach remote provider | Probable Cause: Remote provider server may be down or not accepting requests. | Suggested Remediation: Make sure remote provider server is healthy and accepting requests." app=meshery code=meshery-server-1301 probable-cause="Remote provider server may be down or not accepting requests." severity=2 short-description="Could not reach remote provider" suggested-remediation="Make sure remote provider server is healthy and accepting requests."
time="2026-07-24T15:14:28-05:00" level=warning msg=".Get \"https://kickstart.metabit.com/capabilities?os=meshery&playground=false\": dial tcp: lookup kickstart.metabit.com: no such host | Short Description: Could not reach remote provider | Probable Cause: Remote provider server may be down or not accepting requests. | Suggested Remediation: Make sure remote provider server is healthy and accepting requests." app=meshery code=meshery-server-1301 probable-cause="Remote provider server may be down or not accepting requests." severity=2 short-description="Could not reach remote provider" suggested-remediation="Make sure remote provider server is healthy and accepting requests."
time="2026-07-24T15:14:28-05:00" level=warning msg=".Get \"https://collab.eti.cisco.com/capabilities?os=meshery&playground=false\": dial tcp: lookup collab.eti.cisco.com: no such host | Short Description: Could not reach remote provider | Probable Cause: Remote provider server may be down or not accepting requests. | Suggested Remediation: Make sure remote provider server is healthy and accepting requests." app=meshery code=meshery-server-1301 probable-cause="Remote provider server may be down or not accepting requests." severity=2 short-description="Could not reach remote provider" suggested-remediation="Make sure remote provider server is healthy and accepting requests."
time="2026-07-24T15:14:48-05:00" level=info msg="For registrant artifacthub imported 304 models, 2682 components, 52 relationships." app=meshery
time="2026-07-24T15:14:48-05:00" level=info msg="For registrant github imported 262 models, 2922 components, 1128 relationships." app=meshery
time="2026-07-24T15:14:48-05:00" level=info msg="For registrant meshery imported 18 models, 2312 components, 8 relationships." app=meshery
time="2026-07-24T15:14:48-05:00" level=error msg="open : no such file or directory | Short Description: error reading file | Probable Cause: File does not exist in the location .Insufficient permissions | Suggested Remediation: Verify that file exist at the provided location.Verify sufficient file permissions." app=meshery code=meshkit-11171 probable-cause="File does not exist in the location .Insufficient permissions" severity=2 short-description="error reading file" suggested-remediation="Verify that file exist at the provided location.Verify sufficient file permissions."
time="2026-07-24T15:15:07-05:00" level=warning msg="DEPRECATION: the local provider name \"None\" is the legacy alias for \"Local\" and will be removed in a future Meshery release. Update meshery-provider cookies, the PROVIDER environment variable, and ~/.meshery/config.yaml to use \"Local\". This warning is logged once per server process." app=meshery
```
</details>
<details>
<summary>Evidence: Reset-path result: GET connections after DELETE /database/reset now HTTP 200 (was 500), env join tables recreated</summary>

```text
DELETE /api/system/database/reset -> HTTP 200; then WITHOUT restart GET /api/integrations/connections -> HTTP 200 totalCount=4 (github source/git, artifacthub source/registry, meshery, artifacthub sibling); filter=type source -> HTTP 200 totalCount=2; sqlite tables now include environment_connection_mappings, environments, workspaces_environments_mappings
```
</details>
- Outcome: 🔧 2 issues found → auto-fixed ✅ across 2 runs (1h9m11s)

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>✅ **intent** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Rebase** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Review** - 5 infos</summary>

- 🚨 `server/models/seed_connections.go:166` - Seeding stamps the definition identity onto EVERY registrant row of a kind, and there is more than one per kind in practice - producing duplicate, indistinguishable connections. The user intent states this as a required outcome: &#34;Idempotency across restarts was a required outcome and is proven&#34; and &#34;Updating keeps one row&#34;; docs/content/en/project/contributing/models/connections.md:175 states it as a guarantee: &#34;can never produce a second Connection for a kind&#34;. Both are false against real data. Registrant IDs are content-addressed md5 over the whole v1beta1 Connection (schemas connection_helper.go), and the shipped registrant blobs are not uniform: 208 models/*/model.json with kind=artifacthub and 54 with kind=github carry &#34;user_id&#34;: &#34;00000000-0000-0000-0000-000000000000&#34; while the rest omit it, which unmarshals to UserID=&amp;uuid.Nil vs nil and hashes to a different row. A live meshery DB confirms it: two artifacthub connection rows (50bef83c… and 1cad20cd…), both referenced by registries.registrant_id (4517 and 36 refs), created 56ms apart in the same boot. registrantConnectionsByKind returns both, isSeedable passes, and the loop at line 166 updates both to name=&#34;Artifact Hub&#34;, type=&#34;source&#34;, sub_type=&#34;registry&#34; - identical in every user-visible field. The test fixture (seed_connections_test.go:39) registers exactly one model dir per kind, so the multi-registrant case can never arise there, and connectionsByKind&#39;s duplicate guard is never exercised against the real registrant population. Needs a decision: pick one canonical registrant per kind to carry the identity, or dedupe/merge registrant rows (which would also require repointing registries.registrant_id).
- ⚠️ `server/models/seed_connections.go:191` - updates[&#34;type&#34;] writes a column the server&#39;s canonical connection model does not read, so the seeded type never reaches the API or the UI. The connections table carries BOTH columns (verified against a live meshery DB: `type` and `connection_type`): meshkit&#39;s NewRegistryManager AutoMigrates v1beta1.Connection (field Type -&gt; column `type`) and server/cmd/main.go:260 AutoMigrates connections.Connection, which is v1beta3 (field ConnectionType, no gorm column tag -&gt; column `connection_type`). ConnectionPersister.GetConnections/GetConnection scan into v1beta3, so the JSON `type` field is read from `connection_type`, which stays NULL for these rows (all four rows in the live DB have `type` populated and `connection_type` empty). Net effect: filtering by &#39;source&#39; now works, because GetConnections filters with raw SQL on `type` (connection_persister.go:53 and dynamicKeys at :57), but ui/components/connections/ConnectionTable.columns.tsx:322 renders the &#39;Category&#39; column from the JSON `type`, which stays blank for the seeded Artifact Hub and GitHub connections. The root cause is schema drift in meshery/schemas (v1beta3 Connection.ConnectionType needs gorm:&#34;column:type&#34;); until that lands, seeding should write the column the canonical model actually reads.
- ⚠️ `server/models/seed_connections.go:202` - The &#34;must never carry an owner&#34; invariant is written to the wrong column and is also unreachable. v1beta1.Connection.UserID maps to GORM column `user_id`, but the canonical v1beta3 model owns `owner` (both columns exist in the live table). Two consequences: (a) the read at line 91 uses the v1beta1 struct, so the `owner` column never scans into conn.UserID and the guard at line 201 is always false - the branch is dead; (b) if it ever did fire it would clear `user_id`, a column no read path in the server consumes, leaving `owner` untouched. Either read/write `owner` (matching connectionv1beta3.Connection.Owner, which is what RegistrantHostToV1beta1 populates UserID from at registry.go:226) or drop the branch rather than leaving an invariant that silently does nothing.
- ⚠️ `server/models/seed_connections.go:124` - Rule 1 (&#34;Meshery already sources content through the kind&#34;) treats the existence of a registrant row as a Meshery-internal signal, but a registrant row of an arbitrary kind can be created through a public API. server/handlers/connection_definition_handler.go:144 builds `connectionv1beta1.Connection{Kind: def.ModelReference.Registrant.Kind}` straight from the POST /api/registry/connections request body and registers it, creating a registrant row for whatever kind the caller names. Of the shipped definitions, only Kubernetes declares a required credential; Grafana and Prometheus have `credentialSchema.required = None`, so rule 2 does not exclude them. An authenticated caller registering a connection definition under registrant kind grafana or prometheus therefore causes the next boot to seed a system-owned, endpoint-less Grafana/Prometheus connection - exactly the outcome the comment at lines 111-115 says rule 1 exists to prevent. Consider gating on registrant rows Meshery itself created (e.g. a system-owned marker) rather than on registrant existence alone.
- ℹ️ `server/models/seed_connections.go:27` - A nil db or registry manager returns silently with no log line, so a boot path that fails to seed leaves no trace at all - every other failure mode in this function logs a MeshKit error. A single log.Warn naming which dependency was nil keeps the resilience property while making the skip diagnosable.

🔧 Fix: Seed one canonical registrant per kind; write both schema halves
5 issues (3 warnings, 2 infos) still open:
- ⚠️ `server/models/seed_connections.go:290` - Seeding clears `credential_id` on every boot, silently downgrading a connection the user authenticated. Both seedable definitions ship an OPTIONAL credential that the wizard offers: models/meshery-core/0.7.2/v1.0.0/connections/ArtifactHubConnection.json declares apiKeyId/apiKeySecret (&#34;Leave both key fields empty to use the public, unauthenticated API&#34;) and GitHubConnection.json declares a token (&#34;for public repositories it is optional and raises the API rate limit&#34;). requiresCredential only gates on `required`, so these kinds are seedable precisely because their credential is optional - meaning a user can legitimately attach one. When they do, the next restart wipes credential_id to the nil UUID while `status` is deliberately left alone, leaving a connection that reads as `connected` but has quietly reverted to anonymous (and a GitHub connection that silently loses its rate-limit relief). This is consistent with the intent&#39;s principle that these connections &#34;carry NO credentials&#34;, but it contradicts the reasoning behind the status carve-out in the same function (&#34;re-asserting it on every boot would undo a user who connected&#34;) - the credential is part of that same user action. Either leave an existing credential alone the way status is left alone, or make the clearing explicit to the user; the credential row itself is also left orphaned in `credentials`.
- ⚠️ `server/models/seed_connections.go:79` - Accumulated duplicate connection definitions cause the same row to be written once per duplicate on the first boot, and inflate the reported count. connection_definition_dbs grows by one row per definition per boot: ConnectionDefinition.Create (schemas v1beta3 connection_definition_helper.go) sets a fresh uuid.NewV4() and does an unconditional Create, and registration re-registers every model dir on every boot. A live meshery DB confirms it - 3 rows each for grafana, kubernetes and prometheus (9 total for 3 shipped definitions). registeredConnectionDefinitions therefore returns N copies per kind, and because byKind (line 66) and legacyByID (line 72) are snapshotted BEFORE the loop, seedUpdatesFor recomputes the same non-empty diff for every copy: the canonical row is UPDATEd N times in one pass, `seeded` counts N, and the log line reports N connections seeded when one row changed. Worse, if two accumulated definitions of a kind ever disagree - a definition edited in a later models release, with the old row never deleted - the row is rewritten twice on every boot forever, breaking the &#34;a restart writes nothing&#34; property. Fix inside this change: handle each kind at most once per pass (skip a kind after it has been stamped, or dedupe defs by kind before the loop). The unbounded definition growth itself is pre-existing (PR #20926 / schemas) and worth a separate issue.
- ⚠️ `server/models/seed_connections.go:254` - &#34;Lowest id&#34; is deterministic for a fixed registrant set, but not stable when the set grows, and nothing ever un-stamps a previously canonical row. Registrant ids are md5 hashes of the hand-authored model.json `registrant` blob; the shipped blobs already come in 8 distinct spellings across kinds. A future models release that introduces a new spelling whose hash sorts below the current canonical id moves the stamp to the new row, while the previously stamped row keeps name/type/subType from the definition forever - leaving two `source`-typed &#34;Artifact Hub&#34; rows, the exact outcome the single-canonical rule exists to prevent, permanently and with no self-healing on later boots. A stable-under-growth rule that still satisfies the owner&#39;s determinism and never-touch-siblings constraints: prefer a registrant that already matches the definition, falling back to lowest id. Related observation on the pick itself - in a live DB the lowest-id artifacthub registrant (1cad20cd…) owns 6 model entities while the sibling it leaves stale (50bef83c…) owns 450, so the definition identity lands on the minor registrant; ordering by model count desc then id asc (explicitly allowed by the owner&#39;s instruction) would put it on the dominant one.
- ℹ️ `.remember/tmp/save-session.pid:1` - Commit 10ecde45 includes .remember/tmp/last-save-ts and .remember/tmp/save-session.pid, which carry this machine&#39;s local session timestamp (1784849965 -&gt; 1784907383) and PID (26253 -&gt; 20008). They are unrelated to the change and are churning again in the working tree right now. This is the same class of accident AGENTS.md calls out for ui/tsconfig.tsbuildinfo - stage explicit paths rather than sweeping tracked scratch files into the commit.
- ℹ️ `server/models/seed_connections.go:32` - Worth stating in the PR description: narrowing rule 1 to `registries.type = &#39;model&#39;` changed which scenario the credential rule guards. server/models/meshmodel/core/register.go registers ComponentDefinitions under the kubernetes registrant, so that registrant owns only `component` entries and is now excluded by rule 1 - the cluster-connect path the intent cites as proof that &#34;rule 2 is load-bearing&#34; no longer reaches rule 2 at all. The regression test was correspondingly rewritten (seed_connections_test.go:378 now synthesizes a kubernetes registrant owning a `model` entity), which covers a different real path: importing a model package whose registrant kind is kubernetes. The end behavior is unchanged and now has defense in depth, but the intent&#39;s description of rule 2&#39;s role is no longer accurate as written.

🔧 Fix: Preserve user credentials; sticky canonical pick; dedupe definitions
5 infos still open:
- ℹ️ `.remember/tmp/save-session.pid:1` - The scratch-file cleanup did not hold. Commit 35462fd5640 (&#34;[Server] Drop session scratch files from the branch diff&#34;) restored .remember/tmp/last-save-ts and .remember/tmp/save-session.pid to the base content as instructed, but the next commit ad67c78b5b7 swept them straight back in - the net branch diff against 3127035 still shows last-save-ts 1784849965 -&gt; 1784910022 and save-session.pid 26253 -&gt; 63925 - and `git status` shows both dirty again right now (modified / deleted). The `.remember` save hook rewrites these files continuously in the background, so restoring them mid-branch cannot survive another commit: either stage explicit source paths only (never `git add -A`, which is what re-swept them), or make the restore the final action immediately before push. As before, do not `git rm` them or add gitignore entries - they are tracked on the base branch.
- ℹ️ `docs/content/en/project/contributing/models/connections.md:179` - The contributor guide still documents the canonical-pick rule that this round replaced. The alert says seeding stamps &#34;exactly one canonical registrant per kind - deterministically, the lowest id, so the same row is picked on every boot&#34;, but canonicalRegistrantFor (server/models/seed_connections.go:312) now prefers a registrant that already carries the definition identity and only falls back to lowest id when none does. That change exists precisely because &#34;the lowest id&#34; does NOT give &#34;the same row on every boot&#34; once the registrant set grows, so the docs now assert the property the code deliberately stopped relying on. Update the sentence to describe the sticky rule (already-stamped row wins; lowest id only for the first stamp) and why - a later models release can introduce a `registrant` spelling that hashes below the stamped row.
- ℹ️ `server/models/seed_connections.go:192` - registrantConnectionsByKind (line 160) and legacyRegistrantColumnsByID (line 192) issue byte-for-byte the same query - `SELECT * FROM connections WHERE EXISTS (SELECT 1 FROM registries WHERE registries.registrant_id = connections.id AND registries.type = &#39;model&#39;)` - and differ only in the scan target, then are rejoined by id in Go. The correlated EXISTS has no supporting index (meshkit&#39;s Registry declares no index on registrant_id) and `registries` grows every boot (14.5k rows after three boots in a live DB), so this scans it once per connection row, twice. Scanning both column sets in one pass - a small struct carrying id, name, kind, connection_type, type, sub_type, owner, user_id - removes the second scan, the id map and the two-snapshot skew window between the reads.
- ℹ️ `server/models/seed_connections.go:312` - Stickiness closes the registrant-growth window but not the definition-change one. carriesDefinitionIdentity matches on name+connectionType+subType, so if a later models release edits a shipped definition (say GitHubConnection.json&#39;s name or subType), no registrant carries the new identity and the pick falls back to registrants[0] - which, in exactly the multi-registrant case the stickiness exists for, is not the previously stamped row. The old row keeps the previous definition identity forever (siblings are never un-stamped, by design) and the new lowest-id row gets the new one, leaving two `source`-typed rows for the kind. Narrow - it needs a multi-registrant kind plus a definition edit - and consistent with the deliberate no-destructive-reverts rule, so noting it rather than proposing a change; it is another instance of what #20950 ultimately resolves.
- ℹ️ `server/models/seed_connections.go:34` - Worth stating in the PR/issue: the seeded rows only surface on the Connections page in local-provider mode. RemoteProvider.GetConnections (server/models/remote_provider.go:4510) proxies to the Remote Provider&#39;s PersistConnection endpoint, so under the default Meshery Cloud provider the page is served by the provider and never reads the local `connections` table that seeding writes. The defect the change targets is stated in terms of ConnectionPersister.GetConnections, which is the local persister, so this is consistent with the intent - but it bounds what &#34;available out of the box&#34; means for #20932: local-provider installs and the local registry views get the corrected identity; remote-provider installs do not.

</details>

<details>
<summary>🔧 **Test** - 2 issues found → auto-fixed ✅</summary>

- ⚠️ `server/handlers/database_handlers.go:165` - Pre-existing, but it blocks the user-visible half of the database-reset seeding path this change deliberately wires up. After `DELETE /api/system/database/reset`, seeding does run and writes the correct rows (verified in sqlite: `github|GitHub|source|git|source`), but `GET /api/integrations/connections` returns HTTP 500 (`meshery-server-1032`) until the server is restarted. Cause: `ResetSystemDatabase` drops every table and re-migrates only a subset - `environments` and `environment_connection_mappings` are never recreated - while `ConnectionPersister.GetConnections` joins `environment_connection_mappings` to batch-load environments. Confirmed pre-existing: the base commit `3127035f9ec` has the same AutoMigrate list minus `connections.Connection{}`, and the join predates it. Flagging it because this diff already edits exactly that AutoMigrate list, for exactly this class of reason (the comment added there explains the canonical-columns gap), so extending it with `environment.Environment{}` and `environment.EnvironmentConnectionMapping{}` would be a one-line fix in the same place - but widening scope is the author&#39;s call.
- ℹ️ `server/handlers/connections_seeded_api_test.go` - new test file written by agent: server/handlers/connections_seeded_api_test.go
- <code>`go test ./models/ -run &#39;TestSeed|TestIsSeedable&#39; -v` - all 12 seed tests pass (registrant scoping, credential rule, canonical pick, idempotency, status and credential carve-outs)</code>
- <code>`go test ./handlers/ -run &#39;TestConnectionsAPI&#39; -count=1 -v` - new test driving the real `GetConnections` handler; records the response body before and after seeding and across a simulated restart</code>
- <code>`go build -o /tmp/.../meshery-server ./server/cmd` then booted it with `HOME`/`USER_DATA_FOLDER`/`KUBECONFIG_FOLDER` pointed at a throwaway dir, `PROVIDER=None PORT=9081 REGISTER_STATIC_K8S=false`</code>
- <code>Boot log check: `grep &#39;Seeded&#39; server-boot1.log` -&gt; `Seeded 2 system-owned connection(s) from registered connection definitions`</code>
- <code>`curl -s -b &#39;meshery-provider=None&#39; &#39;http://localhost:9081/api/integrations/connections?pageSize=100&#39;` -&gt; HTTP 200, 4 rows, GitHub and Artifact Hub carrying the definition identity</code>
- <code>`curl -s -b &#39;meshery-provider=None&#39; --get --data-urlencode &#39;filter=type source&#39; .../api/integrations/connections` -&gt; HTTP 200, exactly the 2 seeded connections</code>
- <code>`sqlite3 config/mesherydb.sql &#34;select kind,name,type,sub_type,connection_type from connections;&#34;` - both schema halves written on the seeded rows, siblings untouched</code>
- <code>Restart against the same database: `grep -c &#39;Seeded&#39; server-boot2.log` -&gt; 0 (zero writes) and the API returns the same 4 rows with identical ids</code>
- <code>`curl -X DELETE .../api/system/database/reset` then re-checked registration, seeding and the connections API</code>
- <code>Status carve-out: set the GitHub connection to `connected`, restarted, confirmed `status=connected` survived and seeding wrote nothing</code>
- <code>`git show 3127035f9ec:server/handlers/database_handlers.go` and `:server/models/connection_persister.go` to confirm the reset-path 500 predates this branch</code>

🔧 Fix: Unify system-table migration list across reset paths
✅ Re-checked - no issues remain.
- `go test ./server/models/ -run 'TestSeedConnections|TestIsSeedable|TestSeedUpdatesFor|TestSystemDatabaseModels|TestAutoMigrateSystemTables' -v (GOROOT-matched go1.26.4)`
- `go test ./server/handlers/ -run 'TestConnectionsAPISurfacesSeededSystemConnections|TestConnectionsAPISeededConnectionsSurviveRestart' -v (captured before/after connections JSON)`
- `go build ./cmd/... ./internal/graphql/resolver/... ./handlers/... ./models/... (all three SeedComponents call sites compile with the new *database.Handler signature)`
- `Built and booted the real meshery-server binary; exercised GET /api/integrations/connections unfiltered and filter=type source on first boot, after DELETE /api/system/database/reset, and after a restart; inspected persisted sqlite state`

</details>

<details>
<summary>✅ **Document** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Lint** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Push** - passed</summary>

✅ No issues found.

</details>

## Summary by CodeRabbit

* **New Features**
  * System-owned connections (for example Artifact Hub and GitHub) are seeded at startup and exposed via the integrations API.
  * Connection seeding now runs as part of the broader startup/component seeding flow with deterministic registrant selection.

* **Bug Fixes**
  * Seeded system connections remain idempotent across restarts and hard resets (no duplicate results).
  * Improved reliability of system database reset/migrations using a unified system-table migration set.

* **Documentation**
  * Expanded guidance on connection definition registration/consumption, seeding eligibility, and ownership behavior.

* **Tests**
  * Added API and model tests covering seeded connection visibility and restart survivability.


marblom007 (approved):
Reviewed as a maintainer.

Registry-driven boot seeding of system-owned Artifact Hub + GitHub connections. Seeds only kinds that hold a registrant AND whose credentialSchema requires nothing; no kinds hard-coded in server code; in-place writes so no duplicate rows; user/status connections untouched.

Verified through the full validation pipeline, which surfaced and resolved several real issues (captured here per our review practice):
- Duplicate-registrants-per-kind (absent vs zero-UUID content-addressed hash): resolved with a deterministic canonical registrant per kind + softened doc guarantee; the underlying identity fix is filed as follow-up #20950 (now open as schemas PR #1093).
- Registrant rule externally inducible via POST /api/registry/connections: narrowed to require the registrant to own real model entities.
- Credential invariant: applied only at initial materialization (mirrors the status carve-out) so a user-attached optional credential is never silently reverted on reboot.
- Sticky canonical pick (prefer already-stamped, tie-break lowest id) so the stamp doesn't move when the registrant set grows.
- Legacy-vs-canonical column bug (type/user_id vs connection_type/owner) fixed so the seeding is not a no-op on the wire.
- Root-fix: one shared SystemDatabaseModels migration list unifying boot + reset + GraphQL hard-reset, killing the drop-all/re-migrate-subset divergence that left the connections API 500ing after a DB reset.

8 unit tests incl. restart-idempotency and kubernetes-registrant exclusion; docs added. CI green (17 passed, 0 failed; only advisory CodeRabbit pending). Approving.

PR #20965, opened 2026-07-25
[Server] fix: handle negative pagination page size

### LINKED ISSUE #20964: [Server] Negative pagesize bypasses pagination limit across API handlers
### Current Behavior

While testing the connection definitions endpoint locally, I noticed that a negative `pagesize` value bypasses the expected pagination behavior.

With a normal page size:

```text
GET /api/meshmodels/connections?pagesize=1
```

I get:

```text
HTTP: 200

{
  "page": 0,
  "pageSize": 1,
  "totalCount": 3,
  "returnedDefinitions": 1
}
```

This works as expected and returns one connection definition.

However, with:

```text
GET /api/meshmodels/connections?pagesize=-1
```

I get:

```text
HTTP: 200

{
  "page": 0,
  "pageSize": -1,
  "totalCount": 3,
  "returnedDefinitions": 3
}
```

All 3 available connection definitions are returned, and the response reports `pageSize` as `-1`.

I also checked `pagesize=0` and malformed values such as `pagesize=abc`. Both fall back to the default page size of `25`.

### What I traced

The behavior appears to originate in the shared `getPaginationParams()` helper.

A negative integer is valid input to `strconv.Atoi()`, so `-1` is parsed successfully.

The helper currently handles a zero limit:

```go
if limit == 0 {
    limit = defaultPageSize
}
```

but a negative value passes through unchanged.

For the connection definitions endpoint, the path I traced is:

```text
GET /api/meshmodels/connections?pagesize=-1
        |
        v
getPaginationParams()
        |
        | limit = -1
        v
GetConnectionDefinitions()
        |
        v
ConnectionFilter{Limit: -1}
        |
        v
RegistryManager.GetEntities()
        |
        v
ConnectionFilter.Get()
        |
        v
GORM Limit(-1)
```

GORM treats `Limit(-1)` as removing the limit condition. Because of this, the query returns the complete result set instead of applying pagination.

The handler then uses the same `limit` value when constructing the response, resulting in:

```json
{
  "pageSize": -1
}
```

`getPaginationParams()` is shared by multiple server handlers, so the underlying validation issue does not appear to be specific to the connection definitions endpoint. This endpoint is where I reproduced and traced the behavior.

### Expected Behavior

Negative numeric `pagesize` values should not bypass pagination.

For example:

```text
GET /api/meshmodels/connections?pagesize=-1
```

should fall back to the default page size, consistent with how `pagesize=0` and malformed values are currently handled.

The existing `pagesize=all` behavior should remain unchanged.

### Steps to Reproduce

1. Run Meshery Server locally with multiple connection definitions available.

2. Request:

```text
GET /api/meshmodels/connections?pagesize=1
```

The response contains:

```text
pageSize: 1
totalCount: 3
returnedDefinitions: 1
```

3. Request:

```text
GET /api/meshmodels/connections?pagesize=-1
```

The response contains:

```text
pageSize: -1
totalCount: 3
returnedDefinitions: 3
```

4. Observe that the negative page size removes the effective pagination limit and is also returned as negative pagination metadata.

### Possible Fix

The validation can be handled in the shared `getPaginationParams()` helper.

The current condition:

```go
if limit == 0 {
    limit = defaultPageSize
}
```

could account for negative numeric values as well, for example:

```go
if limit <= 0 {
    limit = defaultPageSize
}
```

This check is inside the existing numeric parsing path guarded by:

```go
if limitstr != "all"
```

so the existing `pagesize=all` behavior would remain unchanged.

The `page` parameter already has similar handling for negative values:

```go
if page < 0 {
    page = 0
}
```

### Suggested Regression Coverage

A table-driven test around `getPaginationParams()` could cover:

- `pagesize=-1` -> default page size
- `pagesize=0` -> default page size
- `pagesize=abc` -> default page size
- `pagesize=1` -> `1`
- `pagesize=all` -> preserve the existing unlimited behavior

An endpoint-level regression test for `/api/meshmodels/connections` could additionally verify that a negative page size no longer returns `pageSize: -1` or removes pagination.

### Screenshots / Logs

Normal pagination:

```text
REQUEST:
GET /api/meshmodels/connections?pagesize=1

HTTP: 200

{
  "page": 0,
  "pageSize": 1,
  "totalCount": 3,
  "returnedDefinitions": 1
}
```

Negative page size:

```text
REQUEST:
GET /api/meshmodels/connections?pagesize=-1

HTTP: 200

{
  "page": 0,
  "pageSize": -1,
  "totalCount": 3,
  "returnedDefinitions": 3
}
```

### Environment

- Host OS: Linux
- Meshery Server: local development build

SuryanshGarg04:
## Description

Fixes #20964

This PR fixes the handling of negative `pageSize` values in the shared `getPaginationParams()` helper.

While testing the connection definitions endpoint, I found that a negative page size such as:

```text
GET /api/meshmodels/connections?pagesize=-1
```

was parsed successfully by `strconv.Atoi()` and passed downstream as `limit = -1`.

The existing pagination validation only handled `limit == 0`:

```go
if limit == 0 {
    limit = defaultPageSize
}
```

When the negative value reached GORM as `Limit(-1)`, GORM removed the limit condition, causing the complete result set to be returned. The response also reported `"pageSize": -1`.

This change normalizes non-positive numeric page sizes to `defaultPageSize`:

```go
if limit <= 0 {
    limit = defaultPageSize
}
```

The special `pageSize=all` behavior remains unchanged because it bypasses this normalization block and continues to use `limit = 0`.

The helper supports both the canonical `pageSize` parameter and the legacy `pagesize` spelling, so regression coverage includes both.

## Current Behavior

Before the fix:

```text
REQUEST:
GET /api/meshmodels/connections?pagesize=-1

HTTP: 200

{
  "page": 0,
  "pageSize": -1,
  "totalCount": 3,
  "returnedDefinitions": 3
}
```

A negative page size removes the query limit and the API reports an invalid negative `pageSize`.

## Expected Behavior

Negative numeric page sizes should fall back to the default page size, consistent with zero and malformed page-size values.

After the fix:

```text
REQUEST:
GET /api/meshmodels/connections?pagesize=-1

HTTP: 200

{
  "page": 0,
  "pageSize": 25,
  "totalCount": 26,
  "returnedDefinitions": 25
}
```

The number of connection definitions in my local environment increased between the before and after runs. The relevant values here are the reported `pageSize` and the number of returned definitions relative to the requested page size.

## No Regression

Positive page sizes continue to work as expected:

```text
REQUEST:
GET /api/meshmodels/connections?pagesize=1

HTTP: 200

{
  "page": 0,
  "pageSize": 1,
  "totalCount": 26,
  "returnedDefinitions": 1
}
```

The existing unlimited behavior is also preserved:

```text
REQUEST:
GET /api/meshmodels/connections?pagesize=all

HTTP: 200

{
  "page": 0,
  "pageSize": 26,
  "totalCount": 26,
  "returnedDefinitions": 26
}
```

## Tests

Added table-driven regression coverage for:

- canonical `pageSize=-1`
- legacy `pagesize=-1`
- `pageSize=0`
- malformed `pageSize=abc`
- positive `pageSize=1`
- `pageSize=all`

Test result:

```text
=== RUN   TestGetPaginationParams
=== RUN   TestGetPaginationParams/negative_canonical_pageSize_defaults
=== RUN   TestGetPaginationParams/negative_legacy_pagesize_defaults
=== RUN   TestGetPaginationParams/zero_pageSize_defaults
=== RUN   TestGetPaginationParams/malformed_pageSize_defaults
=== RUN   TestGetPaginationParams/positive_pageSize_preserved
=== RUN   TestGetPaginationParams/all_preserves_unlimited_behavior
--- PASS: TestGetPaginationParams (0.00s)
PASS
ok      github.com/meshery/meshery/server/handlers      1.194s
```

## Notes

The fix is made in the shared pagination helper rather than in the connection definitions handler, so invalid negative page sizes are normalized at the point where pagination parameters are parsed.

No changes are made to MeshKit's `ConnectionFilter` behavior or to the existing `pageSize=all` semantics.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid pagination `pageSize` values, including zero and negative inputs, now consistently fall back to the default page size.
  * Positive `pageSize` and `pageSize=all` (unlimited) behavior remain unchanged.

* **Tests**
  * Added coverage for valid, invalid, zero, legacy, and unlimited pagination inputs.


KumarNirupam1 (approved):
thanks for this small fix 
lgtm

ritzorama (commented):
When is the negative pagination size occurring?

SuryanshGarg04:
@ritzorama  I traced where the negative value could come from. It isn't generated by the normal UI or CLI flow. The case I reproduced was through a direct API request, for example `GET /api/meshmodels/connections?pagesize=-1`.

Since `getPaginationParams()` only handled limit == 0, -1 was passed through to GORM, where `Limit(-1)` removes the query limit. So this is mainly API side input hardening to make sure negative page sizes fall back to the default instead of changing the query behavior.

reaper8055 (approved):
Lgtm

lekaf974 (approved):

PR #20970, opened 2026-07-26
[mesheryctl] fix: return error on non-200 spreadsheet fetch response

### LINKED ISSUE #20902: [mesheryctl] registry generate/update report success (exit 0) when the spreadsheet fetch returns a non-200
### Description

Both `mesheryctl registry generate` and `mesheryctl registry update` fetch the integration spreadsheet and then guard the result like this:

`internal/cli/root/registry/generate.go` (~L196):
```go
resp, err := srv.Spreadsheets.Get(spreadsheeetID).Fields().Do()
if err != nil || resp.HTTPStatusCode != 200 {
    utils.Log.Error(ErrUpdateRegistry(err, outputLocation))
    return err
}
```

`internal/cli/root/registry/update.go` (~L88) has the same shape.

Two defects:

1. **Failure reported as success.** When the fetch returns a non-200 status but `err == nil` (e.g. 403/404/429 from the Sheets API), the branch is taken but `return err` returns `nil`, so the command exits 0 as if it succeeded - no models were generated/updated, yet nothing signals failure.
2. **Misleading error log.** In that same case `ErrUpdateRegistry(err, ...)` is logged with a `nil` error, producing a confusing/empty cause. (`generate.go` also passes `outputLocation` rather than the model location.)

### Expected

A non-200 response should be reported as an error and cause a non-zero exit, with the HTTP status included in the message.

### Actual

Non-200 responses are logged with a `nil` cause and the command still exits successfully.

### Suggested fix

Construct an explicit error when `resp.HTTPStatusCode != 200` (including the status code) and return it, e.g.:
```go
if err != nil {
    utils.Log.Error(ErrUpdateRegistry(err, modelLocation))
    return err
}
if resp.HTTPStatusCode != 200 {
    e := ErrUpdateRegistry(fmt.Errorf("unexpected HTTP status %d fetching spreadsheet", resp.HTTPStatusCode), modelLocation)
    utils.Log.Error(e)
    return e
}
```

### Environment

- mesheryctl built from `master`.

Vijesh-CDmaster:
## Description

`mesheryctl registry generate` and `mesheryctl registry update` fetch the
integration spreadsheet before generating/updating models. Previously,
when the fetch returned a non-200 HTTP status (e.g. 403, 404, 429) with
a nil `error`, both commands logged a confusing error with a nil cause
and then returned `nil`/`err` (which was itself nil) — exiting 0 as if
the operation succeeded, even though no models were actually
generated or updated.

## Changes

- Extracted the fetch-validation logic into a shared `checkSpreadsheetFetch`
  helper (in `generate.go`, used by both `generate.go` and `update.go`)
  that returns an explicit, non-nil error for both network failures and
  non-200 HTTP responses, including the actual status code in the message.
- Added `spreadsheet_fetch_test.go` with unit tests covering: a network
  error, a 403 response, a 429 response, and a 200 (success) response.
- In `update.go`, `RunE` no longer swallows the error returned by
  `InvokeCompUpdate()` — it now propagates it instead of returning `nil`,
  so update failures also result in a non-zero exit code.

## Testing

- `go build ./...` and `go vet ./...` pass cleanly.
- Existing test suite in `internal/cli/root/registry/` passes (two
  pre-existing, Windows-environment-only failures in `purge_test.go`
  are unrelated to this change — confirmed present on `master` prior
  to this PR).
- New `TestCheckSpreadsheetFetch` covers all 4 scenarios described above.

Fixes #20902

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved spreadsheet download validation for registry generation and updates.
  - Added clear error messages for network failures and unexpected HTTP responses, including rate limiting (HTTP 429).
  - Registry update failures are now properly reported instead of being silently treated as successful.
  - Component processing errors are now surfaced during registry updates.
  - Successful spreadsheet requests continue through the normal generation and update workflows.

### DIFF
--- a/mesheryctl/internal/cli/root/registry/generate.go
+++ b/mesheryctl/internal/cli/root/registry/generate.go
@@ -58,11 +58,15 @@ var (
 )
 
 // checkSpreadsheetFetch validates the result of a spreadsheet fetch and
-// returns an explicit, non-nil error for network failures or non-200
-// HTTP responses, so callers never mistake a failed fetch for success.
+// returns an explicit, non-nil error for network failures, a nil response,
+// or non-200 HTTP responses, so callers never mistake a failed fetch for
+// success or dereference a nil response.
 func checkSpreadsheetFetch(resp *sheets.Spreadsheet, err error, location string) error {
 	if err != nil {
 		return ErrUpdateRegistry(err, location)
+	}
+	if resp == nil {
+		return ErrUpdateRegistry(fmt.Errorf("received nil response fetching spreadsheet"), location)
 	}
 	if resp.HTTPStatusCode != 200 {
 		return ErrUpdateRegistry(fmt.Errorf("unexpected HTTP status %d fetching spreadsheet", resp.HTTPStatusCode), location)
--- a/mesheryctl/internal/cli/root/registry/generate.go
+++ b/mesheryctl/internal/cli/root/registry/generate.go
@@ -56,6 +56,20 @@ var (
 	// Whether to generate only the latest version of each model
 	latestVersionOnly bool
 )
+
+// checkSpreadsheetFetch validates the result of a spreadsheet fetch and
+// returns an explicit, non-nil error for network failures or non-200
+// HTTP responses, so callers never mistake a failed fetch for success.
+func checkSpreadsheetFetch(resp *sheets.Spreadsheet, err error, location string) error {
+	if err != nil {
+		return ErrUpdateRegistry(err, location)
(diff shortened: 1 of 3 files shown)


ishwar170695 (commented):
I agree with the existing review here. Updating the caller to `return err` is an improvement, but `InvokeCompUpdate()` still appears to swallow some internal failures by logging and returning `nil`. Those paths should also propagate an error so the command consistently exits non-zero on update failures.

PR #20984, opened 2026-07-28
fix(server): allow connection register before row is persisted

### LINKED ISSUE #20983: [Server] Prometheus/Grafana connection create fails with record not found
### Current Behavior

Creating a **Prometheus** or **Grafana** connection from the **Create Connection** wizard fails on **Review & Create**, even when the target endpoint is running and credential verification is bypassed.

Observed UI errors:

- `record not found` (banner on the Review step)
- `Failed to retrieve the connection with id <uuid> to update status.`
- `Failed to create Prometheus connection: record not found` (and the same shape for Grafana)

**What is happening under the hood:**

The wizard calls `POST /api/integrations/connections/register` twice:

1. `status: "register"` - runs kind-specific verification (Prometheus/Grafana `RegisterAction`)
2. `status: "connect"` - should persist via `DefaultConnectAction`

On `register`, the state machine advances in-memory to `registered` but **does not insert a connection row**. After the transition, `SendEvent` still calls `GetConnectionByID` to update status. Because no row exists yet, the provider returns GORM `record not found`, and the request fails before `connect` can persist the connection.

Additionally, Prometheus/Grafana `RegisterAction` returns `machines.Exit` on success, which is not a valid edge in the default connection state graph (unlike other actions that return `NoOp`), which further confuses the post-transition path.

**Relevant code:**

- `server/machines/models.go` - post-transition `GetConnectionByID` status update
- `server/machines/prometheus/register.go` / `server/machines/grafana/register.go` - verify-only register; returns `Exit`
- `server/machines/actions.go` - `DefaultConnectAction` persists on connect
- `ui/components/connections/wizard/genericSteps.tsx` - `register` then `connect`

### Expected Behavior

1. Completing **Create Connection** for Prometheus or Grafana **creates and persists** the connection (and credential when provided).
2. With verification enabled, unreachable endpoints or bad credentials fail with a **clear verification error**, not `record not found`.
3. With verification bypassed, registration succeeds without requiring a live probe, and the connection is still persisted.
4. Status updates after state-machine transitions either:
   - operate on a connection row that already exists for that lifecycle step, or
   - soft-skip when the connection has not been persisted yet (early `register` phase),
   and must **not** surface a generic `record not found` as a create failure.

**Acceptance criteria**

- [ ] Wizard create for Prometheus succeeds end-to-end
- [ ] Wizard create for Grafana succeeds end-to-end
- [ ] Bypass verification does not require a reachable URL
- [ ] Failed probe (verification on) returns a health/scan error, not `record not found`
- [ ] No reliance on UI filter PRs (#20617) for create to work

### Screenshots/Logs

Image

### Environment

- **Host OS:**  Windows

### To Reproduce
1. Open **Create Connection**.
2. Choose **Prometheus** (or Grafana).
3. Configure connection (name/URL) and credential (or bypass verification).
4. On **Review & Create**, click **Create Connection**.
5. Registration fails; connection is not created.

---

chaitanyamedidar:
**Notes for Reviewers**

- This PR fixes #20983 

Register verifies before DefaultConnectAction inserts the row. Treat GetConnectionByID 404 as not-yet-persisted instead of failing wizard create. Return NoOp from prom/grafana register actions.

Screenshot:
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Grafana and Prometheus registration now complete cleanly without triggering unintended follow-up transitions.
  * Registration succeeds during early setup when connection records are not yet available.
  * Connection status synchronization now handles missing connection records while still reporting genuine lookup errors.
  * Provider failures continue to be surfaced without attempting invalid status updates.

* **Tests**
  * Added coverage for early registration, missing connections, and provider error handling.


ishwar170695 (commented):
1. Can we scope the 404 skip specifically to the pre-persistence `register` flow? A 404 on an already-persisted connection could be a real error and shouldn't get silently skipped.

2. Can we add a test for the actual Prometheus/Grafana `RegisterAction` returning `NoOp`? The current test only covers the generic `SendEvent` path.

SuryanshGarg04:
Went through the PR and couldn't find any blocking issues apart from the points already raised by @ishwar170695. LGTM once those comments are addressed.

server/machines/grafana/register.go
```diff
@@ -71,7 +71,8 @@ func (ra *RegisterAction) Execute(ctx context.Context, machineCtx interface{}, d
 	if err != nil && !connPayload.SkipCredentialVerification {
 		return machines.NoOp, eventBuilder.WithMetadata(map[string]interface{}{"error": models.ErrGrafanaScan(err)}).Build(), models.ErrGrafanaScan(err)
 	}
-	return machines.Exit, nil, nil
+	// Stay in REGISTERED. Exit is not a machine edge and was misread as a follow-up event.
```
banana-three-join: Correct change. `machines.Exit` was incorrectly being used as an `exit` to the execute. The comment above the return isn't necessary. It could be abstracted away into the PR's description since this is more of a description to the fix rather than an explanation of the behavior within the execute's return (Which isn't needed as now due to it following the default `return machines.NoOp, nil, nil` pattern)

server/machines/prometheus/register.go
```diff
@@ -72,7 +72,8 @@ func (ra *RegisterAction) Execute(ctx context.Context, machineCtx interface{}, d
 	if err != nil && !connPayload.SkipCredentialVerification {
 		return machines.NoOp, eventBuilder.WithMetadata(map[string]interface{}{"error": models.ErrPrometheusScan(err)}).Build(), models.ErrPrometheusScan(err)
 	}
-	return machines.Exit, nil, nil
+	// Stay in REGISTERED. Exit is not a machine edge and was misread as a follow-up event.
```
banana-three-join: Same feedback as in `grafana/register.go`. Good fix but this can do without the comment above the `return machines.NoOp, nil, nil`

server/machines/models.go
banana-three-join, line 1: **The block got harder to follow**
 
The old code was three early returns and then the happy path at top level. The new `if / else if / else if / else` mixes arms that fall through with arms that return, and the twenty lines of actual work end up in the trailing `else`, indented further than anything else in the function. The two returning arms also build the same "Failed to retrieve the connection" event twice.
 
Inverting keeps the original body untouched:
 
```
notYetPersisted := sm.PersistsOnConnect &&
    originalEventType == Register &&
    statusCode == http.StatusNotFound

if notYetPersisted {
    sm.Log.Debugf("%s: connection %s not yet persisted after register; skipping status update", sm.Name, sm.ID)
} else {
    // original body, guard clauses intact
}
```
 
`connectionMissing` folds into that condition. With one call site and a single comparison it's hard for it to earn a function, and its own doc comment notes it isn't sufficient on its own — naming the full rule instead says more at the point where it matters. The "status code is the contract, don't string-match error text" note is worth keeping, just on the branch.
 
**Two comments to adjust**
 
"Prometheus/grafana verify on register but only insert the row on connect" reads as a two-kind special case, but per `getMachine` it's how the default machine behaves for every non-k8s kind.
 
On the `connection == nil` arm, "Non-404 empty result — not the early-lifecycle missing case" describes something that can't happen: both providers return a non-nil error whenever they return a nil connection, so that arm is unreachable. The original comment about guarding against a nil dereference was accurate.

banana-three-join (changes requested):

server/machines/grafana/register.go
chaitanyamedidar: Agreed.

server/machines/prometheus/register.go
chaitanyamedidar: Done.

server/machines/models.go
chaitanyamedidar: Took the invert so the original early returns stay at the top and the status write is not buried in a trailing else. Folded `connectionMissing` into `Register && 404`; one comparison did not need a helper.

Kept the pre-existing nil-connection comment (deref guard). Dropped the persist-on-connect / 404-skip comment we had added; `notYetPersisted` already states that rule.

Did not add `sm.PersistsOnConnect`. That is a new field on every machine. The skip only needs to fire on register-before-persist; a persist-on-connect flag is separate lifecycle work.

server/machines/models_test.go
```diff
@@ -513,6 +520,94 @@ func (a *failingExitAction) ExecuteOnExit(context.Context, interface{}, interfac
...
+	event, err := sm.SendEvent(ctx, Register, nil)
+	if err == nil {
+		t.Fatal("expected GetConnectionByID provider error to fail SendEvent, got nil")
+	}
+	if !errors.Is(err, getErr) {
+		t.Fatalf("expected the original provider error, got %v", err)
+	}
+	if event == nil {
+		t.Fatal("expected an error event when status retrieval fails, got nil")
+	}
```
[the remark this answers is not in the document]
chaitanyamedidar: The test already asserts the skip boundary, the original provider error returns and we do not call `UpdateConnectionById`.

server/machines/models.go
```diff
@@ -310,44 +311,52 @@ func (sm *StateMachine) SendEvent(ctx context.Context, eventType EventType, payl
...
-		if connectionPayload.MetaData == nil {
-			connectionPayload.MetaData = map[string]interface{}{}
-		}
+			connection, err = sm.Provider.UpdateConnectionById(token, connectionPayload, sm.ID.String())
 
-		connection, err = sm.Provider.UpdateConnectionById(token, connectionPayload, sm.ID.String())
+			if err != nil {
+				// In this case should the current state be again set to previous state i.e. should we rollback. But not only state should be rollback but other actions as well, rn we don't rollback state.
+				return events.NewEvent().WithDescription(fmt.Sprintf("Operation succeeded but failed to update the status of the connection to %s.", sm.CurrentState)).WithMetadata(map[string]interface{}{"error": err}).FromSystem(*sysID).FromOwner(userUUID).ActedUpon(sm.ID).WithCategory("connection").WithAction("update").Build(), err
+			}
```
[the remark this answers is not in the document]
chaitanyamedidar: The (*Connection, error) signature can be (nil, nil), so that Debugf could panic. In-tree local/remote providers do not return that (connection or err). The log line is also pre-existing. Not adding a nil check in this 404-skip PR.

server/machines/grafana/register_test.go
```diff
@@ -0,0 +1,62 @@
...
+	userID, err := uuid.NewV4()
+	if err != nil {
+		t.Fatalf("user uuid: %v", err)
+	}
+	sysID := core.Uuid(uuid.Nil)
+	ctx := context.Background()
+	ctx = context.WithValue(ctx, models.UserCtxKey, &models.User{ID: core.Uuid(userID)})
+	ctx = context.WithValue(ctx, models.SystemIDKey, &sysID)
+	return ctx
+}
```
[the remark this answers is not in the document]
chaitanyamedidar: The helpers are duplicated. Leaving them in-package next to each RegisterAction. A shared testutil for about 12 lines is more machinery than it saves, the health URLs still differ.

banana-three-join (changes requested):
I got curious about the fix so, I went through the surrounding code and I think the fix belongs somewhere other than where the PR puts it. That's not a reflection on the patch - the diagnosis in #20983 is correct and the `Exit` → `NoOp` change in the register actions is right and worth keeping. It's that the reconciliation step turns out to be the wrong place to solve it. Here's what I found and the options as I see them.

**The current pipeline**

`SendEvent` ends with a block (`server/machines/models.go:311`) that syncs the machine's state onto the connection row: fetch by ID, compare status, write back. It runs after every transition and assumes a row exists.

For the default machine, nothing writes a row until the connect event —
`DefaultConnectAction.Execute` (`server/machines/actions.go:94`) is the only writer.
Kubernetes is the opposite: `SaveK8sContext` writes the row in the handler during cluster import, before any event is dispatched, and `DiscoverAction` writes again on entry to `DISCOVERED`.

So on a register event for any non-Kubernetes kind, the machine advances to `REGISTERED`, the probe succeeds, reconciliation asks for a row that has never been written, and the 404 becomes an error. The wizard (`ui/components/connections/wizard/genericSteps.tsx:276`) awaits register before connect, so the `catch` fires and the connection is never created.

Root cause: two legitimate states of the world - a connection mid-registration, and a connection whose row was lost - are indistinguishable at the point where the code has to decide what to do.

Regarding the current approach, there's a huge issue which is that it makes
the fetch's status code load-bearing, and that code doesn't mean the same thing on both providers.

`DefaultLocalProvider.GetConnectionByID` synthesizes 404 from `gorm.ErrRecordNotFound` - that genuinely is a statement about persistence. `RemoteProvider.GetConnectionByID` (`server/models/remote_provider.go:4608`) returns `resp.StatusCode` verbatim for any non-200 response, and discards the errors from both `url.Parse` and `http.NewRequest`. A stale capability endpoint, a gateway, or a provider mid-deploy all arrive as 404 and would now be read as "not yet persisted" and logged at debug.

That's true regardless of which fix we choose, and it's why I don't want to merge the skip as written.

**The options**

**A. Tolerate absence, inferring when it's expected.** What the PR does now: decide from the event name (or the kind, or the machine) that a missing row is acceptable here. The problem is that each of those correlates imperfectly with the thing we actually need to know. `Register` reaches Kubernetes too, via the status-update path in `connections_handlers.go:564`. And a default machine can have a row: after a cache miss, `InitializeMachineWithContext` (`server/machines/helpers/helpers.go:129`) builds a fresh instance at whatever initial state the caller passes - `connections_handlers.go:74` always passes `DISCOVERED` - without ever reading the database. Narrowing the condition reduces the mistake without removing it.

**B. Tolerate absence unconditionally.** No inference: if there's no row, there's nothing to sync, so warn and continue - every kind, every event. Honest and small. What we give up is any signal when a row genuinely vanished, and we accept that registration can succeed with nothing in the database, which relocates the surprise rather than removing it.

**C. Guarantee the row exists, from the handler.** Persist in `connections_handlers.go` before dispatching the register event, using the payload it already holds. Reconciliation then stays strict and a 404 means what it says.

This is the one I'd argue for, mainly because it isn't a new pattern: it's what the
Kubernetes path already does. `SaveK8sContext` writes in the handler and then sends the event. Applying that consistently makes both lifecycles say the same thing.

**D. Guarantee the row exists, from a new register action.** Same guarantee, attached to the `REGISTERED` state instead. I looked at this first and moved away from it: `State` holds a single `Action` (`server/machines/state.go:6`), so grafana and prometheus would need a composite action to run both their probe and the write, with ordering and short-circuit semantics that are easy to get subtly wrong. It also bundles in a separate behaviour change (see below).

One thing that rules out a fifth option worth mentioning: the insert can't live inside the reconciliation block itself. That block builds its payload from the row it just fetched - `Kind` and `MetaData` both come from `connection` - so it has no independent source for a connection's identity. Making it create a row would mean type-asserting `SendEvent`'s `payload interface{}`, which is a `ConnectionPayload` on register but `nil` on the background discovery calls from `middlewares.go`. The step is a status writer; it reads identity from the thing it writes to.

For the final decision regarding the most "proper" fix, any of C or D needs these answered first:

**Credential attachment.** `ConnectionPersister.SaveConnection`
(`server/models/connection_persister.go:151`) upserts, and on collision heals only `Kind`, `Name`, `ConnectionType`, `SubType`, `Status`, and `Metadata`. `CredentialID` isn't in that list. A row created at register without a credential would never get one, because connect's `SaveConnection` takes the collision branch. That heal logic was written for Kubernetes re-import, where deterministic IDs collide — it wasn't designed for a create-then-enrich flow, and we'd be borrowing a contract written for something else.

**Remote provider behaviour.** `RemoteProvider.SaveConnection` (`:4491`) is a plain POST. Register-then-connect posts the same ID twice, and whether that upserts or duplicates lives in Meshery Cloud, not in this repo. Notably `UpdateConnectionById` (`:4763`) already does a GET-then-merge before its PUT specifically to avoid duplicate rows, which suggests the remote side has needed care here before.

**Orphan rows.** If connect fails or someone abandons the wizard after register, a
`REGISTERED` row stays behind where today nothing persists. Kubernetes already leaves `DISCOVERED` rows around, so there's precedent, but it changes what appears in the connections list and there's no cleanup path.

**Whether register and connect should stay two round-trips.** The wizard sends them back to back with the same payload and does nothing in between. If register exists to verify before persisting, the split is meaningful. If it's historical, collapsing them server-side removes this window entirely.

**Which brings us to the scope of this PR**

Keep: the `NoOp` change in `grafana/register.go` and `prometheus/register.go`. That's correct independent of everything above.

Drop: the 404 skip in `SendEvent`.

I'd rather not have you build option C until the credential and remote-provider questions are answered, since either could change its shape. I'll get those settled. If you'd like to stay on this, the most useful thing right now would be confirming the non-probe-kind failure above against a running instance - it's a quick check and it decides how much else moves.

Maanvi212006 (commented):
So, I went through this PR in detail. All tests passing. 

@banana-three-join I dug into your Option C concern around credential attachment and can confirm it's a real issue. Looked at `ConnectionPersister.SaveConnection`, and here's what I found. On a fresh insert, it just saves the connection as it is, including whatever `CredentialID` is on the payload, which would be empty if this is called during Register. On a collision, meaning the row already exists, it only heals `Kind`, `Name`, `ConnectionType`, `SubType`, `Status`, and `Metadata` from the incoming payload. `CredentialID` is never touched, so whatever value the existing row already has just stays as it is.

So if Register ends up creating the row early with no credential attached, and Connect later tries to save the same connection ID with the real credential, the collision branch would silently keep the empty `CredentialID` forever. Option C would need the healing logic in `SaveConnection` extended to also backfill `CredentialID` when it's empty, or it would need a separate update path altogether.

I agree with keeping the `Exit` to `NoOp` change in `grafana/register.go` and `prometheus/register.go`, and dropping the 404 skip in `SendEvent` for now until the credential and remote provider questions are settled.

I'd like to review this PR again and I'm happy to help further with this PR or any other reviews going forward.

PR #20987, opened 2026-07-28
[Chore]: Bump meshery/schemas to v1.3.56

l5io:
Bump `github.com/meshery/schemas` to v1.3.56

_This pull request has been auto-generated by [meshery-ci](https://github.com/meshery-ci)_


pontusringblom:
Hold before merging: once this bump picks up v1.3.53 it carries the new Environment \`purpose\` property into Meshery Server.

Merging it before meshery/meshery#21802 lands will break environment creation at runtime. The property is specified as \`NOT NULL DEFAULT 'user'\`, and GORM's AutoMigrate cannot apply that constraint to an existing table, so the model and the schema diverge on the first write.

Sequence: land #21802 first, then this bump. The contract is documented in \`docs/environment-purpose-contract.md\` in meshery/schemas.

Marking draft so it is not merged on the routine chore path in the meantime.

PR #20996, opened 2026-07-28
[bug]: Fix docs preview URL not being commented on PRs

banana-three-join:
## Problem

The preview deploys, but no comment is ever posted.

`pr-preview-action` delegates commenting to `marocchino/sticky-pull-request-comment`, passing only `header` and `message` — never `number`. That action resolves its target as:

```ts
context?.payload?.pull_request?.number || +core.getInput("number")
```

We deploy from `workflow_run`, where `payload.pull_request` is undefined, and the unset input coerces to `0`. The step logs `no pull request numbers given: skip step` and exits successfully — a green run with no comment.

The action's `pr-number` input doesn't help: it only feeds the preview path and is not wired to the comment step, so this can't be fixed through the action's inputs.

## Changes

**`comment: false` + explicit comment step** — disables a step that is a guaranteed no-op under `workflow_run`, and posts the comment ourselves with `number:` from the PR metadata artifact. `header: pr-preview` is kept so the comment shares a marker with the prune notice and updates in place instead of stacking.

**`pages-base-url: docs.meshery.io`** — when unset, the action derives the base URL from the repo slug alone, yielding `meshery.github.io/meshery`. It never reads the Pages config, so it doesn't know about the `gh-pages` CNAME. That URL only worked via GitHub's redirect to the custom domain. Setting it explicitly makes the commented URL canonical.

**`umbrella-dir: pr-preview`** — no behavioural change; this is the existing default. Stated explicitly because it must equal the `/pr-preview/` prefix the build bakes via `--baseURL`. Hugo decides what URLs go into the HTML, the action decides where files land, and nothing enforces agreement — if they diverge the deploy still reports success while assets resolve against the production site root.

## Summary by CodeRabbit

* **Documentation**
  * Documentation preview links are now posted or updated in a sticky pull request comment.
  * Preview comments are managed without generating duplicate comments.


.github/workflows/deploy-docs-preview.yml
```diff
@@ -60,13 +60,25 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Deploy PR preview
+        id: preview
         uses: rossjrw/pr-preview-action@ffa7509e91a3ec8dfc2e5536c4d5c1acdf7a6de9 #1.8.1
         with:
           source-dir: ./docs-preview-build/docs/public
+          pages-base-url: docs.meshery.io # otherwise meshery.github.io/meshery is used as a default
+          umbrella-dir: pr-preview # must match the /pr-preview/ baseURL the build bakes
```
ianrwhitney, line 68: this is default option for umbrella-dir. lets keep inputs lean if they are defaults

ianrwhitney (approved):
lgtm. one nit. lets merge and see if it works and we can continue to iterate.

PR #21000, opened 2026-07-29
Fix 20901 model import schema error

### LINKED ISSUE #20901: [mesheryctl] model init scaffolds v1beta3 components that released servers reject with an opaque "No valid component" error
### Description

`mesheryctl model init` scaffolds component definitions with `schemaVersion: components.meshery.io/v1beta3` (the version shipped in the `@meshery/schemas` component template). When that model is packaged with `mesheryctl model build` and registered into a **released** Meshery Server with `mesheryctl model import`, every component is silently dropped and the import fails with a generic, non-actionable error:

```
Error: No valid component or relationship found in the Model provided. A Model can be only imported if it contains at least one valid Component or Relationship.
```

Root cause: MeshKit's registration classifier (`models/registration/utils.go` `getEntity`) switches on `schemaVersion`. Released servers (verified on `playground.meshery.io`, v1.0.62) only recognize `components.meshery.io/v1beta1`; a component carrying the newer `v1beta3` string falls through to the `default` branch and is discarded. Because the message never names the offending `schemaVersion`, the user has no way to discover that the scaffold output is simply ahead of the target server.

### Steps to reproduce

1. `mesheryctl model init my-icons --version v0.1.0`
2. Fill in `my-icons/v0.1.0/model.json` and add one or more component JSON files under `components/` (the scaffolded files keep `schemaVersion: components.meshery.io/v1beta3`).
3. `mesheryctl model build my-icons/v0.1.0`
4. `mesheryctl model import -f my-icons-v0-1-0.tar` against a released server.

### Expected

Either the scaffold emits a component `schemaVersion` the current release accepts, or the import error names the unsupported `schemaVersion` (e.g. "component 'x' uses unsupported schemaVersion components.meshery.io/v1beta3") so the fix is discoverable.

### Actual

All components are dropped; the import fails with a generic "No valid component or relationship found" message. Rewriting the components to `components.meshery.io/v1beta1` (and the model to `models.meshery.io/v1beta2`) makes the identical model import cleanly with all components.

### Suggested fixes

- Surface the unrecognized `schemaVersion` in the registration classifier's error instead of silently dropping the entity.
- Consider having `model init` scaffold the `schemaVersion` accepted by the release channel the CLI targets, or document the version requirement in the [Create Models](https://docs.meshery.io/guides/configuration-management/creating-models) guide (a compatibility note has been added there in the accompanying PR).

### Environment

- mesheryctl built from `master`; server: playground.meshery.io (v1.0.62)

Hemanth-katariya:
Fixes #20901

Older Meshery Servers (e.g. `v1.0.62` and prior) only support `components.meshery.io/v1beta1` schemaVersions, and silently reject newer components without returning an error. `mesheryctl model import` then surfaces this as a generic `ErrInvalidModel` ("No valid component or relationship found"). 

This PR updates `ErrInvalidModel` to mention `schemaVersion` incompatibility as a probable cause and provides a remediation step to verify the schemaVersion in use.

## Summary by CodeRabbit

- **New Features**
  - Added TOON output support across Meshery CLI views, exports, evaluations, and performance results.
  - TOON-formatted output can be displayed or saved to a file.
  - Updated command options and examples to document JSON, YAML, and TOON formats.
  - Added clearer errors and guidance for unsupported formats and model imports.

- **Tests**
  - Added coverage for displaying and saving TOON output, including missing-file-path errors.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -369,6 +369,7 @@ require (
 	github.com/tonistiigi/go-csvvalue v0.0.0-20240814133006-030d3b2625d0 // indirect
 	github.com/tonistiigi/units v0.0.0-20180711220420-6950e57a87ea // indirect
 	github.com/tonistiigi/vt100 v0.0.0-20240514184818-90bafcd6abab // indirect
+	github.com/toon-format/toon-go v0.0.0-20251202084852-7ca0e27c4e8c // indirect
 	github.com/valyala/fastjson v1.6.5 // indirect
 	github.com/vbatts/tar-split v0.12.2 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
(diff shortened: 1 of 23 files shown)


Hemanth-katariya:
Closing this PR because it was based on my previous feature branch. I've opened a clean PR containing only the changes for #20901.

PR #21033, opened 2026-07-31
[mesheryctl] Exit instead of hanging when the dashboard port-forward drops

### LINKED ISSUE #21020: [mesheryctl] system dashboard --port-forward hangs after the tunnel drops instead of exiting
### Current Behavior

If the port forward behind `mesheryctl system dashboard --port-forward` dies while it's
running pod restart, rollout, network blip the command doesn't notice. The browser
starts getting connection refused, the CLI still prints that the UI is available, and the
process sits there until you Ctrl C it. Nothing reports that the tunnel is gone.

It looks like the error never makes it out of `run()` in
[`mesheryctl/pkg/utils/portforward.go#L148-L171`]:

```go
failure := make(chan error)          // unbuffered

go func() {
    if err := pf.run(); err != nil {
        failure <- err
    }
}()

// The `select` statement below depends on one of two outcomes from `pf.run()`:
// 1) Succeed and block, causing a receive on `<-pf.readyCh`
// 2) Return an err, causing a receive `<-failure`
select {
case <-pf.readyCh:
case err := <-failure:
    return err
}
```

The comment says one of two outcomes, but client-go has a third `ForwardPorts()` closes
`Ready` *first* and only then blocks, so it can succeed and *then* fail
(`k8s.io/client-go/tools/portforward/portforward.go`, v0.35.3):

```go
if pf.Ready != nil {
    close(pf.Ready)
}

// wait for interrupt or conn closure
select {
case <-pf.stopChan:
case <-pf.streamConn.CloseChan():
    return ErrLostConnectionToPod
}
```

By then `Init()` has already returned and nothing is receiving on `failure`, so the send
parks forever and `ErrLostConnectionToPod` is never surfaced. `stopCh` is only ever closed
by `Stop()`, so [`dashboard.go#L182`](mesheryctl/internal/cli/root/system/dashboard.go#L182)'s

`<-portforward.GetStop()` never returns. `ForwardPorts` also does `defer pf.Close()`, which
shuts the local listener that's the connection-refused part.

Wanted to check whether this is intentional: the `failure` channel handles the pre-ready
case correctly, so it's possible the post-ready path was just out of scope at the time.

### Expected Behavior

When the tunnel drops, the command should unblock and exit, reporting the underlying
`lost connection to pod` error, instead of hanging silently while advertising a URL that
no longer answers.

### Screenshots/Logs

`run()` is unexported and needs a live cluster, so I couldn't unit-test the real type. I
modelled `Init()` as written and substituted client-go's actual `forward()` sequencing
(close `Ready`, block, then return `ErrLostConnectionToPod`):

```
=== RUN   TestInitLeaksRunGoroutineWhenTunnelDiesAfterReady
    LEAK CONFIRMED: run goroutine parked forever on `failure <- err`;
                    ErrLostConnectionToPod is never surfaced
    HANG CONFIRMED: stopCh still open, dashboard stays parked on
                    <-portforward.GetStop() with a dead tunnel
--- PASS: TestInitLeaksRunGoroutineWhenTunnelDiesAfterReady (2.20s)
=== RUN   TestInitDeliversPreReadyError
--- PASS: TestInitDeliversPreReadyError (0.00s)
PASS
```

The second test covers the pre-ready failure path that works today, so a fix can be shown
not to regress it.

### Environment

- **Host OS:** any
- **Platform:**  Kubernetes
- **Meshery Server Version:** n/a

---

Priyanshu6968:
**Notes for Reviewers**

- This PR fixes #21020 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dashboard port-forward lifecycle handling.
  - Port-forward failures and unexpected tunnel drops are now reported clearly.
  - Stopping port forwarding is safe to repeat and no longer causes shutdown errors.

- **Tests**
  - Added coverage for startup failures, tunnel interruptions, clean shutdowns, and repeated stop requests.

### DIFF
--- a/mesheryctl/internal/cli/root/system/dashboard.go
+++ b/mesheryctl/internal/cli/root/system/dashboard.go
@@ -160,11 +160,13 @@ Note: Meshery's web-based user interface is embedded in Meshery Server and is av
 				// ticker for keeping connection alive with pod each 10 seconds
 				ticker := time.NewTicker(10 * time.Second)
 				go func() {
+					defer ticker.Stop()
 					for {
 						select {
 						case <-signals:
 							portforward.Stop()
-							ticker.Stop()
+							return
+						case <-portforward.GetStop():
 							return
 						case <-ticker.C:
 							keepConnectionAlive(mesheryURL)
@@ -180,6 +182,11 @@ Note: Meshery's web-based user interface is embedded in Meshery Server and is av
 				}
 
 				<-portforward.GetStop()
+				if err := portforward.Err(); err != nil {
+					err = ErrRunPortForward(err)
+					utils.Log.Error(err)
+					return err
+				}
 				return nil
 			}
 
--- a/mesheryctl/pkg/utils/portforward.go
+++ b/mesheryctl/pkg/utils/portforward.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"sync"
 
 	meshkitkube "github.com/meshery/meshkit/utils/kubernetes"
 	log "github.com/sirupsen/logrus"
@@ -31,6 +32,9 @@ type PortForward struct {
 	emitLogs   bool
 	stopCh     chan struct{}
 	readyCh    chan struct{}
+	stopOnce   sync.Once
+	errMx      sync.Mutex
+	err        error
 	config     *rest.Config
 }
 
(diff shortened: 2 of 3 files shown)


Maanvi212006 (changes requested):
Please write a detailed description explaining what this PR does and how it fixes the linked issue

Priyanshu6968:
client-go closes the port-forward's `Ready` channel *before* it starts waiting on the
connection, so `ErrLostConnectionToPod` always comes back after `Init()` has already
returned. At that point the unbuffered `failure` channel in `Init()` had no receiver left,
so the goroutine blocked on the send forever and the error went nowhere. `stopCh` stayed
open too, which is what left `<-portforward.GetStop()` in `dashboard.go` hanging so , the
tunnel was dead but the CLI kept advertising the UI until you Ctrl-C.
The fix buffers that channel so the goroutine can always exit, records the error, and closes
`stopCh` once the forwarder returns, so `GetStop()` finally behaves the way its comment
already claimed. Added `Err()` so the caller can tell a clean `Stop()` apart from a dropped
tunnel, and made `Stop()` idempotent since the forwarder goroutine and the SIGINT handler
can both reach it now. `dashboard.go` checks `Err()` after the wait and exits with
`ErrRunPortForward` instead of returning nil.
`Init()` delegates to an unexported `start(run func() error)` so the tests can drive the
lifecycle without a cluster. The four tests in `portforward_test.go` cover failure before
readiness, a drop after readiness, a clean stop, and repeated `Stop()` calls, the first
three fail against the old code with "timed out waiting for the port-forward to stop",
which is the hang itself.

dhruveshmishra:
@Priyanshu6968 add the details in pr description not in comment

leecalcote (approved):

PR #21061, opened 2026-08-01
Updated mesheryctl-e2e.yaml with correct reference

Salmaan-M:
**Notes for Reviewers**

## Description

The scheduled `mesheryctl-e2e` workflow currently references
`meshery-extensions/mesheryctl-service-mesh-patterns-action`,
which can no longer be resolved (`repository not found`) and causes the workflow to fail during the **Set up job** phase.

This PR updates the scheduled workflow to use
`layer5io/meshery-cloud-native-patterns-action`, which exposes the same action inputs (`provider_token`, `platform`, `pattern_url`, `pattern_file`).

The intent of this change is to validate whether the active action is the correct replacement for the unavailable action and restore execution of the scheduled workflow.

## Related

Discussion with maintainers on Slack indicated that validating this replacement through CI is the appropriate next step.

## Summary by CodeRabbit

* **Chores**
  * Updated end-to-end testing to run consistently across workflow executions.
  * Kubernetes metrics and LLM GPU pattern validation now use the latest cloud-native patterns integration.


marblom007:
The action-repo reference rename itself looks fine, but the PR description only mentions that rename - it doesn't mention that the diff also changes `scheduled-test`'s trigger guard from `github.event_name == 'schedule' && github.repository == 'meshery/meshery'` to `always()`. I agree with CodeRabbit's Major-severity flag on this (restoring the guard plus `!cancelled()`), and want to underline why it matters beyond style: this workflow runs against the dedicated remote-provider test-user token, and `always()` means this job now fires on every trigger this workflow responds to, not just the intended schedule on the canonical repo - that's a meaningful, undisclosed change in when a credentialed job runs, not a reference-path fix.

I also see CodeRabbit's other two open points (the unpinned `@master` third-party action exposing `PROVIDER_TOKEN`, and that action declaring the unsupported Node12 runtime) are still unresolved - agreeing those should block too, not re-explaining them since the fixes are already spelled out there.

Given this touches CI/CD trigger semantics and credential exposure, I don't think this should merge as-is without the description being corrected to disclose the guard change and a maintainer explicitly signing off on the `always()` behavior - flagging this rather than ruling on it myself.

Salmaan-M:
@marblom007 Thanks for pointing this out. The always() change was only added temporarily to allow the scheduled-test job to run in my fork for validation, since the original schedule guard prevents that job from executing outside meshery/meshery. It wasn't intended to be part of the final fix.

I've restored the original schedule-only guard in the latest commit. The scope of this PR is only to update the unavailable action reference. Thanks for catching that.

The switch to meshery-cloud-native-patterns-action was made because the previously referenced meshery-extensions/mesheryctl-service-mesh-patterns-action repository is no longer available.

Regarding the Node12 runtime and pinning to a commit SHA, I'll wait for maintainer guidance on whether this is the intended replacement action before making those changes, since they relate to the replacement action itself rather than the missing repository reference.

PR #21088, opened 2026-08-02
[server] Replace log.Fatal with regexp.MustCompile in filler

### LINKED ISSUE #21087: [Server] Process crashes abruptly due to log.Fatal in pattern filler initialization
### Current Behavior
In `server/models/pattern/stages/filler.go`, the `init()` function uses `log.Fatal()` to handle errors from `regexp.Compile(FillerPattern)`. If the regular expression were to fail to compile for any reason, this would cause an abrupt exit of the entire Meshery server process, which is an anti-pattern for Go applications.

### Expected Behavior
The package-level regex initialization should be refactored to use `regexp.MustCompile()`. This is the idiomatic Go approach for global regular expressions. It removes the need for a manual `init()` block and avoids the use of `log.Fatal()`.

### Screenshots/Logs
N/A - This is a preventative code quality and best-practices improvement.

### Environment

- **Host OS:** Windows
- **Platform:** Docker
- **Meshery Server Version:** latest (master branch)
- **Meshery Client Version:** latest (master branch)

KushagraB424:
**Notes for Reviewers**
Removed the `init()` function and `log.Fatal` in `server/models/pattern/stages/filler.go`. Refactored `FillerRegex` to be initialized directly using `regexp.MustCompile()`, which adheres to Go best practices for package-level regular expressions and avoids abrupt program crashes from `log.Fatal`.

- This PR fixes #21087

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup handling for invalid filler pattern configuration, ensuring failures are surfaced immediately and consistently.

### DIFF
--- a/server/models/pattern/stages/filler.go
+++ b/server/models/pattern/stages/filler.go
@@ -2,7 +2,6 @@ package stages
 
 import (
 	"fmt"
-	"log"
 	"regexp"
 	"strings"
 
@@ -15,15 +14,7 @@ import (
 
 const FillerPattern = `\$\(#ref\..+\)`
 
-var FillerRegex *regexp.Regexp
-
-func init() {
-	var err error
-	FillerRegex, err = regexp.Compile(FillerPattern)
-	if err != nil {
-		log.Fatal("failed to compile filler design regex")
-	}
-}
+var FillerRegex = regexp.MustCompile(FillerPattern)
 
 // Filler - filler stage processes the pattern to subsitute Pattern
 func Filler(skipPrintLogs bool) ChainStageFunction {


marblom007:
This is a correct, idiomatic simplification. `FillerPattern` is a hardcoded constant (never user input), so this only ever fails at build/startup regardless of approach - the old `init()` + `log.Fatal` never had a request-time crash risk to begin with, and this doesn't introduce one. Net code reduction, matches Go convention for a should-never-fail package-level regex. CI is green.

One accuracy nit on the description, not the code: it says this "avoids abrupt program crashes from log.Fatal," but `regexp.MustCompile` still panics (and still halts startup) on a bad pattern - so the real benefit here is idiomatic style and a better stack trace on failure, not crash avoidance. Worth a one-line correction to the PR description so the record is accurate, but doesn't affect the merge.

marblom007 (approved):
Approving on the exact head SHA (9da2720) after re-verifying checks (server unit tests, integration tests, golangci-lint, DCO) are green against this specific commit. Confirmed this is a correct, safe idiomatic simplification - see my review comment above (one non-blocking accuracy nit on the description only, not on the code).

PR #21092, opened 2026-08-03
fix: update Notification filter chips on predefined filter selection

### LINKED ISSUE #15221: [UI] Ensure that Notification Center filter chips are updated upon selection of predefined filter
### Current Behavior

https://github.com/user-attachments/assets/0c761c8f-d929-4e1f-91ca-ed959b1b2705

### Desired Behavior
Filter chips should change when users click on one of the predefined event view buttons.

### Environment

- Meshery Server Version: stable-v0.8.110

---

KhushamBansal:
**Notes for Reviewers**

- This PR fixes #15221

After:

https://github.com/user-attachments/assets/f5bf62b9-b820-4b75-ab74-86e339a10d10

## Summary by CodeRabbit

* **New Features**
  * Added Redux-backed filtering to the Notification Center.
  * Displayed active status and severity filters as selectable chips.
  * Restored the default unread filter when all filters are cleared.
  * Improved handling of empty, unknown, and single-value filter options.

* **Bug Fixes**
  * Improved filter updates, clearing, and value normalization.

* **Tests**
  * Added coverage for filter-to-chip conversion and filter state restoration.


ui/components/layout/NotificationCenter/filter.tsx
```diff
@@ -40,17 +41,21 @@ const useFilterSchema = () => {
...
       handleFilter={handleFilter}
       filterSchema={filterSchema}
-      defaultFilters={[
-        {
-          type: 'STATUS',
-          value: 'unread',
-          label: 'status: unread',
-        },
-      ]}
+      defaultFilters={selectedFilters}
```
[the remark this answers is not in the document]
KhushamBansal: Fixed!

Bharath314 (commented):
Previously, the notification center's default filter was unread messages. Now there is no default filter at all. Is this an intended change?

Also isn't the implementation a bit bug-prone and unoptimized right now?

TypingFilter manipulates selectedFilter and syncs changes with redux. Filter uses redux to set defaultFilters in TypingFilter, which use listens for changes on defaultFilter using useEffect to change selectedFilter, causing a double rerender.

Don't you think it's just better to remove the `defaultFilters` prop and just control all state of TypingFilter internally? I'm also skeptical that the Filter Component is even necessary because all it is doing is wrap TypingFilter while adding minimal functionality and isn't reused anywhere.

KhushamBansal:
@Bharath314 Thanks for reviewing. Yes, I agree. My intention was to fix #15221 in this PR.
Yes, the current approach isn't ideal. Let me refactor this to simplify the implementation.

leecalcote:
@KhushamBansal, still draft?

marblom007:
I see Bharath314 already raised "now there is no default filter at all - is this intended?" and you agreed to refactor but that hasn't landed yet, so I traced through why, rather than repeating the open question.

It's a real regression, not just an architectural smell: `events.ts` initializes `current_view.filters` to `{ initial: true }`, not `{ status: 'unread' }` - the actual "status: unread" default lives in `view_to_fetch_on_open.filters` and only gets copied into `current_view.filters` after `loadEvents`'s async fetch resolves on mount. Since the new chip-derivation logic in `filter.tsx` maps `Object.entries(currentFilters)` through `key.toUpperCase()` against `filterSchema` and drops any key without a schema match, the `initial` key produces an empty `selectedFilters` array - so on every Notification Center open, chips are empty until the fetch resolves, whereas before they always rendered "status: unread" immediately.

The new `filter.test.tsx` doesn't catch this because its `beforeEach` seeds `sliceState.filters = { status: 'unread' }` directly, never exercising the real initial Redux shape (`{ initial: true }`).

Suggested fix: fall back to `view_to_fetch_on_open.filters` (or an explicit `{ status: 'unread' }`) whenever `currentFilters.initial` is true, and add a test case that seeds the actual initial slice state rather than a post-fetch one - that would have caught this before it reached review.

marblom007 (changes requested):
Requesting changes for a real regression traced above, not the open architectural question Bharath314 already raised (not repeating that).

`current_view.filters` initializes to `{ initial: true }`, and the new chip-derivation logic drops any key that doesn't match `filterSchema` - so on every Notification Center open, filter chips render empty until the async `loadEvents` fetch resolves, whereas before this PR they always showed "status: unread" immediately. The new `filter.test.tsx` doesn't catch this because its `beforeEach` seeds a post-fetch state (`{ status: 'unread' }`), not the real initial Redux shape.

Concrete fix: fall back to `view_to_fetch_on_open.filters` (or an explicit `{ status: 'unread' }`) whenever `currentFilters.initial` is true, and add a test case that seeds the actual initial slice state rather than a post-fetch one.

Happy to re-review as soon as that's in - the Redux-derivation approach itself is the right direction.

KhushamBansal:
**Fixed Initial State Fallback**: Added filtersToChips mapper to ensure the filter correctly defaults to status: unread on startup (when currentFilters is { initial: true }), preventing the empty chips regression.

**Fixed Double-Rerender Loop**: Replaced the fragile useEffect sync with a key={filtersKey} remount pattern on <TypingFilter /> and memoized useFilterSchema.

**Added Regression Tests**: Updated filter.test.tsx to seed the true initial Redux state ({ initial: true }) and added tests asserting that the unread fallback works before the async fetch resolves.

Ready for re-review.

ui/components/layout/NotificationCenter/filter.tsx
```diff
@@ -1,56 +1,115 @@
...
+      handleFilter={onFilterChange}
       filterSchema={filterSchema}
-      defaultFilters={[
-        {
-          type: 'STATUS',
-          value: 'unread',
-          label: 'status: unread',
-        },
-      ]}
+      defaultFilters={selectedFilters}
```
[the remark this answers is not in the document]
KhushamBansal: Fixed!

ui/components/layout/NotificationCenter/filter.tsx
```diff
@@ -1,56 +1,135 @@
...
+  const filtersKey = useMemo(() => JSON.stringify(selectedFilters), [selectedFilters]);
+
+  // Clear restores unread — the product default — not an unfiltered fetch.
+  const onFilterChange = (filters: Record<string, unknown>) => {
+    const normalized = normalizeFilterPayload(filters || {}, filterSchema);
+    if (Object.keys(normalized).length === 0) {
+      handleFilter({ status: STATUS.UNREAD });
+      return;
+    }
+    handleFilter(normalized);
```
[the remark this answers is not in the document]
KhushamBansal: fixed!

Bharath314:
Hey @KhushamBansal, I'm unable to clear the unread filter now. Can you please look into it?

https://github.com/user-attachments/assets/93af6744-7e15-4f7d-9211-ba57f5104fe4

KhushamBansal:
> Hey @KhushamBansal, I'm unable to clear the unread filter now. Can you please look into it?
> 
>  2026-08-04.14-52-55.mp4

@Bharath314 Thanks for reviewing. I have fixed that now.

ui/components/layout/NotificationCenter/filter.test.tsx
```diff
@@ -1,70 +1,159 @@
...
+let typingFilterMountCount = 0;
+
+vi.mock('react-redux', () => ({
+  useSelector: (sel: any) =>
+    sel({
+      events: {
+        current_view: { filters: sliceState.filters },
+      },
+    }),
+}));
```
Bharath314: The Redux docs [advise against mocking redux](https://redux.js.org/usage/writing-tests#:~:text=Do%20not%20try%20to%20mock%20selector%20functions%20or%20the%20React%2DRedux%20hooks%21%20Mocking%20imports%20from%20libraries%20is%20fragile%2C%20and%20doesn%27t%20give%20you%20confidence%20that%20your%20actual%20app%20code%20is%20working).

ui/components/layout/NotificationCenter/filter.tsx
```diff
@@ -1,56 +1,143 @@
...
+  };
+
   return (
     <TypingFilter
-      handleFilter={handleFilter}
+      key={filtersKey}
+      handleFilter={onFilterChange}
```
Bharath314, line 139: React devtools tells me that this `TypingFilter` is being rendered multiple times because of this prop changing.
Image

`Filter` itself is rendering multiple times as well because of its own `handleFilter` prop

Image

Can you investigate this and determine whether this can be fixed in this PR or if it is out of scope?

KhushamBansal: This is fixed now.

ui/components/layout/NotificationCenter/filter.test.tsx
KhushamBansal: Thanks, I have updated the implementation.

ui/components/layout/NotificationCenter/filter.tsx
```diff
@@ -1,59 +1,160 @@
...
+        handleFilter={onFilterChange}
+        filterSchema={filterSchema}
+        defaultFilters={selectedFilters}
+        placeholder="Filter Notifications"
+      />
+    );
+  },
+);
+
+Filter.displayName = 'Filter';
```
Bharath314:
```suggestion
```

ui/components/layout/NotificationCenter/filter.test.tsx
Bharath314, line 1: While it is a bad idea to mock internal components, this pattern already seems to exist all over the codebase. A clean-up drive is needed to remove unnecessary mocking in all the tests sometime. For now, this can pass.

Bharath314 (approved):
Fantastic job with fixing the re-renders. There is one line I suggest removing, but otherwise LGTM!

ui/components/layout/NotificationCenter/filter.test.tsx
KhushamBansal: Acknowledged 👍

KhushamBansal:
@marblom007 Could you please re-review this PR when you get a chance? Thanks!

PARTH-TUSSLE (approved):
Hey @KhushamBansal, really nice work addressing all the feedbacks. I reviewed the changes and preview. The predefined filters correctly update the corresponding filter chips, and the chip state stays in sync when switching between filters.
The implementation and test coverage also look good to me. LGTM, approving!
Thanks @Bharath314 🦖 .

ishwar170695 (commented):
Is `resetVersion` still needed here? Since clearing now updates the filters to `{}`, `selectedFilters` should change to `[]`, which already changes `filtersKey`. Just wondering if there's a case where the extra reset counter is still required.

banana-three-join (commented):
Please look into the following points before I go ahead and approve the PR:
 
**1. Clearing severity (or any multi-select) leaves the chip bar out of sync with Redux.**
 
`normalizeFilterPayload` early-returns for anything that isn't `multiple: false`, so the empty-array `delete` only ever runs for `STATUS`. Emptying `severity` gives:
 
- `normalizeFilterPayload({ severity: [] })` → `{ severity: [] }`, key survives
- `Object.keys(normalized).length === 0` is false → not treated as a clear, no `resetVersion` bump
- Redux becomes `{ severity: [] }`
- `filtersToChips` doesn't early-return (one key), then `values = []` → no chips. So Redux reports an active filter, the chip bar renders empty, and no remount fires. The `empty-status` test shows TypingFilter does emit `{ status: [] }` on clear, so I'd expect `{ severity: [] }` to behave the same; worth confirming in the UI.
 
Fix is small — hoist the empty-array delete out of the `multiple: false` guard so it runs for every dimension, plus a test seeding `{ severity: [] }`.
 
**2. The real `TypingFilter` still isn't exercised by any test.**
 
CodeRabbit flagged this on the first pass and I don't think it got picked up. `filter.test.tsx` mocks `TypingFilter` throughout, so `expect(typingFilterMountCount).toBe(2)` confirms the *mock* remounts —
it can't confirm the real component re-seeds chips from the new `defaultFilters`, which is the mechanism the whole fix depends on. One test in `typing-filter`'s own suite that renders the real component, re-renders with changed defaults, and asserts the visible chips would close it. (Separate from the general mocking discussion — this is one specific untested path.)
 
**3. `key` changes on every filter mutation, not just clears.**
 
`filtersKey` includes `selectedFilters`, so the key changes whenever any chip is added or removed. The comment and the tests both read as remount-on-clear, but it's remount-on-every-change — meaning `TypingFilter` is destroyed and rebuilt on the primary interaction path, taking in-progress typed text and focus with it.
 
**4. `filtersToChips` and `normalizeFilterPayload` disagree on which value wins.** For a
`multiple: false` dimension holding an array, the render path takes the first value
(`values.slice(0, 1)`) and the dispatch path takes the last (`value[value.length - 1]`). Masked today since normalization runs before dispatch — but `filtersToChips` defends against that state anyway, so either it's reachable and they should agree, or the `slice` is dead code.

KhushamBansal:
@ishwar170695 @banana-three-join Thanks for your review. I'll update this PR.

ishwar170695 (approved):

Issue #21094, opened 2026-08-03
[Server] Misleading ErrQueryGet remediation on connection GET failures + verify remote-provider delete of a never-issued UUID (TC-1066)


marblom007:
## Summary

Two connection-handler concerns surfaced while diagnosing the Connection
Lifecycle e2e failures:

1. **(Confirmed code defect)** The connection **list** path reports every
   `GetConnections` failure through `ErrQueryGet`, which drops the underlying
   error and attaches a misleading "Query parameter is not a part of the
   request" remediation - so provider/DB/auth failures are mislabeled.
2. **(Open question for the server / remote-provider team)** Whether deleting a
   *never-issued but well-formed* connection UUID against the **remote
   provider** cleanly resolves to "No connection with ID …" (404 / not-found) or
   instead attempts a save/de-register that can fail (`meshery-server-1051`).

> Context: the test-side of this was already cleaned up in #21093 (merged) -
> see "Test-side already handled" below. This issue tracks the remaining
> **server-side** items, which #21093 did not touch.

---

## 1. Confirmed defect - misleading `ErrQueryGet` remediation on get failures

`server/handlers/connections_handlers.go:322` (connection **list** path):

```go
connectionsPage, err := provider.GetConnections(...)
obj := "connections"
if err != nil {
    h.log.Error(ErrQueryGet(obj))              // line 323
    writeMeshkitError(w, ErrQueryGet(obj), http.StatusInternalServerError)
    return
}
```

`ErrQueryGet` (`server/handlers/error.go:383`):

```go
func ErrQueryGet(obj string) error {
    return errors.New(ErrQueryGetCode, errors.Alert,
        []string{"unable to get: ", obj},
        []string{},                                             // underlying err dropped
        []string{"Query parameter is not a part of the request"},
        []string{"Make sure to pass the query parameter in the request"})
}
```

Two problems on this path:

1. **It drops the underlying `err`** (empty cause slice), so the real failure
   (provider/DB/auth, e.g. a `401`) never surfaces - the caller only sees
   `unable to get: .connections`.
2. **It attaches a "Query parameter is not a part of the request" probable
   cause / remediation to _any_ `GetConnections` failure**, which is misleading
   for provider/DB/auth failures that have nothing to do with a query parameter.

The **count** path a few lines up (line 301) already does the right thing -
`ErrGetConnections(err)` - which wraps the underlying error. The list path
should use the same `ErrGetConnections(err)` so the true cause is reported.

[...]

PR #21095, opened 2026-08-03
[mesheryctl] Fix stale TC-1080 connection view invalid-id assertions (Invalid UUID / meshery-server-1432)

marblom007:
## Description

Fixes the two stale **TC-1080** assertions in the mesheryctl Connection
Lifecycle e2e suite.

`03-connection-view.bats` asserted `--partial "Invalid connection ID"` for
`connection view --save <invalid-id>` and `connection view <invalid-id>
--output-format json`. The server's invalid-identifier error was changed to
`ErrInvalidUUID` — short description **"Invalid UUID"**, code
**`meshery-server-1432`** (`server/models/error.go:388`) — so the old substring
no longer matches and both tests fail. The CLI is behaving correctly; the tests
were stale.

Both assertions now check for the stable **`meshery-server-1432`** error code
rather than a human-readable message, which is robust to future wording drift.

Verified against the exact CLI output captured in the failing CI run
(`mesheryctl-e2e`, job `91602906279`):

```
Error: Response Status Code 400. Server emitted an error:
{"error":"Invalid UUID","code":"meshery-server-1432", …}
```

The two TC-1080 tests reach the server and receive this 400 directly (they are
not affected by the separate remote-provider auth problem), so the code-based
assertion matches the real output.

## Notes / context

This is one of three captain-approved follow-ups from a 12-failure Connection
Lifecycle e2e diagnosis:

- **This PR** — the stale TC-1080 assertions (test-only fix).
- Filed separately: #21094 — a possible server product bug (TC-1066: deleting a
  non-existent connection UUID attempts a provider save/de-register instead of
  returning "No connection with ID …"), plus the misleading `ErrQueryGet`
  remediation on `GetConnections` failures.
- The remaining failures are an expired remote-provider token in CI (being
  handled out of band).

## Type of change

- Bug fix (test correctness; no production code change)

## Test coverage

- `mesheryctl/tests/e2e/007-connection/03-connection-view.bats` parses cleanly
  (`bats --count` = 12). The change is a substring swap in two
  `assert_output --partial` lines; the new substring is taken verbatim from the
  captured server response.

Signed-off-by: marblom007 <158522975+marblom007@users.noreply.github.com>

## Summary by CodeRabbit

* **Tests**
  * Updated connection validation tests to verify the stable `meshery-server-1432` error code for invalid connection IDs.

### DIFF
--- a/mesheryctl/tests/e2e/007-connection/03-connection-view.bats
+++ b/mesheryctl/tests/e2e/007-connection/03-connection-view.bats
@@ -57,7 +57,10 @@ require_connection_id() {
     run $MESHERYCTL_BIN connection view --save "$NONEXISTENT_ID"
     assert_failure
     assert_output --partial "Error"
-    assert_output --partial "Invalid connection ID"
+    # Server rejects a non-parseable/nil connection id with ErrInvalidUUID
+    # ("Invalid UUID", meshery-server-1432). Assert on the stable error code
+    # rather than the human-readable message, which has drifted before.
+    assert_output --partial "meshery-server-1432"
 }
 
 @test "[TC-1080][cut=Kubernetes Connection][tg=Connection Lifecycle] given no argument is provided when running mesheryctl connection view connection-id --output-format then a message error is displayed" {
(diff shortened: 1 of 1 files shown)


marblom007:
Closing as superseded. #21093 ("fix connection tests", merged ~06:18 UTC) already landed the same TC-1080 fix in `03-connection-view.bats` - it replaces the stale `"Invalid connection ID"` assertions with `"Invalid UUID"` + `"invalid connection ID"` (and also fixes the adjacent `--output-format foo` assertions), which is a strict superset of the two assertions this PR touched. I branched from a master that predated that merge, so this is now redundant and would conflict. No action needed here; the stale assertions are already fixed on master.

PR #21105, opened 2026-08-03
[UI] Fix design import: correct fileName wire field and prevent duplicate submissions

NSTKrishna:
## Description

This PR fixes two related bugs in the design import flow (`POST /api/pattern/import`):

1. **400 Bad Request on every file upload** — the request body sent `file_name` (snake_case) instead of `fileName` (camelCase), so the server's oneOf decoder never matched the File variant.
2. **Duplicate imports on rapid clicks** — clicking "Import" multiple times while a request was in-flight fired one request per click, creating duplicate designs.

---

### Root Cause

The server's `resolveImportVariant` decodes the body against `MesheryPatternImportRequestBody`. The File variant is accepted only when both `file` **and** `fileName` are present. The Go struct generated from the OpenAPI schema uses the camelCase wire name:

```go
// github.com/meshery/schemas/.../design.go
type MesheryPatternImportFilePayload struct {
    File     []byte  `json:"file"`
    FileName string  `json:"fileName"`   // ← camelCase per naming conventions
    Name     *string `json:"name,omitempty"`
}
```

The UI was serialising the key as `file_name` (snake_case), so `FileName` decoded as `""`, `hasFile` was `false`, and the handler fell into the `!hasFile && !hasURL` branch → 400 every time.

---

### Root Cause

`handleImportDesign` is `async` and fires `importPattern` (an RTK mutation) with no in-flight guard. Rapid clicks each completed validation and each fired a separate POST, creating duplicate design records.

### Fix

The guard lives in `ImportDesignModal` — the single component both call-sites go through — so no per-caller changes are needed. Both `handleImportDesign` implementations now `return` the `importPattern` Promise chain so the modal can `await` it accurately

### Before

https://github.com/user-attachments/assets/a1d874b6-ded8-4f0c-acd5-0fd57e8f05cc

### After

https://github.com/user-attachments/assets/4acf4252-bb9e-4518-a7f8-d583589ce6eb

**Notes for Reviewers**

- This PR fixes #

## Summary by CodeRabbit

* **New Features**
  * Design imports now display an “Importing…” state while processing.
  * Import actions and dialog closing are temporarily disabled to prevent duplicate submissions.
  * Import workflows support asynchronous completion handling and reset correctly afterward.

* **Bug Fixes**
  * Improved import request compatibility by using the correct file name field.
  * Import completion and error handling remain synchronized across design and workspace flows.

### DIFF
--- a/ui/components/designs/ImportDesignModal.tsx
+++ b/ui/components/designs/ImportDesignModal.tsx
@@ -52,7 +52,7 @@ const ImportDesignModalComponent: FC<ImportDesignModalProps> = ({
   return (
     <FormModal
       isOpen
-      onClose={handleClose}
+      onClose={isSubmitting ? () => {} : handleClose}
       title="Import Design"
       headerIcon={<DesignModalHeaderIcon />}
       size="sm"
--- a/ui/components/designs/ImportDesignModal.tsx
+++ b/ui/components/designs/ImportDesignModal.tsx
@@ -11,7 +11,7 @@
  * handles the RJSF ref + submit plumbing, so callers only need to provide
  * a close handler and a submit callback receiving the form data.
  */
-import { FC, memo, useCallback, useState } from 'react';
+import { FC, memo, useCallback, useRef, useState } from 'react';
 import { importDesignSchema, importDesignUiSchema } from '@sistent/sistent';
 import { FormModal } from '@/components/shared/Modal';
 import { DesignModalHeaderIcon } from './design-modal-header';
@@ -32,18 +32,21 @@ const ImportDesignModalComponent: FC<ImportDesignModalProps> = ({
   handleImportDesign,
 }) => {
   const [isSubmitting, setIsSubmitting] = useState(false);
+  const isSubmittingRef = useRef(false);
 
   const handleSubmit = useCallback(
     async (formData: unknown) => {
-      if (isSubmitting) return;
+      if (isSubmittingRef.current) return;
+      isSubmittingRef.current = true;
       setIsSubmitting(true);
       try {
         await handleImportDesign(formData);
       } finally {
+        isSubmittingRef.current = false;
         setIsSubmitting(false);
       }
     },
-    [isSubmitting, handleImportDesign],
+    [handleImportDesign],
   );
 
   return (
--- a/ui/components/designs/ImportDesignModal.tsx
+++ b/ui/components/designs/ImportDesignModal.tsx
@@ -11,7 +11,7 @@
  * handles the RJSF ref + submit plumbing, so callers only need to provide
  * a close handler and a submit callback receiving the form data.
  */
-import { FC, memo } from 'react';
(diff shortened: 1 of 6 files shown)


marblom007:
Good fix and good follow-through on review feedback. The `fileName` correction is right - the Go payload struct already declares `json:"fileName"` (camelCase, correct per this repo's wire-naming convention), so the bug was purely the UI serializing `file_name` client-side; this wasn't a schemas violation, just a client-side typo.

I also verified the duplicate-submission guard is a real fix, not cosmetic: `isSubmittingRef` is a ref used as the synchronous guard (not just `useState`, which only drives UI), and both the Cancel button and modal-close route through the same guard via `FormModal`. Since `importPattern(...)` is now correctly `return`ed and the chain ends in a non-rethrowing `.catch()`, the promise always resolves, so `isSubmitting` reliably clears in `finally` on both success and failure paths.

I see CodeRabbit/Copilot already caught the ref-vs-state issue, the Cancel-button gap, and the stale `file_name` test assertions in earlier rounds - all three look resolved in the current diff, so I won't re-raise them. Nice iteration.

No open flaws from my read. One tiny, non-blocking note: the PR checklist's "commits are signed off" box is unchecked but `git log` shows all three commits do carry `Signed-off-by` - just a checklist oversight, nothing to fix.

NSTKrishna:
> Good fix and good follow-through on review feedback. The `fileName` correction is right - the Go payload struct already declares `json:"fileName"` (camelCase, correct per this repo's wire-naming convention), so the bug was purely the UI serializing `file_name` client-side; this wasn't a schemas violation, just a client-side typo.
> 
> I also verified the duplicate-submission guard is a real fix, not cosmetic: `isSubmittingRef` is a ref used as the synchronous guard (not just `useState`, which only drives UI), and both the Cancel button and modal-close route through the same guard via `FormModal`. Since `importPattern(...)` is now correctly `return`ed and the chain ends in a non-rethrowing `.catch()`, the promise always resolves, so `isSubmitting` reliably clears in `finally` on both success and failure paths.
> 
> I see CodeRabbit/Copilot already caught the ref-vs-state issue, the Cancel-button gap, and the stale `file_name` test assertions in earlier rounds - all three look resolved in the current diff, so I won't re-raise them. Nice iteration.
> 
> No open flaws from my read. One tiny, non-blocking note: the PR checklist's "commits are signed off" box is unchecked but `git log` shows all three commits do carry `Signed-off-by` - just a checklist oversight, nothing to fix.

Checked the box now, thanks for catching that, yeah i missed it

marblom007 (approved):
Approving on the exact head SHA (234513e) after re-verifying checks (e2e, ui-unit-tests, eslint, DCO) are green against this specific commit. Verified fix correctness in my review comment above - real fileName wire fix plus a real ref-based duplicate-submit guard, and prior bot findings are already resolved in this diff.

PR #21107, opened 2026-08-03
[Server] Fix more response body leaks in remote auth and local provid…

### LINKED ISSUE #21106: [Server] Response body leaks in remote auth and local provider HTTP methods
### Current Behavior
Four methods in the server codebase make HTTP requests but fail to properly close the response body on all paths, leading to resource leaks (open connections/file descriptors).

1. `server/models/remote_auth.go`: `revokeToken()` - Makes an `http.Post` request but completely omits closing `r.Body`.
2. `server/models/remote_auth.go`: `introspectToken()` - Makes an `http.Post` request but completely omits closing `r.Body`.
3. `server/models/default_local_provider.go`: `genericHTTPPatternFile()` - Evaluates `if resp.StatusCode != http.StatusOK` and returns an error *before* the `defer SafeClose(resp.Body, log)` statement is reached, leaking the body whenever a file is not found (status != 200).
4. `server/models/default_local_provider.go`: `genericHTTPFilterFile()` - Same defect as above; the `defer` is placed after the status code check.

### Expected Behavior
All `http.Get` and `http.Post` response bodies should be reliably closed on all execution paths to prevent connection exhaustion. This is typically done by adding `defer SafeClose(resp.Body, l.Log)` immediately after checking the request error (`if err != nil`).

### Environment

- **Host OS:** Windows

KushagraB424:
This fixes additional response body leaks where `defer SafeClose(resp.Body, log)` was either missing entirely or called after checking the response status code and returning an error.
- `revokeToken` and `introspectToken` in `server/models/remote_auth.go`
- `genericHTTPPatternFile` and `genericHTTPFilterFile` in `server/models/default_local_provider.go`

**Notes for Reviewers**

- This PR fixes #21106 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of HTTP response resources during provider and authentication operations.
  * Ensured response bodies are closed reliably, including when requests return errors.

### DIFF
--- a/server/models/default_local_provider.go
+++ b/server/models/default_local_provider.go
@@ -1972,11 +1972,11 @@ func genericHTTPPatternFile(fileURL string, log logger.Handler) ([]MesheryPatter
 	if err != nil {
 		return nil, err
 	}
+	defer SafeClose(resp.Body, log)
+
 	if resp.StatusCode != http.StatusOK {
 		return nil, fmt.Errorf("file not found")
 	}
-
-	defer SafeClose(resp.Body, log)
 
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
@@ -2009,11 +2009,11 @@ func genericHTTPFilterFile(fileURL string, log logger.Handler) ([]MesheryFilter,
 	if err != nil {
 		return nil, err
 	}
+	defer SafeClose(resp.Body, log)
+
 	if resp.StatusCode != http.StatusOK {
 		return nil, fmt.Errorf("file not found")
 	}
-
-	defer SafeClose(resp.Body, log)
 
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
--- a/server/models/remote_auth.go
+++ b/server/models/remote_auth.go
@@ -383,6 +383,7 @@ func (l *RemoteProvider) revokeToken(tokenString string) error {
 		l.Log.Error(err)
 		return err
 	}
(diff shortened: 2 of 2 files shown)


marblom007:
Solid fix. I checked both files: in `default_local_provider.go` (`genericHTTPPatternFile`, `genericHTTPFilterFile`) and `remote_auth.go` (`revokeToken`, `introspectToken`), the `defer` on the response body close is now placed before the status-code check, which brings these two functions in line with the pattern already used elsewhere in the same files (e.g. `refreshToken`, `UpdateJWKs`) rather than introducing a new convention. `resp.Body` is guaranteed non-nil here whenever `err == nil`, so no nil-guard is needed.

I also checked for sibling leak sites in both files - `doRequestHelper` returns `resp` without closing it, but that's correct since closing is the caller's responsibility there, not a leak.

One minor, non-blocking ask: this is exactly the kind of fix that's easy to regress silently later. A small `httptest`-backed test asserting the body gets closed even on a non-200 status (for one of the two functions) would make it stick. Not blocking this PR on it, just worth an issue/follow-up if you have the bandwidth.

marblom007 (approved):
Approving on the exact head SHA (3a96df9) after re-verifying all checks (server unit tests, integration tests, golangci-lint, DCO) are green against this specific commit, not an aggregated view. Matches the fix pattern already used elsewhere in these files - see my review comment above.

PR #21118, opened 2026-08-04
[mesheryctl] Propagate registry update errors

### LINKED ISSUE #21045: [mesheryctl] registry update exits successfully after a component CSV setup failure
### Current Behavior

`mesheryctl registry update` can log an error from its component-update workflow but still exit successfully.

In `mesheryctl/internal/cli/root/registry/update.go`:

- `updateCmd.RunE` logs an error returned by `InvokeCompUpdate()` and returns `nil`.
- `InvokeCompUpdate()` logs a wrapped component-CSV reading error and returns `nil`.

This makes automation unable to detect a failed registry update from the command exit status.

This finding is based on source review. It specifically concerns errors that reach `InvokeCompUpdate`; it does not claim to fix malformed CSV-row validation inside MeshKit.

### Desired Behavior

When component CSV setup or reading fails, `mesheryctl registry update` should:

1. Log the structured MeshKit error.
2. Return that error from the command.
3. Exit with a non-zero status.

### Proposed Fix

Return `err` instead of `nil` in both error branches:

- `updateCmd.RunE`
- `InvokeCompUpdate`

### Test Plan

Add a regression test that uses a nonexistent component CSV path, calls `InvokeCompUpdate()`, and verifies that it returns a non-nil error.

### Screenshots/Logs

Not applicable — this was identified through code inspection.

### Environment

- Host OS: macOS
- Meshery Server Version: Not applicable
- Meshery Client Version: Current `master`
- Platform: Local CLI execution

ChaitanyaSai-Meka:
`mesheryctl registry update` could log a component-CSV setup/parsing error and still exit successfully (exit 0), hiding failures from automation and CI.
 
In `mesheryctl/internal/cli/root/registry/update.go`:
* `updateCmd.RunE` logged the error returned by `InvokeCompUpdate()` but returned `nil`.
* `InvokeCompUpdate()` logged a wrapped component-CSV reading error but also returned `nil`.
This PR fixes both branches to return the error instead of swallowing it, so a failed registry update now surfaces as a non-zero exit status.
 
Fixes #21045
 
## Changes
 
* `updateCmd.RunE`: return the error from `InvokeCompUpdate()` instead of `nil`.
* `InvokeCompUpdate()`: return the wrapped `ErrUpdateRegistry` instead of `nil` when component-CSV setup/parsing fails.
## Test Plan
 
Added a regression test that:
1. Sets `componentCSVFilePath` to a nonexistent path.
2. Calls `InvokeCompUpdate()`.
3. Asserts a non-nil error is returned (`ErrUpdateRegistry` or equivalent).
Verified with:
`go test ./mesheryctl/internal/cli/root/registry`
 
All tests pass locally, no regressions.
 
**Notes for Reviewers**
 
* This PR fixes #21045

## Summary by CodeRabbit

* **Bug Fixes**
  * Registry update failures are now correctly reported instead of appearing successful.
  * Improved error handling when the component data file is missing or cannot be read.

* **Tests**
  * Added regression coverage to verify appropriate error reporting for failed registry updates.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved registry update error reporting when component data cannot be loaded or parsed.
  * Update failures now surface correctly instead of appearing successful.
  * Errors include the affected component data location and a consistent registry update error code.


ayanasarkar (commented):
solid fix
good catch on the swallowed error test actually covers the real failure path too, nice.
Only thing — does anything downstream rely on registry update always exiting 0 (like a CI script that ignores the exit code)? 
Might be worth a quick grep before merge just in case.
Otherwise LGTM 👍

ChaitanyaSai-Meka:
Hey @ayanasarkar,

Ran a full repo grep for bth th CLI invocation nd the underlying InvokeCompUpdate(), covering Go code, workflows, Makefiles, shell scripts, docs, and the BATS e2e tests. Only downstream consumer is model-generator.yml, which already has an if: failure() step expecting a non-zero exit on failure nd th existing e2e tests already assert failure on invalid input. So ths fix just makes CSV parsing failures behave consistently wth tht existing pattern. Nothing relies on the old silent-exit-0 behavior, so it's a strict improvement, not a breaking change.

ayanasarkar:
> Hey @ayanasarkar,
> 
> Ran a full repo grep for bth th CLI invocation nd the underlying InvokeCompUpdate(), covering Go code, workflows, Makefiles, shell scripts, docs, and the BATS e2e tests. Only downstream consumer is model-generator.yml, which already has an if: failure() step expecting a non-zero exit on failure nd th existing e2e tests already assert failure on invalid input. So ths fix just makes CSV parsing failures behave consistently wth tht existing pattern. Nothing relies on the old silent-exit-0 behavior, so it's a strict improvement, not a breaking change.

Nice, that's a thorough check. Good to know it's a strict improvement with no breaking consumers. 
lgtm👍

ChaitanyaSai-Meka:
Thanks! Since approvals are required to merge here, could you hit the Approve button when you get a chance?

ayanasarkar:
> Thanks! Since approvals are required to merge here, could you hit the Approve button when you get a chance?

Ah, I don't think I actually have write access on this repo so my approve won't satisfy that requirement unfortunately.
A maintainer with write access should be by soon to approve ! No worries :)

ChaitanyaSai-Meka:
Hi @leecalcote  @rishiraj38 , could you review this PR whenever you get a chance? Thanks!

mesheryctl/internal/cli/root/registry/update.go
```diff
@@ -132,14 +132,16 @@ func InvokeCompUpdate() error {
 
 	componentCSVHelper, err := meshkitRegistryUtils.NewComponentCSVHelper(url, "Components", sheetGID, componentCSVFilePath)
 	if err != nil {
+		err = ErrUpdateRegistry(err, modelLocation)
+		utils.Log.Error(err)
```
lekaf974:
```suggestion
```

mesheryctl/internal/cli/root/registry/update.go
```diff
@@ -132,14 +132,16 @@ func InvokeCompUpdate() error {
...
 		return err
 	}
 
 	err = componentCSVHelper.ParseComponentsSheet(modelName)
 	if err != nil {
 		err = ErrUpdateRegistry(err, modelLocation)
 		utils.Log.Error(err)
```
lekaf974:
```suggestion
```

mesheryctl/internal/cli/root/registry/update.go
```diff
@@ -96,7 +96,7 @@ mesheryctl registry update --spreadsheet-id 1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdw
 		err = InvokeCompUpdate()
 		if err != nil {
 			utils.Log.Error(err)
```
lekaf974:
```suggestion
```

mesheryctl/internal/cli/root/registry/update_test.go
```diff
@@ -0,0 +1,49 @@
...
+	logDirPath = t.TempDir()
+	modelLocation = t.TempDir()
+	modelName = ""
+	sheetGID = 1
+
+	err = InvokeCompUpdate()
+	if err == nil {
+		t.Fatal("expected an error, got nil")
+	}
+	if meshkiterrors.GetCode(err) != ErrUpdateRegistryCode {
```
lekaf974: use existing assertion method https://github.com/meshery/meshery/blob/98f3143d6018202497cba832878d53512d0771e6/mesheryctl/pkg/utils/testing.go#L475

lekaf974 (changes requested):
@ChaitanyaSai-Meka thanks, I added some comments

ChaitanyaSai-Meka:
@lekaf974  I have made the requested changes.

ChaitanyaSai-Meka:
@lekaf974  AssertMeshkitErrorsEqual doesn't fit here because the long description embeds temp directory paths, causing a mismatch. Used a nil check + error code check instead.

mesheryctl/internal/cli/root/registry/update.go
```diff
@@ -132,14 +131,14 @@ func InvokeCompUpdate() error {
 
 	componentCSVHelper, err := meshkitRegistryUtils.NewComponentCSVHelper(url, "Components", sheetGID, componentCSVFilePath)
 	if err != nil {
+		err = ErrUpdateRegistry(err, modelLocation)
 		return err
```
lekaf974:
```suggestion
return ErrUpdateRegistry(err, modelLocation)
```
since it is not use after

mesheryctl/internal/cli/root/registry/update.go
```diff
@@ -132,14 +131,14 @@ func InvokeCompUpdate() error {
...
 	}
 
 	err = componentCSVHelper.ParseComponentsSheet(modelName)
 	if err != nil {
 		err = ErrUpdateRegistry(err, modelLocation)
-		utils.Log.Error(err)
-		return nil
+		return err
```
lekaf974:
```suggestion
        return ErrUpdateRegistry(err, modelLocation)
```
same

mesheryctl/internal/cli/root/registry/update_test.go
lekaf974: please check meshkit error like it is done everywhere https://github.com/meshery/meshery/blob/03f50faab81786352545b4ebfe17b1dc2ca5257b/mesheryctl/internal/cli/root/model/import_test.go#L65

lekaf974 (changes requested):
Please check latest comments

lekaf974:
> @lekaf974 AssertMeshkitErrorsEqual doesn't fit here because the long description embeds temp directory paths, causing a mismatch. Used a nil check + error code check instead.

I did not see this comment before doing the latest review but this should not be the case, I'll re-read the changes made

mesheryctl/internal/cli/root/registry/update.go
```diff
@@ -132,14 +131,14 @@ func InvokeCompUpdate() error {
 
 	componentCSVHelper, err := meshkitRegistryUtils.NewComponentCSVHelper(url, "Components", sheetGID, componentCSVFilePath)
 	if err != nil {
+		err = ErrUpdateRegistry(err, modelLocation)
```
lekaf974:
```suggestion
```

meshkitRegistryUtils.NewComponentCSVHelper returns proper meshkit error

mesheryctl/internal/cli/root/registry/update.go
```diff
@@ -132,14 +131,14 @@ func InvokeCompUpdate() error {
...
 		return err
 	}
 
 	err = componentCSVHelper.ParseComponentsSheet(modelName)
 	if err != nil {
 		err = ErrUpdateRegistry(err, modelLocation)
```
lekaf974:
```suggestion
```
componentCSVHelper.ParseComponentsSheet returns proper meshkit error

PR #21129, opened 2026-08-04
[Server] fix: use requested pageSize instead of len(fetched)

### LINKED ISSUE #21128: [bug] Local Provider: pageSize in paginated responses reflects len(fetched) instead of requested page size, breaking hasMore / infinite-scroll logic
### Current Behavior
In the **Local Provider** (built-in), the `pageSize` field in paginated list responses is populated with `len(fetched)` — the actual number of records returned for the current page — instead of the **requested** page size. 

Because of this, the `pageSize` field is incorrect on the last page of results. The UI uses this field directly to decide whether more pages exist (e.g. `hasMore = totalCount > pageSize * (page + 1)`). When `pageSize` shrinks to the number of items on the last page, the math evaluates to `true` incorrectly, causing the UI to continuously try fetching more pages that do not exist, leading to infinite empty network requests or a broken "Load More" state.

**Affected Files/Fields:**
- `server/models/workspace_persister.go` : `WorkspacePage.PageSize`
- `server/models/workspace_persister.go` : `MesheryDesignPage.PageSize`
- `server/models/workspace_persister.go` : `MesheryViewPage.PageSize`
- `server/models/workspace_persister.go` : `TeamPage.PageSize`
- `server/models/environment_persister.go` : `EnvironmentPage.PageSize`

### Expected Behavior
`pageSize` in the response should always reflect the **requested** page size (i.e., the value parsed from the `pagesize` query parameter), regardless of how many records are actually returned on that specific page. This matches standard REST pagination conventions and the behavior of the Remote Provider (meshery-cloud).

### Screenshots/Logs
If you curl the API directly when you have more than 10 items total:
```bash
# Assuming you have 13 designs in a workspace, fetching the second page (page 1):
curl -s "http://localhost:9081/api/workspaces/<workspace-id>/designs?page=1&pagesize=10" | jq '.pageSize'
```
**Current Output:** `3`
**Expected Output:** `10`

### Environment

- **Host OS:** Windows
- **Platform:** Docker
- **Meshery Server Version:** latest master
- **Meshery Client Version:** latest master

### To Reproduce
1. Run Meshery with the **Local Provider** (default).
2. Create **more than 10** designs, views, or workspaces (e.g. 13 designs).
3. Open the Spaces Switcher or the Workspaces page in the UI.
4. Scroll to the bottom to trigger the infinite scroll.
5. Observe in the browser network tab that it continuously fires empty requests for pages that don't exist because `hasMore` calculates incorrectly on the last page.

---

anushkagupta200615-jpg:
Fixes incorrect `pageSize` tracking in paginated responses which broke the UI `hasMore` calculations on the last page, causing infinite empty network requests.

Fixes #21128.

**Notes for Reviewers**

- This PR fixes the pagination bug across all 5 affected local provider endpoints (Workspaces, Designs, Views, Teams, and Environments). 
- Previously, the `PageSize` field in the JSON response was incorrectly set to `len(fetched)`. On the final page of results, this caused the UI's `hasMore` calculation (`totalCount > pageSize * (page + 1)`) to evaluate to true instead of false.
- We have updated the logic to always return the *requested* `pageSize` back to the client, or fallback to `totalCount` if `pageSize="all"`.
- We also added a comprehensive new unit test suite (`server/models/pagination_pagesize_test.go`) to ensure this pagination math is strictly validated going forward.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected pagination responses so page size consistently reflects the requested value.
  - “All” requests now report the total number of available records as the page size.
  - Improved “has more” indicators across first, middle, last, and exact-fit pages.
  - Updated environment, workspace, design, view, team, and connection listings for consistent pagination behavior.

- **Tests**
  - Added coverage for pagination edge cases and verified page-size data remains intact when responses are converted to and from JSON.


server/models/environment_persister.go
ishwar170695: Should this also use `resolvedPageSize` for consistency with the other paginated endpoints updated in this PR?

anushkagupta200615-jpg: Great catch, thanks for spotting this! You are absolutely right—I missed this 7th endpoint in the initial sweep. I have just pushed a new commit that updates GetEnvironmentConnections to use resolvedPageSize so it is now completely consistent with the others

ishwar170695 (approved):

server/models/workspace_persister.go
```diff
@@ -81,19 +81,22 @@ func (wp *WorkspacePersister) GetWorkspaces(orgID, search, order, page, pageSize
 	query.Table("workspaces").Count(&count)
 
 	workspacesFetched := []workspace.Workspace{}
-	pageUint, err := strconv.ParseUint(page, 10, 32)
+	pageUint, err := strconv.ParseUint(page, 10, strconv.IntSize)
 	if err != nil {
 		return nil, err
 	}
```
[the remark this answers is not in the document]
Maanvi212006: Address this comment as `ParseUint` is still allowing values up to 2^64-1, which then overflow to negative when cast to int. So, I believe it must be accepted. Also leave a comment below all the bot's reviews justufying why you accepted/rejected that review

Maanvi212006 (changes requested):
Test is added, but it only checks the `hasMore` formula on a hand-built `pageResponse` struct — it never calls `GetWorkspaces`/`GetEnvironments`/etc. So if the actual fix in the persister files gets reverted back to `len(fetched)`, this test suite would still pass. Would be good to have at least one test that actually invokes a persister method and asserts on its real output.

Maanvi212006 (changes requested):
Thanks for adding the real persister test — this resolves the immediate concern for `GetWorkspaces`. However, the same regression risk (revert to `len(fetched)` going undetected) still exists for the other 6 fixed methods (`GetEnvironments`, `GetEnvironmentConnections`, `GetWorkspaceEnvironments`, `GetWorkspaceDesigns`, `GetWorkspaceViews`, `GetWorkspaceTeams`), which only have formula-level coverage, not real-invocation coverage. Could you extend the same pattern to all?

PR #21142, opened 2026-08-05
DO NOT MERGE: temporary proof that the E2E gate fires (closes itself)

CodeAhmedJamil:
## DO NOT MERGE - DO NOT REVIEW

Temporary, deliberately-failing PR. It exists only to prove that the E2E gate
restored in #21140 actually fails the job, and it will be closed and its branch
deleted as soon as the E2E job reports `failure`.

It contains #21140's commits plus one added file,
`ui/tests/e2e/gate-demo.spec.ts`, whose only test asserts `1 === 2`.

**Expected result:** `tests-e2e / E2E Tests` concludes **failure**, and the step
"Fail job when Playwright reports test failures" runs rather than being skipped.

Before #21140, that job concluded `success` on 19 of the last 19 runs that reached
it, 8 of them with real test failures. If this PR goes green, the gate is still
not firing and #21140 is not finished.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -8,7 +8,22 @@ on:
       - "docs/**"
       - "mesheryctl/**"
       - "*.md"
-      - ".github/**"
+      # Inert .github paths only. This used to exclude ".github/**" wholesale,
+      # which meant a PR changing this workflow - or test-e2e.yml, or the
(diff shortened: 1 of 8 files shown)


jijillery:
Purpose served: #21140 is merged, and this branch demonstrated the gate firing (run 30982664574 job=failure, gate step=failure) against its green control. Closing as planned - it was never for merge.

PR #21161, opened 2026-08-05
[mesheryctl] Fix response body leak and nil pointer dereference in au…

### LINKED ISSUE #21159: [mesheryctl] Fix panic and response body leak in pkg/utils/auth.go
### Current Behavior
While reviewing `mesheryctl/pkg/utils/auth.go`, two bugs were found related to HTTP request handling:

**1. Nil Pointer Dereference (Panic) in `UpdateAuthDetails`**
Around line 235, the `defer` statement for closing the response body is placed *before* checking the error. Because `defer` arguments are evaluated immediately, this will cause a panic if `client.Do(req)` fails (since `resp` will be `nil`):
```go
client := &http.Client{}
resp, err := client.Do(req)
defer SafeClose(resp.Body) // <-- BUG: panics immediately if err != nil

if err != nil {
	err = errors.Wrap(err, "error dispatching there request: ")
	return err
}
```

### Desired Behavior
In UpdateAuthDetails, defer SafeClose(resp.Body) should be placed after checking if err != nil.
In GetProviderInfo, defer func() { _ = resp.Body.Close() }() should be added immediately after the if err != nil check for the http.Get request.
Screenshots/Logs
N/A - Code analysis bug.

### Environment
Host OS: Windows / Mac / Linux
Meshery Server Version: stable-v
Meshery Client Version: stable-v
Platform: Docker / Kubernetes
Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
🧰 [mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
🙋🏽 [Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)
📓 [mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63slGAuCnleDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
📚 [Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
🎯 [Meshkit errors Guide](https://docs.meshery.io/project/contributing/contributing-error)
🙋🏾‍♀️ Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io/)
👩🏽‍💻 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

KushagraB424:
**Notes for Reviewers**

This PR addresses two bugs in `mesheryctl/pkg/utils/auth.go`:
1. **Nil Pointer Dereference:** Moved `defer SafeClose(resp.Body)` in `UpdateAuthDetails` to after the `err != nil` check. Previously, it would panic if the request failed because `resp` was nil.
2. **Response Body Leak:** Added `defer func() { _ = resp.Body.Close() }()` to `GetProviderInfo` to properly close the response body.

- This PR fixes #21159

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication request handling to prevent errors when responses are unavailable.
  * Ensured provider response resources are properly closed after successful requests.

### DIFF
--- a/mesheryctl/pkg/utils/auth.go
+++ b/mesheryctl/pkg/utils/auth.go
@@ -233,12 +233,12 @@ func UpdateAuthDetails(filepath string) error {
 
 	client := &http.Client{}
 	resp, err := client.Do(req)
-	defer SafeClose(resp.Body)
 
 	if err != nil {
 		err = errors.Wrap(err, "error dispatching there request: ")
 		return err
 	}
+	defer SafeClose(resp.Body)
 
 	data, err := io.ReadAll(resp.Body)
 	if err != nil {
@@ -352,6 +352,7 @@ func GetProviderInfo(mctCfg *config.MesheryCtlConfig) (map[string]Provider, erro
 	if err != nil {
 		return nil, err
 	}
+	defer func() { _ = resp.Body.Close() }()
 
 	if err := json.NewDecoder(resp.Body).Decode(&res); err != nil {
 		return nil, err


kunalKumar-13 (commented):
This has been open since 5 August without feedback, so first: both bugs you identified were real. Unfortunately the situation has shifted underneath it and I think it needs a rebase rather than a review.

**The nil dereference has already been fixed on master.** Commit `5e6d53d00`, *"fix(mesheryctl): prevent nil response panic in auth update"*, landed on **9 September**:

```
$ git blame -L 236,240 mesheryctl/pkg/utils/auth.go
5e6d53d00e (Kanika0306 2026-09-09 236)   err = errors.Wrap(err, "error dispatching the request: ")
^16adee5bd (marblom007 2026-07-19 237)   return err
^16adee5bd (marblom007 2026-07-19 238) }
5e6d53d00e (Kanika0306 2026-09-09 239) defer SafeClose(resp.Body)
```

Same fix as yours — the `defer` moved below the `err != nil` check. That change also corrected the "dispatching there request" typo, which is why your diff won't apply cleanly.

For what it's worth your diagnosis was right and 35 days early. `SafeClose(resp.Body)` evaluates its argument at the `defer` statement, not when the function returns, so a failed `client.Do` panicked immediately on `nil.Body`. That is exactly what it was.

**The other half is still needed, and nobody has fixed it.** `GetProviderInfo` on master today:

```go
// mesheryctl/pkg/utils/auth.go:349
resp, err := http.Get(mctCfg.GetBaseMesheryURL() + "/api/providers")
if err != nil {
    return nil, err
}

if err := json.NewDecoder(resp.Body).Decode(&res); err != nil {
    return nil, err
}

return res, nil
```

No `Close` on any path. And it has three call sites — `system/provider.go:126`, `system/provider.go:207`, and `auth.go:300` — so it isn't a one-shot at startup.

So I'd suggest rebasing onto master and reducing this to just the `GetProviderInfo` change. That leaves a one-line PR that is still worth merging.

One small thing if you do: the file's own convention is `SafeClose`, which logs a close error rather than discarding it —

```go
// helpers.go:428
func SafeClose(co io.Closer) {
    if cerr := co.Close(); cerr != nil {
        Log.Error(cerr)
    }
}
```

`UpdateAuthDetails` uses it a few lines above. `defer func() { _ = resp.Body.Close() }()` does close the body correctly, but `defer SafeClose(resp.Body)` matches the surrounding code and keeps the error visible.

Sorry it sat this long — the remaining half is a real leak and worth finishing.

Issue #20995, opened 2026-07-28
Grounding AI in the Meshery Registry: Semantic Retrieval and Context Assembly


miacycle:
## Description

The Meshery Registry holds thousands of components across hundreds of Models spanning Kubernetes, the CNCF landscape, and the major clouds, together with the Relationship definitions that describe how those components legitimately connect. That registry is precisely the knowledge a language model needs in order to generate a design Meshery will accept, and it is orders of magnitude too large to fit in a context window. Today there is no defined mechanism for selecting registry context for AI prompts, and even ordinary component search lacks a model filter. Generation therefore falls back on whatever the model happens to remember about Kubernetes, which is how agents invent components that do not exist and propose relationships the policy engine will reject.

This project builds the retrieval layer that grounds every other AI capability in Meshery. The mentee implements an embedding index over Models, components, and Relationships with a pluggable backend so that operators can choose a hosted embedding service or a fully local one, incremental re-indexing as models are registered or updated, and a ranked context assembly API that accepts an intent and returns a compact, token-budgeted slice of the registry for the AI Adapter and the Meshery MCP Server to consume. The same index pays off directly for humans: semantic search across the registry pages in Meshery UI, and search improvements in `mesheryctl` including the missing model filter for component search. Effectiveness is measured, not asserted, using the design validation and evaluation work running in parallel this term.

## Expected Outcomes

- An embedding index over Meshery Models, components, and Relationships, incrementally rebuilt on registry changes, with a pluggable hosted or local embedding backend.
- A context selection API that returns a ranked, token-budgeted registry slice for a given intent, consumed by both the AI Adapter and the Meshery MCP Server.
- Semantic search in the Meshery UI registry experience and in `mesheryctl`, including the currently missing `--model` filter on component search.
- Measured improvement in component resolution rate and generated design validity against a retrieval-free baseline, reported through the term's evaluation harness.
- Documentation of the retrieval architecture, index schema, backend configuration, and tuning guidance.

## Technologies

Golang, React, vector search, REST, SQL

## Required/Desirable Skills

Kubernetes CRDs

*(The required skills are not the same as the technologies listed above.)*

sooovamm:
Hi @miacycle I'd like to work on this issue. Could you please assign this to me.

pontusringblom:
> Hi [@miacycle](https://github.com/miacycle) I'd like to work on this issue. Could you please assign this to me.

[...]

PR #21180, opened 2026-08-06
feat(mesheryctl): add --model filter to component search

smishraa30:
## What this does

`component search` and `component list` both call `GET /api/registry/components`,
which never read a `model` query parameter — even though `ComponentFilter.ModelName`
and its DB-level filtering have existed since the `/models/{model}/components` route
was added. This wires that same field into the search endpoint's handler and adds a
`--model` flag to the CLI to send it.

## Why this is low-risk

The `ModelName` filter itself isn't new logic — it's the same field and query path
already exercised in production by `GetMeshmodelComponentByModel`. This change reuses
it on a second endpoint rather than introducing new filtering behavior.

## Testing

Added a CLI test asserting `--model` builds the correct query string
(`model=X&search=Y&page=...`) — passes locally (`go test ./internal/cli/root/components/... -run TestSearchComponent -v`).
`go build ./server/...` and `go vet ./server/...` are clean. I didn't have a full
local server + seeded registry running to verify the live DB query end-to-end; happy
to if a maintainer wants that before merge.

Relates to #20995 (`the currently missing --model filter on component search`).

## Summary by CodeRabbit

* **New Features**
  * Added an optional model filter to component searches using the `--model` or `-m` flag.
  * Search results can now be narrowed by model name and display relevant component details, including names, versions, and pagination information.

* **Bug Fixes**
  * Improved component search handling so model filters are correctly applied and results match the selected model.


mesheryctl/internal/cli/root/components/search_test.go
```diff
@@ -58,6 +58,14 @@ func TestSearchComponent(t *testing.T) {
 			ExpectedResponse: "components.search.output.golden",
 			ExpectError:      false,
 		},
+		{
+			Name:             "given valid name and model flag when component search then filter by model",
```
Maanvi212006, line 62: This only checks the `--model` flag builds the right URL. It reuses the shared fixture (`MODEL: component-test-0`), so it can't actually verify server-side filtering with `kubernetes`. Maybe rename to reflect that, or use a matching fixture.

Maanvi212006 (changes requested):

mesheryctl/internal/cli/root/components/search_test.go
smishraa30: Updated. Added a dedicated fixture (components.api.response.model.golden) where the component actually belongs to model kubernetes, plus a matching golden output (components.search.model.output.golden). The test now verifies the response is actually filtered by model, not just that the URL is built correctly. All 5 subtests pass locally. PTAL 🙏

smishraa30:
@miacycle — this closes the `--model` filter item named in #20995's expected outcomes. 
Maanvi's earlier feedback on the test fixture is addressed (dedicated fixture with a 
matching model, so it verifies actual filtering rather than just URL construction). 
CI is green. Could you or another maintainer take a look when you have a chance?

Maanvi212006 (approved):
LGTM!

smishraa30:
Hi @Maanvi212006 — thanks again for catching that the model-filter test was reusing a shared fixture that couldn't actually prove server-side filtering. I've fixed it to use a matching fixture instead, and kept the branch synced with master since. DCO's green — happy to make any further changes, otherwise is this good to merge?

PR #21219, opened 2026-08-07
docs(contributing): enable section-index page titles

### LINKED ISSUE #21178: [Docs]Section-index (_index.md) pages never render their own title — display_title: false suppresses it with no fallback
## Current behaviour

Every section-index page under `docs/content/en/project/contributing/`
(`cli/_index.md`, `ui/_index.md`, `models/_index.md`, and `contributing-docs/_index.md`)
sets `display_title: false` in its front matter. Per `docs/layouts/_default/baseof.html`
line 76, this suppresses the page's auto-generated `<h1>{{ .Title }}</h1>`:

```go
{{ if ne (.Params.display_title | default true) false }}
  <h1>{{ .Title }}</h1>
{{ end }}
```

The section's child-page listing comes from `docs/layouts/partials/section-index.html`,
which only prints the *child* pages' titles as links — it never renders the current
page's own title:

```go
{{ range .Sections }}
  <a href="{{ .RelPermalink }}">{{ .Title }}</a>
{{ end }}
{{ range .RegularPages }}
  <a href="{{ .RelPermalink }}">{{ .Title }}</a>
{{ end }}
```

Net effect: the page's own title never appears anywhere in the main content area —
only in the breadcrumb and the (already-highlighted) sidebar entry. Visiting any of
these section-index pages lands you directly on a bare bullet list of child pages
with no heading introducing them.

**Steps to reproduce:**
1. Visit any of `/project/contributing/contributing-cli/`,
   `/project/contributing/contributing-ui/`, `/project/contributing/contributing-models/`,
   or `/project/contributing/contributing-docs/`
2. Note there's no visible page heading above the child-page list.

## Desired behaviour

Each section-index page should give some visible confirmation of "you are here" in
the main content area — not just the breadcrumb and sidebar. The simplest fix is
removing `display_title: false` from these pages so the `<h1>` renders normally, but
it's worth confirming with a maintainer first: the setting was applied consistently
across every section in this directory, so it may have been an intentional choice
(e.g. to avoid a heading feeling redundant next to the child-page list) rather than
an oversight. If it's intentional, an alternative might be a smaller heading rendered
specifically on section-index pages instead of full removal of the suppression.

## Screenshots

Image

## Context

Surfaced during review of #20498, where restructuring
`project/contributing/contributing-docs/` into a section (matching the existing
`cli/`, `ui/`, `models/` pattern) reproduced the same missing-heading behavior on the
new `_index.md`.

cc @YASHMAHAKAL

smishraa30:
## Description

Section-index pages under `docs/content/en/project/contributing/` (`cli/`, `ui/`, `models/`, `contributing-docs/`) set `display_title: false`, which suppresses the auto-generated `<h1>` in `baseof.html`. Since these pages have no body content, their title never rendered anywhere in the main content area — only in the breadcrumb and sidebar.

Removes `display_title: false` from those four files so the existing `<h1>` renders. No template changes.

## Why the content-level fix

Per review feedback on the earlier approach, this is fixed at the content level rather than by adding a fallback to the global `list.html`.

Checked every `_index.md` on master that sets `display_title: false` — 26 files. Twenty-two supply their own heading in body content (`# Contributing`, `# mesheryctl adapter`, the homepage hero), so the suppression there is deliberate and unchanged. The only four with empty bodies are the ones in this PR. Their parent `project/contributing/_index.md` has both the flag and its own `# Contributing` heading, which is likely what got copied down without the body that justified it.

## Screenshots

Screenshot 2026-08-29 161519

Screenshot 2026-08-29 161504

https://docs.meshery.io/pr-preview/pr-21219/project/contributing/cli/

Fixes #21178

## Summary by CodeRabbit

* **Documentation**
  * Page titles are now visible by default on the CLI, documentation, models, and UI contributing guides.
  * Contributing documentation pages now present their titles consistently.


smishraa30:
Following up since it's been a few days — this is ready for review now, not just a starting point for discussion. Addressed CodeRabbit's feedback (the fallback heading now checks the same display_title condition baseof.html uses, not just empty page content), all checks are green, and it's a minimal, scoped 3-line change to list.html only.

@YASHMAHAKAL — could you take a look when you have a moment? Mainly want your read on whether display_title: false on these section-index pages was intentional (in which case this scoped heading fallback should be the right call), or if you'd prefer a different heading level or treatment. Happy to adjust based on your take.

Preview: https://docs.meshery.io/pr-preview/pr-21219/project/contributing/cli/

ishwar170695 (commented):
I think this issue is better fixed at the content level by removing `display_title: false` from the affected section `_index.md` files, rather than adding a fallback specifically for that condition in the global `list.html` template.

Those pages appear to suppress their titles because of inherited or copied frontmatter, so enabling the existing `<h1>` behavior keeps the template logic simpler and avoids introducing special-case behavior for all list pages.

smishraa30:
> better fixed at the content level by removing `display_title: false` from the affected section `_index.md` files, rather than adding a fallback spe

Agreed — that's the cleaner fix, and it was what I originally proposed in #21178. Switching this PR to remove `display_title: false` from the four section-index pages and reverting the `list.html` change. Will push shortly.

ishwar170695 (approved):
LGTM. This correctly restores the titles only for the four section-index pages where `display_title: false` had no replacement heading.

smishraa30:
> LGTM. This correctly restores the titles only for the four section-index pages where `display_title: false` had no replacement heading.

Thanks for the review, kindly merge it whenever you get free.

Katotodan (approved):

Katotodan:
LGTM

banana-three-join:
Before approving and merging, I believe there's some context missing as to why these parameters were willingly added (because these headings were present before). Could you look and see if there's any context as to why the `display_title: false` was added? @smishraa30

Bhumikagarggg:
> Before approving and merging, I believe there's some context missing as to why these parameters were willingly added (because these headings were present before). Could you look and see if there's any context as to why the `display_title: false` was added? @smishraa30

I checked the history. `display_title: false` was introduced in commit `680157ef` from PR #20333 (`Docs: restructure contributing docs`) when these section `_index.md` files were created. I don’t see discussion indicating that suppressing the titles on these empty section pages was intentional.It looks more likely that the setting was copied from `docs/content/en/project/contributing/_index.md`, where it makes sense because that page manually renders `# Contributing` in the body. These child section pages do not have body content/headings, so removing `display_title: false` seems like the correct scoped fix.

banana-three-join:
Ty! @Bhumikagarggg

banana-three-join (approved):

PR #21262, opened 2026-08-09
[Server] Preference sync never activates if a remote provider is unreachable at boot, and /api/system/sync then hangs forever

### LINKED ISSUE #21260: [Server] Preference sync never activates if a remote provider is unreachable at boot, and /api/system/sync then hangs forever
### Current Behavior

If a remote provider is unreachable when the server boots but comes back later, saving user
preferences hangs forever. `/api/system/sync` is one of those requests, so the UI never
finishes loading.

`syncChan` is only created by `SyncPreferences()`, which runs once at boot and returns early
unless the provider already advertises `SyncPrefs` (`server/models/remote_provider.go:421`).
If the boot probe failed, capabilities are empty at that moment, so the channel is never
created. Capabilities do load later, on the next `/api/providers/stream` refresh  and from
then on `RecordPreferences` sends on the still-nil `syncChan`
(`server/models/remote_provider.go:4100`), which blocks forever.

`StopSyncPreferences` already nil-checks its channel for exactly this case ("the race where
SyncPrefs became supported but SyncPreferences never ran"). `RecordPreferences` doesn't, so
it looks like the guard was added to one sender and missed on the other, wanted to check if
that's the right reading..

### Expected Behavior

A preference write either syncs, or fails with an error. It shouldn't hang. Ideally the
sync worker also starts once capabilities load, instead of only at boot.

### Environment

---

Priyanshu6968:
**Notes for Reviewers**

- This PR fixes #21260 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of remote provider preference synchronization during startup, capability updates, and shutdown.
  * Prevented preference updates from blocking indefinitely when synchronization is unavailable or busy.
  * Ensured provider URL and capability settings remain consistent during login.
  * Improved handling of concurrent synchronization activity and repeated lifecycle operations.

### DIFF
--- a/server/models/remote_provider.go
+++ b/server/models/remote_provider.go
@@ -112,6 +112,13 @@ type RemoteProvider struct {
 	// provider and token cookie expiry bound
 	CookieDuration time.Duration
 
+	// syncMu guards the preference-sync worker's lifecycle: the two channels
+	// below and the flag recording whether the worker is running. Activation
+	// is no longer a boot-time-only event (capabilities can load at any time,
+	// see SyncPreferences), so the request path and the capability-load path
+	// can race to start it.
+	syncMu       sync.Mutex
+	syncActive   bool
 	syncStopChan chan struct{}
 	syncChan     chan *userSession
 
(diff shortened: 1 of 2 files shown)


Priyanshu6968:
Good catch, fixed. `RecordPreferences` now takes a single `GetProviderProperties()`
snapshot and reads both the capability set and the name off it, so the error can't name a
provider whose capabilities came from a different revision of the struct. Also logging the
error that's actually returned now instead of `ErrOperationNotAvailable`.

Worth noting the same raw read was in code this PR added `enqueuePrefSync`'s warn line
called `l.Name()`, which is just `return l.ProviderName` with no lock. Fixed that too.

One thing I've left alone: `Name()` is unsynchronized for all its other callers, and about
a dozen other `ErrInvalidCapability` sites in this file read `l.ProviderName` directly.
Locking inside `Name()` would cover all of them in one line and I don't see a caller
holding `propertiesMu`, but it's a wider change than a hang fix should carry.

PR #21314, opened 2026-08-10
[UI] Controllers settings UX scanability

chaitanyamedidar:
**Notes for Reviewers**

- This PR fixes #

https://github.com/user-attachments/assets/c7c1c461-a36a-4606-8e6f-6e12d01a034c

## New Features
- Added visual selection for inherited, Operator, and embedded controller modes.
- Added watch-list configuration with inherited, whitelist, and blacklist options.
- Added searchable operator version selection.
- Organized MeshSync and Broker settings into expandable Operator-only sections.
- Added contextual icons, tooltips, and clearer section guidance.

## Improvements
- Improved responsive sizing, accessibility, inherited-value visibility, and deployment-mode explanations.
- Clarified server-wide defaults and per-cluster controller overrides.

## Tests
- Expanded coverage for modes, accordions, inherited settings, and watch-list configuration.


ui/tests/e2e/controllers-config.testmap.ts
ishwar170695, line 207: `meshsyncGrid` still scopes to the `MeshSync version` container, but `Output namespaces` and `Output resources` are now rendered in the separate `Discovery filters` section by this PR. This causes the current E2E test to fail with `no control rendered for meshsync.outputNamespaces`. Please update this locator/test mapping to cover the new structure, then rerun the E2E tests.

ui/components/configuration/ControllersConfigForm.tsx
```diff
@@ -1,27 +1,184 @@
...
+        {WATCH_MODE_OPTIONS.map((option) => (
+          <MenuItem key={option.value} value={option.value}>
+            {option.label}
+          </MenuItem>
+        ))}
+      </TextField>
+
+      {watchMode === 'whitelist' && (
+        <Box sx={{ marginTop: '1rem', width: '100%' }}>
+          {whitelist.map((row, index) => (
```
ishwar170695, line 106: Consider avoiding `index` as the React key here. Since whitelist rows are editable and can be removed from the middle, index keys can cause React to reuse the wrong row state. A stable row identifier would be safer.

ui/tests/e2e/controllers-config.testmap.ts
chaitanyamedidar: Updated the testmap with `meshsyncFilters` and the render-all-controls case to use it. Thanks.

ui/components/configuration/ControllersConfigForm.tsx
chaitanyamedidar: Agreed. Whitelist rows now use client-only stable keys.

banana-three-join (changes requested):
A few things I'd like addressed before merge:

### 1. `crypto.randomUUID()` will throw on non-localhost HTTP origins
 
``ui/components/configuration/ControllersConfigForm.tsx:67, 172``
 
`crypto.randomUUID` is a secure-context-only API. It's available on HTTPS and `http://localhost`, which covers `mesheryctl system start` and `kubectl port-forward`, so this won't show up in local dev or CI (jsdom provides it). But we document NodePort exposure and `http://<meshery-server>:9081` access, and on those origins `crypto.randomUUID` is `undefined` — clicking **Add resource** throws `TypeError: crypto.randomUUID is not a function` and takes the form down.
 
`uuid@^14.0.1` is already a `ui/` dependency:
 
```ts
import { v4 as uuidv4 } from 'uuid';
 
// line 67
...Array.from({ length: whitelist.length - whitelistKeysRef.current.length }, () => uuidv4()),
 
// line 172
whitelistKeysRef.current = [...whitelistKeysRef.current, uuidv4()];
```
 
### 2. Operator-only accordions start collapsed on the server-wide defaults page
 
``ui/components/configuration/ControllersConfigForm.tsx:251-269``
 
With an empty draft, `serverDefaultDeploymentMode` returns `mode: 'embedded'` (`deploymentMode.ts:213`, `:40`), so `operatorModeApplies` is `false` and both accordions initialise collapsed. But `isInertIn` (`deploymentMode.ts:231`) only returns `true` for `scope: 'connection'` — on the server-defaults layer none of those fields are inert, they're fully editable and they do reach connections that override to Operator.
 
The result is that `SERVER_DEFAULT_SCOPE_NOTE` renders directly above a collapsed section chipped "Operator mode only," which reads as if the settings don't apply there. Nothing is lost or disabled, so this isn't a correctness bug — it's a discoverability one, and it's the opposite of the scanability goal on that page.
 
Suggest gating the collapse on scope:
 
```ts
const operatorModeApplies =
  modeForDisclosure === 'operator' || deploymentMode?.scope !== 'connection';
```
 
Related, in the same block: the `useEffect` at `:266-269` re-runs on every `operatorModeApplies` transition and overwrites the accordion state unconditionally, so a user who manually expands a section, switches mode to Operator and back, loses their expansion. Worth tracking whether the user has touched the accordion and skipping the auto-set if so.

ui/components/connections/wizard/kubernetesDeploymentMode.tsx
```diff
@@ -114,13 +83,13 @@ export const MeshsyncDeploymentModePicker = ({
...
-      );
-    })}
-  </Box>
+              </>
+            }
+          />
+        </Box>
+      </CustomTooltip>
+    ))}
+  </RadioGroup>
```
Bharath314: I had a little trouble following this implementation. Is there a reason why we are not using a `Select` here instead?

ui/components/configuration/ControllersConfigFields.tsx
```diff
@@ -0,0 +1,495 @@
...
+          alignItems: 'center',
+          gap: 1.25,
+          px: 1.5,
+          py: 1.5,
+          '& .MuiFormControlLabel-label': { fontWeight: 600, fontSize: '0.875rem', minWidth: 0 },
+        }}
+      />
+    </Card>
+  );
+});
```
Bharath314, line 75: This is really generic component. I think this should be moved into /components/shared as its own file. That way, it can be reused anywhere.

ui/components/configuration/DeploymentModeNotices.tsx
```diff
@@ -17,10 +17,28 @@ import {
...
+/** Product icons already used on Connections controller chips. */
+export const CONTROLLER_SECTION_ICONS: Record<ConfigSection, string> = {
+  operator: '/static/img/integrations/meshery-operator.svg',
+  meshsync: '/static/img/extensions/meshsync.svg',
+  broker: '/static/img/integrations/nats-icon-color.svg',
+};
+
+/** Section help copy — shown via InfoTooltip next to the title, not as body text. */
+export const CONTROLLER_SECTION_PURPOSE: Record<ConfigSection, string> = {
+  operator:
```
Bharath314: I'm not sure about the naming of `CONTROLLER_SECTION_ICONS` and `CONTROLLER_SECTION_PURPOSE`. Meshery Broker doesn't _control_ anything and there is no category called controllers as far as I am aware. Can these be renamed to just `SECTION_ICONS` and `SECTION_PURPOSE` since their scope is limited to just this file?

ui/components/configuration/DeploymentModeNotices.tsx
```diff
@@ -61,24 +72,47 @@ export const DeploymentModeBanner: React.FC<{ governance?: DeploymentModeGoverna
...
+        gap: '0.75rem',
+        marginBottom: '0.75rem',
+        scrollMarginTop: '1rem',
+        flexWrap: 'wrap',
+      }}
       data-testid={`controllers-config-section-${section}`}
     >
+      <Box
+        component="img"
+        src={CONTROLLER_SECTION_ICONS[section]}
```
Bharath314, line 101: There is a TypeScript error here:

```
Type '{ component: "img"; src: string; alt: string; "aria-hidden": true; sx: { width: number; height: number; flexShrink: number; objectFit: "contain"; }; "data-testid": string; }' is not assignable to type 'IntrinsicAttributes & Omit<BoxProps, "ref"> & RefAttributes<HTMLDivElement>'.
  Property 'src' does not exist on type 'IntrinsicAttributes & Omit<BoxProps, "ref"> & RefAttributes<HTMLDivElement>'
```
This seems to be a type error in either sistent or mui. Can a ts-ignore be added here as a stopgap?


```suggestion
        // TODO: fix this type error in sistent
        // @ts-ignore
        src={CONTROLLER_SECTION_ICONS[section]}
```

ui/components/configuration/DeploymentModeNotices.tsx
```diff
@@ -17,10 +17,28 @@ import {
   SERVER_DEFAULT_SCOPE_NOTE,
   dormantPathsIn,
   isInertIn,
+  modeConsequence,
   type ConfigSection,
   type DeploymentModeGovernance,
 } from './deploymentMode';
 
+/** Product icons already used on Connections controller chips. */
```
Bharath314, line 25: Remove unnecessary context added by LLM
```suggestion
/** Product icons */
```

ui/components/configuration/DeploymentModeNotices.tsx
```diff
@@ -17,10 +17,28 @@ import {
...
+export const CONTROLLER_SECTION_ICONS: Record<ConfigSection, string> = {
+  operator: '/static/img/integrations/meshery-operator.svg',
+  meshsync: '/static/img/extensions/meshsync.svg',
+  broker: '/static/img/integrations/nats-icon-color.svg',
+};
+
+/** Section help copy — shown via InfoTooltip next to the title, not as body text. */
```
Bharath314, line 32: Remove unnecessary context added by LLM
```suggestion
/** Section information shown via InfoTooltip next to the title */
```

ui/components/configuration/DeploymentModeNotices.tsx
Bharath314, line 1: This comment isn't related to this PR, but a general observation about this file: Isn't this file's name misleading? I think a name like `ConfigFormSectionComponents` would be a more accurate descriptor of this file's components.

ui/components/connections/wizard/kubernetesDeploymentMode.tsx
Bharath314, line 1: These are refactoring changes not related to the title of the PR. Should these be kept?

ui/components/configuration/deploymentMode.ts
```diff
@@ -229,3 +244,23 @@ export const dormantPathsIn = (
...
+
+const longestLine = (value: string) => Math.max(1, ...value.split('\n').map((line) => line.length));
+const chars = (...parts: Array<string | number | undefined | null>) =>
+  Math.max(1, ...parts.map((part) => longestLine(String(part ?? ''))));
+
+export const fitWidth = (...parts: Array<string | number | undefined | null>) =>
+  `calc(${chars(...parts)}ch + 2.75em)`;
+
+export const fitNumberWidth = (...parts: Array<string | number | undefined | null>) =>
+  `calc(${chars(...parts)}ch + 4.75em)`;
```
Bharath314, line 266: These are generic utility functions. Should these be moved to /utils? Then they can be used anywhere.
For example, `fitWidth` can be given a parameter width to make it flexible to use for any width in any component or even different widths in the same component for responsiveness.

ui/components/configuration/ControllersConfigForm.tsx
```diff
@@ -1,27 +1,209 @@
...
+  if (whitelist.length === 0) {
+    whitelistKeysRef.current = [];
+  } else if (whitelistKeysRef.current.length < whitelist.length) {
+    whitelistKeysRef.current = [
+      ...whitelistKeysRef.current,
+      ...Array.from({ length: whitelist.length - whitelistKeysRef.current.length }, () =>
+        crypto.randomUUID(),
+      ),
+    ];
+  }
```
Bharath314, line 70: Using Ref like this is an [anti-pattern](https://react.dev/reference/react/useRef#usage:~:text=Do%20not%20write%20or%20read%20ref%2Ecurrent%20during%20rendering,-%2E). It's preferable to use a stable identifier like [Ishwar said](https://github.com/meshery/meshery/pull/21314#discussion_r3791944616).

Bharath314 (changes requested):

ui/components/configuration/ControllersConfigForm.tsx
```diff
@@ -1,27 +1,209 @@
...
+  fieldRowSx,
+} from './ControllersConfigFields';
+
+function ControllersConfigWatchList({
+  label,
+  value,
+  disabled,
+  onChange,
+}: {
+  label: ReactNode;
```
Bharath314: `ControllersConfigWatchList` isn't reused. There is no need to pass `label` as prop and it can be hardcoded inside this component

Bharath314, line 50: Since this is a one-time component, label can be hard-coded instead of being passed as a prop.

Bharath314:
image

Would it be a good idea to disable the inputs in these accordions or simply not render them if operator mode is not selected?

Bharath314:
Forcibly changing user input is generally considered bad UX. Can you display an error state instead if the numbers are out of range?

https://github.com/user-attachments/assets/73b5fcc9-730e-456e-a081-0d754b46fd1a

Off-topic: I think [this component](https://mui.com/material-ui/react-number-field/#spinner-field) would fit better here though it hasn't been implemented in sistent yet. Maybe we can replace it in a future PR?

PR #21338, opened 2026-08-11
mesheryctl: bypass interactive pagination when stdout is redirected (#21334)

### LINKED ISSUE #21334: mesheryctl pagination hangs when stdout is redirected
## Description

`mesheryctl` pagination remains interactive when stdout is redirected to a file or another non-terminal destination.

This causes the command to continue waiting for keyboard input even though its output is no longer being displayed on the terminal. This makes the command unsuitable for non-interactive usage such as shell scripts, redirected output, and CI/automation environments.

## Current Behavior

When pagination is triggered using a small page size, `mesheryctl` waits for keyboard input between pages.

For example:

```bash
./mesheryctl connection list --pagesize 1
```

When stdout is connected to a terminal, this interactive behavior is expected.

However, when stdout is redirected:

```bash
./mesheryctl connection list --pagesize 1 > /tmp/mesheryctl-output.txt
```

the command does not return control to the shell. The pagination flow continues to wait for keyboard input from the controlling terminal even though stdout has been redirected.

This means a command that is expected to behave as a non-interactive CLI invocation can remain blocked waiting for user input.

## Expected Behavior

`mesheryctl` should not wait for keyboard input when stdout is redirected, piped, or otherwise not connected to an interactive terminal.

When stdout is connected to a terminal, the existing interactive pagination behavior should be preserved.

When stdout is non-interactive:

* The command should not wait for keyboard input.
* Interactive pagination controls should be skipped.
* Terminal-specific screen-clearing operations should be skipped.
* Pagination should continue automatically until all available pages have been processed.
* The complete output should be written to the redirected destination.

## Reproduction

1. Build/run `mesheryctl` locally from the repository.
2. Ensure the Meshery server is running and has enough connections/results to produce multiple pages.
3. Force pagination with a small page size:

```bash
./mesheryctl connection list --pagesize 1
```

4. Confirm that the command displays a page and waits for keyboard input before proceeding.
5. Redirect stdout to a file:

```bash
./mesheryctl connection list --pagesize 1 > /tmp/mesheryctl-output.txt
```

6. Observe that the command remains running instead of returning to the shell.
7. The pagination flow still responds to keyboard input from the controlling terminal.

## Root Cause

The pagination implementation performs terminal-specific operations without first checking whether the command is running interactively.

In `internal/cli/pkg/display/pagination.go`, `utils.ClearLine()` is invoked when moving to subsequent pages:

```go
if currentPage > 0 {
    utils.ClearLine()
}
```

In `internal/cli/pkg/display/pagination_handlers.go`, keyboard input is requested whenever another page is available:

```go
keysEvents, err := keyboard.GetKeys(10)
```

There is currently no check to determine whether stdout is attached to an interactive terminal before these operations are performed.

### Interactive terminal

Image
### Redirected stdout

Image

## Environment

* **Host OS:** macOS
* **Meshery Server:** Local Meshery server running through Docker
* **Meshery Client:** Local build from the `mesheryctl` source repository
* **Platform:** Docker
* **Command tested:** `./mesheryctl connection list --pagesize 1`

Srujansa:
### Description
Fixes #21334

Bypasses interactive pagination in `mesheryctl connection list` when stdout is attached to a non-interactive terminal (`utils.IsInteractiveTerminal()`), preventing CLI commands from hanging in CI/scripting environments.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection listing when output is redirected or run in non-interactive environments.
  * Interactive pagination is now bypassed for non-interactive terminal output.

### DIFF
--- a/mesheryctl/internal/cli/root/connections/list.go
+++ b/mesheryctl/internal/cli/root/connections/list.go
@@ -81,6 +81,11 @@ mesheryctl connection list --count
 			DisplayCountOnly: connectionListFlagsProvided.count,
 		}
 
+		// Bypass interactive pagination if terminal is non-interactive (stdout redirected)
+		if !utils.IsInteractiveTerminal() {
+			data.PageSize = 0
+		}
+
 		return display.ListAsyncPagination(data, processConnectionData)
 	},
 }


suttonskate:
Correct diagnosis and a minimal fix - gating `data.PageSize = 0` on `!utils.IsInteractiveTerminal()` matches the described symptom (CLI hanging in CI/scripting when stdout is redirected and an interactive pager can never receive input).

One thing I'd want confirmed before merge, since I can't see `ListAsyncPagination`'s handling of `PageSize: 0` in this diff: does `0` mean "no pagination, return everything," or could it be interpreted as "zero page size" and silently return no results in some code path? If that's already an established convention elsewhere in this codebase (worth a quick grep for other `PageSize: 0` call sites), this is a clean fix; if `0` isn't otherwise a meaningful sentinel here, worth being explicit about that behavior in a comment so the next reader doesn't have to trace it.

Verdict: one thing to confirm (the `PageSize: 0` semantics) but the diagnosis and fix shape are correct; not blocking if that convention already exists elsewhere.

PR #21341, opened 2026-08-11
fix(ui): fix operator precedence in MeshSync connection stepper (#21337)

### LINKED ISSUE #21337: [UI] Verify Connection button never disables in the MeshSync registration stepper (operator precedence)
## Summary

In the MeshSync connection-registration stepper, the **Verify Connection button is never
disabled**. It is enabled even with no credential selected and no form input, because an
operator-precedence mistake makes the guard always evaluate `true`.

## Where

`ui/components/connections/meshSync/Stepper/StepperContent.tsx` (the
`CredentialDetailContent` step):

```js
useEffect(() => {
  if (selectedCredential !== null || (formState !== null && formState['secret']) !== undefined) {
    setDisableVerify(false);
  } else {
    setDisableVerify(true);
  }
}, [selectedCredential, formState]);
```

## Why it always passes

The parentheses close after `formState['secret']`, so the `&&` is grouped **before** the
`!== undefined` comparison:

```js
(formState !== null && formState['secret']) !== undefined
```

- `formState === null` -> left side is `false`, and `false !== undefined` is `true`.
- `formState !== null` -> left side is `formState['secret']`, and any value other than
  literal `undefined` is `true`.

So the right-hand operand of the `||` is `true` in every case where it is reached, and
`setDisableVerify(false)` always runs. The `else` branch is unreachable.

The intended condition is almost certainly:

```js
if (selectedCredential !== null || (formState !== null && formState['secret'] !== undefined)) {
```

## Impact

A user can press **Verify Connection** with nothing filled in. The registration event is
sent with an empty `credentialSecret`, so verification fails at the server instead of the
button being unavailable - a confusing failure in place of a disabled control.

## Reproduction

1. Connections -> register a MeshSync-discovered Prometheus or Grafana connection.
2. Reach the **Credential** step. Select no existing credential and type nothing.
3. The **Verify Connection** button is enabled; pressing it attempts verification and fails.

Expected: the button stays disabled until a credential is selected or a secret is entered.

## Provenance

Pre-existing; not introduced by any recent change. Surfaced by review during
https://github.com/meshery/meshery/pull/PLACEHOLDER (the credential `secret` tolerant-read
change, the blocking half of a coordinated change with layer5io/meshery-cloud#5918).
Deliberately scoped out of that PR: it is a user-visible behaviour change in an unrelated
flow that deserves its own reproduction and review rather than a ride-along in a
critical-path PR.

mayannkkk19:
### Summary
Fixes an operator precedence bug in `StepperContent.tsx` where an improperly parenthesized condition caused `false !== undefined` to evaluate to `true`. This prevented the "Verify Connection" button from being disabled in the MeshSync registration stepper when credentials/secrets were empty.

- Fix parenthesis grouping in disableVerify logic to check formState['secret'] !== undefined
- Ensure Verify Connection button remains disabled when no credential or secret is provided

Fixes #21337

**Notes for Reviewers**

- This PR fixes #21337 

## Summary by CodeRabbit

* **Bug Fixes**
  * New credential verification is now enabled only when the secret field contains non-empty, trimmed text.
  * Verification for existing credentials continues to be available when an existing credential is selected.
  * Prevents verification from being triggered when a new credential’s secret is blank or contains only whitespace.

### DIFF
--- a/ui/components/connections/meshSync/Stepper/StepperContent.tsx
+++ b/ui/components/connections/meshSync/Stepper/StepperContent.tsx
@@ -312,7 +312,7 @@ export const CredentialDetails = ({ sharedData, handleNext, handleRegistrationCo
   };
 
   useEffect(() => {
-    if (selectedCredential !== null || (formState !== null && (formState['secret'] !== undefined))) {
+    if (selectedCredential !== null || (formState !== null && formState['secret'] !== undefined)) {
       setDisableVerify(false);
     } else {
       setDisableVerify(true);
--- a/ui/components/connections/meshSync/Stepper/StepperContent.tsx
+++ b/ui/components/connections/meshSync/Stepper/StepperContent.tsx
@@ -312,7 +312,8 @@ export const CredentialDetails = ({ sharedData, handleNext, handleRegistrationCo
   };
 
   useEffect(() => {
-    if (selectedCredential !== null || (formState !== null && formState['secret'] !== undefined)) {
+    const hasSecret = typeof formState?.secret === 'string' && formState.secret.trim().length > 0;
+    if (selectedCredential !== null || hasSecret) {
       setDisableVerify(false);
     } else {
       setDisableVerify(true);
--- a/ui/components/connections/meshSync/Stepper/StepperContent.tsx
+++ b/ui/components/connections/meshSync/Stepper/StepperContent.tsx
@@ -312,7 +312,7 @@ export const CredentialDetails = ({ sharedData, handleNext, handleRegistrationCo
   };
 
   useEffect(() => {
-    if (selectedCredential !== null || (formState !== null && formState['secret']) !== undefined) {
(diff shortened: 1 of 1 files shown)


suttonskate:
This is exactly the kind of bug that's invisible in review and only shows up when you trace operator precedence by hand: `(formState !== null && formState['secret']) !== undefined` evaluates its left side to a boolean or the secret's value, and then `!== undefined` is true for nearly every outcome of that left side (including `false`), so the guard was effectively always true regardless of whether a secret was actually present - explaining why "Verify Connection" stayed enabled with empty credentials. The fix doesn't just re-parenthesize the same check; it replaces it with an explicit `typeof formState?.secret === 'string' && formState.secret.trim().length > 0`, which is both correct and stricter than the original intent (a whitespace-only secret no longer counts as present, which is the right call for a credential field).

No changes required - correct root-cause fix.

Issue #10736, opened 2024-04-16
[Models] Relationships: Expliciting define "intermediary" component in binding relationships


leecalcote:
### Current Situation
"to" and "from" are supported with an implicit intermediary component.

### Proposed Change
Make the intermediary component explicit.

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://layer5.io/community/handbook)

- [Models Repository](https://github.com/meshery/meshery/tree/master/server/meshmodel)
- 🖥 [Models Overview](https://docs.google.com/presentation/d/1SQMfyu5shjpGKlYONdVzOtd7UYTgLWBcgUvHMLCZ2tY/edit#slide=id.g226f5de5180_19_259)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](http://discuss.meshery.io) and [Community Slack](https://slack.meshery.io)

## Contributing to Meshery Models

!! See https://docs.meshery.io/project/contributing/contributing-models

### Instructions for Models

_Forthcoming_

### Instructions for Components
While the default shape for new components is a circle, each component should be considered for its best-fit shape.
1. Review and familiarize with the available set of predefined relationship types. Refer the Cytoscape [node types](https://js.cytoscape.org/demos/node-types/) for a list of possible shapes. 
1. Propose a specific shape, best-suited to visually represent the Component. _Example - Deployment as a pentagon._
1. Proposee a specific icon, best-suited to visually represent the Component. _Example - DaemonSet as a skull icon._

### Instructions for Relationships
1. Identify the relationship and any specific constraints to be enforced between the two specific components, their models or potentially other components, models, or environmental considerations.
1. Propose a specific visual representation for the relationship. See example on [Modeling Relationships.](https://docs.meshery.io/project/contributing/contributing-models)
1. Prospose the appropriate relationship type, using one of the predefined set of relationship types or suggest a new relationship where an existing type does not fit.
1. Create a Relationship Definition (yaml).
1. Create a policy for evaluation of the relationship (rego).
1. Review a prior pull request as an example of how to define a Relationships: https://github.com/meshery/meshery/pull/9880/files

### Instructions for Policies
1. _Forthcoming_

Omswastik-11:
Hi @leecalcote !!
Is it open ? Can I work on it ?

jamieplu:
Yes, and yes, @Omswastik-11. We will wait to assign it to you until you've demonstrated competency.

Omswastik-11:
[...]

PR #21343, opened 2026-08-11
feat(policy): evaluate explicit via intermediary binding relationships

Junnygram:
### Description
This PR updates the relationship policy evaluator to support explicit intermediary components (the `via` selector field introduced in meshery/schemas) inside binding relationships.

### Changes
* **Rego Policy Engine:**
  * Updated [identify_binding_policy.rego](file:///Users/junioroyewunmi/Desktop/meshery/meshery/meshery/models/meshery-core/0.7.2/v1.0.0/policies/identify_binding_policy.rego) to identify binding components using the new `allow.via` field when defined.
  * Added a seamless fallback union block to query the legacy implicit `match` selectors if the `via` block is not present (ensuring full backward-compatibility with older models).

### Associated Issue
This resolves the backend policy phase of [meshery/meshery#10736](https://github.com/meshery/meshery/issues/10736).

https://github.com/user-attachments/assets/fb80ba7b-ce27-4d4c-8a49-09c840cc8e36

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved relationship identification when optional connection selectors are provided.
  * Preserved existing behavior when connection selectors are unavailable.

### DIFF
--- a/models/meshery-core/0.7.2/v1.0.0/policies/identify_binding_policy.rego
+++ b/models/meshery-core/0.7.2/v1.0.0/policies/identify_binding_policy.rego
@@ -37,8 +37,13 @@ identify_relationship(
 	from := extract_components(design_file.components, from_selectors)
 	to := extract_components(design_file.components, to_selectors)
 
-	# should consider model and versions (regex/operator/normal string)
+	via_selectors := object.get(selector_set.allow, "via", [])
 	binding_comps := {type |
+		count(via_selectors) > 0
+		some via_selector in via_selectors
+		type = via_selector.kind
+	} union {type |
+		count(via_selectors) == 0
 		some match_selector in selector_set.allow.from[_].match
 
 		match_selector[0].kind != "self"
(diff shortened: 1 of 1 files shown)


suttonskate:
The `union(...)` of the new `via`-based selector set and the legacy `match`-based set, gated by `count(via_selectors) > 0` vs `== 0`, is the right shape for backward compatibility - existing relationship definitions that only use implicit `match` selectors keep evaluating exactly as before, and new definitions using the explicit `via` field get picked up without needing every existing relationship rewritten. That's a real fallback, not just a same-behavior no-op guard.

I can't run the Rego test suite from a diff read alone, so I'd want to confirm this is covered by `server/policies`-level tests exercising both the `via`-present and `via`-absent cases (mirroring the schema-conformance test pattern used elsewhere in this repo for relationship policy correctness) - if that coverage exists elsewhere in the PR and I'm just not seeing it in this diff excerpt, disregard.

No hard blockers based on the policy logic itself - please confirm test coverage for both selector paths before merge.

PR #21350, opened 2026-08-12
[Server] Check the error GetToken returns on connection and credential updates

Eason09053360:
**Notes for Reviewers**

`UpdateConnection` and `UpdateUserCredential` discard the token error and then guard on a stale one:

```go
_creds, err := json.Marshal(connection)
if err != nil { return nil, err }   // non-nil marshal error already returned here

tokenString, _ := l.GetToken(req)   // token error discarded
if err != nil {                     // so this tests a necessarily-nil err
    l.Log.Error(ErrGetToken(err))   // ...and this shows the check was meant for GetToken
    return nil, err
}
```

The guard can never fire, so a request with no session cookie travels on with an empty token and is `PUT` to the remote provider anyway. A missing token then surfaces as whatever the provider makes of an unauthenticated write — and against a provider answering `200`, as no error at all.

**Changes**

- `server/models/remote_provider.go`: `_` → `err` at both sites. The bare `err` return is kept because `GetToken` already wraps in `ErrGetToken`, matching the correct call sites at 5433/5889.
- Regression test asserting the remote provider is **never reached** when the token is absent.

Two other sites discard the token error and are deliberately left alone: `UpdateToken` returns the empty string by design, and `RecordPreferences` has no dead guard to establish intent — and sits in the region #21262 is editing.

**Verification**

Both tests fail without the fix, and the failure is the bug:

```
--- FAIL: TestRemoteProviderUpdateConnectionRequiresToken
    expected an error when the request carries no token, got connection
    &{ID:00000000-0000-0000-0000-000000000000 Name: ... Kind: ...}
```

No error — a blank object returned as a successful update. With the fix, both pass and the package is green (`go test ./server/models/` → `ok 20.728s`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Remote connection and credential updates now stop safely when a provider session token is unavailable.
  * Prevents update requests from being sent when token retrieval fails and reports the error appropriately.
  * Applies consistently to both remote connections and user credentials.

* **Tests**
  * Added coverage confirming tokenless updates return the expected error and do not contact the remote provider.

### DIFF
--- a/server/models/remote_provider.go
+++ b/server/models/remote_provider.go
@@ -4713,9 +4713,9 @@ func (l *RemoteProvider) UpdateConnection(req *http.Request, connection *connect
 	bf := bytes.NewBuffer(_creds)
 	remoteProviderURL, _ := url.Parse(fmt.Sprintf("%s%s/%s", l.RemoteProviderURL, ep, connection.Kind))
 	cReq, _ := http.NewRequest(http.MethodPut, remoteProviderURL.String(), bf)
-	tokenString, _ := l.GetToken(req)
+	tokenString, err := l.GetToken(req)
 	if err != nil {
-		l.Log.Error(ErrGetToken(err))
+		l.Log.Error(err)
 		return nil, err
 	}
 
@@ -5199,9 +5199,9 @@ func (l *RemoteProvider) UpdateUserCredential(req *http.Request, credential *Cre
 	bf := bytes.NewBuffer(_creds)
 	remoteProviderURL, _ := url.Parse(l.RemoteProviderURL + ep)
 	cReq, _ := http.NewRequest(http.MethodPut, remoteProviderURL.String(), bf)
-	tokenString, _ := l.GetToken(req)
+	tokenString, err := l.GetToken(req)
 	if err != nil {
-		l.Log.Error(ErrGetToken(err))
+		l.Log.Error(err)
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


suttonskate:
Sharp catch, and the PR description names the bug precisely: `tokenString, _ := l.GetToken(req)` discarded the real error, then `if err != nil` tested a stale `err` left over from the marshal call a few lines up - so a missing/invalid session token silently fell through to sending an empty token string to the remote provider instead of failing fast. `TestRemoteProviderUpdateConnectionRequiresToken`'s `countingProvider` asserting the remote endpoint is never hit (rather than just asserting on the error) is the right assertion - it proves the old code's actual failure mode (an empty-token PUT reaching the provider and coming back 200 with a zero-valued object and no error) can't happen anymore, not just that an error is now returned.

One small regression to flag: the fix drops the `ErrGetToken(err)` wrapper and logs the raw `err` instead (`l.Log.Error(err)` in both `UpdateConnection` and `UpdateUserCredential`). That loses the MeshKit structured error's code/cause/remediation that `ErrGetToken` was providing - worth keeping the wrapper (`l.Log.Error(ErrGetToken(err))`) so this failure mode is still identifiable by error code in logs, consistent with how the rest of this codebase surfaces MeshKit errors.

Verdict: one small change worth making (restore the `ErrGetToken` wrap in the log calls) but not a blocker - the core fix and its test are correct and this is close to ready.

PR #21363, opened 2026-08-13
[mesheryctl] Enforce 0600 permissions on auth file

### LINKED ISSUE #21239: [mesheryctl] Auth token file (~/.meshery/auth.json) created with overly permissive 0666/0644 mode instead of 0600
### Current Behavior

This is a low-severity local hardening issue (requires another local user
account on the same machine) rather than a remotely exploitable
vulnerability, but it deviates from how other CLIs (`kubectl`, `gh`, `aws`)
handle local credential files.

`mesheryctl system login` creates the authentication token file at
`~/.meshery/auth.json` with `0644` permissions when the system uses the
default umask of `022`.

The token is currently written in `internal/cli/root/system/login.go` using:

    os.WriteFile(token.GetLocation(), tokenData, 0o666)

With a umask of `022`, the resulting file permissions are:

    0666 & ~0022 = 0644

This leaves the authentication token readable by the file's group and
other users according to the Unix file permission bits.

I reproduced this using a locally built `mesheryctl` against a locally
running Meshery instance, authenticating through the remote `cloud.meshery.io`
provider so the token field is a real, populated JWT (1823 characters) rather
than an empty placeholder — confirming a genuine live credential is exposed
by this file mode, not just an empty field.

### Desired Behavior

The authentication token should be created with owner-only permissions,
such as `0600`:

    -rw------- 600 ~/.meshery/auth.json

The file creation could use:

    os.WriteFile(token.GetLocation(), tokenData, 0o600)

This would ensure that the authentication token is not readable by group
or other users through the file's Unix permission bits.

### Steps to Reproduce

1. Build mesheryctl from source:

       cd meshery/mesheryctl
       go build -o mesheryctl .

2. Start Meshery locally using Docker:

       ./mesheryctl system start -p docker

3. Set a remote provider (so the token field is populated):

       ./mesheryctl system provider set cloud.meshery.io

4. Remove any existing authentication file:

       rm ~/.meshery/auth.json

5. Log in (completes via browser OAuth against cloud.meshery.io):

       ./mesheryctl system login

6. Check the umask and resulting file permissions:

       umask
       ls -l ~/.meshery/auth.json
       stat -f '%Sp %Lp %N' ~/.meshery/auth.json

7. Confirm the token field is a real, non-empty credential (checking length
   only, never printing the value):

       python3 -c "
       import json
       d = json.load(open('$HOME/.meshery/auth.json'))
       for k, v in d.items():
           print(k, '->', len(str(v)), 'chars')
       "

### Screenshots/Logs

The attached screenshot shows the complete reproduction:

- `mesheryctl system provider set cloud.meshery.io` succeeds.
- `mesheryctl system login` authenticates via the `cloud.meshery.io` provider.
- The system umask is `022`.
- The resulting `~/.meshery/auth.json` has `0644` permissions.
- The `token` field is confirmed to be `1823` characters — a real JWT, not
  an empty/placeholder value.

Example output:

    === ENVIRONMENT ===
    022

    === AUTH FILE ===
    -rw-r--r-- ... ~/.meshery/auth.json
    -rw-r--r-- 644 ... ~/.meshery/auth.json

    meshery-provider -> 7 chars
    token -> 1823 chars

### Environment

- Host OS: macOS
- Meshery Server Version: stable-latest
- Meshery Client Version: locally built from source
- Platform: Docker
- Provider: cloud.meshery.io (remote)
- Shell: zsh
- umask: 022

### Additional Observation

A similar permissive file write exists in `internal/cli/root/system/logout.go`:

    os.WriteFile(token.GetLocation(), []byte{}, 0666)

I'd like to submit a PR fixing this (and the related `logout.go` write) if
that's welcome.

Image

Debagithub567:
## Notes for Reviewers

Fixes #21239

### Summary

`mesheryctl system login` and `mesheryctl system logout` previously wrote
`~/.meshery/auth.json` with `0666` permissions.

With a default `umask` of `022`, the resulting file permissions were
`0666 & ~0022 = 0644`, leaving the authentication token readable by group
and other users.

This PR introduces a shared `writeAuthTokenFile` helper that:

- Creates the authentication file with `0600` permissions.
- Explicitly applies `0600` using `os.Chmod` so existing files created by older versions are also corrected.
- Uses the same secure write path for both `login` and `logout`.

#### Manual validation

The original behavior was reproduced with:

```text
umask
022

stat -f '%Lp %N' ~/.meshery/auth.json
644 /Users/.../.meshery/auth.json


## Summary by CodeRabbit

* **Security**
  * Authentication tokens are now saved with restricted file permissions.
  * Token updates use safer atomic writes to prevent partial or corrupted files.
* **Bug Fixes**
  * Improved handling of login, logout, and token refresh failures.
  * Added clearer guidance when token storage fails, including checking write access to the Meshery configuration directory.
* **Tests**
  * Added coverage for secure permissions, permission correction, successful updates, and temporary-file cleanup.

### CODE REVIEW

**Review by ishwar170695 — COMMENTED** [review]
Aside from the coderabbit comments which are need to be done.
One question: `UpdateAuthDetails()` in `mesheryctl/pkg/utils/auth.go` still uses `os.WriteFile(..., os.ModePerm)`. Should that path also be updated to use the new helper?

File: mesheryctl/internal/cli/root/system/auth_file.go
Comment (lekaf974) [inline mesheryctl/internal/cli/root/system/auth_file.go:1]: I think if it is only for on method we could keep add this method in system.go for now and the tests in system_test.go

File: mesheryctl/pkg/utils/auth_file_test.go [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+	})
+
+	t.Run("corrects existing insecure permissions to 0600", func(t *testing.T) {
+		path := filepath.Join(t.TempDir(), "auth.json")
+
+		err := os.WriteFile(path, []byte(`{"token":"test-token"}`), 0o644)
+		require.NoError(t, err)
+
+		err = writeAuthTokenFile(path, []byte(`{"token":"updated-token"}`))
+		require.NoError(t, err)
```
lekaf974: use assert

mesheryctl/internal/cli/root/system/auth_file.go
```diff
@@ -0,0 +1,11 @@
+package system
+
+import "os"
+
+func writeAuthTokenFile(path string, data []byte) error {
+	if err := os.WriteFile(path, data, 0o600); err != nil {
+		return err
```
lekaf974: I think it should be a dedicated Meshkit Error ErrInvalidFilePermission with proper details

lekaf974 (changes requested):
Please resolve all comments already fix and check those I added

Debagithub567:
> Aside from the coderabbit comments which are need to be done. One question: `UpdateAuthDetails()` in `mesheryctl/pkg/utils/auth.go` still uses `os.WriteFile(..., os.ModePerm)`. Should that path also be updated to use the new helper?

Yes, I updated UpdateAuthDetails() to use the same WriteAuthTokenFile helper. This ensures login, logout, and token refresh all use the same atomic 0600 write path and avoids leaving the os.ModePerm write in place.

mesheryctl/internal/cli/root/system/auth_file.go
Debagithub567: > I think if it is only for on method we could keep add this method in system.go for now and the tests in system_test.go

I moved the helper to pkg/utils because UpdateAuthDetails() in pkg/utils/auth.go also writes the auth file and now uses the same helper. Keeping WriteAuthTokenFile in pkg/utils gives all three callers a single implementation and avoids introducing a dependency from pkg/utils back into the system package.

mesheryctl/pkg/utils/auth_file_test.go
Debagithub567: Addressed. I initially used assert as suggested, but the latest repository lint rule/CodeRabbit check requires the standard Go testing library for mesheryctl tests, so the final test uses t.Fatalf/t.Errorf instead.

mesheryctl/internal/cli/root/system/auth_file.go
Debagithub567: I kept a dedicated ErrWriteAuthTokenFile for the helper because the helper can fail at multiple stages (temp-file creation, chmod, write, close, or rename), and not all of those represent an invalid file permission. The permission-specific failure is only the Chmod(0600) step. Using ErrWriteAuthTokenFile for the overall operation keeps the error semantics accurate while still exposing the underlying error.

I also moved the helper to pkg/utils because UpdateAuthDetails() uses the same auth-file write path, so login, logout, and token refresh now share one implementation.

mesheryctl/pkg/utils/auth_file.go
lekaf974, line 1: why create a new file for auth since auth.go exists ?

mesheryctl/pkg/utils/auth_file_test.go
lekaf974, line 1: should be in auth_test.go file

mesheryctl/pkg/utils/error.go
```diff
@@ -17,6 +17,7 @@ var (
 	ErrAttachAuthTokenCode            = "mesheryctl-1093"
 	ErrUnmarshalCode                  = "mesheryctl-1094"
 	ErrFileReadCode                   = "mesheryctl-1095"
+	ErrWriteAuthTokenFileCode         = "mesheryctl-1116"
```
lekaf974, line 20: I don't see the file generate by errorutil tools modified so this has been added manually, please check meshery error code documentation

lekaf974 (changes requested):
Please check my comments

suttonskate:
Correct fix for a real information-disclosure issue - `0666` with a `022` umask left the auth token world/group-readable at `0644`, and the fix does more than just chmod the final file: `WriteAuthTokenFile` writes to a temp file in the same directory, chmods it to `0600` *before* writing the token data, then atomically renames it into place. That ordering matters - it closes the window where a partially-written or default-permission file could be readable before the final chmod, which a naive "write then chmod" fix would not.

Test coverage checks both the fresh-file case and correcting pre-existing insecure permissions on an existing file, which is exactly the scenario that matters for users upgrading from an older mesheryctl with an already-644 auth.json sitting on disk.

No changes required - solid security fix, ready as-is.

lekaf974 (changes requested):
new error code needs to use errorutil tool not be added manually

Issue #21077, opened 2026-08-02
[Test Plan] Test #2: Install mesheryctl and start Meshery on an invalid platform


yi-nuo426:
Automate and verify test case **#2** from the [Meshery Test Plan](https://docs.google.com/spreadsheets/d/13Ir4gfaKoAX9r8qYjAFFl_U9ntke4X5ndREY1T7bnVs/edit?gid=838298230#gid=838298230) ("Latest" tab), currently marked **Absent (Help Wanted)**.

- **Test #:** 2
- **Group:** Installation
- **Component under test:** Installer / Bash Script

**Scenario:** Install mesheryctl and start Meshery on an invalid platform.

**Action:**
```sh
curl -L https://git.io/meshery | PLATFORM=bob bash -
```

**Expected outcome:** Throws an error that the platform is invalid (unless a valid platform is provided).

---
This test case is not yet automated. This issue tracks implementing it (e.g. as a BATS case under `mesheryctl/tests/e2e/`, or an install-script test) and updating the Test Plan status once it lands. Help wanted.

hiyach28:
Hi! @yi-nuo426 I'd like to work on this issue.

I've investigated this specific test case and reviewed the platform validation flow in the `meshery.io` install script.

**Build & environment:** I've already verified my local build environment and the existing E2E testing setup, so I'm ready to work on this scenario.

**Implementation approach:**
- Add a new BATS test covering this platform validation scenario, following the existing E2E conventions.
- Verify that an invalid platform is rejected with the expected error message.
- Verify that the installer either exits appropriately after repeated invalid attempts or successfully proceeds once a valid platform is supplied, matching the current installer behavior.
- Ensure the test performs proper cleanup so it remains isolated from the rest of the E2E suite.

Confirmed that the current installer already validates invalid platform values and retries user input through its interactive prompt. The automated test will focus on verifying both the failure and recovery paths so this behavior remains protected against future regressions.

Please let me know if this aligns and I'd be happy to work on it.

Katotodan:
I would like to work on this please

PR #21395, opened 2026-08-14
[Test] Add e2e tests for mesheryctl installation with invalid platform and valid platform

Katotodan:
## Description

This test simulates the installation of `mesheryctl` using both `curl` and Homebrew.

It covers the following scenarios:

* Verifies that the installer rejects invalid platform values after multiple invalid attempts.
* Verifies that `mesheryctl` can be installed without deploying Meshery when `DEPLOY_MESHERY=false`.
* Verifies that Meshery is successfully installed when a valid platform, such as Docker or Kubernetes, is provided.
* Verifies that the Homebrew installation of `mesheryctl` completes successfully.

A CI workflow has also been added to run these E2E tests automatically. This helps prevent future changes to the installation process from introducing regressions and ensures that the installer continues to work as expected.

The CI environment runs the tests in an isolated environment, with cleanup performed after each test to remove resources and artifacts from the previous installation. A defensive cleanup is also performed before the test suite starts in case a previous run left any resources behind.

- This PR fixes ##21077

## Summary by CodeRabbit

* **Tests**
  * Added automated end-to-end coverage for Meshery installer workflows across Docker, Kubernetes, and Homebrew.
  * Validated invalid platform handling and installation behavior when deployment is disabled.
  * Added safeguards against stalled test runs and comprehensive cleanup between scenarios.
  * Added automated failure diagnostics for Kubernetes, Docker, and Helm environments to simplify troubleshooting.


mesheryctl/tests/e2e/010-installation/00-cli-install-invalid-platform.bats
```diff
@@ -0,0 +1,41 @@
...
+    $MESHERYCTL_BIN system stop 2>/dev/null || true
+    sleep 1
+}
+
+@test "[TC-0001][tg=Installation] install mesheryctl and start Meshery on an invalid platform" {
+    # Test that installing mesheryctl with an invalid platform fails gracefully
+    
+    run bash -c 'curl -L https://meshery.io/install | PLATFORM=bob bash - < /dev/null'
+    
+    assert_output --partial "curl: Failed writing body"
```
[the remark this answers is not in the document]
Katotodan: Removing < /dev/null keeps the test in a pending state forever.

mesheryctl/tests/e2e/010-installation/00-cli-install-invalid-platform.bats
```diff
@@ -0,0 +1,43 @@
...
+  # Extract ONLY the getopts block into a variable for focused testing
+  export GETOPTS_BLOCK
+  GETOPTS_BLOCK=$(sed -n '/while getopts/,/done/p' "$INSTALL_SCRIPT")
+}
+
+teardown_file() {
+  rm -f "$INSTALL_SCRIPT"
+}
+
+@test "Supported platform (e.g. 'docker') is defined in getopts cases" {
```
lekaf974: please check other e2e test naming that use given, when then (same comment for other tests that follow)

mesheryctl/tests/e2e/010-installation/00-cli-install-invalid-platform.bats
```diff
@@ -0,0 +1,43 @@
...
+  # Check if 'docker)' case exists specifically inside the getopts block
+  run grep -E '^\s*docker\)' <<< "$GETOPTS_BLOCK"
+
+  [ "$status" -eq 0 ]
+  [[ "$output" =~ "docker" ]]
```
lekaf974: please use assert like it is done in other tests

mesheryctl/tests/e2e/010-installation/00-cli-install-invalid-platform.bats
```diff
@@ -0,0 +1,43 @@
...
+
+teardown_file() {
+  rm -f "$INSTALL_SCRIPT"
+}
+
+@test "Supported platform (e.g. 'docker') is defined in getopts cases" {
+  # Check if 'docker)' case exists specifically inside the getopts block
+  run grep -E '^\s*docker\)' <<< "$GETOPTS_BLOCK"
+
+  [ "$status" -eq 0 ]
```
lekaf974: please use assert like in other tests

lekaf974 (changes requested):
@Katotodan this looks good, small comments added mainly to follow the implementation standard

SuryanshGarg04:
I may be missing something here, but the current tests only use `grep` to check the script text. They don't actually test what happens when an invalid platform like `bob` is passed.
Could we test the actual platform validation logic with an invalid value, without running the full installer? That would give us more direct coverage for #21077

lekaf974:
> I may be missing something here, but the current tests only use `grep` to check the script text. They don't actually test what happens when an invalid platform like `bob` is passed. Could we test the actual platform validation logic with an invalid value, without running the full installer? That would give us more direct coverage for #21077

Good point

KumarNirupam1 (changes requested):
one concern with the test setup: `GETOPTS_BLOCK` is extracted using `sed` from `while getopts` to `done`, so the test depends on the exact structure of the install script rather than the validation behavior itself. a small refactor of that block could make these tests fail even when the platform handling is still correct.

can we make the test less dependent on the script's formatting/structure while keeping the same focused scope?

Katotodan:
@SuryanshGarg04 , @KumarNirupam1 

The initial approach I considered was to write the test like this:

\`\`\`bash
@test "[TC-0002][tg=Installation] install mesheryctl and start Meshery on a valid platform" {
    # Test that installing mesheryctl with a valid platform succeeds
    
    run bash -c 'curl -L https://meshery.io/install | PLATFORM=docker bash - < /dev/null'
    
    assert_success  
}
\`\`\`

While this approach is straightforward, it faces an issue when an invalid `PLATFORM` is passed: rather than failing immediately, the script hangs while waiting for interactive terminal input via `< /dev/tty` and eventually exits after 5 failed attempts. Because an automated test environment cannot interactively respond, this causes the test suite to freeze.

To avoid this, I used static analysis on the downloaded installer script. `ALLOWED_CASES` and `ENV_ALLOWED_CASES` extract and validate all platform values accepted by the `-p` flag and the `PLATFORM` environment variable. 

The test strictly asserts that:
1. Exactly two platform options exist.
2. Those options are strictly `docker` and `kubernetes`.

If a contributor modifies the installer script to introduce an unsupported or unauthorized platform, the case count and value checks will fail automatically without hanging the test process.

Katotodan:
To better test this locally, run curl -L https://meshery.io/install  and save the output inside a local file.
Then test these variables locally 
```
ALLOWED_CASES=$(sed -n '/case $OPTARG in/,/esac/p' "$INSTALL_SCRIPT" | grep -E '^\s+[a-zA-Z0-9_-]+\)' | tr -d ' )' | grep -v '^\*$' | sort -u)

  # 2. Extracts platform cases from the environment variable check block ($PLATFORM)
  ENV_ALLOWED_CASES=$(awk '/case "\$PLATFORM" in/,/esac/' "$INSTALL_SCRIPT" | grep -E '\)' | tr -d ' );' | tr '|' '\n' | grep -v '^\*' | grep -v '^$' | sort -u)
```
Change $INSTALL_SCRIPT to the path of the installer script. 
Echo those variables to be sure of their values.

Try to modify the saved installer script to support more platform and rerun the ALLOWED_CASES, and ENV_ALLOWED_CASES again and view how it adapts. 

This only confirms those two variables are well configured.

KumarNirupam1 (approved):
@Katotodan

I ran the checks you mentioned and everything looks good.

- Ran your extraction commands against the live install script — both ALLOWED_CASES and ENV_ALLOWED_CASES correctly show only docker and kubernetes.
- Also ran the BATS tests (make e2e-no-build BATS_FOLDER_PATTERN=010-installation) — both pass, 2/2.
- Then I edited a copy of the install script to add a fake platform (windows) and re-ran the extraction — both variables picked it up, so the test would correctly fail if someone adds an unsupported platform. Good to see it adapts.

That addresses my concern about the test being too dependent on the script structure. LGTM 👍

image


Image

mesheryctl/tests/e2e/010-installation/00-cli-install-invalid-platform.bats
```diff
@@ -0,0 +1,43 @@
...
+  # 2. Extracts platform cases from the environment variable check block ($PLATFORM)
+  ENV_ALLOWED_CASES=$(awk '/case "\$PLATFORM" in/,/esac/' "$INSTALL_SCRIPT" | grep -E '\)' | tr -d ' );' | tr '|' '\n' | grep -v '^\*' | grep -v '^$' | sort -u)
+}
+
+teardown_file() {
+  rm -f "$INSTALL_SCRIPT"
+}
+
+
+@test "[TC-0100][tg=installation]verify getopts flag -p contains only expected supported platforms (docker or kubernetes only)" {
```
lekaf974: we are using `given, when, then` naming convention from BDD in e2e tests

mesheryctl/tests/e2e/010-installation/00-cli-install-invalid-platform.bats
```diff
@@ -0,0 +1,43 @@
...
+  run echo "$ALLOWED_CASES"
+
+  assert_success
+  assert_line --index 0 "docker"
+  assert_line --index 1 "kubernetes"
+  assert_equal "${#lines[@]}" 2
+}
+
+@test "[TC-0101][tg=installation]verify PLATFORM environment variable block contains only expected supported platforms (docker or kubernetes only)" {
```
lekaf974: here why are you testing environment variables defines previously ?

lekaf974 (changes requested):
You are not doing E2E test, what must be done is to execute the install script and ensure it is behaving as expected.

also this will not work properly in this e2e suite since meshery install is done in CI in previous steps before the suite is running so this needs to be refactored for example with one folder to run mesheryctl subcommands e2e and another one for the installation process of meshery it self

Katotodan:
> You are not doing E2E test, what must be done is to execute the install script and ensure it is behaving as expected.
> 
> also this will not work properly in this e2e suite since meshery install is done in CI in previous steps before the suite is running so this needs to be refactored for example with one folder to run mesheryctl subcommands e2e and another one for the installation process of meshery it self

Thank you @lekaf974 for your reviews,
I have changed the approach to mimic the actual user interaction and make sure the test covers running:

`curl -L https://meshery.io/install | PLATFORM=bob bash -`

with an invalid platform.

The expected error message should be displayed whenever a user attempts to install Meshery using an unsupported platform.
Note that the installation failed gracefully. 

This test is specifically focused on verifying how the installation process behaves when an invalid platform is provided through the `curl` installation command. The other tests are already added for `mesheryctl` cover the system and how to start Meshery using `mesheryctl`.

I am planning to add more tests for other installation methods, such as Homebrew, but for now, I am using this test to validate the approach and make sure I am heading in the right direction.

Since this test only covers invalids platform, there is no `.gz` artifact being added to the codebase. Therefore, I don't think there should be any conflict with `MESHERY_BIN` or any other environment variable.

suttonskate:
Good diagnosis in the PR description - correctly identifying that piping a *valid* platform through the live installer hangs because it proceeds into the real install flow, and scoping this test to only exercise the invalid-platform rejection path (never sending a valid selection) avoids that hang while still testing real logic.

That said, this has the same shape of external dependency as #21485 in this sweep: `curl -L https://meshery.io/install | PLATFORM=bob bash -` pulls the live production installer at test time. If the prompt copy ("Invalid platform. Enter a valid platform...") or the platform-selection flow changes upstream, this test breaks with zero code change here - and this test's technique of piping straight into `bash` (rather than #21485's isolated block-extraction) means an installer change is even more likely to introduce unrelated side effects during a CI run before the invalid-platform loop is reached. Given both PRs hit this same fragility independently, it might be worth the maintainers deciding on one canonical approach (e.g., a pinned/local copy of the installer script for bats fixtures) rather than two different live-fetch strategies landing separately.

Not a blocker on its own claim (regression coverage for the invalid-platform-exits-after-5-attempts behavior), but I'd want the live-fetch-at-test-time question answered the same way as #21485 before merge.

Katotodan:
@lekaf974, and @suttonskate, thank you for your reviews.
I have made some changes to the approach. Instead of testing against the live production installer, I now download the installer script and test only the platform validation logic, without installing or attempting to install Meshery or mesheryctl.
I spliced the installer script so that only the part responsible for validating the PLATFORM value is executed.
Please review this approach and let me know if you think it is more effective than the previous approaches.
Although the test runs against only part of the installer, it still mimics the relevant user experience and verifies how the installer handles an invalid platform.
Thanks!

Katotodan:
@coderabbitai Please perform a full review of this PR.

Focus on:
- Correctness and potential bugs
- Test coverage and test reliability
- CI/CD workflow correctness
- E2E test isolation and cleanup
- Cross-platform compatibility
- Failure and timeout handling
- Security and supply-chain concerns
- Maintainability and readability
- Potential flaky tests
- Any edge cases or regressions

Please review the entire PR, including both the test code and the GitHub Actions workflow, and report any issues you find. Do not limit the review to changed lines if additional surrounding code is relevant.

install/tests/e2e/001-installation/installer.bats
```diff
@@ -0,0 +1,285 @@
...
+  run timeout "$wall_timeout" expect -c "$script"
+}
+
+
+@test "Given wrong platforms provided when installing then installer exits after 5 invalid attempts" {
+  run_expect '
+    set timeout 150
+    log_user 1
+
+    spawn bash -c {curl -L https://meshery.io/install | PLATFORM=bob bash -}
```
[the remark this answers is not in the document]
Katotodan: No, this PR aims to test against the production installer.

install/tests/e2e/001-installation/installer.bats
```diff
@@ -0,0 +1,285 @@
...
+      -re {(?i)select.*platform|platform.*:} { send "\r" }
+      timeout { exit 2 }
+    }
+
+    expect {
+      -re {(?i)select.*platform|platform.*:} { send "bob\r" }
+      timeout { exit 2 }
+    }
+
+    expect eof
```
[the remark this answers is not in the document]
Katotodan: On wrong platform, the installation failed gracefully, it just provides an error message. And this test check if that error message was provided and that meshery did not install successfully (refute_output check this)

.github/workflows/meshery-installer-e2e-test.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+      # ubuntu-24.04 runners do not ship with `expect` preinstalled, and none
+      # of the setup actions above (BATS, minikube, Homebrew) provide it —
+      # the installer.bats tests rely on it directly for interactive prompts.
+      - name: Install expect
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y expect
+
+      - name: Verify dependencies
```
lekaf974, line 67: this step is not useful, those dependencies are builtin in github runner's images and except those explicitly installed before

install/tests/e2e/001-installation/installer.bats
```diff
@@ -0,0 +1,289 @@
...
+  ' 180
+
+  assert_success
```
lekaf974, line 192: from what I am seeing in the script `exit 1` in the loop this should be assert_failure

install/tests/e2e/001-installation/installer.bats
```diff
@@ -0,0 +1,289 @@
...
+  assert_output --partial "Too many invalid attempts. Please try again."
+  # The script should reject each bad value before ultimately bailing out,
+  # not just print the final message with a lucky earlier match.
+  refute_output --partial "Starting Meshery..."
+}
+
+
+@test "Given docker as the platform when installing then docker is accepted" {
```
lekaf974, line 200:
```suggestion
@test "Given docker as the platform when installing then succeed" {
```

install/tests/e2e/001-installation/installer.bats
```diff
@@ -0,0 +1,289 @@
...
+  run bash -lc "command -v mesheryctl"
+  assert_success
+
+  run bash -lc "mesheryctl system status"
+  assert_success
+ 
+}
+
+
+@test "Given kubernetes as the platform when installing then kubernetes is accepted" {
```
lekaf974, line 225:
```suggestion
@test "Given kubernetes as the platform when installing then succed" {
```

install/tests/e2e/001-installation/installer.bats
```diff
@@ -0,0 +1,289 @@
...
+# Helper: run an `expect` script with a hard wall-clock timeout so a hang
+# in the installer cannot stall the whole suite.
+run_expect() {
+  local script="$1"
+  local wall_timeout="${2:-300}"
+  run timeout "$wall_timeout" expect -c "$script"
+}
+
+
+@test "Given wrong platforms provided when installing then installer exits after 5 invalid attempts" {
```
lekaf974, line 157:
```suggestion
@test "Given wrong platforms provided when installing then failed after 5 invalid attempts" {
```

install/tests/e2e/001-installation/installer.bats
```diff
@@ -0,0 +1,289 @@
...
+
+  assert_success
+  assert_output --partial "Meshery deployed on Kubernetes."
+
+  # Functional check: verify the Meshery pods actually exist and reach
+  # Ready state, rather than trusting only the printed success message.
+  # Wait on all pods in the namespace rather than a specific label selector
+  # — the chart's actual labels aren't guaranteed to match a guessed
+  # app.kubernetes.io/part-of value, and --all avoids depending on that.
+  run bash -lc "kubectl get pods -n meshery --no-headers"
```
lekaf974, line 242: there is a detik bats library (already installed via the bats-core/bats-action) to validate kubernetes stuff like this https://github.com/meshery/meshery/blob/7a635f4b4d5e49e8a800e2d5bf4ef4298b4e42a7/mesheryctl/tests/e2e/000-prerequisites/00-servert.bats#L24

install/tests/e2e/001-installation/installer.bats
```diff
@@ -0,0 +1,289 @@
...
+  assert_success
+
+  run bash -lc "kubectl get pods -n meshery --no-headers"
+  assert_success
+  refute_output --partial "CrashLoopBackOff"
+  refute_output --partial "Error"
+}
+
+
+@test "Given DEPLOY_MESHERY=false when installing, Only install mesheryctl binary" {
```
lekaf974, line 256:
```suggestion
@test "Given DEPLOY_MESHERY=false when installing then mesheryctl binary is present" {
```

install/tests/e2e/001-installation/installer.bats
```diff
@@ -0,0 +1,289 @@
...
+    expect eof
+  ' 320
+
+  assert_success
+  assert_output --partial "mesheryctl installed."
+  assert_output --partial "Run \"mesheryctl system start\" to start Meshery."
+
+  # DEPLOY_MESHERY=false should mean nothing was actually deployed — verify
+  # that no meshery namespace/resources were created as a side effect.
+  run bash -lc "kubectl get namespace meshery"
```
lekaf974, line 271: use detik

install/tests/e2e/001-installation/installer.bats
```diff
@@ -0,0 +1,289 @@
...
+  assert_failure
+
+  # Assert that no Meshery Docker container exists after installation.
+  run bash -lc 'test -z "$(docker ps -aq --filter name=meshery)"'
+  assert_success
+}
+
+
+@test "Given Homebrew when installing Meshery CLI then mesheryctl is installed" {
```
lekaf974, line 280: I would have spiit in 2 files script tests in one file, homebrew in another one

lekaf974 (changes requested):
@Katotodan I added some comments mainly cosmetic, LGTM overall

lekaf974 (approved):
LGTM

Issue #20994, opened 2026-07-28
Feature: Adapter for AI and LLMs


hortison:
### Description
Meshery is the open-source cloud native manager that empowers platform engineers to design and operate infrastructure. As infrastructure complexity grows, the need for intelligent assistance becomes critical.

This project focuses on developing and enhancing a dedicated AI Adapter and AI Connections for Meshery. This adapter serves as the bridge between Meshery’s core orchestration engine and various Large Language Models (LLMs). The goal is to enable "Natural Language to Infrastructure" capabilities, allowing users to describe their architectural intent (e.g., "Deploy a highly available Kubernetes cluster on AWS with Prometheus monitoring") and have Meshery auto-generate the visual topology and configuration manifests.

Decouple the AI logic from the core platform, allowing users to "Bring Your Own Model" (BYOM)—supporting both cloud-based providers (OpenAI, Anthropic) and local inference runners.

### Implementation
  - Co-design and implement the interface for the AI Adapter in Go to communicate with the Meshery Server.
  - Implement support for connecting to local LLMs to ensure data privacy for users who cannot send infrastructure data to the public cloud.
  - Improve the "System Prompt" and context-window management to feed the LLM relevant data regarding Meshery Models (schema definitions) so the AI generates valid infrastructure configurations.
  - Write unit and integration tests to ensure the reliability of the adapter.
  - Create user guides on how to configure the adapter with different AI providers.

### Acceptance Tests
  - A fully functional AI Adapter (or Connection) integrated into the Meshery ecosystem.
  - Demonstrable capability for users to swap between at least two different LLM providers (e.g., OpenAI vs. a local Llama 3 model).
  - Implementation of a feature where natural language queries result in a rendered design.
  - Merged pull requests (PRs) including code, tests, and documentation.

### Mockups

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and designs for Meshery UI in [Figma](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI)
[...]

PR #21428, opened 2026-08-16
connections: add minimal OpenAI Connection/Credential types

Faisal77666:
## What this is

A minimal starting point for #20994 (BYOM: Adapter for AI and LLMs) 
`OpenAIConn`/`OpenAICred` following the existing `PromConn`/`PromCred`
pattern in `server/models/connections/connections.go`.

## Why this shape

While reading the existing pattern, I noticed `PromCred` and
`GrafanaCred` expose their secret via an exported, json-tagged field
(`APIKeyOrBasicAuth string \`json:"secret,omitempty"\``)  meaning the
raw credential is reachable through `json.Marshal`. `OpenAICred` fixes
this for the new provider: the secret is unexported and only reachable
via a `Secret()` accessor for internal callers, verified by
`TestOpenAICred_SecretNeverMarshaled`.

This seemed directly relevant to the issue's requirement that "secrets
... never appear in logs or events"  happy to open a separate issue
for the existing Prometheus/Grafana gap if that's useful, rather than
scope-creep it into this PR.

## What's here

- `OpenAIConn` / `OpenAICred` structs
- `Secret()` accessor
- A test proving the secret is absent from marshaled JSON

## What's intentionally not here yet

- No provider client / HTTP calls
- No health-check wiring (would follow the `RegisterAction.Execute`
  pattern used in `server/machines/prometheus/register.go`)
- No UI, no generation path

Opening this as a draft to get early feedback on whether this is the
right shape before extending it  not requesting merge yet.

### DIFF
--- a/server/models/connections/openai.go
+++ b/server/models/connections/openai.go
@@ -1,0 +1,24 @@
+package connections
+
+// OpenAIConn mirrors the existing PromConn/GrafanaConn pattern.
+type OpenAIConn struct {
+	URL  string `json:"url,omitempty"`
+	Name string `json:"name,omitempty"`
+}
+
+// OpenAICred fixes the secret-exposure gap present in PromCred/GrafanaCred:
+// the raw key is never a marshalable field.
+type OpenAICred struct {
+	Name string `json:"name,omitempty"`
+	apiKeyOrBasicAuth string
+}
+
+func NewOpenAICred(name, secret string) OpenAICred {
+	return OpenAICred{Name: name, apiKeyOrBasicAuth: secret}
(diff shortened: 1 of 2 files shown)


leecalcote:
Still draft or ready for review?

Faisal77666:
@leecalcote  Ready for review on the shape/approach  not requesting merge yet. It's intentionally minimal (see PR description) as a starting point for #20994; wanted early feedback on whether this pattern is right before extending it to the other providers. Happy to mark it non-draft if that's a clearer signal for review.

PR #21441, opened 2026-08-17
[mesheryctl] Fix system logs --follow spinning forever after the log stream ends

### LINKED ISSUE #21440: mesheryctl system logs --follow busy-spins forever once the log stream reports EOF
In `mesheryctl/internal/cli/root/system/logs.go`, the follow-mode read loop checks `numBytes == 0` and issues `continue` *before* checking whether `err == io.EOF`:

```go
for {
	buf := make([]byte, BYTE_SIZE)
	numBytes, err := logs.Read(buf)
	if numBytes == 0 {
		continue
	}
	if err == io.EOF {
		break
	}
	...
}
```

Per the `io.Reader` contract, once a stream's data is exhausted it's expected to return `(0, io.EOF)` on every subsequent call. The `continue` above runs first every time, so the `io.EOF` check is never reached - the loop spins on `Read()` indefinitely instead of exiting.

Reproduced with a fake `io.Reader` that returns one chunk of data, then `(0, io.EOF)` on every call after: **353,605,192 iterations in 3 seconds** against a closed stream. Root cause is in the check ordering, not platform-specific (confirmed the logic on Windows; the missing check applies identically on Linux, where this actually ships).

In practice, this means `mesheryctl system logs --follow <pod>` never returns and burns a full CPU core once a container's log stream ends (container restart, pod deletion, etc.) instead of returning to `wg.Wait()` and letting the command exit cleanly.

Will open a PR shortly with the fix + regression test.

Atishyy27:
Fixes #21440

## What
The `system logs --follow` read loop checked `numBytes == 0` and issued `continue` before checking `err == io.EOF`. A `Reader` is allowed to return `(0, io.EOF)` once exhausted - the `continue` skipped past that check every time, so the loop never reached its break condition.

Extracted the loop into `followLogs(logs io.Reader, podName string)` so it's unit-testable without a real Kubernetes client, and reordered the checks to match the `io.Reader` contract: process `n > 0` bytes first, then check `io.EOF`, then other errors.

## Why it matters
Once a container's log stream ends (restart, pod deletion, etc.), `mesheryctl system logs --follow <pod>` never returns - it spins on `Read()` indefinitely, burning a full CPU core, instead of reaching `wg.Wait()` and exiting cleanly.

## Test
Added `TestFollowLogsReturnsOnEOF`, using a `Reader` that mimics the same `(n>0, nil)` then `(0, io.EOF)` sequence a real pod log stream produces, and asserts `followLogs` returns within 2s. Standalone repro (outside the test suite, against the pre-fix inline loop) measured 353,605,192 iterations in 3 seconds against a closed stream.

## Metrics
- 2 files changed: `mesheryctl/internal/cli/root/system/logs.go`, `mesheryctl/internal/cli/root/system/logs_test.go`
- Before: unbounded busy-loop once the stream reports EOF. After: returns within a read cycle.
- `go build ./...` and `go test ./internal/cli/root/system/...` both pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved log-following behavior when a log stream ends after returning data.
  * Prevented follow mode from hanging when Kubernetes log streams close normally.
  * Preserved available log output before stream termination and improved handling of other read errors.

* **Tests**
  * Added coverage for log streams that return data followed by an end-of-file signal.
  * Verified that final log chunks are displayed even when stream termination occurs simultaneously.

### DIFF
--- a/mesheryctl/internal/cli/root/system/logs.go
+++ b/mesheryctl/internal/cli/root/system/logs.go
@@ -53,6 +53,26 @@ func printLogs(logs string, podName string) {
 	}
 }
 
+// followLogs reads from logs until it hits EOF or an error, printing each chunk
+// as it arrives. It must process any bytes returned before checking the error,
+// since a Reader can return (n > 0, io.EOF) on the same call.
+func followLogs(logs io.Reader, podName string) {
+	for {
+		buf := make([]byte, BYTE_SIZE)
+		numBytes, err := logs.Read(buf)
+		if numBytes > 0 {
+			printLogs(string(buf[0:numBytes]), podName)
+		}
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			utils.Log.Errorf("error occurred while processing logs %v", err)
+			break
+		}
+	}
+}
+
 type cmdSystemLogsFlags struct {
 	Follow bool `json:"follow" validate:"boolean"`
 }
@@ -240,22 +260,7 @@ mesheryctl system logs meshery-istio
 						wg.Add(1)
 						go func() {
 							defer wg.Done()
-							for {
-								buf := make([]byte, BYTE_SIZE)
-								numBytes, err := logs.Read(buf)
-								if numBytes == 0 {
-									continue
-								}
-								if err == io.EOF {
-									break
-								}
-								if err != nil {
-									utils.Log.Errorf("error occurred while processing logs %v", err)
-									break
-								}
-								logBuf = buf[0:numBytes]
-								printLogs(string(logBuf), name)
-							}
+							followLogs(logs, name)
 						}()
 					}
 				}
--- a/mesheryctl/internal/cli/root/system/logs_test.go
+++ b/mesheryctl/internal/cli/root/system/logs_test.go
@@ -14,7 +14,41 @@
 
 package system
 
-import "testing"
+import (
+	"io"
+	"testing"
+	"time"
+
+	"github.com/meshery/meshery/mesheryctl/pkg/utils"
+)
+
+// eofOnceReader returns one chunk of data, then (0, io.EOF) on every call after,
+// the same way a k8s pod log stream reports its end.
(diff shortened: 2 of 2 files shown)


emphor11:
@Atishyy27 The actual bug is correctly fixed, the change is focused, the extraction makes the behavior testable, and the regression test covers the original infinite-loop scenario. The only meaningful improvement is adding coverage for (n > 0, io.EOF) so the test protects the ordering that prevents the final log chunk from being lost.

suttonskate:
This is the correct fix and you found the actual root cause, not just a symptom: an `io.Reader` is explicitly allowed to return `(n > 0, io.EOF)` on the same call per the `io.Reader` contract, and the old `if numBytes == 0 { continue }` before the EOF check meant that final non-empty EOF chunk got processed but the loop never re-checked `err` afterward on subsequent zero-byte EOF reads - so it spun forever instead of breaking. Reordering to check `err == io.EOF` unconditionally (not gated on `numBytes == 0`) fixes it correctly.

Extracting `followLogs` for unit-testability was the right call, and `eofOnceReader` in the test is a good minimal reproduction of the exact contract violation this fixes - the 2-second timeout guard against a real hang in CI rather than a flaky infinite wait is a nice touch.

No changes required - this is ready as-is, and I'd call this the cleanest fix in this batch.

Atishyy27:
@emphor11 added that case, thanks. the new test drives a reader that returns the data and io.EOF on the same call, and it asserts the chunk still lands in the log output rather than only that followLogs returns.

checked it fails the other way round before trusting it: with the error checked ahead of the bytes, that same reader produces empty output. so the test pins the ordering, not just the exit.

SetupMeshkitLoggerTesting already returns the buffer it attaches to the logger, so asserting on the output needed no new plumbing. this is the same case coderabbit flagged.

emphor11 (approved):
The fix looks correct and well scoped. LGTM

ishwar170695 (approved):

Issue #21408, opened 2026-08-14
[Models] Add meaningful relationships between AWS SNS `Topic` and SQS `Queue`


hortison:
## Current Situation

AWS SNS topics cannot be visually connected to SQS queues in Meshery UI, limiting the ability to design and understand event-driven fan-out messaging architectures.

## Proposed Change

Add meaningful relationships between AWS SNS `Topic` and SQS `Queue` to enable visual fan-out messaging pattern management in Meshery UI.

## Potential Relationships

- Topic (AWS SNS) → Queue (AWS SQS)

## Help Your PR Get Merged Faster :

For easier review and approval, please provide one of the following to confirm the relationships work as expected:

- Screenshots showing your design in Meshery UI with the relationships working.
- A short video demonstrating the relationship functionality.
- After your PR is merged: Get your work recognized by publishing your design to the Meshery Catalog! 🎉

---

### General Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)

- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)

### Understanding Meshery Models

- 📚 [Models](https://docs.meshery.io/concepts/logical/models)
  - 📚 [Components](https://docs.meshery.io/concepts/logical/components)
  - 📚 [Relationships](https://docs.meshery.io/concepts/logical/components)
- 👨‍💻[Models Repository](https://github.com/meshery/meshery/tree/master/server/meshmodel)

 ### Contributing to Meshery Models
 - [Contributing Models](https://docs.meshery.io/project/contributing/contributing-models)
   - [Contributing Components](https://docs.meshery.io/project/contributing/contributing-components)
   - [Contributing Relationships](https://docs.meshery.io/project/contributing/contributing-relationships)
   - 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

iamdeepaktiwari08:
Please assign me to this issue @hortison

ishwar170695:
> Please assign me to this issue [@hortison](https://github.com/hortison)

Go for it.

PR #21460, opened 2026-08-18
feat(meshmodel): add AWS SNS to SQS relationship

iamdeepaktiwari08:
Adds a non-binding network relationship between an AWS SNS Subscription and an AWS SQS Queue for aws-sns-controller v1.10.0.

* Uses Subscription spec.endpoint as the mutated reference.
* Uses the SQS Queue ARN/URL as the mutator reference.
* Uses relationships.meshery.io/v1beta3.
* Adds only the new SNS → SQS relationship file.

 Summary by CodeRabbit

* **New Features**
  * Added enabled, non-binding network relationships between AWS SQS queues and AWS SNS subscriptions across supported model versions.
  * Supports event-driven fan-out messaging by connecting queues with SNS subscription endpoints.
  * Automatically updates connection display names and subscription endpoints for clearer, more accurate relationship configuration.


vedant21-ctr (changes requested):
The useEffect that resets menuOpen when open is false is duplicated here. Could we remove the duplicate and keep a single effect for this behavior?

vedant21-ctr:
and,
@iamdeepaktiwari08  I noticed this PR also adds the VPC network relationship definitions. Since the PR is mainly about the SNS → SQS relationship, is the VPC change intentionally part of this PR, or should it be split into a separate change?

iamdeepaktiwari08:
> And, @iamdeepaktiwari08 I noticed this PR also adds the VPC network relationship definitions. Since the PR is mainly about the SNS → SQS relationship, is the VPC change intentionally part of this PR, or should it be split into a separate change?

Yes, this was created by Mistek VPC. In another PR, this VPC part was addressed and reviewed @vedant21-ctr.

iamdeepaktiwari08:
Shall I create a new pr because here VPC mistakenly I added? @vedant21-ctr

vedant21-ctr:
Got it, thanks for clarifying! No worries. Please remove the VPC-related changes from this PR so that it stays focused on the SNS → SQS relationship. Also, please improve the grammar and spellings in the comments  and else where needed

Maanvi212006 (commented):
Does this PR address an existing issue? If so, please include the issue number in the PR description.

iamdeepaktiwari08:
I removed the VPC file from the PR, now it is only addressing the meaningful relationships between AWS SNS Topic and SQS Queue
https://github.com/meshery/meshery/issues/21408
@Maanvi212006 @vedant21-ctr

suttonskate:
Clean cross-model relationship (aws-sns-controller `Topic` → aws-sqs-controller `Queue`, non-binding/network) - both selector sides correctly leave `patch: null` since this is describing an association rather than mutating either component's spec, which is the right shape for a fan-out/messaging relationship where there's no field to patch. Validated via Kanvas import per the description.

No changes required.

ui/components/workspaces/SpacesSwitcher/WorkspaceSwitcher.tsx
ishwar170695, line 1: Do we need the `WorkspaceSwitcher.tsx` changes in this PR?

ishwar170695 (commented):
I noticed this relationship is only added under `v1.7.1`, while the existing SNS relationships are present under `v1.7.1`, `v1.8.0`, and `v1.9.0`. Should this relationship be added to the newer model versions as well?

iamdeepaktiwari08:
> I noticed this relationship is only added under `v1.7.1`, while the existing SNS relationships are present under `v1.7.1`, `v1.8.0`, and `v1.9.0`. Should this relationship be added to the newer model versions as well?

There are only two versions there: v1.7.1 and v1.8.0. I added only v1.7.1. 
Should I add this to the latest version?  @ishwar170695

ishwar170695:
> There are only two versions there: v1.7.1 and v1.8.0. I added only v1.7.1.

v1.9.0 was added 5days ago
v1.9.1 was added only 6hours ago so its understandable.

ishwar170695:
> Should I add this to the latest version? @ishwar170695

Yah ig.

iamdeepaktiwari08:
Deepaks-MacBook-Air:aws-sns-controller deepaktiwari$ ls
v1.7.1  v1.8.0
Deepaks-MacBook-Air:aws-sns-controller deepaktiwari$ pwd
/Users/deepaktiwari/meshery/models/aws-sns-controller
after fetched latest change in the repo
I can see only these versions. @ishwar170695

ishwar170695:
@iamdeepaktiwari08 Maybe pull the latest master?
image

iamdeepaktiwari08:
Thanks, I got it apart from v1.7.1. I am going to add SNS relationships to the remaining versions. @ishwar170695

iamdeepaktiwari08:
Thanks for the review. I’ve added the SNS-to-SQS relationships for the remaining versions as discussed. Please review the latest changes when you get a chance. Thank you @ishwar170695 @vedant21-ctr @suttonskate @Maanvi212006

ui/components/workspaces/SpacesSwitcher/WorkspaceSwitcher.tsx
ishwar170695: What about this file?

iamdeepaktiwari08: The WorkspaceSwitcher.tsx changes were not part of my original changes for this PR. Khusham has already addressed the related UI changes in the  commits.

iamdeepaktiwari08:
Hi, thanks for reviewing the PR. I’ve addressed the feedback, and the checks are passing. Could someone with write access please take a final look and approve it when possible? Thank you

hiyach28 (changes requested):
please remove the[ UI commits](ui/components/workspaces/SpacesSwitcher/WorkspaceSwitcher.tsx) from this branch so this PR *only* contains the relationship changes.

vedant21-ctr (changes requested):
Thanks for addressing the previous feedback and adding the SNS → SQS relationship across the remaining model versions.

One blocker remains: the branch still contains the unrelated UI commits/changes (workspace switcher and Kanvas icon changes). Please remove those changes so this PR contains only the SNS → SQS relationship work.

Once the branch is cleaned up, please also confirm the relationship has been validated for the supported SNS model versions.

iamdeepaktiwari08:
Thanks for the feedback. I’ve removed the unrelated UI/workspace and Kanvas changes from the branch, so the PR now contains only the SNS → SQS relationship work.

I had already validated the relationship for the supported SNS model versions before this update, and the changes cover the required versions. @vedant21-ctr @hiyach28

hiyach28 (approved):
LGTM

vedant21-ctr (approved):
LGTM!

ishwar170695 (approved):

ishwar170695:
The PR LGTM.

But I am not exactly sure what to do here cuz as time passes new version folders are being added and thus their are new folders which will not be having the relationship.
@banana-three-join

iamdeepaktiwari08:
Hey @ishwar170695 i added relationships in all versions of SNS.

banana-three-join (commented):
So, this is my first time reviewing this particular area of the codebase so the following findings can be taken with a grain of salt; nevertheless, from what I researched, these seem like valid concerns in regards of the changes in implemented in this PR:

**The Topic and the Queue don't reference each other**

In SNS/SQS, whether a topic delivers to a queue isn't a property of either
resource. It's determined by a `Subscription`, which holds `spec.topicARN`
(the topic) and `spec.endpoint` (the queue's ARN).

Checking the component schemas in this repo confirms it: `Topic.spec` has no field naming a queue, and `Queue.spec` has no field naming a topic. 

(`Topic.spec.sqsSuccessFeedbackRoleARN` looks related but is an IAM role for delivery-status logging, not a queue reference.)

That's why this definition ends up with `patch: null` and `match: {}` on both sides — there is no field to read and none to write.

**Consequence: the definition is inert to the evaluation engine**

With `patch: null` on both selector sides, `matchingMutators` in `server/policies/eval_rules.go` returns false immediately:

``go
if fromClause.RelationshipDefinitionSelectorsPatch == nil ||
   toClause.RelationshipDefinitionSelectorsPatch == nil {
 	return false
}
```

So `EdgeNonBindingPolicy.IdentifyRelationship` never proposes a pair, and `patchMutatorsAction` never patches anything (same nil-patch guard).

The definition's only effect is to make a hand-drawn Topic→Queue edge *feasible* in Kanvas — for any topic and any queue — because `isRelationshipFeasible` in `server/policies/feasibility.go` checks component kind and model name only, with no patch requirement.

That also means the validation in the description doesn't demonstrate what it looks like it does. "Automatically rendered" and "0 errors" are both what you'd expect from a definition the engine never identifies; it would behave identically for a topic and queue with no subscription between them.

**Suggested change: anchor on `Subscription`**

This model already establishes that pattern:

- `edge-binding-network-bqddq.json` — `Topic ↔ Subscription` via `spec.topicARN`
- `edge-non-binding-network-qwafj.json` — `PlatformEndpoint ↔ Subscription` via `spec.endpoint`

A `PlatformEndpoint` is a push-notification subscriber — structurally the same slot an SQS queue occupies. The missing piece in the corpus is the SQS equivalent:

```json
"from": [ { "kind": "Queue", "model": { "name": "aws-sqs-controller", "...": "..." },
             "patch": { "mutatedRef": [["displayName"]],
                        "patchStrategy": "replace" } } ],
"to":   [ { "kind": "Subscription", "model": { "name": "aws-sns-controller", "...": "..." },
             "patch": { "mutatorRef": [["configuration","spec","endpoint"]],
                        "patchStrategy": "replace" } } ]
```

Note this follows `bqddq`'s `["displayName"]` sink rather than `qwafj`'s. `qwafj` writes to `configuration.status.ackResourceMetadata.arn`, and none of these components have a `status` block in their schema — that path doesn't resolve, so it's not worth copying.

**Version coverage**

`models/aws-sns-controller/` has five versions on master — v1.7.1, v1.8.0, v1.9.0, v1.9.1, v1.10.0 — all carrying the same four relationship files. This PR adds to three, skipping v1.7.1 and v1.10.0, including the newest. Whether the relationship exists would depend on which model version a user happens to have registered.

***Description mismatch***
The body says the relationship was added for v1.7.1, which is the one version the diff doesn't touch.

***Nit***
`model.registrant.kind` is `"github"` here; the four existing files in this model use `""` at that level.

iamdeepaktiwari08:
@banana-three-join 
Thanks for the detailed review. I agree with the core concern here.
I was initially validating the relationship from the Kanvas rendering side, but after looking at the relationship evaluation flow and the existing SNS Topic → Subscription / PlatformEndpoint → Subscription definitions, I see that the current Topic → Queue definition with patch: null doesn’t actually establish the SNS/SQS association through the Subscription resource.

I’ll rework this to anchor the relationship on Subscription, using spec.topicARN for the Topic side and spec.endpoint for the Queue side, and I’ll verify the appropriate Queue ARN field against the SQS component schema rather than copying the existing PlatformEndpoint path.

I’ll also correct the version coverage/description mismatch and review the supported SNS controller versions before updating the relationship files.

iamdeepaktiwari08:
Hey @banana-three-join

Adds an AWS SNS-to-SQS relationship through the SNS Subscription.
Updated relationship to Queue → Subscription using `Subscription.spec.endpoint`.
Added support for SNS controller versions v1.7.1, v1.8.0, v1.9.0, and v1.9.1.
Validated all relationship files.

Junnygram:
There are a few configuration issues in this file that need to be addressed:

**1. Reversed `mutatorRef` and `mutatedRef` Directions**
Currently, the configuration directions are swapped, which will cause the visual arrows on the canvas to point backwards and the patch to fail. 
In a typical fan-out setup, the data flows from the `Subscription` down to the `Queue`. Therefore, the Queue must provide its ARN/URL (`mutator`), and the Subscription must receive it into its endpoint (`mutated`).
*   **Current:** `from` (Queue) has `mutatedRef`, and `to` (Subscription) has `mutatorRef`.
*   **Required Fix:** Completely swap the `from` and `to` blocks so that **`from`** is the `Subscription` (using `mutatedRef`) and **`to`** is the `Queue` (using `mutatorRef`). This is required for drag-and-drop orchestration to function correctly.

**2. Invalid Patch Path for Queue**
For the `Queue` component, the patch references `[["displayName"]]`.
*   AWS Controllers for Kubernetes (ACK) schemas typically require the Queue's ARN or URL for an endpoint, not the displayName.
*   **Required Fix:** Please update this to point to the correct field, which is likely `[["metadata", "name"]]` (for design-time patching) or `[["status", "ackResourceMetadata", "arn"]]`.

**3. Missing Latest Versions**
I noticed you added the relationship for versions `v1.7.1` through `v1.9.1`.
*   **Required Fix:** Please ensure this relationship is only propagated to the **latest** version of the model, which is actually **`v1.10.0`**. It's only the latest version that is evaluated by the engine, whereas the older versions are superseded. 

**4. Schema Version Check**
The file declares `"schemaVersion": "relationships.meshery.io/v1beta2"`.
*   **Note:** We are prioritizing `v1beta1` (or `v1beta3` for `via` intermediaries) as the canonical standard. Please make sure not to just copy legacy `v1beta2` structures.

Here is a visual representation of how the data flows (from SNS Topic ➡️ Subscription ➡️ SQS Queue) to help clarify the direction for Point 1!

Screenshot 2026-09-01 at 11 11 23



**Note on infrastructure provisioning flow vs data flow:** 
In reality, when provisioning this infrastructure, we create the `SNS Topic` first, then the `SQS Queue`, and *finally* we create the `Subscription` as a bridge to link them. Because the Subscription connects the two, it must know *which* Queue to target, which is exactly why the Subscription needs to extract the Queue's ARN (making the Queue the `mutator` and the Subscription the `mutated`).

Junnygram:
Small correction on my earlier comment, the latest published schema version is actually v1beta3. So please upgrade the schemaVersion to relationships.meshery.io/v1beta3. All relationship definitions should use the latest published schema version.

models/aws-sns-controller/v1.10.0/v1.0.0/relationships/edge-non-binding-network-sns-to-sqs.json
```diff
@@ -0,0 +1,100 @@
...
+              "name": "aws-sqs-controller",
+              "registrant": {
+                "kind": "github"
+              },
+              "version": ""
+            },
+            "patch": {
+              "mutatorRef": [
+                [
+                  "metadata",
```
Maanvi212006, line 82: Tested `configurationForComponentAtPath(["metadata","name"], ...)` against the real `Queue.json`, it returns nil, since that path hits the component envelope, not the K8s resource. Likely needs `["configuration", "metadata", "name"]` instead, per convention elsewhere (e.g. `edge-binding-permission-vcrwt.json`).

Maanvi212006 (changes requested):

iamdeepaktiwari08:
@Maanvi212006 
Thanks for catching this. I updated the mutatorRef path to ["configuration", "metadata", "name"] and pushed the fix.

models/aws-sns-controller/v1.10.0/v1.0.0/relationships/edge-non-binding-network-sns-to-sqs.json
iamdeepaktiwari08: issue resolved

Maanvi212006 (dismissed):

Junnygram:
Thanks @iamdeepaktiwari08 for updating the direction, patching endpoint, and upgrading the schemaVersion to relationships.meshery.io/v1beta3 as discussed.

A couple of quick follow-up items before this can be merged:

1. Target Model Version:
Since my earlier comment, master has updated with new model versions (v1.10.1 and v1.10.2). Following the principle that only the latest model version is evaluated by the engine, please rebase with master and move this relationship definition into v1.10.2:
`models/aws-sns-controller/v1.10.2/v1.0.0/relationships/edge-non-binding-network-sns-to-sqs.json`

2. Schema Hygiene:
Under canonical v1beta3, please clean up the unused placeholder fields (`id: "00000000-0000-0000-0000-000000000000"`, `"match": {}`, and `"match_strategy_matrix": null`).

iamdeepaktiwari08:
@Junnygram  as per feedback i made changes please take a look whenever you free
.

PR #21464, opened 2026-08-18
[UI] Add E2E tests for design import flow (PR #21105)

NSTKrishna:
## Description

Adds Playwright E2E tests covering the two issues fixed in meshery/meshery#21105.

### Background

PR #21105 fixed two bugs in the Design Import flow (`POST /api/pattern/import`):

1. **Wire-format regression** – the file-upload request body was serialising the file name as `file_name` (snake_case) instead of `fileName` (camelCase) as required by the server schema and the schemas-generated `ImportDesignApiArg` type.
2. **Duplicate-submission** – rapid clicks on the Import button could fire the POST more than once before the `isSubmittingRef` guard was added to `ImportDesignModal`.

### New Files

| File | Purpose |
|---|---|
| `ui/tests/e2e/design_import.testmap.ts` | Traceability map (same shape as `connections.testmap.ts`) — defines `DESIGN_IMPORT_CASES`, the `Design Import` Allure `testGroup` label, and annotation/tag helpers |
| `ui/tests/e2e/design_import.spec.ts` | Four Playwright tests covering the full import flow |

### Test Cases

| ID | Title | What is asserted |
|---|---|---|
| TC-DI-001 | Open Import Design modal | Modal opens from toolbar; both radio options visible; Cancel dismisses cleanly |
| TC-DI-002 | URL import wire contract | POST body contains `url` + `name`; **no `file_name`** (snake_case regression guard) |
| TC-DI-003 | File upload wire contract | POST body contains `fileName` (camelCase); **no `file_name`** – primary regression guard for #21105 |
| TC-DI-004 | Duplicate-submission prevention | Button becomes `disabled` + shows "Importing…" in-flight; only **one** POST fired |

### Supporting Change

- **`ui/tests/e2e/pages/ProviderSelectionPage.js`**:
  Updated `getProviderMenuItem(providerName)` to match the provider menu item rendered as `${providerName} More`. This change is necessary to resolve a Playwright strict-mode violation encountered when running the E2E suite locally. This is strictly a supporting E2E/test-infrastructure change and does not alter the Design Import flow itself.

### Design Decisions

- All `POST /api/pattern/import` requests are **stubbed via `page.route()`** — tests are self-contained and run without a real backend.
- TC-DI-004 uses a 300 ms stub delay to hold the in-flight window open long enough to assert the disabled state.
- TC-DI-002/003 self-skip (not fail) when the import request is never fired, mirroring the connections suite's infra-less degradation contract.
- The `180_000 ms` describe timeout matches the connections spec rationale (slow CI + two sequential navigation waits).

---

> **Related**: meshery/meshery#21105

## Linked Issues

closes #21105

## Screenshots / Recordings

N/A – network-layer and UI-state assertions; no visual diff.

---

**Signed-off-by**: NSTKrishna <krishnagehlot936@gmail.com>

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for the Design Import flow.
  * Verified import modal controls, URL and file-upload data, success notifications, and request handling.
  * Confirmed the Import button is disabled and shows “Importing…” during submission to prevent duplicates.
  * Added traceability metadata for tracked test cases.

* **Bug Fixes**
  * Improved provider selection to recognize menu entries labeled with the provider name followed by “More.”


ui/tests/e2e/pages/ProviderSelectionPage.js
```diff
@@ -5,7 +5,7 @@ export class ProviderSelectionPage {
   }
 
   getProviderMenuItem(providerName) {
-    return this.page.getByRole('menuitem', { name: providerName });
+    return this.page.getByRole('menuitem', { name: `${providerName} More` });
```
NSTKrishna: changes in providerSelectionPage.js are necessary to fix a strict mode violation that only happens when running tests locally in your environment.

vedant21-ctr (changes requested):
@NSTKrishna  

1: waitForImportRequest still says it returns Promise<Request>, but the Request import from Playwright got removed. So right now that type is pointing to the browser's built-in Request, not Playwright's — they're not the same thing. It's not breaking the tests since Playwright doesn't strictly type-check at runtime, but it's still incorrect. Could you add import { Request } from '@playwright/test'; back?
2: The comment at the top still says the schema-anchoring came from #21105 — that was actually done in #21114. And it still says "100ms delay" even though the code uses 300ms now. Just needs the wording updated to match.

Everything else looks solid.

ui/tests/e2e/design_import.spec.ts
```diff
@@ -0,0 +1,315 @@
+import { expect, Page } from '@playwright/test';
+import * as allure from 'allure-js-commons';
```
hiyach28, line 2: design_import.spec.ts and design_import.testmap.ts import allure-js-commons directly, but it isn't declared in ui/package.json; only allure-playwright is. This currently relies on the transitive dependency being hoisted by npm. Could we maybe either import Allure through allure-playwright or explicitly declare allure-js-commons as a dev dependency?

ui/tests/e2e/design_import.testmap.ts
```diff
@@ -0,0 +1,130 @@
...
+export type DesignImportCaseKey = keyof typeof DESIGN_IMPORT_CASES;
+
+/** Playwright tags for a tracked case — grep-safe and surfaced as Allure tag labels. */
+export function designImportTags(key: DesignImportCaseKey): string[] {
+  const c = DESIGN_IMPORT_CASES[key];
+  return ['@design-import', `@${c.testId}`, '@cut:ui-design-import-modal', '@client:ui'];
+}
+
+/** Untracked tag (no sheet row yet). */
+export function designImportTagsUntracked(component: string): string[] {
```
hiyach28: We have removed the `annotateDesignImportCaseUntracked` and `designImportTagsUntracked` imports/usages from design_import.spec.ts, but their function definitions still remains. Could you verify whether these functions are still referenced elsewhere in the test suite? If they have no remaining callers, i think it would be cleaner to remove them?

suttonskate:
This is thorough, well-targeted regression coverage. A few things done right:

- TC-DI-003 asserts on the raw wire body for `fileName` **and** the absence of `file_name` - that's the actual regression guard for #21105, not just a happy-path check, and it'll catch the exact defect (snake_case creeping back in) rather than a generic "import succeeded" assertion that could pass even with the wrong field name.
- TC-DI-004's duplicate-submission test uses a 300ms-delayed stub specifically to hold the in-flight window open long enough to assert the disabled state, then asserts `importCallCount === 1` - that's testing the actual guard behavior, not just that the button visually looks disabled.
- Stubbing `POST /api/pattern/import` via `page.route()` keeps these deterministic and not dependent on a real backend import succeeding.

One small thing worth a look, unrelated to the design-import tests themselves: the diff also changes `ProviderSelectionPage.js`'s `getProviderMenuItem` to match `${providerName} More` instead of `providerName` alone. If that's an unrelated fix needed to unbreak another existing test on this branch, it'd be cleaner called out explicitly in the PR description (or split out) so a reviewer doesn't have to guess why a design-import PR touches provider-selection page objects.

No changes required on the design-import tests themselves - solid coverage. Just flag the ProviderSelectionPage change so it doesn't slip through unexplained.

ui/tests/e2e/design_import.spec.ts
```diff
@@ -0,0 +1,315 @@
...
+  await expect(page.getByRole('dialog')).toBeVisible({ timeout: 10_000 });
+}
+
+/**
+ * Wait for a POST to /api/pattern/import.
+ * The promise rejects/times-out if no request fires within 30 s,
+ * which fails the test loudly instead of silently skipping.
+ */
+function waitForImportRequest(page: Page): Promise<Request> {
+  return page.waitForRequest(
```
Bharath314, line 83: Can you fix the typescript error on this line?

ui/tests/e2e/design_import.spec.ts
```diff
@@ -0,0 +1,315 @@
...
+      // Capture the outgoing request BEFORE clicking Import.
+      const importReqPromise = waitForImportRequest(page);
+
+      await dialog.getByRole('button', { name: 'Import' }).click();
+
+      // Collect the request — if it never fires the test fails with a clear timeout message.
+      const importReq = await importReqPromise;
+
+      // ── Wire contract assertions ──────────────────────────────────────────
+      const body = importReq.postDataJSON() as Record<string, unknown>;
```
Bharath314: typescript error on this line as well.

ui/tests/e2e/design_import.spec.ts
```diff
@@ -0,0 +1,315 @@
...
+
+      await dialog.getByRole('button', { name: 'Import' }).click();
+
+      // Collect the request — if it never fires the test fails with a clear timeout message.
+      const importReq = await importReqPromise;
+
+      // ── Wire contract: fileName must be camelCase ─────────────────────────
+      // The request is sent as multipart/form-data or JSON depending on the
+      // RJSF widget. Inspect whichever form the body takes.
+      const rawBody = importReq.postData() ?? '';
```
Bharath314, line 254: typescript error on this line as well.

ui/tests/e2e/design_import.spec.ts
Bharath314: I second this comment.

Bharath314 (changes requested):

ishwar170695 (commented):
Could we confirm whether the strict-mode violation occurs consistently in the standard E2E setup or CI, or only in a specific local environment? Since this changes a shared page-object selector used across multiple tests, it’d be helpful to understand why multiple matches are being rendered and make sure this isn't an environment-specific workaround.

ui/tests/e2e/design_import.spec.ts
```diff
@@ -0,0 +1,321 @@
...
+      const importReq = await importReqPromise;
+
+      // ── Wire contract assertions ──────────────────────────────────────────
+      // postDataJSON() returns null when the content-type is not JSON; fail
+      // fast with a clear message rather than letting assertions run on null.
+      const rawJson = importReq.postDataJSON();
+      if (rawJson === null) {
+        throw new Error('Import request body was not JSON — cannot inspect wire contract fields');
+      }
+      const body = rawJson as Record<string, unknown>;
```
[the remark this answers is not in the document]
NSTKrishna: Addressed in commit 017dd7f — `postDataJSON()` now gets a null-check before the `as` cast. If the body isn't JSON the test fails with a clear error message instead of silently running assertions on null.

```typescript
const rawJson = importReq.postDataJSON();
if (rawJson === null) {
  throw new Error('Import request body was not JSON — cannot inspect wire contract fields');
}
const body = rawJson as Record<string, unknown>;
```

ui/tests/e2e/pages/ProviderSelectionPage.js
```diff
@@ -5,7 +5,11 @@ export class ProviderSelectionPage {
   }
 
   getProviderMenuItem(providerName) {
-    return this.page.getByRole('menuitem', { name: providerName });
+    // The MUI menu item renders as "${providerName}" in CI and as
+    // "${providerName} More" locally (the suffix comes from an additional icon
+    // button inside the menu item). Using a regex that matches either form
+    // avoids a Playwright strict-mode violation in both environments.
+    return this.page.getByRole('menuitem', { name: new RegExp(`^${providerName}(\\s+More)?$`) });
```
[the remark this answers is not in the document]
NSTKrishna: Good catch in the general case, but not a real risk here — the only two callers pass `'Local'` and `'Meshery'` (defined in `PROVIDERS` in `local.setup.js` / `remote.setup.js`), neither of which contains regex metacharacters. Adding an escape helper for two known-safe string literals would be over-engineering. If a provider name with special characters is ever added, the regex will break loudly (test failure), making it easy to catch and fix at that point.

ui/tests/e2e/design_import.spec.ts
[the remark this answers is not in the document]
NSTKrishna: Good catch! Playwright `postDataJSON()` does indeed attempt to parse `application/x-www-form-urlencoded` payloads, so checking `rawJson === null` isn't a strict guarantee that the payload was JSON.

Added an explicit assertion for the `content-type` header to ensure it matches `^application/json(?:;|$)` before calling `postDataJSON()`. Addressed in commit 9df30dc (or whatever the latest commit hash is).

Issue #21078, opened 2026-08-02
[Test Plan] Test #3: Install mesheryctl and select the platform to deploy on prompt


yi-nuo426:
Automate and verify test case **#3** from the [Meshery Test Plan](https://docs.google.com/spreadsheets/d/13Ir4gfaKoAX9r8qYjAFFl_U9ntke4X5ndREY1T7bnVs/edit?gid=838298230#gid=838298230) ("Latest" tab), currently marked **Absent (Help Wanted)**.

- **Test #:** 3
- **Group:** Installation
- **Component under test:** Installer / Bash Script

**Scenario:** Install mesheryctl and select the platform to deploy, chosen at an interactive prompt.

**Action:**
```sh
curl -L https://git.io/meshery | bash -
```

**Expected outcome:** Shows a prompt for the user to enter the platform from a list of available ones; starts if a valid platform is chosen, terminates if invalid.

---
This test case is not yet automated. This issue tracks implementing it (e.g. as a BATS case under `mesheryctl/tests/e2e/`, or an install-script test) and updating the Test Plan status once it lands. Help wanted.

Om-Beast:
Hi @yi-nuo426, I'd like to work on this issue.

I'll automate Test #3 by adding an end-to-end test for the interactive installer flow, covering both valid and invalid platform selections while following the existing test structure and conventions in the repository.

Looking forward to submitting a PR. Thanks!

PR #21485, opened 2026-08-18
[Test] Automate mesheryctl platform selection prompt

Om-Beast:
## Description

Automates Test #3 from #21078 by covering the interactive platform-selection flow in the Meshery installer.

## What changed

- Added Bats coverage under `mesheryctl/tests/e2e/010-installation/`.
- Verifies that the installer prompts for a deployment platform.
- Verifies a valid `docker` selection is accepted.
- Verifies an invalid selection is rejected and the installer accepts a subsequent valid `kubernetes` selection.
- Isolates the installer prompt block before installation side effects.
- Uses `expect` to exercise the installer's interactive prompt deterministically.

## Validation

Local test result:

```text
1..2
ok 1 installer accepts docker platform selection
ok 2 installer rejects invalid platform and accepts kubernetes

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for the Meshery installation platform prompt.
  * Verified successful Docker and Kubernetes selections.
  * Verified invalid platform input is rejected.

### DIFF
--- a/mesheryctl/tests/e2e/010-installation/00-cli-install-platform-prompt.bats
+++ b/mesheryctl/tests/e2e/010-installation/00-cli-install-platform-prompt.bats
@@ -4,9 +4,13 @@ setup() {
     INSTALLER="$BATS_TEST_TMPDIR/meshery-install.sh"
     HARNESS="$BATS_TEST_TMPDIR/platform-prompt.sh"
 
-    curl -fsSL https://meshery.io/install -o "$INSTALLER"
+    curl -fsSL --connect-timeout 10 --max-time 60 https://meshery.io/install -o "$INSTALLER"
 
-    awk '/^####### COMMON FUNCTIONS/{exit} {print}' "$INSTALLER" > "$HARNESS"
+    awk '
+    /^####### COMMON FUNCTIONS/ { found = 1; exit }
+    { print }
+    END { exit !found }
+' "$INSTALLER" > "$HARNESS" || return 1
 
     cat >> "$HARNESS" <<'EOF'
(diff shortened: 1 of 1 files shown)


Om-Beast:
Thanks for the review. I’ve addressed the previously reported stability concerns and pushed the fixes in commit 15e6fb2.

The updated test was revalidated locally with:

1..2
ok 1 installer accepts docker platform selection
ok 2 installer rejects invalid platform and accepts kubernetes

There are no remaining actionable review comments on the updated commit.

suttonskate:
The isolation approach is nice - trimming the installer script down to just the platform-selection block via the `awk`/marker-comment split (`awk '/^####### COMMON FUNCTIONS/ ...'`) means you're testing the actual prompt logic without needing a full install to run in CI, and the invalid-then-valid-selection case is exactly the edge that's easy to skip when writing this kind of test.

One thing worth flagging: `setup()` does `curl -fsSL https://meshery.io/install` at test run time. That makes the test dependent on the live production installer script at whatever moment CI happens to run - if that script's marker comment (`####### COMMON FUNCTIONS`) or its prompt copy ("Enter a platform to deploy Meshery") ever changes, this test breaks with no code change in this repo, and conversely a change to that script that breaks the CLI could ship without this test catching it if CI ran against a stale fetch beforehand. Worth checking whether the repo already has a policy for testing against the live installer elsewhere (if so, this is consistent and fine) - if not, pinning to a local copy of the installer script under test would make this deterministic and catch regressions to the script itself.

Not a blocker given comment/review activity already exists on this PR, but I'd want that question answered before merge: is fetching the live installer at test time an accepted pattern here?

PR #21486, opened 2026-08-18
fix(mesheryctl): support model init templates in model generate

### LINKED ISSUE #21402: [mesheryctl] UX and Architectural issues with model generate flag and -t template logic

## Bug Description

---

#### Current Behavior

#### Desired Behavior

---

#### Current Behavior

`type Model struct {`
`// ...`

#### Desired Behavior

### Environment

- Host OS: Mac Linux Windows
- Platform: Docker Kubernetes

---

hiyach28:
### Description

This PR fixes #21402 

Fixes the template parsing issue in mesheryctl model generate where a model.json generated by mesheryctl model init could not be passed directly to the -t flag.
The model init command produces a nested `ModelDefinition`, while model generate previously expected the flattened `v1beta1.Model` structure. This resulted in unmarshalling errors for fields such as registrant and category.

### Changes

- Added support for parsing the nested `modelv1beta2.ModelDefinition` structure produced by mesheryctl model init.
- Explicitly map the nested model fields to the flattened structure expected by the model generation pipeline.
- Preserve support for existing flat template files through a fallback unmarshalling path.
- Added a real model init template fixture for regression testing.
- Added test coverage for:
    - existing flat templates
    - nested model init templates
    - default model registration
    - `--skip-registration`

### Validation

- go test -v -short -run TestModelGenerate ./internal/cli/root/model -timeout 30s
- All TestModelGenerate cases pass.
- git diff --check passes.

## Summary by CodeRabbit

* **New Features**
  * Added support for importing standard model definitions with identity, registrant, category, subcategory, and metadata.
  * Preserved compatibility with legacy model formats through automatic fallback handling.
  * Added default and minimal model templates to streamline model creation.

* **Tests**
  * Improved validation coverage for model import requests and generated model details.


suttonskate:
This closes a real gap - `model generate -t` silently choking on the exact file `model init` produces is the kind of papercut that makes the two commands feel disconnected even though they're meant to compose. Adding a fixture (`model_template.json`) that mirrors the nested `ModelDefinition` shape is the right way to pin the parsing contract so it doesn't regress silently later.

Question before merge: does the parser fall back cleanly to the old flattened `v1beta1.Model` shape for existing model.json files created before this change, or does detecting "nested vs flat" rely on a field that's only present in the new template? If there's a large body of model.json files out there already using the flattened shape, I'd want to see a test asserting both shapes still parse - if that's already covered and I missed it in the diff, ignore this.

One question I'd want answered before merge, otherwise this looks solid.

hiyach28:
Hi @suttonskate thank you for reviewing! 

Yes, it falls back cleanly, and this exact scenario is covered in the tests.
The fallback detection relies on the `name` field. The new nested template (`modelv1beta2.ModelDefinition`) uses `json:"name"`, whereas the legacy flat template (`v1beta1.Model`) used `json:"model"`. 
The logic specifically checks for this:
```go
var modelDef modelv1beta2.ModelDefinition
if err := encoding.Unmarshal(data, &modelDef); err == nil && modelDef.Name != "" {
   //Nested structure matched(new model init template)
} else {
    //Fallback: Legacy flat template; unmarshal directly into the flat struct
    err := encoding.Unmarshal(data, &importRequest.ImportBody.Model)
}

**Review by ishwar170695 — COMMENTED** [review]
Could we use `schemaVersion` to distinguish the nested template from the legacy flat template instead of relying on `Name`? It seems safer since the two formats could potentially overlap on fields.

Also, a test for a legacy template with `registrant`/`category `omitted? That would make sure it still takes the legacy fallback path.

**Comment by hiyach28** [comment]
@ishwar170695 thank you for your review!

i agree that schemaVersion is a better discriminator than relying on Name being populated, since it explicitly identifies the nested schema rather than depending on a field that could overlap between the two representations.

the legacy-template case is also worth covering with a minimal fixture where fields like registrant and category are omitted. that should give us confidence that the fallback continues to work for sparse legacy templates as well.

i'll make those adjustments and add the regression coverage.

File: mesheryctl/internal/cli/root/model/generate_test.go [hunk 1]
```diff
@@ -62,12 +67,57 @@ func TestModelGenerate(t *testing.T) {
...
+					t.Fatalf("Failed to parse request body: %v", err)
+				}
+
+				if importReq.ImportBody.Model.Model != "minimal-model" {
+					t.Errorf("Expected model 'minimal-model', got '%s'", importReq.ImportBody.Model.Model)
+				}
+				if importReq.ImportBody.Model.ModelDisplayName != "Minimal Model" {
+					t.Errorf("Expected displayName 'Minimal Model', got '%s'", importReq.ImportBody.Model.ModelDisplayName)
+				}
+			},
```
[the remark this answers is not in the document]
hiyach28: Resolved them through commit: 122630a

hiyach28:
I’ve updated the nested-template discriminator to use schemaVersion instead of relying on Name, and added coverage for a minimal legacy template with the optional registrant/category fields omitted. I also updated the test to validate the actual POST payload so the fallback mapping is explicitly verified rather than only checking that the command succeeds.

The targeted TestModelGenerate suite passes, and the changes remain isolated to this issue only.

mesheryctl/internal/cli/root/model/import.go
```diff
@@ -166,9 +169,47 @@ func registerModel(data []byte, componentData []byte, relationshipData []byte, f
 		importRequest.ImportBody.ModelFile = data
 	default:
 		if data != nil {
-			err := encoding.Unmarshal(data, &importRequest.ImportBody.Model)
-			if err != nil {
-				return utils.ErrUnmarshal(err)
+			var modelDef modelv1beta2.ModelDefinition
+			if err := encoding.Unmarshal(data, &modelDef); err == nil && modelDef.SchemaVersion != "" {
```
rishiraj38: `SchemaVersion != ""` alone is a weak sentinel. A legacy template with a stray `schemaVersion` field might silently take the wrong path. Consider checking the actual version prefix, for example, `strings.HasPrefix(modelDef.SchemaVersion, "models.meshery.io/")`, or validating at least one nested object field such as `modelDef.Category.Name != ""`.

mesheryctl/internal/cli/root/model/generate_test.go
```diff
@@ -62,11 +115,104 @@ func TestModelGenerate(t *testing.T) {
...
+		},
+		{
+			Name:             "model generate: from URL with default registration",
+			Args:             []string{"generate", "--file", "https://github.com/cert-manager/cert-manager/releases/download/v1.13.0/cert-manager.crds.yaml", "--template", filepath.Join(fixturesDir, "templates", "template.json")},
 			URL:              apiURL,
 			Fixture:          "generate.api.ok.response.golden",
 			ExpectedResponse: "generate.dir.skipped.output.golden",
 			HttpCode:         200,
+			ValidateRequest: func(req *http.Request, t *testing.T) {
+				bodyBytes, _ := io.ReadAll(req.Body)
```
rishiraj38: The other `ValidateRequest` closures correctly use `t.Fatalf` for this error. This one swallows it, which is inconsistent and might hide a real failure. Use `t.Fatalf` here as well, like the others.

mesheryctl/internal/cli/root/model/import.go
hiyach28: Thank you for pointing it out! Resolved it,and will take care next time onwards!

mesheryctl/internal/cli/root/model/generate_test.go
hiyach28: Thank you for pointing it out! have resolved it and will remember this!

Issue #21490, opened 2026-08-18
12 kubernetes relationship definitions patch paths that do not exist in their component schemas


leecalcote:
### Current Behavior

The relationship schema-conformance guard added for #21482 (server/policies/relationship_schema_conformance_test.go) pins 12 pre-existing kubernetes definitions whose `configuration.`-rooted mutation paths do not resolve against the component JSON schemas shipped in the same model-version directory. Known members include edge-non-binding-reference-aatvf (ServiceAccount `spec.signerName`), -djklo (Node `spec.nodeName`), -gbkch (ClusterRoleBinding `spec.roleRef.name`), -keceb (RoleBinding `spec.roleRef.name`), -lnrwq (Node `spec.nodeSelector`), -mszif (PriorityClass); the authoritative, current enumeration is the test's allowlist.

Each is the same defect class as the fixed Ingress backend definition: forming the relationship patches fields the component schema does not have, producing configuration that is invalid against the schema it ships beside.

### Desired Behavior

Each allowlisted definition is repaired (correct path for the intended semantic, or the pair removed when no valid sink exists) and removed from the allowlist. Because the in-tree corpus is regenerated nightly from the Meshery Integrations spreadsheet, every repair must land in the corresponding sheet row as well (same workflow as the row-446 fix documented on #21482) or the nightly `registry generate` reintroduces it.

### Context

Found by review during #21482. The guard fails on any new definition with a non-resolving path; the allowlist exists solely so it lands green while these are repaired, and each entry cites this issue.

### Environment
- meshery/meshery, models/kubernetes/v1.37.0-rc.0

leecalcote:
Scale context from the review pass that produced the guard: running the same resolver across the newest version directory of all 476 models in models/ checks 1,688 configuration-rooted mutation paths and finds 349 unresolvable - roughly 21% of the corpus carries this defect class. The 12 entries this issue records are only the kubernetes + cert-manager slice the guard currently gates. Widening the guard's model scope should proceed together with (or after) sheet-side repairs, since a corpus-wide allowlist would be ~349 lines of pinned defects.

AyushSharma0209:
Hi I was goning through the files of this intended issue...While analyzing the file edge-non-binding-reference-gbkch.json (ClusterRoleBinding), I noticed an unusual design layout that I'd like to get some consensus.


[...]

Issue #21489, opened 2026-08-18
[Server] schemaDefaultAtPath cannot traverse allOf, so component schema defaults never resolve for Kubernetes kinds


leecalcote:
### Current Behavior

`schemaDefaultAtPath` (server/policies/eval_rules.go) walks only `cursor["properties"][segment]` - no `allOf`/`oneOf`/`anyOf` flattening and no array-index or `_` segment handling. Every Kubernetes component schema wraps `spec` in `allOf` (in models/kubernetes/v1.37.0-rc.0/v1.0.0/components: 57 components have `properties.spec.allOf`, 0 have `properties.spec.properties`), so for any Kubernetes component the walk dead-ends on the first hop below `spec` and schema defaults silently never resolve during relationship evaluation.

### Why it is filed rather than fixed inline

Surfaced during review of the Ingress/cert-manager relationship coverage branch (#21482 follow-on), whose new conformance test implements the correct allOf-flattening traversal in the same package (server/policies/relationship_schema_conformance_test.go) - the fix is mechanically small (share that traversal), but it is behavioral: schema defaults would begin resolving for the first time across the whole corpus, changing evaluation output in ways that need their own test coverage and review, not a ride-along in a models PR.

### Desired Behavior

`schemaDefaultAtPath` flattens `allOf` wrappers (and treats numeric/`_` segments as array-item descents) using the same traversal the conformance test uses, with evaluation-level tests pinning which defaults start applying.

### Environment
- meshery/meshery master, server/policies/eval_rules.go

Myself-Praveen:
Hi @leecalcote, I see you assigned this to yourself, but I'd love to take a crack at it if you haven't started yet!

miacycle:
@Myself-Praveen see the PR above? Review it.

Myself-Praveen:
Done reviewing! @miacycle @leecalcote

I verified the relevant policy tests locally, and they pass successfully. The new schema conformance checks cover nested allOf and array-descent behavior as required.

The fix looks correct and clean for the issue described.

Non-blocking note: schemaAlternatives currently has no cycle guard, so a self-referential/circular schema could theoretically recurse indefinitely. Adding a visited-set guard would make it more robust if circular references ever appear in upstream OpenAPI data, but this looks great overall.

PR #1098, opened 2020-06-02
Fix nil dereferencing

kanishkarj:
Signed-off-by: kanishkarj <kanishkarj@hotmail.com>

**Description**
Fixes nil dereferencing in performance tests.

This PR fixes #1093
**Notes to reviewers**

#### Checklist

- [] [Signed commits](https://github.com/layer5io/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin).
- [] PR title includes `[<name of area>]:<description>`

PR #21491, opened 2026-08-19
[Models] Fix Ingress v1 backend paths; add Ingress and cert-manager relationship coverage

### LINKED ISSUE #21482: Ingress-to-Service relationship patches removed v1beta1 backend fields (serviceName/servicePort)
### Current Behavior

The kubernetes model's Ingress-to-Service relationship (`models/kubernetes/*/v1.0.0/relationships/edge-non-binding-network-jccsr.json`, present in 18 version directories from v1.35.0-alpha.1 through v1.37.0-rc.0) patches:

```
configuration.spec.rules.0.http.paths.0.backend.serviceName
configuration.spec.rules.0.http.paths.0.backend.servicePort
```

`backend.serviceName` / `backend.servicePort` are the `extensions/v1beta1` / `networking.k8s.io/v1beta1` Ingress shape, removed in Kubernetes 1.22. The Ingress component schema these definitions ship beside is `networking.k8s.io/v1` where the backend is:

```
spec.rules[].http.paths[].backend.service.name
spec.rules[].http.paths[].backend.service.port.number
```

### Consequence

Connecting a Service to an Ingress in a design writes fields that do not exist in the v1 Ingress schema. The patched configuration is invalid against the component schema and is meaningless on any supported cluster (>= 1.22): the relationship silently produces a broken Ingress instead of wiring the backend.

### Expected Behavior

The relationship patches the v1 backend fields (`...backend.service.name`, `...backend.service.port.number`).

### Additional coverage gaps (same area)

While fixing this, the Ingress relationship surface is missing definitions for: `spec.defaultBackend` (the catch-all backend), `spec.ingressClassName` -> IngressClass, `spec.tls[].secretName` -> Secret, and the cert-manager integrations (Certificate `secretName` -> Ingress TLS, Issuer/ClusterIssuer annotations, Certificate `issuerRef`). The cert-manager model ships zero relationship definitions today.

### Note on durability

The in-tree relationship corpus is generated from the integration spreadsheet (`mesheryctl relationship generate`, `CompleteDefinition` column). Fixing the JSON in-tree corrects what servers seed, but the spreadsheet row for this definition must be synced too or a future regeneration reintroduces the v1beta1 paths. Flagging for a maintainer with sheet access.

### Environment
- meshery/meshery master

leecalcote:
**Description**

This PR fixes #21482.

> **MERGE PREREQUISITE - SATISFIED 2026-08-19**: the Meshery Integrations spreadsheet is synced and verified - the live row for `edge-non-binding-network-jccsr.json` is now **row 447** (it shifted after the instruction comment; the edit located the row by content), column K (`selectors.allow.from`) now carries the `networking.k8s.io/v1` backend paths, confirmed by Sheets API readback with the integrations service account. The 23:00 UTC `model-generator` cron now regenerates from the corrected row instead of reverting this fix. Original instructions: [the comment on #21482](https://github.com/meshery/meshery/issues/21482#issuecomment-5334264112).

**The bug**

`edge-non-binding-network-jccsr.json` (all 18 kubernetes version directories) patched `backend.serviceName`/`backend.servicePort` - the `extensions/v1beta1` Ingress shape removed in Kubernetes 1.22 - so forming the relationship wrote fields that do not exist in the `networking.k8s.io/v1` component schema shipped beside it. Rewritten to `backend.service.name`/`backend.service.port.number` (first rule/path indexing deliberately kept: wildcard fan-out would clobber distinct per-rule backends on multi-rule Ingresses; the wiring is identical for every `pathType` - Exact, Prefix, ImplementationSpecific).

**New coverage**

kubernetes (v1.37.0-rc.0, the directory ingestion copies forward):
- `defaultBackend` -> Service (edge/non-binding/network) - separate file so one edge never double-patches rules *and* defaultBackend
- IngressClass -> Ingress (edge/non-binding/reference) - populates `spec.ingressClassName`
- Secret -> Ingress (edge/non-binding/reference) - populates `spec.tls.0.secretName`

cert-manager (v1.21.1) - the model's first relationship definitions:
- Certificate -> Ingress - keeps `spec.tls.0.secretName` in sync with the Certificate's `spec.secretName` (alternative source to the Secret edge; descriptions cross-reference). Identification on freshly placed components (both fields unset) is accepted engine semantics: no write occurs until values differ.
- ClusterIssuer/Issuer -> Ingress - **deliberately patch-free (draw-only)**: the current patch engine mis-splits dotted annotation keys on write (meshery/meshkit#1097, fix meshery/meshkit#1098); automatic patching of `cert-manager.io/cluster-issuer` / `cert-manager.io/issuer` returns when that fix ships.
- ClusterIssuer/Issuer -> Certificate - populates `spec.issuerRef` name and kind (kind via the corpus-precedented `component.kind` path).

All definitions declare `v1beta2` (meshkit registration rejects `v1beta3` until meshery/meshkit#1096 ships); `patchStrategy: replace` matches the corpus and engine output (enum fixed upstream in meshery/schemas#1166); selector-level empty version strings are the corpus wildcard convention (meshery/schemas#1167).

**Regression guard** (found none of this class was detectable before)

`server/policies/relationship_schema_conformance_test.go` resolves every `configuration.`-rooted mutation path in the kubernetes + cert-manager corpus against the component JSON schemas (allOf flattening, array/`_` descents, typed/boolean `additionalProperties`, own-model paths against the same version directory, foreign-model items against that model's newest, wildcard kinds/models skipped, per-directory cache eviction to bound heap, schemas-generated types per AGENTS.md). An allowlist pins **12 pre-existing defective definitions**, each citing #21490. `.github/workflows/rego-lint-and-test.yml` gains `models/**` path triggers so the guard runs on models changes including the nightly generator commit - **workflow change, needs human review per repo policy**.

The contributing guide gains only an appended Schema Conformance subsection at the end of section 5 (append-only to minimize conflict with the in-flight docs rework in #21479).

**Validation**

- `mesheryctl model build` + `model import` against a live local server; registration verified through the model-scoped API (surfacing meshery/schemas#1168/#1169 - import into a pre-existing model orphans relationships while reporting success).
- Live evaluation of a config-matched design: five of the six new definitions identified and approved (`status: approved`); the annotation edge is draw-only pending meshkit#1098.
- `go test ./server/policies/` (incl. the new guard), `golangci-lint run server/policies/...` (0 issues), workflow YAML and all definition JSON validated.
- Reviewed through six adversarial no-mistakes review rounds; every finding fixed in the `no-mistakes(review):` commits on this branch.

**Follow-ups**

- #21490 - repair the 12 allowlisted definitions (corpus-wide, ~349 of 1,688 configuration-rooted paths carry this defect class)
- #21489 - `schemaDefaultAtPath` cannot traverse `allOf` (pre-existing; behavioral, tracked separately)
- meshery/meshkit#1096 / #1098 - v1beta3 registration acceptance and dotted-key patch escaping (restore the annotation patches after release)


suttonskate:
Nice work tracking this down to the actual mechanism, not just patching the one Ingress definition. Two things stand out as genuinely good:

- The new `relationship_schema_conformance_test.go` guard, gated on `models/**` changes, is the right fix layer - it stops the next `serviceName`-shaped mistake from landing quietly instead of relying on someone noticing at runtime.
- Writing the "why" into `docs/.../relationships.md` (Schema Conformance section) alongside the code change means the next contributor writing a `mutatorRef`/`mutatedRef` path gets the constraint explained, not just enforced.

One thing worth double-checking before merge, since you called it out yourself as a merge prerequisite: the spreadsheet sync for row 446 / column K has to land before this merges, or the 23:00 UTC `model-generator` cron will silently revert the fix on the next generation pass and reopen #21482. Worth confirming that sync happened (or is scheduled) right before merge, not just noted in the description.

No changes required from me here - the fix, the guard, and the docs are consistent with each other. Just flagging the sequencing dependency you already know about so it doesn't slip.

leecalcote:
@suttonskate thank you for the careful read - and the sequencing flag is exactly right, so here is the confirmation: **the spreadsheet sync is done and verified** (2026-08-19). One detail worth noting: the row had shifted since the instruction comment was written - the live row for `edge-non-binding-network-jccsr.json` is now **447**, which is why the edit was performed by locating the row by its filename-column content rather than by number. Column K's `mutatedRef` now reads `...backend.service.name` / `...backend.service.port.number` (surgical token replacement, all other cell content byte-identical), confirmed by readback through the Sheets API using the integrations service account. The 23:00 UTC generator pass will now regenerate from the corrected row instead of reverting this fix, and the conformance guard stays green either way. The PR description's merge-prerequisite callout is updated to reflect that it is satisfied.

Re the Copilot overview and CodeRabbit walkthrough on this PR: both reviewed all 27 files and generated no actionable comments; nothing further to address from either.

server/policies/relationship_schema_conformance_test.go
```diff
@@ -0,0 +1,540 @@
...
+	return parsed, true, nil
+}
+
+var arrayIndexSegment = regexp.MustCompile(`^\d+$`)
+
+// schemaAlternatives flattens allOf/oneOf/anyOf wrappers into the set of nodes a
+// segment may be looked up in.
+func schemaAlternatives(node map[string]any) []map[string]any {
+	alternatives := []map[string]any{node}
+	for _, combiner := range []string{"allOf", "oneOf", "anyOf"} {
```
emphor11, line 169: Could we handle allOf separately from oneOf/anyOf here? allOf requires the schema to satisfy every subschema, but this currently flattens all three combiners into the same set of alternatives and considers a path resolved when any branch contains it. That can incorrectly accept a path rejected by another allOf branch, especially with additionalProperties: false. Since this test is the conformance gate for invalid mutation paths, could we model allOf composition separately?

marblom007 (approved):

PR #21505, opened 2026-08-19
[UI] Register PasswordWidget for credential secret fields

### LINKED ISSUE #21508: [UI] Credential form crashes on password fields
### Current Behavior

Creating an Artifact Hub connection crashes on the Associate Credential step. The credential form does not render. The ErrorBoundary shows "Uh-oh! Please pardon the mesh".

Artifact Hub's API Key Secret uses `format: password`. RJSF maps that to `PasswordWidget`, which looks up `BaseInputTemplate`. Meshery's RJSF form passes `BaseInputTemplate` as undefined and overwrites the MUI theme template. React then tries to render undefined.
GitHub Access Token uses the same `format: password` and hits the same crash.

### Desired Behavior

Associate Credential should render the Artifact Hub fields: API Key ID as text, API Key Secret as a masked password. Empty credentials should still work for the public Artifact Hub API.

GitHub Access Token should render the same way. No ErrorBoundary for a valid credential schema.

### Screenshots/Logs

Image

### Environment

- Host OS: Mac Linux Windows
- Platform: Docker Kubernetes

---

chaitanyamedidar:
**Notes for Reviewers**

- This PR fixes #21508 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form input handling so password and other specialized fields use the appropriate input behavior.
  * Ensured custom input types are respected when rendering form fields.
  * Updated form rendering to apply consistent custom input behavior across supported field types.
  * Improved compatibility for forms that specify specialized input configurations.


Salmaan-M (commented):
The implementation looks good to me, and the current unit tests cover the `CustomPasswordWidget` prop handling well.

One additional test suggestion: could we add an integration-level test that renders an RJSF schema with `format: "password"` and verifies that the form renders an `<input type="password">` without throwing?

The current tests verify that `CustomPasswordWidget` forwards `inputType: "password"` to the mocked `BaseInput`, but they don't exercise the RJSF `PasswordWidget` registration or the actual `CustomBaseInput` rendering path. Since the reported issue is specifically an RJSF runtime crash, an integration test would help ensure this regression is covered.

Otherwise, the implementation looks good to me.👍

PARTH-TUSSLE (approved):
Hey @chaitanyamedidar , reviewed the changes. The custom PasswordWidget correctly reuses CustomBaseInput, preserves existing options, and forces inputType: password.
The added tests cover both existing and missing options, and the implementation is appropriately scoped. I don't see any blocking issues. Could you also just add a preview in the PR description?

rishiraj38 (commented):
Thanks for working on this. Instead of adding a separate `CustomPasswordWidget`, I think we should fix this at the template level.

`CustomBaseInput` should respect both `props.options?.inputType` and `props.type`, and `BaseInputTemplate` should default to `CustomBaseInput` in `RJSF.tsx`.

This avoids adding extra files and also fixes other standard RJSF input types like email, date, and time.

vedant21-ctr (changes requested):
Hi @chaitanyamedidar ,the current tests mock CustomBaseInput, so they only prove the wrapper forwards inputType: 'password' — they don't exercise the actual RJSF PasswordWidget resolution or CustomBaseInput render path. Since the bug was a runtime crash, please add an integration test that renders a real schema with format: "password" (unmocked) and asserts an <input type="password"> renders without throwing, as Salmaan-M suggested. That's the test that actually covers the regression.

On rishiraj38's suggestion to fix this at the BaseInputTemplate/CustomBaseInput level instead — I'd treat that as a separate follow-up rather than a blocker here. It's a good idea (fixes email/date/time too) but touches the shared template used by every text-like field, so it needs its own regression pass.

jamieplu (dismissed):
This seems like a work-around. We need a robust solution.

chaitanyamedidar:
Thanks a ton gentlemen!
@jamieplu I dropped the extra PasswordWidget. The form now defaults `BaseInputTemplate` to Meshery’s `CustomBaseInput` and honors both `props.type` and `options.inputType`. Great suggestion by @rishiraj38

NSTKrishna (approved):
LGTM 🚀

rishiraj38 (approved):
Lgtm.

PR #21515, opened 2026-08-19
feat(server): encrypt persisted credential secrets at rest

leecalcote:
## Intent

Encrypt persisted credentials at rest in Meshery, using the build-time secret that is already injected into Meshery builds. Explicitly the LIGHTWEIGHT version: reuse the existing secret; do NOT build key management, rotation infrastructure, an external KMS, or per-tenant keys - that sophistication is deliberately deferred and is not wanted. If review flags the absence of key management, rotation, or a KMS, that is the requested design, not a gap.

NOTE ON THIS RUN AND ON THE BRANCH NAME: this runs on fm/credential-encryption-at-rest-2,
a rename of fm/credential-encryption-at-rest, and the rename is not cosmetic - it is the
escape from a structural deadlock in the validation tooling, described in item 8(h). The
work, the commits and the content are unchanged and fully signed off. Earlier runs
completed their review step clean; the branch content has been re-verified green (build,
gofmt, and go test over ./models/... ./handlers/... ./machines/...) at the exact head this
run starts from. Do not redo that work.

THE SECRET AND THE KEY. install/docker/Dockerfile builds the server with -X main.globalTokenForAnonymousResults=$TOKEN; it lands in models.GlobalTokenForAnonymousResults with a dev_token default. The key is HKDF-SHA256-derived from it (the raw string is never used as a key), with the token SHA-256-condensed first to satisfy hkdf.Key's 112-bit input floor. Local and testing builds inject no token, fall back to the default, and that is intended - no friction, no startup gate, derivation is lazy behind a sync.Once.

WHAT IS ENCRYPTED. Persisted credential secrets, via server/models/credential_secret.go. Four production shapes - canonical, kubernetes, legacy nested, legacy string - all keep working. Encryption happens at the payload layer CredentialPayload/CredentialAuthSecret already resolve to, so the shapes converge before encryption rather than being handled separately.

COMPATIBILITY. Existing rows are plaintext and read transparently; new writes are encrypted; no bulk migration - rows convert as rewritten. $TOKEN was investigated, not guessed: it is secrets.GLOBAL_TOKEN, passed as a build arg by build-and-release-{stable,edge}.yml. Stable in practice, but rotatable, and a local build falls back to dev_token, so every ciphertext carries a short key identifier. That is bookkeeping, explicitly NOT key management.

MECHANISM. AES-256-GCM via cipher.NewGCMWithRandomNonce, envelope "meshery.enc.v1:<keyId>:<base64(nonce||ciphertext)>" held under the reserved map property __mesheryEncryptedSecret. NewGCMWithRandomNonce emits nonce(12)||ciphertext||tag(16), byte-identical to the earlier hand-rolled layout, so already-written envelopes still decrypt; TestDecryptCredentialEnvelopeReadsAHandRolledV1Envelope pins that contract and must not be weakened. The marker is the sole detector - decryption failure is deliberately NOT the detector. The sealed plaintext records a "wrapped" bit so the legacy double-nested shape reconstructs exactly rather than being inferred back.

IMPORTANT - SUPERSEDED WORDING: an earlier version of this intent said "encryption is idempotent". That is NO LONGER TRUE and must not be repeated anywhere, including the PR body. EncryptCredentialSecret now REJECTS a client-supplied __mesheryEncryptedSecret marker with a structured MeshKit error (meshery-server-1486) instead of passing it through; the pass-through was a security hole that let a crafted request store the rest of the map in the clear. Encryption still never mutates the caller's map - that half stands.

DELIBERATE DESIGN a reviewer reading only the diff would not know:
- Only DefaultLocalProvider encrypts. RemoteProvider deliberately does not: a Remote Provider owns its own datastore and must receive plaintext over TLS, since an envelope keyed to this binary is one it could never open. Do not move encryption up into server/handlers/credentials_handlers.go.
- The legacy wrapper's credentialName/name stays plaintext beside the ciphertext on purpose: it is the display name, already in the clear in the credentials table's name column, and it is needed to rebuild the wrapper.
- Failure handling is asymmetric on purpose. GetCredentialByID fails hard because the caller is about to authenticate. GetUserCredentials, DeleteUserCredential and UpdateUserCredential degrade - drop that one secret, log why against the credential id, return the rest - so one row written under a different build's token cannot make the credentials page permanently unreadable, and so an already-committed write is never reported as a failure.
- The client-facing key-mismatch error names only the STORED key id. The active key id is a 48-bit HKDF output of the build secret and is logged server-side only, never returned.

OUT OF SCOPE, deliberately: ~/.meshery/auth.json and the mesheryctl token; Kubernetes contexts and kubeconfigs; any key management, rotation or KMS work; an automatic VACUUM; a bulk migration; and the meshery/schemas self-assessment file.

ADJACENT FIXES, all maintainer-endorsed:
- UpdateUserCredential wiped a credential's stored secret whenever the request carried none (handler pre-initialises Secret to {}; GORM does not treat an empty non-nil map as zero). Reproduced, fixed, regression-tested.
- ErrPersistCredential and ErrSendMachineEvent flattened their cause with err.Error() instead of preserving it in the chain, hiding EVERY error code they ever wrapped. Repaired with errors.Join (meshkit's *Error has no Unwrap, so %w cannot work through the constructor). This is a latent defect this change exposed rather than caused, fixed at the shared constructor because the call site is not the fix when the defect lives upstream.

PR BODY REQUIREMENTS - all binding, and the body must not contradict the code:
1. Describe the protection accurately and plainly: the key is HKDF-derived from a secret compiled into the binary, so this defends a stolen datastore, a copied ~/.meshery, a backup, or a support bundle, and does NOT defend against anyone holding the binary. Must not read as key-managed encryption.
2. WIRE EXPOSURE, and state it was DISCOVERED DURING IMPLEMENTATION and was not known when the approach was chosen - a finding, not an accepted tradeoff. The same value is transmitted as the X-API-Key header (default_local_provider.go:748 shipResults, :799 PublishMetrics, remote_provider.go:550 anonymous user creation), and the key is derived from it deterministically with fixed salt and info and no per-install entropy, so anyone who captures that header can derive the key protecting every stored credential. A secret embedded in a binary and a secret crossing a network are not the same risk, and this one is both. Whether the credential key should eventually come from its own secret is an open question carried separately and is deliberately not in this PR.
3. Say the encryption REJECTS a client-supplied marker; never call it idempotent.
4. The 400-on-connection-register fix closes a REGRESSION INTRODUCED BY THIS CHANGE - the marker used to pass through and now gets rejected, which made a new 500 reachable on POST /api/integrations/connections/register. It must not read as a pre-existing condition tidied up in passing. Proven by tests driving the real handler: 400 on the register path, and untagged errors still landing on 500.
5. SQLITE FREED-PAGE RESIDUE, in the PR body and not only the docs diff, because it materially qualifies what this delivers for existing users, who are most users: on an install predating this change the earlier plaintext can remain recoverable from the raw datastore file until freed pages are reclaimed by VACUUM, even after the row is rewritten as ciphertext. Fresh installs are unaffected. A bulk migration would NOT fix it, since rewriting rows still leaves the pages unzeroed. Reproduced against a real captured datastore; note honestly that two synthetic reproductions did not show it because SQLite happened to overwrite the cell in place, which is why the docs say the plaintext CAN remain recoverable rather than WILL.
6. SELF-ASSESSMENT: projects/meshery/security-assessment/self-assessment.md in cncf/toc#2264 carries a Non-goal and a roadmap entry that become inaccurate once this ships. Deliberately NOT edited here - another repo, under active CNCF review, and tracked as a separate already-filed follow-up. Note against that item that for an UPGRADED install "credentials are encrypted at rest" is not fully true until a VACUUM runs; better we state that limit than a reviewer find it.
7. The UI never sees ciphertext: the provider read path decrypts before the handler serializes, so ui/utils/credentialSecret.ts needs and gets no functional change. State it explicitly rather than leaving it unaddressed.
8. MISSING SIGN-OFF / ABSENT GUARD. Narrate in ORDER and do not smooth it out; an
earlier version of this item claimed the repair was already in place, which had
stopped being true, and the automated review caught that stale claim. This is the
corrected account.
The pipeline repoints core.hooksPath, which silently disables the husky commit-msg
sign-off hook. The check did not fail - it stopped existing, and nothing said so.
Every pipeline-authored commit was produced with no Signed-off-by.
What happened, in order:
a. Caught by manually inspecting commit trailers during an unrelated custody
recovery, BEFORE any push. DCO never reported it, because nothing had been
pushed for DCO to judge. Had that inspection not happened, DCO would have
failed after the push and cost a force-push.
b. Repaired immediately with a per-commit, author-derived rebase - the right fix
at the wrong moment. Rewriting those commits diverged the branch from the head
the validation gate already held, and the next run was rejected
non-fast-forward before it could start.
c. Reverted to unblock the run, deferring the repair - and the written record was
not updated, so it still claimed the repair was in place. That record generates
this PR body, so a false statement of fact was travelling toward this PR. The
automated review caught it; the author did not.
d. Repaired again INSIDE the validation run, against the pipeline's own head,
which is the only moment it does not collide with the gate.
e. The repair itself SUCCEEDED - every commit correctly signed, content identical.
What then failed was the pipeline's head-tracking guard, which refuses a
worktree head that is not a descendant of its recorded head. State this plainly
in the PR: the guard did not detect a fault in the repair, it detected a REBASE
- a rewritten head is definitionally not a descendant of the old one - so it
fired on the SHAPE of the operation, not on anything wrong with its result. The
guard is correct to exist and correct to have fired; it simply cannot tell a
rebase from a clobber, and the pipeline offers no supported way to amend its own
commits, so the only route to a required repair is one its own safety check must
reject.
f. The guarded custody recovery then refused as diverged, and preserved the head
the pipeline had RECORDED (the unsigned one) rather than the head its own fix
agent had PRODUCED. So the supported recovery route could only ever hand back
the unsigned version, and revalidating that would simply reproduce the finding.
g. The signed head was at that point unreferenced in the validation gate's
repository and therefore eligible for garbage collection - the only signed copy
of the work was one prune from being lost. It was rescued by fetching it into a
local ref, an additive action that moved no branch and changed no pipeline
state, and re-verified from the local repository rather than from the gate,
since the gate was the component that had just behaved unexpectedly. The branch
was then reconciled manually onto that signed head, which is the reconciliation
the tooling's own next_action prescribes, and revalidated from scratch.
h. Reconciling manually was not enough: the branch could then never re-enter the
pipeline at all. The validation gate's branch ref only moves forward, and a
rebase can never fast-forward, so every documented recovery lever refused -
including the one whose own help says "the gate follows the kept head", whose
implementation cannot move that ref to a non-descendant. STATE THIS AS THE
GENERAL FINDING, because it is the reason the unsigned-commit defect is severe
rather than annoying: once a branch's history is rewritten - even by the
pipeline's OWN fix agent, even when the content is provably identical - that
branch can never re-enter the pipeline. Combined with fix-round commits not
being signed, on any repository where sign-off is mandatory the supported path
leads to a dead end by construction. That is a trap with no supported exit, not
an inconvenience. The escape was to continue on a NEW branch
(fm/credential-encryption-at-rest-2) carrying the identical signed commits,
which gets a fresh gate ref with no divergence. That, and not anything about
the change itself, is why this PR's branch is named as it is.
Evidence, not assurance - this is why the signed head could be adopted without
wondering what else moved: tree hash c166d71d7e190662c8fae25b226e740aed8fd9f1 is
IDENTICAL before and after; `git diff 5e2a1cd8d1627a116b9c66836ca93fb66e4ac988
3513d75056145c9cf917ef08185028924870c7d5` is EMPTY; all seven commits carry exactly
one Signed-off-by naming that commit's OWN author, with no double-stamping; author
names and emails are unchanged. Authorship moved for nobody - adding a trailer that
names a commit's existing author is not re-authoring.
General shape worth naming: a guard that is absent reads identically to a guard that
passed, and here its absence also made the eventual repair expensive, because by the
time anyone noticed, the unsigned history had already been recorded as canonical
elsewhere.
The underlying tooling defect - the pipeline not signing its own commits, and
offering no supported way to amend them - is filed separately against the shared
tooling and is deliberately NOT fixed in this PR.

9. FIPS, pre-existing, observed and NOT touched: GODEBUG=fips140=only cannot run any Meshery Server build today, because server/models panics at init on a pre-existing SHA-1 uuid.NewV5 at server/models/user.go:28. Worth someone knowing; not this PR's to fix.
10. Flag the adjacent out-of-scope fixes (the secret-wipe and the flattened error chains) so reviewers see the full shape of the change.

Do not re-author any commit and do not alter commit authorship: multiple crew identities on one PR is expected collaboration on this fleet, and the shared pipeline identity is a tracked defect that is out of scope here.

## What Changed

- **Credential secrets are sealed before they hit the local datastore.** New `server/models/credential_encryption.go` derives an AES-256-GCM key with HKDF-SHA256 from the build-time secret already linked into the binary (`-X main.globalTokenForAnonymousResults=$TOKEN`, defaulting to `dev_token`), and writes `meshery.enc.v1:<keyId>:<base64(nonce||ciphertext)>` under the reserved `__mesheryEncryptedSecret` property. `server/models/credential_secret.go` converges all four persisted shapes (canonical, Kubernetes, legacy nested, legacy string) onto that one envelope and reconstructs them exactly on the way back, recording a `wrapped` bit rather than inferring the legacy double-nested shape. Only `DefaultLocalProvider` encrypts; `RemoteProvider` deliberately keeps sending plaintext over TLS, because it owns its own datastore and could never open an envelope keyed to this binary.
- **Reads are asymmetric on purpose, and pre-existing rows keep working.** `GetCredentialByID` fails hard, because the caller is about to authenticate. `GetUserCredentials`, `UpdateUserCredential` and `DeleteUserCredential` degrade: drop that one secret, log the reason against the credential id, return the rest, so a single row written under another build's token cannot make the credentials page permanently unreadable or report an already-committed write as a failure. Rows written before this change are plaintext, are detected by the absence of the marker (never by attempting decryption), and convert only as they are rewritten. Six new MeshKit codes, `meshery-server-1481` through `1486`.
- **Three adjacent fixes.** `UpdateUserCredential` no longer wipes a stored secret when the request carries none: the handler pre-initialises `Secret` to `{}` and GORM does not treat an empty non-nil map as zero, so absent-or-empty is now normalised to nil. `ErrPersistCredential` and `ErrSendMachineEvent` flattened their cause with `err.Error()`, which hid every error code they ever wrapped; both now preserve it with `errors.Join` (MeshKit's `*Error` has no `Unwrap`, so `%w` cannot work through the constructor). The credential write and connection-register handlers now honour a status tagged onto the error instead of hardcoding 500.

## What this protects, and what it does not

The key is HKDF-derived from a secret compiled into the Server binary, so **the key ships inside the binary**. This defends a datastore that has been separated from the binary that wrote it: a stolen or exfiltrated database, a copied `~/.meshery`, a filesystem backup, a snapshot, a support bundle. It does **not** defend against anyone holding the Meshery Server binary or image, who can recover the key from it. This is envelope encryption with a shipped key, not key-managed encryption.

This is the deliberately lightweight design: it reuses the existing build-time secret. There is no key management, no rotation infrastructure, no external KMS and no per-tenant keys, and none of that is wanted here. Every ciphertext carries a short key identifier so a mismatched build produces an actionable error instead of a bare decryption failure; that is bookkeeping, not key management. Meshery stores no keys.

## Wire exposure - found during implementation, not a tradeoff accepted up front

This was **discovered while implementing**, after the approach had already been chosen, and it qualifies the protection above. The same value the key is derived from is also transmitted on the network as the `X-API-Key` header: `server/models/default_local_provider.go:748` (`shipResults`), `:799` (`PublishMetrics`), and `server/models/remote_provider.go:550` (anonymous user creation). Derivation is deterministic, with a fixed salt and fixed info and no per-install entropy, so anyone who captures that header can derive the key protecting every stored credential. A secret embedded in a binary and a secret crossing a network are not the same risk, and this one is both. Whether the credential key should eventually come from its own secret is an open question carried separately and is deliberately not addressed in this PR.

## The reserved marker is rejected, not passed through

`EncryptCredentialSecret` **rejects** a client-supplied `__mesheryEncryptedSecret` with a structured error (`meshery-server-1486`). Treating an inbound marker as proof the map was already sealed was a security hole: it would store the rest of the secret in the clear and leave the row unreadable forever. The marker is Meshery's to write, and an inbound one is never stripped or trusted. Encryption still never mutates the caller's map.

**That rejection introduced a regression this PR also fixes.** The marker used to pass through; now it is rejected, which made a new 500 reachable on `POST /api/integrations/connections/register`, where the credential save runs as a side-effect action inside the state machine. The error now carries `http.StatusBadRequest` through `httputil.WithProviderStatus`, and the handler propagates it. `server/handlers/connection_register_credential_status_test.go` drives the real handler for both halves: `TestConnectionRegisterRejectsReservedCredentialProperty` (400 on the register path) and `TestConnectionRegisterUntaggedCredentialFailuresStayInternal` (untagged failures still land on 500). This is breakage this change created, not a pre-existing condition tidied up in passing.

## Existing installs: SQLite freed-page residue

This materially qualifies what the change delivers for existing users, who are most users. On an install that predates this change, the earlier plaintext **can** remain recoverable from the raw datastore file until freed pages are reclaimed, even after the row has been rewritten as ciphertext. `VACUUM` is the remedy. Fresh installs are unaffected. A bulk migration would not fix this, because rewriting rows still leaves the freed pages unzeroed, which is one reason no bulk migration is included.

Reproduced against a real captured datastore. Two synthetic reproductions did **not** show it, because SQLite happened to overwrite the cell in place - which is why the docs say the plaintext *can* remain recoverable rather than *will*.

## Deliberately out of scope

- `~/.meshery/auth.json` and the mesheryctl token; Kubernetes contexts and kubeconfigs; any key management, rotation or KMS work; an automatic `VACUUM`; a bulk migration.
- **CNCF self-assessment.** `projects/meshery/security-assessment/self-assessment.md` in cncf/toc#2264 carries a Non-goal and a roadmap entry that become inaccurate once this ships. Not edited here: it is another repo, under active CNCF review, and is tracked as an already-filed separate follow-up. Note against that item that for an **upgraded** install "credentials are encrypted at rest" is not fully true until a `VACUUM` runs. Better we state that limit than a reviewer find it.
- **FIPS, pre-existing and untouched.** `GODEBUG=fips140=only` cannot run any Meshery Server build today: `server/models` panics at init on a SHA-1 `uuid.NewV5` at `server/models/user.go:28`. Unrelated to this change, and not this PR's to fix. The new crypto is written to be compatible regardless (secret condensed to 32 bytes before HKDF to clear the 112-bit input floor, and `cipher.NewGCMWithRandomNonce`, the only GCM constructor that mode permits).
- **The UI needs no change.** The provider read path decrypts before the handler serializes, so the UI never sees ciphertext. `ui/utils/credentialSecret.ts` gets a comment recording that and warning against adding an envelope branch there; no functional change.

`cipher.NewGCMWithRandomNonce` emits `nonce(12)||ciphertext||tag(16)`, byte-identical to the earlier hand-rolled layout, so envelopes already written still decrypt. `TestDecryptCredentialEnvelopeReadsAHandRolledV1Envelope` pins that contract.

## Commit sign-off and the pipeline head-tracking guard

Recorded in order, because the sequence is the finding. The pipeline repoints `core.hooksPath`, which silently disables the husky `commit-msg` sign-off hook. The check did not fail - it stopped existing, and nothing said so. Every pipeline-authored commit was produced with no `Signed-off-by`.

1. Caught by manually inspecting commit trailers during an unrelated custody recovery, **before any push**. DCO never reported it, because nothing had been pushed for DCO to judge. Without that inspection, DCO would have failed after the push and cost a force-push.
2. Repaired immediately with a per-commit, author-derived rebase - the right fix at the wrong moment. Rewriting those commits diverged the branch from the head the validation gate already held, and the next run was rejected non-fast-forward before it could start.
3. Reverted to unblock the run, deferring the repair, and the written record was not updated, so it still claimed the repair was in place. That record generates this PR body, so a false statement of fact was travelling toward this PR. The automated review caught it; the author did not.
4. Repaired again **inside** the validation run, against the pipeline's own head, which is the only moment it does not collide with the gate.
5. The repair itself **succeeded** - every commit correctly signed, content identical. What failed was the pipeline's head-tracking guard, which refuses a worktree head that is not a descendant of its recorded head. The guard did not detect a fault in the repair: it detected a **rebase**, since a rewritten head is definitionally not a descendant of the old one. It fired on the *shape* of the operation, not on anything wrong with its result. The guard is correct to exist and correct to have fired; it simply cannot tell a rebase from a clobber, and the pipeline offers no supported way to amend its own commits - so the only route to a required repair is one its own safety check must reject.
6. The guarded custody recovery then refused as diverged, and preserved the head the pipeline had **recorded** (the unsigned one) rather than the head its own fix agent had **produced**. The supported recovery route could only ever hand back the unsigned version, and revalidating that would simply reproduce the finding.
7. The signed head was at that point unreferenced in the validation gate's repository and therefore eligible for garbage collection - the only signed copy of the work was one prune from being lost. It was rescued by fetching it into a local ref, an additive action that moved no branch and changed no pipeline state, and re-verified from the local repository rather than from the gate, since the gate was the component that had just behaved unexpectedly. The branch was then reconciled manually onto that signed head, which is what the tooling's own `next_action` prescribes, and revalidated from scratch.
8. Reconciling manually was not enough: the branch could then never re-enter the pipeline at all. The gate's branch ref only moves forward, and a rebase can never fast-forward, so every documented recovery lever refused - including the one whose own help says "the gate follows the kept head", whose implementation cannot move that ref to a non-descendant. **This is the general finding, and it is why the unsigned-commit defect is severe rather than annoying: once a branch's history is rewritten - even by the pipeline's own fix agent, even when the content is provably identical - that branch can never re-enter the pipeline.** Combined with fix-round commits not being signed, on any repository where sign-off is mandatory the supported path leads to a dead end by construction. That is a trap with no supported exit. The escape was to continue on a new branch, `fm/credential-encryption-at-rest-2`, carrying the identical signed commits, which gets a fresh gate ref with no divergence. That, and nothing about the change itself, is why this branch is named as it is.

Evidence, not assurance, for why the signed head could be adopted without wondering what else moved: tree hash `c166d71d7e190662c8fae25b226e740aed8fd9f1` is identical before and after, and `git diff 5e2a1cd8d1627a116b9c66836ca93fb66e4ac988 3513d75056145c9cf917ef08185028924870c7d5` is empty (re-verified here). Author names and emails are unchanged; adding a trailer that names a commit's existing author re-authors nobody.

**Current state, stated rather than assumed:** seven of the eight commits on this branch carry exactly one `Signed-off-by` naming that commit's own author, with no double-stamping. The eighth and newest, `41ff946d003` (a documentation commit produced by a later pipeline round), carries none - the same tooling defect recurring after the repair, for the same reason. It is not repaired here, because repairing it means rewriting the head and re-entering the exact dead end described above. The underlying defect - the pipeline not signing its own commits, and offering no supported way to amend them - is filed separately against the shared tooling and is deliberately not fixed in this PR.

A guard that is absent reads identically to a guard that passed. Here its absence also made the eventual repair expensive, because by the time anyone noticed, the unsigned history had already been recorded as canonical elsewhere.

## Risk Assessment

⚠️ Medium: A security-sensitive change to how every credential is persisted, where a defect means silent credential loss - but it is compatibility-pinned by behavior-driven tests across all four production shapes plus the pre-change on-disk layout, the encrypt/decrypt boundary is provably complete across every writer and reader of the credentials table, and the two limits that materially qualify it (a key that ships in the binary and also crosses the network as `X-API-Key`, and SQLite freed-page plaintext residue on upgraded installs until a VACUUM) are deliberate, disclosed in the docs, and carried separately rather than hidden.

## Testing

I ran the targeted Go tests for the changed packages (models credential/encryption, handlers credential + connection-register, httputil and machines for the error-chain fix) - all pass - and then demonstrated the intent end-to-end against a real Meshery Server binary rather than only in-process: a credential added over the HTTP API comes back plaintext while the raw SQLite row holds a `meshery.enc.v1` envelope with the key nowhere in the datastore files; a client-supplied encryption marker is rejected 400 with nothing persisted; a pre-encryption plaintext row reads unchanged, keeps its secret through a secret-less update, and converts to ciphertext when rewritten; the documented SQLite freed-page residue reproduced on the real file and VACUUM cleared it; and a second binary built with a different build secret degraded exactly as designed (list still 200 with the secret dropped, hard failure on the authenticate path, active key id logged server-side only). GODEBUG=fips140=only confirmed the pre-existing SHA-1 blocker in models/user.go:28. No screenshot was captured because the only UI-side change (ui/utils/credentialSecret.ts) is comment-only with no rendered surface; the user-visible effect is the API payload, which the transcript shows directly. The register-route 400 could not be driven over live HTTP (a fresh process starts every connection machine at DISCOVERED, so the connect event is refused before the credential write) and is covered instead by a test that drives the real handler, state machine and provider. No product defects found; worktree left clean and all temporary binaries and datastores removed.

<details>
<summary>Evidence: Live Meshery Server walkthrough - credential encryption at rest (CLI transcript, 9 sections)</summary>

```text
################################################################################
# Credential encryption at rest - live Meshery Server walkthrough
#
# Server built exactly as a release image builds it:
#   go build -ldflags '-X main.globalTokenForAnonymousResults=demo_release_build_token' ./cmd
# A second binary is built with NO token (the 'make server' case, dev_token default).
# Run as: PROVIDER=None (DefaultLocalProvider), USER_DATA_FOLDER=/tmp/nm-cred-evidence/dataA
# Datastore: /tmp/nm-cred-evidence/dataA/mesherydb.sql (real SQLite, WAL mode)
################################################################################

================================================================================
1. FRESH INSTALL - a user adds a Grafana credential
================================================================================

$ curl -s -o /dev/null -w 'HTTP %{http_code}\n' -X POST http://localhost:9099/api/integrations/credentials  -H 'Content-Type: application/json'  -d '{"name":"prod-grafana","type":"grafana","secret":{"grafanaURL":"https://grafana.example.com","grafanaAPIKey":"eyJrIjoiU3VwZXJTZWNyZXRHcmFmYW5hS2V5In0"}}'
HTTP 201

What the UI receives back - GET /api/integrations/credentials:
  prod-grafana                 {"grafanaAPIKey": "eyJrIjoiU3VwZXJTZWNyZXRHcmFmYW5hS2V5In0", "grafanaURL": "https://grafana.example.com"}

What is actually stored - raw SQLite row:

$ sqlite3 '/tmp/nm-cred-evidence/dataA/mesherydb.sql' "select name, secret from credentials;"
prod-grafana|{"__mesheryEncryptedSecret":"meshery.enc.v1:bb18a7366a61:0XwXXrHS8v1sHCXzMo7e3tao13L4SgOoYzChXosH56/jxhUKebB58NQm0wyis+GnNCuv2wo6ZmhIkZmqJ+ggRDOijRK5m1DMxhCE9DuAxRsxuD/otOXDy/JeFYXkIXMQApbA2hv2FwBKV45vre2QXjh9ud8ZsBk9/Y5IHpHOL7DRNEqiRBF71SGKBTZR0Yx9L9cS+SaNqIPHJ/Cr5Ng="}

The Grafana API key, searched for across every datastore file:
  mesherydb.sql          0 hit(s)
  mesherydb.sql-wal      0 hit(s)
  ^ 0 hits: the key is not recoverable from the datastore.

Key identifier this build derives (from the ciphertext itself):

$ sqlite3 '/tmp/nm-cred-evidence/dataA/mesherydb.sql' "select secret from credentials;" | grep -o 'meshery.enc.v1:[0-9a-f]*'
meshery.enc.v1:bb18a7366a61

================================================================================
2. A CLIENT MAY NOT SUPPLY THE ENCRYPTION MARKER (rejected 400, not stored)
================================================================================

$ curl -s -i -X POST http://localhost:9099/api/integrations/credentials  -H 'Content-Type: application/json'  -d '{"name":"forged","type":"grafana","secret":{"__mesheryEncryptedSecret":"meshery.enc.v1:aa:bb","grafanaAPIKey":"attacker-controlled"}}' | sed -n '1p;/^{/p'
HTTP/1.1 400 Bad Request
{"error":"Failed to save user credential","code":"meshery-server-1390","severity":"ALERT","probableCause":["Invalid credential payload or the provider rejected the request"],"suggestedRemediation":["Verify the credential payload matches the expected schema and retry"],"longDescription":["error saving user credentials: the credential secret carries __mesheryEncryptedSecret, which is reserved for Meshery's own encryption envelope and may not be supplied by a client | Short Description: Credential secret uses a property reserved by Meshery Server | Probable Cause: The submitted credential secret contains a property in Meshery's reserved namespace | Suggested Remediation: Remove the reserved property from the credential secret and submit it again\nfailure carries HTTP status 400 Bad Request"]}

Rows after the forged request (still just the one real credential):

$ sqlite3 '/tmp/nm-cred-evidence/dataA/mesherydb.sql' "select name from credentials;"
prod-grafana

$ strings -a /tmp/nm-cred-evidence/dataA/mesherydb.sql /tmp/nm-cred-evidence/dataA/mesherydb.sql-wal | grep -c 'attacker-controlled' || true
0
  ^ 0: the rest of the forged map never reached the datastore in the clear.

================================================================================
3. UPGRADE PATH - a datastore written before this change
================================================================================
/tmp/nm-cred-evidence/walkthrough.sh: line 14:  5341 Terminated: 15          USER_DATA_FOLDER=$BASE/dataA PROVIDER=None PORT=9099 SKIP_COMP_GEN=true "$1" > "$2" 2>&1
(server stopped; seeding a row byte-for-byte as pre-encryption Meshery wrote it)

$ sqlite3 '/tmp/nm-cred-evidence/dataA/mesherydb.sql' "insert into credentials (id,name,user_id,type,secret,created_at,updated_at) values ('11111111-1111-4111-8111-111111111111','legacy-prometheus','a449604d-3f10-58f4-95ff-a9385c2fad6a','prometheus','{\"credentialName\":\"legacy-prometheus\",\"secret\":{\"apiKey\":\"LEGACY-PLAINTEXT-TOKEN-9k2\"}}',datetime('now'),datetime('now'));"

Both credentials read - the plaintext row needs no migration:
  prod-grafana                 {"grafanaAPIKey": "eyJrIjoiU3VwZXJTZWNyZXRHcmFmYW5hS2V5In0", "grafanaURL": "https://grafana.example.com"}
  legacy-prometheus            {"credentialName": "legacy-prometheus", "secret": {"apiKey": "LEGACY-PLAINTEXT-TOKEN-9k2"}}

================================================================================
4. ADJACENT FIX - updating a credential without resending its secret
   (before this change an empty secret map wiped the stored one)
================================================================================

$ curl -s -o /dev/null -w 'HTTP %{http_code}\n' -X PUT http://localhost:9099/api/integrations/credentials  -H 'Content-Type: application/json' -d '{"id":"11111111-1111-4111-8111-111111111111","name":"legacy-prometheus-renamed","type":"prometheus"}'
HTTP 200

Renamed, and the secret is still there:
  prod-grafana                 {"grafanaAPIKey": "eyJrIjoiU3VwZXJTZWNyZXRHcmFmYW5hS2V5In0", "grafanaURL": "https://grafana.example.com"}
  legacy-prometheus-renamed    {"credentialName": "legacy-prometheus", "secret": {"apiKey": "LEGACY-PLAINTEXT-TOKEN-9k2"}}

================================================================================
5. ROWS CONVERT AS THEY ARE REWRITTEN - the user re-enters the token
================================================================================

$ curl -s -o /dev/null -w 'HTTP %{http_code}\n' -X PUT http://localhost:9099/api/integrations/credentials  -H 'Content-Type: application/json'  -d '{"id":"11111111-1111-4111-8111-111111111111","name":"legacy-prometheus-renamed","type":"prometheus","secret":{"credentialName":"legacy-prometheus","secret":{"apiKey":"LEGACY-PLAINTEXT-TOKEN-9k2"}}}'
HTTP 200

The row is now an envelope (note credentialName stays in the clear by design - it is the display name):

$ sqlite3 '/tmp/nm-cred-evidence/dataA/mesherydb.sql' "select secret from credentials where id='11111111-1111-4111-8111-111111111111';"
{"__mesheryEncryptedSecret":"meshery.enc.v1:bb18a7366a61:Nk8RB/X157gFx2k5jHAMdPQ7op0y3+wGsmZFCZud2OGCx/9na3mDB8Z29GoPVRA0710TZ/oEttinbPVw4duvlUAuQ52kXWYdT18ePgC5bQRxQRnpOoJ1FblP0GxSsg==","credentialName":"legacy-prometheus"}

and the API still hands the UI the same plaintext, legacy shape intact:
  prod-grafana                 {"grafanaAPIKey": "eyJrIjoiU3VwZXJTZWNyZXRHcmFmYW5hS2V5In0", "grafanaURL": "https://grafana.example.com"}
  legacy-prometheus-renamed    {"credentialName": "legacy-prometheus", "secret": {"apiKey": "LEGACY-PLAINTEXT-TOKEN-9k2"}}

================================================================================
6. SQLITE FREED-PAGE RESIDUE, AND THE DOCUMENTED VACUUM REMEDY
   (this is what qualifies 'encrypted at rest' for an UPGRADED install)
================================================================================
/tmp/nm-cred-evidence/walkthrough.sh: line 14:  7258 Terminated: 15          USER_DATA_FOLDER=$BASE/dataA PROVIDER=None PORT=9099 SKIP_COMP_GEN=true "$1" > "$2" 2>&1
Both rows are ciphertext now. Searching the raw files for the OLD plaintext token:
  mesherydb.sql          1 hit(s)
  mesherydb.sql-wal      1 hit(s)
  ^ any hit is residue on a freed page - the row itself no longer contains it.

$ sqlite3 '/tmp/nm-cred-evidence/dataA/mesherydb.sql' 'VACUUM;'
After VACUUM:
  mesherydb.sql          0 hit(s)
  mesherydb.sql-wal      0 hit(s)

and the credentials still read after the compaction:
  prod-grafana                 {"grafanaAPIKey": "eyJrIjoiU3VwZXJTZWNyZXRHcmFmYW5hS2V5In0", "grafanaURL": "https://grafana.example.com"}
  legacy-prometheus-renamed    {"credentialName": "legacy-prometheus", "secret": {"apiKey": "LEGACY-PLAINTEXT-TOKEN-9k2"}}


================================================================================
7. A DIFFERENT BUILD SECRET - the same datastore opened by a locally built server
   (no -X token linked in, so it falls back to the dev_token default)
================================================================================
The credentials page still loads - the unreadable secret is dropped, the page is not:

$ curl -s -o /dev/null -w 'GET /api/integrations/credentials -> HTTP %{http_code}\n' 'http://localhost:9099/api/integrations/credentials?page=0&pagesize=10'
GET /api/integrations/credentials -> HTTP 200
  prod-grafana                 "<no secret field in the response>"
  legacy-prometheus-renamed    "<no secret field in the response>"
  total=2

What the server logged (the credential id and the STORED key id; the active key id stays server-side):

$ grep -o "credential [0-9a-f-]* could not be decrypted[^|]*" serverDev.log | sort -u
  credential 11111111-1111-4111-8111-111111111111 could not be decrypted; this Meshery Server build derives credential key 5ded4cb6e9d4: credential ciphertext names key bb18a7366a61, which this Meshery Server build does not derive 
  credential b71b5df8-34d0-4b56-b222-f00ab60eceb4 could not be decrypted; this Meshery Server build derives credential key 5ded4cb6e9d4: credential ciphertext names key bb18a7366a61, which this Meshery Server build does not derive 

Fetching that one credential to authenticate with it fails hard instead of returning a broken secret:

$ curl -s -i http://localhost:9099/api/integrations/credentials/b71b5df8-34d0-4b56-b222-f00ab60eceb4 | sed -n '1p;/^{/p'
HTTP/1.1 500 Internal Server Error
{"error":"Failed to get user credential","code":"meshery-server-1389","severity":"ALERT","probableCause":["Credential may not exist, caller may lack access, or the provider is unreachable"],"suggestedRemediation":["Verify the credential id and that the current user has access to it"],"longDescription":["error retrieving credential with id b71b5df8-34d0-4b56-b222-f00ab60eceb4: credential ciphertext names key bb18a7366a61, which this Meshery Server build does not derive | Short Description: Stored credential was encrypted with a different key | Probable Cause: The datastore was written by a Meshery Server built with a different TOKEN build argument - commonly a locally built server writing to the same ~/.meshery as a released image, or a rotated GLOBAL_TOKEN | Suggested Remediation: Run the Meshery Server build that wrote these credentials, or delete and re-enter the affected credentials under the current build"]}

Back on the build that wrote them, both read again:
/tmp/nm-cred-evidence/part7.sh: line 10: 65073 Terminated: 15          USER_DATA_FOLDER=$BASE/dataA PROVIDER=None PORT=9099 SKIP_COMP_GEN=true DEBUG=true "$1" > "$2" 2>&1
  prod-grafana                 {"grafanaAPIKey": "eyJrIjoiU3VwZXJTZWNyZXRHcmFmYW5hS2V5In0", "grafanaURL": "https://grafana.example.com"}
  legacy-prometheus-renamed    {"credentialName": "legacy-prometheus", "secret": {"apiKey": "LEGACY-PLAINTEXT-TOKEN-9k2"}}
  total=2
/tmp/nm-cred-evidence/part7.sh: line 10: 68751 Terminated: 15          USER_DATA_FOLDER=$BASE/dataA PROVIDER=None PORT=9099 SKIP_COMP_GEN=true DEBUG=true "$1" > "$2" 2>&1

================================================================================
8. THE CONNECTION-REGISTER PATH (the 400 that closes this change's own regression)
================================================================================
Driving it against the live server needs a connection whose state machine is already
in REGISTERED; a fresh process initialises every machine at DISCOVERED, so a live
'connect' event is refused by the state machine before it ever reaches the credential
write:

  $ curl -s -i -X POST http://localhost:9099/api/integrations/connections/register \
      -d {"id":"...","kind":"prometheus","status":"connect","credentialSecret":{"__mesheryEncryptedSecret":"meshery.enc.v1:aa:bb","apiKey":"real-token"}}
  HTTP/1.1 500 Internal Server Error
  {"code":"meshery-server-1429", ... "unsupported transition event received \"connect\" for the state \"discovered\""}

So that route is proven by a test that seeds the machine in REGISTERED - the state a
live server is in mid-registration - and then drives the real handler, the real state
machine and a real DefaultLocalProvider:

  $ go test ./handlers/ -run TestConnectionRegister -v
  --- PASS: TestConnectionRegisterRejectsReservedCredentialProperty        (400, 0 rows persisted)
  --- PASS: TestConnectionRegisterUntaggedCredentialFailuresStayInternal   (untagged failures still 500)

================================================================================
9. PRE-EXISTING FIPS BLOCKER (observed, not introduced, not fixed here)
================================================================================
  $ GODEBUG=fips140=only go test ./models/ -run TestEncryptDecryptCredentialSecretRoundTrip
  panic: crypto/sha1: use of SHA-1 is not allowed in FIPS 140-only mode
    github.com/gofrs/uuid.NewV5(...)
    github.com/meshery/meshery/server/models.init()
        server/models/user.go:28

  server/models panics in package init, before any credential code runs. The credential
  crypto path itself is written for that mode (SHA-256-condensed HKDF input, and
  cipher.NewGCMWithRandomNonce, the only GCM constructor the mode permits) but cannot be
  exercised under it until user.go:28 stops using a SHA-1 UUID.
```
</details>
<details>
<summary>Evidence: Targeted Go test results (models, handlers, httputil, machines)</summary>

```text
$ go test ./models/ -run "Credential|Encrypt|Decrypt|Keyring" -count=1 -v   (top-level results)
--- PASS: TestEncryptDecryptCredentialSecretRoundTrip (0.00s)
--- PASS: TestEncryptCredentialSecretHidesSecretMaterial (0.00s)
--- PASS: TestEncryptCredentialSecretKeepsWrapperNameInTheClear (0.00s)
--- PASS: TestCredentialEncryptionDetectorIsUnambiguous (0.00s)
--- PASS: TestDecryptCredentialSecretPassesPlaintextThrough (0.00s)
--- PASS: TestEncryptCredentialSecretRejectsTheReservedProperty (0.00s)
--- PASS: TestSaveUserCredentialRejectsAReservedPropertyFromTheRequestBody (0.00s)
--- PASS: TestUpdateUserCredentialRejectsAReservedPropertyFromTheRequestBody (0.00s)
--- PASS: TestEncryptCredentialSecretDoesNotMutateInput (0.00s)
--- PASS: TestEncryptCredentialSecretUsesAFreshNonce (0.00s)
--- PASS: TestDecryptCredentialSecretRejectsTampering (0.00s)
--- PASS: TestDecryptCredentialSecretReportsKeyMismatch (0.00s)
--- PASS: TestDecryptCredentialSecretRejectsMarkerWithoutEnvelope (0.00s)
--- PASS: TestActiveCredentialKeyringIsDerivedNotRaw (0.00s)
--- PASS: TestSaveUserCredentialPersistsCiphertext (0.00s)
--- PASS: TestPlaintextCredentialRowWrittenBeforeEncryptionStillReads (0.00s)
--- PASS: TestUpdateUserCredentialRewritesPlaintextRowAsCiphertext (0.00s)
--- PASS: TestGetUserCredentialsDropsUndecryptableSecretsWithoutFailingThePage (0.00s)
--- PASS: TestUpdateUserCredentialWithoutSecretKeepsTheStoredOne (0.00s)
--- PASS: TestDecryptCredentialEnvelopeReadsAHandRolledV1Envelope (0.00s)
--- PASS: TestUpdateUserCredentialSurvivesAnUndecryptableStoredSecret (0.00s)
--- PASS: TestCredentialPayload (0.00s)
--- PASS: TestCredentialAuthSecret (0.00s)
--- PASS: TestSaveUserCredentialReturnsPopulatedCredential (0.00s)
--- PASS: TestGetUserCredentialsRejectsUnsanitizedOrderInput (0.00s)
--- PASS: TestK8sContextFromConnectionCredentialShapes (0.00s)
--- PASS: TestRemoteProviderDeleteUserCredential_SendsCanonicalCredentialIdParam (0.00s)
--- PASS: TestSeedConnectionsPreservesUserAttachedCredential (3.56s)
--- PASS: TestSeedUpdatesForClearsOwnerButKeepsCredential (0.00s)
PASS
ok  	github.com/meshery/meshery/server/models	5.678s

$ go test ./handlers/ -run "Credential|ConnectionRegister" -count=1 -v   (top-level results)
--- PASS: TestConnectionRegisterRejectsReservedCredentialProperty (0.00s)
--- PASS: TestConnectionRegisterUntaggedCredentialFailuresStayInternal (0.00s)
--- PASS: TestCredentialAPIReturnsPlaintextWhileTheRowIsCiphertext (0.00s)
--- PASS: TestCredentialAPIServesAPreEncryptionRowUnchanged (0.00s)
--- PASS: TestCredentialAPIRenameKeepsTheStoredSecret (0.00s)
--- PASS: TestDeleteUserCredential_QueryParamContract (0.00s)
--- PASS: TestDeleteUserCredential_RejectsMissingOrInvalidID (0.00s)
--- PASS: TestSaveUserCredential_ClientSuppliedUserIdCannotOverride (0.00s)
--- PASS: TestUpdateUserCredential_ClientSuppliedUserIdCannotOverride (0.00s)
--- PASS: TestCredentialWriteFailureStatusCodes (0.00s)
--- PASS: TestGrafanaClientForConnectionCredentialShapes (0.01s)
--- PASS: TestPrometheusClientForConnectionCredentialShapes (0.01s)
PASS
ok  	github.com/meshery/meshery/server/handlers	1.705s

$ go test ./models/httputil/... ./machines/... -count=1   (packages touched by the error-chain fix)
ok  	github.com/meshery/meshery/server/models/httputil	0.445s
ok  	github.com/meshery/meshery/server/machines	1.340s
?   	github.com/meshery/meshery/server/machines/grafana	[no test files]
ok  	github.com/meshery/meshery/server/machines/helpers	2.549s
ok  	github.com/meshery/meshery/server/machines/kubernetes	1.401s
?   	github.com/meshery/meshery/server/machines/prometheus	[no test files]
```
</details>
<details>
<summary>Evidence: The crux: what the API returns vs what is on disk (live server, real SQLite)</summary>

```text
$ curl -s -X POST http://localhost:9099/api/integrations/credentials -d '{"name":"prod-grafana","type":"grafana","secret":{"grafanaURL":"https://grafana.example.com","grafanaAPIKey":"eyJrIjoiU3VwZXJTZWNyZXRHcmFmYW5hS2V5In0"}}'
HTTP 201

GET /api/integrations/credentials (what the UI receives)
prod-grafana {"grafanaAPIKey": "eyJrIjoiU3VwZXJTZWNyZXRHcmFmYW5hS2V5In0", "grafanaURL": "https://grafana.example.com"}

$ sqlite3 mesherydb.sql "select name, secret from credentials;"
prod-grafana|{"__mesheryEncryptedSecret":"meshery.enc.v1:bb18a7366a61:0XwXXrHS8v1sHCXzMo7e3tao13L4SgOoYzChXosH56/jxhUKebB58NQm0wyis+GnNCuv2wo6ZmhIkZmqJ+ggRDOijRK5m1DMxhCE9DuAxRsxuD/otOXDy/JeFYXkIXMQApbA2hv2FwBKV45vre2QXjh9ud8ZsBk9/Y5IHpHOL7DRNEqiRBF71SGKBTZR0Yx9L9cS+SaNqIPHJ/Cr5Ng="}

$ strings -a mesherydb.sql mesherydb.sql-wal | grep -c 'eyJrIjoiU3VwZXJTZWNyZXRHcmFmYW5hS2V5In0'
0 <- the API key is not recoverable from the datastore
```
</details>
<details>
<summary>Evidence: Forged encryption marker is rejected 400, not stored</summary>

```text
$ curl -s -i -X POST /api/integrations/credentials -d '{"name":"forged","type":"grafana","secret":{"__mesheryEncryptedSecret":"meshery.enc.v1:aa:bb","grafanaAPIKey":"attacker-controlled"}}'
HTTP/1.1 400 Bad Request
{"error":"Failed to save user credential","code":"meshery-server-1390", ... "the credential secret carries __mesheryEncryptedSecret, which is reserved for Meshery's own encryption envelope and may not be supplied by a client ... failure carries HTTP status 400 Bad Request"}

$ sqlite3 mesherydb.sql "select name from credentials;"
prod-grafana <- the forged credential was not persisted
$ strings -a mesherydb.sql* | grep -c 'attacker-controlled'
0 <- and none of it landed in the clear
```
</details>
<details>
<summary>Evidence: SQLite freed-page residue reproduced on a real datastore, and cleared by VACUUM</summary>

```text
Both rows are ciphertext now. Searching the raw files for the OLD plaintext token:
mesherydb.sql 1 hit(s)
mesherydb.sql-wal 1 hit(s)
^ residue on freed pages - the row itself no longer contains it

$ sqlite3 mesherydb.sql 'VACUUM;'
After VACUUM:
mesherydb.sql 0 hit(s)
mesherydb.sql-wal 0 hit(s)

and the credentials still read after the compaction:
prod-grafana {"grafanaAPIKey": "...", "grafanaURL": "https://grafana.example.com"}
legacy-prometheus-renamed {"credentialName": "legacy-prometheus", "secret": {"apiKey": "LEGACY-PLAINTEXT-TOKEN-9k2"}}
```
</details>
<details>
<summary>Evidence: Key mismatch: degraded list vs hard failure, and where each key id appears</summary>

```text
(same datastore, opened by a server built with NO token - the dev_token default)

GET /api/integrations/credentials -> HTTP 200
prod-grafana <no secret field in the response>
legacy-prometheus-renamed <no secret field in the response>
total=2

server log (server-side only):
credential b71b5df8-... could not be decrypted; this Meshery Server build derives credential key 5ded4cb6e9d4: credential ciphertext names key bb18a7366a61, which this Meshery Server build does not derive

GET /api/integrations/credentials/b71b5df8-... -> HTTP 500
{"code":"meshery-server-1389", ... "credential ciphertext names key bb18a7366a61, which this Meshery Server build does not derive"}
^ names only the STORED key id; the active key id (5ded4cb6e9d4) never leaves the server
```
</details>

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>✅ **intent** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Rebase** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Review** - 3 infos</summary>

- ℹ️ `server/models/credential_encryption.go:198` - The AES-GCM envelope carries no associated data binding it to the row it belongs to: `ring.aead.Seal(nil, nil, plaintext, nil)` authenticates only the ciphertext, not the credential id, user id or type. A `secret` column value therefore stays valid if copied onto a different credential row, or if an older row&#39;s envelope is restored from a backup - `decryptCredentialEnvelope` and `DecryptCredentialSecret` will both report success. This is outside the documented threat model (an attacker with write access to `mesherydb.sql` already controls the local install, which is single-user), so it is not a defect in what this delivers - recording it so that whenever a `meshery.enc.v2` is cut, passing the credential id as AAD is on the list. Changing it now would break the v1 on-disk contract the intent pins.
- ℹ️ `server/models/credential_secret.go:257` - `ErrCredentialDecrypt` is used for two different failures. In `credential_encryption.go:234` it wraps an AEAD `Open` failure, where its probable cause (&#34;The stored ciphertext has been altered or truncated since it was written&#34;) and remediation (&#34;Restore the datastore from a known-good backup&#34;) are correct. Here it wraps a `json.Unmarshal` failure of the *already authenticated* plaintext - GCM has passed, so the bytes provably were not altered or truncated; the row was sealed by a Meshery version whose envelope this one cannot parse. That is exactly `ErrMalformedCredentialEnvelope`&#39;s stated cause, and the backup advice is actively misleading for it. Swapping this one call to `ErrMalformedCredentialEnvelope(&#34;sealed credential envelope is not valid JSON&#34;)` keeps 1483 meaning &#34;the ciphertext is damaged&#34;.
- ℹ️ `server/models/default_local_provider.go:1770` - The best-effort decrypt added to `DeleteUserCredential` has no observable effect: its only caller, `handlers.DeleteUserCredential`, discards the returned credential (`_, err = provider.DeleteUserCredential(...)`) and responds with a fixed body. On a key-mismatch row its sole outcome is an extra error log line per delete. It is defensible as provider-interface hygiene (the method is exported on the `Provider` interface and a future caller would otherwise get a raw envelope), and the comment says as much - flagging it only so it is not mistaken for a step the delete path depends on.

</details>

<details>
<summary>✅ **Test** - passed</summary>

✅ No issues found.
- <code>`go test ./models/ -run &#39;Credential|Encrypt|Decrypt|Keyring&#39; -count=1 -v` - 29 top-level tests pass (round trip, four persisted shapes, reserved-property rejection, hand-rolled v1 envelope compatibility, key mismatch, degradation paths)</code>
- <code>`go test ./handlers/ -run &#39;Credential|ConnectionRegister&#39; -count=1 -v` - 12 tests pass, including `TestConnectionRegisterRejectsReservedCredentialProperty` (400 + 0 rows) and `TestConnectionRegisterUntaggedCredentialFailuresStayInternal` (untagged failures stay 500)</code>
- <code>`go test ./models/httputil/... ./machines/... -count=1` - packages touched by the errors.Join error-chain fix</code>
- <code>Live server: built `./cmd` twice (`-ldflags &#39;-X main.globalTokenForAnonymousResults=demo_release_build_token&#39;` and with no token) and ran with `PROVIDER=None PORT=9099 USER_DATA_FOLDER=&lt;tmp&gt; SKIP_COMP_GEN=true DEBUG=true`</code>
- <code>`curl -X POST /api/integrations/credentials` then `sqlite3 mesherydb.sql &#39;select name, secret from credentials&#39;` - API plaintext vs ciphertext row</code>
- <code>`strings -a mesherydb.sql mesherydb.sql-wal | grep -c &#39;&lt;grafana api key&gt;&#39;` - 0 hits across the whole datastore</code>
- <code>`curl -X POST /api/integrations/credentials` with `__mesheryEncryptedSecret` in the body - HTTP 400 meshery-server-1390, 0 rows added, no `attacker-controlled` residue</code>
- `Seeded a pre-encryption legacy-nested plaintext row via sqlite3, restarted the server, listed credentials - reads with no migration`
- <code>`curl -X PUT /api/integrations/credentials` with no `secret` - HTTP 200, rename applied, stored secret preserved</code>
- <code>`curl -X PUT /api/integrations/credentials` re-supplying the secret - row converts to an envelope with `credentialName` still in the clear</code>
- <code>`strings -a` scan for the legacy plaintext token after conversion (residue reproduced), then `sqlite3 mesherydb.sql &#39;VACUUM;&#39;` and rescan (0 hits), then re-listed credentials</code>
- <code>Started the dev_token binary against the same datastore - `GET /api/integrations/credentials` 200 with the secret omitted, `GET /api/integrations/credentials/{id}` 500 naming only the stored key id, server log naming credential id + active key id</code>
- <code>`GODEBUG=fips140=only go test ./models/ -run TestEncryptDecryptCredentialSecretRoundTrip` - confirms the pre-existing panic at server/models/user.go:28</code>
- <code>Attempted the register route live (`POST /api/integrations/connections/register` with status connect) - refused by the state machine at DISCOVERED before the credential write, recorded in the transcript</code>

</details>

<details>
<summary>⚠️ **Document** - 1 info</summary>

- ℹ️ `docs/content/en/concepts/logical/credentials.md:79` - Judgment call left as-is: the VACUUM remedy in credentials.md tells users to compact ~/.meshery/config/mesherydb.sql with the server stopped, but the datastore is opened with _journal_mode=WAL (server/models/sql-utils.go:102). A clean shutdown checkpoints and removes the -wal sidecar, so the documented procedure is correct as written; after an unclean shutdown the -wal file can still hold the earlier plaintext and VACUUM alone would not reclaim it. I did not add that qualifier - it narrows an already-hedged security note and expanding it was outside what this change made stale. Worth a follow-up if maintainers want the guidance to cover crash-recovery state.

</details>

<details>
<summary>⚠️ **Lint** - 1 info</summary>

- ℹ️ `server/models/default_local_provider_test.go:73` - golangci-lint reports six pre-existing staticcheck SA5011 findings in server/models test files (default_local_provider_test.go:73/76, meshsync_register_queue_test.go:15/18, preference_map_persister_test.go:32/36). All are on unchanged lines that predate this change, and all are false positives - each guard ends in t.Fatal, which terminates the goroutine, so the flagged dereference is unreachable. CI hides them via only-new-issues on pull_request, so they do not gate this change. Left unfixed because this phase must not modify tests; resolving them would mean adding //nolint:staticcheck or restructuring the assertions.

</details>

<details>
<summary>✅ **Push** - passed</summary>

✅ No issues found.

</details>

## Summary by CodeRabbit

- **Security**
  - Credential secrets are now encrypted at rest using authenticated encryption.
  - Existing plaintext credentials remain readable and are upgraded to encrypted storage when updated.
  - Secrets are decrypted before being returned to authorized callers and are excluded from server logs.
  - Remote providers remain responsible for protecting credentials they receive.

- **Bug Fixes**
  - Credential and connection-registration errors now return appropriate client-facing HTTP statuses.
  - Added clearer diagnostics for invalid, corrupted, or incompatible encrypted credentials.

- **Documentation**
  - Expanded guidance on credential security, backups, key rotation, provider handling, and error behavior.


server/handlers/credentials_handlers_test.go
```diff
@@ -32,12 +38,18 @@ func newCredentialSpyProvider() *credentialSpyProvider {
 func (m *credentialSpyProvider) SaveUserCredential(_ string, c *models.Credential) (*models.Credential, error) {
 	cp := *c
 	m.observedSave.Store(&cp)
+	if m.saveErr != nil {
+		return nil, fmt.Errorf("error saving user credentials: %w", m.saveErr)
```
lekaf974, line 42: wondering if it should be a meshkit error ?

server/handlers/credentials_handlers_test.go
```diff
@@ -32,12 +38,18 @@ func newCredentialSpyProvider() *credentialSpyProvider {
...
+	}
 	return c, nil
 }
 
 func (m *credentialSpyProvider) UpdateUserCredential(_ *http.Request, c *models.Credential) (*models.Credential, error) {
 	cp := *c
 	m.observedUpdate.Store(&cp)
+	if m.updateErr != nil {
+		return nil, fmt.Errorf("error updating user credential: %w", m.updateErr)
```
lekaf974, line 51: same here

server/models/httputil/provider_status.go
```diff
@@ -21,12 +27,13 @@ import (
 type ProviderStatus int
 
 func (s ProviderStatus) Error() string {
-	return fmt.Sprintf("remote provider responded with HTTP %d %s", int(s), http.StatusText(int(s)))
+	return fmt.Sprintf("failure carries HTTP status %d %s", int(s), http.StatusText(int(s)))
```
lekaf974, line 30: should we convert to meshkit error ?

server/models/credential_secret.go
```diff
@@ -124,3 +126,161 @@ func authSecretFromPayload(payload map[string]interface{}) string {
...
+	}
+
+	plaintext, err := decryptCredentialEnvelope(ciphertext)
+	if err != nil {
+		return nil, err
+	}
+
+	var envelope credentialSecretEnvelope
+	if err := json.Unmarshal(plaintext, &envelope); err != nil {
+		return nil, ErrCredentialDecrypt(err)
```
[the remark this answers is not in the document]
lekaf974: this seems legit

server/models/default_local_provider.go
```diff
@@ -1575,10 +1575,30 @@ func (l *DefaultLocalProvider) SaveUserCredential(token string, credential *Cred
...
 	}
-	result := l.GetGenericPersister().Table("credentials").Create(&credential)
+
+	// Persist the secret encrypted. EncryptCredentialSecret returns a fresh map,
+	// so the caller's in-memory credential (and the connection payload it was
+	// built from) keeps its plaintext secret, and only the row is ciphertext.
+	stored := *credential
+	encryptedSecret, err := EncryptCredentialSecret(credential.Secret)
+	if err != nil {
+		return nil, fmt.Errorf("error saving user credentials: %w", err)
```
lekaf974, line 1585: EncryptCredentialSecret return a meshkit error should we return it as is ?

server/models/default_local_provider.go
```diff
@@ -1635,18 +1671,89 @@ func (l *DefaultLocalProvider) GetUserCredentials(_ *http.Request, userID string
 }
 
 func (l *DefaultLocalProvider) UpdateUserCredential(_ *http.Request, credential *Credential) (*Credential, error) {
+	// Encrypt into a copy for the same reason SaveUserCredential does: the
+	// caller keeps the credential it handed in, and only the row is ciphertext.
+	stored := *credential
+	encryptedSecret, err := EncryptCredentialSecret(credential.Secret)
+	if err != nil {
+		return nil, fmt.Errorf("error updating user credential: %w", err)
```
lekaf974, line 1679: same EncryptCredentialSecret return a meshkit error should we return it as is ?

server/models/default_local_provider.go
```diff
@@ -1635,18 +1671,89 @@ func (l *DefaultLocalProvider) GetUserCredentials(_ *http.Request, userID string
...
+
+// credentialDecryptionFailure renders a credential decryption failure for the
+// server log, naming the key identifier this build derives.
+//
+// That identifier stays out of the error returned to the caller: it is a
+// function of the build-time token, and the API error already names the key the
+// stored ciphertext was written under, which is the half an operator acts on.
+// Pairing the two is a diagnosis the operator does from the server log.
+func credentialDecryptionFailure(credentialID core.Uuid, err error) error {
+	return fmt.Errorf("credential %s could not be decrypted; this Meshery Server build derives credential key %s: %w", credentialID, activeCredentialKeyID(), err)
```
lekaf974, line 1754: should we return a meshkit error ?

lekaf974 (commented):
Added some comments

server/models/remote_provider.go
```diff
@@ -5063,6 +5063,13 @@ func (l *RemoteProvider) GetKubeClient() *mesherykube.Client {
 }
 
 // SaveCredential - to save a creadential for an integration
+//
+// The secret is sent to the Remote Provider as plaintext, over TLS, and is
+// deliberately NOT put through EncryptCredentialSecret. At-rest encryption
+// belongs to whoever owns the datastore: here that is the Remote Provider, which
+// has no way to derive this binary's key and would be handed an envelope it
+// could never open. Only DefaultLocalProvider, which owns Meshery's own
+// datastore, encrypts. See credential_encryption.go.
```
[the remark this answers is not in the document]
leecalcote: No, doing so will hinder local development of remote providers.

server/models/default_local_provider.go
leecalcote: Good eye, @lekaf974!

PR #21521, opened 2026-08-20
[mesheryctl] Validate config path is not empty

### LINKED ISSUE #2867: [Feature] Prevent mesheryctl panic/crash by adding Go path validation check
### Current Behavior
While exploring mesheryctl logic, I noticed that if a configuration context path evaluates to an empty string (""), the CLI tool doesn't handle it gracefully. It throws raw system panic logs or a sudden crash instead of properly warning the user.

### Desired Behavior
I want to inject a quick, structural safety check (if-validation) in the Go engine to catch any empty configurations right at the entry point. Instead of crashing, it should return a clean error like: 
"error: Meshery configuration context path cannot be empty."

I am currently an incoming Cybersecurity student applying for the CNCF LFX Mentorship (GitHub: https://github.com/alina-cloudsec). I have already built basic firewall simulations and RBAC logic locally, and I would like to turn that logic into high-value code here. 

Please assign this issue to me so I can submit a clean Pull Request immediately.

### Screenshots / Mockups
(No screenshots needed for this CLI logic change.)

---
<h4>Contributor [Guides](https://docs.meshery.io/project/contributing) and Resources</h4>

- 🖥 [Contributing to Meshery Website](https://github.com/meshery/meshery.io#contributing-to-the-mesheryio-website)

atanishka308:
**Notes for Reviewers**

- This PR fixes   https://github.com/meshery/meshery.io/issues/2867 (filed in meshery.io, but mesheryctl's code lives in this repo)
- Root cause: `viper.SetConfigFile()` silently ignores an empty string instead of erroring, so an empty `--config` path slips through and later surfaces as a raw error like `Error: stat : no such file or directory`
- Fix: added `config.ValidateConfigPath()`, called in `initConfig()` (root.go) and in `AddTokenToConfig`, `DeleteTokenFromConfig`, `SetTokenToConfig`, `AddContextToConfig` (config.go)
- Now returns: `error: Meshery configuration context path cannot be empty`
- Verified with: `mesheryctl --config="" system context create test-ctx`
- `go build ./...` and `go vet ./...` both pass clean

**Before:**
BEFORE

**After:**
AFTER

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for empty or whitespace-only configuration paths.
  * mesheryctl now stops initialization with a clear error and remediation guidance when the configuration path is invalid.
  * Configuration changes are prevented from proceeding when the path is missing or invalid.


mesheryctl/internal/cli/root/config/config.go
```diff
@@ -54,6 +54,18 @@ type Context struct {
 	EnvVars    map[string]any `yaml:"env,omitempty" mapstructure:"env,omitempty"`
 }
 
+// ValidateConfigPath checks that a Meshery configuration context path is
+// not empty (or all whitespace) before it is handed off to viper or the
+// filesystem. Left unchecked, an empty path is silently ignored by
+// viper.SetConfigFile and can surface later as a raw panic or a confusing
+// low-level OS error instead of a clean, actionable message.
+func ValidateConfigPath(configPath string) error {
```
banana-three-join: The name of the function is a bit misleading. The validation is only scoped to the emptiness of the config file but due the name of the function, this intent is not clear.

banana-three-join (changes requested):

atanishka308:
Good catch — renamed to `EnsureConfigPathNotEmpty` to make it clear the check only covers emptiness, not general path validity. Updated across all call sites in config.go and root.go.

PR #21530, opened 2026-08-21
Fix Welcome First-Time Contributors workflow failure

sangramrath:
The Welcome First-Time Contributors workflow has been failing with the below error:
Screenshot 2026-08-21 at 7 32 05 AM

The cause if the error is because the workflow is trying to use `secrets` in a context GitHub Actions does not allow.

**Notes for Reviewers**
Removing condition checking for `SLACK_BOT_TOKEN` in failure alert step as **this token is always present**.

- This PR fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved failure notifications for the first-time contributor workflow by ensuring alerts are attempted whenever the workflow fails.
  * Notification errors continue to be handled without interrupting the workflow.

### DIFF
--- a/.github/workflows/welcome-first-time-contributors.yml
+++ b/.github/workflows/welcome-first-time-contributors.yml
@@ -144,7 +144,6 @@ jobs:
       - name: 🚨 Alert on Failure
         if: failure()
         uses: slackapi/slack-github-action@v2.1.1
-        # Only run if the Slack bot token secret is configured; skip gracefully otherwise.
         continue-on-error: true
         with:
           method: chat.postMessage
--- a/.github/workflows/welcome-first-time-contributors.yml
+++ b/.github/workflows/welcome-first-time-contributors.yml
@@ -142,7 +142,7 @@ jobs:
       # is alerted to any automation breakage promptly.
       # -----------------------------------------------------------------------
       - name: 🚨 Alert on Failure
-        if: failure() && secrets.SLACK_BOT_TOKEN != ''
+        if: failure()
         uses: slackapi/slack-github-action@v2.1.1
         # Only run if the Slack bot token secret is configured; skip gracefully otherwise.
         continue-on-error: true


ishwar170695 (approved):
LGTM. The `secrets.SLACK_BOT_TOKEN` check can't be used in the `if:` condition, so` if: failure()` is the right fix. The Slack config is unchanged, and `continue-on-error:` true makes sure a Slack failure doesn't fail the workflow. I don't see any security issues with this change.

.github/workflows/welcome-first-time-contributors.yml
Bharath314, line 16: I think this line should be changed to `pull_request` or [this line](https://github.com/meshery/meshery/pull/21530/changes#diff-857c93d0fc5445a13afebfc7ec0a39563266e55a051e1c17d022ef1148f3f8afL103) should be changed to `if: github.event_name == 'pull_request_target'`. I'm not sure which is the correct change but I think the First PR step won't trigger currently due to mismatch in event name.

Bharath314 (changes requested):

emphor11 (approved):
LGTM! The fix looks correct and the checks are passing.
One separate follow up: could we open an issue for the existing pull_request_target / github.event_name == 'pull_request' mismatch?

Junnygram (approved):

.github/workflows/welcome-first-time-contributors.yml
sangramrath: @emphor11 is this what you are referring to as well?

dhruveshmishra (approved):
LGTM! This fixes the GitHub Actions syntax error where `secrets` cannot be used directly in step-level `if:` conditions. `continue-on-error: true` handles the fallback properly.

Katotodan (approved):

.github/workflows/welcome-first-time-contributors.yml
Bharath314: I just checked and realised that the hyperlink wasn't working as expected. The link was supposed to point to line 103 which is currently `if: github.event_name == 'pull_request'`. Since the event types defined here are only `issues` and `pull_request_target`, I think that if block will never execute.

banana-three-join: I agree with Bharath. Also believe the conditional should be changed to `if: github.event_name == 'pull_request_target'`. This is due to the `pull_request_target` event providing the necessary permissions for `actions/first-interaction` to work. Looking at the history of the action, the only working run was the [first one](https://github.com/meshery/meshery/actions/runs/31912628022) which immediately got a fix afterwards due the `actions/first-interaction` step not working, leaving me to think that the `pull_request` was left on accident so it should be changed to `pull_request_target`. 

Also, it should be safe to use since there aren't any checkouts in the workflow but the hashed version for `actions/first-interaction` should be used instead just to be on the safe side.

hiyach28 (approved):

yi-nuo426:
Correct fix. .github/workflows/welcome-first-time-contributors.yml:145 now uses the supported failure() condition rather than referencing secrets directly in if, preventing GitHub Actions expression validation from rejecting the workflow. The token remains passed only through the action input at line 150, and continue-on-error: true ensures a Slack delivery failure cannot mask the original workflow failure. No changes requested.

PR #21556, opened 2026-08-21
[UI] Fix extensions failing to render on refresh

### LINKED ISSUE #21531: [UI] Extensions fail to load on refresh
### Current Behavior
The extension contents are not rendered after refreshing the Extensions page. The page remains without the extension content until the Extensions button is clicked again, after which the contents load correctly.

### Desired Behavior
The extension contents should be rendered correctly after refreshing the Extensions page without requiring any additional interaction.

### Screenshots/Logs

https://github.com/user-attachments/assets/e7284973-3c65-44d0-b259-4c3aa5425e9f

### Environment

- Host OS: Mac Linux Windows
- Platform: Docker Kubernetes

---

ayanasarkar:
Fixes #21531

## Problem
On refresh, the Extensions page relied on `extensionType` and `providerCapabilities` state to decide what to render. These were populated asynchronously, so on a hard refresh the page would briefly (or indefinitely) fall through to a state where nothing rendered until the user re-triggered navigation by clicking into Extensions again.

## Fix
Derive render-readiness directly from `matchedExtension` (computed via `useMemo` from `providerCapabilities`) instead of from separately-updated state. The page now waits for `isReady = !isLoading && !!matchedExtension` before rendering, so it renders correctly on first load/refresh without requiring extra interaction. Also removed the now-unused `CircularProgress` import and loading branch.

**Notes for Reviewers**
- This PR fixes #21531
- Verified locally: refreshing the Extensions page now loads content immediately without needing to click into Extensions again. Video of  the desired behavior is attached.

RECORDING AFTER CHANGES:

https://github.com/user-attachments/assets/2f4b1800-a4c5-4122-8116-c34942c28973

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved extension rendering while provider information is loading.
  - Prevented extensions from appearing before loading completes.
  - Displayed a permission error page for unmatched extension routes.
  - Improved handling of unavailable or inaccessible extensions to avoid blank screens.
  - Preserved sandboxed rendering for navigator extensions and remote components.
  - Ensured loading and error states transition reliably after provider information is received.


ayanasarkar:
@Sbragul26  could u please review this once

dhruveshmishra (commented):
Could we verify that providerCapabilities is available before isLoading becomes false? Otherwise, matchedExtension could still be null and nothing would be rendered.

ayanasarkar:
> Could we verify that providerCapabilities is available before isLoading becomes false? Otherwise, matchedExtension could still be null and nothing would be rendered.

Good catch — clarifying: isLoading only flips to false once the RTK Query
request has settled (success or error), so there's no race where
providerCapabilities is still pending while isLoading is false.

You're right there was a real gap though: if the query errors, or the URL
doesn't match any registered extension, matchedExtension stayed null
indefinitely and the user saw a persistent spinner with no feedback.
Pushed a fix — isReady now also resolves on isError, and we show
DefaultError instead of an infinite spinner in that case.

ui/pages/extension/[...component].tsx
```diff
@@ -100,6 +108,8 @@ function RemoteExtension() {
     };
   }, [dispatch, matchedExtension]);
 
+  const isReady = !isLoading && (!!matchedExtension || isError);
+
```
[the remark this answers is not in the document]
Comment (ayanasarkar) [reply ui/pages/extension/[...component].tsx]: Good catch — simplified further. Readiness now just depends on the query
having settled (!isLoading), since matchedExtension will naturally be
null in both the error case and the "no match" case. Fallback now shows
DefaultError whenever there's no match, covering both scenarios.

dhruveshmishra (approved):
Thanks for the clarification and the additional fix. The loading and error cases are handled now. LGTM 👍

ayanasarkar:
@Sbragul26 could you please review this once

ui/pages/extension/[...component].tsx
```diff
@@ -100,6 +108,8 @@ function RemoteExtension() {
     };
   }, [dispatch, matchedExtension]);
 
+  const isReady = !isLoading;
```
Comment (Katotodan) [inline ui/pages/extension/[...component].tsx]: Since isReady is used only once, there is no need to declare it.
Use only the isLoading variable

Katotodan (changes requested):

Katotodan:
Here is a screen recording, when I run this PR locally.
I don't know if I am the only one encountering this error when running your PR locally. 

https://github.com/user-attachments/assets/dc632fad-5767-495d-8fc3-097a91c7504a

ui/pages/extension/[...component].tsx
Comment (ayanasarkar) [reply ui/pages/extension/[...component].tsx]: Done — inlined isReady, removed the unused extensionType selector it was tangled up with too. Pushed.
 in the latest commit.

ayanasarkar:
> Here is a screen recording, when I run this PR locally. I don't know if I am the only one encountering this error when running your PR locally.
> 
>  Screen.Recording.2026-08-25.at.19.33.07.mov

Re-tested locally after the latest push — recorded it again, no errors on refresh this time. 

https://github.com/user-attachments/assets/4995d230-7ddf-4918-9e5e-ae2fa7b010bd

Let me know if you're still seeing the issue on your end after pulling latest.

ui/pages/extension/[...component].tsx
```diff
@@ -109,20 +116,21 @@ function RemoteExtension() {
...
-                type={extensionType}
+                type={matchedExtension.name}
                 Extension={(url) => RemoteComponent({ url })}
               />
             )}
           </NoSsr>
-        ) : !isLoading ? null : (
-          <CircularProgress />
+        ) : (
+          <DefaultError permissionKey={Keys.ExtensibilityViewExtensions} />
```
Comment (rishiraj38) [inline ui/pages/extension/[...component].tsx:133]: Don't use the default error here. We are already handling it here: https://github.com/meshery/meshery/pull/21556/changes#diff-47d04d2fa0cd33559397b752ddaeaaf95938eca20fcaa004ce28e7897488b8e9R110-R111

rishiraj38 (changes requested):

ui/pages/extension/[...component].tsx
Comment (ayanasarkar) [reply ui/pages/extension/[...component].tsx]: @rishiraj38 Good catch, agreed — that was reusing the permission-denied UI for an unrelated case. Removed it, now falls through to null since the permission check above already covers real access issues.

yi-nuo426:
ui/pages/extension/[...component].tsx:116-136 now renders from matchedExtension, derived directly from fetched capabilities and the current route, rather than waiting for asynchronously updated Redux extensionType. Loading remains driven by the capabilities query, so a hard refresh no longer depends on a second render after dispatch. No changes requested.

ayanasarkar:
> ui/pages/extension/[...component].tsx:116-136 now renders from matchedExtension, derived directly from fetched capabilities and the current route, rather than waiting for asynchronously updated Redux extensionType. Loading remains driven by the capabilities query, so a hard refresh no longer depends on a second render after dispatch. No changes requested.

thanks for confirming, @yi-nuo426 appreciate you taking a look.

@Katotodan @rishiraj38 — looping back on your review comments specifically, since I believe both are addressed now:

- @Katotodan: the runtime error from your recording should be fixed in 9487f06 (inlined `isReady`, removed the unused `extensionType` selector). Re-tested locally after that push and didn't see it reproduce — let me know if it's still happening after pulling latest.
- @rishiraj38: removed the `DefaultError` reuse you flagged, in ec0f06e — it now falls through to `null` for the unmatched-route case, since the permission check above already covers real access issues.

Could either of you re-review when you get a chance so we can get this unblocked? Happy to make further changes if anything's still off. Thanks for the patience on this one!

PR #21563, opened 2026-08-22
[mesheryctl] Fix connection leak in dashboard port-forward keep-alive

HariomPtdr:
**Notes for Reviewers**

`keepConnectionAlive` is driven by a 10 second ticker for as long as `mesheryctl system dashboard --port-forward` is running:

```go
ticker := time.NewTicker(10 * time.Second)
...
case <-ticker.C:
    keepConnectionAlive(mesheryURL)
```

It discarded the response without closing it:

```go
_, err := http.Get(url)
if err != nil {
    utils.Log.Debugf("connection request failed %v", err)
}
utils.Log.Debug("connection request success")
```

Because the body is never drained or closed, the connection is not returned to the transport's idle pool, so every tick dials a new one and leaks the old one for the life of the command — roughly 360 connections per hour of port forwarding. That is the opposite of what the function is there to do: it is meant to hold one connection open, not accumulate them.

There is a second, smaller problem in the same function: there is no `return` after the error branch, so a failed request also logs `connection request success`.

**Fix**

Drain and close the body so the connection is reused, and return early on error.

**Tests**

Added `dashboard_test.go` with two cases:

- `TestKeepConnectionAliveReusesTheConnection` — starts an `httptest` server, tracks `http.StateNew` via `ConnState`, calls the function 5 times and asserts exactly one connection was opened.
- `TestKeepConnectionAliveDoesNotLogSuccessOnError` — points at a closed listener and asserts the failure is logged and success is not.

Both fail on `master` and pass with this change:

```
$ go test ./internal/cli/root/system/ -run TestKeepConnectionAlive     # before
    dashboard_test.go:59: expected 5 calls to share a single connection, but 5 were opened
--- FAIL: TestKeepConnectionAliveReusesTheConnection
--- FAIL: TestKeepConnectionAliveDoesNotLogSuccessOnError

$ go test ./internal/cli/root/system/ -run TestKeepConnectionAlive     # after
--- PASS: TestKeepConnectionAliveReusesTheConnection (0.00s)
--- PASS: TestKeepConnectionAliveDoesNotLogSuccessOnError (0.00s)
ok      github.com/meshery/meshery/mesheryctl/internal/cli/root/system  1.536s
```

`gofmt -l` and `go vet ./internal/cli/root/system/` are both clean. The remaining failures in this package are `TestSystemCmdIntegration`, which needs a live Docker/Kubernetes environment and fails identically on an unmodified `master` checkout here.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dashboard connection handling to prevent connection leaks during ongoing sessions.
  * Added clearer failure handling when dashboard requests or responses cannot be processed.

* **Tests**
  * Added coverage for connection reuse and request failure scenarios.


lekaf974 (approved):
LGTM

PR #21564, opened 2026-08-22
fix(mesheryctl): add 10s timeout to http.Client in system check

### LINKED ISSUE #19994: [mesheryctl] fix: add timeout to http.Client in system/check.go to prevent CLI from hanging
**Current Behavior**
`mesheryctl/internal/cli/root/system/check.go` creates `http.Client{}` 
without a timeout at two locations (lines ~503 and ~695). Health check 
requests can hang indefinitely if the Meshery server is unresponsive, 
blocking the CLI with no feedback to the user.

**Expected Behavior**
HTTP clients in `runServerChecks` and `runAdapterHealthChecks` should 
have a timeout (e.g., 10 seconds) so `mesheryctl system check` returns 
a clear error instead of hanging forever.

**Fix**
Change `client := &http.Client{}` to 
`client := &http.Client{Timeout: 10 * time.Second}` at both locations.

**Affected Component(s)**
- mesheryctl

**Steps To Reproduce**
Run `mesheryctl system check` against an unresponsive endpoint — 
the command never returns.

LavyaTandel:
## What
Add a 10s timeout to the HTTP clients in `runServerChecks` and `runAdapterHealthChecks` so `mesheryctl system check` fails fast instead of hanging forever against an unresponsive server.

## Evidence
mesheryctl/internal/cli/root/system/check.go:504 and :715 create `&http.Client{}` with no timeout.

## Fix
`client := &http.Client{Timeout: 10 * time.Second}` at both sites.

## Test plan
`go build ./...` passes. Health check against a non-listening endpoint returns an error within 10s rather than blocking.

## Duplicate Scan
No open PR references #19994.

## Risk
Only changes request timeout; no behavior change when the server responds.

Closes #19994

## Summary by CodeRabbit

* **Bug Fixes**
  * Added 10-second timeouts to server version and adapter health checks.
  * Prevented checks from hanging indefinitely when a service is unresponsive.

### DIFF
--- a/mesheryctl/internal/cli/root/system/check.go
+++ b/mesheryctl/internal/cli/root/system/check.go
@@ -23,6 +23,7 @@ import (
 	"runtime"
 	"slices"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 
@@ -501,7 +502,7 @@ func (hc *HealthChecker) runMesheryVersionHealthChecks() error {
 		return err
 	}
 
-	client := &http.Client{}
+	client := &http.Client{Timeout: 10 * time.Second}
 	resp, err := client.Do(req)
 	// failed to fetch response for server version
 	if err != nil || resp.StatusCode != 200 {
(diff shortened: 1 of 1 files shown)


banana-three-join:
No. You haven't read the contributor guidelines (Didn't ask for the issue to be assigned to you nor signed your commit). Go do that first and then you'll be presented with an issue to fix.

PR #21568, opened 2026-08-22
fix(mesheryctl): close kubernetes log stream promptly after consumption

### LINKED ISSUE #21474: [mesheryctl] Close Kubernetes log streams promptly in system logs
### Current Behavior

`mesheryctl system logs` creates Kubernetes log streams while processing container logs. The stream is currently closed using `defer logs.Close()` within the container-processing flow.

Because Go's `defer` executes when the surrounding function returns, a stream created for one container can remain open while subsequent containers are being processed. This can result in multiple Kubernetes log streams being held concurrently for longer than necessary.

A similar stream-lifecycle consideration exists in `--follow` mode, where streams remain active while logs are being followed and need to be closed when the consumer finishes, encounters an error, or is cancelled.

### Desired Behavior

Each Kubernetes log stream should have a clear owner and should be closed as soon as its consumer is finished with it.

- Streams from completed containers should be released promptly.
- `--follow` streams should remain open while actively being followed.
- Streams should be closed when following terminates.
- Errors and context cancellation should also trigger cleanup.
- Existing `mesheryctl system logs` output and behavior should remain unchanged.

### Implementation

Refactor the log-stream lifecycle in `mesheryctl system logs` so that `logs.Close()` is tied to the lifetime of the stream consumer rather than the surrounding command function.

The implementation could use a smaller function scope for `defer logs.Close()`, explicit cleanup after stream consumption, or transfer stream ownership to the goroutine responsible for consuming it.

The chosen approach should preserve the existing normal and `--follow` behavior while avoiding premature stream closure, blocked consumers, goroutine leaks, or double-close issues.

### Acceptance Tests

- [ ] Kubernetes log streams are closed after their consumers finish.
- [ ] Streams from previously processed containers are not unnecessarily retained while later containers are processed.
- [ ] Multiple containers can be processed without unnecessarily accumulating active streams.
- [ ] `--follow` keeps streams open while they are actively being followed.
- [ ] Follow-mode streams are closed when following terminates.
- [ ] Streams are cleaned up when consumption returns an error.
- [ ] Streams are cleaned up when the associated context is cancelled.
- [ ] No goroutines remain blocked after stream consumption terminates.
- [ ] Existing `mesheryctl system logs` output and behavior remain unchanged.
- [ ] Regression tests are added for stream ownership and cleanup.

LavyaTandel:
## What
Close each Kubernetes log stream as soon as its consumer finishes instead of when the whole `mesheryctl system logs` command returns.

## Evidence
logs.go:231 `defer func(){ _ = logs.Close() }()` runs at function return, so streams for finished containers stay open across the pod loop.

## Fix
For non-follow: close after io.ReadAll. For --follow: close in the consumer goroutine on exit.

## Test plan
`go build ./...` passes; `mesheryctl system logs` opens and closes one stream per container; `--follow` closes on termination.

## Duplicate Scan
No open PR references #21474.

## Risk
Output identical; only stream lifecycle tightened.

Closes #21474

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes log stream cleanup for non-following logs.
  * Ensured following log streams close properly when streaming ends.

### DIFF
--- a/mesheryctl/internal/cli/root/system/logs.go
+++ b/mesheryctl/internal/cli/root/system/logs.go
@@ -228,10 +228,12 @@ mesheryctl system logs meshery-istio
 					if err != nil {
 						return err
 					}
-					defer func() { _ = logs.Close() }()
 					var logBuf []byte
 					if !systemLogsFlags.Follow {
 						logBuf, err = io.ReadAll(logs)
+						// ponytail: close this container's stream as soon as it is consumed,
+						// not when the whole pod loop returns — scoped defer.
+						_ = logs.Close()
 						if err != nil {
 							return utils.ErrReadResponseBody(err)
 						}
@@ -240,6 +242,7 @@ mesheryctl system logs meshery-istio
 						wg.Add(1)
 						go func() {
 							defer wg.Done()
+							defer func() { _ = logs.Close() }()
 							for {
 								buf := make([]byte, BYTE_SIZE)
 								numBytes, err := logs.Read(buf)


banana-three-join:
No. You haven't read the contributor guidelines (Didn't ask for the issue to be assigned to you nor signed your commit). Go do that first and then you'll be presented with an issue to fix.

banana-three-join:
Here. https://docs.meshery.io/project/contributing/

After you're done with that and showcase that you went through it, you can go ahead and properly request a good first issue to be assigned to you. If you require further help, here's the [slack](https://mesheryio.slack.com/ssb/redirect#/shared-invite/email) channel that we have to answer any of your inquires.

PR #21569, opened 2026-08-22
[mesheryctl] Make imgs-output-path optional in registry publish

### LINKED ISSUE #21558: [mesheryctl] Registry publish  fails when using the 4 arguments suggested in its official meshery example

### Current Behavior
The internal documentation (help text) for `mesheryctl registry publish` provides the following example for publishing to the Meshery Server:

    // Publish To Meshery
    mesheryctl registry publish meshery GoogleCredential GoogleSheetID [repo]/models

This example relies on passing exactly 4 arguments. However, the command's `PreRunE` validation block (`mesheryctl/internal/cli/root/registry/publish.go`) strictly enforces that exactly **5 arguments** are provided:

    if len(args) != 5 {
        return errors.New(utils.RegistryError("[ system, google sheet credential, sheet-id, models output path, imgs output path] are required...", "publish"))
    }

Because of this hardcoded `len(args) != 5` check, if a user runs the exact 4-argument command provided in the example, it immediately fails and demands an `imgs-output-path`. 

Furthermore, the `RunE` block blindly executes `imgsOutputPath = args[4]`, which means simply removing the `PreRunE` check without safely handling the argument bounds would cause a panic.

### Desired Behavior
There should be no mismatch between the official examples and the validation logic. 
- If `imgs-output-path` is truly optional for the `meshery` system, the validation and `RunE` code should gracefully handle 4 arguments. 
- If `imgs-output-path` is strictly mandatory for all systems, the example documentation needs to be updated to include it so users aren't misled.

### Screenshots/Logs
N/A

### Environment

- Host OS: Mac / Linux / Windows
- Platform: Docker / Kubernetes

---

Gurkaran18:
**Notes for Reviewers**

- This PR fixes #21558 

As the official help command examples show, publishing to the `meshery` server only strictly requires 4 arguments. The `PreRunE` logic used to blindly enforce `len(args) == 5`, immediately failing the exact example given to users. This PR relaxes the constraint to allow 4 or 5 arguments and conditionally initializes `imgsOutputPath` to prevent an out-of-bounds panic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the publish command to accept four or five positional arguments for the `meshery` system.
  * Required an image output path when publishing to `remote-provider` or `website`.
  * Kept the image output path optional for other supported publishing scenarios.
  * Improved invalid-argument errors with a consistent error code and remediation guidance.

* **Tests**
  * Added system-specific validation coverage for valid and invalid argument counts and image-path requirements.

* **Documentation**
  * Added a reference entry for invalid `registry publish` arguments.


vedant21-ctr (changes requested):
@Gurkaran18 
one issue is still left:

The current validation allows 4 or 5 arguments for all systems, but `remote-provider` and `website` still require `imgsOutputPath`. Could you make the validation system-specific so that:

- `meshery` accepts 4 or 5 arguments
- `remote-provider` and `website` require all 5 arguments

Please also add boundary tests for these argument combinations so we cover the regression.

ishwar170695 (commented):
One small thing beyond the existing comments: could we add a test for the 4-argument `meshery` case and the rejected 4-argument `remote-provider`/`website` cases? This is an argument-validation change, so it would help prevent the regression from coming back. Otherwise looks good to me.

Gurkaran18:
@ishwar170695 and @vedant21-ctr, thanks for the reviews.  I have pushed the changes.

I updated the validation so it checks based on the system name now. `meshery` allows 4 or 5 arguments, but `remote-provider` and `website` strictly require all 5. 

I also added tests for all these boundary cases in `publish_test.go` (11 test cases in total) and they are all passing locally. Also, I updated the tests to use the standard Go `testing` API instead of `testify`, and I've switched the validation errors to use a proper MeshKit error constructor as requested by coderabbitai. 

Let me know if something else needs to be Updated

mesheryctl/internal/cli/root/registry/publish.go
```diff
@@ -81,8 +81,15 @@ mesheryctl registry publish website "$CRED" 1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdw
...
+		// All systems require at least: system, credential, sheet-id, models-output-path.
+		if len(args) < 4 || len(args) > 5 {
+			return ErrPublishInvalidArgs("[ system, google sheet credential, sheet-id, models output path ] are required. [ imgs output path ] is optional for 'meshery' but required for 'remote-provider' and 'website'.\n\nUsage: \nmesheryctl registry publish [system] [google-sheet-credential] [sheet-id] [models-output-path] [imgs-output-path]\nmesheryctl registry publish [system] [google-sheet-credential] [sheet-id] [models-output-path] [imgs-output-path] -o [output-format]\nRun 'mesheryctl registry publish --help'")
+		}
+
+		// remote-provider and website always write icons, so imgs-output-path is mandatory.
+		sys := args[0]
+		if (sys == "remote-provider" || sys == "website") && (len(args) < 5 || args[4] == "") {
```
ishwar170695, line 95: Should we also validate unknown `system` values here? Right now an invalid system with 4 arguments passes `PreRunE` and only fails later in `RunE`.

Gurkaran18: Yes , Since the valid systems are strictly meshery, remote-provider, and website, we should check args[0] against those three. If it does not match any of them, we can immediately return a validation error. I can add this check to PreRunE along with a boundary test to ensure unknown systems are rejected. 

Let me know if you want me to go ahead and push this. @ishwar170695

ishwar170695 (approved):
Thanks for addressing the feedback. The system-specific validation and boundary tests look good, and the unknown system validation is now handled early as well. LGTM.

lekaf974:
IMO this need more refactoring because the command does not follow UX I think that those args should be flags as example see `registry generate` command

@leecalcote any thought ?

Gurkaran18:
@lekaf974 The flags for `system`, `google-sheet-credential`, `sheet-id`, `models-output-path`, and `imgs-output-path` are already defined in `publish.go`’s `init()` — they are just commented out.

There’s also a note from @theBeginner86 explaining that they switched to args because having five required flags made the command pretty cumbersome to type.

So it wasn’t really an oversight flags were considered and intentionally moved away from. I am fine with switching back if that’s what we decide, but I’d rather avoid reworking it only to change it again. I will wait for @leecalcote reply.

docs/data/errorref/mesheryctl_errors_export.json
lekaf974, line 1: this file is autogenerated and should not be committed

mesheryctl/internal/cli/root/registry/publish.go
lekaf974, line 1: Wondering if we should move to flags instead of args like it is done in other commands for UX consistency

ritzorama: Yes, ➕ . I agree.

vedant21-ctr (approved):

docs/data/errorref/mesheryctl_errors_export.json
lekaf974: @Gurkaran18 :arrow_up:

lekaf974 (changes requested):
remove unnecessary file to resolve conflict

Gurkaran18:
Thanks for the review @lekaf974!

Removed docs/data/errorref/mesheryctl_errors_export.json and merged the latest master, so the conflict is resolved. master had already taken mesheryctl-1255, so I renumbered ErrPublishInvalidArgs to mesheryctl-1256. Could you re-review?

Gurkaran18:
@ritzorama @lekaf974 I have switched registry publish to flags. It now uses --system, --spreadsheet-cred, --spreadsheet-id, --models-output-path and --imgs-output-path (plus the existing -o). The spreadsheet flag names match registry generate for consistency. --imgs-output-path is optional for meshery and required for remote-provider and website, and the old positional syntax now returns a clear error pointing to the flags. I updated the unit tests, the e2e bats tests and cmds.yml as well. Could you please take another look?

PR #21570, opened 2026-08-22
[mesheryctl] Close the response body in GetProviderInfo and IsMesheryRunning

HariomPtdr:
**Notes for Reviewers**

Follow-up to #21563, which fixed the same class of bug in `mesheryctl system dashboard`. Two more places in `pkg/utils` issue an HTTP request and leave the response body open.

**`GetProviderInfo`** (`auth.go`) decodes the body but never closes it, and skips it entirely when decoding fails:

```go
resp, err := http.Get(mctCfg.GetBaseMesheryURL() + "/api/providers")
if err != nil {
    return nil, err
}

if err := json.NewDecoder(resp.Body).Decode(&res); err != nil {
    return nil, err
}
```

**`IsMesheryRunning`** (`healthcheck.go`) only reads the status code, so the body is never touched at all.

An unclosed body keeps the connection out of the transport's idle pool. Neither of these runs on a timer the way `keepConnectionAlive` did, so the impact is smaller than #21563 — but they are on paths that run on most `mesheryctl` invocations, and the decode-error path in `GetProviderInfo` leaks unconditionally.

**Fix**

Close both with the `SafeClose` helper already used throughout `pkg/utils` (`platform.go` uses it for exactly this). `IsMesheryRunning` guards on `resp` being non-nil, because a failed request there is expected and deliberately falls through to the platform checks below.

**Tests**

`TestGetProviderInfoClosesTheResponseBody` swaps in a `http.DefaultTransport` wrapper that counts `Close` calls on the response body, then asserts it was closed exactly once.

Worth noting why it is written that way: I first tried asserting connection reuse via `ConnState`, and that test passed even *without* the fix — Go's transport recycles a connection once the body is read to EOF, which `json.Decode` already does. Counting `Close` directly is what actually distinguishes the two.

```
$ go test ./pkg/utils/ -run TestGetProviderInfoClosesTheResponseBody     # before
    auth_test.go:159: expected the response body to be closed once, got 0
--- FAIL: TestGetProviderInfoClosesTheResponseBody (0.00s)

$ go test ./pkg/utils/ -run TestGetProviderInfoClosesTheResponseBody     # after
--- PASS: TestGetProviderInfoClosesTheResponseBody (0.00s)
ok      github.com/meshery/meshery/mesheryctl/pkg/utils  1.334s
```

`IsMesheryRunning` is not covered by a new test — it reads the active context and branches on the configured platform, so exercising it needs a live Docker/Kubernetes environment.

`go test ./pkg/utils/` passes, `go vet` is clean, and `golangci-lint run --config=../.github/.golangci.yml ./pkg/utils/...` reports nothing on the changed files. It does report two pre-existing `staticcheck` deprecation warnings in `table.go`, which this PR does not touch.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when checking service health, including safer handling of failed requests.
  - Ensured provider information requests properly release response resources after completion.

- **Tests**
  - Added coverage verifying response resources are closed correctly while provider data is decoded successfully.


HariomPtdr:
Good catch, applied in 5532325.

I had guarded on `resp != nil`, which is equivalent in the common case but not in the one `net/http` documents: when `CheckRedirect` fails, `http.Get` returns a non-nil response *and* a non-nil error, and that body is already closed. Guarding on `err == nil` avoids the second close.

The nil-response handling and the fallthrough to the platform checks are unchanged.

PR #21574, opened 2026-08-22
Automate Test Case #9 (Homebrew installation)

### LINKED ISSUE #21084: [Test Plan] Test #9: Install mesheryctl via Homebrew
Automate and verify test case **#9** from the [Meshery Test Plan](https://docs.google.com/spreadsheets/d/13Ir4gfaKoAX9r8qYjAFFl_U9ntke4X5ndREY1T7bnVs/edit?gid=838298230#gid=838298230) ("Latest" tab), currently marked **Absent (Help Wanted)**.

- **Test #:** 9
- **Group:** Installation
- **Component under test:** Installer / Homebrew

**Scenario:** Install mesheryctl via Homebrew.

**Action:**
```sh
brew install mesheryctl
```

**Expected outcome:** Installs the latest version of mesheryctl from Homebrew.

---
This test case is not yet automated. This issue tracks implementing it (e.g. as a BATS case under `mesheryctl/tests/e2e/`, or an install-script test) and updating the Test Plan status once it lands. Help wanted.

hiyach28:
### Summary

This PR fixes #21084 

This PR automates the Meshery Test Plan test case 9 (Homebrew installation).
It introduces a focused, standalone BATS test that executes `brew install mesheryctl` and natively validates the installation. To prevent false positives from pre-existing system binaries, the test explicitly queries Homebrew's internal path (`brew --prefix`), runs the newly installed binary, and strictly requires the binary's reported version to match the exact string Homebrew claims to have installed.
utilizes the standard `mesheryctl` BATS helper logic and strictly adheres to the QA Allure traceability contract by prefixing the test name with `[TC-9][tg=Installation]`.

### Testing
- Verified `mesheryctl` BATS convention conformity.
- The test natively guards itself (`if ! command -v brew`) to gracefully skip execution instead of failing when invoked on Windows or local environments where Homebrew is not installed. 

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for installing `mesheryctl` through Homebrew.
  * Verifies the installation location, installed version, and `mesheryctl version` output.
  * Automatically skips when Homebrew is unavailable.

### DIFF
--- a/mesheryctl/tests/e2e/010-installation/01-brew.bats
+++ b/mesheryctl/tests/e2e/010-installation/01-brew.bats
@@ -24,6 +24,5 @@ setup() {
 
     run "$brew_prefix/bin/mesheryctl" version
     assert_success
-    assert_output --partial "Client"
-    assert_output --partial "$brew_version"
+    assert_line --regexp "Client.*$brew_version"
 }
--- a/mesheryctl/tests/e2e/010-installation/01-brew.bats
+++ b/mesheryctl/tests/e2e/010-installation/01-brew.bats
@@ -24,5 +24,7 @@ setup() {
 
     run "$brew_prefix/bin/mesheryctl" version
     assert_success
-    assert_line --regexp "Client.*$brew_version"
+    local client_version
+    client_version=$(echo "$output" | awk '/^Client/ {print $2}')
+    assert_equal "$client_version" "v$brew_version"
 }
--- a/mesheryctl/tests/e2e/010-installation/01-brew.bats
+++ b/mesheryctl/tests/e2e/010-installation/01-brew.bats
@@ -1,0 +1,29 @@
+#!/usr/bin/env bats
+
+setup() {
+    load "$E2E_HELPERS_PATH/bats_libraries"
+    _load_bats_libraries
+}
+
+@test "[TC-9][tg=Installation] Install mesheryctl via Homebrew" {
+    if ! command -v brew &> /dev/null; then
+        skip "brew is not installed on this system"
+    fi
+
+    run brew install mesheryctl
+    assert_success
+
+    run brew --prefix mesheryctl
+    assert_success
+    local brew_prefix="$output"
+
+    run brew list --versions mesheryctl
+    assert_success
+    local brew_version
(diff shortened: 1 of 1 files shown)


mesheryctl/tests/e2e/010-installation/01-brew.bats
```diff
@@ -0,0 +1,29 @@
...
+    load "$E2E_HELPERS_PATH/bats_libraries"
+    _load_bats_libraries
+}
+
+@test "[TC-9][tg=Installation] Install mesheryctl via Homebrew" {
+    if ! command -v brew &> /dev/null; then
+        skip "brew is not installed on this system"
+    fi
+
+    run brew install mesheryctl
```
ishwar170695, line 13: Apart from what coderabbit already flagged,
One thing I'm wondering about: do we need to handle the case where `mesheryctl` is already installed? The E2E suite setup doesn't appear to establish a clean Homebrew state, so in that case this may not actually exercise a fresh installation. Should we account for that here?

mesheryctl/tests/e2e/010-installation/01-brew.bats
```diff
@@ -0,0 +1,29 @@
...
+    assert_success
+
+    run brew --prefix mesheryctl
+    assert_success
+    local brew_prefix="$output"
+
+    run brew list --versions mesheryctl
+    assert_success
+    local brew_version
+    brew_version=$(echo "$output" | awk '{print $2}')
```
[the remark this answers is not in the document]
hiyach28: For this E2E test, we're relying on the clean CI runner and installing mesheryctl as part of the test, so there should only be one installed version in the environment being tested.

mesheryctl/tests/e2e/010-installation/01-brew.bats
hiyach28: i don't think we should uninstall an existing mesheryctl installation as part of the test though, since that could unexpectedly remove a developer's local setup. The test is intended to run against the clean CI environment, where the installation starts from a fresh runner, so i think leaving the setup unchanged would be better

PR #21575, opened 2026-08-23
Refactor clipboard logic and improve tooltip behavior

### LINKED ISSUE #21518: [UI] Copy button state shared across fields
### Current Behavior

On the Remote [Provider Session page](https://playground.meshery.io/user/preferences), clicking the copy button for one field also shows the "Copied!" state on other copy buttons when hovered.
For example, after copying the `PackageUrl`, hovering over the `ProviderUrl` copy button also shows **"Copied!"** for around 3–4 seconds.

The actual copy functionality works correctly, but the UI state is incorrectly applied to multiple copy buttons.

### Desired Behavior

The **"Copied!"** state should only be displayed for the specific copy button that was clicked. Other copy buttons should retain their normal state.

### Screenshots/Logs

https://github.com/user-attachments/assets/e6c4a77b-0af4-4686-897d-47d39ad29f53

### Environment

- Host OS: Mac Linux Windows
- Platform: Docker Kubernetes

---

RaIyAn90:
**Notes for Reviewers**

- This PR fixes #21518

The "Copy" tooltip in the Remote Provider Info tab was not correctly showing the last-clicked provider when multiple copy actions overlapped. This PR addresses the underlying shared-state and race-condition issues so the "Copied!" tooltip always tracks the most recent user action.

**Changes**
- Changed `copied` state from `boolean` to `string | null` so it can track which provider was copied (the original `boolean` comparison against `providerName` was always false, meaning the tooltip never actually said "Copied!").
- Added `timeoutRef` to properly manage the 2-second tooltip timer across renders.
- Added `requestIdRef` sequence counter to prevent out-of-order clipboard promise resolutions from overwriting the current state.
- Added guards inside both the `.then()` callback and the `setTimeout` callback so stale resolutions and stale timers cannot mutate state belonging to a newer copy action.
- Added an unmount cleanup that clears the pending timer and invalidates pending requests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved copy feedback in user preferences.
  * Copy status now correctly follows the selected provider.
  * Copy feedback resets automatically after two seconds, preventing stale “Copied!” messages.
  * Updated tooltips to clearly indicate which provider information was copied.
  * Prevented outdated copy notifications from appearing after switching providers or leaving the preferences view.


ui/components/user-preferences/index.tsx
```diff
@@ -267,15 +267,15 @@ const UserPreference: React.FC<UserPreferenceProps> = (props) => {
...
       navigator.clipboard
         .writeText(text)
         .then(() => {
-          setCopied(true);
+          setCopied(text);
 
           setTimeout(() => {
-            setCopied(false);
+            setCopied(null);
           }, 2000);
```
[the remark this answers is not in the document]
RaIyAn90: Changes applied

dhruveshmishra (commented):
Please address ai suggestions and can u give preview for ur changes

ui/components/user-preferences/index.tsx
```diff
@@ -267,15 +267,15 @@ const UserPreference: React.FC<UserPreferenceProps> = (props) => {
...
   const RemoteProviderInfoTab = () => {
-    const [copied, setCopied] = useState(false);
-    const copyToClipboard = (text) => {
+    const [copied, setCopied] = useState<string | null>(null);
+    const copyToClipboard = (text: string): void => {
       navigator.clipboard
         .writeText(text)
         .then(() => {
-          setCopied(true);
+          setCopied(text);
```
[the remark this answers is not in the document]
kirbx01: The remaining issue is that copied is now keyed by providerName, but CustomTooltip still compares it against provider. I can help address the remaining review comments if you'd like, @RaIyAn90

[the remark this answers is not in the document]
RaIyAn90: Applied. Updated copyToClipboard to accept a key parameter and updated the tooltip check to copied === providerName

ui/components/user-preferences/index.tsx
```diff
@@ -357,9 +361,12 @@ const UserPreference: React.FC<UserPreferenceProps> = (props) => {
                               {provider}
                             </Typography>
 
-                            <CustomTooltip title={copied ? 'Copied!' : 'Copy'} placement="top">
+                            <CustomTooltip 
+                              title={copied === provider ? 'Copied!' : 'Copy'}
```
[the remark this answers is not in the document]
RaIyAn90: Removed trailing whitespaces.

ui/components/user-preferences/index.tsx
```diff
@@ -357,7 +357,7 @@ const UserPreference: React.FC<UserPreferenceProps> = (props) => {
                               {provider}
                             </Typography>
 
-                            <CustomTooltip title={copied ? 'Copied!' : 'Copy'} placement="top">
+                            <CustomTooltip title={copied === provider ? 'Copied!' : 'Copy'} placement="top">
```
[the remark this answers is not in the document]
RaIyAn90: JSX is reformatted

ishwar170695 (commented):
Minor suggestion: since `timeoutRef` persists across renders, consider clearing the pending timeout on component unmount as well. This isn't blocking for this PR.

Also please fix the formatting so that the CI can pass.

ishwar170695:
> ⚠️ Outside diff range comments (1)
> > ui/components/user-preferences/index.tsx (1)> `275-283`: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Ignore stale clipboard completions.**
> > > If two copy actions overlap, a late completion can overwrite `copied` and cancel the active provider’s timer. Track a request sequence and ignore completions from older clicks. Add a test that resolves mocked clipboard promises out of order.
> > > 
> > > 🤖 Prompt for AI Agents

Also this too.

ui/components/user-preferences/index.tsx
```diff
@@ -267,15 +267,19 @@ const UserPreference: React.FC<UserPreferenceProps> = (props) => {
   }
 
   const RemoteProviderInfoTab = () => {
-    const [copied, setCopied] = useState(false);
-    const copyToClipboard = (text) => {
+    const [copied, setCopied] = useState<string | null>(null);
```
banana-three-join, line 270: Copied serves as the conditional state of when the component was copied. It cannot be changed to string or null.

ui/components/user-preferences/index.tsx
```diff
@@ -267,15 +267,19 @@ const UserPreference: React.FC<UserPreferenceProps> = (props) => {
...
+    const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
```
banana-three-join, line 271: From the looks of it, the timeoutRef is acting as a secondary layer to keep the reference of the timeout function to clearout the timeout but the problem is that, as far as I'm concerned, the state across both icons isn't shared and the timeout is behaving accordingly. It's just that the timeout across the components is a bit long so it gives the impression that they're being shared. 

If my assumption is correct, then the fix is only to reduce the timeout timer so users don't get the wrong impression. 

Also both the issue and the PR need to be updated if I'm correct so please look into that

banana-three-join (changes requested):
Some changes are required and could you look into my assumption? If I'm correct then @Sbragul26 you would need to change the description of the issue as I mentioned within the review

ui/components/user-preferences/index.tsx
ishwar170695: I think the state is shared because both icons use the same `copied` state from `RemoteProviderInfoTab`.

Previously, clicking either icon set `copied` to `true`, so both showed `Copied!`. Reducing the timeout would only shorten that behavior.

Using `string | null` identifies which provider was copied, so only that icon shows `Copied!`. I don't think the issue or PR description needs to change.

banana-three-join: You're right, I misread the scope of the `copied` state, ty for correcting me! @ishwar170695 

Could you add a cleanup on the unmount and finx the lint checks? @RaIyAn90 

Ty!

ui/components/user-preferences/index.tsx
RaIyAn90: With the original `useState(false)` it refers to `someProviderName` which is always false, tooltip is not saying "Copied!" for any provider. The type change is required for the copied state to identify which provider just copied.

ui/components/user-preferences/index.tsx
RaIyAn90: The state is shared across all buttons in the same tab render, where reducing the timeout would hide the visible symptom but the underlying shared state overwrite still happens.

RaIyAn90:
@dhruveshmishra, @ishwar170695, @banana-three-join thank you all for your suggestions.

@ishwar170695 I've added a cleanup on the unmount and fixed the formatting and The `copyToClipboard` function now uses a `requestIdRef` counter that increments on every click which tracks a request sequence and ignore completions from older clicks.

ishwar170695 (changes requested):
The cleanup and formatting fixes look good, but the `requestIdRef` handling is still incomplete.

`requestId` is incremented on each copy action, but it isn't checked when `navigator.clipboard.writeText()` resolves. If two clipboard requests overlap and an older request resolves after a newer one, the older completion can still overwrite `copied` and replace the newer timeout.

Could you add a guard at the start of the `.then()` callback?

```tsx
if (requestId !== requestIdRef.current) {
  return;
}
```

The cleanup should also invalidate pending requests on unmount, for example:

```tsx
requestIdRef.current++;
```

inside the cleanup function.

It would also be good to guard the timeout callback with the same request ID, so an older timer cannot clear state belonging to a newer copy action.

RaIyAn90:
@ishwar170695 Thanks for the catch, you're right, the guard was missing.

Added `if (requestId !== requestIdRef.current) return;` at the start of the .then() callback.

Added the same guard inside the `setTimeout` callback so a stale timer can't clear newer state.

The unmount cleanup now increments `requestIdRef.current` to invalidate any pending requests.

ishwar170695 (approved):

banana-three-join (changes requested):
The `.then()` guard is right. The one in the timer isn't — if a second copy rejects or resolves after 2s, the first timer bails on the id check and `copied` never clears, leaving a stuck "Copied!".

Make the reset idempotent on the key instead:

```ts
timeoutRef.current = setTimeout(() => {
  setCopied((prev) => (prev === key ? null : prev));
  timeoutRef.current = null;
}, 2000);
```

Everything else looks good to merge!

RaIyAn90:
@banana-three-join thanks for your suggestion. I've made changes accordingly, updated the timer to use functional `setCopied`. Using key checks means slow or failed requests won't accidentally overwrite the state anymore.

banana-three-join (dismissed):
LGTM! Could you squash it into only one commit? Ty.

alexquincy:

- `ui/components/user-preferences/index.tsx`

**What to do:**

RaIyAn90:
Hi @banana-three-join, I squashed commits into one commit. While squashing I faced merge block and to  resolve that I accidentally dismissed your review. Could you please re-approve this? And sorry about that.

banana-three-join (approved):

banana-three-join:
That's okay! @RaIyAn90 ty for the fix and the effort. I suggest joining the [Slack community](https://slack.meshery.io/) so we can help you situated with any other issues or any other areas that you might look into improving in!

PR #412, opened 2019-10-27
Changes docs page Playgroud to Management

dellamina:
Closes #407

It does also change the url to match the page name so a redirect (HTTP 301) from the old url to the new one would be required to prevent 404's.

PR #21589, opened 2026-08-24
[ci]: improvement to the failure notification for the Model Generator workflow.

banana-three-join:
## Summary

Previously, a failure produced an email with a fixed subject ("GitHub Actions - Workflow Failure") and a body naming only the repository and workflow. Nothing identified which run failed or why, so every failure required opening Actions, locating the run, and reading the logs before triage could start.

Three changes:

- The generate step's console output is captured with `tee`. `mesheryctl` prints the failure reason to the console, but that output was previously discarded when the runner was destroyed — so the actual reason never survived the run.
- A new `Classify failure` step maps the captured output to a short label and extracts the tool's own error message. `root.go` sets `SilenceUsage` but not `SilenceErrors`, so cobra prints `Error: <message>`, which is used verbatim for the body.
- The failure email leads with the label in the subject, the tool's error text in the body, a direct link to the failed run attempt, and the console output as an attachment. It's also marked `continue-on-error` so an SMTP problem can't replace the real failure in the run history.

Subjects now read like `Spreadsheet credential rejected (run #412)` instead of the fixed string, so failures are distinguishable and triageable from the inbox.

The credential was also moved from the command line into `env:` so it is no longer inlined into the script text.

## Notes

Labels cover the paths that actually produce a non-zero exit — `registry generate` only fails hard on setup errors (credential rejected, spreadsheet unreachable, sheet parse failure, invalid arguments), since per-model failures are logged and swallowed upstream. Anything unmatched falls through to `Unrecognized failure` with the raw message still shown in the body, so no failure loses information.

`Send Email on Registry Generate Error` is intentionally untouched — its condition is muted pending the registry error backlog, per the existing inline comment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automated generation failure reporting.
  * Failure notifications now include a clearer reason, diagnostic details, and run information.
  * Notifications continue processing even if email delivery encounters an error.

* **Chores**
  * Improved generation command logging and credential handling for more reliable workflow execution.

### DIFF
--- a/.github/workflows/model-generator.yml
+++ b/.github/workflows/model-generator.yml
@@ -30,9 +30,14 @@ jobs:
           make
 
       - name: Generate Models and Components
+        env:
+          CRED: ${{ secrets.INTEGRATION_SPREADSHEET_CRED }}
         run: |
           cd mesheryctl
-          ./mesheryctl registry generate --spreadsheet-id "1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw" --spreadsheet-cred "${{ secrets.INTEGRATION_SPREADSHEET_CRED }}"
+          ./mesheryctl registry generate \
+            --spreadsheet-id "1DZHnzxYWOlJ69Oguz4LkRVTFM79kC2tuvdwizOJmeMw" \
+            --spreadsheet-cred "$CRED" \
+            2>&1 | tee "$RUNNER_TEMP/generate.log"
 
       - name: Upload artifacts
         if: always()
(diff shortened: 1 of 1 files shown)


ishwar170695 (commented):
Apart from coderabbit's comment,
The detailed notification currently runs for any workflow failure, while the diagnostics are only collected from the generation step. Since the goal here is to make failures actionable from the email itself, should we either scope this detailed notification to generation failures or capture the failing stage as well? 
Otherwise, a later updater or commit failure could be reported using a successful `generate.log`.

leecalcote:
> Apart from coderabbit's comment, The detailed notification currently runs for any workflow failure, while the diagnostics are only collected from the generation step. Since the goal here is to make failures actionable from the email itself, should we either scope this detailed notification to generation failures or capture the failing stage as well? Otherwise, a later updater or commit failure could be reported using a successful `generate.log`.

@ishwar170695, please propose your changes. You can open a PR/branch off of this one.

banana-three-join:
Requested follow-up work to be done over at https://github.com/meshery/meshery/pull/21603

Issue #21477, opened 2026-08-18
[CI] Flaky mesheryctl unit tests: parallel packages share one writable TestConfig.yaml


miacycle:
## Symptom

`mesheryctl Unit tests` fails intermittently on unrelated PRs. Latest instance: [run 32157497911](https://github.com/meshery/meshery/actions/runs/32157497911) on #21447, whose only new commit was a documentation edit.

The failure has a distinctive shape - the package fails with **no `--- FAIL` line for any subtest**:

```
=== RUN   TestListWorkspaces/given_organization_ID_when_workspace_list_then_return_workspaces_under_that_organization
Valid context is not available in meshconfig
FAIL	github.com/meshery/meshery/mesheryctl/internal/cli/root/workspaces	0.295s
```

`Valid context is not available in meshconfig` comes from `CheckIfCurrentContextIsValid` (`mesheryctl/internal/cli/root/config/config.go:100`), which returns it when `mc.CurrentContext == ""`.

## Root cause

`SetupContextEnv` (`mesheryctl/pkg/utils/testing.go:300`) points the **global** viper at a **single shared, writable file** for every test package:

```go
viper.Reset()
viper.SetConfigFile(path + "/../../../../pkg/utils/TestConfig.yaml")
DefaultConfigPath = path + "/../../../../pkg/utils/TestConfig.yaml"
```

`go test ./mesheryctl/...` runs packages **in parallel processes**. Command code that the tests exercise writes that same file through `viper.WriteConfig()` - `internal/cli/root/system/{context,provider,channel}.go` and `internal/cli/root/config/config.go` all do. So one package's test can be rewriting `TestConfig.yaml` while another package's test reads it, and a read that lands mid-write sees no `current-context`.

That matches every observed property:

- **Intermittent.** Depends on inter-process timing, not on the code under test.
- **Unrelated package.** The failing test never touches provider or context config itself.
- **Passes locally.** `go test --short ./mesheryctl/... -race -coverprofile=... -covermode=atomic` (the exact CI command) passes on the same commit; different machine timing, different interleaving.
- **Passed on the previous push of the same branch**, with more substantial changes than the push that failed.

## Why it is newly visible

The `Run mesheryctl coverage` step was recently corrected so a failing `go test` actually fails the job - previously `| tee` swallowed the exit status (see the comment at `.github/workflows/go-testing-ci.yml:139-148`). This flake was presumably occurring before and simply could not turn the job red.

## Suggested fix

[...]

PR #21592, opened 2026-08-24
[mesheryctl] Fix inverted server version check in system check

HariomPtdr:
**Notes for Reviewers**

`runMesheryVersionHealthChecks` handles the server version result in two ways depending on `hc.Options.PrintLogs` — log it, or return it as an error. The error branch has the condition inverted (`mesheryctl/internal/cli/root/system/check.go`):

```go
if hc.Options.PrintLogs { // log if we're supposed to
    if !*isOutdated {
        utils.Log.Infof("✓ Meshery Server is up-to-date (stable-%s)", serverVersion.GetBuild())
    } else {
        utils.Log.Info("!! Meshery Server is not up-to-date")
    }
} else { // else we grab the error
    if !*isOutdated {
        return errors.New("!! Meshery Server is not up-to-date")
    }
}
```

`handlers.CheckLatestVersion` sets `isOutdated` to `true` when the running build differs from the latest release:

```go
isOutdated := false
...
if latestVersion != serverVersion {
    isOutdated = true
```

So `!*isOutdated` means the server *is* up to date — which is how the logging branch directly above reads it. The error branch therefore raises "not up-to-date" in exactly the case where the server is current, and returns `nil` when it is genuinely outdated:

| Server state | `--print-logs` | Without `--print-logs` (before this change) |
|---|---|---|
| Up to date | `✓ up-to-date` | returns error "not up-to-date" |
| Outdated | `!! not up-to-date` | returns `nil` |

The CLI version check about twenty lines further down the same function uses the same `PrintLogs`/error shape and gets it right — it logs on `latest == version` and errors on `latest != version` — which is what the server check was presumably meant to mirror.

**Fix**

Raise the error on `*isOutdated`.

**Testing**

`go build ./...` passes and `golangci-lint run --config=../.github/.golangci.yml ./internal/cli/root/system/...` reports 0 issues.

I have not added a unit test. `runMesheryVersionHealthChecks` calls `handlers.CheckLatestVersion` and `meshkitutils.GetLatestReleaseTagsSorted`, both of which reach the GitHub releases API with no injection point, so a test around this branch would depend on the network — and #21477 suggests flaky `mesheryctl` tests are already a concern. Happy to add one if you would like the version lookup made injectable first, or if there is an existing pattern for stubbing it that I have missed.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the server version check so it reports an error when the server is outdated, while allowing up-to-date servers to pass successfully.

### DIFF
--- a/docs/assets/scss/_elements.scss
+++ b/docs/assets/scss/_elements.scss
@@ -8,6 +8,7 @@ a > * {
 }
 img {
   max-width: 100%;
+  height: auto;
   border-style: none;
 }
 img.center {
@@ -19,6 +20,14 @@ video.center {
   display: block;
   margin-left: auto;
   margin-right: auto;
+}
+iframe {
+  max-width: 100%;
+}
+iframe:not(.iframe-container iframe) {
+  max-width: 100%;
+  aspect-ratio: 16 / 9;
+  height: auto;
 }
 
 p {
--- a/docs/content/en/concepts/logical/policies/index.md
+++ b/docs/content/en/concepts/logical/policies/index.md
@@ -63,7 +63,9 @@ It may happen that certain eval decisions contain results such that two differen
 
 For an in-depth review, watch this meeting recording.
 
-<iframe width="560" height="315" src="https://www.youtube.com/embed/XrLpBVsm6nk?si=j-igwtd_gQ0N61vV&amp;start=453" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen></iframe>
+<div class="iframe-container">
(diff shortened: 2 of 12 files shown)


ishwar170695:
Please attach the issue link if its solving any of the open issues.

ishwar170695 (approved):
LGTM. Verified the `isOutdated` semantics and the `PrintLogs=false` call path. The previous condition was inverted, and this change correctly makes outdated servers return an error while allowing up-to-date servers to proceed.

HariomPtdr:
Thanks for the review.

There's no existing issue for this one — I found it while reading `check.go`, so there's nothing to link. Happy to file one retroactively if you'd prefer the tracker to have a record.

One thing worth flagging honestly, since you looked at the call path: I don't think the branch is reachable in current `mesheryctl`. `RunVersionChecks` is only set in `systemCheckCmd`'s `RunE`, which is also the only place that builds a `HealthCheckOptions` with `PrintLogs: true`; the `PrintLogs: false` construction in `system start` goes through `RunPreflightHealthChecks()`, which only runs the docker and kubernetes checks and never calls `runMesheryVersionHealthChecks`. So this is latent rather than something users are hitting today.

I still think it's worth fixing — the branch exists to be used, and the CLI version check below it shows the intended shape — but I didn't want to overstate the impact.

banana-three-join (approved):
LGTM. Ty!

PR #21606, opened 2026-08-25
[CI] Add raw error and run link to model-generator failure email

arjunmehta-git:
## Summary

The `model-generator.yml` "Send Email on Workflow Failure" step only ever said the run had failed - no error text, and its "link" was just the plain workflow name (`${{ github.workflow }}`), not a URL. That made the overnight failure email nearly useless for triage.

- **New `Capture Failure Log` step** (`if: failure()`): a single `run:` block that calls `gh run view "$RUN_ID" --log-failed`, the gh CLI's built-in flag for printing only the failed step(s)' raw log output. Writes that (tail-capped at 300 lines as a safety net) plus the run URL into a markdown file.
- **Send Email step**: reads that file via `body: file://...` with `convert_markdown: true`, so the run link renders clickable and the raw output renders as a code block.

Went through an Octokit-based version first (list jobs, download the full job log, slice by timestamp window, compute a step-relative line anchor) - that was overkill for what `gh run view --log-failed` already does in one call, so it's gone in favor of this much smaller diff.

Left the second email ("Send Email on Registry Generate Error") alone - it already links to the run and attaches the registry error log, which was working correctly.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/model-generator.yml'))"` - YAML parses.
- [x] `actionlint .github/workflows/model-generator.yml` - no new findings (pre-existing shellcheck SC2086 note on an untouched step only).
- [x] `gh run view --help` - confirmed `--log-failed` is a real, supported flag.
- [ ] Next real failure of this workflow (or a manual `workflow_dispatch` forced to fail) - confirm the email arrives with the raw log excerpt and a working run link.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow failure notifications with detailed logs from the failed run.
  * Email alerts now include the repository name and run number for easier identification.
  * Failure details are formatted in markdown for improved readability.

### DIFF
--- a/.github/workflows/model-generator.yml
+++ b/.github/workflows/model-generator.yml
@@ -67,95 +67,21 @@ jobs:
           commit_message: "Models generated and updated"
           branch: master
 
-      - name: Gather Failure Details
-        id: gather-failure-details
+      - name: Capture Failure Log
         if: failure()
-        uses: actions/github-script@v8
-        with:
-          github-token: ${{ secrets.GH_ACCESS_TOKEN }}
-          script: |
-            const fs = require('fs');
-            const outFile = '${{ runner.temp }}/failure-details.md';
-            const runUrl = `https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
-
-            let jobUrl = runUrl;
-            let failedStepName = null;
-            let logExcerpt = null;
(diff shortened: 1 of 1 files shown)


simihablo (approved):
lgtm

PR #21554, opened 2026-08-21
[server] Fix malformed doc comments in models/remote_provider.go

wahidrizka:
## What this fixes

Same issue as #21549, but in `server/models` this time.
`.github/.golangci.yml` has `ST1020`, `ST1021`, and `ST1022` marked
"temporarily disabled" under staticcheck, the checks that require an
exported symbol's doc comment to start with the symbol's own name.

Running staticcheck with just those three checks against
`server/models/remote_provider.go` turns up 9 violations. This PR
fixes all of them, comment text only, no behavior changes.

## Six of the nine are stale names, not just style

| Function | Comment used to say | Note |
|---|---|---|
| `DownloadProviderExtensionPackage` | lowercase `downloadProviderExtensionPackage` | old unexported name |
| `PublishCatalogPattern` | `PublishMesheryPattern` | |
| `UnPublishCatalogPattern` | `UnPublishMesheryPattern` | |
| `PublishCatalogFilter` | `CloneMesheryFilter` | |
| `UnPublishCatalogFilter` | `UnPublishMesheryFilter publishes...` | described as publishing, not unpublishing |
| `SaveUserCredential` | `SaveCredential` | also fixed a "creadential" typo |
| `GetUserCredentials` | `GetCredentials` | |

`UnPublishCatalogFilter` is the one worth double-checking: its old
comment said it "publishes" a filter to catalog. Confirmed against the
function body that it's genuinely the unpublish path, it sends an
HTTP DELETE (its sibling `PublishCatalogFilter` sends POST), and its
own log line reads `"attempting to unpubish filter"`.

The other two, `GetUsersKeys` and `HandleUnAuthenticated`, just needed
the name added. `HandleUnAuthenticated`'s comment was literally just
the bare name on its own line with nothing after it, merged into a
proper first sentence instead.

## Verification

```
$ staticcheck -checks="ST1020,ST1021,ST1022" ./server/models/...
# before: 9 hits in remote_provider.go
# after:  0 hits in remote_provider.go (grep confirmed, other files
#         in this package still have their own separate violations,
#         out of scope for this PR)
```

Also clean: `go build ./server/...`, `go vet ./server/models/...`.

## Summary by CodeRabbit

* **Documentation**
  * Updated internal API comments to accurately describe provider downloads, authentication, catalog filtering, and credential handling.
  * No user-facing functionality or behavior changed.

PR #21613, opened 2026-08-26
[server] Fix malformed doc comments in server/machines (batch 6)

wahidrizka:
Continues the doc-comment cleanup from #21549, #21554, #21578, #21579, #21580 and #21588, this time in `server/machines` (the connection state machine and its kubernetes actions).

## What changed

`staticcheck -checks="ST1020,ST1021,ST1022" ./server/machines/...` reported 8 hits across 5 files. All are comment-text-only changes, no behavior changes.

- `EventType`, `Events` (models.go), `StateType`, `States` (state.go): each comment described the type but never named it, so it read like a random sentence rather than documentation. Reworded to start with the type name.
- `SendEvent` (models.go): the existing comment was accurate and useful (explains when to use which context values) but never named the method. Reworded the first line to start with `SendEvent`, and fixed a `useID` typo in the same sentence while I was already rewording it.
- `ExecuteOnEntry` on `ConnectAction`, `DiscoverAction` and `RegisterAction` (server/machines/kubernetes/): all three carried the exact same leftover note, copy-pasted across the three files:

  ```go
  // Execute On Entry and Exit should not return next eventtype i suppose, look again.
  ```

  That note is a design question about the `Action` interface's return signature, not a description of what any specific `ExecuteOnEntry` does, so it never satisfied the lint rule. Checked each implementation's body: all three unconditionally `return machines.NoOp, nil, nil`, and their real logic lives entirely in `Execute`. Added a first line that says exactly that, and kept the original note as a second line rather than deleting it, since it may still be a question worth someone answering.

## Verification

- `staticcheck -checks="ST1020,ST1021,ST1022" ./server/machines/...`: 8 hits before, 0 after.
- `go build ./server/machines/...` and `go build ./server/...`: clean.
- `go vet ./server/machines/...` and `go vet ./server/...`: clean.
- `go test ./server/machines/...`: all passing (`machines`, `machines/helpers`, `machines/kubernetes`).

## Summary by CodeRabbit

* **Documentation**
  * Clarified how Kubernetes connection, discovery, and registration actions are executed.
  * Improved documentation for machine events, event publication, and identifier selection.
  * Clarified the roles of machine state types and state collections.

### DIFF
--- a/server/machines/kubernetes/connect.go
+++ b/server/machines/kubernetes/connect.go
@@ -17,6 +17,7 @@ import (
 
 type ConnectAction struct{}
 
+// ExecuteOnEntry is a no-op; ConnectAction performs all of its work in Execute.
 // Execute On Entry and Exit should not return next eventtype i suppose, look again.
 func (ca *ConnectAction) ExecuteOnEntry(ctx context.Context, machineCtx interface{}, data interface{}) (machines.EventType, *events.Event, error) {
 	return machines.NoOp, nil, nil
--- a/server/machines/kubernetes/discover.go
+++ b/server/machines/kubernetes/discover.go
@@ -13,6 +13,7 @@ import (
 
 type DiscoverAction struct{}
 
+// ExecuteOnEntry is a no-op; DiscoverAction performs all of its work in Execute.
 // Execute On Entry and Exit should not return next eventtype i suppose, look again.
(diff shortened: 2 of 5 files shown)


wahidrizka:
Closing this in favor of #21626, which consolidates this PR together with the other open doc-comment batches into one, per the feedback on #21615. Thanks for the review time already spent here, and sorry for the noise of splitting this into so many PRs.

PR #21615, opened 2026-08-26
[server] Fix malformed doc comments in server/models (batch 8)

wahidrizka:
Continues the doc-comment cleanup from #21549, #21554, #21578, #21579, #21580, #21588, #21613 and #21614, back in `server/models` this time (batch 2 and batch 3 covered part of this package already; this covers 7 of the files that were left).

## What changed

`staticcheck -checks="ST1020,ST1021,ST1022" ./server/models/` (the direct package, not its subpackages) reported 7 hits, one per file. All are comment-text-only changes, no behavior changes.

Two are wrong-name bugs rather than just missing names:

- `MesheryCloneFilterRequestBody` (meshery_filter.go) carried this comment:

  ```go
  // MesheryCatalogFilterRequestBody refers to the type of request body
  // that CloneMesheryFilterHandler would receive
  type MesheryCloneFilterRequestBody struct {
  ```

  `MesheryCatalogFilterRequestBody` is a real, different type declared a few lines above in the same file, with its own correct comment. The sentence itself was accurate (`CloneMesheryFilterHandler` really does receive a `*models.MesheryCloneFilterRequestBody`, confirmed in `server/handlers/meshery_filter_handler.go`), so only the leading name needed to change.

- `SessionPreferencePersister` (preference_persister.go) was documented as `PreferencePersister`, the name of the interface it implements (`PreferencePersister`, defined separately in `preference.go`), not its own name.

The rest (`RegistrationFailureLog`, `PreferenceParams`, `MeshSyncRegistrationQueue`, `GetIndexForRegisterCol`, `GetConnection`) just never named the symbol they were documenting; their descriptions were otherwise accurate and are kept as-is.

## Verification

- `staticcheck -checks="ST1020,ST1021,ST1022" ./server/models/`: 7 hits before, 0 after (9 remain in subpackages under `server/models/...`, unrelated to this PR's files).
- `go build ./server/models/...` and `go build ./server/...`: clean.
- `go vet ./server/models/...`: clean.
- `go test ./server/models/`: 20 pre-existing failures on this machine, all with the identical root cause `Binary was compiled with 'CGO_ENABLED=0', go-sqlite3 requires cgo to work`, none touching the files or symbols this PR changes. Confirmed unrelated by stashing this PR's changes and re-running the same tests against unmodified master: identical failures.

## Summary by CodeRabbit

* **Documentation**
  * Clarified and corrected internal developer documentation for connection lookups, registration queues, preferences, session persistence, key registration, filter requests, and failure logs.
  * No user-facing functionality or behavior changed.

### DIFF
--- a/server/models/connection_persister.go
+++ b/server/models/connection_persister.go
@@ -264,8 +264,7 @@ func (cp *ConnectionPersister) UpdateConnectionByID(connection *connections.Conn
 	return connection, nil
 }
 
-// Get connection by ID
-// If kind is provided filter with kind too
+// GetConnection gets a connection by ID. If kind is provided, it also filters by kind.
 func (cp *ConnectionPersister) GetConnection(id core.Uuid, kind string) (*connections.Connection, error) {
 	connection := connections.Connection{}
 	query := cp.DB.Where("id = ?", id)
--- a/server/models/keys_helper.go
+++ b/server/models/keys_helper.go
@@ -35,7 +35,7 @@ func NewKeysRegistrationHelper(dbHandler *database.Handler, log logger.Handler) 
 	return krh, err
 }
 
-// returns the spreadsheet column index that captures whether the key should be registered.
+// GetIndexForRegisterCol returns the spreadsheet column index that captures whether the key should be registered.
 func (kh *KeysRegistrationHelper) GetIndexForRegisterCol(cols []string) int {
 	if shouldRegisterColIndex != -1 {
 		return shouldRegisterColIndex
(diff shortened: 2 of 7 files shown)


banana-three-join:
While I do appreciate the help. Could you please compress all of these PRs into one through cherry picking? A change like this shouldn't have grown into 9 batches and it's starting to add up as noise to the fix.

wahidrizka:
Closing this in favor of #21626, which consolidates this PR together with the other open doc-comment batches into one, per the feedback on #21615. Thanks for the review time already spent here, and sorry for the noise of splitting this into so many PRs.

PR #21635, opened 2026-08-27
[UI] Fix Workspace Edit/Delete stale closure bug

### LINKED ISSUE #21629: [UI] TypeError on workspace Edit/Delete due to stale closure in WorkspaceDataTable
### Current Behavior
Clicking the "Edit Workspace" or "Delete Workspace" action buttons from the three-dot menu in the Workspace data table view immediately crashes the UI with a runtime `TypeError: Cannot read properties of undefined (reading 'name')`.

**Technical Details:**
The crash is caused by a stale closure in `ui/components/workspaces/WorkspaceDataTable.tsx`. The `columns` definition (specifically the `customBodyRender` for the Actions column) captures the `workspacesData` array in its lexical scope. Because `tableCols` is initialized using `useState(columns)` on the first render: when the RTK Query has not yet returned data and the array is empty (`[]`): the closure is permanently frozen over the empty array. 

When a user clicks the action buttons after the table data loads, the handler evaluates `workspacesData[tableMeta.rowIndex]` against the empty array, which results in `undefined`. This `undefined` value is then passed into `handleWorkspaceModalOpen` or `handleDeleteWorkspaceConfirm`, causing the crash when the code attempts to read the `.name` property.

### Expected Behavior
- Clicking "Edit Workspace" should successfully open the edit modal pre-populated with the selected workspace's details. 
- Clicking "Delete Workspace" should successfully open the confirmation modal displaying the selected workspace's name.
- The Actions column should read the workspace object from a mutable ref or directly from the updated row data rather than relying on a stale `useState(columns)` closure.

### Screenshots/Logs

Image

Image

### Environment

---

hiyach28:
**Notes for Reviewers**

- This PR fixes #21629 

### Description
Clicking the "Edit" or "Delete" actions on a Workspace in the UI resulted in an `Uncaught TypeError: Cannot read properties of undefined (reading 'name')` crash. This occurred because `WorkspaceDataTable` retained a stale closure over the initial empty `workspacesData` array (due to the `tableCols` state initialization required by `ResponsiveDataTable`), causing the `customBodyRender` callback in the Actions column to look up data against `[]`. 

### Changes
- Imported and utilized `useRef` to maintain a stable, mutable reference to the latest `workspacesData` array returned by the RTK query across renders.
- Updated the Actions column's `customBodyRender` to pull `selectedWorkspace` from `workspacesDataRef.current[tableMeta.rowIndex]` instead of the stale closure.
- Left the existing `tableCols` state entirely unmodified to ensure Sistent's `ResponsiveDataTable` column-visibility lifecycle is perfectly preserved.

### Screen Recording
Showing the fixed working of edit and delete buttons 

https://github.com/user-attachments/assets/0fd26799-36f9-49c9-b69d-766404775f0f 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace table actions to consistently use the most up-to-date workspace data.
  * Prevented actions from relying on outdated workspace lists after table updates.

### DIFF
--- a/.github/workflows/rego-lint-and-test.yml
+++ b/.github/workflows/rego-lint-and-test.yml
@@ -7,6 +7,7 @@ on:
     paths:
       - "**.rego"
       - "server/policies/**"
+      - "models/**"
       - ".github/workflows/rego-lint-and-test.yml"
   push:
     branches:
@@ -14,6 +15,7 @@ on:
     paths:
       - "**.rego"
       - "server/policies/**"
+      - "models/**"
   workflow_dispatch:
   workflow_call:
 
--- a/VISION.answers.md
+++ b/VISION.answers.md
@@ -1,0 +1,164 @@
+# Meshery vision review answers
+
+This file preserves the author's recorded decisions that calibrate `VISION.md`.
+
+## Round 1 - 2026-08-26
+
+### H-1 - A generic resource importer without a model
+
+Verdict: In vision.
+
+Reasoning: Users can use annotation-only components and annotation-only relationships (like those offered in the meshery-core and meshery-shapes models) to visually depict their full architecture or to visually and collaboratively convey a concept with the acknowledgement that those annotation-only components and relationships are not semantically meaningful; they do not bear configuration properties that can be orchestrated or deployed.
+
+Reasoning: That annotation-only components are excluded from deployment operations.
+
+### H-2 - A marketplace extension with a trust banner
+
+Verdict: In vision.
+
+### H-3 - A temporary local API endpoint
+
+Verdict: Off mission.
+
+Reasoning: We need to ratchet away from the local API endpoints.
+
+Reasoning: Meshery is a schema-driven project.
+
+Reasoning: The source of truth for all API endpoints shall be sourced from https://github.com/meshery/schemas.
+
+### H-4 - Canonicalize credentials on the next save
+
+Verdict: In vision.
+
+### H-5 - Keep a permission-denied design entry visible
+
+Verdict: Off mission.
+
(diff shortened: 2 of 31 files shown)


rishiraj38 (commented):
The core fix looks correct. `useState(columns)` captures the initial `workspacesData`, causing the action column to use stale data. Using a ref to access the latest data resolves that issue without affecting the table's column state.

A few things still need attention:

1. Add a guard in `customBodyRender` before passing the workspace to `WorkspaceActionList`. Using `workspacesDataRef.current?.[tableMeta.rowIndex]` is also slightly safer than directly indexing into `.current`.
2. Add defensive guards in `handleDeleteWorkspaceConfirm` and `handleWorkspaceModalOpen` since both handlers assume the workspace object exists.
3. The issue appears to be in how the component consumes the RTK query data, not in the schemas generated RTK client itself.

Overall, the approach is correct. Adding these guards would make the fix more robust when the workspace data is temporarily unavailable.

hiyach28:
@rishiraj38 I checked the guards against similar action patterns like Filters.columns.tsx and MesheryPatterns.columns.tsx directly resolve data `tableMeta.rowIndex` and pass the domain object to their action handlers without defensive guards. Since `ResponsiveDataTable `is rendering from the same server side data array and the ref now tracks that live RTK response, if we see there isn't an expected path for an invalid workspace here. so i think adding those if conditions won't address a reachable state, hence i think we should leave it as it is. 
Please let me know what you think

rishiraj38 (approved):
LGTM !!!!!

banana-three-join (approved):
LGTM. The reference is frozen in place, so the way around it is to use the current ref to get the updated value of the frozen reference

PR #21646, opened 2026-08-28
fix(server): add nil check for machineInst to prevent potential crash

### LINKED ISSUE #20821: [Server] Unguarded SendEvent on possibly-nil instance in auto_register (dormant latent crash)

Retr0-XD:
## Description
Fixes a potential nil pointer dereference crash in auto registration flow.

## Issue
Fixes #20821

## Problem
In auto_register.go, if InitializeMachineWithContext fails, machineInst can be nil. However, the code continues and attempts to call SendEvent() on the nil pointer without checking, causing a panic.

## Solution
- Added nil check after machine initialization
- Added continue statement to skip to next URL on failure
- Prevents nil pointer dereference while maintaining intended error handling behavior

## Changes
- Line 108: Updated condition to check both err != nil AND machineInst == nil
- Line 109: Added continue to skip to next iteration if initialization failed

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic registration reliability by handling unavailable machine instances as failures.
  - Failed registration attempts are now logged, while processing continues with other eligible machines.

### DIFF
--- a/server/machines/helpers/auto_register.go
+++ b/server/machines/helpers/auto_register.go
@@ -105,8 +105,9 @@ func (arh *AutoRegistrationHelper) processRegistration() {
 
 						machineInst, err := InitializeMachineWithContext(connectionPayload, ctx, id, data.MeshsyncDataHandler.UserID, arh.smInstanceTracker, arh.log, data.MeshsyncDataHandler.Provider, machines.DISCOVERED, connType, nil)
 
-						if err != nil {
+						if err != nil || machineInst == nil {
 							arh.log.Error(ErrAutoRegister(err, connType))
+							continue
 						}
 
 						_, err = machineInst.SendEvent(ctx, machines.Register, connectionPayload)


banana-three-join:
For next time, please don't raise a PR when it has been assigned to someone already. Request for the PR to be assigned to you if the currently assigned contributor has gone silent and the PR becomes stale. I'll go ahead and close this PR. 

On another note, by adding an `or` to the `err != nil` with `machineInst == nil` then let me ask you something, if in that sequence the `err` gets logged, what would happen if `err == nil`? So, wrong approach and somethnig to keep in mind. 

If you would like to work on another issue, please look for one which hasn't been assigned to someone yet or look for ones that have stale PRs and I'll gladly assign you, but for the time being, I'll go ahead and close this PR.

Issue #21205, opened 2026-08-07
Workspace edit and delete actions throw runtime error when clicked


rishiraj38:
### Description

Clicking the **Edit** or **Delete** action for a [workspace](https://playground.meshery.io/management/workspaces) throws a runtime error, preventing the action from completing.

### Expected Behavior

* **Edit** should open the workspace edit modal with the selected workspace details.
* **Delete** should open the confirmation dialog for the selected workspace.

### Actual Behavior

A runtime error is thrown:

```text
TypeError: Cannot read properties of undefined (reading 'name')
```

### Error Details

**Edit Action**

<img width="1207" height="807" alt="Image" src="https://github.com/user-attachments/assets/0e3da7fe-1b0f-471e-82e9-83883f1b5dc4" />

**Delete Action**

<img width="1405" height="765" alt="Image" src="https://github.com/user-attachments/assets/49d47ebc-db10-4dea-9c55-61c1c311c626" />

### Environment

- **Host OS:** Mac Linux Windows
- **Platform:** Docker or Kubernetes
- **Meshery Server Version:** stable-v
- **Meshery Client Version:** stable-v

---
### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)
- 🛠 [Meshery Build & Release Strategy](https://docs.meshery.io/project/contributing/build-and-release)
- 📚 [Instructions for contributing to documentation](https://docs.meshery.io/project/contributing/contributing-docs)
   - Meshery documentation [site](https://docs.meshery.io/) and [source](https://github.com/meshery/meshery/tree/master/docs)
- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://meshery.io/community#community-forums) and [Community Slack](https://slack.meshery.io)

ayanasarkar:
can i work on this issue?

hiyach28:
@rishiraj38 I would love to be assigned on this issue. Will investigate it, find the root cause, and subsequently raise a PR after discussing if it goes in the right approach!

hiyach28:
[...]

PR #21649, opened 2026-08-28
fix: Add missing workspaceID validation in workspace handlers (#21205)

Retr0-XD:
## Summary

Add validation to ensure workspaceID is not empty in both DeleteWorkspaceHandler and UpdateWorkspaceHandler before passing to provider methods.

## Problem

Missing validation causes runtime errors when the route parameter 'id' is missing or empty. This results in empty workspace IDs being passed to provider methods, causing unexpected behavior.

## Solution

Added validation in both handlers to return a proper 400 Bad Request error with ErrWorkspaceMissingInput when workspaceID is empty, matching the pattern used in GetWorkspacesHandler.

## Testing

The fix prevents empty workspaceID values from reaching provider methods and returns proper error responses to the client.

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded workspace ID validation across workspace, environment, design, view, and team operations.
  * Requests missing a workspace ID now receive a clear HTTP 400 Bad Request response before processing.
  * Workspace update and deletion requests validate missing IDs consistently.
  * Invalid requests no longer trigger unnecessary provider operations, improving reliability and preventing unintended processing.

### DIFF
--- a/server/handlers/workspace_handlers.go
+++ b/server/handlers/workspace_handlers.go
@@ -13,6 +13,17 @@ import (
 	workspace "github.com/meshery/schemas/models/v1beta3/workspace"
 )
 
+// validateWorkspaceID validates that workspaceID is not empty
+func (h *Handler) validateWorkspaceID(w http.ResponseWriter, workspaceID string) bool {
+	if workspaceID == "" {
+		missingInput := models.ErrWorkspaceMissingInput()
+		h.log.Error(missingInput)
+		writeMeshkitError(w, missingInput, http.StatusBadRequest)
+		return false
+	}
+	return true
+}
+
 // workspacePayloadWire is a handler-local dual-accept wrapper around
 // workspace.WorkspacePayload (now v1beta3, canonical-camelCase). The
 // canonical wire form emits `organizationId`; the legacy `organization_id`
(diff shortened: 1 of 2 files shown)


Junnygram (commented):
A few gaps if the intent of #21205 is that every single-workspace handler validates workspaceId:

- `GetWorkspaceByIdHandler` still reads `mux.Vars(r)["id"]` with no empty check, even though `ErrWorkspaceMissingInput` says workspaceId is required for single-workspace operations.
- The nested handlers (environments, designs, views, teams) have the same unguarded `workspaceID := mux.Vars(req)["id"]`.
- Gorilla mux will not match `/api/workspaces/{id}` with an empty `id` in normal routing, so this is mostly defensive. If we are adding it, it should be consistent — or extracted to one helper — rather than only Delete/Update.
- No test covers the 400 path.

Not blocking if #21205 was scoped to just those two handlers; otherwise please extend the check (and a small handler test) before merge.

Retr0-XD:
@Junnygram addressed those gaps, 
All handlers validate workspaceID consistently now + test coverage for the 400 error paths, and everything's passing

yi-nuo426:
server/handlers/workspace_handlers.go:16-24 consistently returns a 400 before provider calls when workspaceID is empty, and server/handlers/workspace_handlers_test.go:461-463 correctly fails if the provider is invoked on that invalid path. No changes requested.

Retr0-XD:
@Junnygram the gaps you called out have been addressed: workspaceID validation is now consistent across handlers and the test verifies the 400 path is triggered before the provider is invoked. @yi-nuo426 confirmed the changes on Sep 14. Could you take another look when you have a moment?

PR #21665, opened 2026-08-28
[CI] Migrate integration test mesh configuration to Registry identifier

Salmaan-M:
**Notes for Reviewers**

### Description

The scheduled mesheryctl-e2e workflow currently uses the legacy SMP representation for the mesh in .github/integration-test-config.yaml:

mesh:
  type: 3

The current PerformanceTestConfigFile expects mesh to be a string representing a Meshery Registry model name. Since type: 3 corresponds to Istio and the current Registry identifier is istio, this updates the configuration to:

mesh: istio

This resolves the configuration type mismatch encountered during mesheryctl perf apply.

Testing

Validated the configuration against the current PerformanceTestConfigFile implementation. The E2E workflow will be used to verify the change in CI.

## Summary by CodeRabbit

* **Chores**
  * Updated integration test configuration to identify Istio as the service mesh using the current configuration format.

### DIFF
--- a/.github/integration-test-config.yaml
+++ b/.github/integration-test-config.yaml
@@ -24,5 +24,4 @@ test:
 
 # Service mesh under test in Service Mesh Performance Spec format
 # See: https://github.com/service-mesh-performance/service-mesh-performance/blob/master/protos/service_mesh.proto
-mesh:
-  type: 3
+mesh: istio


yi-nuo426:
Nice migration to the Registry identifier. Please update or remove the two preceding comments in .github/integration-test-config.yaml:25-27: they still say mesh uses the SMP protobuf representation and link to the SMP enum, while the value is now the Registry model name istio. Leaving them unchanged makes the canonical E2E input self-contradictory. Replace them with wording such as Meshery Registry model name under test.

PR #21670, opened 2026-08-29
[Docs] Generate llms.txt and llms-full.txt for the documentation site

Gauravtiwari31:
**Notes for Reviewers**

- This PR addresses meshery/meshery.io#2921 (the issue is filed on the website repo, but the docs Hugo config lives here).

Publishes `/llms.txt` and `/llms-full.txt` per the [llmstxt.org](https://llmstxt.org) spec, plus a raw-Markdown rendition of every page at `<page-url>index.md`, so AI agents can consume the docs without scraping rendered HTML.

### What changed

- Registers three output formats — `llms`, `llmsfull`, `md` — and enables them for the `home`, `page` and `section` kinds. Taxonomy and term output stay disabled, as they are today.
- Adds a `[params.llms]` block for the title, exclusions and the "Optional" groupings.
- Adds the generation templates under `layouts/` and `layouts/partials/llms-txt/`.
- Raises the Hugo floor from 0.110.0 to 0.117.0 (see below). CI already builds with 0.157.0 and `package.json` pins 0.158.0, so nothing in the pipeline moves.

### Why the templates are vendored rather than the community module imported

The issue suggested importing [`gethugothemes/hugo-modules/llms-txt`](https://github.com/gethugothemes/hugo-modules/tree/master/llms-txt). I evaluated it and it is not usable here as published, for one decisive reason: it renders page bodies with `.RawContent`, which does not expand shortcodes.

This site leans on shortcodes exactly where it matters — **35 of 37** installation pages, **43 of 53** guides, **116 of 119** reference pages and **20 of 22** concepts contain them, and the `ref` shortcode alone appears **1,113 times**. Every internal cross-link would have reached an agent as literal template syntax instead of a URL. Worse, pages whose body is a single shortcode (`installation/accessing-meshery-ui.md`, for one) would have produced no content at all.

These templates use `.RenderShortcodes` instead, which is what moves the Hugo floor to 0.117.0. Four of the module's five templates would have needed overriding anyway, which would have left only its output-format declarations actually in use while masking any upstream fixes — hence vendoring. The templates are derived from that module and retain its MIT notice.

### Why some pages are grouped under "Optional"

`project/compatibility`, `project/releases` and `extensions/models` are **2,508 of the site's 2,811 pages and 69% of its Markdown by volume**, and nearly all of them have no body at all — they are front matter rendered by a custom layout. Enumerating them would bury the hand-written docs while giving an agent nothing to read, so each is linked once from the spec's conventional `## Optional` section. Per-page `index.md` output is still generated for them, so nothing becomes unreachable. These three groupings are configuration, not hard-coded, and are the part of this PR I'd most expect review feedback on.

### Verification

Verified against a full production build (5,684 pages, clean build):

| | |
|---|---|
| `llms.txt` | 71 KB, 303 entries across 8 sections, **0 unresolved links** |
| `llms-full.txt` | 3.32 MB, 296 stanzas |
| per-page `index.md` | 2,810 files at a predictable path |
| unrendered shortcode syntax | **0** (57 occurrences, all from the two pages that *document* shortcodes and escape it deliberately) |

A structural validator checks `llms.txt` against the spec — H1, blockquote summary, H2 file lists, `- [name](url): notes` item grammar, absolute URLs, `Optional` last — plus that every listed URL was actually published. It is negative-tested (a corrupted H1, a malformed item and a dead link are all rejected), so the pass is meaningful.

### Notes and known limitations

- `llms-full.txt` at 3.32 MB (~870K tokens) exceeds any current model's context even after filtering. It suits ingestion pipelines rather than single-request reading; happy to drop it or trim it further if maintainers prefer.
- 224 of 2,810 `index.md` files contain block HTML, because Docsy/Meshery shortcodes emit HTML — 153 wrap code in `<pre class="codeblock-pre">` rather than fenced blocks. Valid Markdown and readable, but not pristine. Possible follow-up.
- The partials live in `layouts/partials/llms-txt/`, not `layouts/partials/llms/`, because the bare `LLMs` entry at `.gitignore:106` matches a directory named `llms` on case-insensitive filesystems. Linux CI is unaffected, but the files would have silently vanished for contributors on Windows and macOS. Worth tightening that ignore rule separately.
- Not yet done: pointing a live ChatGPT/Claude session at the generated files. That needs a preview deploy; I verified navigation deterministically against the build instead (index entry → `index.md` → does the page actually answer the question).

## Summary by CodeRabbit

* **New Features**
  * Added machine-readable Markdown versions of documentation pages.
  * Added “View as Markdown” links and Markdown discovery metadata.
  * Added `llms.txt` with page links, descriptions, exclusions, and optional bulk collections.
  * Added `llms-full.txt`, containing the complete documentation corpus in Markdown.
  * Added configurable titles, notes, exclusions, and bulk page sets.

* **Improvements**
  * Updated documentation builds to generate these formats.
  * Raised the minimum supported Hugo version to 0.117.0.

### DIFF
--- a/.github/workflows/build-docs-preview.yml
+++ b/.github/workflows/build-docs-preview.yml
@@ -90,7 +90,7 @@ jobs:
           PREVIEW_BASE_URL: "/pr-preview/pr-${{ github.event.pull_request.number }}/"
         run: |
           cd docs
-          npm run build:production -- --gc --baseURL "${PREVIEW_BASE_URL}"
+          npm run build:production:llms -- --gc --baseURL "${PREVIEW_BASE_URL}"
           find public -maxdepth 1 -type d -name 'v0.*' -exec rm -rf {} +
           cat > public/robots.txt <<'ROBOTS'
           User-agent: *
--- a/.github/workflows/hugo.yml
+++ b/.github/workflows/hugo.yml
@@ -92,7 +92,7 @@ jobs:
           HUGO_ENVIRONMENT: production
           HUGO_ENV: production
         run: |
-          make -C docs build-production BASE_URL="/"
+          make -C docs build-production-llms BASE_URL="/"
           cp docs/CNAME docs/public/CNAME
           touch docs/public/.nojekyll
           if [ -d gh-pages-maintenance/pr-preview ]; then
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -33,6 +33,10 @@
 #   build              Build locally with draft, future, and expired content.
 #   build-preview      Build for a deploy preview (uses DEPLOY_PRIME_URL).
 #   build-production   Build for production. Pass BASE_URL=... to set the base URL.
+#   build-llms         Build locally, also emitting llms.txt, llms-full.txt and
+#                      the per-page Markdown renditions.
+#   build-production-llms
+#                      Build for production with those files. Pass BASE_URL=...
 #   site               Serve locally with live reload.
 #   serve              Serve locally once, file watcher off (no live reload).
 #   clean              Empty the build cache, reinstall dependencies, run 'site'.
@@ -97,6 +101,25 @@ build-production: check-go check-deps
 		npm run build:production -- --baseURL "$$base_url"; \
 	else \
 		npm run build:production; \
+	fi
+
+## Build docs.meshery.io locally including the LLM-facing files: /llms.txt,
+## /llms-full.txt, and an index.md rendition beside every page. These are kept
+## out of `build` so the standard contributor flow is not slowed by ~2,810
+## extra files; config-llms.toml layers the output formats on when asked for.
+build-llms: check-go check-deps
+	npm run build:llms
+
+## Build docs.meshery.io for production including the LLM-facing files.
+## Pass BASE_URL=... to set the base URL, exactly as build-production does.
+build-production-llms: check-go check-deps
+	set -e; \
+	if [ -n "$(BASE_URL)" ]; then \
+		base_url="$(BASE_URL)"; \
+		base_url="$${base_url%/}/"; \
+		npm run build:production:llms -- --baseURL "$$base_url"; \
+	else \
+		npm run build:production:llms; \
 	fi
 
 ## Build and run docs.meshery.io on your local machine (draft and future content enabled).
--- a/docs/config-llms.toml
+++ b/docs/config-llms.toml
@@ -1,0 +1,17 @@
+# Enables the LLM-facing outputs. Layered on top of hugo.toml:
+#
+#     hugo --config hugo.toml,config-llms.toml
+#
+# which the `build-llms` and `build-production-llms` make targets do for you.
+# Kept out of hugo.toml so that the standard contributor build (`make build`,
+# `make site`) does not pay for 2,810 extra index.md files, and so that a
+# contributor editing one page sees exactly the output they saw before.
+#
+# Each key here replaces its counterpart in hugo.toml wholesale, so the
+# pre-existing formats are restated rather than appended to.
+[outputs]
+  home = ["HTML", "RSS", "SITEMAP", "llms", "llmsfull", "md"]
+  page = ["HTML", "md"]
+  section = ["HTML", "RSS", "md"]
+  taxonomy = []
+  term = []
--- a/docs/hugo.toml
+++ b/docs/hugo.toml
@@ -101,6 +101,34 @@ As a self-service engineering platform, Meshery enables collaborative design and
   taxonomy = []
   term = []
 
+# The formats below are declared but not switched on here. Declaring an output
+# format generates nothing by itself; config-llms.toml adds the [outputs]
+# entries that enable them, and is layered in by the build-llms make targets.
+# This keeps llms.txt, llms-full.txt and the 2,810 per-page index.md files out
+# of the standard contributor build.
+#
+# "llms" and "llmsfull" render once, for the home page. "md" publishes an
+# index.md beside every page's index.html; those are the URLs llms.txt points
+# at, so an agent that follows one receives Markdown prose rather than a themed
+# HTML page. Taxonomy and term pages stay unpublished, so neither file lists
+# them.
+[outputFormats]
+  [outputFormats.llms]
+    baseName = "llms"
+    mediaType = "text/plain"
+    isPlainText = true
+    notAlternative = true
+  [outputFormats.llmsfull]
+    baseName = "llms-full"
+    mediaType = "text/plain"
+    isPlainText = true
+    notAlternative = true
+  [outputFormats.md]
+    baseName = "index"
+    mediaType = "text/markdown"
+    isPlainText = true
+    notAlternative = true
+
 [sitemap]
   changefreq = "daily"
   priority = 0.5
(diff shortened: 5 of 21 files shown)


Gauravtiwari31:
Preview is live, and the last item from the PR description ("point a live AI agent at the generated files") is now done against it.

**Preview:** https://docs.meshery.io/pr-preview/pr-21670/llms.txt

### One fix worth calling out, in case you looked at an earlier preview

The first preview of this PR published a **broken** `llms.txt` — 2,812 entries at 467 KB, with all 1,288 `project/compatibility` pages and the site search page leaking in.

Cause: the exclusion patterns are written relative to the site root (`/project/compatibility/`), but `.RelPermalink` carries whatever path `baseURL` has. Preview deploys under `/pr-preview/pr-NNNN/`, so nothing matched and every excluded page was included. Production deploys with `BASE_URL="/"` and would have been fine, which is exactly why this did not show up in local builds — but any versioned or subpath deploy would have hit it too.

`is-included.html` now strips the site root before matching, so the patterns mean the same thing wherever the site is served from. There is a regression test that builds with `--baseURL http://example.com/pr-preview/pr-1/` and asserts the entry count matches a root deploy.

Current preview output:

| | before | after |
|---|---|---|
| entries | 2,812 | **303** |
| size | 467 KB | **71 KB** |
| `project/compatibility` links | 1,288 | **1** (the Optional entry) |
| site search page | present | **excluded** |

### Live agent navigation test

Pointed a model at the preview `llms.txt` with no other context, then had it follow whichever link it chose.

| task | page it picked | answered? |
|---|---|---|
| Install on Kubernetes with Helm | `installation/kubernetes/helm/index.md` | yes — returned `helm repo add meshery …` / `helm install meshery …` verbatim |
| Install CLI on Windows | `installation/mesheryctl/windows/index.md` | yes — Scoop bucket commands plus the direct-download path |
| Access the web UI | `installation/accessing-meshery-ui/index.md` | yes — `http://localhost:9081` and `mesheryctl system dashboard` |

3/3 routed correctly, 3/3 answered from the Markdown alone.

The third one is the case that motivated vendoring rather than importing the module: that page's entire body is a single shortcode, so under the module's `.RawContent` rendering it would have contained nothing but `{{< installation/accessing-meshery-ui >}}` and answered nothing.

### Verification summary

Full production build, 5,684 pages, clean:

- `llms.txt` — 303 entries, 8 sections, **0 unresolved links**
- `llms-full.txt` — 3.32 MB, 296 stanzas
- per-page `index.md` — 2,810 files
- unrendered shortcode syntax — **0** (57 occurrences, all from the two pages that document shortcodes and escape it deliberately)

Also addressed the CodeRabbit review comment: Markdown link-label characters in page titles are now escaped. No current page title contains a bracket, so output is unchanged today — it is a latent break, not an active one.

### Still open for maintainer input

- `llms-full.txt` at 3.32 MB (~870K tokens) is past any current model's context even after filtering. It suits ingestion pipelines rather than single-request reading — happy to drop it or trim it further if you would prefer.
- The three `[params.llms] bulk` groupings are the judgement call in this PR. They are configuration, not hard-coded, so they are easy to change.
- Unrelated to this PR: the bare `LLMs` entry at `.gitignore:106` matches any directory named `llms` on case-insensitive filesystems. It silently hid these partials on Windows until I renamed the directory to `llms-txt`. Worth tightening separately.

Gauravtiwari31:
Two follow-ups pushed (`6302d49`).

### `.Date` — made explicit, but it was not producing wrong output

`full-entry.html` now tests `.Date.IsZero` instead of leaning on `with .Date`:

```
{{- if not .Date.IsZero -}}
  {{- $lines = $lines | append (printf "date: %s" (.Date.Format "2006-01-02")) -}}
{{- end -}}
```

Worth being precise about why, because the obvious reading is wrong. In plain Go templates a struct is *always* truthy, so `with .Date` would emit `date: 0001-01-01` on every undated page. Hugo's own truthiness honours the `IsZero` interface, so it was already skipping them correctly — I checked the built output before changing anything: **296 stanzas, 2 date lines, zero occurrences of `0001-01-01`**.

So this is not a bug fix, it is removing an undocumented dependency on Hugo-specific truth rules for a line that would otherwise be wrong on 294 of 296 stanzas. Output is unchanged.

### Bulk traversal — measured, and deliberately left as-is

The section loop rescans the full page list once per top-level section, which reads as O(sections × pages). I profiled it rather than guessing, with `--templateMetrics` over the full 5,684-page build:

```
cumulative     average      total  template
 1.7291427s  615.572µs       2809  _partials/llms-txt/page-markdown.html
934.5643ms  934.5643ms          1  index.llmsfull.txt
646.5488ms   2.155162ms       300  _partials/llms-txt/full-entry.html
329.2836ms  329.2836ms          1  index.md
128.3805ms  128.3805ms          1  index.llms.txt
 80.3769ms    14.312µs       5616  _partials/llms-txt/is-included.html
 16.7969ms    55.435µs        303  _partials/llms-txt/entry.html
  7.9276ms    13.256µs        598  _partials/llms-txt/md-url.html
```

The call count is the answer. `is-included` ran **5,616 times, not ~34,000**. The `hasPrefix` guard sits before the partial call, and a page lives under exactly one top-level section, so the partial fires once per page per template (~2,808 × 2), while the nested loop only costs cheap string comparisons. It totals **80ms — 0.004% of the build**.

Everything llms-related together is ~3.9s of template time. The dominant cost is `page-markdown.html` at 1.73s across 2,809 calls, which is the per-page `index.md` output itself — inherent to the feature, not the traversal.

So: **justifying, not fixing.** Bucketing pages by section in one pass would save a few thousand string comparisons and cost a map build plus a second place to get the baseURL-prefix stripping right — the thing that already caused the subpath bug earlier in this PR. I have put the numbers in a comment above the loop so the next reader does not have to re-derive them before "optimising" it.

Happy to flatten it anyway if you would rather have the single-pass version; it is a contained change.

leecalcote (changes requested):
Thank you for working on this, @Gauravtiwari31. 

1. Missing hyperlinks to these files. These should be visible to site visitors on each page). 
2. Exclude LLM txt and markdown generation from the standard contributor flow. Instead augment the make targets to facilitate building of these files. 
3. Update the website’s GitHub workflow configuration to include generation of these files.

Gauravtiwari31:
Thanks @leecalcote  — all three make sense. Confirming how I plan to take them:

1. Adding a per-page "View as Markdown" link plus a site-wide /llms.txt link. Should the per-page link sit in the page footer next to "Edit this page", or would you rather it were in the header near the title?

2. Gating the three output formats behind a Hugo config environment so they're off in the default hugo server flow, with a dedicated make target to build them. Worth noting the default build currently produces 2,810 index.md files and a 3.32 MB llms-full.txt, so this should measurably speed up local iteration.

3. Wiring that make target into the docs deploy workflow so production still publishes them.

One interaction between 1 and 2: if generation is off by default, the per-page Markdown links would 404 in a standard contributor build. I'll make the link conditional on the same flag so it only renders when the outputs exist. Say the word if you'd rather the links always render.

Also folding in the two outstanding CodeRabbit findings in the same push — the zero-date guard, and my reasoning for keeping the bulk groups out of llms-full.txt.

yi-nuo426:
docs/config-llms.toml:12-17 scopes the LLM-specific output formats to a dedicated config, and .github/workflows/hugo.yml:96 explicitly opts the deployment workflow into it. This preserves normal contributor builds while ensuring the published site includes the new outputs. No changes requested.

PR #21678, opened 2026-08-30
[mesheryctl] Check HTTP status code before unmarshaling GitHub API response in platform.go

### LINKED ISSUE #21675: [mesheryctl] GetManifestTreeURL and ListManifests silently misinterpret non-200 GitHub API responses
### Current Behavior

`GetManifestTreeURL()` and `ListManifests()` in
[`[mesheryctl/pkg/utils/platform.go](https://github.com/meshery/meshery/blob/master/mesheryctl/pkg/utils/platform.go)`](https://github.com/meshery/meshery/blob/master/mesheryctl/pkg/utils/platform.go)
call `http.Get()` against the GitHub API and immediately `json.Unmarshal`
the response body — without ever checking `resp.StatusCode`.

When GitHub rate-limits the request (`403`) or returns any other non-200
response, the error body (e.g. `{"message": "API rate limit exceeded..."}`)
silently unmarshals into an empty `ManifestList{}`. `GetManifestTreeURL`
then falls through to a completely unrelated, misleading error:

```
could not find path: install/deployment_yamls/k8s in the manifest tree
```

A user hitting a GitHub rate limit (which is common — GitHub's
unauthenticated API limit is only 60 requests/hour) sees this "not found"
message instead of anything indicating what actually went wrong, making it
very hard to self-diagnose.

### Desired Behavior

Both functions should check `resp.StatusCode` after reading the response
body and, if it isn't `http.StatusOK`, return an error that includes the
actual status code and response body — so users see something like:

```
unexpected status 403 from https://api.github.com/repos/...: {"message": "API rate limit exceeded for ..."}
```

instead of the misleading "could not find path" message.

### Screenshots/Logs

Reproduced by pointing the function at a mock server returning `403`:

```go
resp, err := http.Get(url)
if err != nil {
    return "", errors.Wrapf(err, "failed to make GET request to %s", url)
}
defer SafeClose(resp.Body)
body, err := io.ReadAll(resp.Body)
if err != nil {
    return "", errors.Wrap(err, "failed to read response body")
}
// <-- no resp.StatusCode check here before unmarshal
var manLis ManifestList
err = json.Unmarshal([]byte(body), &manLis)
```

Confirmed against current `master` (commit `1863b6e`) — both
`GetManifestTreeURL` (~L109-127) and `ListManifests` (~L137-153) share this
gap.

### Environment

- Host OS: Mac Linux Windows
- Meshery Client Version: stable-v (issue is in the client-side `mesheryctl` binary, independent of server version)
- Platform: Docker Kubernetes

---

### Contributor [[Guides](https://docs.meshery.io/project/contributing)](https://docs.meshery.io/project/contributing) and [[Handbook](https://meshery.io/community#handbook)](https://meshery.io/community#handbook)
- 📚 [[mesheryctl Contributing Guide](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)](https://github.com/meshery/meshery/blob/master/mesheryctl/README.md)
- 📺 [[Beginner's guide to contributing to Meshery and mesheryctl](https://youtu.be/hh_kFLZx3G4)](https://youtu.be/hh_kFLZx3G4)
- ⌨️ [[mesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- ⌨️ [[Meshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- ⌨️ [[Meshkit errors Guide](https://docs.meshery.io/project/contributing/contributing-error)](https://docs.meshery.io/project/contributing/contributing-error)
- 🙋🏾🙋🏼 Questions: [[Discussion Forum](https://discuss.meshery.io/)](https://discuss.meshery.io) and [[Community Slack](https://slack.meshery.io/)](https://slack.meshery.io)
- 📺 [[Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)](https://meshery.io/talks-and-trainings#trainings)

Indrapal-70:
**Notes for Reviewers**

* This PR fixes [[mesheryctl] GetManifestTreeURL and ListManifests silently misinterpret non-200 GitHub API responses #21675](https://github.com/meshery/meshery/issues/21675)

`GetManifestTreeURL()` and `ListManifests()` in `mesheryctl/pkg/utils/platform.go` unmarshaled the HTTP response body into `ManifestList` without checking `resp.StatusCode`. When the GitHub API rate-limited (403) or returned any other error status, the error JSON body was silently parsed as an empty tree, causing `GetManifestTreeURL()` to fall through to a misleading `"could not find path: install/deployment_yamls/k8s in the manifest tree"` error instead of surfacing the real cause.

Fix: added a `resp.StatusCode != http.StatusOK` check in both functions, placed after reading the response body so the actual status code and response message are included in the returned error. Added `TestGetManifestTreeURL` and `TestListManifests` covering the 403 rate-limit case and the existing happy path, using `httptest.NewServer`.

Note: `TestNavigateToBrowser` fails in this local dev environment due to missing xdg-open/browser binaries (headless WSL) — confirmed pre-existing and unrelated to this change via `git stash` comparison.

[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)

[Before]
Screenshot 2026-08-30 135010

[After]Screenshot 2026-08-30 134453

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of unexpected GitHub API responses, including clearer errors for rate limits and other failures.
  - Prevented failures when processing pod names with fewer than two segments.
  - Added guidance in API error messages to help diagnose service availability and rate-limiting issues.

- **Tests**
  - Expanded coverage for successful and failed GitHub API requests.
  - Added validation for standard, numeric, single-segment, and empty pod names.


ishwar170695 (commented):
LGTM. One minor test suggestion: since the fix is meant to surface the actual GitHub API error, could we also assert that the response message is included in the returned error, not just the `403` status?

```go
if !strings.Contains(err.Error(), "API rate limit exceeded") {
    t.Errorf("expected error to contain rate limit message, got: %v", err)
}
```

Otherwise, this looks good to me.

Indrapal-70:
Hi @ishwar170695 , thanks for the review . I added that assertion to both TestGetManifestTreeURL and TestListManifests and pushed both of those changes in the PR could you review it again .

mesheryctl/pkg/utils/platform.go
```diff
@@ -119,6 +120,10 @@ func GetManifestTreeURL(version string) (string, error) {
 		return "", errors.Wrap(err, "failed to read response body")
 	}
 
+	if resp.StatusCode != http.StatusOK {
+		return "", errors.Errorf("unexpected status %d from %s: %s", resp.StatusCode, url, string(body))
```
lekaf974: Since you are introducing a new error this must be a meshkit error

mesheryctl/pkg/utils/platform.go
```diff
@@ -146,6 +151,10 @@ func ListManifests(url string) ([]Manifest, error) {
 		return nil, errors.Wrap(err, "failed to read response body")
 	}
 
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.Errorf("unexpected status %d from %s: %s", resp.StatusCode, url, string(body))
```
lekaf974: Since you are introducing a new error this must be a meshkit error

mesheryctl/pkg/utils/platform_test.go
```diff
@@ -15,18 +15,120 @@
...
+		origURL := gitHubBaseURL
+		gitHubBaseURL = server.URL
+		defer func() { gitHubBaseURL = origURL }()
+
+		_, err := GetManifestTreeURL("v0.6.0")
+		if err == nil {
+			t.Fatalf("expected error, got nil")
+		}
+
+		if !strings.Contains(err.Error(), "403") {
```
lekaf974: Use assert not string validation in error

lekaf974 (changes requested):
HI good starting point, can you address comments added

Indrapal-70:
Hi @lekaf974, thanks for the review .  I've updated the PR to address both points:

1. **MeshKit structured error**: Added a new structured error `ErrGithubAPIResponse` (`mesheryctl-1254`) in `mesheryctl/pkg/utils/error.go`, following the existing convention in this file (see `ErrResponseStatusBody` as the closest sibling). Replaced the `errors.Errorf(...)` calls in `GetManifestTreeURL` and `ListManifests` with calls to `ErrGithubAPIResponse`. I opted for a new error rather than reusing `ErrResponseStatus`/`ErrResponseStatusBody` since their remediation text points users toward `mesheryctl system status`, which wouldn't make sense for a GitHub API failure.
2. **Test assertions**: Replaced the manual `strings.Contains` + `t.Errorf` checks in `platform_test.go` with `testify`'s `require`/`assert`, matching the pattern already used elsewhere in this package.

mesheryctl/pkg/utils/platform.go
```diff
@@ -33,6 +33,7 @@ var (
 	// ManifestsFolder is where the Kubernetes manifests are stored
 	ManifestsFolder = "manifests"
 	ReleaseTag      string
+	gitHubBaseURL   = "https://api.github.com"
```
lekaf974, line 36:
```suggestion
	githubBaseURL   = "https://api.github.com"
```

mesheryctl/pkg/utils/platform_test.go
```diff
@@ -15,18 +15,93 @@
...
+		}))
+		defer server.Close()
+
+		origURL := gitHubBaseURL
+		gitHubBaseURL = server.URL
+		defer func() { gitHubBaseURL = origURL }()
+
+		_, err := GetManifestTreeURL("v0.6.0")
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "403")
```
lekaf974: please use dedicated meshkit error test function https://github.com/meshery/meshery/blob/c6e1a32a9761216c59d512bc6a4efe92bda082f7/mesheryctl/pkg/utils/testing.go#L505

mesheryctl/pkg/utils/platform_test.go
```diff
@@ -15,18 +15,93 @@
...
+	t.Run("returns error on non-200 HTTP status code (e.g. 403 Rate Limit)", func(t *testing.T) {
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusForbidden)
+			_, _ = w.Write([]byte(`{"message": "API rate limit exceeded"}`))
+		}))
+		defer server.Close()
+
+		_, err := ListManifests(server.URL)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "403")
```
lekaf974: same here

lekaf974 (changes requested):
please reuse existing testing method

ishwar170695 (commented):
The tests are still using `assert/require` and checking the error string. Since `AssertMeshkitErrorsEqual` was specifically requested here, can we use the existing helper for both error cases and compare against the expected `ErrGitHubAPIResponse(...)` instead?

Also, the new `mesheryctl-1255` error is still missing from `docs/data/errorref/mesheryctl_errors_export.json`. Can we regenerate/update the error reference export as well?

lekaf974:
> docs/data/errorref/mesheryctl_errors_export.json

this file is auto generated on merge in master

Indrapal-70:
@lekaf974 @ishwar170695 I have done the changes required could you please give the final review ?

mesheryctl/pkg/utils/platform.go
lekaf974: still waiting for this variable name to be updated

vedant21-ctr (commented):
The core fix looks good and the tests/CI are passing. Just two small things before approval:

Could we rename gitHubBaseURL to githubBaseURL for consistent GitHub casing?
Please add a short Go doc comment for the new exported ErrGitHubAPIResponse function.

lekaf974 (approved):
LGTM

yi-nuo426:
mesheryctl/pkg/utils/platform.go:123 and :154 now read GitHub API error bodies and reject non-200 responses before unmarshaling them as manifest trees. This preserves the real rate-limit or API failure instead of returning a misleading missing-path result. No changes requested.

PR #21680, opened 2026-08-30
[UI] Forward options through initiateQuery so getDesign force-refetches

### LINKED ISSUE #21164: [UI] initiateQuery silently drops its options argument, so getDesign never force-refetches
### Description

`initiateQuery` accepts two parameters, but two call sites pass a third options argument that is **silently dropped**. The most consequential one is `getDesign`, which believes it is forcing a refetch and is not.

```js
// ui/rtk-query/utils.ts:9
export const initiateQuery = async (query, variables) => {
  try {
    return await store.dispatch(query.initiate(variables));
    //                                        ^ no options forwarded
```

```js
// ui/rtk-query/design.ts:193-200
export const getDesign = async ({ design_id }) => {
  return await initiateQuery(
    designsApi.endpoints.getDesign,
    { design_id },
    {
      forceRefetch: true,   // <-- dropped on the floor
    },
  );
};
```

JavaScript discards extra arguments without error, so there is no warning at any layer - not at author time, not at lint (`initiateQuery` is untyped JS), and not at runtime.

### Impact

`getDesign()` never force-refetches; it resolves from RTK Query's cache whenever a cached entry exists. The consumer is the design info modal:

```js
// ui/components/workspaces/SpacesSwitcher/MainDesignsContent.tsx:122-127
const handleInfoModal = async (design) => {
  const selectedDesignWithPatternFile = await getDesign({ design_id: design?.id });
  setSelectedDesign(selectedDesignWithPatternFile?.data);
```

The whole reason this helper exists is to fetch the full design **including `patternFile`**, which the list payload does not carry. With the cache serving the request, a design that was edited elsewhere - or whose cached entry predates an update - shows stale content in the info modal.

`ui/rtk-query/user.ts:373` passes a third argument too (`initiateQuery(userApi.endpoints.getAccessToken, {}, {})`). That one is inert since the object is empty, but it is the same misunderstanding of the signature and is worth cleaning up alongside.

### Expected

`initiateQuery` should forward options through to `query.initiate`:

```js
export const initiateQuery = async (query, variables, options) =>
  await store.dispatch(query.initiate(variables, options));
```

### Why this is filed rather than fixed inline

Found during the design-import canonical-sourcing audit (follow-up to #21114). `initiateQuery` is a shared utility used across many `ui/rtk-query/*` modules and `ui/machines/validator/designValidator.ts`. Making the parameter live changes refetch behaviour for `getDesign` immediately, and any future call site that passes options starts behaving differently than it does today - so it warrants its own change with its own testing, not a drive-by in a wire-contract audit.

Note for whoever picks this up: `ui/rtk-query/__tests__/utils.test.ts` covers `initiateQuery` but only ever calls it with two arguments, so the existing suite will not catch the regression class either way. Add a case asserting options reach `query.initiate`.

Gauravtiwari31:
## Description

`initiateQuery` in `ui/rtk-query/utils.ts` accepted only `(query, variables)`, so the third argument passed by two call sites was silently discarded — JavaScript drops extra arguments without error, and the helper is untyped JS, so nothing flagged it at author time, at lint, or at runtime.

The consequential one is `getDesign` in `ui/rtk-query/design.ts`, which passes `{ forceRefetch: true }` and never actually forced a refetch. It resolved from the RTK Query cache whenever a cached entry existed. That helper exists precisely to fetch the full design including `patternFile`, which the list payload does not carry, so the design info modal (`ui/components/workspaces/SpacesSwitcher/MainDesignsContent.tsx`) could show stale content for a design edited elsewhere.

## Changes

- **`ui/rtk-query/utils.ts`** — add the `options` parameter and forward it to `query.initiate(variables, options)`; document it in the JSDoc. RTK Query's `initiate` defaults its second parameter to `{}`, so the existing two-argument call sites (`user.ts`, `view.ts`, `meshModel.ts`, `machines/validator/designValidator.ts`) are unaffected.
- **`ui/rtk-query/user.ts`** — drop the inert `{}` third argument from `getUserAccessToken`, as noted in the issue.
- **`ui/rtk-query/__tests__/utils.test.ts`** — add a case asserting options reach `query.initiate`. The existing `dispatches query.initiate(variables)` case pinned the exact argument list via `toHaveBeenCalledWith`, so it is updated to `toHaveBeenCalledWith(variables, undefined)`.

`ui/rtk-query/design.ts` needs no change — `getDesign`'s `forceRefetch` starts taking effect once the utility forwards it, which is the point of the fix.

## Testing

- `vitest run rtk-query/__tests__/utils.test.ts` — 5 passed (5).
- Regression check: temporarily reverting `query.initiate(variables, options)` back to `query.initiate(variables)` fails 2 of the 5, including the new case, confirming the test pins the behaviour rather than passing vacuously.
- `prettier --check` and `eslint` clean on all three files.

Fixes #21164

## Summary by CodeRabbit

* **New Features**
  * Query initiation now supports optional configuration options, which are forwarded when requests are started.
* **Bug Fixes**
  * Improved consistency in user access-token requests by using the standardized request argument format.
  * Added coverage to verify requests behave correctly when options are omitted or explicitly provided.
  * Preserved existing query behavior when no additional configuration is supplied.


yi-nuo426:
ui/rtk-query/utils.ts:43 now forwards options as the second argument to query.initiate, making getDesign's existing forceRefetch: true option in ui/rtk-query/design.ts:190 effective. Callers that omit options preserve RTK Query's default behavior. No changes requested.

PR #21682, opened 2026-08-30
fix(mesheryctl): prevent nil response panic in auth update

### LINKED ISSUE #21681: [Bug] mesheryctl: Fix nil-pointer panic in UpdateAuthDetails when server is unreachable
### Current Behavior

When `UpdateAuthDetails` is executed in `mesheryctl` while the Meshery server is unreachable or offline, `client.Do(req)` returns an error alongside a `nil` HTTP response (`resp == nil`).

`UpdateAuthDetails` currently executes `defer SafeClose(resp.Body)` before checking if `err != nil`. Evaluating `resp.Body` on a `nil` response pointer causes a runtime nil-pointer dereference panic, which crashes the CLI.

### Desired Behavior

`UpdateAuthDetails` should evaluate `err` immediately after `client.Do(req)` and return the error before accessing `resp.Body`. When the server is unreachable or the network request fails, `mesheryctl` should exit with a readable error message instead of panicking.

### Screenshots/Logs

```text
panic: runtime error: invalid memory address or nil pointer dereference

[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x...]

goroutine 1 [running]:

github.com/meshery/meshery/mesheryctl/pkg/utils.UpdateAuthDetails(0x...)

    /mesheryctl/pkg/utils/auth.go:235 +0x...



```

### Environment

- Host OS: macOS, Linux, Windows
- Meshery Server Version: edge / master
- Meshery Client Version: edge / master
- Platform: Docker, Kubernetes

---

- ⌨️ [svgmesheryctl Command Tracker](https://docs.google.com/spreadsheets/d/1q63sIGAuCnIeDs8PeM-0BAkNj8BBgPUXhLbe1Y-318o/edit#gid=0)
- ⌨️ [svgMeshery CLI Commands and Documentation](https://docs.google.com/document/d/1xRlFpElRmybJ3WacgPKXgCSiQ2poJl3iCCV1dAalf0k/edit#heading=h.5fucij4hc5wt)
- ⌨️ [Meshkit Errors Guide](https://docs.meshery.io/project/contributing/contributing-error)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.meshery.io/) and [Community Slack](https://slack.meshery.io/)

Kanika0306:
### Description

`client.Do(req)` can return `resp == nil` alongside an error when a network request fails, such as when the Meshery server is unreachable or offline. The previous implementation in `UpdateAuthDetails` evaluated `defer SafeClose(resp.Body)` before checking `if err != nil`, causing a runtime nil-pointer dereference panic in `mesheryctl`.

This PR fixes the issue by moving `defer SafeClose(resp.Body)` immediately after the `if err != nil` check, ensuring `resp.Body` is only accessed after a successful HTTP request.

### Current Behavior

When `UpdateAuthDetails` is executed in `mesheryctl` while the Meshery server is unreachable, `client.Do(req)` can return `(nil, err)`. The function previously attempted to access `resp.Body` before handling the error, resulting in a runtime panic:

`panic: runtime error: invalid memory address or nil pointer dereference`

### Desired Behavior

`UpdateAuthDetails` now evaluates the request error immediately after `client.Do(req)`. If the request fails, the function returns the wrapped error before accessing `resp.Body`.

When the server is unreachable or the network request fails, `mesheryctl` now exits cleanly with a readable error message instead of panicking.

Example error:

`error dispatching request: connection refused`

### Changes Made

- Updated `mesheryctl/pkg/utils/auth.go`.
- Moved `defer SafeClose(resp.Body)` to execute only after the request error has been checked.
- Added deterministic regression tests in `mesheryctl/pkg/utils/auth_test.go`.
- Added coverage for the `(nil, error)` response scenario.
- Verified successful authentication and update behavior remains unchanged.
- Verified existing HTML response error handling remains intact.
- Verified the non-existent token-file error path remains intact.

### Regression Tests Added

The following scenarios are covered:

1. `client.Do` returns `(nil, error)`
   - `UpdateAuthDetails` returns the error cleanly.
   - No nil-pointer panic occurs.

2. Successful HTTP response
   - Authentication details are processed successfully.
   - Token file is updated as expected.

3. HTML error response
   - Invalid response body handling continues to work correctly.

4. Non-existent token file
   - The expected token-file error is returned.

### Test Results

All `TestUpdateAuthDetails` regression tests pass successfully, including:

- Nil response with request error
- Successful response and token update
- HTML response error handling
- Non-existent token file handling

Test result: `PASS`

The regression tests for `UpdateAuthDetails` pass successfully:

    === RUN   TestUpdateAuthDetails
    === RUN   TestUpdateAuthDetails/Given_client.Do_returns_nil_response_and_error,_When_UpdateAuthDetails_is_called,_Then_it_returns_error_without_panic
    === RUN   TestUpdateAuthDetails/Given_client.Do_returns_success_response,_When_UpdateAuthDetails_is_called,_Then_token_is_updated_successfully
    === RUN   TestUpdateAuthDetails/Given_client.Do_returns_HTML_response,_When_UpdateAuthDetails_is_called,_Then_it_returns_invalid_body_error
    === RUN   TestUpdateAuthDetails/Given_non-existent_token_file,_When_UpdateAuthDetails_is_called,_Then_it_returns_error_reading_token
    --- PASS: TestUpdateAuthDetails (0.03s)

All test cases passed successfully.

### Environment

- Host OS: macOS, Linux, Windows
- Meshery Server Version: edge / master
- Meshery Client Version: edge / master
- Platform: Docker, Kubernetes

### Verification

The fix has been committed and pushed to the remote branch `fix/update-auth-details-panic`.

Commit: `958a14ba3da`

Commit message: `fix(mesheryctl): prevent nil response panic in auth update`

All added regression tests pass successfully, and existing successful-response and error-handling behavior remains unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected authentication request error messages for improved clarity.
  * Improved handling of authentication response validation and request failures.
  * Added more reliable authentication processing when refreshing credentials.
  * Improved handling of missing authentication token files.

* **Tests**
  * Expanded coverage for successful refreshes, request failures, invalid responses, and missing token files.
  * Improved test isolation and cleanup for more consistent results.

### DIFF
--- a/mesheryctl/pkg/utils/auth.go
+++ b/mesheryctl/pkg/utils/auth.go
@@ -232,12 +232,11 @@ func UpdateAuthDetails(filepath string) error {
 
 	client := &http.Client{}
 	resp, err := client.Do(req)
-	defer SafeClose(resp.Body)
-
 	if err != nil {
-		err = errors.Wrap(err, "error dispatching there request: ")
+		err = errors.Wrap(err, "error dispatching the request: ")
 		return err
 	}
+	defer SafeClose(resp.Body)
 
 	data, err := io.ReadAll(resp.Body)
 	if err != nil {


banana-three-join (commented):
Appreciate the fix!

`defer` arguments are evaluated at the point the `defer` statement executes, so `defer SafeClose(resp.Body)` on `master` dereferences `resp` before the `err != nil` check ever runs. That's the panic. And there's no leak risk in the reordered version: per the `Client.Do` contract, on error the response can be ignored, and the only case where a non-nil response accompanies a non-nil error is a `CheckRedirect` failure, where the body is already closed. Moving the `defer` below the error check is exactly the idiomatic ordering. The error-message typo fix is a welcome bonus.

A few suggestions on the tests:

**1. Avoid mutating `http.DefaultTransport`.** Reassigning a stdlib global makes these subtests unable to run with `t.Parallel()` and lets them race with anything else in the package that uses the default client. Injecting the client scopes the stub to the test:

```go
// auth.go
func UpdateAuthDetails(filepath string) error {
	return updateAuthDetails(&http.Client{}, filepath)
}

func updateAuthDetails(client *http.Client, filepath string) error {
	// existing body, minus `client := &http.Client{}`
}
```

If changing the signature isn't desirable, a package-level `var httpClient = &http.Client{}` swapped in tests is still an improvement — the blast radius stays inside the package.

**2. Prefer `httptest.NewServer` for the success and HTML cases.** Hand-assembling an `*http.Response` with a `NopCloser` skips real serialization, header handling, and body closing. A test server exercises the actual path:

```go
srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
	fmt.Fprint(w, `{"token":"refreshed-token","meshery-provider":"Local"}`)
}))
defer srv.Close()
```

The `roundTripFunc` stub is still the right tool for the `(nil, error)` case, since that failure can't be produced by a real server.

**3. Isolate the token file per subtest.** All four subtests share one `tokenFile`, and the success case rewrites it — that makes the ordering between subtests load-bearing. A small helper that writes a fresh token file into `t.TempDir()` per subtest removes the coupling.

**4. `viper.Reset()` + reassigning `DefaultConfigPath`** is the same global-state pattern as (1). If the config path can be threaded through instead, the `t.Cleanup` dance goes away entirely.

None of these block the fix itself — the behavior change is right and the added coverage is genuinely valuable. These are about making the tests order-independent and parallel-safe.

Kanika0306:
@banana-three-join 
could you review this 
I have made the changes in this.
Thankyou

banana-three-join (changes requested):
Please keep the solution scoped only to:
```
if err != nil {
		err = errors.Wrap(err, "error dispatching the request: ")
		return err
	}
defer SafeClose(resp.Body)
```

After that's done, squash the changes into one commit so I can approve and merge, ty!

banana-three-join (approved):

Kanika0306:
Hi @banana-three-join  How do I connect to the meshery Communication channel could you guide me 
Thankyou

banana-three-join:
Ty! Really appreciate it. I would recommend joining our Slack channel over at https://slack.meshery.io/ so we can provide you better guidance.

yi-nuo426:
mesheryctl/pkg/utils/auth.go:239 now defers response-body cleanup only after client.Do succeeds. Since a network failure can return a nil response, this ordering returns the dispatch error without dereferencing resp.Body and eliminates the CLI panic. No changes requested.

PR #21685, opened 2026-08-30
[UI] Let LegacyRJSFModal disable Submit while it is pending

### LINKED ISSUE #21227: LegacyRJSFModal: no way to disable Submit while the modal is pending (enabled button is inert)
### Current Behavior

`LegacyRJSFModal` renders `PrimaryActionButtons` in `ModalFooter` **outside** the `loadingSchema` branch (`components/shared/Modal/LegacyRJSFModal.tsx:171-181`), so the Submit button is on screen and enabled while the modal is still in its pending state.

Its only disable input is `canNotSubmit`, which is derived solely from the modal title matching `'untitled design'` / `'Untitled'` / `'lfx'`. Any other title - for example `"Help & Support"` - renders Submit enabled throughout.

Clicking Submit during that window reaches `handleFormSubmit`, whose `if (formRef.current && ...)` guard is false because `RJSFWrapper` is not mounted yet. The click therefore does nothing at all: no close, no submit, no feedback.

So a consumer that legitimately opens the modal shell before its form is ready gets an enabled control that silently does nothing - a dead click - and has no way to prevent it.

### Desired Behavior

`LegacyRJSFModal` should let a consumer disable Submit for reasons other than the title heuristic - for example an explicit `submitDisabled` (or `canNotSubmit`) prop that is OR-ed with the existing title-derived value, and/or disabling Submit automatically whenever the modal is rendering its `loadingSchema` pending state, since in that state the submit path is inert by construction.

Either would let consumers gate Submit on their own readiness rather than presenting an enabled button that cannot work.

### Context

Found in Kanvas (layer5labs/meshery-extensions) while making Help & Support open its modal shell immediately with the host's pending affordance, so that a slow profile lookup never makes the click read as a no-op.

The safety property there is intact - the submission cannot proceed without a resolved identity, and the extension-side guard covering that is unreachable precisely because the host's own `formRef` guard already blocks the click. The remaining gap is purely presentational: an enabled button during the pending window that does nothing when pressed.

Kanvas is shipping with its documentation corrected to describe the current host behaviour rather than working around it, and will gate on this prop once it exists.

Gauravtiwari31:
## Description

`LegacyRJSFModal` renders `PrimaryActionButtons` in `ModalFooter` **outside** the `loadingSchema` branch, so Submit is on screen and enabled while the modal is still pending. Its only disable input was `canNotSubmit`, derived solely from the title matching `'untitled design'` / `'Untitled'` / `'lfx'` — so any other title (for example `"Help & Support"`) rendered Submit enabled throughout.

Clicking it during that window reaches `handleFormSubmit`, whose `if (formRef.current && ...)` guard is false because `RJSFWrapper` is not mounted yet. The click therefore does nothing at all: no submit, no close, no feedback — a dead click a consumer had no way to prevent.

## Changes

`ui/components/shared/Modal/LegacyRJSFModal.tsx`, both suggestions from the issue:

- **`loadingSchema`** is OR-ed into the disabled state. In that state the submit path is inert by construction, so this only stops an inert button from presenting as clickable.
- **A new `submitDisabled` prop** (default `false`) lets a consumer gate Submit on its own readiness, OR-ed with the modal's own reasons.

| component | before | after |
|---|---|---|
| `Modal` (also used by Meshery extensions) | `canNotSubmit` | `canNotSubmit \|\| loadingSchema \|\| submitDisabled` |
| `RJSFModalWrapper` | `canNotSubmit \|\| isSubmitting` | `canNotSubmit \|\| loadingSchema \|\| isSubmitting \|\| submitDisabled` |

`RJSFModalWrapper` already disabled on `isSubmitting` but not on `loadingSchema`, so it gains both inputs for consistency.

Nothing enabled today becomes disabled once a schema resolves: `loadingSchema` flips to `false` and the button behaves exactly as before. `submitDisabled` defaults to `false`, so consumers that do not pass it are unaffected.

## Testing

New `ui/components/shared/Modal/__tests__/LegacyRJSFModal.submitDisabled.test.tsx`, 5 cases: disabled while pending, enabled once the schema resolves, disabled when `submitDisabled` is set, `submitDisabled` defaulting to false, and the existing title-derived heuristic still firing. It renders the real Sistent components inside `SistentThemeProvider` (per the note in `vitest.config.ts` about shallow `@sistent/sistent` stubs), mocking only `RJSFWrapper` and `useNotification`.

Verified the tests are not vacuous: reverting the `Modal` disable expression to `canNotSubmit` fails exactly the two cases that cover the new behaviour, and passes the other three.

Fixes #21227

## Summary by CodeRabbit

- **New Features**
  - Added an optional setting to disable modal submission buttons when needed.
  - Submission buttons now remain disabled while schemas are loading or submissions are in progress.

- **Bug Fixes**
  - Prevented submission when modal validation conditions indicate it is unavailable.

- **Tests**
  - Added coverage for disabled, enabled, loading, and validation-related submission states.

### DIFF
--- a/ui/components/shared/Modal/LegacyRJSFModal.tsx
+++ b/ui/components/shared/Modal/LegacyRJSFModal.tsx
@@ -75,6 +75,9 @@ const SchemaVersion: FC<SchemaVersionProps> = ({ schema_array, type, schemaChang
  * @param {Object} props.showInfoIcon - Determines whether to show the info icon adjacent to the modal button.
  * @param {string} props.submitBtnText - The text for the submit button.
  * @param {Object} props.uiSchema - The UI schema for the form fields.
+ * @param {boolean} props.submitDisabled - Disables the submit button for a consumer-owned
+ * reason, OR-ed with the modal's own reasons (a forbidden title, and the pending state while
+ * the schema is still loading). Defaults to false.
  */
 
 // Meshery extensions also uses this modal
@@ -94,6 +97,7 @@ function Modal(props) {
     helpText,
     RJSFWrapperComponent = null,
     initialData = {},
+    submitDisabled = false,
   } = props;
 
   const [canNotSubmit, setCanNotSubmit] = useState(false);
@@ -175,7 +179,7 @@ function Modal(props) {
             secondaryText="Cancel"
             primaryButtonProps={{
               onClick: handleFormSubmit,
-              disabled: canNotSubmit,
+              disabled: canNotSubmit || loadingSchema || submitDisabled,
             }}
             secondaryButtonProps={{
               onClick: handleClose,
@@ -201,6 +205,7 @@ function RJSFModalWrapper({
   helpText,
   widgets = {},
   isSubmitting = false,
+  submitDisabled = false,
 }) {
   const formRef = useRef();
   const formStateRef = useRef();
@@ -270,7 +275,7 @@ function RJSFModalWrapper({
           secondaryText="Cancel"
           primaryButtonProps={{
             onClick: handleFormSubmit,
-            disabled: canNotSubmit || isSubmitting,
+            disabled: canNotSubmit || loadingSchema || isSubmitting || submitDisabled,
             startIcon: isSubmitting ? <CircularProgress size={16} /> : undefined,
           }}
           secondaryButtonProps={{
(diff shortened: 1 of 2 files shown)


yi-nuo426:
ui/components/shared/Modal/LegacyRJSFModal.tsx:182 and :278 gate Submit on loadingSchema in both modal implementations, preventing the previously inert click before RJSFWrapper assigns formRef. The defaulted submitDisabled prop leaves existing callers enabled after schema resolution. No changes requested.

PR #21686, opened 2026-08-30
[UI] Keep the Notification Center icon out of Redux as a serializable name

### LINKED ISSUE #16873: [UI] Redux Warning: Non-serializable value in Notification Center state (events.ui.icon)
### Current Behavior
While inspecting the Redux state in the browser console during development, a consistent warning appears regarding a "Non-serializable value detected in the state," specifically at the path `events.ui.icon`.

The application is currently storing React Components (SVG Icons) directly in the Redux store. This violates Redux best practices (which state that the store should only contain serializable data like plain objects, arrays, strings, and numbers) and causes performance warnings in the console.

### Desired Behavior
The Redux store should only contain serializable data. The `events.ui.icon` field should store a string identifier (e.g. info-icon, warning-icon) or be removed from the store entirely.

The mapping of that identifier to the actual React Icon Component (SVG) should happen at the UI component level (e.g., inside `Notification.js` or `NotificationCenter.js`) during rendering, not inside the global state.

### Screenshots/Logs

Image

Error Log:

```plaintext
installHook.js:1 
A non-serializable value was detected in the state, in the path: `events.ui.icon`. 
Value: ({width:e=V,height:t=W,...r})=>(0,DT.jsx)("svg",{width:e,height:t,xmlns:"http://www.w3.org/2000/svg",viewBox:"0 0 24 24",...r ...
...
Take a look at the reducer(s) handling this action type: mesheryApi/executeQuery/pending.
```

### Environment

- Browser: Chrome  
- Host OS: Windows
- Meshery Server Version: latest (master branch)
- Platform: Docker 

---

Gauravtiwari31:
## Description

`components/layout/NotificationCenter/index.tsx:219` read its icon straight out of Redux:

```js
const Icon = uiConfig.icon || BellIcon;
```

so a React SVG component sat in the store at `events.ui.icon` and tripped RTK's serializable-state check — the warning in the issue.

It gets there from outside this repo. `store/index.ts` subscribes to `MESHERY_EXTENSION_EVENT.DispatchToMesheryStore` and forwards whatever a Meshery extension puts on the event bus straight to `dispatch`; the `openNotificationCenter` reducer then spread `action.payload.ui` into `state.ui` unfiltered. Nothing in this repo dispatches it, which is why the value only appears with extensions loaded.

## Changes

| File | Change |
|---|---|
| `components/layout/NotificationCenter/notificationIcons.ts` *(new)* | Name→component registry (`bell`, `alert`, `error`, `info`) and `resolveNotificationIcon` |
| `components/layout/NotificationCenter/index.tsx` | `uiConfig.icon \|\| BellIcon` → `resolveNotificationIcon(uiConfig.icon)` |
| `store/slices/events.ts` | Screen non-serializable values out of `ui` in `openNotificationCenter`; export `openNotificationCenter` |

Two deliberate choices:

- **The screen is general, not icon-specific.** `isSerializable` rejects functions and symbols and recurses through objects and arrays, so a React element (caught by its `$$typeof` symbol) or a stray callback under any `ui` key is dropped, not just `icon`.
- **`resolveNotificationIcon` is total.** An unknown name, a missing one, or a component sent by an older extension all resolve to the bell, so the header always has something to render.

Also exports `openNotificationCenter` from the slice — it was the only reducer missing from the action-creator export list, so no in-repo code could dispatch it by creator.

## Behaviour change worth reviewing

An extension still passing a React component now renders the bell instead of its own icon, until that icon is added to `NOTIFICATION_ICONS`. That is the trade the issue asks for ("store a string identifier … the mapping … should happen at the UI component level"), and adding an icon is a one-line entry — but it is extension-visible, so I would rather flag it here than have it found later.

## Testing

- `components/layout/NotificationCenter/__tests__/notificationIcons.test.ts` — resolution for registered names, unknown names, `null`/`undefined`/`''`, a legacy component, and a totality check that no input yields `undefined`.
- `store/slices/__tests__/events.test.ts` — a component, a React element, and a nested function are all dropped while serializable siblings survive; the raw action object is screened too (the event-bus path never goes through the action creator); `closeNotificationCenter` still resets to serializable defaults.

20/20 targeted tests pass. Verified they are not vacuous: reverting the screen in the reducer fails exactly 4 of them. Full vitest suite run twice — 2788 passed, 0 failures both times. ESLint clean on all five files.

Fixes #16873

## Summary by CodeRabbit

- **New Features**
  - Added support for configurable Notification Center icons, including bell, alert, error, and info options.
  - Invalid or legacy icon values now safely fall back to the default bell icon.
  - Notification Center UI settings now retain only serializable values for more reliable behavior.

- **Bug Fixes**
  - Prevented unsupported data such as callbacks, components, and undefined values from affecting notification state.

### DIFF
--- a/ui/components/layout/NotificationCenter/__tests__/notificationIcons.test.ts
+++ b/ui/components/layout/NotificationCenter/__tests__/notificationIcons.test.ts
@@ -1,0 +1,66 @@
+// The Notification Center header used to read a React component straight out of
+// Redux (`uiConfig.icon || BellIcon`), which is what put a non-serializable
+// value at `events.ui.icon` (issue #16873). The store now carries a name and the
+// header resolves it here, so this is the half that has to stay total: every
+// input, including the component an older extension might still send, must
+// produce a renderable component rather than undefined.
+
+import { describe, expect, it } from 'vitest';
+import BellIcon from '../../../../assets/icons/BellIcon';
+import AlertIcon from '../../../../assets/icons/AlertIcon';
+import {
+  DEFAULT_NOTIFICATION_ICON,
+  NOTIFICATION_ICONS,
+  resolveNotificationIcon,
+} from '../notificationIcons';
+
+describe('resolveNotificationIcon', () => {
+  it('resolves a registered name to its component', () => {
+    expect(resolveNotificationIcon('bell')).toBe(BellIcon);
+    expect(resolveNotificationIcon('alert')).toBe(AlertIcon);
+  });
+
+  it('falls back to the bell for an unknown name', () => {
+    expect(resolveNotificationIcon('no-such-icon')).toBe(BellIcon);
+  });
+
+  it('falls back to the bell when no icon is set', () => {
+    expect(resolveNotificationIcon(undefined)).toBe(BellIcon);
+    expect(resolveNotificationIcon(null)).toBe(BellIcon);
+    expect(resolveNotificationIcon('')).toBe(BellIcon);
+  });
+
+  it('falls back to the bell for a component, which the store no longer keeps', () => {
+    const LegacyIcon = () => null;
+    expect(resolveNotificationIcon(LegacyIcon)).toBe(BellIcon);
+  });
+
+  it('never returns undefined, so the header always has something to render', () => {
+    const inputs = [
+      'bell',
+      'alert',
+      'error',
+      'info',
+      'nope',
+      '',
+      null,
+      undefined,
+      42,
+      {},
+      () => null,
+    ];
+    inputs.forEach((input) => {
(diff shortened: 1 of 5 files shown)


yi-nuo426:
Thanks for moving icon resolution out of Redux. ui/store/slices/events.ts:49 still accepts non-plain objects with no enumerable properties: new Date(), new Map(), and new Set() all pass Object.values(value).every(...) even though RTK considers them non-serializable. A circular object also recurses until the reducer throws. Please restrict values to arrays and plain objects, track ancestors with a WeakSet, and add regression cases for Date, Map, and a circular object.

PR #21706, opened 2026-09-01
fix(cli): typo fix and panic fix in mesheryctl perf/environment commands

LYK-02:
## Description
Two small fixes in mesheryctl:

1. **Typo fix**: Corrected "environmnents" → "environments" in the 
   `environment view` command's help description.
2. **Panic fix**: Fixed a crash when running `mesheryctl perf ""` (empty 
   string argument). The code accessed `args[0][0]` without checking that 
   `args[0]` was non-empty, causing `index out of range [0] with length 0`.

## Before

$ mesheryctl perf ""
panic: runtime error: index out of range [0] with length 0

## After

$ mesheryctl perf ""
Error: '' is an invalid command for 'perf'. Use 'mesheryctl perf --help' to display usage guide.

## Testing
Manually verified both fixes by building locally and running the affected 
commands. Note: the existing `TestPerfCmd` test harness only supports 
asserting error cases containing specific substrings; adding a case for 
this scenario would require extending the harness itself, which felt out 
of scope for this fix — happy to follow up if maintainers want that added.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the environment viewing command description in command-line help and reference documentation.
  - Improved handling of empty or invalid performance command input to prevent errors.
  - Made invalid-command suggestions safer by validating input before comparing characters.


LYK-02:
Thanks for the review, @copilot! Good catches on both:

1. I'll extend `TestPerfCmd` with a case for `Args: []string{""}` to cover the regression — asserting the error contains "invalid" and confirming no panic.
2. I'll regenerate the mesheryctl docs so `environment/view.md` picks up the corrected help text.

Will push both updates shortly.

PARTH-TUSSLE:
Thanks for addressing the review points. I don't see the follow-up commits for them yet though. Once the regression test and regenerated CLI docs are pushed, I'll take another look👍

LYK-02:
Hi @PARTH-TUSSLE, thanks for checking in! Pushed the docs fix now — regenerated the CLI reference so environment view picks up the corrected typo.

On the test — I confirmed importing root into perf's test file causes an actual import cycle (root imports perf, so perf can't import root back), and other approaches (SetArgs+Execute, calling RunE directly) hit unrelated Cobra quirks with how empty-string args get parsed before reaching my fix. I don't want to force in a test that ends up testing the wrong thing. Would you prefer I keep digging, or ship this now and open a follow-up issue for extending the test harness? Happy either way.

emphor11:
@LYK-02 Could we add a regression case to `TestPerfCmd` for `Args: []string{""}`?

This is the exact input that previously caused the `args[0][0]` panic, and the existing table-driven test already calls `PerfCmd.SetArgs(tt.Args)` followed by `PerfCmd.Execute()`, so it should fit without introducing a new harness.

That would make sure this specific panic cannot silently return in a future change.

LYK-02:
Hey @emphor11, I actually tried that first — SetArgs([]string{""}) + Execute() doesn't reach the panic line at all, since Cobra strips empty-string args before dispatching, so it just falls through to cmd.Help() with err == nil. Trying a different approach now (external test package to go through root.RootCmd properly) — will push shortly.

LYK-02:
Found something while testing: IsValidSubcommand compares against s.CalledAs(), which returns "" for uninvoked commands — so it wrongly treats an empty arg as valid and returns nil instead of erroring, even though real CLI usage does show the correct error. Seems like a separate bug worth its own issue rather than expanding this PR's scope. Happy to file that follow-up and ship this as-is (panic fix + docs fix) unless you'd rather I keep digging here.

PARTH-TUSSLE:

yi-nuo426:
The length guard in mesheryctl/internal/cli/root/perf/perf.go:62 correctly prevents the empty argument from reaching args[0][0]. Please add Args: []string{""} to mesheryctl/internal/cli/root/perf/perf_test.go:19 and assert the returned error contains invalid; this makes the reported panic scenario a permanent regression case.

LYK-02:
Hey @yi-nuo426, I actually tried exactly that — SetArgs([]string{""}) + Execute() doesn't reach the panic line, because Cobra strips empty-string args before dispatching, so it falls through to cmd.Help() with err == nil instead. More details in my earlier comments above. Still open to suggestions on how to test this cleanly, or happy to ship as-is and file a follow-up issue.

PR #21748, opened 2026-09-02
[mesheryctl] Fix index out of range panic in system check and start for portless endpoints

### LINKED ISSUE #21696: [mesheryctl]  System check/start panics with index out of range when a context's endpoint has no port
### Current Behavior
Running `mesheryctl system check` (or `system start`, which calls the same health-check path) panics with `runtime error: index out of range [1] with length 1` if the active context's endpoint has no port — e.g. endpoint is empty (`""`) or just a hostname like `"localhost"` (no `:port`).

The panic happens in `mesheryctl/internal/cli/root/system/check.go`, in `runDockerHealthChecks()`:

```go
endpointParts := strings.Split(hc.context.GetEndpoint(), ":")
...
if endpointParts[1] != "//localhost" {   // panics here if endpointParts has length 1
```

`endpointParts[1]` is accessed in three places in this function without first checking `len(endpointParts) > 1`. Since the config schema (`endpoint` in `config.go`) allows an empty/omitted endpoint, and a context can end up with a portless endpoint through means outside the normal `context create` flow (e.g. manually editing `~/.meshery/config.yaml`), this is reachable in practice, not just theoretically.

### Desired Behavior
`mesheryctl system check`/`system start` should detect a missing or malformed endpoint and return a clear error message (e.g. "context endpoint is invalid, expected host:port") instead of panicking with a raw Go runtime error.

### Screenshots/Logs
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
.../mesheryctl/internal/cli/root/system.(*HealthChecker).runDockerHealthChecks(...)
	.../mesheryctl/internal/cli/root/system/check.go:276

Reproduced with a minimal standalone Go snippet confirming `strings.Split("", ":")` and `strings.Split("localhost", ":")` both yield a length-1 slice, so `[1]` panics.

Related: a very similar unguarded-index pattern exists in `SubscribeToBroker` on the server side (#20434) — different file/component, but worth a cross-reference.

### Environment

- Host OS: macOS 26.6.2
- mesheryctl: built from source (master branch), local dev build — version string isn't embedded outside official releases
- Meshery Server: not required to reproduce — the panic is in mesheryctl's own endpoint-parsing code, confirmed independently with the standalone Go snippet in Screenshots/Logs
- Platform: Docker (via Docker Compose)

---

Gurkaran18:
**Notes for Reviewers**

- This PR fixes #21696

`mesheryctl system check` and `mesheryctl system start` both parsed the current context's `endpoint` by splitting it on `":"` and indexing the result without checking the slice length. Any endpoint without a port — `""`, `localhost`, or a hand-edited `~/.meshery/config.yaml` — yields a length-1 slice, so `[1]` panicked with a raw Go runtime error instead of reporting a bad configuration.

There are **two independent panics**, not one:

| Path | Site | Reached via |
|---|---|---|
| `system check` | `runDockerHealthChecks`, `endpointParts[1]` (2 accesses) | `system check`, and `system start`'s PreRunE health checks |
| `system start` | `resolveDockerEndpoint`, `userPort[1]` | `startDockerDeployment` → `configureDockerServices` |

The second is only *reachable* once the first is fixed, so fixing `check.go` alone leaves `mesheryctl system start` still panicking.

**Changes**

- `check.go`: replaced the index-based `endpointParts[1] != "//localhost"` checks  with `isLocalMesheryEndpoint`, which resolves the host via `net/url` and falls   back to `net.SplitHostPort` for schemeless values.
- `start.go`: added `parseContextEndpoint`, which parses the endpoint once and returns the new `ErrInvalidEndpoint` (`mesheryctl-1255`) when it is empty, has  no port, or is malformed. The resolved address and port are passed through
  instead of the raw `[]string`.

Two latent bugs fall out of parsing correctly:

- the port mapping written to `docker-compose.yaml` used the last `":"` segment, so a portless endpoint silently produced a mapping of `localhost:9081` rather than failing.
- the declined-confirmation branch set `endpoint.Address` to the *scheme* — `"http"` for `http://example.com:9081`. It now receives `http://example.com`.

**Before**

With `endpoint: localhost` in the current context:

```
$ mesheryctl system check --preflight

Docker
--------------
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
...system.(*HealthChecker).runDockerHealthChecks(...)
	/mesheryctl/internal/cli/root/system/check.go:339 +0x590
...system.(*HealthChecker).RunPreflightHealthChecks(...)
	/mesheryctl/internal/cli/root/system/check.go:275 +0x60
```

**After**

```
$ mesheryctl system check --preflight

Docker
--------------
!! Docker is not running
✓ docker-compose is available

Kubernetes API
--------------
✓ can initialize Kubernetes client
!! cannot query the Kubernetes API

Kubernetes Version
--------------
!! cannot check Kubernetes version
!! exit status 1

--------------
--------------
!! Meshery prerequisites not met
```

A genuinely malformed endpoint on the `system start` path now reports `mesheryctl-1255` with a probable cause and remediation instead of panicking.

**Testing**

- New `TestIsLocalMesheryEndpoint` and `TestParseContextEndpoint` cover every  panic-triggering input from the issue (`""`, `"localhost"`, `"http://localhost"`, whitespace-only), each wrapped in `assert.NotPanics`.
- `go build`, `go vet`, `gofmt` clean across `mesheryctl`; full  `go test --short ./...` suite passes.
- Ran the `errorutil` sequence from `.github/workflows/error-codes-updater.yaml`; `analyze` emits 0 bytes, so that gate passes. `next_error_code` bumped to 1256  and `docs/data/errorref/mesheryctl_errors_export.json` regenerated in the same commit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Mesheryctl now validates context endpoints during Docker startup and health checks, reporting clear configuration errors instead of crashing or surfacing Docker failures.
  * Endpoint parsing safely handles localhost, loopback, IPv6, schemes, ports, and surrounding whitespace.
  * Invalid endpoints—including missing, malformed, non-numeric, zero, negative, or out-of-range ports—are rejected consistently.

* **Documentation**
  * Added guidance for resolving invalid endpoint configuration errors.

### DIFF
--- a/mesheryctl/internal/cli/root/system/check.go
+++ b/mesheryctl/internal/cli/root/system/check.go
@@ -19,7 +19,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
+	"net/url"
 	"runtime"
 	"slices"
 	"strings"
@@ -300,17 +302,34 @@ func wrapDockerStartError(err error) error {
 	return ErrDockerStart(errors.Wrapf(err, "failed to start Docker "))
 }
 
+// isLocalMesheryEndpoint reports whether endpoint points at localhost/127.0.0.1.
+//
+// It tolerates endpoints that lack a scheme, a port, or both (e.g. "localhost",
+// "" or a manually edited ~/.meshery/config.yaml) instead of assuming a fixed
+// number of ":"-separated segments, which panicked on any such endpoint (#21696).
+func isLocalMesheryEndpoint(endpoint string) bool {
+	if u, err := url.Parse(endpoint); err == nil && u.Hostname() != "" {
+		return u.Hostname() == "localhost" || u.Hostname() == "127.0.0.1"
+	}
+	host := endpoint
+	if h, _, err := net.SplitHostPort(endpoint); err == nil {
+		host = h
+	}
+	host = strings.TrimPrefix(host, "//")
+	return host == "localhost" || host == "127.0.0.1"
+}
+
 // Run healthchecks to verify if docker is running and active
 func (hc *HealthChecker) runDockerHealthChecks() error {
 	if hc.Options.PrintLogs {
 		utils.Log.Info("\nDocker \n--------------")
 	}
-	endpointParts := strings.Split(hc.context.GetEndpoint(), ":")
+	isLocal := isLocalMesheryEndpoint(hc.context.GetEndpoint())
 
 	// Check whether docker daemon is running using Docker client API
 	dockerCli, err := dockerclient.NewClientWithOpts(dockerclient.FromEnv, dockerclient.WithAPIVersionNegotiation())
 	if err != nil {
-		if endpointParts[1] != "//localhost" {
+		if !isLocal {
 			return wrapDockerContextError(err, hc)
 		}
 		if hc.Options.IsPreRunE { // if this is PreRunExec we trigger self installation
@@ -336,7 +355,7 @@ func (hc *HealthChecker) runDockerHealthChecks() error {
 		// Try to ping the Docker daemon to verify it's actually running
 		_, err = dockerCli.Ping(context.Background())
 		if err != nil {
-			if endpointParts[1] != "//localhost" {
+			if !isLocal {
 				return wrapDockerContextError(err, hc)
 			}
 			if hc.Options.IsPreRunE { // if this is PreRunExec we trigger self installation
--- a/mesheryctl/internal/cli/root/system/check_test.go
+++ b/mesheryctl/internal/cli/root/system/check_test.go
@@ -12,6 +12,35 @@ import (
 
 var update = flag.Bool("update", false, "update golden files")
 
+// TestIsLocalMesheryEndpoint guards against #21696: runDockerHealthChecks used to
+// split the endpoint on ":" and index into the result without checking its length,
+// panicking with "index out of range [1] with length 1" whenever the configured
+// endpoint had no port (e.g. "localhost") or was empty.
+func TestIsLocalMesheryEndpoint(t *testing.T) {
+	tests := []struct {
+		name     string
+		endpoint string
+		want     bool
+	}{
+		{"scheme, host and port", "http://localhost:9081", true},
+		{"scheme and host, no port", "http://localhost", true},
+		{"bare host, no scheme or port", "localhost", true},
+		{"host and port, no scheme", "localhost:9081", true},
+		{"empty endpoint", "", false},
+		{"loopback IP with port", "127.0.0.1:9081", true},
+		{"remote host", "http://example.com:9081", false},
+		{"non-http scheme", "tcp://localhost:2375", true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


hiyach28 (changes requested):
please address all codebrabbit comments

Gurkaran18:
@hiyach28 I have addressed the coderabbit comments and has made changes

hiyach28 (approved):
reviewed the changes and verified the execution paths:

- Both index out of range panics in `system check` and `system start` are resolved.
- Endpoints are properly parsed with `net/url` and `net.SplitHostPort`, catching missing, non-numeric, or out-of-range ports before Docker Compose or health checks run.
- Proactively resolves the latent scheme truncation bug (`userPort[0]`) on the declined localhost prompt.
- Clean regression tests in `check_test.go` and `start_test.go` using the standard `testing` package.
- MeshKit error `mesheryctl-1255` is properly registered and documentation regenerated in the same commit.

LGTM!

ishwar170695 (commented):
Would it make sense to keep the explicit port requirement limited to the Docker startup path? `parseContextEndpoint()` is also used by `runDockerHealthChecks()`, which would reject valid context URLs like `https://meshery.example.com` without an explicit port.

Gurkaran18:
@ishwar170695 ,  The mandatory-port thing in parseContextEndpoint is really only there for configureDockerServices — it needs a real port to build the Docker port mapping. runDockerHealthChecks doesn't actually care about the port, it just needs to bail out early if the endpoint's empty/garbage. Reusing the strict parser there broke valid endpoints like https://meshery.example.com with no explicit port.

Would it make sense to split it up — keep the strict port requirement for the system start path (configureDockerServices/resolveDockerEndpoint), and add a lighter check for runDockerHealthChecks that just catches empty/malformed endpoints without needing a port? I can push that plus a regression test for a portless endpoint on system check if that sounds right to you.

ritzorama:
Excellent replacement of panic-prone parsing with an actionable error contract and coverage. One IPv6 reconstruction case remains in `mesheryctl/internal/cli/root/system/start.go:314-316`: `SplitHostPort` returns `::1` for `http://[::1]:9081`, and concatenating it as `scheme + "://" + host` produces invalid `http://::1`. Preserve brackets for colon-containing hosts and add a `[::1]` regression case; once that and the existing thread are addressed, this looks strong.

Gurkaran18:
@ritzorama Thanks! Both issues are addressed:

- **IPv6:** `parseContextEndpoint` now re-brackets colon-containing hosts before rejoining the scheme (be5ea3d), so `http://[::1]:9081` resolves to `http://[::1]`. `TestParseContextEndpoint` covers `http://[::1]:9081`, `[::1]:9081` and `http://[2001:db8::1]:9081`.
- **Portless endpoints (@ishwar170695's thread):** the parsing is now split (99b7f82). `system start` still requires an explicit port for the Docker mapping, while `runDockerHealthChecks` uses `validateContextEndpoint`, which accepts `https://meshery.example.com`. `TestRunDockerHealthChecksPortlessSchemedEndpoint` is the regression test for it.

Could you take another look?

Issue #21167, opened 2026-08-05
Follow-up: four accepted review items from the Operator/MeshSync/Broker settings PRs (#21144, #21145, #21146)


CodeAhmedJamil:
Four review items were **accepted** across #21144, #21145 and #21146 but **not yet
implemented** — work stopped at a budget boundary with every thread answered and
no fix applied. The thread replies are scattered across three PRs and easy to
miss on a diff, so this is the single place to look.

Two of the four matter; the other two are small.

---

## 1. `#21146` — unparseable override triggers an unwanted MeshSync redeploy ⚠️ **matters**

**File:** `server/handlers/controllers_config_handlers.go` (~line 441)
**Raised by:** CodeRabbit (graded *low*; it is not low)

When `ControllersConfigFromMetadata` fails to parse a connection's stored
override, the code sets `override = nil`. Resolution then yields the server-wide
default, that differs from the previous materialization, and the handler persists
the new value and calls `reconcileMeshsyncDeploymentMode`.

**A connection whose override is merely corrupt is therefore torn down and
redeployed into a deployment mode the user never chose.** That is a silent
destructive action taken on the strength of unreadable data — the same failure
class #21146 exists to remove.

**Change:** treat an unparseable override as *unknown* intent, not *absent*
intent. Skip reconciliation for that connection and surface the parse failure;
the per-connection `GET` already reports it.

---

## 2. `#21144` — rollout wording overstates the restart guarantee ⚠️ **matters**

**File:** `docs/content/en/project/contributing/contributing-controllers-config.md` (~line 86)
**Raised by:** CodeRabbit (*medium*)

The doc says a watch-list change rolls MeshSync pods and *"any other change must
not roll pods."* That is wrong. The env and args the same section documents —
`MESHSYNC_REDACT_SECRETS`, `MESHSYNC_BROKER_CONTENT_DEDUP`, `DEBUG`,
`--outputNamespaces`, `--outputResources` — live in the Deployment pod template,
so changing any of them triggers an ordinary rolling update by definition.

The guarantee that actually holds is narrower: the `meshery.io/restarted-at`
annotation is stamped **only** on a watch-list change, and the previously-applied
value is carried forward otherwise, so the apply itself never rolls pods
gratuitously.

**Change:** distinguish an annotation-triggered restart from an ordinary
pod-template rollout. The implementation is correct; the documentation is not.

---

## 3. `#21146` — document the accepted camelCase metadata alias

[...]

PR #21753, opened 2026-09-02
Document the camelCase meshsyncDeploymentMode read alias

MU5A:
Addresses item 3 of #21167 (originally raised on #21146).

`ui/components/connections/metadata.tsx` and `ui/components/connections/wizard/kubernetesDeploymentMode.tsx` both read `metadata?.meshsyncDeploymentMode ?? metadata?.meshsync_deployment_mode`, accepting a camelCase spelling on top of the snake_case key. The server only ever writes snake_case, and the contributing doc previously documented only that spelling — leaving the alias looking like an undocumented second setting rather than an accepted read-side spelling of the same materialized value.

Checked the other three items from #21167 against current master first: items 1, 2, and 4 are already fixed in code/docs. This PR covers the one that wasn't.

## Test plan
- [x] Confirmed both read sites still contain the camelCase fallback on current master
- [x] Confirmed the doc previously had no mention of the camelCase spelling

## Summary by CodeRabbit

- **Documentation**
  - Clarified deployment mode metadata handling, including supported snake_case and legacy camelCase spellings.
  - Explained that both spellings refer to the same deployment mode value and preserve compatibility with older clients.
  - Documented the distinction between the materialized connection metadata value and the camelCase kubeconfig import setting used to resolve it.


ritzorama:
Thanks for precisely distinguishing the materialized metadata key from the kubeconfig context contract; this prevents a subtle compatibility rule from looking like a second setting.

PR #21754, opened 2026-09-02
chore(deps): bump golang.org/x/crypto to v0.56.0

omlahore:
## The problem

`golang.org/x/crypto` is pinned at `v0.53.0` as an **indirect** requirement. That version carries two SSH denial-of-service advisories, and `govulncheck` finds real call paths to both rather than just flagging the module:

**GO-2026-6355** and **GO-2026-6354**, deadlocked established and undecided channels in `golang.org/x/crypto/ssh`:

```
#1: mesheryctl/pkg/utils/compose.go:119:21
    utils.ComposeClient.Up -> compose.composeService.Up -> ssh.Dial
#2: server/models/default_local_provider.go:2002:7
    models.githubRepoFilterScan -> walker.Git.Walk -> ssh.NewClientConn
```

Both reach code that talks to a remote over SSH, so they are live rather than theoretical.

## Why Dependabot has not already done this

`.github/dependabot.yml` runs a monthly `gomod` job on `/`, and Dependabot proposes updates for **direct** requirements. `golang.org/x/crypto` is indirect here, so it only moves when something that requires it moves, and nothing has. That is why it has drifted three minor versions behind its fixed release.

## The change

```
go get golang.org/x/crypto@v0.56.0
go mod tidy
```

`go mod tidy` carries the rest of the `golang.org/x` family forward as MVS resolves them, which is why `term`, `text`, `tools`, `mod`, `net`, `sync` and `sys` also move. That is not incidental: the `x/text` bump from `v0.38.0` to `v0.41.0` clears a third advisory in the same run.

Only `go.mod` and `go.sum` change. No source files.

## Verification

`go build ./...` is clean.

`govulncheck ./...` before and after:

| | before | after |
|---|---|---|
| Findings your code is affected by | **22** | **19** |
| GO-2026-6355 traces | 2 | 0 |
| GO-2026-6354 traces | 2 | 0 |
| `golang.org/x/text` findings | 1 | 0 |

## What this does not fix

One `x/crypto` finding survives the bump, and it should: **GO-2026-5932**, which says `golang.org/x/crypto/openpgp` is unmaintained and unsafe by design. Its `Fixed in` is `N/A`, so no version clears it. It reaches you through `DownloadMesheryPatternHandler -> HelmConverter.Convert -> armor.Encode` at `server/handlers/meshery_pattern_handler.go:930`. That needs a decision about the OpenPGP dependency itself rather than a version bump, so I have left it alone and am flagging it separately.

The remaining 19 findings sit in `containerd`, `docker/cli`, `docker/compose`, `moby/buildkit`, `moby/go-archive`, `in-toto`, `oras-go` and `otel`. Several have fixed versions available and are indirect for the same reason. Can send those as separate PRs if you want them split that way, since bumping the Docker and containerd family together is a much larger blast radius than this one.

---

## Update: the wasm module needed tidying too

The first push failed `build-meshery / Build image`. `server/policies/wasm` is a **separate module** that does `replace github.com/meshery/meshery => ../../../`, so bumping the root's dependencies left its own `go.mod` pinning the superseded versions. The Dockerfile cross-compiles that module for `GOOS=js GOARCH=wasm`, which then failed with `updates to go.mod needed`.

Fixed in a second commit by running `go mod tidy` inside `server/policies/wasm`. Verified locally:

```
cd server/policies/wasm && GOOS=js GOARCH=wasm go build -trimpath -ldflags="-w -s" -o policy_engine.wasm .
  -> 55 MB artifact produced
go build ./...   (root)
  -> clean
```

It is the only nested module in the repo that replaces the root, so nothing else is affected.

## Summary by CodeRabbit

* **Chores**
  * Updated underlying Go components to newer versions.
  * No user-facing features or behavior changes.


ritzorama:
Thanks for promptly carrying the security remediation through both Go modules and refreshing the compatible transitive `x/*` dependencies. The focused dependency-only scope and passing validation make this straightforward to review.

Issue #21175, opened 2026-08-06
[UI] workspace.ts endpoints are silently discarded by injectEndpoints - expandInfo counts, owner resolution and infinite scroll never run


hamza-mohd:
## Summary

`ui/rtk-query/workspace.ts` injects eight endpoints whose names `@meshery/schemas` already defines. It injects **without** `overrideExisting: true`, so RTK Query silently discards every one of them (dev-only console warning) and the schemas definitions serve all calls.

Three of those local definitions are not simple re-declarations - they carry behaviour the app depends on, and **that behaviour never runs**.

Found during the schema-consumer audit in #PR_PLACEHOLDER, which fixed the three broken workspace CRUD mutations in the same file. The read-side endpoints below were left for a follow-up because restoring them changes what the UI renders and needs a running server to verify.

## Dead local definitions

| endpoint | dead local behaviour |
|---|---|
| `getWorkspaces` | `expandInfo` fan-out that fills `designCount` / `environmentCount` / `viewCount` / `teamCount` |
| `getDesignsOfWorkspace` | `expandUser` owner-profile resolution, `normalizePaginatedCollectionResponse`, infinite-scroll `merge`/`serializeQueryArgs`/`forceRefetch` |
| `getViewsOfWorkspace` | same `expandUser` + infinite-scroll machinery |
| `assignViewToWorkspace`, `unassignViewFromWorkspace`, `getTeamsOfWorkspace`, `assignTeamToWorkspace`, `unassignTeamFromWorkspace` | plain re-declarations, except they target `/api/extensions/api/workspaces/...` while schemas (and the server, `server/router/server.go`) use `/api/workspaces/...` |

Callers that pass the now-inert options:

- `ui/components/workspaces/WorkspaceDataTable.tsx` - `expandInfo: true`
- `ui/components/workspaces/SpacesSwitcher/{RecentContent,MyDesignsContent,SharedContent}.tsx`, `ui/components/dashboard/widgets/MyDesignsWidget.tsx` - `expandUser: true`
- `ui/components/workspaces/SpacesSwitcher/WorkspaceContent.tsx` - `infiniteScroll: true`

## Expected user-visible symptoms

- Workspace cards / table show no design, environment, view or team counts.
- Design and view lists show no owner name (the `[deleted]` / blank fallback) instead of the resolved profile.
- Infinite scroll on the workspace switcher does not accumulate pages.

Worth confirming against a running server before deciding the fix - it is possible some of these render acceptably by other means.

[...]

PR #21755, opened 2026-09-03
fix(ui): release the RTK Query subscription in initiateQuery

omlahore:
## The problem

`initiateQuery` dispatches an RTK Query `initiate()` and awaits the result, but never unsubscribes:

```ts
export const initiateQuery = async (query, variables) => {
  try {
    return await store.dispatch(query.initiate(variables));
  } catch (error) {
    ...
  }
};
```

A manually dispatched `initiate()` returns a `QueryActionCreatorResult`, which holds a cache subscription until `unsubscribe()` is called. Every call here therefore pins its cache entry for the life of the session, and `keepUnusedDataFor` never gets a chance to apply.

There are **13 live call sites** across `rtk-query/user.ts`, `view.ts` and `design.ts`, and several of them run repeatedly rather than once:

```
user.ts:368   getProviderCapabilities
user.ts:373   getAccessToken
user.ts:378   getLoggedInUser
user.ts:383   getSystemVersion
user.ts:388   getAllUsers
view.ts:79    getView
design.ts:190 ...
```

Each caller reads the returned object and drops it, so nothing is relying on the subscription staying open.

## The change

Capture the subscription, await it as before, release it in `finally` so the error path is covered too.

The `unsubscribe` call is optional-chained as well as the value. `query` is untyped in this helper, so a caller passing something that is not an endpoint keeps the current behaviour instead of getting a `TypeError` from the cleanup. The existing tests in `utils.test.ts` mock `dispatch` with `mockResolvedValue`, which returns a plain promise with no `unsubscribe`, so without that guard this change would have broken them.

The repo already uses this pattern in `rtk-query/__tests__/design-import-invalidation.test.ts:143`, which dispatches `initiate()` and then calls `listSub.unsubscribe()` once it is done with the entry.

## Verification

Three tests in `rtk-query/__tests__/utils.test.ts`: the success path, the rejection path, and one pinning that `query.initiate()` is called exactly once, the thunk is dispatched exactly once, the awaited value is returned unchanged, and the single subscription is released.

```
npx vitest run rtk-query/
  Test Files  30 passed (30)
  Tests       301 passed (301)
```

## Not included

`rtk-query/workspace.ts` has six more unreleased `initiate()` dispatches, at lines 105, 114, 121, 128, 174 and 239. I wrote that fix first and then reverted it, because the file's own comment documents those endpoints as dead code under #21175: `injectEndpoints` without `overrideExisting: true` silently discards the colliding names, so the `@meshery/schemas` definitions serve every call and the `expandInfo` and `expandUser` paths never run. Worth fixing whenever #21175 is resolved, but fixing it now would change nothing at runtime.

---

## Addressed review, 2026-09-03

**The `Filters.tsx` reference was wrong and has been removed.** @TheRealVeni is right: `fetchCatalogFilter({...}).subscribe({ next, error })` there is an RxJS-style observable subscription, not a `QueryActionCreatorResult`. `design-import-invalidation.test.ts:143` is the only genuine `initiate()` plus `unsubscribe()` pair in the repo, so that is what the description cites now.

**Coverage added as requested**, in `test(ui): assert initiateQuery dispatches once per call`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved query lifecycle handling by releasing cache subscriptions after successful or failed requests.
  - Preserved existing query results and error responses.
  - Non-RTK Query endpoints continue to behave as before.
  - Cache retention settings can now take effect as expected after queries complete.

- **Tests**
  - Added coverage for subscription cleanup following both successful and rejected queries.
  - Added verification that queries are initiated and dispatched correctly, with results returned unchanged.

### DIFF
--- a/ui/rtk-query/__tests__/utils.test.ts
+++ b/ui/rtk-query/__tests__/utils.test.ts
@@ -30,6 +30,33 @@ describe('initiateQuery', () => {
     expect(result).toEqual({ data: { ok: true } });
   });
 
+  it('releases the cache subscription that initiate() opened', async () => {
+    // The real dispatch returns a QueryActionCreatorResult, which is a promise
+    // that also carries unsubscribe(). Leaving it unsubscribed pins the cache
+    // entry for the life of the session.
+    const unsubscribe = vi.fn();
+    const subscription = Object.assign(Promise.resolve({ data: { ok: true } }), { unsubscribe });
+    dispatch.mockReturnValue(subscription);
+    const initiate = vi.fn().mockReturnValue({ type: 'thunk-action' });
+
+    const result = await initiateQuery({ initiate }, { id: 'abc' });
+
+    expect(result).toEqual({ data: { ok: true } });
+    expect(unsubscribe).toHaveBeenCalledTimes(1);
+  });
+
+  it('releases the subscription even when the query rejects', async () => {
+    const unsubscribe = vi.fn();
+    const subscription = Object.assign(Promise.reject(new Error('boom')), { unsubscribe });
+    dispatch.mockReturnValue(subscription);
+    const initiate = vi.fn().mockReturnValue({ type: 'thunk-action' });
+
+    const result = await initiateQuery({ initiate }, undefined);
+
+    expect(result.isError).toBe(true);
+    expect(unsubscribe).toHaveBeenCalledTimes(1);
+  });
+
   it('returns the awaited dispatch result on success', async () => {
     const fulfilled = {
       data: { foo: 'bar' },
--- a/ui/rtk-query/__tests__/utils.test.ts
+++ b/ui/rtk-query/__tests__/utils.test.ts
@@ -45,6 +45,27 @@ describe('initiateQuery', () => {
     expect(unsubscribe).toHaveBeenCalledTimes(1);
   });
 
+  it('dispatches initiate exactly once and releases that one subscription', async () => {
+    // Guards against a refactor that dispatches twice, or that releases a
+    // different subscription than the one it awaited.
+    const unsubscribe = vi.fn();
+    const fulfilled = { data: { id: 'one' } };
+    const subscription = Object.assign(Promise.resolve(fulfilled), { unsubscribe });
+    dispatch.mockReturnValue(subscription);
+    const thunk = { type: 'thunk-action' };
+    const initiate = vi.fn().mockReturnValue(thunk);
+    const variables = { id: 'one' };
+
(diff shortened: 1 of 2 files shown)


TheRealVeni (changes requested):
The cleanup approach looks correct, and I agree that initiateQuery is being used as a one-shot imperative fetch at these call sites. Before merging, could we add/verify coverage that the helper dispatches query.initiate() exactly once and returns the same fulfilled/rejected result while releasing that single subscription?

Also, the Filters.tsx reference in the PR description appears to be an RxJS subscription rather than an RTK Query QueryActionCreatorResult, so I think that example should be corrected.

I'm fine with keeping the workspace.ts calls out of scope given the endpoint-collision issue described in #21175.

omlahore:
Both fair.

You are right about `Filters.tsx`. That is `fetchCatalogFilter(...).subscribe({ next, error })`, an RxJS-style observable, not a `QueryActionCreatorResult`, so it is not the same pattern and I should not have cited it. I went looking for a real precedent and `rtk-query/__tests__/design-import-invalidation.test.ts:143` is the only one in the repo that dispatches `initiate()` and then calls `unsubscribe()`. The description now cites that instead.

Coverage added in a second commit. It pins one dispatch, one release, same result: `initiate` called once with the variables, the thunk dispatched once, the awaited value returned by identity, and `unsubscribe` called exactly once. That sits alongside the existing success and rejection tests, both of which already assert the single release.

`npx vitest run rtk-query/` is 30 files, 301 tests, all passing.

TheRealVeni:
Awesome! Will review it now. Thanks for the quick response!

Yup, approved from me, thanks!

TheRealVeni (approved):

santusht06 (commented):
LGTM! Great catch on the memory leak in `initiateQuery`, @omlahore.

Calling `subscription.unsubscribe()` inside a `finally` block ensures that RTK Query internal ref counts reach 0 on both fulfillment and rejection without holding dead subscriptions in the cache. The two new Vitest unit tests in `utils.test.ts` verify the unsubscribe lifecycle thoroughly.

omlahore:
Thanks @TheRealVeni. And @santusht06, the ref count reaching 0 on both paths is exactly it, that's why the release is in `finally` rather than after the await.

ritzorama:
Great catch and well-contained fix. Using `finally` ensures the imperative helper releases the exact RTK Query subscription on both fulfillment and failure, and the focused regression tests make the intended lifecycle clear.

PR #21756, opened 2026-09-03
[UI] Fixing partially clipped Save Button in Design Info on small screens. 

### LINKED ISSUE #21732: [UI] Save button is partially clipped in Design Info on small screens
### Current Behavior
In [Meshery Playground -> Designs](https://playground.meshery.io/configuration/designs?view=grid), when opening the Info view for a design on smaller/mobile screens, the Save button extends beyond the right edge of the card/container.
As a result, the button is only partially visible and cannot be accessed properly.

### Expected Behavior
The Save button should be fully contained within the card/container and remain completely visible and accessible on smaller/mobile screens.

### Screenshots/Logs

Image

### Environment

---

cutiepie-10:
**Notes for Reviewers**

- This PR fixes #21732

It took me sometime to setup the system, hence a delayed PR. Steps I did before making this PR:-
1. Replicated the problem by inspect and changing the size of the viewport.
2. Found the specific component that is related to the problem. It was the LegacyInfoModal.

Problem:-
1. The Save button was being clipped.
2. Found out that the content was overflowing the modal footer.

Solution:
1. In the LegacyInfoModal, the ActionContainer was not wrapping properly.
2. Changed the property flexWrap to wrap from no-wrap.
3. Decreased the gap as well as changed overflow to auto.

Screenshots:-
Before the fix:-
save-button-prob

Afer the fix:

For small screens
save-button-fix-sm

For screens with extra small width
save-button-fix-xs

## Summary by CodeRabbit

* **UI Improvements**
  * Added an approval-flow help tooltip to the information modal footer, opened by selecting the information icon.
  * Adjusted footer action spacing for improved responsive layout.

### DIFF
--- a/ui/components/shared/Modal/Information/LegacyInfoModal.tsx
+++ b/ui/components/shared/Modal/Information/LegacyInfoModal.tsx
@@ -26,7 +26,7 @@ import {
   useHasPermission,
   useTheme,
 } from '@sistent/sistent';
-import { Close, Lock, Public } from '@/assets/icons';
+import { Close, Lock, Public, InfoOutlinedIcon } from '@/assets/icons';
 import * as yaml from 'js-yaml';
 import _ from 'lodash';
 import { useSnackbar } from 'notistack';
@@ -326,6 +326,7 @@ const InfoModal_: FC<InfoModalProps> = React.memo((props) => {
   const [imageError, setImageError] = useState(false);
   const version = getDesignVersion(selectedResource);
   const canChangeVisibility = !isPublished && isOwner;
+  const [showModalFooterInfo, setModalFooterInfo] = useState(false);
   const handleError = () => {
     setImageError(true);
   };
@@ -486,13 +487,35 @@ const InfoModal_: FC<InfoModalProps> = React.memo((props) => {
             </Grid>
           </Grid>
         </ModalBody>
-        <ModalFooter
-          helpText={
-            'Upon submitting your catalog item, an approval flow will be initiated. [Learn More](https://docs.meshery.io/concepts/catalog)'
-          }
-          variant="filled"
-        >
-          <ActionContainer>
+        <ModalFooter variant="filled">
+          <Box
+            style={{
+              alignContent: 'center',
+              justifyContent: 'space-between',
+              minWidth: 0,
+            }}
+          >
+            <CustomTooltip
+              title={
+                'Upon submitting your catalog item, an approval flow will be initiated. [Learn More](https://docs.meshery.io/concepts/catalog)'
+              }
+              open={showModalFooterInfo}
+              placement="top"
+              onClose={() => setModalFooterInfo(false)}
+            >
+              <IconButton
+                aria-label="Show catalog approval flow help"
+                onClick={() => setModalFooterInfo(true)}
+              >
+                <InfoOutlinedIcon fill={theme.palette.common.white} style={iconMedium} />
+              </IconButton>
+            </CustomTooltip>
+          </Box>
+          <ActionContainer
+            sx={{
+              gap: { xs: '0.25rem', sm: '0.5rem', md: '1rem' },
+            }}
+          >
             <TooltipButton title={'Copy Design Link'} onClick={handleCopy}>
               <CopyLinkButton>Copy Link</CopyLinkButton>
             </TooltipButton>
--- a/ui/components/shared/Modal/Information/InfoModal.test.tsx
+++ b/ui/components/shared/Modal/Information/InfoModal.test.tsx
@@ -114,6 +114,7 @@ vi.mock('@/assets/icons', () => ({
   Close: () => <svg data-testid="close-icon" />,
   Lock: () => <svg data-testid="lock-icon" />,
   Public: () => <svg data-testid="public-icon" />,
+  InfoOutlinedIcon: () => <svg data-testid="info-outlined-icon" />,
 }));
 
 vi.mock('@/components/meshery-mesh-interface/PatternService/RJSF_wrapper', () => ({
--- a/ui/components/shared/Modal/Information/styles.tsx
+++ b/ui/components/shared/Modal/Information/styles.tsx
@@ -9,7 +9,6 @@ export const CreatAtContainer = styled(Typography)(({ isBold }) => ({
 export const ActionContainer = styled(Box)({
   width: '100%',
   display: 'flex',
-  gap: '1rem',
   justifyContent: 'end',
 });
 


cutiepie-10:
Hi @hiyach28 ,
You may review and tell me any changes that I need to make in this PR.

hiyach28 (changes requested):
Please define width px too for the images. 
1. can you attatch a screen recording with changing width so we can see behaviour on all layouts 
2. in the smaller screen width image, the info button should remain top aligned as in the larger screen

cutiepie-10:
Hi @hiyach28,
Thanks a lot for your feedback!
On searching the screen size I have found out that screen width ranges between 320px to 480px. 
On thinking about the changes you have asked I have three different approaches:-
## Approaches
### Approach 1:-
 Original approach which I have committed, where I am applying flexWrap:wrap for the actionContainer
```
<ActionContainer
  sx={{
    overflow: { xs: 'auto' },
    flexWrap: { xs: 'wrap' },
    gap: { xs: '0.5rem' },
  }}
>
```

In this approach the buttons are stacked on top of each other which is not required. As we can see in this video 


[flexWrap:wrap.webm](https://github.com/user-attachments/assets/8acb0265-2733-4fa9-a01e-9652dad13498)


### Approach 2:- 
Now if I remove the flexWrap and just add the overflow property. The copy link button is clipped away when we reduce the screen size.
```
<ActionContainer
        sx={{
          overflow: { xs: 'auto' },
          gap: { xs: '0.5rem' },
        }}
      >

```

[no-wrap.webm](https://github.com/user-attachments/assets/d42a9363-6986-4a51-9d01-0dc3238f2c0d)

### Approach 3:- 
I can remove the helpText from ModalFooter as it adds a tooltip button for info and add a custom tool tip button to provide the same help. With this approach the when the screen is as small as 320px the info button and the rest of the buttons are visible.

```
<CustomTooltip 
            title={'Upon submitting your catalog item, an approval flow will be initiated. [Learn More](https://docs.meshery.io/concepts/catalog)'}
  open={showModalFooterInfo}
  placement='top'
  onClose={()=>setModalFooterInfo(false)}>
      <IconButton aria-label='HelpOutlineOutlined' onClick={()=>setModalFooterInfo(true)}>
          <HelpOutlineOutlined fill={theme.palette.common.white} style={iconMedium }/>
      </IconButton>
</CustomTooltip>
</Box>
<ActionContainer
  sx={{
    gap: { xs: '0.25rem' },
  }}
>
  
  <TooltipButton title={'Copy Design Link'} onClick={handleCopy}>
    <CopyLinkButton>Copy Link</CopyLinkButton>
  </TooltipButton>
```
[custom-tooltip-button.webm](https://github.com/user-attachments/assets/11498bb5-adae-42b5-9cd4-eaa3524490c5)

## Conclusion

According to me the 3rd approach looks better than the other two.
Can you please tell me which changes should I commit?

hiyach28:
Approach 3 looks good to me, however we would want to keep the tooltip info icon button consistent : `i`
Also please address all coderabbit comments
@Maanvi212006 @Katotodan

cutiepie-10:
Thanks a lot of replying @hiyach28 !
I am done working on the AI conversation 2 and 3 but the first suggestion says to add wrapping to the modalFooter which will let the buttons to stack vertically over each other as shown in my previous message. 

[flexWrap:wrap.webm](https://github.com/user-attachments/assets/088b7914-50ef-4911-9f06-8222551816ab)

This change is not required.
But as the AI mentioned the footer the info button and copy button will have almost no space. It is true at 320px of width which is seen in the older legacy phones like iPhone SE.

I have also added the info-outline-icon instead of the help-outlined-icon. 
Hence the final outcome looks like this:-

[custom-tooltip-button-info.webm](https://github.com/user-attachments/assets/f6743d7c-bc44-4f58-bc31-bfb5fada341a)

Maanvi212006 (commented):
Yes, approach 3 makes sense and it looks way better now!

One thing: CodeRabbit asked for the footer/container to wrap or shrink, but what shipped is a smaller gap + moving the help icon out of the footer. Works for current labels, but nothing shrinks/wraps if a label gets longer later. Worth checking it still holds at 320px with slightly longer text.

Also:
1. Add a line in the PR description on why gap-reduction was chosen over wrap/shrink, so it's documented.
2. Just a consistency note, not a bug: this modal's tooltip is click-to-open, but `DeployStepper`'s help tooltip (`common.tsx`) is hover-based. Two patterns for the same "help icon" concept now exist in the codebase. Maybe a follow-up would be good. Thoughts @Katotodan @hiyach28 ?

cutiepie-10:
Thanks a lot for replying @Maanvi212006 !
I had looked into the Coderabbitai's suggestion before committing. The ModalButtonPrimary and ModalButtonSecondary  are shrinking but not wrapping the text inside though I had removed ``` width:100% ``` from ActionContainer. I tried adding attributes like ``` textOverflow:'ellipsis', whiteSpace:'nowrap' ``` yet these do not work and the modal footer just collapses. Hence creating a buggy layout 320px onwards.
Hence I have reduced the gap between the buttons.

Katotodan:
Agree with @hiyach28 and @Maanvi212006.
Approach 3 looks good, one suggestion, reduce on the padding top and padding bottom on those buttons (only on small devices).
This will make them look more like rectangles than squares.

ritzorama:
Thank you for addressing the small-screen action-bar layout with a targeted change. The labeled help control and responsive action spacing retain the footer functionality while giving the buttons more usable room.

ui/components/shared/Modal/Information/LegacyInfoModal.tsx
```diff
@@ -486,13 +487,35 @@ const InfoModal_: FC<InfoModalProps> = React.memo((props) => {
...
+              minWidth: 0,
+            }}
+          >
+            <CustomTooltip
+              title={
+                'Upon submitting your catalog item, an approval flow will be initiated. [Learn More](https://docs.meshery.io/concepts/catalog)'
+              }
+              open={showModalFooterInfo}
+              placement="top"
+              onClose={() => setModalFooterInfo(false)}
```
hiyach28, line 504: This is a controlled `Tooltip`, but MUI's touch handling still calls `onClose` after `leaveTouchDelay` (1500ms). So on phones, the help text closes about 1.5s after tapping and "Learn More" is hard to reach. On desktop it closes as soon as the pointer leaves the icon. The previous `helpText` used sistent's `HelperTextPopover`, which stays open until you click away. Could you use `HelperTextPopover` from `@sistent/sistent` here, or keep `helpText`? Then the `useState` and the `CustomTooltip`/`IconButton` wrapper can be removed.

ui/components/shared/Modal/Information/LegacyInfoModal.tsx
```diff
@@ -486,13 +487,35 @@ const InfoModal_: FC<InfoModalProps> = React.memo((props) => {
...
+            >
+              <IconButton
```
hiyach28, line 506: The default `IconButton` padding plus `iconMedium` makes this button about 40px wide, while the `HelperTextPopover` it replaces is `size="small"` with 2px padding. So this swap doesn't free footer width; the smaller `ActionContainer` gap does. Could you check whether keeping `helpText` with just the new gap fits at 320px? If it does, the custom tooltip isn't needed.

ui/components/shared/Modal/Information/LegacyInfoModal.tsx
```diff
@@ -486,13 +487,35 @@ const InfoModal_: FC<InfoModalProps> = React.memo((props) => {
...
-            'Upon submitting your catalog item, an approval flow will be initiated. [Learn More](https://docs.meshery.io/concepts/catalog)'
-          }
-          variant="filled"
-        >
-          <ActionContainer>
+        <ModalFooter variant="filled">
+          <Box
+            style={{
+              alignContent: 'center',
+              justifyContent: 'space-between',
```
hiyach28, line 494: Nit: `alignContent` / `justifyContent` have no effect here, since this `Box` isn't `display: 'flex'` and has a single child. Please remove them, or add `display: 'flex'` if you meant them to apply.

hiyach28 (changes requested):
Thanks for the updates, @cutiepie-10. The sized screenshots, the recordings, the info icon and the aria-label fix are all in. I'm fine with the smaller responsive gap instead of wrapping, since the footer no longer wraps. That also covers my earlier top-alignment point.

One issue before merging: the new custom tooltip regresses on touch devices, and sistent already has a component for this (inline below). Also:
- Please update the PR description. It still mentions `flexWrap: wrap` / `overflow: auto`, which aren't in the diff, and @Maanvi212006's request to explain choosing a smaller gap over wrapping is still open.
- Please consider @Katotodan's suggestion to reduce the vertical padding on the buttons on small screens.

PR #21766, opened 2026-09-03
fix: deduplicate embedded design style properties

### LINKED ISSUE #21744: [Bug]: Generated embedded design contains invalid CSS and duplicate style properties
## Current Behavior

The Meshery embedded design export generates several issues in the output JavaScript:

1. The generated CSS contains the misspelled value `inheirt` instead of `inherit`.
2. The generated CSS uses `"transparent"` with quotes instead of the valid CSS keyword `transparent`.
3. The serialized Cytoscape `elementStyles` contains duplicate style properties in both kebab-case and camelCase forms, for example:
   - `background-color` / `backgroundColor`
   - `border-width` / `borderWidth`
   - `text-opacity` / `textOpacity`

These issues are present in the generated export and appear to originate from the upstream design/embed export process rather than manual changes to the generated file.

## Expected Behavior

The embedded design export should:

- Generate valid CSS values such as `inherit` and `transparent`.
- Avoid unnecessary duplicate style properties in the serialized `elementStyles`.
- Produce a clean generated JavaScript file without requiring manual post-processing.

## Steps to Reproduce

1. Create or open a Meshery design in Kanvas.
2. Export the design as an embedded design.
3. Open the generated embedded JavaScript file.
4. Inspect the generated CSS and serialized `elementStyles`.
5. Observe the issues described above.

## Evidence

Image
- Screenshot 1: Shows the generated CSS issues: the `inheirt` typo and the quoted `"transparent"` value. Both are highlighted with red boxes.

Image
- Screenshot 2: Shows duplicate `elementStyles` properties, including kebab-case and camelCase versions, highlighted with red boxes.

## Additional Context

I encountered these issues while working on Meshery PR #21648. I intentionally did not modify the generated JavaScript manually, following maintainer guidance that generated embed JavaScript should be fixed at the source/generator level rather than edited directly.

This issue is being opened so the underlying export/generator behavior can be investigated and fixed.

AdityaOP007:
#21744 

## Summary

This PR fixes duplicate style properties in Meshery's embedded design export.

While investigating issue #21744, I found that some component styles could be serialized in both `kebab-case` and `camelCase`, resulting in duplicate properties in the generated CytoscapeJS `elementStyles` payload.

### What changed

* Updated `ToCytoscapeJS` to remove duplicate `camelCase` style properties when their canonical `kebab-case` equivalents already exist.
* Legitimate standalone `kebab-case` properties are preserved.
* Added a regression test covering properties such as `backgroundColor`, `borderWidth`, and `textOpacity`.
* Kept the change limited to the backend serialization logic.
* No generated JavaScript files were modified.

### Upstream note

The `inheirt` typo and quoted `transparent` CSS values were traced to the upstream `@layer5/meshery-design-embed` package. They are intentionally not patched in this PR, since modifying generated artifacts would only mask the upstream issue.

### Testing

* `git diff --check` ✅
* Regression test added
* Go tests/build: not run locally because the Go toolchain is unavailable in the current environment; these should be validated by CI.

Closes #21744

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cytoscape component data handling for Unicode content.
  * Resolved duplicate style properties by consistently retaining kebab-case keys when equivalent camelCase keys are present.
  * Preserved legitimate style properties and unrelated component values during data conversion.

* **Tests**
  * Added coverage for duplicate style-key resolution and component data conversion.

### DIFF
--- a/server/models/pattern/core/pattern.go
+++ b/server/models/pattern/core/pattern.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"path/filepath"
 	"strings"
+	"unicode"
 
 	"github.com/gofrs/uuid"
 	"github.com/meshery/meshery/server/models/pattern/utils"
@@ -189,13 +190,25 @@ func ToCytoscapeJS(patternFile *pattern.PatternFile, log logger.Handler) (cytosc
 			return cy, err
 		}
 
+		var data interface{} = *cmp
+		cmpByt, err := json.Marshal(cmp)
+		if err == nil {
+			var cmpMap map[string]interface{}
+			if err := json.Unmarshal(cmpByt, &cmpMap); err == nil {
+				if styles, ok := cmpMap["styles"].(map[string]interface{}); ok {
+					resolveDuplicateStyles(styles)
+				}
+				data = cmpMap
+			}
+		}
+
 		elem := cytoscapejs.Element{
 			Data:       elemData,
 			Position:   &elemPosition,
 			Selectable: true,
 			Grabbable:  true,
-			Scratch: map[string]component.ComponentDefinition{
-				"_data": *cmp,
+			Scratch: map[string]interface{}{
+				"_data": data,
 			},
 		}
 
@@ -540,4 +553,30 @@ func getCytoscapeJSPosition(component *component.ComponentDefinition, log logger
 	}
 
 	return pos, nil
+}
+
+func resolveDuplicateStyles(styles map[string]interface{}) {
+	for k := range styles {
+		kebab := toKebabCase(k)
+		if kebab != k {
+			if _, exists := styles[kebab]; exists {
+				delete(styles, k)
+			}
+		}
+	}
+}
+
+func toKebabCase(s string) string {
+	var res string
+	for i, r := range s {
+		if unicode.IsUpper(r) {
+			if i > 0 {
+				res += "-"
+			}
+			res += string(unicode.ToLower(r))
+		} else {
+			res += string(r)
+		}
+	}
+	return res
 }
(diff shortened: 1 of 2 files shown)


santusht06 (commented):
Thanks for addressing the style duplication issue, @AdityaOP007!

A couple of observations from a deep review of this change:

1. **`Scratch` Type Change (`map[string]interface{}` vs `ComponentDefinition`)**:
   In `ToCytoscapeJS`, `elem.Scratch` was previously strongly typed as `map[string]component.ComponentDefinition`, but is now `map[string]interface{}` holding a generic `map[string]interface{}` after `json.Unmarshal`.
   - In `server/models/pattern/core/pattern.go` around line 295 (`processCytoElementsWithPattern`), callers unmarshal via `json.Marshal(&data)` then `json.Unmarshal(..., &declaration)`. While that path serializes from interface, any extension or consumer expecting a concrete `ComponentDefinition` in `_data` may experience assertion issues.
   
2. **Performance Overhead on Large Designs**:
   Marshalling and unmarshalling `cmp` into `map[string]interface{}` per-component inside the loop adds serialization overhead for designs containing hundreds of nodes. Could we perform the kebab-case deduplication directly on `cmp.Styles.AdditionalProperties` before Cytoscape export?

3. **Acronym Casing**:
   `toKebabCase` handles standard camelCase (`backgroundColor` -> `background-color`), but may produce unintended hyphens on acronyms or consecutive capitals (e.g. `URLPath` -> `u-r-l-path`).

The unit test addition is great. Addressing the direct map mutation would avoid the JSON round-trip entirely!

ritzorama:
Thank you for adding focused regression coverage around canonical style serialization. The output-level fix is well scoped; please also resolve the remaining required CI failures.

PR #21776, opened 2026-09-04
[Server] Drop the quadratic kind rescan and per-call regex in mutator identify

omlahore:
There are three identify paths in `server/policies`. f5a32461c optimised two of them and left the third alone, and this finishes that work.

`policy_matchlabels.go:51-55` documents its own rewrite as "Reduces the dominating cost from O(N²·F) to O(N·F)", and `identify_bench_test.go:76-79` documents the binding path going "O(n³) → O(n²)". `identifyRelationshipsBasedOnMatchingMutatorAndMutatedFields` is the third, it is used by `HierarchicalParentChildPolicy`, `EdgeNonBindingPolicy` and `HierarchicalWalletPolicy`, and it had neither a fix nor a benchmark.

## Two compounding defects

**The kind filter was inside the pair loop.** `matchComponentPairs` scanned every component pair and re-tested `componentMatchesKind` on the inner one, so the kind check itself cost O(C²) even when only a handful of components match either kind.

**`matchName` compiled a regex on every call.** It falls through to `regexp.MatchString`, which compiles its pattern each time, and the mismatch path is the common path in that inner loop. So a design paid C² regex compilations per selector pair.

## The change

Bucket by kind once, reusing the existing `filterComponentsByKindTyped` from `policy_binding.go` rather than adding a helper. It appends in slice order, so pair order is unchanged.

For `matchName`, a literal fast path:

```go
if regexp.QuoteMeta(pattern) == pattern {
    return strings.Contains(name, pattern)
}
```

`QuoteMeta` returns its input unchanged and allocation-free when there is nothing to escape, so this is a zero-alloc metacharacter test. An unanchored literal regex is exactly a substring match, which is why this is `strings.Contains` and not `==`: it preserves the existing behaviour where `Pod` matches inside `PodDisruptionBudget`. Patterns that fail to compile still return false.

I considered a compiled-regex cache and rejected it. It measured slower than this, and it adds unbounded shared state for no gain.

## Measured

Fixture is n Deployments and n ConfigMaps, so only half the components match either side. `go1.26.7`, Ryzen 7 7840HS, `-benchtime 50x`.

| components | before | after | |
|---|---|---|---|
| 40 | 704,234 ns / 373 KB / 4,281 allocs | 493,686 ns / 149 KB / 2,284 allocs | 1.4x |
| 200 | 8,097,244 ns / 6.6 MB / 66,889 allocs | 2,846,866 ns / 1.0 MB / 16,854 allocs | 2.8x |
| 800 | 91,653,835 ns / 96.3 MB / 928,672 allocs | 17,589,188 ns / 7.1 MB / 128,121 allocs | 5.2x |

`matchName` on the mismatch path: 2,214 ns and 21 allocs, down to 18.6 ns and 0.

Being precise about what improved: the hoisting does not reduce the number of `matchingMutators` calls, since those were already gated behind the kind checks. The entire win is collapsing C² kind tests to C, plus the regex compiles that went with them. At small component counts the absolute saving is modest; it is the growth curve that matters, and the cost is paid up to `MAX_RE_EVALUATION_DEPTH` times per request by three of the six registered policies.

## Tests

`identify_mutator_bench_test.go` adds the scaling benchmark this path was missing, mirroring `identify_bench_test.go`.

`TestMatchNameEquivalence` pins the fast path against a copy of the original implementation across a corpus built from an alphabet of regex metacharacters, roughly 300k name/pattern pairs, asserting they never disagree. The table in `TestMatchName` also gains the unanchored cases, the empty pattern, invalid patterns like `[` and `a{2,1}`, anchors, alternation, and unicode.

`go test ./policies/ -count=1` passes in 194s, including the Rego-vs-Go conformance and catalog-alignment suites, so this is behaviour-preserving.

203 insertions across 4 files, well under the 500-line gate.

## Not in this PR

`matchName` being unanchored means a selector for kind `Pod` also matches `PodDisruptionBudget`, and `ConfigMap` matches `ConfigMapList`. This PR deliberately preserves that so it stays behaviour-preserving. Whether it should be anchored is a semantics question for a maintainer, and I would rather raise it separately than bundle it here.

Note `server/policies` is not currently `gofmt` clean on master (15 files), so I have left formatting alone rather than bury this in a whitespace diff. My changed lines are clean.

## Summary by CodeRabbit

* **Performance**
  * Improved policy evaluation performance when matching component pairs.
  * Accelerated matching for literal names while preserving existing wildcard and regular-expression behavior.

* **Reliability**
  * Expanded validation across empty patterns, invalid expressions, special characters, and Unicode names.
  * Added benchmarks covering relationship identification and component matching across varying workloads.

### DIFF
--- a/server/policies/eval_rules.go
+++ b/server/policies/eval_rules.go
@@ -410,15 +410,16 @@ func matchComponentPairs(
 	fromKind := selectorItemKind(fromSel)
 	toKind := selectorItemKind(toSel)
 
-	for _, compFrom := range comps {
-		if !componentMatchesKind(compFrom, fromKind) {
-			continue
-		}
-		for _, compTo := range comps {
+	// Bucket by kind once instead of re-testing the kind of every candidate
+	// inside the inner loop, which made the kind check itself O(C^2).
+	// filterComponentsByKindTyped appends in slice order, so pair order is
+	// unchanged.
+	fromComps := filterComponentsByKindTyped(comps, fromKind)
+	toComps := filterComponentsByKindTyped(comps, toKind)
+
+	for _, compFrom := range fromComps {
+		for _, compTo := range toComps {
 			if compFrom.ID == compTo.ID {
-				continue
-			}
-			if !componentMatchesKind(compTo, toKind) {
 				continue
 			}
 			if !matchingMutators(compFrom, compTo, fromSel, toSel, design) {
--- a/server/policies/identify_mutator_bench_test.go
+++ b/server/policies/identify_mutator_bench_test.go
@@ -1,0 +1,130 @@
+package policies
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/meshery/schemas/models/v1beta1/component"
+	"github.com/meshery/schemas/models/v1beta1/pattern"
+	"github.com/meshery/schemas/models/v1beta2/relationship"
+)
+
+// buildMutatorBenchFixture returns n Deployments and n ConfigMaps, so only half
+// the components match either side of the selector pair. That is the shape the
+// pre-fix code paid most for: the kind of every component was re-tested inside
+// the inner loop, making the kind check itself O(C²) even though the eventual
+// pair set is much smaller.
+//
+// Each Deployment's mutator path holds a ConfigMap name, and one ConfigMap
+// carries the matching value, so matchingMutators does real work rather than
+// short-circuiting on the first ref.
+//
+// Pre-fix work: O(C²) kind tests, each one a regexp compile on the mismatch path.
+// Post-fix work: O(C) kind tests, and the pair enumeration is unchanged.
+func buildMutatorBenchFixture(n int) (*pattern.PatternFile, *relationship.RelationshipDefinition, relationship.SelectorItem, relationship.SelectorItem) {
+	comps := make([]*component.ComponentDefinition, 0, 2*n)
+	for i := 0; i < n; i++ {
+		d := &component.ComponentDefinition{
+			Component: component.Component{Kind: "Deployment"},
+			Configuration: map[string]interface{}{
+				"spec": map[string]interface{}{
+					"configMapName": fmt.Sprintf("cm-%d", i%(n/4+1)),
+				},
+			},
+		}
+		d.ID = staticUUID(fmt.Sprintf("deploy-%d", i))
+		comps = append(comps, d)
+
+		c := &component.ComponentDefinition{
+			Component: component.Component{Kind: "ConfigMap"},
+			Configuration: map[string]interface{}{
+				"metadata": map[string]interface{}{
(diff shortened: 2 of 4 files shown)


kunalKumar-13 (commented):
This has been open since 4 September with no feedback, which is a shame because both changes hold up. I checked the two that could have quietly changed behaviour.

**The `matchName` fast path is exactly equivalent.** The claim that an unanchored literal regex is a substring match is the load-bearing one, so I ran both versions over a grid of names against patterns — literals, wildcards, anchors, alternation, character classes, `\d+`, and an unbalanced `(`:

```
13 names × 15 patterns = 195 combinations, 0 differences
```

Worth noting the invalid-pattern case survives too: `QuoteMeta("(")` is `\(`, which is not equal to `"("`, so it falls through to `regexp.MatchString`, which errors, and `err == nil && matched` gives false — same as before. The guard is doing real work there rather than accidentally swallowing a bad pattern.

Your comment about `"Pod"` still matching inside `"PodDisruptionBudget"` is the important line in the whole PR, incidentally. That is unintuitive behaviour to preserve deliberately, and someone reading `strings.Contains` cold would reasonably assume it was a tightening.

**The bucketing preserves pair order and the predicate is identical.** Two things I wanted to confirm before believing "pair order is unchanged":

- `filterComponentsByKindTyped` (`policy_binding.go:389`) is a plain range-and-append, so slice order survives. It also already existed — reusing it rather than adding a second filter is the right call.
- It tests with `matchName(comp.Component.Kind, kind)`, and the inline check it replaces was `componentMatchesKind`, which is a one-line wrapper around exactly that (`eval_rules.go:368`). Same predicate, so the same components are selected.

The one reordering is that the `compFrom.ID == compTo.ID` check now happens after the kind filter rather than interleaved with it. Both are pure predicates and both must pass, so the result set is unchanged — just worth saying out loud since the diff makes it look like a reordering of logic rather than of loops.

**On the framing** — pointing at `policy_matchlabels.go:51-55` and `identify_bench_test.go:76-79` as the two paths already fixed, and this being the third, makes the case much better than a bare "this is O(n²)". It is a stated intent that was left half-finished rather than a new opinion about performance.

Adding `identify_mutator_bench_test.go` alongside is the right move too — the other two paths have benchmarks documenting their complexity, and this one had none, which is presumably part of why it got skipped.

LGTM.

omlahore:
Hey, thanks. The substring thing was one of the claims in here I wasn't fully sure of, so I appreciate you running it.

And I guess you're right about the ID check. Sorry, I should have put that in the description instead of letting people find it in the diff. I'm sort of a beginner myself so things like that slip past me.

ritzorama:
Thank you for completing this optimization path and documenting it with both equivalence coverage and benchmarks. The localized change makes the performance improvement easy to validate and maintain.

Issue #21265, opened 2026-08-10
Ensure latest accepted connection lifecycle transition wins


SuryanshGarg04:
## Problem

Connection lifecycle work initiated by an earlier `Connect` can continue after a newer `Disconnect` or `Delete` transition has been accepted.

The use of `context.WithoutCancel()` is intentional because connection work should not be tied to the lifetime of the HTTP request. However, there is no connection-lifecycle owner or invalidation mechanism that makes work belonging to an earlier lifecycle obsolete when a newer lifecycle transition is accepted.

As a result, asynchronous Connect work, controller configuration, MeshSync setup, or reconciliation can overlap with Disconnect/Delete cleanup and subsequently mutate runtime state after the connection has already transitioned.

### Example

A concrete failure sequence is the MeshSync data-handler path:

1. `Connect` observes that a MeshSync data handler does not exist.
2. `Connect` creates and starts the handler.
3. Before the handler is published to `MesheryControllersHelper`, `Disconnect` is accepted.
4. `Disconnect` performs its cleanup.
5. The earlier `Connect` operation later publishes the handler.

This can leave runtime resources inconsistent with the connection's current lifecycle state.

## Root Cause

There are two related lifecycle problems:

1. **No lifecycle-level ownership or invalidation**
   - Background work started by `Connect` is not tied to a lifecycle generation.
   - A newer `Disconnect` or `Delete` does not invalidate the older work.

2. **FSM event dispatch does not preserve accepted request ordering**
   - FSM events are dispatched in goroutines.
   - The FSM mutex serializes whichever event acquires it first, but does not preserve the ordering of accepted lifecycle requests.

Therefore, a sequence such as:

```text
Connect → Disconnect
```

can potentially be processed as:

```text
Disconnect → Connect
```

allowing the older `Connect` event to restore the persisted state to `CONNECTED`.

## Invariant

For a given connection:

**The latest accepted lifecycle transition must win.**

Once a newer lifecycle transition such as `Disconnect` or `Delete` is accepted:
- work belonging to an earlier lifecycle must become stale;
- stale work must not subsequently mutate the connection lifecycle;
- stale work must not recreate resources belonging to the previous lifecycle.

## Scope

This issue focuses on connection lifecycle ordering and stale asynchronous work.

The following related findings remain separate follow-up issues:
- Initialization TOCTOU
- Deletion-path differences
- Shared mutable controller state

The existing `CONNECTED` boundary must be preserved.

[...]

PR #21817, opened 2026-09-05
[Server] Guard MesheryControllersHelper state with an RWMutex

SuryanshGarg04:
ctxControllerHandlers, ctxOperatorStatus and ctxMeshsyncDataHandler are written by connection lifecycle side effects (connect, disconnect, delete, MeshSync mode changes) and read concurrently by HTTP handlers and the background status polling. Protect them with an RWMutex, take the handler out of the helper under the lock before stopping it so readers are not blocked behind Stop(), and return early from AddMeshsyncDataHandlers when no database handler is configured instead of dereferencing nil.

This does not change the check-then-act in AddMeshsyncDataHandlers, which remains a separate follow-up.

Part of #21265.

**Notes for Reviewers**

- This PR fixes #

### DIFF
--- a/server/models/meshery_controllers.go
+++ b/server/models/meshery_controllers.go
@@ -70,6 +70,10 @@ type MesheryControllersHelper struct {
 	// needs a context identifier takes one from its caller (the K8sContext is
 	// available at every call site) rather than reading this field.
 	contextID string
+	// mu protects ctxControllerHandlers, ctxOperatorStatus, and ctxMeshsyncDataHandler from concurrent reads and writes
+	// across the connection lifecycle side-effects and background polling HTTP handlers.
+	mu sync.RWMutex
+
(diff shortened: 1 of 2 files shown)


ritzorama:
Thanks for tackling this race-prone lifecycle path. I inspected the new RWMutex coverage, MeshSync add/remove flow, status polling call sites, and concurrency tests. The accessor locking is a solid improvement, but please serialize the full add/remove lifecycle: `AddMeshsyncDataHandlers` currently unlocks after observing a nil handler at `meshery_controllers.go:328-333`, before startup and assignment at `:379-398`, so concurrent reconcilers can create duplicate handlers. Also move the `dbHandler == nil` guard before embedded startup; the current guard returns after a `stopFunc` may have been created and leaks the started MeshSync run.

PR #21819, opened 2026-09-05
Split/21365/3 lifecycle generation ctx

SuryanshGarg04:
**Notes for Reviewers**

- This PR fixes #


ritzorama:
I inspected the lifecycle-context rotation/rollback logic, state-machine tracker behavior, and controller-helper synchronization in this draft. One concurrency issue remains in `AddMeshsyncDataHandlers`: its read-check/create/write sequence at `meshery_controllers.go:328-395` is not atomic, so concurrent callers can each start a handler and leak the losing pipeline. Use a dedicated initialization mutex (or hold an exclusive lock) across the nil check, construction, startup, and assignment. Thanks for adding focused lifecycle-generation coverage while this series is still in draft.

PR #21825, opened 2026-09-05
[UI] Surface administrative environment purpose badge and restrict edit/delete affordances

NiftyOracle:
Added tests for rendering and interaction of Administrative environments in EnvironmentCard component.

**Notes for Reviewers**

- This PR fixes #

**[Signed commits](https://github.com/meshery/meshery/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

## Summary by CodeRabbit

- **Tests**
  - Expanded coverage for environment cards, including administrative environment badges.
  - Verified that editing and deletion are disabled for administrative environments, with explanatory tooltips shown.
  - Confirmed that edit and delete actions remain available for non-administrative environments.

Issue #21803, opened 2026-09-04
[UI] Surface the new Environment purpose designation, and offer no control that sets it


pontusringblom:
### Current Behavior

`meshery/schemas` is adding a first-class `purpose` property to the Environment construct (`v1beta3`). It designates what an environment exists for:

| Value | Meaning |
| --- | --- |
| `user` | An ordinary environment. People create these to logically group Connections and their Credentials. |
| `administrative` | An environment the platform provisions to hold organization-level configuration. Resolvers of that configuration trust it. |
| absent | Identical to `user`. Every environment that predates the property is in this state. |

**The property is not client-settable.** It is absent from `EnvironmentPayload` - the schema every `POST`/`PUT` `requestBody` references - and absent from the RJSF create-or-edit form schema. Permission to create an environment must not confer the ability to make one administrative.

The environments UI shows every environment identically, so an administrative environment is indistinguishable from an ordinary one. Someone can open the ordinary edit and delete affordances on an environment the platform depends on without any indication that it is different.

Full contract: `docs/environment-purpose-contract.md` in `meshery/schemas`.

### Desired Behavior

1. **Surface the designation.** Show administrative environments as such in `ui/components/environments/` - the list and `environment-card.tsx` - so it is visible before someone acts on one. Use the Sistent design system rather than a bespoke badge.

2. **Do not offer a control that sets it.** `ui/components/environments/index.tsx` already drives the create-or-edit modal from the canonical RJSF schema (`createAndEditEnvironmentSchema` out of `@meshery/schemas`), which does not carry `purpose` - so nothing needs adding to keep it off the form, and nothing may be added. No hand-authored form field, no toggle, no override in `initialData`.

3. **Read the value with a positive test.** Compare against the literal `"administrative"`. Do not test for "not `user`": the property is optional on the wire and absent for every environment that predates it, so a negative test renders those as administrative.

4. **Consider what the ordinary affordances should do** on an administrative environment - editing its name, deleting it, reassigning its connections. Whatever the server ends up allowing, the UI should not present an action that will be refused as though it will succeed. Worth deciding alongside the server behaviour rather than independently.

5. Cover the new rendering in `environment-card.test.tsx` / `index.test.tsx`, including an environment with the property absent.

### Additional context

Blocked on the `@meshery/schemas` release that carries `Environment.purpose`. Companion issues cover Meshery Server and mesheryctl.

NiftyOracle:
[...]

PR #21828, opened 2026-09-05
[UI] Surface administrative environment purpose badge and restrict edit/delete affordances

NiftyOracle:
…dit/delete affordances

Added purpose prop to environmentDetails and updated administrative checks for editing and deleting environments.

**Notes for Reviewers**

- This PR fixes #

## Summary by CodeRabbit

- **New Features**
  - Administrative environments are now clearly labeled with an “Administrative” badge.
  - Administrative environments cannot be edited or deleted.
  - Disabled actions continue to provide explanatory tooltips.
  - Updated environment card layouts improve title and control alignment.

### DIFF
--- a/ui/components/environments/environment-card.tsx
+++ b/ui/components/environments/environment-card.tsx
@@ -11,6 +11,7 @@ import {
   useTheme,
   IconButton,
   CustomTooltip,
+  Chip,
 } from '@sistent/sistent';
 import { iconMedium } from '../../css/icons.styles';
 
@@ -63,6 +64,10 @@ export const TransferButton = ({ title, count, onAssign, permissionKey }) => {
  * @param {Object} props.environmentDetails - The details of the environment.
  * @param {string} props.environmentDetails.name - The name of the environment.
  * @param {string} props.environmentDetails.description - The description of the environment.
+ * @param {string} [props.environmentDetails.purpose] - The purpose of the environment.
+ *   When equal to "administrative", the environment is platform-provisioned and shown
+ *   with an "Administrative" badge. Edit and delete affordances are disabled for such
+ *   environments because actions on them are likely to be refused by the server.
  * @param {Function} props.onDelete - Function to delete the environment.
  * @param {Function} props.onEdit - Function to edit the environment.
  * @param {Function} props.onSelect - Function to select environment for bulk actions.
@@ -88,6 +93,11 @@ const EnvironmentCard = ({
     { skip: !environmentDetails.id },
   );
   const environmentConnectionsCount = environmentConnections?.totalCount || 0;
+
+  // Compare against the literal "administrative".
+  // Do NOT test for "not user": the property is optional and absent for every
+  // environment that predates it, so a negative test would render those as administrative.
+  const isAdministrative = environmentDetails?.purpose === 'administrative';
 
   // this allows to handle both cases when deleted at is:
   // - timestamp or null
@@ -121,10 +131,25 @@ const EnvironmentCard = ({
               borderRadius: 2,
             }}
           >
-            <Grid2 sx={{ display: 'flex', flexDirection: 'row', pb: 1 }}>
+            <Grid2 sx={{ display: 'flex', flexDirection: 'row', pb: 1, alignItems: 'center' }}>
               <Name variant="body2" onClick={(e) => e.stopPropagation()}>
                 {environmentDetails?.name}
               </Name>
+              {isAdministrative && (
+                <Chip
+                  label="Administrative"
+                  size="small"
+                  sx={{
+                    ml: 1,
+                    backgroundColor: 'rgba(0, 179, 159, 0.12)',
+                    color: '#00B39F',
+                    border: '1px solid #00B39F',
+                    fontWeight: 600,
+                    fontSize: '0.7rem',
+                    height: '22px',
+                  }}
+                />
+              )}
             </Grid2>
             <Grid2
               sx={{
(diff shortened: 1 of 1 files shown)


NiftyOracle:
This PR addresses #21803. Changes include:
Added Administrative badge using Sistent Chip on front and back card faces
Positive test against literal 'administrative' as specified
Edit/delete disabled for administrative environments with descriptive tooltips
Tests covering badge rendering, absent purpose, and disabled affordances

ui/components/environments/environment-card.tsx
```diff
@@ -89,6 +94,11 @@ const EnvironmentCard = ({
   );
   const environmentConnectionsCount = environmentConnections?.totalCount || 0;
 
+  // Compare against the literal "administrative".
+  // Do NOT test for "not user": the property is optional and absent for every
+  // environment that predates it, so a negative test would render those as administrative.
+  const isAdministrative = environmentDetails?.purpose === 'administrative';
```
hortison, line 100: This is always `false` today. `environmentDetails.purpose` doesn't exist on the `Environment` type - I checked the pinned `meshery/schemas` v1.3.37 struct (`models/v1beta3/environment/environment.go`) and grepped the server tree; there's no `purpose`/`administrative` concept anywhere in the backend or in the schema this type is generated from. Until the server actually returns this field, the badge and the disabled edit/delete state below are unreachable for every environment.

leecalcote:
Is this a duplicate PR, @NiftyOracle?

ui/components/environments/environment-card.tsx
```diff
@@ -63,6 +64,10 @@ export const TransferButton = ({ title, count, onAssign, permissionKey }) => {
  * @param {Object} props.environmentDetails - The details of the environment.
  * @param {string} props.environmentDetails.name - The name of the environment.
  * @param {string} props.environmentDetails.description - The description of the environment.
+ * @param {string} [props.environmentDetails.purpose] - The purpose of the environment.
```
hortison, line 67: This documents `purpose` as a real response field, but it isn't defined in `meshery/schemas` yet. Per this repo's AGENTS.md, any new or changed API response field must be added to schemas first and consumed via the generated client - otherwise there's no shared source of truth pinning the literal `'administrative'` value used below, and the two can drift in spelling or casing.

ui/components/environments/environment-card.tsx
```diff
@@ -121,10 +131,25 @@ const EnvironmentCard = ({
               borderRadius: 2,
             }}
           >
-            <Grid2 sx={{ display: 'flex', flexDirection: 'row', pb: 1 }}>
+            <Grid2 sx={{ display: 'flex', flexDirection: 'row', pb: 1, alignItems: 'center' }}>
               <Name variant="body2" onClick={(e) => e.stopPropagation()}>
                 {environmentDetails?.name}
               </Name>
+              {isAdministrative && (
+                <Chip
```
hortison, line 139: This `sx` block (fontWeight, fontSize, height) is duplicated verbatim on the second `Chip` further down. `styles.tsx` in this directory defines every label as a named styled component - worth factoring this into one so the front and back badges don't drift apart on a future sizing tweak.

ui/components/environments/environment-card.tsx
```diff
@@ -121,10 +131,25 @@ const EnvironmentCard = ({
...
+                  label="Administrative"
+                  size="small"
+                  sx={{
+                    ml: 1,
+                    backgroundColor: 'rgba(0, 179, 159, 0.12)',
+                    color: '#00B39F',
```
hortison, line 145: Hardcoded hex/rgba here instead of `theme.palette` tokens, unlike the rest of this file's `styles.tsx` (which consistently derives colors from `theme.palette`). A future theme or rebrand change won't touch this badge, so it can end up clashing with the rest of the card.

hortison (changes requested):
Note: #21829, opened shortly after this one with an equivalent diff to this same file, has been closed as a duplicate - continuing review here.

Inline notes below cover the styling issues. Two things worth calling out that don't map to a diff line:

**The feature can't do anything yet.** `isAdministrative` is derived from `environmentDetails.purpose`, which doesn't exist anywhere in the `Environment` type. I checked the pinned `meshery/schemas` v1.3.37 `Environment` struct (`models/v1beta3/environment/environment.go`) and grepped the whole server tree - there's no "purpose" or "administrative" concept in the backend or in the schema the UI type is generated from. So `isAdministrative` is always `false` for every real environment: the badge never renders and edit/delete are never disabled by this logic.

**Test gap left for the next contributor.** `environment-card.test.tsx`'s `@sistent/sistent` mock doesn't export `Chip`. This PR doesn't add a test with `purpose: 'administrative'`, so this is silently masked today. Whoever eventually wires up a real `purpose` value and writes a test that exercises the `isAdministrative` branch will hit `Chip` resolving to `undefined` in the mock - an "Element type is invalid" crash instead of a passing assertion.

To move this forward: add `purpose` to `meshery/schemas` first (see this repo's AGENTS.md on the schema-first requirement), regenerate the client, then wire this component up to the generated type instead of an ad hoc field.

NiftyOracle:
Hi @leecalcote! Thank you for reviewing. After checking the existing issues and PRs, I don't believe this is a duplicate. Issue #21803 specifically introduces the new purpose property from meshery/schemas and requires surfacing the administrative designation in the UI — which is distinct from any general environment UI work. Happy to clarify or adjust if I've misunderstood the scope!

yi-nuo426:
The explicit purpose === administrative guard correctly avoids treating older environments with no purpose field as administrative, while consistently disabling both edit and delete controls for the protected case. No changes requested.

leecalcote:
> Hi @leecalcote! Thank you for reviewing. After checking the existing issues and PRs, I don't believe this is a duplicate. Issue #21803 specifically introduces the new purpose property from meshery/schemas and requires surfacing the administrative designation in the UI — which is distinct from any general environment UI work. Happy to clarify or adjust if I've misunderstood the scope!

This is a duplicate. #21825

Issue #20445, opened 2026-07-03
[Server] Add SafeGo and recovery middleware to prevent background panic crashes


mightbeanshuu:
### Current Behavior

Meshery Server starts many background goroutines directly with `go ...` across `server/`. A panic in any goroutine that is not recovered inside that same goroutine terminates the whole `meshery-server` process.

This has already shown up as a recurring bug class:

- #20391: event broadcaster can panic the server.
- #18775: nil-pointer dereference in a connection-status goroutine crashes the server.
- #20434 / #20435: `SubscribeToBroker` panics on missing or portless broker endpoint.
- #20441 / #20442: control-plane subscription goroutine panics on a container-less pod spec.

Those issues fix individual call sites, but the root exposure remains. A quick scan excluding tests and generated GraphQL code shows roughly 212 `go` launches in `server/`; about 89 of those are not direct `Publish` calls. The router currently installs OpenTelemetry middleware, but no recovery middleware:

```go
gMux.Use(otelmux.Middleware("meshery-server"))
```

The codebase already has a localized version of this defensive pattern in `runRelationshipEvaluation`, which wraps a spawned evaluation goroutine with `recover()` and stack logging so one relationship-evaluation panic becomes a request error instead of a process outage.

### Desired Behavior

Meshery should treat background goroutine panics as isolated task failures, not process-wide outages. HTTP handler panics should be converted to structured `500` responses with server-side logs, and newly introduced background goroutines should be easy to spot during review.

### Proposal

Add defense in depth in three reviewable stages:

1. Add a `SafeGo` helper, for example `utils.SafeGo(ctx, name string, log logger.Handler, fn func())`, that launches `fn` in a goroutine with `defer recover()`, structured logging, and stack capture.
2. Add a gorilla/mux recovery middleware for HTTP handler paths, returning a structured `500` while logging the panic and stack server-side.
3. Add a small lint or CI guard that flags new bare `go` statements under `server/` unless explicitly exempted, so the pattern does not regress.

### Why This Is Worth Doing

- It turns one malformed cluster object, broker event, or edge-case input into one logged task failure instead of a full server crash.
- It consolidates a repeated class of accepted panic bugs instead of fixing the next unguarded index, type assertion, nil dereference, or send-on-closed-channel one at a time.
- The migration can be split by package: add helper and middleware first, then migrate high-risk goroutine sites such as broker subscriptions, MeshSync/control-plane listeners, FSM transitions, and long-running evaluation or import flows.

[...]

PR #21833, opened 2026-09-06
[Server] Decode environment connections filter once, fixing the unreachable owner filter

### LINKED ISSUE #21826: [Server] Conflicting filter formats in GetEnvironmentConnections make the owner/organization_id filters unreachable and panic on a non-boolean assigned
<html><head></head><body><h2>Current Behavior</h2>
<p><code>GetEnvironmentConnections</code> (<code>server/models/environment_persister.go</code>) feeds a single <code>filter</code> query parameter to two parsers with incompatible grammars:</p>
<pre><code class="language-go">// line 245 — parser 1: filter must be a JSON object
err := json.Unmarshal([]byte(filter), &amp;filterMap)
if err != nil {
    return nil, err
}
if assignedVal, ok := filterMap["assigned"]; ok {
    isAssigned = assignedVal.(bool)          // line 251 — unchecked
}
...
// line 276 — parser 2: filter must be "key value"
dynamicKeys := []string{"owner", "organization_id"}
query = utils.ApplyFilters(query, filter, dynamicKeys)
</code></pre>
<p><code>ApplyFilters</code> (<code>server/helpers/utils/utils.go:378</code>) splits on a space and reads <code>filterArr[0]</code> as the column. The two grammars are mutually exclusive, so each disables the other:</p>
<ul>
<li><strong>JSON input</strong> (<code>{"assigned":false}</code> — what Meshery UI sends) arrives at <code>ApplyFilters</code> as one token with no space, so <code>filterKey</code> is the whole JSON blob. It matches neither <code>deleted_at</code> nor any <code>dynamicKeys</code> entry, and is a silent no-op.</li>
<li><strong><code>ApplyFilters</code> input</strong> (<code>owner abc</code>, <code>organization_id x</code>, <code>deleted_at Deleted</code>) fails <code>json.Unmarshal</code> on line 245 first, and the raw parse error is returned to the client.</li>
</ul>
<p><strong>The <code>owner</code> and <code>organization_id</code> filters declared on line 276 are therefore unreachable.</strong> No value of <code>filter</code> can invoke them: the JSON parse rejects every input that would reach them, and every input that survives the JSON parse is inert by the time it arrives.</p>
<p><code>GetEnvironments</code> in the same file (lines 49–50) declares the identical <code>dynamicKeys := []string{"owner", "organization_id"}</code> and calls <code>ApplyFilters</code>, but does not JSON-parse <code>filter</code>. So <code>GET /api/environments?filter=owner abc</code> filters correctly while <code>GET /api/environments/{id}/connections?filter=owner abc</code> returns a parse error — the same parameter name and the same declared keys behave oppositely on sibling endpoints, which suggests the JSON parse was added to this one function without noticing the existing <code>ApplyFilters</code> contract.</p>
<p>Separately, line 251 asserts <code>assignedVal.(bool)</code> without the comma-ok idiom, so a non-boolean <code>assigned</code> panics the handler goroutine. There is no panic-recovery middleware on the HTTP path (see #20445), so the request is aborted with no response and a stack trace is logged. <code>net/http</code> recovers per-connection, so the process itself survives — this aborts the request, it does not crash the server.</p>
<p><code>filter</code> is declared in <code>meshery/schemas</code> (<code>constructs/v1beta1/environment/api.yml</code>, operation <code>getEnvironmentConnections</code>) as an optional query parameter of type <code>string</code> with no description — an untyped string with no documented grammar, which is how two grammars came to coexist behind it. It is taken verbatim from <code>q.Get("filter")</code> at <code>server/handlers/environments_handlers.go:244</code> and passed straight through <code>DefaultLocalProvider.GetConnectionsOfEnvironment</code>, so any authenticated client controls it fully.</p>
<p><strong>Scope:</strong> Meshery UI only ever sends <code>'{"assigned":false}'</code> (<code>ui/components/environments/index.tsx:134</code>), a valid boolean, so neither defect affects the first-party UI today. This is an API-contract bug affecting <code>mesheryctl</code> and third-party REST consumers.</p>
<h2>Expected Behavior</h2>
<p><code>filter</code> should have one documented grammar on this endpoint. Either:</p>
<ol>
<li>The JSON object is the contract, and the <code>ApplyFilters</code> call with its <code>owner</code>/<code>organization_id</code> keys is removed or reworked to read them out of the parsed <code>filterMap</code>; or</li>
<li><code>ApplyFilters</code>' <code>key value</code> form is the contract, and <code>assigned</code> is expressed within it.</li>
</ol>
<p>Either way:</p>
<ul>
<li>A malformed value should return a structured <code>400</code>, not a raw <code>encoding/json</code> error and not a panic.</li>
<li>Line 251 should use the comma-ok idiom.</li>
<li>Ideally, the chosen grammar is documented on the <code>filter</code> parameter in <code>meshery/schemas</code> so the contract is visible to generated clients.</li>
</ul>
<h2>Screenshots/Logs</h2>
<p>Calling the exported API directly (<code>EnvironmentPersister</code> with a nil DB, so the panic fires before any DB access on line 258):</p>
<pre><code>{"assigned":false}      PANIC: nil pointer deref   &lt;- control: reached the DB line
{"assigned":"false"}    PANIC: interface conversion: interface {} is string, not bool
{"assigned":0}          PANIC: interface conversion: interface {} is float64, not bool
{"assigned":null}       PANIC: interface conversion: interface {} is nil, not bool
deleted_at Deleted      returned err=invalid character 'd' looking for beginning of value
</code></pre>
<p>The first line is the control case: a valid boolean gets past line 251 and reaches the DB. The last line is the <code>ApplyFilters</code> grammar being rejected by the JSON parser.</p>
<p>The same rejection for each documented <code>ApplyFilters</code> key:</p>
<pre><code>owner abc            -&gt; invalid character 'o' looking for beginning of value
organization_id x    -&gt; invalid character 'o' looking for beginning of value
deleted_at Deleted   -&gt; invalid character 'd' looking for beginning of value
</code></pre>
<h2>Environment</h2>

Field | Value
-- | --
Host OS | platform-independent (source-level defect)
Platform | any; local-provider code path
Meshery Server Version | master @ d279355f9b
Meshery Client Version | n/a

<h2>To Reproduce</h2>
<ol>
<li>
<p>Start Meshery with the default local provider.</p>
</li>
<li>
<p>Send the <code>ApplyFilters</code> grammar, which the sibling endpoint accepts:</p>
<pre><code class="language-bash">curl -G 'http://localhost:9081/api/environments/&lt;envId&gt;/connections' \
  --data-urlencode 'page=0' \
  --data-urlencode 'pagesize=10' \
  --data-urlencode 'filter=owner abc'
</code></pre>
<p>Returns a raw <code>invalid character 'o' looking for beginning of value</code>; the <code>owner</code> filter never runs.</p>
</li>
<li>
<p>Send a non-boolean <code>assigned</code>:</p>
<pre><code class="language-bash">curl -G 'http://localhost:9081/api/environments/&lt;envId&gt;/connections' \
  --data-urlencode 'page=0' \
  --data-urlencode 'pagesize=10' \
  --data-urlencode 'filter={"assigned":"false"}'
</code></pre>
<p>The handler goroutine panics; the connection closes with no response body.</p>
</li>
<li>
<p>For contrast, the sibling endpoint accepts the same value:</p>
<pre><code class="language-bash">curl -G 'http://localhost:9081/api/environments' \
  --data-urlencode 'filter=owner abc'
</code></pre>
<p>Returns <code>200</code> with a normal (here empty) result body — the <code>owner</code> filter is applied rather than rejected. Step 2 sends the identical value to the sibling endpoint and gets a parse error instead.</p>
</li>
</ol>
<h2>Related</h2>
<ul>
<li>#20445 proposes a gorilla/mux recovery middleware, which would convert the panic in step 3 into a structured 500. It does not address the grammar conflict, which is the substance of this report.</li>
<li>The unchecked-assertion class is tracked in #21816, #20437, #21319 and #21051. <code>environment_persister.go:251</code> is not enumerated in any of them.</li>
<li>The <code>page</code>/<code>pagesize</code> parameters are supplied explicitly in the reproduction above because omitting them triggers a separate, unrelated defect in the same function that open PR #21011 already fixes.</li>
</ul></body></html>

Gurkaran18:
**Notes for Reviewers**

- This PR fixes #21826

### What was wrong

`GetEnvironmentConnections` fed one `filter` query parameter to two parsers with incompatible grammars — `json.Unmarshal` (`environment_persister.go:245`) and `utils.ApplyFilters` (`:276`), which splits on a space into `key value` (`helpers/utils/utils.go:378`). Each disabled the other:

- **JSON input** (`{"assigned":false}` — what Meshery UI sends) reached `ApplyFilters` as a single token with no space, matching neither `deleted_at` nor any `dynamicKeys` entry. A silent no-op, so the declared `owner` and `organization_id` filters were **unreachable**.
- **`ApplyFilters` input** (`owner abc`) failed the JSON parse first, and the raw `encoding/json` message was returned to the client.

`assigned` was then read with an unchecked type assertion (`:251`), so a non-boolean panicked the handler goroutine. There is no panic-recovery middleware on the HTTP path (see #20445).

The same file's `GetEnvironments` (lines 49-50) declares the identical `dynamicKeys` and calls `ApplyFilters` **without** a JSON parse, which is why the two sibling endpoints behaved oppositely on the same parameter.

### What this changes

Settles on the **JSON grammar** — the one Meshery UI already sends, so no shipped client changes and no API version bump — and decodes it once in `ParseEnvironmentConnectionsFilter`:

- `owner` now applies as a real predicate (`connections.owner = ?`).
- `assigned` uses the comma-ok idiom; a wrong type is a caller error, not a panic.
- A rejected filter answers **400** (new MeshKit code `meshery-server-1486`) instead of letting `providerStatus` fall back to 502 and blame a remote provider that was never asked — and on the local provider was never involved at all.

Validation stays in the persister rather than the handler on purpose: `RemoteProvider` forwards `filter` opaquely upstream, so validating in the handler would impose Meshery's local grammar on remote providers.

### Two decisions worth your attention

1. **`organization_id` is dropped, not implemented.** The queries run against `connections`, which has an `owner` column but no `organization_id`, so that key could only ever have produced a SQL error had it been reachable. Happy to wire it through a join to `environments` instead if you'd prefer that.
2. **Overlaps #21011**, which adds pagination defaults ~10 lines below this block in the same function. Whichever merges second needs a trivial rebase.

**Follow-up (separate repo):** the `filter` parameter in `meshery/schemas` (`constructs/v1beta1/environment/api.yml`, operation `getEnvironmentConnections`) is still an undescribed bare `string`. Documenting the grammar there is what prevents recurrence.

### Before

Both regressions confirmed to fail against the previous code:

    --- FAIL: TestGetEnvironmentConnectionsOwnerFilterIsApplied (0.00s)
        environment_connections_persister_test.go:128: owner-filtered TotalCount = 2, want 1
    --- FAIL: TestGetEnvironmentConnectionsRejectsMalformedFilter/{"assigned":"false"} (0.00s)
    panic: interface conversion: interface {} is string, not bool [recovered, repanicked]
    FAIL	github.com/meshery/meshery/server/models

### After

    $ go test ./server/models/ ./server/handlers/
    ok  	github.com/meshery/meshery/server/models	23.239s
    ok  	github.com/meshery/meshery/server/handlers	5.605s

    $ golangci-lint run --config=.github/.golangci.yml ./server/...
    0 issues.

    $ go run github.com/meshery/meshkit/cmd/errorutil -d . analyze \
        -i ./server/helpers -o ./server/helpers --skip-dirs mesheryctl | wc -c
    0

Handler status mapping is pinned in **both** directions, so the fix does not disturb the existing provider-status propagation:

    --- PASS: TestGetConnectionsOfEnvironmentHandler_FilterErrorIs400
        --- PASS: /malformed_filter_is_a_client_error         (400)
        --- PASS: /non-JSON_filter_is_a_client_error          (400)
        --- PASS: /provider_403_still_surfaces_as_403         (403)
        --- PASS: /unreachable_provider_still_defaults_to_502 (502)

**Unrelated pre-existing failure:** `server/policies` `TestRelationshipMutationPathsResolveAgainstComponentSchemas` fails with `model "cert-manager" ships no version directories - the scope list is stale`. I verified it fails identically on clean `master` @ `d279355f9b`; it is not caused by this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment connection filtering by owner and assignment status.
  * Invalid, malformed, or unsupported filter values now return a clear HTTP 400 response.
  * Added structured error details and remediation guidance for invalid connection filters.
  * Preserved appropriate status responses for authorization failures and connectivity issues.
  * Improved handling of empty filters with consistent default behavior.


kunalKumar-13 (commented):
The diagnosis here is very good, and I checked the part it hinges on. `ApplyFilters` really does split on a space and take the first token as the key:

```go
// server/helpers/utils/utils.go:378
filterArr := strings.Split(filter, " ")
filterKey := filterArr[0]
```

So `{"assigned":false}` arrives as one token, `filterKey` is the whole JSON string, and it matches neither `deleted_at` nor anything in `dynamicKeys` — it falls through to a no-op. That's exactly your point about `owner` and `organization_id` being unreachable: they're declared as dynamic keys but nothing can ever match them, because the only input that would match is input the JSON parser above rejects first. Two parsers with incompatible grammars on one parameter, each making the other dead.

Decoding once into a typed struct is the right fix rather than trying to make both grammars coexist.

---

**One thing that needs sorting before merge, and it isn't your fault:** this PR and #21879 both claim `meshery-server-1486`.

```
#21833 (this one)  server/models/error.go     ErrInvalidEnvironmentConnectionsFilterCode = "meshery-server-1486"
#21879             server/handlers/error.go   ErrGetConnectionCode                       = "meshery-server-1486"
```

Both also bump `server/helpers/component_info.json` from `next_error_code: 1486` to `1487`, so you'll get a conflict on that file whichever merges second — which is lucky, because it forces someone to look. The trap is that resolving the JSON conflict alone leaves two different errors sharing code 1486 in the codebase, and nothing checks for that.

Worth a comment on both PRs so whoever lands second renumbers to 1487 and bumps the counter to 1488, rather than just taking "theirs" on the JSON.

(I've left the same note on #21879 so it doesn't depend on which of you sees this first.)

---

Smaller things, all fine:

- The 400 rather than a provider error is the right call — a malformed `filter` is a caller mistake, and the remediation string pointing at `filter={"assigned":false}` is genuinely useful in a way most error strings aren't.
- Test coverage across four files for a change this size is more than most PRs this size bring.

Gurkaran18:
@kunalKumar-13 Confirmed — this branch has 1486 with `next_error_code: 1487`, and master is still at 1486, so #21879 hasn't landed either.

Proposal: this PR keeps 1486 (older claim), #21879 takes 1487 / counter 1488. Then the codes are distinct either way and the JSON conflict resolves by one rule — take the higher counter. Same note left on #21879.

kunalKumar-13:
That split works and I'd go with it. This PR keeps 1486 as the older claim, #21879 takes 1487 and the counter goes to 1488, and "take the higher counter" resolves the `component_info.json` conflict without anyone having to think about it.

Worth noting for whoever lands second: resolving the JSON alone is not enough. If both branches still carry 1486 in their own `error.go`, taking the higher counter leaves two errors sharing a code and nothing validates that. The renumber has to happen in the Go file too.

No clash with mine, for what it's worth — #21896 and #21901 are in the `mesheryctl-` namespace (1257 and 1259), separate from `meshery-server-`. Though that counter has the same problem: `mesheryctl-1255` is currently claimed by #21748, #21678 and #21521 at once.

Three collisions across two counters in one week seems like enough to be worth its own issue — a `next_error_code` that every concurrent PR reads and increments is going to keep doing this. Happy to open one if nobody has.

Gurkaran18:
@kunalKumar-13 Agreed, and yes — the renumber has to land in `error.go`, not just the JSON. That's in the note on #21879.

Please do open the counter issue — confirmed on master: `mesheryctl` is at `next_error_code: 1255` with 1254 the highest in use, so 1255 is what every concurrent PR reads. Any fix has to work without a central allocator.

ritzorama:
Nice, focused fix. I inspected the parser, persister query paths, handler status mapping, structured error registration, and new regression tests. The decode-once approach correctly removes the competing filter grammars and the 400 mapping distinguishes caller input errors from provider failures. One cleanup remains: `environment_connections_persister_test.go:14` introduces a deprecated `v1beta1` connection import solely for legacy-column setup. Please build the fixture from v1beta3/local models and explicitly create only the needed legacy columns in the test setup.

PR #21839, opened 2026-09-07
fix(server): safely handle missing kid and exp in JWT validation

### LINKED ISSUE #21816: [Bug]: Unhandled panic in JWT validation crashes server across 150+ APIs (DoS)
#### Description
In `server/models/remote_auth.go:312`, `VerifyToken()` performs an unchecked type assertion:
```go
kid := tokenUP.Header["kid"].(string)
```
If an incoming JWT header omits `kid` or provides a non-string value, Go triggers a runtime panic:
```text
panic: interface conversion: interface {} is nil, not string
```

#### Impact (150+ APIs)
Over **150+ endpoints** in `server/router/server.go` (including `/api/user`, `/api/system/kubernetes`, `/api/perf/profile`, `/api/system/graphql/query`) route through `AuthMiddleware` $\rightarrow$ `VerifyToken`. Sending a token without `kid` to **any** of these routes immediately crashes the entire Meshery server process (Denial of Service).

#### How to Reproduce
Run this curl command against a running instance:
```bash
curl -i -X GET "http://localhost:9081/api/user" \
  -H "Cookie: meshery-provider=localhost:9876; token=eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJzdWIiOiIxMjM0NTY3ODkwIn0."
```
**Result:** Server crashes immediately with `curl: (52) Empty reply from server` and logs the panic at `remote_auth.go:312`.

#### Proposed Fix
Safely check `kid` using the `comma-ok` idiom:
```go
kidVal, ok := tokenUP.Header["kid"]
if !ok {
    return nil, fmt.Errorf("missing 'kid' in token header")
}
kid, ok := kidVal.(string)
if !ok {
    return nil, fmt.Errorf("invalid 'kid' format in token header: expected string")
}
```

suryaff733:
**Notes for Reviewers**

Fixes #21816. Safely handles unchecked type assertions in JWT verification (`remote_auth.go`) to prevent runtime panics on missing/invalid `kid` and `exp`.

| **Missing/invalid `kid`** | `curl: (52) Empty reply`<br>*(Panic crashes request connection)* | `HTTP 302 / Error`<br>*(Graceful redirection / rejection)* |
| **`kid` assertion** | `tokenUP.Header["kid"].(string)` | Comma-ok check `kid, ok := ...` |
| **`exp` assertion** | `int64(jtk["exp"].(float64))` | Type switch supporting `float64`, `int64`, `json.Number` |

### Changes
- `server/models/remote_auth.go`: Comma-ok guards for `kid` and safe type parsing for `exp`.
- `server/handlers/middlewares.go`: Added `recover()` safeguard in `AuthMiddleware`.
- `server/models/remote_auth_test.go`: Added unit tests verifying zero panics on malformed tokens.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved authentication token validation for malformed, incomplete, or invalid tokens.
  - Expired tokens are now rejected consistently, while invalid expiration values fail safely.
  - Prevented authentication-related errors from causing unhandled server crashes; affected requests now receive an HTTP 500 response.
  - Added safeguards to ensure invalid authentication inputs are handled without unexpected failures.
  - Authentication error responses now use a consistent JSON format when unexpected failures occur.

### DIFF
--- a/server/handlers/middlewares.go
+++ b/server/handlers/middlewares.go
@@ -127,6 +127,13 @@ func (h *Handler) NoCacheMiddleware(next http.Handler) http.Handler {
 // AuthMiddleware is a middleware to validate if a user is authenticated
 func (h *Handler) AuthMiddleware(next http.Handler, auth models.AuthenticationMechanism) http.Handler {
 	fn := func(w http.ResponseWriter, req *http.Request) {
+		defer func() {
+			if r := recover(); r != nil {
+				h.log.Error(fmt.Errorf("recovered from panic in AuthMiddleware: %v", r))
+				http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+			}
+		}()
+
 		refURLB64 := GetRefURL(req)
 
 		providerH := h.Provider
--- a/server/handlers/middlewares.go
+++ b/server/handlers/middlewares.go
@@ -130,7 +130,7 @@ func (h *Handler) AuthMiddleware(next http.Handler, auth models.AuthenticationMe
 		defer func() {
 			if r := recover(); r != nil {
 				h.log.Error(fmt.Errorf("recovered from panic in AuthMiddleware: %v", r))
-				http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+				writeJSONError(w, "Internal Server Error", http.StatusInternalServerError)
 			}
 		}()
 
--- a/server/models/remote_auth.go
+++ b/server/models/remote_auth.go
@@ -309,19 +309,45 @@ func (l *RemoteProvider) VerifyToken(tokenString string) (*jwt.MapClaims, error)
 	if err != nil {
 		return nil, ErrPraseUnverified(err)
 	}
-	kid := tokenUP.Header["kid"].(string)
+	kid, ok := tokenUP.Header["kid"].(string)
+	if !ok || kid == "" {
+		return nil, ErrPraseUnverified(fmt.Errorf("token header missing or invalid 'kid'"))
+	}
+
+	if len(x) < 2 {
+		return nil, ErrPraseUnverified(fmt.Errorf("malformed token: missing payload segment"))
+	}
 
 	var jtk map[string]interface{}
-	t, _ := base64.RawStdEncoding.DecodeString(x[1])
+	t, decodeErr := base64.RawStdEncoding.DecodeString(x[1])
+	if decodeErr != nil {
+		t, decodeErr = base64.RawURLEncoding.DecodeString(x[1])
+		if decodeErr != nil {
+			return nil, ErrPraseUnverified(decodeErr)
+		}
+	}
 	if err := json.Unmarshal(t, &jtk); err != nil {
 		return nil, ErrPraseUnverified(err)
 	}
 
 	// TODO: Once hydra fixes https://github.com/ory/hydra/issues/1542
 	// we should rather configure hydra auth server to remove nbf field in the token
-	_, ok := jtk["exp"]
-	if ok {
-		exp := int64(jtk["exp"].(float64))
+	if expRaw, ok := jtk["exp"]; ok {
+		var exp int64
+		switch v := expRaw.(type) {
+		case float64:
+			exp = int64(v)
+		case int64:
+			exp = v
+		case json.Number:
+			parsed, err := v.Int64()
+			if err != nil {
+				return nil, ErrPraseUnverified(fmt.Errorf("token claim 'exp' is not a valid number: %w", err))
+			}
+			exp = parsed
+		default:
+			return nil, ErrPraseUnverified(fmt.Errorf("token claim 'exp' is not a valid number"))
+		}
 		if time.Now().Unix() > exp {
 			return nil, ErrTokenExpired
 		}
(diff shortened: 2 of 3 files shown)


SuryanshGarg04:
` Error: /home/runner/work/meshery/meshery/server/handlers/middlewares.go:133:5: use of `http.Error` forbidden because "Use writeMeshkitError(w, err, status) or writeJSONError(w, msg, status) from server/handlers/utils.go. http.Error writes Content-Type: text/plain and breaks RTK Query's default baseQuery. See docs/content/en/project/contributing/error-contract.md." `

please solve this golang lint error

SuryanshGarg04:
The coderabbitai comment is also valid, please consider that also
It is the fix for the lint failure

suryaff733:
Updated! Replaced http.Error with writeJSONError to comply with the error contract and resolved the forbidigo linter error. Thanks for pointing it out @SuryanshGarg04!

server/handlers/middlewares.go
```diff
@@ -127,6 +127,13 @@ func (h *Handler) NoCacheMiddleware(next http.Handler) http.Handler {
...
+		defer func() {
+			if r := recover(); r != nil {
+				h.log.Error(fmt.Errorf("recovered from panic in AuthMiddleware: %v", r))
+				writeJSONError(w, "Internal Server Error", http.StatusInternalServerError)
+			}
+		}()
+
 		refURLB64 := GetRefURL(req)
 
 		providerH := h.Provider
```
SuryanshGarg04: Could we remove the `recover()` from `AuthMiddleware`?

Since it wraps `next.ServeHTTP(...)`, it can also catch panics from handlers further down the request, even when they're unrelated to authentication. It could also make the response a bit messy if a handler has already started writing before the panic.

Since the original `VerifyToken` panic is already fixed, I think we can drop this recovery and keep the middleware focused on authentication.

kunalKumar-13 (commented):
Both panics are real on master, for anyone reading who wants the line numbers:

```
remote_auth.go:312    kid := tokenUP.Header["kid"].(string)
remote_auth.go:324    exp := int64(jtk["exp"].(float64))
```

A token with no `kid` gives `nil.(string)`, and an `exp` that arrives as anything but `float64` gives the same on line 324 — both take down the connection rather than returning an error, which matches the `curl: (52) Empty reply` in your table.

I spent most of the time on the base64 change, since that's the part where a subtle mistake would be invisible. **The fallback is safe**, though for a slightly non-obvious reason worth writing down:

base64url uses `A-Za-z0-9-_` and standard base64 uses `A-Za-z0-9+/`. A valid base64url segment therefore never contains `+` or `/`, so `RawStdEncoding` can only succeed on the shared `A-Za-z0-9` subset — where both alphabets agree and decode to identical bytes. When the segment does contain `-` or `_`, std fails and your `RawURLEncoding` fallback catches it. I ran three payloads through it (one plain, two producing `-`/`_`) and all three round-trip.

So it's correct. It's just the reverse of the spec order — JWT is base64url by RFC 7519, so `RawURLEncoding` is the primary and std is the odd case. Might read better that way round, but that's taste rather than a defect.

**Something you fixed that isn't in the description**: master had

```go
t, _ := base64.RawStdEncoding.DecodeString(x[1])
```

discarding the decode error entirely. Any payload containing `-` or `_` — which is most real tokens — failed silently there and only surfaced as a confusing `json.Unmarshal` error further down. Checking `decodeErr` is a real fix on its own and worth calling out.

Two smaller notes:

**`len(x) < 2` looks unreachable.** `ParseUnverified` errors out unless the token has exactly three segments, and you already return on that error above, so `x` should always have three by the time you index it. Harmless as defence in depth, but it isn't guarding a live path.

**The blanket `recover()` in `AuthMiddleware` is worth a maintainer's opinion.** Now that both assertions are fixed at source, it's belt-and-braces — and it's safe in the sense that turning a panic into a 500 fails closed rather than open. But it will also swallow any *future* panic anywhere in the auth path and turn it into a generic 500. It does log loudly, which is the right call if you keep it. I'd just want someone who knows the project's conventions to say whether a catch-all there is wanted, or whether the source-level fixes are considered sufficient.

The `exp` type switch covering `float64`, `int64` and `json.Number` is the right shape — `json.Number` in particular only appears if a decoder is configured with `UseNumber()`, which is exactly the sort of thing that changes under someone else's refactor and would have panicked here.

suryaff733:
Thanks @kunalKumar-13 for the thorough review and testing!

- **Base64 order**: Makes total sense. I'll swap it to try `RawURLEncoding` first per RFC 7519, falling back to `RawStdEncoding`.
- **`decodeErr`**: Good catch, I'll update the PR description to mention that we now properly handle the base64 decode error instead of silently discarding it.
- **`len(x) < 2`**: Agreed, it's just defense-in-depth since `ParseUnverified` already checks segments.
- **`recover()` in AuthMiddleware**: Added as defense-in-depth so unexpected panics fail closed with a 500 instead of dropping the connection. Happy to hear maintainers' thoughts on keeping it vs. relying solely on the source fix.

Will push the small base64 ordering tweak and update the description shortly!

PARTH-TUSSLE:

ritzorama:
I inspected JWT parsing, middleware flow, callers, tests, and checks. The claim guards are solid; please scope panic recovery to authentication work so a panic from `next.ServeHTTP` cannot append a 500 JSON body after a downstream handler has already committed a successful response. Re-panic downstream failures (or otherwise leave them to the existing outer recovery) once control enters `next`.

PR #21843, opened 2026-09-08
[UI] Surface administrative environment purpose badge and restrict affordances

NiftyOracle:
Fixes #21803

### environment-card.tsx
- Surface `administrative` purpose badge using Sistent `Chip` component on both front and back card faces
- Positive test against literal `"administrative"` as specified (no negative test against `"user"`)
- Edit and delete buttons disabled for administrative environments with descriptive tooltips
- `disableFlip` reverted to original behavior
- Used `.some()` for Prettier compliance while preserving loose equality behavior

## Summary by CodeRabbit

- **New Features**
  - Administrative environments now display an “Administrative” badge on both sides of their environment cards.
  - Clear tooltips explain why administrative environments cannot be edited or deleted.
- **UI Improvements**
  - Administrative environment names and badges are now centered for improved readability.
  - Edit and delete controls are visibly disabled for administrative environments.

### DIFF
--- a/ui/components/environments/environment-card.tsx
+++ b/ui/components/environments/environment-card.tsx
@@ -11,6 +11,7 @@ import {
   useTheme,
   IconButton,
   CustomTooltip,
+  Chip,
 } from '@sistent/sistent';
 import { iconMedium } from '../../css/icons.styles';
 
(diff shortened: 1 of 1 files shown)


yi-nuo426:
The administrative guard applies to individual edit/delete buttons but not bulk selection. ui/components/environments/environment-card.tsx:236 leaves BulkSelectCheckbox disabled only for deleted environments, so an administrative environment can still be selected and deleted through the parent bulk-delete flow. Please disable selection when isAdministrative is true and guard the bulk-delete handler so administrative IDs cannot reach the mutation.

Issue #21845, opened 2026-09-08
[Docs/QA] mesheryctl BATS e2e suite: 140 of 185 tests untagged for the qa dashboard traceability contract


MU5A:
## Current Behavior

Raised by Lee in #documentation while pointing a newcomer at mesheryctl: "I don't think that we have a dedicated quality report for mesheryctl's end-to-end tests that are written using bash [BATS]."

Checked this against the actual pipeline rather than going on the comment alone. mesheryctl BATS results do reach the `meshery/qa` dashboard (`https://qa.meshery.io`): `bats-to-allure.js` always stamps `project: "mesheryctl"`, and that label is what the dashboard's `allurerc.mjs` filters on to build the "Mesheryctl" report tile. So results aren't missing, they land in one flat, undifferentiated report.

The real gap is upstream of that: the converter's Test Plan traceability (`[TC-n]`, `[cut=...]`, `[tg=...]` title tokens, documented at the top of `bats-to-allure.js`) is opt-in per test. Counted actual coverage:

```
Total @test declarations: 185
Tagged (title starts with a bracket token): 45  (24%)
```

Broken down by suite directory (`mesheryctl/tests/e2e/`):

| Suite | Total | Tagged |
|---|---|---|
| 000-prerequisites | 13 | 0 |
| 001-system | 16 | 2 |
| 002-model | 36 | 0 |
| 003-design | 16 | 0 |
| 004-perf | 6 | 0 |
| 005-component | 13 | 0 |
| 006-registry | 20 | 0 |
| 007-connection | 28 | **28** |
| 008-workspace | 5 | 0 |
| 009-diagnostics | 1 | 1 |
| 999-system | 31 | 14 |

`007-connection` is fully tagged and is clearly where the convention originated (it feeds the existing "Connection Lifecycle" cross-client report in `allurerc.mjs`, which explicitly aggregates UI + CLI results by `testGroup`). The other nine suites are effectively untagged, so they get no `componentUnderTest`/`testGroup` breakdown in the dashboard and no Test Plan deep-link, just a flat pass/fail count.

## Desired Behavior

Tag the remaining ~140 tests with `[cut=...]` and `[tg=...]` (and `[TC-n]` where a corresponding row exists) so each suite gets a meaningful, filterable breakdown in the dashboard, matching what the UI Playwright suite already does.

## Why this isn't one PR

[...]

PR #21846, opened 2026-09-08
Tag the workspace BATS suite for the qa dashboard traceability contract

MU5A:
Small starting slice for #21845.

Lee mentioned in Slack that mesheryctl's BATS e2e suite has no dedicated quality report, unlike the UI Playwright suite. Checked this against the actual `meshery/qa` dashboard config and `bats-to-allure.js`: mesheryctl results do reach the dashboard (`project: "mesheryctl"` is always stamped), but the converter's Test Plan traceability tagging (`[TC-n]`, `[cut=...]`, `[tg=...]` title tokens) is opt-in per test, and only 45 of 185 mesheryctl BATS tests (24%) use it. `007-connection` is fully tagged, clearly the origin of the convention. The other nine suite directories are effectively untagged, full breakdown in #21845.

That's too large to fix in one PR (assigning real `[TC-n]` IDs needs coordination against the shared Test Plan spreadsheet I don't have visibility into). This tags the smallest fully-untagged suite, `008-workspace` (5 tests), as a small, concrete, verified demonstration: `[cut=Workspace][tg=Workspace Management]` on each test.

Deliberately no `[TC-n]`: that token claims a specific spreadsheet row, and fabricating one would be worse than leaving it out. `cut`/`tg` alone are enough to give this suite its own component/Test-Group breakdown in the dashboard without that claim.

## Test plan
- [x] Ran `parseTitleTokens()` (bats-to-allure.js) directly against each new title, confirmed correct `componentUnderTest`/`testGroup` labels with tokens cleanly stripped from the display name
- [x] Existing converter regression suite (`node --test bats-to-allure.test.js`, 27 tests) still passes
- [x] Installed bats-core and confirmed all 5 tests in the suite still parse correctly (`bats --count` reports 5, unchanged)

## Summary by CodeRabbit

- **Tests**
  - Updated workspace end-to-end test names with standardized Workspace Management tags.
  - Test behavior, assertions, and coverage remain unchanged.

### DIFF
--- a/mesheryctl/tests/e2e/008-workspace/00-workspace.bats
+++ b/mesheryctl/tests/e2e/008-workspace/00-workspace.bats
@@ -6,7 +6,7 @@ setup() {
     _load_bats_libraries
 }
 
-@test "given --help flag provided when running mesheryctl workspace --help displays command usage" {
+@test "[cut=Workspace][tg=Workspace Management] given --help flag provided when running mesheryctl workspace --help displays command usage" {
     run $MESHERYCTL_BIN workspace --help
 
     assert_success
--- a/mesheryctl/tests/e2e/008-workspace/01-workspace-create.bats
+++ b/mesheryctl/tests/e2e/008-workspace/01-workspace-create.bats
@@ -6,7 +6,7 @@ setup() {
 	_load_bats_libraries
 }
 
-@test "given an invalid orgId provided as an argument when running mesheryctl workspace create --orgId invalid-org-id then an error message is displayed" {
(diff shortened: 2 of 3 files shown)


MU5A:
Ready for review whenever someone has a moment. Small, self-contained slice, tags the 5-test workspace suite so it gets its own component/Test-Group breakdown on the qa dashboard. Full context and the larger gap this is a starting point for are in #21845.

ritzorama:
Nice, focused traceability update. I inspected all five tagged workspace tests and the BATS-to-Allure token parser; the `[cut=Workspace]` and `[tg=Workspace Management]` prefixes are recognized and preserve the displayed test names after conversion.

PR #20458, opened 2026-07-03
[Server] Transition Kubernetes connection to DISCONNECTED on authorization failure (#14083)

SuryanshGarg04:
Fixes #14083
### Overview
This PR fixes an issue where Kubernetes context discovery could get stuck in an infinite retry loop when the Kubernetes API returned authentication or authorization errors such as `401 Unauthorized` or `403 Forbidden`.
These errors were previously treated as temporary discovery failures, causing Meshery to repeatedly retry discovery instead of moving the connection to `DISCONNECTED`.
---
### Problem & Root Cause
There were three main issues involved:
1. **Kubernetes errors were being masked**
   `AssignServerID()` performs a namespace lookup against the Kubernetes API. When the request failed because of missing permissions or invalid credentials, the original Kubernetes `StatusError` was wrapped in `ErrUnreachableKubeAPI`. Because the MeshKit error type did not implement `Unwrap()`, the underlying Kubernetes error was no longer available to the caller. This meant the HTTP status code could not be inspected.
2. **Authorization failures were treated as `NOTFOUND`**
   `DiscoverAction.Execute()` received the wrapped error and treated it as a generic discovery failure, causing the connection to enter the `NOTFOUND` state.
3. **Discovery continued indefinitely**
   `NOTFOUND` is an active connection state. Background synchronization and Kubernetes middleware continued to manage the connection and trigger discovery again. This resulted in repeated Kubernetes API requests, repeated RBAC/authentication errors, and unnecessary log noise. Furthermore, there was a missing FSM transition from the `Discovered` state to `DISCONNECTED`, which prevented the state machine from directly disconnecting the connection after an authorization failure.
---
### Changes
* **`server/models/k8s_context.go`**
  * `AssignServerID()` now returns the original Kubernetes API error from the namespace lookup.
  * Preserves the `*apierrors.StatusError` type so callers can inspect the HTTP status code.
* **`server/machines/kubernetes/discover.go`**
  * Added explicit handling for `401 Unauthorized` and `403 Forbidden` using `k8serrors.IsUnauthorized()` and `k8serrors.IsForbidden()`.
  * Authentication and authorization failures now trigger a disconnect (`machines.Disconnect`).
  * Transient failures such as timeouts, connection refused errors, and `503 Service Unavailable` continue to use the existing `NOTFOUND` behavior so that discovery can retry automatically.
* **`server/machines/kubernetes/machine.go`**
  * Added the missing `Disconnect: machines.DISCONNECTED` transition for the `Discovered()` state.
[...]

PR #21858, opened 2026-09-08
[Server] Exclude service-account token from K8sContext ID generation

### LINKED ISSUE #21810: Impact of in-cluster service-account token rotation on deterministic connection IDs
## Problem

For in-cluster Kubernetes contexts, the service-account token is included in the `Auth` data used to construct the Kubernetes context.

`K8sContextGenerateID` derives the connection/context identity from that data.

When the projected service-account token rotates, the `Auth` value changes even though the Kubernetes cluster and logical connection remain the same.

Because the generated ID incorporates the token contents, token rotation can therefore produce a **different Kubernetes connection ID for the same in-cluster cluster**.

## Current behavior

The effective flow is:

```text
in-cluster service-account token
        ↓
Kubernetes context Auth
        ↓
K8sContextGenerateID hashes Auth
        ↓
token rotates
        ↓
Auth changes
        ↓
generated ID changes
        ↓
new logical Kubernetes connection
```

The previous connection may remain persisted while the newly generated connection is discovered as a separate context.

## Impact

This can cause the same in-cluster Kubernetes server to accumulate multiple logical connection records over time.

It may also cause discovery and connection-management behavior to repeat when the service-account token rotates, because the newly generated connection does not inherit the previous connection's terminal/managed state.

Potential consequences include:

* duplicate Kubernetes connection records;
* repeated discovery of the same cluster;
* stale disconnected connections accumulating in persistence;
* unnecessary Kubernetes API requests after credential rotation;
* inconsistent connection state for the same logical cluster.

The exact downstream impact should be audited across connection loading, persistence, middleware discovery, and any consumers of Kubernetes connection IDs.

## Expected behavior

Rotation of an in-cluster service-account token should **not change the logical identity of the Kubernetes cluster/connection**.

Credential rotation should update authentication material associated with the existing connection rather than create a new logical connection solely because the token contents changed.

Conceptually:

```text
same Kubernetes cluster
+
rotated service-account token
        ↓
same connection identity
        ↓
updated credentials
```

## Suggested direction

Review the inputs used by `K8sContextGenerateID` and separate:

* stable connection identity;
* mutable authentication credentials.

The generated connection ID should be based on stable cluster/context identity rather than the current service-account token contents.

The implementation should also consider how existing persisted connections are matched when credentials change.

Add regression coverage that simulates service-account token rotation and verifies that:

1. the connection ID remains stable;
2. an existing connection is updated rather than duplicated;
3. connection status is preserved appropriately;
4. repeated discovery does not create additional logical connections.

## Scope / Relationship to #14083

This issue was discovered during the investigation for #14083.

It is **pre-existing** and is not introduced by PR #20458.

PR #20458 fixes the authorization-failure retry loop by preventing automatic rediscovery of a `DISCONNECTED` connection.

This issue is broader: credential rotation can create a **new connection identity**, potentially bypassing the state of the previous connection.

It should therefore be tracked separately and does not block #20458.

Rajkaran-122:
Fixes #21810

Previously, the K8sContextGenerateID function included the service-account token in the hash that generates connection IDs. When Kubernetes rotates in-cluster service-account tokens, this caused the same logical cluster to generate different connection IDs, leading to duplicate connection records and unnecessary rediscovery.

Changes:
- Modified K8sContextGenerateID to exclude tokens from the hash
- For contexts with token auth, the token is omitted from ID generation
- Connection IDs now remain stable across token rotation
- Added test coverage for token rotation scenarios

Testing:
- Added TestK8sContextGenerateID with multiple scenarios
- Verified IDs remain stable after token changes
- Tested both in-cluster and regular contexts

**Notes for Reviewers**

- This PR fixes #

## Summary by CodeRabbit

- **Bug Fixes**
  - Kubernetes in-cluster context identifiers remain stable when service-account tokens or CA certificates rotate.
  - Regular contexts continue receiving distinct identifiers when authentication or CA data changes.
  - Existing deployment types are preserved during kubeconfig discovery.
  - Improved handling and reporting of failures when reading in-cluster credentials.
  - Existing connections and credentials are reconciled and refreshed when updated Kubernetes context details are saved.

- **Tests**
  - Added coverage for token and CA rotation, persistence, reload, and legacy connection reconciliation.

### DIFF
--- a/server/handlers/k8sconfig_handler.go
+++ b/server/handlers/k8sconfig_handler.go
@@ -567,6 +567,7 @@ func (h *Handler) DiscoverK8SContextFromKubeConfig(userID string, token string, 
 			h.log.Warn(ErrNilClusterContext(err))
 			return contexts, err
 		}
+		// DeploymentType is already set by NewK8sContextFromInClusterConfig before ID generation
 		cc.DeploymentType = "in_cluster"
 		conn, err := prov.SaveK8sContext(token, *cc, nil)
 		if err != nil {
@@ -599,7 +600,10 @@ func (h *Handler) DiscoverK8SContextFromKubeConfig(userID string, token string, 
 		metadata["context"] = models.RedactCredentialsForContext(ctx)
 		metadata["description"] = fmt.Sprintf("K8S context \"%s\" discovered with cluster at %s", ctx.Name, ctx.Server)
 		metadata["description"] = fmt.Sprintf("Connection established with context \"%s\" at %s", ctx.Name, ctx.Server)
-		ctx.DeploymentType = "out_of_cluster"
+		// Set deployment type for kubeconfig contexts if not already set
+		if ctx.DeploymentType == "" {
+			ctx.DeploymentType = "out_of_cluster"
+		}
 		conn, err := prov.SaveK8sContext(token, *ctx, nil)
 		if err != nil {
 			h.log.Warn(ErrFailToSaveContext(err))
--- a/server/models/k8s_context.go
+++ b/server/models/k8s_context.go
@@ -403,7 +403,19 @@ func K8sContextGenerateID(kc K8sContext) (string, error) {
 	// For in-cluster contexts, exclude the token from the hash to prevent ID changes
 	// when the service-account token rotates. The token is mutable authentication
 	// material that should not affect the logical connection identity.
-	if kc.Auth != nil {
+	// In-cluster contexts are identified by their server URL pattern (kubernetes.default.svc).
+	isInCluster := false
+	serverURL := kc.Server
+	if serverURL == "" && kc.Cluster != nil {
+		if clusterServer, ok := kc.Cluster["server"].(string); ok {
+			serverURL = clusterServer
+		}
+	}
+	if serverURL != "" && (serverURL == "https://kubernetes.default.svc" || serverURL == "https://kubernetes.default") {
+		isInCluster = true
+	}
+
+	if isInCluster && kc.Auth != nil {
 		if user, ok := kc.Auth["user"].(map[string]interface{}); ok {
 			if _, hasToken := user["token"]; hasToken {
 				// Create a copy of auth without the token for ID generation
--- a/server/models/k8s_context.go
+++ b/server/models/k8s_context.go
@@ -48,6 +48,11 @@ type K8sContext struct {
 	// transition to the connected state. An unreachable context can still be
 	// registered as a (discovered) connection.
 	Reachable bool `json:"reachable" yaml:"-" gorm:"-"`
+	// IsInCluster indicates whether the context was created from in-cluster
+	// configuration. It is transient (never persisted): it is set by
+	// NewK8sContextFromInClusterConfig and used by K8sContextGenerateID to determine
+	// whether to exclude service-account tokens from the ID hash.
+	IsInCluster bool `json:"isInCluster" yaml:"-" gorm:"-"`
 }
 
 // K8sContextFromConnection converts a kubernetes connection into a K8sContext.
@@ -148,6 +153,7 @@ func (kcfg InternalKubeConfig) K8sContext(name string, instanceID *core.Uuid, lo
 		server,
 		instanceID,
 		log,
+		false, // isInCluster = false for kubeconfig contexts
 	)
 }
 
@@ -158,8 +164,9 @@ func NewK8sContextWithServerID(
 	server string,
 	instanceID *core.Uuid,
 	log logger.Handler,
+	isInCluster bool,
 ) (*K8sContext, error) {
-	ctx, _ := NewK8sContext(contextName, clusters, users, server, instanceID, log)
+	ctx, _ := NewK8sContext(contextName, clusters, users, server, instanceID, log, isInCluster)
 
 	// Perform Ping test on the cluster
 	if err := ctx.PingTest(); err != nil {
@@ -349,6 +356,7 @@ func NewK8sContextFromInClusterConfig(contextName string, instanceID *core.Uuid,
 		server,
 		instanceID,
 		log,
+		true, // isInCluster = true for in-cluster config
 	)
 }
 
@@ -365,6 +373,7 @@ func NewK8sContext(
 	server string,
 	instanceID *core.Uuid,
 	log logger.Handler,
+	isInCluster bool,
 ) (K8sContext, string) {
 	ctx := K8sContext{
 		Name:              contextName,
@@ -372,6 +381,7 @@ func NewK8sContext(
 		Auth:              user,
 		Server:            server,
 		MesheryInstanceID: instanceID,
+		IsInCluster:       isInCluster,
 	}
 
 	ID, err := K8sContextGenerateID(ctx)
@@ -403,19 +413,8 @@ func K8sContextGenerateID(kc K8sContext) (string, error) {
 	// For in-cluster contexts, exclude the token from the hash to prevent ID changes
 	// when the service-account token rotates. The token is mutable authentication
 	// material that should not affect the logical connection identity.
-	// In-cluster contexts are identified by their server URL pattern (kubernetes.default.svc).
-	isInCluster := false
-	serverURL := kc.Server
-	if serverURL == "" && kc.Cluster != nil {
-		if clusterServer, ok := kc.Cluster["server"].(string); ok {
-			serverURL = clusterServer
-		}
-	}
-	if serverURL != "" && (serverURL == "https://kubernetes.default.svc" || serverURL == "https://kubernetes.default") {
-		isInCluster = true
-	}
-
-	if isInCluster && kc.Auth != nil {
+	// In-cluster provenance is explicitly set by NewK8sContextFromInClusterConfig.
+	if kc.IsInCluster && kc.Auth != nil {
 		if user, ok := kc.Auth["user"].(map[string]interface{}); ok {
 			if _, hasToken := user["token"]; hasToken {
 				// Create a copy of auth without the token for ID generation
--- a/server/models/k8s_context.go
+++ b/server/models/k8s_context.go
@@ -48,11 +48,6 @@ type K8sContext struct {
 	// transition to the connected state. An unreachable context can still be
 	// registered as a (discovered) connection.
 	Reachable bool `json:"reachable" yaml:"-" gorm:"-"`
-	// IsInCluster indicates whether the context was created from in-cluster
-	// configuration. It is transient (never persisted): it is set by
-	// NewK8sContextFromInClusterConfig and used by K8sContextGenerateID to determine
-	// whether to exclude service-account tokens from the ID hash.
-	IsInCluster bool `json:"isInCluster" yaml:"-" gorm:"-"`
 }
 
 // K8sContextFromConnection converts a kubernetes connection into a K8sContext.
@@ -153,7 +148,6 @@ func (kcfg InternalKubeConfig) K8sContext(name string, instanceID *core.Uuid, lo
 		server,
 		instanceID,
 		log,
-		false, // isInCluster = false for kubeconfig contexts
 	)
 }
 
@@ -164,9 +158,8 @@ func NewK8sContextWithServerID(
 	server string,
 	instanceID *core.Uuid,
 	log logger.Handler,
-	isInCluster bool,
 ) (*K8sContext, error) {
-	ctx, _ := NewK8sContext(contextName, clusters, users, server, instanceID, log, isInCluster)
+	ctx, _ := NewK8sContext(contextName, clusters, users, server, instanceID, log)
 
 	// Perform Ping test on the cluster
 	if err := ctx.PingTest(); err != nil {
@@ -316,29 +309,32 @@ func K8sContextsFromKubeconfigWithOptions(provider Provider, userID string, _ *B
 	return kcs
 }
 
+// Package-level variables for in-cluster file paths, can be overridden in tests
+var (
+	inClusterTokenFile  = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+	inClusterRootCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+)
+
 func NewK8sContextFromInClusterConfig(contextName string, instanceID *core.Uuid, log logger.Handler) (*K8sContext, error) {
-	const (
-		tokenFile  = "/var/run/secrets/kubernetes.io/serviceaccount/token"
-		rootCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
-	)
 	host, port := os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT")
 	if len(host) == 0 || len(port) == 0 {
 		return nil, ErrMesheryNotInCluster
 	}
 
-	token, err := os.ReadFile(tokenFile)
+	token, err := os.ReadFile(inClusterTokenFile)
 	if err != nil {
 		return nil, err
 	}
 
 	server := "https://" + net.JoinHostPort(host, port)
 
-	caData, err := os.ReadFile(rootCAFile)
+	caData, err := os.ReadFile(inClusterRootCAFile)
 	if err != nil {
 		return nil, err
 	}
 
-	return NewK8sContextWithServerID(
+	// Create context with in-cluster deployment type set before ID generation
+	ctx, err := NewK8sContextWithServerID(
 		contextName,
 		map[string]interface{}{
 			"cluster": map[string]interface{}{
@@ -356,8 +352,22 @@ func NewK8sContextFromInClusterConfig(contextName string, instanceID *core.Uuid,
 		server,
 		instanceID,
 		log,
-		true, // isInCluster = true for in-cluster config
 	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Set deployment type to mark this as in-cluster for ID generation
+	ctx.DeploymentType = "in_cluster"
+
+	// Regenerate ID with correct deployment type set
+	newID, err := K8sContextGenerateID(*ctx)
+	if err != nil {
+		return nil, err
+	}
+	ctx.ID = newID
+
+	return ctx, nil
 }
 
 // NewK8sContext takes in name of the context, cluster info of the contexts,
@@ -373,7 +383,6 @@ func NewK8sContext(
 	server string,
 	instanceID *core.Uuid,
 	log logger.Handler,
-	isInCluster bool,
 ) (K8sContext, string) {
 	ctx := K8sContext{
 		Name:              contextName,
@@ -381,7 +390,6 @@ func NewK8sContext(
 		Auth:              user,
 		Server:            server,
 		MesheryInstanceID: instanceID,
-		IsInCluster:       isInCluster,
 	}
 
 	ID, err := K8sContextGenerateID(ctx)
@@ -413,8 +421,8 @@ func K8sContextGenerateID(kc K8sContext) (string, error) {
 	// For in-cluster contexts, exclude the token from the hash to prevent ID changes
 	// when the service-account token rotates. The token is mutable authentication
 	// material that should not affect the logical connection identity.
-	// In-cluster provenance is explicitly set by NewK8sContextFromInClusterConfig.
-	if kc.IsInCluster && kc.Auth != nil {
+	// In-cluster provenance is determined by DeploymentType field, which is persisted.
+	if kc.DeploymentType == "in_cluster" && kc.Auth != nil {
 		if user, ok := kc.Auth["user"].(map[string]interface{}); ok {
 			if _, hasToken := user["token"]; hasToken {
 				// Create a copy of auth without the token for ID generation
--- a/server/models/k8s_context.go
+++ b/server/models/k8s_context.go
@@ -390,12 +390,40 @@ func NewK8sContext(
 // K8sContextGenerateID takes in a kubernetes context and generates an ID for it
 //
 // If the context remains the same, it is guaranteed that the ID will be same
+// For in-cluster contexts, excludes the service-account token from the hash to prevent
+// ID changes when the token rotates (issue #21810)
 func K8sContextGenerateID(kc K8sContext) (string, error) {
 	data := map[string]interface{}{
 		"cluster": kc.Cluster,
 		"auth":    kc.Auth,
 		"meshery": kc.MesheryInstanceID.String(),
 		"name":    kc.Name,
+	}
+
+	// For in-cluster contexts, exclude the token from the hash to prevent ID changes
+	// when the service-account token rotates. The token is mutable authentication
+	// material that should not affect the logical connection identity.
+	if kc.Auth != nil {
+		if user, ok := kc.Auth["user"].(map[string]interface{}); ok {
+			if _, hasToken := user["token"]; hasToken {
+				// Create a copy of auth without the token for ID generation
+				authCopy := make(map[string]interface{})
+				for k, v := range kc.Auth {
+					if k == "user" {
+						userCopy := make(map[string]interface{})
+						for userK, userV := range user {
+							if userK != "token" {
+								userCopy[userK] = userV
+							}
+						}
+						authCopy[k] = userCopy
+					} else {
+						authCopy[k] = v
+					}
+				}
+				data["auth"] = authCopy
+			}
+		}
 	}
 
 	byt, err := json.Marshal(data)
--- a/server/models/k8s_context_test.go
+++ b/server/models/k8s_context_test.go
@@ -274,7 +274,7 @@ func TestK8sContextsFromKubeconfigDiscoversAllContexts(t *testing.T) {
 }
 
 // TestK8sContextGenerateID verifies that K8sContextGenerateID excludes service-account
-// tokens from the hash to prevent ID changes when tokens rotate (issue #21810)
+// tokens from the hash for in-cluster contexts to prevent ID changes when tokens rotate (issue #21810)
 func TestK8sContextGenerateID(t *testing.T) {
 	instanceID, _ := uuid.NewV4()
 
@@ -291,7 +291,7 @@ func TestK8sContextGenerateID(t *testing.T) {
 				Cluster:           sql.Map{"server": "https://k8s.example.com"},
 				MesheryInstanceID: &instanceID,
 			},
-			wantSame: true, // ID should be same after token change
+			wantSame: false, // ID should CHANGE after token change for regular contexts
 		},
 		{
 			name: "in-cluster context with token",
@@ -299,9 +299,21 @@ func TestK8sContextGenerateID(t *testing.T) {
 				Name:              "in-cluster-context",
 				Auth:              sql.Map{"user": map[string]interface{}{"token": "service-account-token"}},
 				Cluster:           sql.Map{"server": "https://kubernetes.default.svc"},
+				Server:            "https://kubernetes.default.svc",
 				MesheryInstanceID: &instanceID,
 			},
-			wantSame: true, // ID should be same after token rotation
+			wantSame: true, // ID should be same after token rotation for in-cluster contexts
+		},
+		{
+			name: "in-cluster context with token (short server URL)",
+			context: K8sContext{
+				Name:              "in-cluster-context-short",
+				Auth:              sql.Map{"user": map[string]interface{}{"token": "service-account-token"}},
+				Cluster:           sql.Map{"server": "https://kubernetes.default"},
+				Server:            "https://kubernetes.default",
+				MesheryInstanceID: &instanceID,
+			},
+			wantSame: true, // ID should be same after token rotation for in-cluster contexts
 		},
 		{
 			name: "context without token",
@@ -311,7 +323,7 @@ func TestK8sContextGenerateID(t *testing.T) {
 				Cluster:           sql.Map{"server": "https://k8s.example.com"},
 				MesheryInstanceID: &instanceID,
 			},
-			wantSame: true,
+			wantSame: true, // ID should be same since no token to change
 		},
 	}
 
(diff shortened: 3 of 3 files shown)


Rajkaran-122:
@SuryanshGarg04 @PARTH-TUSSLE  PTAL  .

kunalKumar-13 (commented):
I replicated `K8sContextGenerateID` both ways and hashed some real context shapes to see what actually changes. The fix does what it says — one thing below is worth a decision before it ships, though.

First, two structural things that could have gone wrong and didn't:

**The `+102/-0` diff works because master already hashes auth.** `server/models/k8s_context.go:396` has `"auth": kc.Auth` in the map, so `data["auth"] = authCopy` overwrites it. If auth hadn't already been in there, this would have *added* it and changed every ID on the system.

**Swapping `sql.Map` for a plain map doesn't shift the hash.** I checked, because a custom marshaller here would have been silent and nasty — `sql.Map` is `map[string]interface{}` with only `Scan` and `Value` (`server/internal/sql/map.go:14`) and no `MarshalJSON`, so `json.Marshal` produces identical bytes either way.

Then the behaviour:

| scenario | before | after |
|---|---|---|
| same cluster, token rotated | different IDs | **same** ✅ |
| cert-auth context, unchanged | same | **same** ✅ no regression |
| same cluster, two different service accounts | different IDs | **same** |
| existing token context, on upgrade | `e40a0aeff42c` | `21a2acdb88f8` |

That last row is what I'd want a maintainer to look at. Every in-cluster connection that already exists will hash to a **new** ID the first time this ships. The old row stays behind under the old ID, a new one gets created — which is the duplicate-connection symptom #21810 describes, happening once on deploy.

The fix is still right; the ID has to change once in order to stop changing forever. But it probably needs either a migration that rewrites existing rows, or a line in the release notes so operators expect one round of rediscovery. Shipping it quietly means the issue looks like it came back a week later.

The third row is a smaller call: two service accounts pointing at the same cluster with the same context name now collapse to one ID rather than two. I think that's intended — your comment says the token is "mutable authentication material that should not affect the logical connection identity", and that reasoning covers it — but it does merge two things that were previously distinct, so better for someone to agree to it than discover it.

Minor style thing, take it or leave it: the nested loop rebuilding `authCopy` is four levels deep for "copy the map, drop one key". Since `sql.Map` is a plain map, `maps.Clone` plus a delete on the cloned `user` map says the same thing in about three lines. Current version is correct either way.

Prasad8830:
Hi @Rajkaran-122, I was reviewing your code, and I think you did not go through the discussion in the issue. 

There are some blocking items; the main one is scoping.

### 1. The change is not scoped to in-cluster contexts,
The agreed scope in the issue discussion was in-cluster only, explicitly said by @SuryanshGarg04:
> keep it scoped to **in-cluster contexts**. Sanitizing auth globally across all contexts risks colliding two distinct users/roles connecting to the same cluster. Out-of-cluster tokens don't rotate on a local filesystem anyway.

The patch keys on the *presence of* `auth.user.token` rather than `DeploymentType == "in_cluster"`, so it also applies to out-of-cluster kubeconfig contexts. It does reach them: `InternalKubeConfig.K8sContext` runs each user entry through `RecursiveCastMapStringInterfaceToMapStringInterface`, whose `ConvertMapInterfaceMapString` recurses into the `map[string]interface{}` branch, so the nested `user` map is `map[string]interface{}`, and the type assertion succeeds.

I confirmed the resulting collision. Two kubeconfig contexts, identical `prod` context / `prod-admin` user / same cluster, differing only in bearer token, built through the real yaml parsing path:

| | alice's token | bob's token |
|---|---|---|
| master | `8e1790952b77341495b279b6b051ea46` | `55b8b7390027a920d6ebe9449b1d3350` |
| with this PR | `0dd0271146dfc11caa44600d8a3d3f70` | `0dd0271146dfc11caa44600d8a3d3f70` |

`DeploymentType` on that context is `""`, and the parsed auth is `sql.Map{"name":"prod-admin", "user":map[string]interface{}{"token":...}}`.

This matters because `ConnectionPersister.SaveConnection` returns the *existing* row on an ID match. Two users on one Meshery instance importing the same context name with their own tokens now share a single connection, and the second import silently adopts the first user's connection and credential. `mesheryInstanceId` is per-install, not per-user, so it does not separate them.

Scoping on `DeploymentType` has one ordering trap worth knowing about: `NewK8sContext` generates the ID before `DeploymentType` is set, it is assigned in `DiscoverK8SContextFromKubeConfig` after the constructor returns, and `SaveK8sContext` recomputes the ID with the field populated. If the branch keys on `DeploymentType`, those two computations take different paths, and `ctx.ID` no longer matches the connection UUID. The field needs to be set inside `NewK8sContextFromInClusterConfig`, before ID generation.

### 2. Credential refresh is not implemented

The issue's expected behaviour is two-sided:

> Credential rotation should update authentication material associated with the existing connection rather than create a new logical connection solely because the token contents changed.

A stable ID satisfies the first half and makes the second half permanent. `SaveMesheryK8sContext` returns `ErrContextAlreadyPersisted` when the row exists, and `SaveK8sContext` discards that error (`_, _ =`), so `k8s_contexts.auth` keeps the token from the first discovery. On the local provider, this is more pronounced, since `DefaultLocalProvider.SaveConnection` ignores `conn.CredentialSecret` entirely, and the k8s token is only ever stored in `k8s_contexts.auth`.

The rotated token needs to be written back to the existing connection. `ReconcileK8sContextServerID` in the same file is the existing precedent for that shape; it self-heals a persisted connection through `UpdateConnectionById` and no-ops when the value is already current.

### 3. Status preservation is unaddressed

`SaveK8sContext` hardcodes `Status: connections.DISCOVERED`. Locally, `ConnectionPersister.SaveConnection` preserves the existing status, so that path is fine. On the remote provider, it is a POST upsert, and remote `LoadAllK8sContext` filters on `CONNECTED`, which is the case that drives repeated rediscovery in the first place. If the provider honours the payload status, a `DISCONNECTED` connection is moved back to `DISCOVERED` on each rediscovery, and the retry loop #20458 closes and reopens even with a stable ID. The issue lists "connection status is preserved appropriately" as a required assertion. Could you confirm the remote path?

### 4. Test coverage does not match what the issue asks for

The issue requests four assertions. Only the first is covered, and only at the hash level rather than through `SaveK8sContext` / `SaveConnection`:

1. Connection ID remains stable - partially covered
2. An existing connection is updated rather than duplicated - not covered
3. Connection status is preserved appropriately - not covered
4. Repeated discovery does not create additional logical connections - not covered

On the cases themselves:

- `"context without token"` has no `token` key, so the "simulate token rotation" block is a no-op. I ran that block verbatim against the fixture, and the auth map is byte-identical before and after, so the case compares an ID to itself.
- `"regular context with token"` and `"in-cluster context with token"` exercise the same single branch, differing only in string values.
- `wantSame` is never `false`, so no case pins that the ID *should* change when the cluster or name changes. That is the assertion that would have caught item 1.
- `"regular context with token"` asserting `wantSame: true` encodes the out-of-cluster behaviour from item 1 as the expected result, so the suite stays green after that issue is fixed. The tests pass today for that reason rather than because the behaviour is right.

Prasad8830:
Also, the PR body still has the template `This PR fixes #` with no number, and the sign-off checkbox is unchecked.

kunalKumar-13:
@Prasad8830's scoping point is correct, and it goes further than the note I left above. I checked the two mechanisms he describes:

**The nested `user` map really is `map[string]interface{}` on the kubeconfig path.** `server/models/k8s_context.go:137` runs the user entry through `RecursiveCastMapStringInterfaceToMapStringInterface`, so `auth["user"].(map[string]interface{})` succeeds for out-of-cluster contexts too. The branch is not in-cluster-only.

**And the ordering trap is real.** `DeploymentType` is assigned in the handler, not the constructor — `server/handlers/k8sconfig_handler.go:570` sets `in_cluster` and `:602` sets `out_of_cluster`, both after `NewK8sContext` has already generated an ID. So keying the branch on `DeploymentType` without also setting it inside `NewK8sContextFromInClusterConfig` would make the constructor and `SaveK8sContext` compute different IDs.

I want to correct my own review while I am here. I did flag that two service accounts on one cluster collapse to a single ID, but I filed it as "probably intended… a smaller judgement call" on the strength of the PR's own framing. That was too generous. Two users on one Meshery instance ending up on a shared connection and credential is not a judgement call, and I should have traced whether it reached out-of-cluster contexts rather than assuming the in-cluster framing held. @Prasad8830 did, and it does.

**One finding of mine that is still open and is not in his four**, so it does not get lost: every in-cluster connection that already exists hashes to a *new* ID the first time this ships.

```
existing token context, on upgrade:  e40a0aeff42c  ->  21a2acdb88f8
```

The old row is orphaned and a new one created — the same duplicate-connection symptom #21810 describes, once, on deploy. That is inherent to any fix here rather than a flaw in this one; the ID has to change once in order to stop changing forever. But it needs a migration or a release note, or the issue will look like it recurred a week after the fix lands.

@Rajkaran-122 — for what it is worth, the diagnosis underneath this is sound, and the two structural things I checked first both hold up: master already hashes `auth` at `k8s_context.go:396`, so the overwrite is correct, and `sql.Map` has no custom `MarshalJSON`, so swapping it for a plain map cannot shift the hash. The gap is scope rather than mechanism. Scoping to in-cluster, setting `DeploymentType` inside `NewK8sContextFromInClusterConfig` before ID generation, and adding a `wantSame: false` case would address most of what is above.

Rajkaran-122:
Thanks @Prasad8830 and @kunalKumar-13 for the detailed review and for tracing the kubeconfig path and ID-generation ordering.  @Prasad8830 @kunalKumar-13 PTAL .

kunalKumar-13:
Thanks for turning it around. Scoping by server URL rather than `DeploymentType` is a reasonable way to dodge the ordering trap — the field really is assigned after `NewK8sContext` has already generated an ID, so keying on it would have needed the extra change in `NewK8sContextFromInClusterConfig` first.

But I think the current check never fires, which would make the whole thing a no-op in exactly the case it targets.

`NewK8sContextFromInClusterConfig` builds the server at `server/models/k8s_context.go:16` of that function:

```go
host, port := os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT")
...
server := "https://" + net.JoinHostPort(host, port)
```

`net.JoinHostPort` always appends `:port`. The two literals compared against have no port, so the equality cannot hold:

| `KUBERNETES_SERVICE_HOST` | resulting `Server` | matches the check |
|---|---|---|
| `10.96.0.1` (the usual ClusterIP) | `https://10.96.0.1:443` | false |
| `172.20.0.1` (EKS) | `https://172.20.0.1:443` | false |
| `kubernetes.default.svc` | `https://kubernetes.default.svc:443` | false |
| `kubernetes.default` | `https://kubernetes.default:443` | false |
| `fd00::1` (IPv6) | `https://[fd00::1]:443` | false |

Two separate reasons, either of which is enough: the port is always there, and `KUBERNETES_SERVICE_HOST` is normally the ClusterIP rather than a DNS name — the kubelet injects the service IP, so `10.96.0.1` or equivalent is what you actually get.

Which means `isInCluster` is `false` for every in-cluster context, `data["auth"]` is never replaced, and the ID still changes on token rotation. The tests do not catch it because they construct `K8sContext` values directly rather than going through `NewK8sContextFromInClusterConfig`, so they can set whatever `Server` the check expects.

I do not think the answer is a smarter string match — the `Server` value is derived from environment variables that vary per cluster, so any pattern is guessing. @Prasad8830's suggestion looks right to me: set `DeploymentType` inside `NewK8sContextFromInClusterConfig` **before** the ID is generated, then branch on it. That makes the signal explicit rather than inferred, and it fixes the ordering problem at the same time, so `NewK8sContext` and `SaveK8sContext` compute the same ID.

A test that goes through `NewK8sContextFromInClusterConfig` with `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` set — rather than building the struct by hand — would have caught this and would keep catching it.

My earlier note about the one-time ID migration still stands separately: whatever the final scoping, existing in-cluster connections hash to a new ID once on upgrade, so it wants a migration or a release note.

Prasad8830:
Thanks for paying some attention, and @kunalKumar-13 for catching the URL-matching no-op.

The scoping problem is genuinely fixed. I verified out-of-cluster contexts differing only by token now produce distinct IDs (`d52aadb79d5de1ec…` vs `d27e107cb2f500ce…`), and in-cluster token rotation is stable. The `wantSame: false` case and its negative assertion are in too.

But the transient `IsInCluster` field reintroduces the duplicate-connection bug via a different path, so I think this needs to be moved back to `DeploymentType`.

### 1. `IsInCluster` is not persisted, and the identity function depends on it 

The field is declared `gorm:"-"` and `yaml:"-"`, so it never survives storage, while `K8sContextGenerateID` now branches on it. The identity of a context, therefore, depends on whether it was just constructed or loaded back.

The FSM path loads from persistence. `KubernetesMiddleware` puts `LoadAllK8sContext`'s result into `AllKubeClusterKey` (`server/handlers/middlewares.go:308` and `:338`), `K8sFSMMiddleware` reads it at `:401` and builds each machine context from it, and `DiscoverAction.Execute` calls `SaveK8sContext` (`server/machines/kubernetes/discover.go:56`), which recomputes the ID at `server/models/default_local_provider.go:513`. On that path, `IsInCluster` is the zero value, so the token is returned to the hash.

I ran that exact sequence against a SQLite-backed `DefaultLocalProvider`, in-cluster discovery, then FSM re-discovery from the reloaded row:

| | discovery ID | FSM ID | connection rows |
|---|---|---|---|
| master | `737ea9ca-7e43-2610-112f-d775b0cc8a46` | `737ea9ca-7e43-2610-112f-d775b0cc8a46` | 1 |
| this PR | `1cd884b2-5e22-885b-534a-43bc09d01bda` | `737ea9ca-7e43-2610-112f-d775b0cc8a46` | **2** |

No token rotation is involved. The FSM path loses the flag and falls back to the token-inclusive hash, so its ID equals the master's; `ConnectionPersister.SaveConnection` finds no existing row and inserts a second one. One cluster, two connection identities, on the next request after discovery.

The reloaded row in that run printed `DeploymentType="in_cluster"`. That field is persisted and already carries the answer, which is why I'd suggest going back to it:

- set `DeploymentType = "in_cluster"` inside `NewK8sContextFromInClusterConfig`, **before** the ID is generated, and
- branch `K8sContextGenerateID` on `DeploymentType == "in_cluster"`.

That fixes the ordering trap and the round-trip problem together, requires no new field, and requires no signature change to `NewK8sContext` / `NewK8sContextWithServerID`. Both are exported, and the added positional bool is an API break for anything outside this repo.

### 2. Nothing tests that the in-cluster wiring is connected

I flipped `true` to `false` in the `NewK8sContextFromInClusterConfig` call and re-ran your tests:

```
--- PASS: TestK8sContextGenerateID (0.00s)
--- PASS: TestNewK8sContextFromInClusterConfigTokenRotation (0.00s)
ok      github.com/meshery/meshery/server/models        0.041s
```

All green with the fix disabled. The whole change hangs on that one literal, and no test covers it.

`TestNewK8sContextFromInClusterConfigTokenRotation`'s comment says it "exercises the actual NewK8sContextFromInClusterConfig → NewK8sContextWithServerID → NewK8sContext → K8sContextGenerateID flow with realistic environment variables", but it constructs `K8sContext` values by hand with `IsInCluster: true` and never calls the constructor or sets any environment variable. That is the test @kunalKumar-13 asked for by name, and it is the one that would have caught both the URL-matching no-op and item 1 above.

To make it writable, the two hardcoded paths in `NewK8sContextFromInClusterConfig` (`tokenFile`, `rootCAFile`) need to be package-level `var`s rather than `const`s, so a test can point them at `t.TempDir()` and set `KUBERNETES_SERVICE_HOST` / `KUBERNETES_SERVICE_PORT`. Then assert the constructor's context yields the same ID before and after the token file is rewritten.

### 3. Still open from the earlier review

- **Credential refresh.** The rotated token still needs to be written back to the existing connection. `SaveMesheryK8sContext` returns `ErrContextAlreadyPersisted` and `SaveK8sContext` discards it (`_, _ =`), so `k8s_contexts.auth` keeps the first token forever. `ReconcileK8sContextServerID` is the precedent for that shape.
- **Status preservation.** `SaveK8sContext` hardcodes `Status: DISCOVERED`. Local is safe; the remote provider POSTs an upsert, and the remote `LoadAllK8sContext` filters on `CONNECTED`, which drives rediscovery. If the provider honours the payload status, a `DISCONNECTED` connection returns to `DISCOVERED` and #20458's loop reopens.
- **Test assertions 2–4 from the issue** (no duplicate, status preserved, idempotent discovery) are still uncovered. Item 1 above is exactly what assertion 2 would have caught.
- **CA data is still hashed.** Rotating only `certificate-authority-data` still changes the in-cluster ID (`18677077a24637bd…` becomes `4cb9cbc6d515f6e2…`). Same rotation class as the token. Hashing `{name, server, mesheryInstanceId}` for in-cluster covers both.
- **The one-time migration** @kunalKumar-13 raised is still open, and independent of the scoping mechanism.

Check out CodeRabbit's  **Linked Issues check**. It is still on `0fda2364` for the same reason:

> It does not implement or verify updating an existing connection, preserving connection status… Resolution: Implement or provide evidence for matching and updating existing connections after token rotation. Preserve connection status, prevent duplicate persisted records, and add regression tests for repeated discovery and connection persistence.

Prasad8830:
Always try to resolve coderabbit's suggestions

Prasad8830:
Hi @Rajkaran-122, are you working on these reviews?

Rajkaran-122:
> Hi @Rajkaran-122, are you working on these reviews? If not, please tell

yes sir .

Prasad8830:
> > Hi @Rajkaran-122, are you working on these reviews? 
> 
> yes sir .

thanks for your reply. I will review your PR

Prasad8830:
This is the right fix, thanks for moving it to `DeploymentType`.

I re-ran the round-trip probe from my last comment against `df29e0621b`:

| | discovery ID | FSM ID (rotated token) | connection rows |
|---|---|---|---|
| `0fda2364` | `1cd884b2-…` | `737ea9ca-…` | 2 |
| `df29e0621b` | `1cd884b2-…` | `1cd884b2-…` | **1** |

The duplicate is gone. Out-of-cluster contexts are still separate by token (`d72e97e9…` vs `e3a6a2b3…`), in-cluster rotation is stable, and the remote provider path is covered too, since `SaveK8sContext` writes `"deploymentType"` into metadata and the struct's json tag matches. `server/models` and `server/handlers` both pass.

### One gap: the wiring is still untested

You made it testable, `inClusterTokenFile` and `inClusterRootCAFile` are package vars now, but nothing uses them. I removed the `ctx.DeploymentType = "in_cluster"` assignment and its ID regeneration from `NewK8sContextFromInClusterConfig` and the whole suite still passed. So the entire fix still rests on one unguarded assignment. Here is the test that closes it. It passes on your branch and fails if that assignment is removed, which is the property you want:

```go
func TestInClusterConfigIDStableAcrossTokenRotation(t *testing.T) {
	mux := http.NewServeMux()
	mux.HandleFunc("/livez", func(w http.ResponseWriter, _ *http.Request) {
		_, _ = w.Write([]byte("ok"))
	})
	mux.HandleFunc("/version", func(w http.ResponseWriter, _ *http.Request) {
		_, _ = w.Write([]byte(`{"major":"1","minor":"29","gitVersion":"v1.29.0"}`))
	})
	mux.HandleFunc("/api/v1/namespaces/kube-system", func(w http.ResponseWriter, _ *http.Request) {
		_, _ = w.Write([]byte(`{"kind":"Namespace","apiVersion":"v1","metadata":{"name":"kube-system","uid":"aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"}}`))
	})
	srv := httptest.NewTLSServer(mux)
	defer srv.Close()

	host, port, err := net.SplitHostPort(srv.Listener.Addr().String())
	if err != nil {
		t.Fatal(err)
	}
	t.Setenv("KUBERNETES_SERVICE_HOST", host)
	t.Setenv("KUBERNETES_SERVICE_PORT", port)

	dir := t.TempDir()
	tokenPath := filepath.Join(dir, "token")
	caPath := filepath.Join(dir, "ca.crt")
	caPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: srv.Certificate().Raw})
	if err := os.WriteFile(caPath, caPEM, 0o600); err != nil {
		t.Fatal(err)
	}

	origToken, origCA := inClusterTokenFile, inClusterRootCAFile
	inClusterTokenFile, inClusterRootCAFile = tokenPath, caPath
	defer func() { inClusterTokenFile, inClusterRootCAFile = origToken, origCA }()

	log, _ := logger.New("t", logger.Options{Format: logger.JsonLogFormat})
	iid := core.Uuid(uuid.FromStringOrNil("11111111-2222-3333-4444-555555555555"))

	build := func(token string) *K8sContext {
		t.Helper()
		if err := os.WriteFile(tokenPath, []byte(token), 0o600); err != nil {
			t.Fatal(err)
		}
		kc, err := NewK8sContextFromInClusterConfig("in-cluster", &iid, log)
		if err != nil {
			t.Fatalf("NewK8sContextFromInClusterConfig: %v", err)
		}
		return kc
	}

	before := build("sa-token-v1")
	after := build("sa-token-v2")

	if before.DeploymentType != "in_cluster" {
		t.Errorf("constructor did not mark the context in-cluster: DeploymentType=%q", before.DeploymentType)
	}
	if before.ID != after.ID {
		t.Errorf("in-cluster ID changed across token rotation: %s -> %s", before.ID, after.ID)
	}
}
```

Imports needed: `encoding/pem`, `net`, `net/http`, `net/http/httptest`, `os`, `path/filepath`.

This is also roughly what CodeRabbit is asking for in its comment on `k8s_context_test.go:422`, though that one additionally wants the save-reload-rotate path through discovery, which would catch the duplicate regression from my last review as well. Both are worth having.

### Remaining open items

CodeRabbit's four new comments cover most of what is left, and I agree with all of them:

- **Credential refresh** (its comment on `k8sconfig_handler.go:609`): this is the one I raised as item 2 twice. Same mechanism: `ConnectionPersister.SaveConnection` returns the existing row unchanged, `DefaultLocalProvider.SaveConnection` never persists `CredentialSecret`, and `K8sContextFromConnection` then reads the stale credential through the existing `CredentialID`. The rotated token needs writing back, with `status` preserved.
- **CA data in the hash** (its outside-diff comment on `421-447`): rotating only `certificate-authority-data` still changes the in-cluster ID. Same rotation class as the token.
- **Legacy ID reconciliation** (its comment on `k8s_context.go:370`): this is @kunalKumar-13's migration point. Existing in-cluster rows were hashed with the token, so they are orphaned once on upgrade. Needs a migration or a release note.
- **MeshKit errors for the two `os.ReadFile` calls** (its comment on `k8s_context.go:324`): pre-existing, but the diff moved those reads, so worth fixing here per the repo's error convention.

Still nothing covering the issue's assertions 2–4 (existing connection updated, not duplicated, status preserved, repeated discovery idempotent). The credential-refresh work above is what assertions 2 and 3 would pin.

PARTH-TUSSLE:

Prasad8830:
> Hey could you add this as an agenda item for websites meet on Monday, we could discuss it there, you can find the link to the meeting minutes in the #websites channel on Slack.

Hi @PARTH-TUSSLE, does only the PR author (@Rajkaran-122) need to add it, or can I add it and discuss it in meet?

PARTH-TUSSLE:
> > Hey could you add this as an agenda item for websites meet on Monday, we could discuss it there, you can find the link to the meeting minutes in the #websites channel on Slack.
> 
> Hi @PARTH-TUSSLE, does only the PR author (@Rajkaran-122) need to add it, or can I add it and discuss it in meet?

No you can add it as well if there's something you want to discuss

SuryanshGarg04 (commented):
@Rajkaran-122 Please address the unresolved CodeRabbit reviews around credential refresh, legacy migration, testing gaps, and MeshKit errors

Also, one more issue: CA certificate rotation can still create duplicate connections.  `certificate-authority-data` is included in the in-cluster context hash, so rotating the CA changes the connection ID. Since CA data is mutable, it should be excluded from the ID for `in_cluster` contexts as well.

ritzorama:
Thanks for iterating on the in-cluster-only scoping, legacy mapping migration, and credential refresh. I inspected the constructor/ID-generation ordering, local persistence transaction, remote path, and regression tests. One remaining blocker is the duplicate error-code allocation: `meshery-server-1464` and `-1465` collide with existing codes, and the Error codes utility cannot export an unambiguous reference. Allocate unique next-reserved values, advance `server/helpers/component_info.json`, and regenerate the `docs/data/errorref/` output with `make error`.

Prasad8830:
Thanks for the push. The CA exclusion works, and the local legacy migration moves the old connection to the stable ID with its status intact. But there are regressions:

**1. Every local k8s save fails (blocking).** `DefaultLocalProvider.SaveConnection` sets `CredentialID` to `new(core.Uuid)` and never creates a credential, so the refresh always calls `GetCredentialByID(uuid.Nil)` and returns:
```
failed to fetch credential for refresh: credential with id 00000000-0000-0000-0000-000000000000 not found
```
This hits in-cluster and out-of-cluster contexts alike, and is likely why "Integration test with meshsync" has failed on every run since `0a623516`. Locally the auth lives in `k8s_contexts.auth`, so refresh that row instead.

**2. Remote refresh panics (blocking).** `UpdateUserCredential(nil, …)` -> `GetToken(req)` -> `req.Cookie()` on a nil request -> nil pointer dereference. FSM discovery runs in a goroutine with no `recover()`. Use the `token` that `SaveK8sContext` already receives.

**3. Error codes still collide (blocking).** `1484` is `ErrInitLoggerCode` and `1485` is `ErrModelNotFoundCode`. Use 1486/1487, bump `component_info.json`, and run `make error`.

**4. Migration leaves the old `k8s_contexts` row** with the expired token and a `connection_id` that no longer exists. Move or delete it in the same transaction.

**5. No test exercises the new code.** Nothing calls `SaveK8sContext`, the refresh, or the migration. The migration test is an empty function inside `/* */`, and in-memory sqlite already works in this package without extra setup. A save -> reload -> rotate test on a sqlite-backed `DefaultLocalProvider` would catch 1st point.

SuryanshGarg04:
@Rajkaran-122  Please respond to each comment

Rajkaran-122:
@Prasad8830 sir PTAL .

server/models/remote_provider.go
```diff
@@ -1150,6 +1156,33 @@ func (l *RemoteProvider) SaveK8sContext(token string, k8sContext K8sContext, add
...
+		// Preserve the existing credential's properties and update only the secret
+		updatedCredential := &Credential{
+			ID:     *connection.CredentialID,
+			Secret: conn.CredentialSecret,
+			UserId: existingCred.UserId, // Preserve the UserId for the update constraint
+			Name:   existingCred.Name,   // Preserve other fields
+			Type:   existingCred.Type,
+		}
+		// Use UpdateUserCredential to update the credential via the remote provider
+		_, updateErr := l.UpdateUserCredential(nil, updatedCredential)
```
Prasad8830: **The remote panic is still there.** `f33bcb9` says it fixed "remote refresh nil request panic", but it only touches `default_local_provider.go`, `error.go` and `component_info.json`. On the head commit, `remote_provider.go:1179` is still `UpdateUserCredential(nil, updatedCredential)` -> `GetToken(req)` -> `req.Cookie()`. A direct GORM update isn't right there either, since the remote provider has no DB, use the `token` that `SaveK8sContext` already receives.

Prasad8830:
Thanks, the 1st regression is gone (saves work again). The error codes are currently (1486/1487 + `component_info.json`), and the migration updates `k8s_contexts.connection_id` in the transaction. Three things remain:

***Credential refresh*** is now a no-op on local, which is the point of the issue.** Guarding on `uuid.Nil` removes the crash, but local `CredentialID` is *always* zero, so the refresh never runs. Rotating the token and re-saving leaves the old one stored:
```
stored token after rotation = tok-v1   (want tok-v2-ROTATED)
credential rows in local DB = 0
```
Local k8s auth lives in `k8s_contexts.auth`, not in a credential row, so the refresh needs to update that row. `SaveMesheryK8sContext` returns `ErrContextAlreadyPersisted` and the caller discards it, so nothing writes the new token today.

***Migration still leaves a stale context row.*** The dangling `connection_id` is fixed, but the legacy row survives with the expired token and now points at the live connection:
```
k8s_context 5a1e476c  connection_id=b464b654  token=tok-v1-EXPIRED  connExists=true
k8s_context b464b654  connection_id=b464b654  token=tok-v2          connExists=true
connections=1  k8s_contexts=2
```
`LoadAllK8sContext` returns every row locally, so the expired-token context keeps being loaded. Delete the legacy row in the same transaction, or carry its ID over instead of inserting a second one.

**Still no tests on any of this.** latest commit adds nothing; the migration test is an empty function inside `/* */`; and nothing calls `SaveK8sContext`, ' Refresh', or `NewK8sContextFromInClusterConfig`. That's why CI is green while 2 and 3 hold.

Rajkaran-122:
@Prasad8830 sir PTAL.

Prasad8830:
The local side looks right now, and I verified it end-to-end: the stored token updates after rotation on both the reloaded and the fresh-discovery paths, and an upgraded install ends with one connection and one context row carrying the fresh token, status preserved. Using `k8s_contexts.auth` and deleting the legacy row is the correct approach. 
Three things left:
**1. The remote panic is still unfixed (blocking).** `remote_provider.go:1179` is still `UpdateUserCredential(nil, updatedCredential)` -> `GetToken(req)` -> `req.Cookie()` on a nil request. `f33bcb9`'s message said this was fixed, but only changed the local provider, and this commit states remote behaviour is unchanged. It needs the `token` that `SaveK8sContext` already receives. This is the one remaining item that affects Meshery Cloud.
**2. `updated_at` is wiped on refresh.** `UpdateMesheryK8sContext` writes `"updated_at": mkc.UpdatedAt`, which is nil on the real in-cluster path, since `NewK8sContextFromInClusterConfig` builds a fresh context each discovery:
```
fresh ctx UpdatedAt = <nil>
updated_at   = NULL
```
`GetMesheryK8sContexts` orders by `updated_at desc` by default, so drop that key and let gorm's auto-timestamp set it.
**3. `TestUpdateMesheryK8sContext` never calls `UpdateMesheryK8sContext`.** It constructs two structs and asserts they differ, including `rotatedCtx.ID != originalCtx.ID` on two values it just set equal. So it passes regardless of what the function does. The migration test is also still an empty body inside `/* */`.

Prasad8830:
On the CGO note in your commit message: CI runs with CGO enabled, and this package already uses `gorm.io/driver/sqlite` in-memory elsewhere, so a real database test will run in CI even if it can't run on your machine.

PR #21863, opened 2026-09-09
[mesheryctl] Add traceability tags to Half A BATS tests

Zayden369:
**Notes for Reviewers**

- This PR addresses **Half A** of #21845.
- It covers `001-system`, `002-model`, `004-perf`, and `005-component`—69 previously untagged tests.

### Changes

- Added `[cut=...]` and `[tg=...]` traceability tokens to all 69 tests.
- Used command-specific Test Groups such as `System Checks`, `Model Import`, `Performance Profile`, and `Component View`.
- Preserved the two existing Connection Lifecycle tags.
- Omitted new `[TC-n]` tokens because corresponding Test Plan rows were not verified.

### Validation

- `node --test bats-to-allure.test.js`: **27/27 passed**
- All 69 new titles emit `componentUnderTest` and `testGroup` labels
- BATS counts remain unchanged:
  - `001-system`: 16
  - `002-model`: 36
  - `004-perf`: 6
  - `005-component`: 13
- `git diff --check`: passed

## Summary by CodeRabbit

- **Tests**
  - Added consistent metadata tags to end-to-end test names across system management, model operations, performance workflows, and component commands.
  - Improved categorization for system checks, dashboards, tokens, model lists, imports, exports, searches, views, generation, deletion, and performance results.
  - Test commands, assertions, and runtime behavior remain unchanged.

- **Bug Fixes**
  - Ensured Allure reports consistently preserve the converter’s framework, language, and project labels.


yi-nuo426:
How do these tags show up in the https://qa.meshery.io/ reports?

yi-nuo426:
@MU5A, please review.

Zayden369:
@yi-nuo426 Thanks for asking. The converter removes the tokens from the displayed test name and maps them to Allure labels:

* `[cut=System]` → `componentUnderTest: System`
* `[tg=System Help]` → `testGroup: System Help`

For example, the report will display the clean name `mesheryctl system help display help`. Opening that result and selecting **Labels** will show the two traceability labels alongside `framework: bats`, `project: mesheryctl`, and `client: CLI`.

These additional labels will appear after this PR is merged and a fresh BATS run is published. The current Mesheryctl report does not group its result tree by these labels, so a separate `meshery/qa` configuration change would be needed if we want visible Component → Test Group sections.
If this grouping would be useful, I’d be happy to create a separate issue for the `meshery/qa` configuration change and work on it after the scope is confirmed.

Zayden369:
@yi-nuo426 @MU5A Please review my pr and if there,s is any anamolies u can tell me, I,ll fix that. The half B is already reviewed but half A is still standalone.

MU5A:
Read through the label-locking change and its test. The approach is right: `resultLabels` only force-overwrites the three converter-owned names (`framework`, `language`, `project`) after the merge, so the existing last-value-wins behavior for everything else (like `client`) is untouched. The new test exercises both ways a result could have been mislabeled, an `ALLURE_LABELS` env override and a `[project=...][framework=...]` title token, in a single case, and asserts exactly one value survives for each. That's the right level of coverage for a fix like this.

Spot-checked a sample of the 69 tag additions against the source `.bats` files, consistent with the pattern, and the unchanged per-file counts (16/36/6/13) plus 27/27 unit tests passing back that up.

leecalcote (approved):
Let's see how well this works...

leecalcote:
@MU5A, @yi-nuo426, thanks much for these reviews. 👍

yi-nuo426:
Thanks for preserving the converter's provenance labels while adding the new test traceability metadata. mesheryctl/bats-to-allure.js now canonicalizes framework, language, and project after deduplication, so neither ALLURE_LABELS nor title tokens can misclassify these BATS results; the added unit test verifies both override paths. The new cut and test-group prefixes are consumed by the existing title-token parser without changing displayed test names. No changes requested.

leecalcote:
@Zayden369, no responses from you on any of your opened PRs. But, we did receive your special request for review of them in Slack. 🤔

Zayden369:
Thank you @leecalcote, @MU5A, and @yi-nuo426 for reviewing and merging this contribution. Your feedback helped ensure that the new traceability tags work correctly without allowing environment variables or title tokens to override the converter-owned labels. I learned a lot while working through the review.

Zayden369:
Thanks for pointing this out, @leecalcote. I’ve only recently started contributing to open source, and I mistakenly assumed that once a PR was merged, no further response was expected from me. I also thought that posting a separate thank-you message after the merge might be unnecessary, so I didn’t follow up.

I now understand that acknowledging reviews and responding to the latest comments is an important part of the contribution process. I’ll make sure to follow up on GitHub and address the feedback there before requesting reviews in Slack. Thank you for helping me understand the expected workflow.

PR #21879, opened 2026-09-09
fix(server): return 404 for missing connection in GetConnectionByID

### LINKED ISSUE #21094: [Server] Misleading ErrQueryGet remediation on connection GET failures + verify remote-provider delete of a never-issued UUID (TC-1066)

paranik-jpg:
**Fixes #21094**

### Description
Refactors `GetConnectionByID` to properly handle non-existent connection IDs by evaluating `models.ErrResultNotFoundCode`, `providerStatus`, and `http.StatusNotFound` to return an explicit HTTP 404 instead of misclassifying the failure as `ErrQueryGet` (which incorrectly suggested invalid query parameters).

### Key Changes
- Created `ErrGetConnection` (`meshery-server-1486`) in `server/handlers/error.go`.
- Added missing `return` statement in the 404 error branch within `GetConnectionByID` in `server/handlers/connections_handlers.go`.
- Synced error code metadata in `server/helpers/component_info.json` using `errorutil`.
- Added regression unit tests in `server/handlers/get_connection_test.go` covering both 404 Not Found and 200 OK paths.

### Testing
Passed unit tests:
```bash
go test -v -run "TestGetConnection" ./server/handlers/

## Summary by CodeRabbit

* **Bug Fixes**
  * Retrieving a connection now returns a clear not-found response when the connection does not exist.
  * Other connection retrieval failures preserve the provider’s status while providing more specific error details.

### CODE REVIEW

**Review by kunalKumar-13 — COMMENTED** [review]
Checked the things that usually go wrong when a PR allocates a new error code, and they're all clean:

- `meshery-server-1486` is free — no occurrences on master, and the highest currently allocated is 1485
- no duplicate codes in `server/handlers/error.go`
- `providerStatus()` exists (`server/handlers/utils.go:72`)
- `models.ErrResultNotFoundCode` exists (`server/models/error.go:113`, `meshery-server-1307`)

The main thing I wanted to flag is the three-way condition, because it reads as redundant and isn't:

```go
if errors.GetCode(err) == models.ErrResultNotFoundCode
   || providerStatus(err) == http.StatusNotFound
   || statusCode == http.StatusNotFound {
```

Each clause covers a different provider shape:

1. `ErrResultNotFoundCode` is the **local** provider's own "record does not exist" code — its database, no upstream involved.
2. `providerStatus(err)` reads a status *embedded in the error* by a **remote** provider, falling back to 502 when there isn't one.
3. `statusCode` is the separate return value from `provider.GetConnectionByID(...)`, which a provider can set without tagging the error at all.

A local not-found only reaches (1). A remote provider that tags its error reaches (2). One that returns the status positionally reaches (3). Drop any one and a real case goes back onto the 500/502 path, which is the bug being fixed. Worth keeping as is, and maybe worth a one-line comment so nobody tidies it later.

The other improvement here is easy to miss reading the diff: master logged `ErrQueryGet(obj)` — generic, no ID — and then wrote a *different* error object to the response. This builds one wrapped error carrying the connection ID and both logs and returns that. Same object in the log and the body, and the ID is in both now.

One question rather than a request: `h.log.Warnf` for the not-found path takes it down from error to warning, which I think is right — a client asking for a row that isn't there isn't a server fault. But the rest of this handler uses `h.log.Error` throughout, so it's worth confirming that's the convention you want rather than an inconsistency someone flags later.

kunalKumar-13:
Following up on my review above with something I only spotted while reading another PR — #21833 also allocates `meshery-server-1486`:

```
#21879 (this one)  server/handlers/error.go   ErrGetConnectionCode                       = "meshery-server-1486"
#21833             server/models/error.go     ErrInvalidEnvironmentConnectionsFilterCode = "meshery-server-1486"
```

Both bump `server/helpers/component_info.json` from `next_error_code: 1486` to `1487`, so whichever merges second gets a conflict on that file. That part is fine — it forces a look.

The trap is that resolving the JSON conflict on its own leaves two different errors sharing 1486, and as far as I can tell nothing validates uniqueness. So if this one lands second, it'd want `meshery-server-1487` and the counter at 1488.

Nothing to change right now — just flagging it so it doesn't get resolved as a plain file conflict. Same note left on #21833.

Gurkaran18:
@paranik-jpg , This PR and #21833 both claim `meshery-server-1486`, and both bump `next_error_code` to 1487 — so whoever merges second gets a conflict on `component_info.json`, and resolving just that leaves two errors sharing 1486. `make error` won't catch the duplicate.

#21833 is the older PR, so could you take 1487 here — `ErrGetConnectionCode = "meshery-server-1487"`, `next_error_code: 1488`, and regenerate `docs/data/errorref/` in the same commit? Then the conflict is just "take the
higher counter".

server/handlers/connections_handlers.go
```diff
@@ -395,8 +395,14 @@ func (h *Handler) GetConnectionByID(w http.ResponseWriter, req *http.Request, _
 	obj := "connection"
 
 	if err != nil {
-		h.log.Error(ErrQueryGet(obj))
-		writeMeshkitError(w, ErrQueryGet(obj), statusCode)
+		if errors.GetCode(err) == models.ErrResultNotFoundCode || providerStatus(err) == http.StatusNotFound || statusCode == http.StatusNotFound {
```
Omkar-Ugal, line 398: why this validation?

paranik-jpg: Different provider implementations return not-found errors differently here:

errors.GetCode(err) == models.ErrResultNotFoundCode: catches missing records from the local provider (direct DB lookup via MeshKit).

providerStatus(err) == http.StatusNotFound: extracts a 404 embedded inside an error object returned by a remote provider.

statusCode == http.StatusNotFound: handles providers that pass the HTTP status back positionally instead of wrapping it in the error.

Without all three, certain provider setups fall through to a 500/502 instead of returning the expected 404.

paranik-jpg:
Sounds good @Gurkaran18 @kunalKumar-13. I’ll rebase on master, bump this code to 1487 (with next_error_code at 1488), and update the error docs. Pushing the update shortly.

server/handlers/connections_handlers.go
Omkar-Ugal: did u check server code?

ritzorama:
This is a clear improvement to missing-connection semantics: the handler now preserves a useful connection ID in the structured error and distinguishes absence from query failures. The focused local tests are helpful; a remote-provider regression case for embedded/positional 404 statuses would further exercise the other two branches.

paranik-jpg:
Thanks @ritzorama! Glad the structured error ID and local test coverage make sense. Definitely down to follow up with mock tests for the remote-provider edge cases once this lands.

PR #21885, opened 2026-09-10
fix(mesheryctl): auto-page lists when stdout is not a TTY

PARZIVAL7498:
## Summary
- When stdout/stdin are not an interactive terminal, list pagination advances through all pages instead of stopping after page 1 or failing on `/dev/tty`.
- Skip `ClearLine` / clear-screen sequences on non-TTY so pipes and redirects are not wiped between pages.
- Leaves the interactive table path unchanged. Related to #20979 (AXI non-interactive slice).

## Test plan
- [ ] Unit: `go test ./internal/cli/pkg/display/ -run TestListPageHandlerAutoPagesWithoutTerminal`
- [ ] Interactive TTY: `mesheryctl connection list` still pages on keypress as before
- [ ] Non-TTY: `mesheryctl connection list </dev/null | cat` (or pipe) completes without hang and prints beyond page 1 when results exceed one page
- [ ] Redirected output has no clear/cls / ClearLine control noise between pages

## Summary by CodeRabbit

* **Bug Fixes**
  * Paginated command output now continues automatically in non-interactive environments, allowing scripts and CI jobs to receive complete lists.
  * Redirected and piped output no longer includes terminal-clearing sequences.
  * Interactive terminal pagination continues to clear the screen between pages as expected.

* **Tests**
  * Updated pagination coverage to verify automatic advancement without a terminal.

PR #21896, opened 2026-09-10
[mesheryctl] Fail rather than pick index 0 when stdin is not interactive

### LINKED ISSUE #21894: [mesheryctl] design deploy/apply silently pick index 0 when stdin is not interactive

### Current Behavior

When `mesheryctl design deploy <name>` or `mesheryctl design apply <name>` matches more than one design, the CLI prints the matches and asks which index to use. With no interactive stdin — a CI job, a pipe, `</dev/null`, a script — it **selects index 0 and deploys it** without any confirmation.

It is not a fallback after retries. It returns on the first iteration:

```go
// mesheryctl/internal/cli/root/design/deploy.go:198
for {
    fmt.Printf("Enter the index of design: ")
    response, err := reader.ReadString('\n')
    if err != nil {
        utils.Log.Info(err)          // EOF logged, not returned
    }
    response = strings.ToLower(strings.TrimSpace(response))
    index, err := strconv.Atoi(response)
    if err != nil {
        utils.Log.Info(err)          // parse failure logged, not returned
    }
    if index < 0 || index >= len(profiles) {
        utils.Log.Info("Invalid index")
    } else {
        return index                 // returns 0
    }
}
```

At EOF `ReadString` gives `("", io.EOF)` and `strconv.Atoi("")` gives `(0, err)`. Both errors are logged and execution continues. `0` is a *valid* index for any non-empty list, so the bounds check passes and the function returns immediately. Every subsequent read would also be EOF, so the loop cannot recover either.

Simulating the loop body:

```
EOF, 2 matching designs  -> returns index 0   (1 iteration)
EOF, 5 matching designs  -> returns index 0   (1 iteration)
stdin = "3\n"            -> returns index 3   (works normally)
```

**Both copies have it.** The two functions are near-identical:

| | |
|---|---|
| `deploy.go:184` `multiplepatternsConfirmation` | called from `deploy.go:128` |
| `apply.go:267` `multiplePatternsConfirmation` | called from `apply.go:107` |

### Desired Behavior

A non-interactive run should not deploy a design nobody selected. Either:

- return an error when `ReadString` fails, so the command exits non-zero with something like *"multiple designs match %q; re-run with a more specific name or an explicit --id"*, or
- distinguish "no input available" (EOF) from "input was unparseable" (retry is meaningful) and only loop for the second.

Worth noting the same read pattern appears at `mesheryctl/pkg/utils/helpers.go:570` and `:794`, though those take the opposite approach — see #21259, which reports them calling `Log.Fatal(err)` and killing the process. The two issues are the mirror image of each other: the helpers over-react to a stdin error, these two under-react. Both come from not separating "there is no user" from "the user typed something wrong", so a shared helper might be the right answer to both.

### Screenshots/Logs

```
$ mesheryctl design deploy "My Design" < /dev/null
Index: 0
Name: My Design
ID: 7b3e...
---------------------
Index: 1
Name: My Design (copy)
ID: 91af...
---------------------
Enter the index of design: EOF
strconv.Atoi: parsing "": invalid syntax
<deploys index 0>
```

### Environment

- Host OS: Mac Linux Windows
- Meshery Server Version: master
- Meshery Client Version: master
- Platform: Docker Kubernetes

---

Found while reviewing #21847. The discussion there — with @akkki007, @SuryanshGarg04 and @peaush07 — settled on dropping a `strings.TrimSpace` from that PR because widening a search increases the chance of hitting this path. That was the right call for that PR, but this hazard is on `master` today and is reachable by any name matching two designs, so it needs its own fix. Opening at @akkki007's suggestion.

I'm happy to take this one if nobody else has it.

kunalKumar-13:
**Notes for Reviewers**

- This PR fixes #21894

`design deploy <name>` and `design apply <name>` list the matches when a name is ambiguous and ask for an index. With no interactive stdin — a script, a pipe, a CI job, `< /dev/null` — they selected index 0 and carried on.

It was not a fallback after retries. It returned on the first iteration:

```go
response, err := reader.ReadString('\n')
if err != nil {
    utils.Log.Info(err)     // EOF logged, not returned
}
index, err := strconv.Atoi(response)
if err != nil {
    utils.Log.Info(err)     // parse failure logged, not returned
}
if index < 0 || index >= len(profiles) {
    utils.Log.Info("Invalid index")
} else {
    return index            // returns 0
}
```

`ReadString` at EOF gives `("", io.EOF)`; `strconv.Atoi("")` gives `(0, err)`. Both are logged and execution continues, and `0` is a valid index for any non-empty list, so the bounds check passes. Every later read is EOF as well, so the loop cannot recover.

Both commands had it — `multiplepatternsConfirmation` (deploy.go) and `multiplePatternsConfirmation` (apply.go) are near-identical.

### What changed

The two kinds of failure are now separated, because they want opposite handling:

- **read error** → there is nobody to ask, so stop. Returns `mesheryctl-1257`, naming the design and how many matched, and suggesting the design ID instead.
- **unparseable or out-of-range answer** → a real user is there and retrying is the point of the prompt, so it loops. The bounds message now says which range is valid rather than just "Invalid index".

Both functions return `(int, error)`. Both call sites are already inside a `RunE`, so they just propagate.

### Testing

```
go build ./...        clean
go vet                clean
gofmt -l              clean
go test -race         ok
```

New tests in `multiple_designs_selection_test.go` swap `os.Stdin` for a pipe, which is what lets the real functions be driven without refactoring them:

- both commands, at 2 and 5 matches, with an immediate EOF
- a valid answer at the first, middle and last index
- whitespace around an answer
- retries past unparseable, out-of-range and negative input
- input that runs out mid-retry

**Mutation-checked.** Restoring the fall-through makes exactly the EOF tests fail:

```
multiple_designs_selection_test.go:95: expected an error with no interactive
    stdin, got index 0 and nil
```

and nothing else fails.

### Notes

The error constant is named `ErrDesignSelectNotInteractiveCode` so it fits the existing alignment in `error.go` — otherwise `gofmt` re-indents the whole `const` block and the diff there becomes 30 changed lines instead of 13 added ones.

`mesheryctl-1257` was picked rather than the counter's `1255`, because #21748, #21678 and #21521 all already claim 1255. 1257 is unclaimed by any open PR as of opening this. Same pattern as the collision between #21833 and #21879 over `meshery-server-1486` — the shared `next_error_code` counter means concurrent PRs reliably pick the same number, which might be worth its own issue.

Found while reviewing #21847, and opened at @akkki007's suggestion.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved design and deployment selection when multiple designs share a name.
  - Invalid, negative, or out-of-range selections now prompt users to try again instead of selecting an unintended design.
  - Whitespace-padded responses and exhausted input are handled more reliably.
  - Non-interactive input now stops the operation with clear guidance to use a design ID, run interactively, or choose a unique name.


kunalKumar-13:
Pushed `cd36db321` — this now uses the helper that already exists rather than inferring the same fact from a read error.

`utils.IsInteractiveTerminal` (`pkg/utils/interactive.go`) checks stdin and stdout with `term.IsTerminal`, and #21885 is already using it for the paging path. I had missed it and reached for EOF detection instead, which was reinventing it.

Asking up front is also better behaviour: the command stops **before** printing the list of matching designs rather than printing them and then failing. In a CI log that is one error line instead of a wall of candidates nobody chose between.

**The read error still needs handling, and mutation testing is what showed why.** The TTY check passing only means a terminal was attached when the prompt started — the stream can still close under it on Ctrl-D. Removing the EOF branch does not fail a test, it **hangs the suite**: `Atoi("")` fails, `continue` goes back to a read that returns EOF again, forever.

Removing the *TTY* check did not fail anything either — which was a gap in my tests rather than evidence it was redundant. The EOF path caught the same cases, so the two mechanisms were indistinguishable. `TestMultipleDesignsSelectionChecksTerminalBeforeReading` closes that by putting a valid answer on stdin with no terminal:

```
--- FAIL: TestMultipleDesignsSelectionChecksTerminalBeforeReading
    expected an error without a terminal even with input available, got index 2
```

Only the up-front check can stop that case.

Since `IsInteractiveTerminal` is a `var`, the tests now say which environment they describe through a small `withTerminal` helper rather than inferring it from the pipe they built — which is what the comment on the var suggests it is there for.

`go build`, `go vet`, `gofmt -l`, `go test` and `go test -race` all clean.

---

@yi-nuo426 — this overlaps #20979 item (4), and I left a note there about a wrinkle worth folding into the design: this prompt did not *hang* on a non-TTY, it silently returned index 0 and deployed. A hang is visible to an operator; a silent selection is not. Whatever the shared handling ends up being, the prompt sites want to fail rather than fall back to a default.

ritzorama:
Great safeguard against an especially risky silent default. Please log the new MeshKit error before returning it so operators receive its full diagnostic details, for example:\n\n```go\nif err != nil {\n    utils.Log.Error(err)\n    return err\n}\n```\n\nAlso regenerate `docs/data/errorref/mesheryctl_errors_export.json`; the new `mesheryctl-1257` code is currently absent.

PR #21903, opened 2026-09-11
fix(server): reject malformed event filters

### LINKED ISSUE #21878: [Server] Return 400 for malformed event filter query parameters
### Current Behavior

The `GET /api/system/events` endpoint silently ignores malformed JSON in event filter query parameters.

For example, when the following request is made:

`GET /api/system/events?category=not-json`

the `getEventFilter` function attempts to parse the `category` parameter as JSON. The parsing fails, but `GetAllEvents` only logs the error and continues processing the request.

The request can then continue to `provider.GetEvents(...)` with an incomplete filter. If the provider request succeeds, the endpoint can return `200 OK` instead of rejecting the malformed input.

### Expected Behavior

Malformed event filter query parameters should result in a `400 Bad Request` response.

The provider should not be called when the event filter cannot be parsed.

For example:

`GET /api/system/events?category=not-json`

should return `400 Bad Request`.

### Screenshots/Logs

Relevant execution path:

GET /api/system/events
→ GetAllEvents
→ getEventFilter
→ json.Unmarshal(category)
→ unmarshal error
→ error is logged but not returned
→ provider.GetEvents(...)
→ potentially 200 OK

Relevant file:

`server/handlers/events_streamer.go`

Functions:

- `GetAllEvents`
- `getEventFilter`

### Environment

- **Host OS:** Windows
- **Platform:** Local development environment
- **Meshery Server Version:** master (local build)
- **Meshery Client Version:** master (local build)

### To Reproduce

1. Start Meshery Server locally from the `master` branch.
2. Send a GET request to the events endpoint with an invalid JSON filter:
   `GET /api/system/events?category=not-json`
3. Observe that the malformed filter error is logged but request processing continues.
4. The request can proceed to the provider layer instead of returning `400 Bad Request`.

### Technical Details

`getEventFilter` returns an error when the event filter query parameter cannot be unmarshalled.

However, `GetAllEvents` currently logs the error and continues:

`filter, err := getEventFilter(req)`

`if err != nil {`
`    h.log.Warn(err)`
`}`

This allows an invalid filter to reach the provider layer.

A regression test should verify that:

1. A malformed event filter returns `400 Bad Request`.
2. The provider's `GetEvents` method is not called.
3. The parsing error is propagated to the HTTP response.

This issue concerns malformed JSON in query parameters for `GET /api/system/events`.

k-anushka14:
**Notes for Reviewers**

- This PR fixes #21878.
- Added a regression test for malformed event filter query parameters.
- `GET /api/system/events?category=not-json` now returns `400 Bad Request`.
- The provider is not called when the event filter cannot be parsed.
- The existing Meshery error handling and error code are preserved.

### Changes

Previously, `GetAllEvents` logged the error returned by `getEventFilter` and continued processing the request. This allowed the provider to be called with an incomplete filter and could result in a successful response for malformed input.

This PR updates `GetAllEvents` to:

- Return `400 Bad Request` when event filter parsing fails.
- Use the existing `writeMeshkitError` error response.
- Preserve the existing `models.ErrUnmarshal` error and Meshery error code.
- Stop request processing before the provider is called.

A regression test was added to verify the malformed `category` filter behavior and ensure that `GetEvents` is not called.

### Testing

```text
go test ./server/handlers -run '^TestGetAllEvents_MalformedCategoryFilter$' -count=1
ok github.com/meshery/meshery/server/handlers 0.118s

go test ./server/handlers
ok github.com/meshery/meshery/server/handlers 4.611s


## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid event filters for category, action, severity, and acted-upon status now return a clear `400 Bad Request` response.
  - Error responses identify the specific filter that is invalid, helping users correct malformed requests.
  - Event retrieval is not attempted when any supported event filter is malformed.

### DIFF
--- a/server/handlers/events_streamer.go
+++ b/server/handlers/events_streamer.go
@@ -55,7 +55,9 @@ func (h *Handler) GetAllEvents(w http.ResponseWriter, req *http.Request, prefObj
 	// eventCategory :=
 	filter, err := getEventFilter(req)
 	if err != nil {
-		h.log.Warn(err)
+		h.log.Error(err)
+		writeMeshkitError(w, err, http.StatusBadRequest)
+		return
 	}
 	filter.Limit = limit
 	filter.Offset = offset
--- a/server/handlers/events_streamer_test.go
+++ b/server/handlers/events_streamer_test.go
@@ -3,14 +3,22 @@ package handlers
 import (
 	"bytes"
 	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
 
+	"github.com/gofrs/uuid"
 	"github.com/meshery/meshery/server/meshes"
+	"github.com/meshery/meshery/server/models"
 	"github.com/meshery/meshkit/logger"
+	"github.com/meshery/meshkit/models/events"
 	_events "github.com/meshery/meshkit/utils/events"
+	"github.com/meshery/schemas/models/core"
 )
 
 const testTimeout = time.Second
@@ -79,6 +87,16 @@ func newTestLogger(t *testing.T) logger.Handler {
 	return log
 }
 
+type eventsSpyProvider struct {
+	*models.DefaultLocalProvider
+	getEventsCalled bool
+}
+
+func (p *eventsSpyProvider) GetEvents(_ string, _ *events.EventsFilter, _ int, _ core.Uuid, _ core.Uuid) (*models.EventsResponse, error) {
+	p.getEventsCalled = true
+	return &models.EventsResponse{}, nil
+}
+
 func waitForPayload(t *testing.T, ch <-chan []byte, name string) []byte {
 	t.Helper()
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by ritzorama** [comment]
Nice focused hardening. I verified malformed category JSON now produces a MeshKit JSON 400 before `GetEvents` is reached, and the regression test pins both outcomes. Extending the test table to the other parsed filters would provide additional contract coverage. Checks are green.

File: server/handlers/events_streamer.go [hunk 1]
```diff
@@ -55,7 +55,9 @@ func (h *Handler) GetAllEvents(w http.ResponseWriter, req *http.Request, prefObj
 	// eventCategory :=
 	filter, err := getEventFilter(req)
 	if err != nil {
-		h.log.Warn(err)
+		h.log.Error(err)
```
hiyach28: this 400 is caused by bad client input, but the log level went from `Warn` to `Error`. Any client sending a bad query string will now produce server error logs. Consider keeping `h.log.Warn(err)` here and saving `Error` for real server-side failures.

server/handlers/events_streamer_test.go
```diff
@@ -102,6 +120,41 @@ func assertNoPayload(t *testing.T, ch <-chan []byte, name string) {
...
 
+func TestGetAllEvents_MalformedCategoryFilter(t *testing.T) {
+	systemID := core.Uuid(uuid.Must(uuid.NewV4()))
+	provider := &eventsSpyProvider{DefaultLocalProvider: &models.DefaultLocalProvider{}}
+	h := &Handler{
+		SystemID: &systemID,
+		log:      newTestLogger(t),
+	}
+	user := &models.User{ID: uuid.Must(uuid.NewV4())}
+	req := httptest.NewRequest(http.MethodGet, "/api/system/events?category=not-json", nil)
```
hiyach28: Optional: only `category` is tested, but `action`, `severity` and `actedUpon`/`acted_upon` go through the same code path. A small table-driven test over all four, as @ritzorama suggested, would stop one of them from quietly going back to "log and continue".

hiyach28 (approved):
LGTM. This fixes the cause of #21878. `getEventFilter` already returned an error for bad `category`/`action`/`severity`/`actedUpon` values, and the handler was ignoring it. It now returns a 400 through `writeMeshkitError` with `ErrUnmarshal` and never calls the provider. The UI always sends these filters as JSON arrays, so existing clients won't start getting 400s. The new test would fail against the old code. CI is green.

I left two optional inline suggestions. Neither blocks the merge.

k-anushka14:
Thanks for the review @ritzorama @hiyach28 ! I’ll address both suggestions:

* Keep the malformed filter logging at `Warn` level since this is caused by invalid client input.
* Extend the regression test into a table-driven test covering `category`, `action`, `severity`, and `actedUpon`/`acted_upon`.

I’ll make these changes and raise a follow-up PR.

PR #21939, opened 2026-09-11
[Server] Decode the workspace list filter once, fixing the unreachable owner filter

### LINKED ISSUE #21897: [Server] Workspace list endpoints parse `filter` with two grammars, leaving `owner` unreachable
### Current Behavior

`WorkspacePersister`'s four workspace-scoped list endpoints hand the **same `filter` query parameter to two parsers with incompatible grammars**:

1. `json.Unmarshal` into `WorkspaceFilter` — the JSON object grammar (`{"assigned":false}`)
2. `utils.ApplyFilters(query, filter, []string{"owner", "organization_id"})` — which does `strings.Split(filter, " ")` and reads a space-separated `"key value"` pair

Each grammar disables the other:

- **JSON input** (what Meshery UI actually sends) reaches `ApplyFilters` as a single token that matches no dynamic key, so it is a **silent no-op** — the declared `owner` and `organization_id` filters are **unreachable**.
- **`ApplyFilters` input** (`owner <uuid>`) fails the JSON decode first, and the raw `encoding/json` message is returned to the caller.

This is the same defect fixed for `GetEnvironmentConnections` in #21826 / #21833, still present in four more functions.

Affected functions — `server/models/workspace_persister.go`:

| Function | Line | `json.Unmarshal` | `ApplyFilters` | Table |
|---|---|---|---|---|
| `GetWorkspaceEnvironments` | 281 | 291 | 322 | `environments` |
| `GetWorkspaceDesigns` | 452 | 463 | 488 | `meshery_patterns` |
| `GetWorkspaceViews` | 618 | 627 | 654 | `meshery_views` |
| `GetWorkspaceTeams` | 746 | 755 | 782 | `teams` |

`GetWorkspaces` (line 47) uses `ApplyFilters` only, with no JSON decode, so it is consistent and not affected.

That the UI sends the JSON grammar is confirmed by `ui/components/environments/index.tsx:134`:

```js
filter: '{"assigned":false}',
```

### Expected Behavior

One grammar, decoded once. The `owner` filter should actually narrow the result set, and a malformed filter should be reported as a structured caller error rendered as HTTP 400 rather than surfacing a raw `encoding/json` message.

### Screenshots/Logs

Two environments assigned to one workspace, owned by two different users. Asking for only `ownerA`'s environments returns both:

```
filter={"assigned":true,"owner":"3ca03df3-2dc0-4b63-b8ea-00d47a4e16f8"}
  -> returned 2 environments (want 1, owner filter applied)
  - env-owned-by-b owner=62b2dd70-4415-41c0-9bb4-4d37b2ed2372   <-- should be excluded
  - env-owned-by-a owner=3ca03df3-2dc0-4b63-b8ea-00d47a4e16f8

--- FAIL: TestProofWorkspaceEnvironmentsOwnerFilterIsDead
    owner filter did not apply: got 2 environments, want 1
```

The other grammar errors out instead:

```
filter="owner abc" -> err=invalid character 'o' looking for beginning of value
```

### Environment

- **Host OS:** macOS (platform-independent — reproduces in a unit test)

### To Reproduce

Against an in-memory SQLite DB with `environments` and `workspaces_environments_mappings` migrated, seed two environments assigned to the same workspace under two different owners, then:

```go
filter, _ := json.Marshal(map[string]interface{}{
    "assigned": true,
    "owner":    ownerA.String(),
})

raw, err := wp.GetWorkspaceEnvironments(wsID, "", "", "0", "10", string(filter))
// err == nil, but the returned page contains BOTH environments —
// the owner predicate never reaches the query.

_, err = wp.GetWorkspaceEnvironments(wsID, "", "", "0", "10", "owner abc")
// err == "invalid character 'o' looking for beginning of value"
```

### Secondary finding — the allow-list names columns that do not exist

For `GetWorkspaceDesigns` the dynamic keys are not merely unreachable, they are wrong: the query runs against `meshery_patterns`, which has neither an `owner` nor an `organization_id` column. `AutoMigrate(&MesheryPattern{})` yields:

```
id name pattern_file location visibility catalog_data type source_content
updated_at created_at view_count share_count download_count clone_count
deployment_count workspace_id org_id
```

```
has "owner":           false
has "organization_id": false
```

`MesheryPattern.UserID` is `gorm:"-"` and `OrgID` is `db:"-"`, so neither is persisted. Had `ApplyFilters` ever been reachable there, it could only have produced a SQL error — the same situation #21833 found for `connections` and `organization_id`, and resolved by dropping the key rather than carrying it forward.

By contrast `environments` and `teams` do have a real `owner` column, so for those the filter is genuinely dead functionality rather than a latent SQL error.

### Suggested Fix

Mirror the approach taken in #21833: settle on the JSON grammar the UI already sends, decode it once into a typed filter, drop `ApplyFilters` from these paths, and per table keep only the keys that map to real columns. Reject a malformed filter as a structured caller error rendered as 400.

### Not a duplicate of

- **#21826** (`GetEnvironmentConnections`) — the same class of defect, but a different function in a different file. Its fix (#21833) touched only `environments_handlers.go`, `environment_connections_filter.go`, `environment_persister.go` and `error.go`; it did **not** touch `workspace_persister.go`, so these four call sites are untouched by it.
- **#21175** (`[UI] workspace.ts endpoints silently discarded by injectEndpoints`) — a UI-layer defect where locally declared endpoints collide with generated names. Its "owner resolution" is `expandUser`, which hydrates an owner *id* into a user profile for display (`getUserProfileById`); it is not a `WHERE owner = ?` predicate. Fixing #21175 would not make this filter work, and fixing this would not revive `expandInfo` or infinite scroll.
- **#17161** (`[mesheryctl] workspace use schema`) — a closed mesheryctl enhancement about schema-driven CLI types, unrelated to server-side filter parsing.

---

Gurkaran18:
**Notes for Reviewers**

- This PR fixes #21897
- Same defect and same fix shape as #21833 (`GetEnvironmentConnections`), ported to the four workspace-scoped list endpoints.

### What was wrong

`GetWorkspaceEnvironments`, `GetWorkspaceDesigns`, `GetWorkspaceViews` and `GetWorkspaceTeams` each fed one `filter` query parameter to two parsers with incompatible grammars:

1. `json.Unmarshal` into `WorkspaceFilter` — the JSON object grammar (`{"assigned":false}`)
2. `utils.ApplyFilters(query, filter, []string{"owner", "organization_id"})` — which does `strings.Split(filter, " ")` and reads a space-separated `"key value"` pair

Each grammar disabled the other:

- **JSON input** (what Meshery UI sends) reached `ApplyFilters` as a single token matching no dynamic key — a silent no-op, so the declared `owner` and `organization_id` filters were **unreachable**.
- **`ApplyFilters` input** (`owner <uuid>`) failed the JSON decode first, and the raw `encoding/json` message was returned to the caller.

`GetWorkspaces` uses `ApplyFilters` only, with no JSON decode, so it is consistent and deliberately left alone.

### What this changes

Settles on the JSON grammar — the one the UI already sends, so no client changes and no API version bump — and decodes it once in `ParseWorkspaceFilter`, which also rejects a top-level `null`: `json.Unmarshal` decodes that into a nil map without error, and every lookup on a nil map succeeds with the zero value, so it would otherwise fall through to the default filter.

- `owner` now applies as a real predicate on the two tables that carry the column, `environments` and `teams`.
- `owner` is deliberately **not** applied to designs or views. `meshery_patterns` has no such column (`MesheryPattern.UserID` is `gorm:"-"`, never persisted), and on `meshery_views` the field is named `UserID`, from which gorm derives `user_id` — the `db:"owner"` tag does not reach gorm. A `v.owner = ?` predicate could only have produced a SQL error.
- `organization_id` is dropped on all four. `meshery_patterns` has no such column; where it does exist (`environments`, `meshery_views`) the enclosing workspace already scopes the query, so the key adds nothing a caller could use.
- A rejected filter is the caller's mistake, so the four handlers answer **400** (new MeshKit code `meshery-server-1488`) instead of letting `providerStatus` blame a remote provider that was never asked.

### Error code coordination

Takes **1488** (`next_error_code: 1489`). Three branches were competing for the `meshery-server` counter: #21833 claims 1486, and #21879 has moved to 1487. 1488 is clear of both in any merge order, so the `component_info.json` conflict resolves by taking the higher counter, with no constant left to renumber afterwards.

`docs/data/errorref/` is not included: `error-codes-updater.yaml` only auto-commits on push to master, and on a PR it just gates on `errorutil analyze` being empty.

### Known pre-existing gap (not addressed here)

`view.View{}` and `team.Team{}` are not in `SystemDatabaseModels()` — only the workspace mapping tables are. So on the local provider those tables do not exist and these two endpoints cannot work regardless of this change. The new `t.owner = ?` predicate is therefore exercised by the unit tests rather than by a working local-provider path. Out of scope here; happy to file it separately.

### Testing

Tests cover the parser, the persister end to end, and the handler status. Both regressions were confirmed to fail against the previous code: the owner filter returned 2 rows where 1 was expected, `owner abc` and `{"assigned":"false"}` came back with error code `"None"` rather than a structured caller error, and `null` produced no error at all.

```
$ go test ./server/models/... ./server/handlers/...
ok  	github.com/meshery/meshery/server/models
ok  	github.com/meshery/meshery/server/handlers

$ golangci-lint run --config=.github/.golangci.yml
0 issues.

$ errorutil analyze -i ./server/helpers -o ./server/helpers --skip-dirs mesheryctl
(no output)
```

## Summary by CodeRabbit

* **New Features**
  * Added workspace listing filters for assignment status, deletion status, and owner.
  * Added clearer validation messages for malformed filters and unsupported value types.
  * Owner filtering is applied to workspace environments and teams.

* **Bug Fixes**
  * Invalid workspace filters now return HTTP 400 responses instead of being treated as provider errors.
  * Provider failures retain their existing status responses across workspace resource listings.


ritzorama:
The shared parser and focused environment/team owner predicates make this behavior substantially easier to reason about, and the 400-status coverage is especially helpful. Please regenerate the server error reference for the new structured error code so its remediation is available in the published documentation (`make error`, committing the relevant `docs/data/errorref/` output).,description:Comment on PR 21939,initial_wait:30}},{

Gurkaran18:
Thanks @ritzorama — done in 1c61b27.

Ran `make error` and committed the regenerated `docs/data/errorref/meshery-server_errors_export.json`. The diff is just the nine lines for `meshery-server-1488`, no churn elsewhere.

For context on how it was missed: `make error` writes to `server/helpers/errorutil_errors_export.json`, which is gitignored (`.gitignore:83`), so on a PR branch nothing visibly changes — the tracked reference under `docs/data/errorref/` is normally only refreshed by the error-codes updater on `master`.

Verified in the PR preview — the entry renders with its remediation:
https://docs.meshery.io/pr-preview/pr-21939/reference/references/error-codes/ (search 1488)

All checks are green. (The two red ones earlier were ~65s infrastructure failures on jobs that normally take 5 and 20 minutes; they're green on the current head.)

One pre-existing quirk visible in the preview: `errorutil` double-escapes embedded quotes, so the entry publishes as `Invalid \"filter\"` rather than `Invalid "filter"`. The Go source is correct (`[]string{"Invalid \"filter\" ..."}`) and nine existing entries in that file render the same way, so I've left the generated file alone rather than hand-editing it.

PR #21955, opened 2026-09-14
[Server] Batch MeshSync resync writes into a single transaction

### LINKED ISSUE #21954: MeshSync resync of one cluster serializes DB writes for all connected clusters via a shared global mutex
### Current Behavior

`MeshsyncDataHandler.persistStoreUpdate` and `meshsyncEventsAccumulator` (`server/models/meshsync_events.go:263-265` and `:218-220`) each wrap a metadata read plus a Create/Update/Delete in `mh.dbHandler.Lock()` / `defer Unlock()`. `dbHandler` is `database.Handler` from meshkit (`*gorm.DB` + `*sync.Mutex`, both pointers). Every connected cluster gets its own `MeshsyncDataHandler`, but `server/models/meshery_controllers.go:370` passes the same dbHandler, shared by the whole server, into all of them. Because the embedded mutex is a pointer, copying the struct by value does not give each cluster its own lock. They all share one.

`subsribeToStoreUpdates` (`meshsync_events.go:157-193`) is what runs on a cluster resync. It receives the full object list for that cluster and loops, calling `persistStoreUpdate` once for each object, synchronously, acquiring and releasing the shared mutex on every single object instead of once for the whole batch.

While that loop is running for one cluster, a live MeshSync event from any other connected cluster (`meshsyncEventsAccumulator`) blocks on the same mutex until the first cluster's entire object backlog finishes persisting.

Root cause: the lock is scoped to a single object instead of a whole batch, and it is shared across all clusters instead of being independent per connection. This turns a constraint meant only to keep SQLite writes to one at a time into full serialization across unrelated clusters.

### Expected Behavior

A resync triggered for one connected cluster should not stall live MeshSync persistence for other connected clusters beyond the time it takes to commit a single batched write.

### Screenshots/Logs

Not applicable. This is a concurrency defect identified through code inspection of the locking pattern, not a symptom with a log signature or UI screenshot. The relevant lock acquisition sites are cited above by file and line.

### Environment

- **Host OS:** Not OS specific, server side Go code path
- **Platform:** Docker or Kubernetes, affects any multi cluster deployment
- **Meshery Server Version:** master (verified against meshkit v1.0.22, the version pinned in go.mod)
- **Meshery Client Version:** Not applicable

### To Reproduce

1. Connect two Kubernetes clusters to the same Meshery Server.
2. Trigger a resync on cluster A (broker restart, operator restart, or a fresh connect with a large object count).
3. While that resync is persisting, emit or observe a real time change on cluster B.
4. Cluster B's event is not persisted until cluster A's entire resync backlog has been written, because both go through the same shared `dbHandler` mutex.

### Impact

In multi cluster deployments, any one cluster resyncing stalls MeshSync persistence for every other connected cluster for the duration of that resync. This gets worse as cluster count and per cluster object count grow, working against the multi cluster scaling MeshSync exists to support.

### Affected Components

`server/models/meshsync_events.go`, `server/models/meshery_controllers.go`

### Acceptance Criteria

- A resync of one cluster's object backlog no longer blocks live persistence for other connected clusters beyond a single batched write.
- A regression test with two `MeshsyncDataHandler`s sharing one `dbHandler` demonstrates bounded lock contention that does not grow with object count.
- No new SQLite "database is locked" errors are introduced.

### Solution Direction

Batch `subsribeToStoreUpdates`'s writes into one transaction per resync payload (a single lock acquisition, one bulk write), and narrow the critical section so it excludes the metadata read where that is safe.

bhuvan-somisetty:
Fixes #21954

### Problem

`MeshsyncDataHandler.persistStoreUpdate` locked and committed once per object during a cluster resync. `dbHandler`'s mutex is shared by every connected cluster's `MeshsyncDataHandler` (`server/models/meshery_controllers.go:370` passes the same handler to all of them), so a resync carrying thousands of objects held that shared lock, and paid SQLite's per-statement commit cost, once per object. Any other connected cluster's real-time MeshSync writes stalled for the whole resync.

### Root Cause

`subsribeToStoreUpdates` looped over every object in a resync payload and called `persistStoreUpdate` per object, each call independently locking, writing, and committing. The lock was scoped to a single object instead of the whole batch.

### Solution

Replaced `persistStoreUpdate` with `persistStoreUpdates`, called once per resync payload instead of once per object. It:

- Acquires `dbHandler`'s lock once for the whole payload instead of once per object.
- Wraps the batch in a single `dbHandler.Transaction(...)`, so the writes commit once instead of once per object.
- Keeps the existing per-object create-then-update-on-conflict semantics (a resync object whose derived ID already exists falls back to `Updates`, exactly as before).
- Keeps the existing best-effort behavior: an individual object's failure is logged and the rest of the batch still persists, rather than aborting the whole resync.

The live single-event path (`meshsyncEventsAccumulator`) is unchanged; only the resync batch path was affected by this issue.

### Testing

Added to `server/models/meshsync_events_test.go`:

- `TestPersistStoreUpdatesCommitsBatchInSingleTransaction`: registers a GORM callback that only counts a commit when that specific call actually opened its own transaction (verified against a real single-object `Create` as a sanity check first, to confirm the instrumentation itself is valid). Asserts a 3-object resync batch produces zero such per-object commits, i.e. everything went through the one outer transaction.
- `TestPersistStoreUpdatesPersistsAllNewObjects`: functional correctness for a batch of new objects.
- `TestPersistStoreUpdatesUpdatesExistingObjectAndContinuesBatch`: a resync object whose ID already exists updates in place (using meshsync's own ID-derivation scheme) and the rest of the batch still persists.
- `TestPersistStoreUpdatesEmptyBatchIsNoop`: an empty payload returns without acquiring the lock.
- `TestPersistStoreUpdatesTwoHandlersShareDBHandlerConcurrently`: two `MeshsyncDataHandler`s constructed with the same `dbHandler` (mirroring production's per-cluster construction) run concurrent resyncs; asserts no deadlock and no lost/corrupted writes.

I traced GORM's actual transaction-nesting behavior (`gorm.io/gorm@v1.31.2`, `callbacks/transaction.go` and `finisher_api.go`'s `Begin`) to confirm that a `Create`/`Updates` call nested inside an already-active `Transaction()` does not open or commit its own transaction, which is what the commit-counting test relies on.

**Local verification note:** this environment doesn't have a working C toolchain for `mattn/go-sqlite3` (cgo), so I could not execute these tests locally end to end. I confirmed `go build ./server/models/...` succeeds, `gofmt` is clean, and manually traced the GORM behavior the tests depend on. CI (which has gcc on its Linux runners) will run the full suite.

### Impact

Resync of one cluster's object backlog no longer blocks other connected clusters' real-time persistence beyond a single batched write, instead of blocking for the whole resync. No change to per-object business logic or the live single-event path.

## Summary by CodeRabbit

- **Bug Fixes**
  - MeshSync resource updates are now persisted in batches within a single transaction, improving consistency and reducing database overhead.
  - Registration updates are published only after successful persistence, preventing incomplete updates.
  - Metadata lookups now use the appropriate database transaction during batch processing.

- **Tests**
  - Added coverage for batch persistence, transaction handling, empty batches, concurrent handlers, and mixed success/failure scenarios.


bhuvan-somisetty:
Been fighting the "server Unit tests" job for a bit, it fails consistently (~14-16 min, exit 1) but I can't actually see which test is failing. The workflow pipes JSON test output straight into a file that never gets uploaded, so the console log just shows module downloads then "exit code 1", nothing else.

Tried a few fixes based on code review (batching the resync writes into one transaction, routing a metadata lookup through the active tx, loosening two test timeouts in case CI's parallel test execution under -race was starving them) but none stuck, so I'm probably fixing guesses instead of the real thing at this point.

@leecalcote could you or someone point me at the actual failure, or know another way to see server-test-output.json for this run? Would really appreciate it 🙏

https://github.com/meshery/meshery/actions/runs/34845783548/job/103982712332

yi-nuo426:
The batching path in `server/models/meshsync_events.go:260-325` correctly keeps component-metadata reads and resource writes on the transaction handle, then delays registration-queue sends until after `Transaction` succeeds. That preserves the previous per-object create/update behavior while avoiding both per-object commits and registrations for a rolled-back batch. The new conflict, empty-batch, and shared-handler concurrency coverage in `server/models/meshsync_events_test.go:526-749` validates the important cases. No change requested.

bhuvan-somisetty:
Thanks for the thorough review! Glad the batching and registration-timing logic checks out.

PR #21960, opened 2026-09-14
[Server] Return 400 for malformed event filter query parameters

### LINKED ISSUE #21878: [Server] Return 400 for malformed event filter query parameters

AnkushUjawane:
This PR fixes GetAllEvents (GET /api/system/events) so that a malformed filter query parameter (e.g. category=not-json instead of a JSON array) returns 400 Bad Request instead of being silently logged and ignored while the request continues on to query the provider anyway.

Before :-
GET /api/system/events?category=not-json
→ 200 OK (filter silently ignored)

After :-
GET /api/system/events?category=not-json
→ 400 Bad Request

Testing :-
Added two tests in events_streamer_test.go, built on the existing DefaultLocalProvider + in-memory SQLite fixture pattern used in delete_connection_test.go:

TestGetAllEventsMalformedFilterReturnsBadRequest — malformed category → expects 400.
TestGetAllEventsValidFilterSucceeds — well-formed category → expects 200 (guards against the fix over-rejecting valid requests).
$ go test ./server/handlers/... -run TestGetAllEvents -v
go: downloading github.com/stretchr/testify v1.11.1
=== RUN   TestGetAllEventsMalformedFilterReturnsBadRequest
--- PASS: TestGetAllEventsMalformedFilterReturnsBadRequest (0.00s)
=== RUN   TestGetAllEventsValidFilterSucceeds
--- PASS: TestGetAllEventsValidFilterSucceeds (0.00s)
PASS
ok      github.com/meshery/meshery/server/handlers      0.059s

Fixes #21878

Notes for Reviewers
This PR fixes #21878

## Summary by CodeRabbit

* **Bug Fixes**
  * Malformed event filters now return a clear **400 Bad Request** response instead of continuing with the request.
  * Valid event filters continue to return successful responses.


yi-nuo426:
The new early return in `server/handlers/events_streamer.go:58-62` correctly prevents `provider.GetEvents` from running after `getEventFilter` rejects malformed JSON, while `writeMeshkitError` gives clients the intended 400 response. The paired malformed and valid-category tests in `server/handlers/events_streamer_test.go:438-466` cover both sides of that contract well. No change requested.

AnkushUjawane:
Hi @yi-nuo426 , thanks so much for the detailed review! Just wanted to check in - Is there anything else needed from my side, or could this get a formal approval / be queued for merge?
Happy to make any further changes if needed. Thanks again for taking the time to review!

PR #22028, opened 2026-09-18
[Server] Reject path traversal and script content in registry SVG writes

### LINKED ISSUE #21966: [Server] Unauthenticated registry component registration writes arbitrary files to the filesystem (path traversal via Model.Name)
### Current Behavior

`POST /api/meshmodel/components/register` and `POST /api/registry/components` accept unauthenticated requests and allow an attacker to write files to the Meshery server's filesystem at arbitrary paths.

Routing (`server/router/server.go`):

```go
// Line 258 — no AuthMiddleware at all
gMux.Handle("/api/meshmodel/components/register",
    h.ProviderMiddleware((http.HandlerFunc(h.RegisterMeshmodelComponents)))).Methods("POST")
// Line 257 — marked NoAuth
handleRegistry("/components",
    h.ProviderMiddleware(h.AuthMiddleware(http.HandlerFunc(h.RegisterMeshmodelComponents), models.NoAuth)), "POST")
```

The handler `RegisterMeshmodelComponents` (`server/handlers/component_handler.go:771`) decodes an arbitrary `registry.MeshModelRegistrantData` body and, for `EntityType == ComponentDefinition`, calls `utils.WriteSVGsOnFileSystem(&c)` **before** any registry validation runs. `WriteSVGsOnFileSystem` (`server/helpers/utils/utils.go:232`) passes the attacker-controlled `comp.Model.Name` straight into `writeSVGHelper` (`utils.go:154`):

```go
func writeSVGHelper(svgColor, svgWhite, svgComplete string, dirname, filename string) ... {
    ...
    path := filepath.Join(UI, dirname, "color")
    err := os.MkdirAll(path, 0777)
    ...
    f, err := os.Create(filepath.Join(path, filename+"-color.svg"))
    ...
    _, err = f.WriteString(svgColor)      // attacker-controlled content
```

where `UI = "../../ui/public/static/img/meshmodels"` (`utils.go:151`). `dirname` is echoed verbatim from the JSON body, so `filepath.Join` does not prevent traversal: a `Model.Name` of `"../../../<anything>/"` escapes the UI directory. No path sanitization, no filename allow-list, no content validation, and (on the legacy route) no authentication whatsoever.

I confirmed the reachable call chain on master (`4706625001e`):

1. Any network caller (no cookie, no token) → `POST /api/meshmodel/components/register`
2. Body: `{"entityType":"ComponentDefinition","connection":{...},"entity":{...component definition with "model": {"name": "../../../tmp/pwn"} ...}}`
3. `RegisterMeshmodelComponents` → `WriteSVGsOnFileSystem` → `os.MkdirAll(".../../../../tmp/pwn/color")` + `os.Create(".../tmp/pwn/color/<kind>-color.svg")` with fully controlled content.

Even a non-traversing `Model.Name` writes attacker-authored `.svg` files under the UI's **served** static path (`ui/public/static/img/meshmodels/...`), which amounts to stored XSS against authenticated UI users (the SVG executes in the UI origin, where session cookies live).

### Expected Behavior

- `POST /api/meshmodel/components/register` must require `ProviderAuth` (or be restricted to loopback/operator-configured tokens) in every deployment mode, matching the sibling `POST /api/registry/...` write endpoints.
- `comp.Model.Name`/`comp.Component.Kind` (and the model-name directories) must be validated as slug-like identifiers; directory creation and file writes must stay confined to `UI`.
- SVG content should be validated/sanitized before being persisted into a server-served path.

### Screenshots/Logs

N/A (verified via source inspection of `server/router/server.go:256-258`, `server/handlers/component_handler.go:771-807`, `server/helpers/utils/utils.go:151-255`).

### Environment

- **Meshery Server Version:** master (`4706625001e`)
- **Meshery Client Version:** any

### To Reproduce

1. Start a Meshery server on an accessible host (default multi-provider mode: unset `PROVIDER`).
2. Without any session cookie or token, send:

```bash
curl -i -X POST http://<host>:9081/api/meshmodel/components/register \
  -H 'Content-Type: application/json' \
  -d '{
    "entityType": "ComponentDefinition",
    "connection": { "kind": "unknown" },
    "entity": {
      "apiVersion": "core.meshery.io/v1beta1",
      "version": "v0.0.1",
      "name": "Probe",
      "displayName": "Probe",
      "type": "Pod",
      "model": {
        "name": "../../../<writable-dir>/meshery-traversal",
        "displayName": "Probe"
      },
      "component": {
        "kind": "probe",
        "schema": "{}"
      },
      "styles": {
        "svgWhite": "<svg xmlns=\"http://www.w3.org/2000/svg\"><script>alert(document.domain)</script></svg>",
        "svgColor": "<svg xmlns=\"http://www.w3.org/2000/svg\"><script>alert(document.domain)</script></svg>",
        "svgComplete": "<svg xmlns=\"http://www.w3.org/2000/svg\"><script>alert(document.domain)</script></svg>"
      }
    }
  }'
```

3. Observe exit code `0`/`200` and that the directory `<cwd>/../<writable-dir>/meshery-traversal/color/` was created on the host with `probe-color.svg` written by the server process user.

### Suggested Fix

- Wrap the legacy route in `AuthMiddleware(..., models.ProviderAuth)` (`server/router/server.go:258`) and flip `/api/registry/components` POST from `NoAuth` to `ProviderAuth`, keeping GET read-only endpoints `NoAuth`.
- Sanitize `Model.Name`/`Component.Kind` (e.g., enforce `^[a-z0-9-]+$`) before `WriteSVGsOnFileSystem`, or rewrite the helper to take validated, flat directory names.
- Strip scripts/event handlers from SVG payloads or serve meshmodel assets with `Content-Disposition: attachment` / a `nosniff` + CSP header.

---

MU5A:
Fixes #21966

## The bug

`RegisterMeshmodelComponents` (backing both `POST /api/registry/components` and the legacy `POST /api/meshmodel/components/register`) accepts an unauthenticated request body and passes `Model.Name`/`Component.Kind` straight into `filepath.Join` with no validation, before writing the request's `SvgColor`/`SvgWhite`/`SvgComplete` content verbatim to disk. A `Model.Name` of `"../../../tmp/pwn"` escapes the intended `ui/public/static/img/meshmodels` directory, and SVG content containing a `<script>` tag is written into the UI's own served static path, a stored XSS against any authenticated UI user who later views that model.

Confirmed both against current master with the exact curl repro from the issue before writing any fix.

## The fix

Added two validators in `server/helpers/utils/utils.go`:
- `validateRegistrySVGPathComponent` rejects anything that isn't a plain alphanumeric slug. Real registrants only ever send names like `istio`, `kubernetes`, `cert-manager`.
- `validateSVGContent` rejects `<script>`, event-handler attributes, `<foreignObject>`, and `javascript:` URIs.

Both run before any filesystem operation.

Also rewrote the actual write path in `writeSVGHelper` to go through `os.Root` instead of bare `os.MkdirAll`/`os.Create`, matching the same symlink-safe, no-TOCTOU-window pattern this codebase already uses for reads in `SafeOpenFile` (`server/handlers/utils.go`). This is deliberate defense in depth on top of the validators above, not a replacement for them, confirmed by testing with the path validator disabled that `os.Root`'s own confinement independently caught the exact traversal payload from the issue.

`WriteSVGsOnFileSystem` now returns an error instead of silently discarding failures. Wired that through its three real callers:
- `RegisterMeshmodelComponents` (the endpoint from the issue) now rejects the request with 400 before any registration happens.
- The two lower-trust internal callers (Kubernetes-cluster component discovery, and ArtifactHub-generated components) log the failure and skip just that one item rather than failing the whole batch.

## What I deliberately did not change, and why

I did not touch the `NoAuth` routing on either endpoint, which the issue's own suggested fix proposes wrapping in `ProviderAuth`. I traced the actual caller before assuming that was safe: `meshery-adapter-library`'s `adapter/register.go` POSTs to this exact endpoint to register a service mesh adapter's components, and adapters have no user session or token available to them. The 2022 commit that added this route says so directly ("left open for adapters and external entities to register until another method of Auth is introduced"), and that method was never introduced. I grepped this org's repos for any adapter secret/token mechanism and found none.

Wrapping the route in `ProviderAuth` would break every current adapter's ability to register at all, a real regression, not a safe fix. That's a genuine, separate architectural gap (how should machine callers with no user session authenticate at all) worth its own issue and maintainer discussion, not something to decide unilaterally inside a bug-fix PR. Happy to file that as a follow-up issue if maintainers agree it's worth tracking separately.

## Test plan

- [x] Added 22 test cases in `utils_test.go`, including one that reproduces the issue's exact curl payload end to end and confirms nothing is written outside the confined directory, and one confirming a legitimate registration (plain slug name, clean SVG) still succeeds unchanged
- [x] Verified each new test actually depends on the fix: with the fix fully reverted the file does not even compile; disabling just the SVG content validator (independent of `os.Root`) makes the corresponding test fail with the exact payload from the issue
- [x] `go build ./server/...`: clean
- [x] `go vet ./server/...`: clean
- [x] `go test ./server/helpers/utils/...`: all pass, including pre-existing tests in the package
- [x] `go test ./server/handlers/...`: two pre-existing failures (`TestConnectionsAPISurfacesSeededSystemConnections`, `TestConnectionsAPISeededConnectionsSurviveRestart`) reproduce identically on a clean, unmodified `upstream/master` checkout and are unrelated to this change

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for uploaded registry icons to block unsafe SVG content and invalid file paths.
  * Invalid icon assets now return clear validation errors and HTTP 400 responses.
  * Server-side file and storage failures now return HTTP 500 responses instead of being reported as invalid uploads.
  * Components with invalid or unwritable icons are skipped and reported during model generation.
  * Valid SVG icons continue to be saved and associated with successfully registered components.


MU5A:
Went through all four, three were real and fixed, one I'm deliberately not adopting.

**component_generation.go (response accuracy)**: confirmed, `responseItem.Components` was set to the raw generated list regardless of which components actually got registered, so a rejected component (invalid icon asset) was reported back as if it had succeeded. Fixed by tracking only components that completed registration; a component that fails `RegisterEntity` itself is now excluded the same way, for consistency.

**component_handler.go / component_generation.go (400 vs 500)**: confirmed, `WriteSVGsOnFileSystem`'s error was always mapped to 400, conflating a rejected input with an actual filesystem/I/O failure. Added `InvalidRegistrySVGAssetError` as a distinct type the validators return, and `errors.As` at each call site, with a new `ErrWriteRegistrySVGAsset` (500) for the filesystem case.

**utils.go (obfuscation bypass)**: confirmed, and I checked this one concretely rather than taking it on faith, ran the old regex against `&#106;avascript:evil()` before changing anything and it returned no match, a real, working bypass. Replaced the raw-string regex with `containsDangerousSVGContent`, which parses the content as XML and HTML-unescapes every attribute value before checking it, closing the entity-encoding and whitespace-split obfuscation classes. It also rejects content that isn't well-formed XML at all rather than falling back to a weaker check, since a browser's SVG parser is far more forgiving of malformed markup than `encoding/xml`, and tolerating that gap would just reopen it.

I didn't go all the way to a full allow-list structural sanitizer as suggested. This codebase has no HTML/SVG sanitization library today, and the vectors this endpoint actually needs to close (script execution and script-like URIs) are a narrow, well-understood set, not the general problem of sanitizing arbitrary untrusted HTML. Happy to revisit if a maintainer wants that larger change.

**register.go (not adopted)**: propagating the error here to skip `RegisterEntity`/`count++` would mean a real Kubernetes CRD whose `Kind` name doesn't match the path-component slug pattern gets silently excluded from the registry entirely, not just missing its icon. The SVG content in this call site is always `models.K8sMeshModelMetadata.Styles`, a fixed server constant that can never trigger the content validator, so the only realistic failure here is the path check against real cluster data, and refusing to register that resource at all over an icon-naming quirk seems worse for the user than registering it without a nice icon. Left the existing log-and-continue behavior.

Added 9 new test cases covering the obfuscation payloads this round closes, all existing tests still pass.

MU5A:
Checked this one carefully before deciding, since it's a legitimate observation but I don't think it belongs in this PR.

Confirmed against upstream/master: the sequential-write-with-no-rollback behavior (each SVG variant written independently, nothing undone if a later variant fails) and the write-then-register-with-no-cleanup pattern (SVGs land on disk before RegisterEntity is even called, and were never removed if registration subsequently failed) both predate this PR entirely, verbatim in the original code. This fix didn't introduce either gap or make them worse. What changed here is that a component that fails validation or registration is now correctly excluded from the API response and, for the endpoint the security issue is actually about, the request is rejected before anything unsafe is ever written.

Making the whole write-then-register sequence atomic (stage writes, commit only after registration succeeds, roll back otherwise) is a real, reasonable thing to want, but it's a data-integrity change to an existing, unrelated part of this registration flow, not a security fix, and it's the "heavy lift" this review's own label says it is: it touches how every one of the three call sites writes and registers, not just the validation this PR adds. I'd rather not fold an architectural change like that into a security PR under review.

Happy to file it as its own issue if a maintainer agrees it's worth tracking separately.

Issue #22031, opened 2026-09-19
[UI] Unquoted YAML scalars are not syntax-highlighted


Ashutoshbind15:
### Current Behavior

In Meshery UI YAML editors, unquoted scalars are left in the default text color while the same value in quotes is highlighted as a string. Side by side, `name: foo` looks plain and `name: "foo"` looks styled, even though both are string values.

This shows up anywhere the shared YAML CodeMirror editor is used (designs, filters, adapter YAML, and similar).

### Desired Behavior

Quoted and unquoted YAML scalars should use the same string highlighting, so `name: foo` and `name: "foo"` read as the same kind of value. Keys, punctuation, and other token colors should stay as they are.

### Screenshots/Logs

<img width="1877" height="790" alt="Image" src="https://github.com/user-attachments/assets/a259a300-e656-4261-8f11-001c9e547679" />

Refer as to how the quoted strings, as dates are highlighted, whilst the unquoted ones look relatively unhighlighted/neutral in style.

### Environment

- Browser: Chrome
- Host OS: Linux
- Meshery Server Version: stable-v
- Platform: Docker

---

### Contributor [Guides](https://docs.meshery.io/project/contributing) and [Handbook](https://meshery.io/community#handbook)

- 🎨 Wireframes and [designs for Meshery UI](https://www.figma.com/file/SMP3zxOjZztdOLtgN4dS2W/Meshery-UI) in Figma [(open invite)](https://www.figma.com/team_invite/redeem/GvB8SudhEOoq3JOvoLaoMs)
- 🖥 [Contributing to Meshery UI](https://docs.meshery.io/project/contributing/contributing-ui)
- 🙋🏾🙋🏼 Questions: [Discussion Forum](https://discuss.meshery.io) and [Community Slack](https://slack.meshery.io)
- 📺 [Self-paced Contributor Trainings](https://meshery.io/talks-and-trainings#trainings)

Ashutoshbind15:
I've reproduced this locally in the shared YAML CodeMirror editor.

The highlighter treats quoted and unquoted scalars as different token types, and the current theme only colors the quoted ones. I can add a small highlight pass so unquoted scalars use the existing string color, with a unit test, without touching JSON mode or lint.

I've read the contributing guide and have a working local UI build. Kindly assign this to me.

pranitaurlam:
Hi  @aabidsofi19  @Yashsharma1911  @sudhanshutech
[...]

PR #22044, opened 2026-09-22
[UI] Color unquoted YAML scalars in shared CodeMirror editor

pranitaurlam:
## Summary
- The shared YAML CodeMirror editor (`ui/components/CodeMirror.tsx`) tags unquoted scalars as `tags.content`, which the `material` theme leaves unstyled, while quoted scalars are tagged `tags.string` and colored green. This made `name: foo` render as plain text next to `name: "foo"`.
- Adds a YAML-only `syntaxHighlighting` extension that colors `tags.content` the same green used for `tags.string`, so quoted and unquoted scalars read as the same kind of value. Keys, punctuation, and JSON mode/lint are untouched.

## Test plan
- [x] Added unit tests in `ui/components/CodeMirror.test.tsx` asserting the highlighting extension is added for `text/x-yaml` mode with the correct tag/color, and is absent for `application/json` mode.
- [x] `npx vitest run components/CodeMirror.test.tsx` — 14/14 passing.
- [x] `npx eslint components/CodeMirror.tsx components/CodeMirror.test.tsx` — clean.

Fixes #22031

## Summary by CodeRabbit

- **Improvements**
  - YAML editing now highlights unquoted scalar values in green, providing more consistent syntax highlighting with quoted strings.
  - JSON editing remains unchanged and does not apply the YAML-specific highlighting.


Salmaan-M (commented):
please sign your commit

